PR #155, opened 2016-05-09
make plot updates work outside ipython notebook

peendebak:
This PR makes plot updates work when working outside ipython notebook.

Current approach is to duplicate the functionality. Another option would be to perform the updates without the notebook functionality.  Note: using ipython is fine with me, but I would like to run my code outside the notebook.


giulioungaretti:
@MerlinSmiles @eendebakpt this is superseed by the new config options and #290 .

PR #255, opened 2016-07-08
Feature: AMI430 driver

Rubenknex:
Driver for the American Magnetics 430 power supply programmer

To do:
- [ ] Testing
- [x] Separate drivers for 2D and 3D magnet operation

@alexcjohnson @giulioungaretti


Rubenknex:
@damazter I implemented the changes we discussed in the last commit.

damazter:
@Rubenknex 
I will post things which are still unclear to me.
So far, it's only small things.
I am missing a docstring for the meaning of the input arguments of the AMI430 driver __init__ (and also for the 2D and 3D driver)

the `AMI430._ramp_to` command does not explicitely check any safety of the magnetic field if it is not part of a parent driver. It only passes the value trough the standard validator if you call the 
```
        self.add_function('ramp_to',
                          call_cmd=self._ramp_to,
                          args=[Numbers(-self._field_rating,
                                        self._field_rating)])
```
function, but not if the user calls `_ramp_to` is called directly. So if the magnet is not part of a larger system, there is still a maximum field for this magnet, which needs to be checked by the `_ramp_to` function.

damazter:
@Rubenknex 
I found another way to quench the magnet:
If you only connect 1 magnet to QCoDes (in systems with only one magnet axis for example)
Then you can call `_set_field` with any input arguments which are not checked (even if `perform_safety_check=True`). So there another check for the value must be added. (I think this check should be performed regardles of the value of `perform_safety_check`)

Rubenknex:
@damazter Yes you're right, I added some checks.

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -0,0 +1,618 @@
...
+        self._coil_constant = coil_constant
+        self._current_rating = current_rating
+        self._current_ramp_limit = current_ramp_limit
+        self._persistent_switch = persistent_switch
+
+        self._field_rating = coil_constant * current_rating
+        self._field_ramp_limit = coil_constant * current_ramp_limit
+
+        # Make sure the ramp rate time unit is seconds
+        if self.ask('RAMP:RATE:UNITS') == '1':
```
damazter: I think this should be 'RAMP:RATE:UNITS?' right?

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -0,0 +1,618 @@
...
+                               'manual up': 4,
+                               'manual down': 5,
+                               'zeroing current': 6,
+                               'quench detected': 7,
+                               'at zero current': 8,
+                               'heating switch': 9,
+                               'cooling switch': 10,
+                           })
+
+        self.add_function('get_error', get_cmd='SYST:ERR?')
```
damazter: get_cmd => call_cmd

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -0,0 +1,618 @@
...
+    one can set the _parent_instrument parameter of the magnet to None.
+    This is done at your own risk, as it skips field strength checks and
+    might result in a magnet quench.
+
+    Example of instantiation:
+
+    magnet = AMI430_3D('AMI430_3D',
+        AMI430('AMI430_X', '192.168.2.3', 0.0146, 68.53, 0.2),
+        AMI430('AMI430_Y', '192.168.2.2', 0.0426, 70.45, 0.05),
+        AMI430('AMI430_Z', '192.168.2.1', 0.1107, 81.33, 0.08))
```
damazter: the field limit is not in this example, which is a required positional argument for AMI430_3D

damazter:
@Rubenknex 
I made a few inline comments, but more generally.
I am a bit confused by your different coordinate systems. It seems that If I set the field in spherical coordinates, the cartesian coordinates are not updated.

I think that it will go wrong if I am wanting to set something in cylindrical coordinates and first set rho, and after that start sweeping over phi, right?

so I think it is important that the different coordinate systems are bound together in some way.

Rubenknex:
@damazter I think I understand what you mean. I made a bunch of changes to how the coordinates are handled internally. Now all the setpoint coordinates are stored internally (`self._x` etc).

Before, when setting a coordinate, the current values were measured and then adjusted. This would result in an incremental error during a loop when for example the magnet does not go fully towards its setpoint.

Now, the coordinates stored internally are independent from the 'actual' magnet fields, and for every `set` they are all updated.

The code could be cleaner however, I could not think of an easy way to have all the coordinate parameters interact such that they all update when one is changed.

damazter:
@Rubenknex 
I am going through the code now, maybe I can come up with something
Maybe some renaming of the coordinates would be appropriate of `field` to 'r' for example
and if phi can be doubly used for spherical and cylindrical, then `r`/`rho` could be used for both as well

damazter:
@Rubenknex 
if I understand correctly the `_update_internal_coords` updates all internal coordinates, based on the value of _x, _y and _z. I guess that this function should then be executed after every `_set_fields` or am I mistaken?

Maybe the code could be more clear if you would explicitely separate the setpoint and the real field values. Then the real field values of the driver could only be x, y and z, fixed as parameters of the driver and only gettable.
And then make a different parameter for the setpoint which is gettable and settable. The problems with coordinate systems would only enter in the setpoint_parameters.
Putting 'setpoint' in th ename of the variable would make it a lot easier to read I think.

I am still thinking about the coordinate system problem. It would be tempting to put the entire x, y, z setpoint parameters as a single tuple parameter, and r, theta, phi setpoint parameters as a single, different parameter. But I don't know if qcodes supports tupled parameters? @giulioungaretti

Rubenknex:
@damazter I agree on separating the actual field and setpoints into separate parameters, that would also make it a lot clearer to the user. But then I guess you would want this for all the parameters (including spherical and cylindrical) right?

If we do this then I think the `_update_internal_coords` being called will depend on what the expected behavior for invalid setpoint will be (probably just keep the old one).

About the tuple parameters, I don't know how that would work with loops. Like would you select one value of the tuple to loop over?

Rubenknex:
@damazter Also another thing, currently in the 3D driver when changing phi and one of the magnets will be out of range, it will still set the others. Maybe it should just stop with an error and not do anything then?

damazter:
Throwing an error earlier might be a good idea, so if you could make that, that would be great.
For the other post.
The actual field parameters only have to exist for x, y and z. I don't see an immediate reason to have all the others as well, because the user will be mainly interested in setting the setpoint and less with what the actual field is in terms of some different coordinate system. But I would not be opposed to adding the other coordinate systems to this.
For the setpoint it is a different matter and I think you will have to make a parameter for everyone of the nine coordinates.
tuples would be nicer, but I also don;t know about the loop integration of this. So we'd have to ask @giulioungaretti .
Another possibility might be to have the x1, x2 and x3 parameters in the driver, and then a single other parameter that defines the cooridnate system of the setpoint such as spherical, cylindrical, cartesian or even helical or weirder. That would make it more natural to distinguish between all the different systems.

In summary you would then have the parameters:
x
y
z
setpoint_r1
setpoint_r2
setpoint_r3
setpoint_coordinate_system (={CARtesian | SPherical | CYLindrical|...})

Then, upon changing the `setpoint_coordinate_system` you could change the units and labels of the coordinates r1, r2 and r3.

internally you might want to keep track of the setpoint in _x, _y and _z as well, but I think those shouldn't be QCodes parameters

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -0,0 +1,636 @@
...
+                                                                 self._field)
+
+        self._set_fields(self._x, self._y, self._z)
+
+    def _get_field(self):
+        return self._measure()['field']
+
+    def _set_field(self, field):
+        self._update_internal_coords()
+        self._field = value
```
damazter: `value` is undefined here, it should be `field`

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -0,0 +1,636 @@
...
+        self._coil_constant = coil_constant
+        self._current_rating = current_rating
+        self._current_ramp_limit = current_ramp_limit
+        self._persistent_switch = persistent_switch
+
+        self._field_rating = coil_constant * current_rating
+        self._field_ramp_limit = coil_constant * current_ramp_limit
+
+        # Make sure the ramp rate time unit is seconds
+        if self.ask('RAMP:RATE:UNITS?') == '1':
```
damazter: @Rubenknex 
is this a stable way of doing this?
I am having trouble with the line endings, (but that can be my fault) I couldn't find any documentation on how this instrument does its line endings.
So if my instruments sends back '1 \r\n', then this line will not match '1' so maybe `.startswith` would be better

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -0,0 +1,636 @@
...
+            self.write('CONF:RAMP:RATE:UNITS 0')
+
+        # Make sure the field unit is Tesla
+        if self.ask('FIELD:UNITS?') == '0':
```
damazter: same here

damazter:
@Rubenknex 
Trying to set a field gives an error:
`station.AMI430_3D.field(0.1)`
yields
```
d:\pycharmprojects\qcodes\qcodes\instrument_drivers\american_magnetics\AMI430.py:481: RuntimeWarning: invalid value encountered in double_scalars
  theta = np.arccos(z / field)
<ipython-input-3-d3a32f90175b> in <module>()
----> 1 station.AMI430_3D.field(0.1)

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in __call__(self, *args)
    133         else:
    134             if self.has_set:
--> 135                 self.set(*args)
    136             else:
    137                 raise NotImplementedError('no set cmd found in' +

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in _validate_and_set(self, value)
    869             e.args = e.args + (
    870                 'setting {} to {}'.format(self.full_name, repr(value)),)
--> 871             raise e
    872 
    873     def _sweep_steps(self, value):

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in _validate_and_set(self, value)
    861             clock = time.perf_counter()
    862             self.validate(value)
--> 863             self._set(value)
    864             self._save_val(value)
    865             if self._delay is not None:

d:\pycharmprojects\qcodes\qcodes\utils\command.py in __call__(self, *args)
    174             raise TypeError(
    175                 'command takes exactly {} args'.format(self.arg_count))
--> 176         return self.exec_function(*args)

d:\pycharmprojects\qcodes\qcodes\instrument_drivers\american_magnetics\AMI430.py in _set_field(self, field)
    586                                                                  self._field)
    587 
--> 588         self._set_fields(self._x, self._y, self._z)
    589 
    590     def _get_rho(self):

d:\pycharmprojects\qcodes\qcodes\instrument_drivers\american_magnetics\AMI430.py in _set_fields(self, x, y, z)
    628         # Finally, ramp up the coils that are increasing
    629         if not swept_x:
--> 630             self._magnet_x.field(x, perform_safety_check=False)
    631 
    632         if not swept_y:

TypeError: ("__call__() got an unexpected keyword argument 'perform_safety_check'", 'setting AMI430_3D_field to 0.1')
```

damazter:
@Rubenknex 
I my opinion, this driver can be a bit more agressive when exceeding the field limits, or having other errors.
Now it will only print to the console of the ipython notebook, but raising an error might be appropiate. (Now, you don't see anything in the ipython notebook)

damazter:
@Rubenknex 
There should be an easier way to set x, y and z simultaniously. now, when I do 
```
AMI.x(0.1)
AMI.y(0.1)
AMI.z(0.1)
```
I need to perform 9 sweeps. And even though 6 of them are trivial (they have the same start as ending point) they take a lot of time because of the sleeps in the `_set_field` function which is a waste of time.

damazter:
@Rubenknex 
I think I managed to quech the magnet again:
using multiprocessing (maybe that is cheating) you can start a sweep in x and y at the same time
all safety checks pass, because although the setpoints have been set, the actual field values are still zero.
so both magnet supplies are ramping to maximum at the same time.

this could be prevented by checking that none of the magnets is ramping when doing the field change. It is also possible that this not a scenario that is realistic. This is also only relevant if the AMI power control unit accepts multiple inputs which I don't know yet.

damazter:
@Rubenknex 
After you try to set an illegal value for one of the parameters you have to manually reset `_x, _y or _z` to be able to set the magnet again. This leads to another way to quench the magnet. By ramping x to 0.9 (with a maximal sphere of 1T) and then setting `_x=0`. Then calling `AMI.y(0.9)` will not see the security risk and ramp the y magnet.
I think this problem will be solved once all the stuff with the coordinate transformation and internal variables is solved. But just so you know what can lead to problems I have already included this in this discussion

qcodes/instrument_drivers/american_magnetics/AMI430.py
Rubenknex: This is something I didn't know at the time. I think it can be figured out by asking the power supply an `*IDN?` or any other command with the visa application.

Rubenknex:
@damazter Do you mean the `logging.error` calls don't show up in the notebook? Also raising an exception would probably stop your entire measurement (if you don't handle them), would this be wanted behavior?

The `_set_fields` function could probably check if the setpoints are identical to the last ones, preventing a pointless ramp. But maybe this is better to implement when we decide on how all the coordinate setpoints/current values should be represented. I can probably start implementing your idea of having separate parameters measuring (get only) and setpoints.

Rubenknex:
@damazter About the multiprocessing, it's probably a good idea to only have one magnet ramping at any time, which could be implemented easily I think by checking the state of the other two.

About the internal parameters, yes I think once we separate them into measured and setpoints, we don't have to store an internal state anymore.

qcodes/instrument_drivers/american_magnetics/AMI430.py
damazter: You're right that it can be figured out, but maybe it would be good to have the code a little bit more resilient to changes of the terminator. Some check that the answer is either ("0" or "1") instead of only checking =="1". If something would go wrong with this temrinator, the user will not find out I think if there is no check. If the magnet returns "0\r\n" suddenly, then this will leave the magnet with the wrong units. So that would be hard to debug

damazter:
@Rubenknex 
> @damazter Do you mean the logging.error calls don't show up in the notebook? Also raising an exception would probably stop your entire measurement (if you don't handle them), would this be wanted behavior?

The errors now don't show up in the notebook, which could be confusing. I am not opposed to just throwing an error. If you enter a wrong type (such as '0' instead of 0), an error is thrown, so throwing an error when the field is potentially exceded does feel natural to me.

> The _set_fields function could probably check if the setpoints are identical to the last ones, preventing a pointless ramp. But maybe this is better to implement when we decide on how all the coordinate setpoints/current values should be represented. I can probably start implementing your idea of having separate parameters measuring (get only) and setpoints.

Sounds good!

> About the internal parameters, yes I think once we separate them into measured and setpoints, we don't have to store an internal state anymore.

Same thing right? Again, sounds good!

> @damazter About the multiprocessing, it's probably a good idea to only have one magnet ramping at any time, which could be implemented easily I think by checking the state of the other two.

Yeah, we indeed don't want two simultanious sweeps in the near future (and possible long term future), It would be super hard to prevent quenches hwen doing that, so I agree with your solution

Rubenknex:
@damazter I implemented most of the changes.
- get-only `*_measured` parameters for returning a measured value
- Internal state is stored in `__x/y/z`, this makes the parameters inaccessible. I could not store the setpoints the way I wanted using only the  `Parameter` class.
- `_set_fields` is now responsible for performing all checks
- If one or more magnets will exceed its limits, the call is aborted
- New x/y/z values are checked for being equal with the `np.isclose` function, I don't know how accurate these magnets are, but by default it checks for differences larger than 1e-5 or 10 micro Tesla

The code has also become a bit cleaner.

damazter:
@Rubenknex 
I am not completely done with reading the driver, but some starting remarks
> Internal state is stored in __x/y/z, this makes the parameters inaccessible. I could not store the setpoints the way I wanted using only the Parameter class.

This is not exactly true, the parameters are still accesable, but the names are just mangled. But I guess that should be enough

> New x/y/z values are checked for being equal with the np.isclose function, I don't know how accurate these magnets are, but by default it checks for differences larger than 1e-5 or 10 micro Tesla

I think you should add quite some zero's there, just to be sure. And it should be possible to change this

Finally, I saw that you didn't go along with:
> Another possibility might be to have the x1, x2 and x3 parameters in the driver, and then a single other parameter that defines the cooridnate system of the setpoint such as spherical, cylindrical, cartesian or even helical or weirder. That would make it more natural to distinguish between all the different systems.
In summary you would then have the parameters:
x
y
z
setpoint_r1
setpoint_r2
setpoint_r3
setpoint_coordinate_system (={CARtesian | SPherical | CYLindrical|...})
Then, upon changing the setpoint_coordinate_system you could change the units and labels of the coordinates r1, r2 and r3.
internally you might want to keep track of the setpoint in _x, _y and _z as well, but I think those shouldn't be QCodes parameters

I personally think this might be easier than the system that is in place now, any thoughts?

Rubenknex:
@damazter 
Ah ok, I'm not sure about a  better solution, python is not good at hiding stuff from users haha.

I could add a parameter for the setpoint checking precision.

About your idea with reusing the parameters, I'm not sure which approach is better, but here are a few things:
- Say a user changes his coordinate system somewhere without setting it every time he sets a parameter. Now he might set an 'x' with an angle or something without this being immediately obvious.
- From the code it will not be very clear what is happening when there is `setpoint_r2(0.5)`
- I personally prefer an interface (what the parameters actually do) that is static instead of dynamic if it is not absolutely neccesary (for some intruments the state matters a lot, but here we can still do everything simultaniously)

But let me know what you think of this

damazter:
@Rubenknex 
You're right, hiding stuff is hard

parameters for precision checking would work
you know about ManualParameter which implements that?

Also I see that the _measure function only returns a single value. This is quite a waste of time if you wan't to know all three, isn't it?

line 593:
```python
    def _set_field(self, field):
        phi, theta, field, rho = self._get_non_cartesian()

        field = value
```
value not defined (i think the argument in the def should be changed form field to value)

The biggest problem I can think of now for not having the `r1, r2, r3` method is when sweeping, you will accumulate rounding errors when sweeping in spherical coordinates because you transform back and forth to cartesian instead of only going in one direction.
Implementing other coordinate systems is then also difficult, but i guess they should interface directly to `_set_fields` Maybe it would make sense making that method nonprivate.

Also, say you want to sweep two spherical coordinates at the same time, you can only make subsequent calls to change them. This means that the sweepingroute is far from optimal (take the case where in the simultaneous case, one of the cartesian coordinates is constant). A solution would be is that all methods in the driver only update the setpoint and that it will start ramping when you call some extra `ramp` function which then does the actual ramping of the field.
Then with the `r1, r2, r3` method, `r1, r2` and `r3` would just be manual parameters. This would make the code of the driver quite a lot shorter. I would even be in favor of making a single parameter `setpoint` that contains the 3-dimensional setpoint. Right now, the loop cannot really deal with that, but it will have to in the future so it wouldn't be that weird to already implement it in this driver. (Sorry @giulioungaretti )

If we stick with the current implementation, it would be sort of similar to what it now is, but it would be possible to have a more optimal path when changing two parameters simultaneously (without ramping two direction at the same time)

Rubenknex:
@damazter 

I also thought about the rounding errors, but a simple (not rigorous) test shows that sweeping phi over 360 degrees in 10000 steps gives only an error on the order of 1e-13 compared to the starting x/y/z components. Maybe a better test is needed, but I don't think this is a problem.

That's a nice idea, first setting setpoints and then ramping, it would make the code a lot cleaner and indeed solve most of the problems until now. One thing I can think of right now; how would one do a loop? You cannot do `Loop(magnet.phi[0, 180, 10])` anymore.

Rubenknex:
@damazter @giulioungaretti 

Tuple parameters have been added to the driver in the form of

`get() -> (x,y,z)`
`set((x,y,z))`

for all coordinate systems. The code also became a bit cleaner.

damazter:
@Rubenknex 
Two quick notes,
on line 616 there is a line `field = value` but value is not defined
secondly the `np.isclose` still needs to be changed such that it becomes more predictable what close means. A safe bet would be to set no relative_tolerance but only an absolute tolerance of 10^-8

damazter:
everytime the instrument is initialized I get the error:
```python
qcodes\qcodes\instrument_drivers\american_magnetics\AMI430.py:540: RuntimeWarning: invalid value encountered in double_scalars
  theta = np.arccos(z / field)
```

So I think it would make more sense to define `theta` to be `0` or something when there is no field

damazter:
The validator on the `cartesian` parameter makes it impossible to use the parameter, because the vector that you supply to the `cartesian` parameter is not a number:

```python
<ipython-input-5-e5ddc0628e4a> in <module>()
----> 1 station.AMI430_3D.cartesian((0.5,0,0))

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in __call__(self, *args)
    133         else:
    134             if self.has_set:
--> 135                 self.set(*args)
    136             else:
    137                 raise NotImplementedError('no set cmd found in' +

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in _validate_and_set(self, value)
    868             e.args = e.args + (
    869                 'setting {} to {}'.format(self.full_name, repr(value)),)
--> 870             raise e
    871 
    872     def _sweep_steps(self, value):

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in _validate_and_set(self, value)
    859         try:
    860             clock = time.perf_counter()
--> 861             self.validate(value)
    862             self._set(value)
    863             self._save_val(value)

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in validate(self, value)
    335             context = self.name
    336 
--> 337         self._vals.validate(value, 'Parameter: ' + context)
    338 
    339     def sweep(self, start, stop, step=None, num=None):

d:\pycharmprojects\qcodes\qcodes\utils\validators.py in validate(self, value, context)
    158         if not isinstance(value, self.validtypes):
    159             raise TypeError(
--> 160                 '{} is not an int or float; {}'.format(repr(value), context))
    161 
    162         if not (self._min_value <= value <= self._max_value):

TypeError: ('(0.5, 0, 0) is not an int or float; Parameter: AMI430_3D.cartesian', 'setting AMI430_3D_cartesian to (0.5, 0, 0)')
```

The same holds for `cylindrical` and `spherical`

damazter:
Unfortunately the default validator is Numbers, so I guess you have to set it to the `any` validator to circumvent the validator all together

damazter:
I am having a new problem and although I think I know what caused it, I am not fully sure how to solve it: The return value af `ask('QU?')` is parsed as an int but checked against string values.

Would it be possible to preparse everything to either a string or an int just to be sure what we are getting? (the same is true for the other parameters i guess)

It is strange that I have not seen this error before, I am not sure why that wouyld be.
Note that I am running agains a dummy version of the instrument, so it is a little hard to check that my dummy version works perfectly, but I think it would be good to build stability for `str` vs `int` anyway
error:
```python
KeyError                                  Traceback (most recent call last)
d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in _valmapping_get_parser(self, val)
    831             val = int(val)
--> 832             return self._get_mapping[val]
    833         except (ValueError, KeyError):

KeyError: 0

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
<ipython-input-5-e5ddc0628e4a> in <module>()
----> 1 station.AMI430_3D.cartesian((0.5,0,0))

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in __call__(self, *args)
    133         else:
    134             if self.has_set:
--> 135                 self.set(*args)
    136             else:
    137                 raise NotImplementedError('no set cmd found in' +

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in _validate_and_set(self, value)
    868             e.args = e.args + (
    869                 'setting {} to {}'.format(self.full_name, repr(value)),)
--> 870             raise e
    871 
    872     def _sweep_steps(self, value):

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in _validate_and_set(self, value)
    860             clock = time.perf_counter()
    861             self.validate(value)
--> 862             self._set(value)
    863             self._save_val(value)
    864             if self._delay is not None:

d:\pycharmprojects\qcodes\qcodes\utils\command.py in __call__(self, *args)
    175             raise TypeError(
    176                 'command takes exactly {} args'.format(self.arg_count))
--> 177         return self.exec_function(*args)

d:\pycharmprojects\qcodes\qcodes\instrument_drivers\american_magnetics\AMI430.py in _set_fields(self, values)
    668             swept_x = True
    669         elif np.abs(self._magnet_x.field()) < np.abs(x):
--> 670             self._magnet_x._set_field(x, perform_safety_check=False)
    671             swept_x = True
    672 

d:\pycharmprojects\qcodes\qcodes\instrument_drivers\american_magnetics\AMI430.py in _set_field(self, value, perform_safety_check)
    177             return
    178 
--> 179         if self._can_start_ramping():
    180             self.pause()
    181 

d:\pycharmprojects\qcodes\qcodes\instrument_drivers\american_magnetics\AMI430.py in _can_start_ramping(self)
    136         Check the current state of the magnet to see if we can start ramping
    137         """
--> 138         if self.is_quenched():
    139             logging.error(__name__ + ': Could not ramp because of quench')
    140             return False

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in __call__(self, *args)
    127         if len(args) == 0:
    128             if self.has_get:
--> 129                 return self.get()
    130             else:
    131                 raise NotImplementedError('no get cmd found in' +

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in get(self)
    809         except Exception as e:
    810             e.args = e.args + ('getting {}'.format(self.full_name),)
--> 811             raise e
    812 
    813     def _valmapping_get_parser(self, val):

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in get(self)
    804     def get(self):
    805         try:
--> 806             value = self._get()
    807             self._save_val(value)
    808             return value

d:\pycharmprojects\qcodes\qcodes\utils\command.py in __call__(self, *args)
    175             raise TypeError(
    176                 'command takes exactly {} args'.format(self.arg_count))
--> 177         return self.exec_function(*args)

d:\pycharmprojects\qcodes\qcodes\utils\command.py in call_by_str_parsed_out(self, *args)
    128     def call_by_str_parsed_out(self, *args):
    129         """Execute a formatted string with output parsing."""
--> 130         return self.output_parser(self.exec_str(self.cmd_str.format(*args)))
    131 
    132     def call_by_str_parsed_in(self, arg):

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in _valmapping_get_parser(self, val)
    832             return self._get_mapping[val]
    833         except (ValueError, KeyError):
--> 834             raise KeyError('Unmapped value from instrument: {!r}'.format(val))
    835 
    836     def _valmapping_with_preparser(self, val):

KeyError: ('Unmapped value from instrument: 0', 'getting AMI430_X_is_quenched', 'setting AMI430_3D_cartesian to (0.5, 0, 0)')
```

Rubenknex:
@damazter 
I am a bit confused about why this error occurs. The code in parameter.py looks like:

```
# Try and match the raw value from the instrument directly
        try:
            return self._get_mapping[val]
        except KeyError:
            pass

        # If there is no match, we can try to convert the parameter into a
        # numeric value
        try:
            val = int(val)
            return self._get_mapping[val]
        except (ValueError, KeyError):
            raise KeyError("Unmapped value from instrument: {!r}".format(val))
```

So apparently it's failing to do `self._get_mapping[val]` where `self._get_mapping={'0':False, '1':True}`. This is strange because `'0'` is being returned by the instrument right?

How are you testing it as a dummy instrument by the way?

damazter:
@Rubenknex 
My dummy instrument connects over TCPIP, so that is the same as the instrument intself. The instrument returns `'0\r\n'` but apperantly that ends up as being `0` in the driver. Wouldn't it be easiest to convert everything to a string before doing the value mapping. (Is there an option to do something like that in qcodes parameters?)

Rubenknex:
@damazter 
Can you check if changing the `terminator='\n'` to `'\r\n'` fixes it? Maybe your `'0\r\n'` is being stripped by qcodes to `'0\r'` which then cannot be found in the dict.

Does the magnet return it's messages with \r\n? If so and the above fixes it then I will change the terminator to that.

damazter:
@Rubenknex Indeed that sovled the problem. I changed the line terminator before, but I accidently overwrote it when pulling the new version of your driver. Stupid me.

would it be reasonable to check for this? But maybe this is also up to the user, to configure this correctly.
I think it would make sense to change the behaviour of the driver a little bit though. If terminator is given as a kwarg to the constructor, I think this given terminator should be used in favor of '\n' (which is not the case now, as you overwrite it when calling `super().__init)`)

Rubenknex:
@damazter 
Yes I think it's a good idea for the user to be able to change it, I updated it in the latest commit.

damazter:
@Rubenknex 
getting there, but I am still finding some errors:
I think the problem is somewhere in the `partial` function, but I am not sure

executing `station.AMI430_3D.field()`
gives
```python
KeyError                                  Traceback (most recent call last)
<ipython-input-6-1e06fb7e6062> in <module>()
----> 1 station.AMI430_3D.field()

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in __call__(self, *args)
    127         if len(args) == 0:
    128             if self.has_get:
--> 129                 return self.get()
    130             else:
    131                 raise NotImplementedError('no get cmd found in' +

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in get(self)
    809         except Exception as e:
    810             e.args = e.args + ('getting {}'.format(self.full_name),)
--> 811             raise e
    812 
    813     def _valmapping_get_parser(self, val):

d:\pycharmprojects\qcodes\qcodes\instrument\parameter.py in get(self)
    804     def get(self):
    805         try:
--> 806             value = self._get()
    807             self._save_val(value)
    808             return value

d:\pycharmprojects\qcodes\qcodes\utils\command.py in __call__(self, *args)
    175             raise TypeError(
    176                 'command takes exactly {} args'.format(self.arg_count))
--> 177         return self.exec_function(*args)

d:\pycharmprojects\qcodes\qcodes\instrument_drivers\american_magnetics\AMI430.py in _get_setpoints(self, *names)
    578 
    579     def _get_setpoints(self, *names):
--> 580         return self._from_xyz(self.__x, self.__y, self.__z, names)
    581 
    582     def _set_x(self, value):

d:\pycharmprojects\qcodes\qcodes\instrument_drivers\american_magnetics\AMI430.py in _from_xyz(self, x, y, z, *names)
    563         }
    564 
--> 565         returned = tuple(coords[name] for name in names)
    566 
    567         if len(returned) == 1:

d:\pycharmprojects\qcodes\qcodes\instrument_drivers\american_magnetics\AMI430.py in <genexpr>(.0)
    563         }
    564 
--> 565         returned = tuple(coords[name] for name in names)
    566 
    567         if len(returned) == 1:

KeyError: (('field',), 'getting AMI430_3D_field')
```

Rubenknex:
@damazter 
I made a small type, it should work now.

damazter:
@Rubenknex 
Thanks, that works now.
I found however that when initializing the driver, the setpoints are set to zero, but the magnet might not be. This leads to very strange behaviour as it becomes impossible to actually set the magnet to zero because of 
```
        if np.isclose(self.__x, x, rtol=0, atol=1e-8):
            swept_x = True
```
etc.

I think this can be solved by setting `__x`, `__y` and `__z` to the meaured field values at the initialization.
Problems will still arise when the magnet does stuff which the driver is unaware off, but I don't know if that can happen.

Rubenknex:
@damazter 
That's an interesting point, the values are now initialized and I added a method `update_internal_setpoints` that can be called when stuff has been changed manually.

damazter:
@Rubenknex 
nice
You forgot to define the magnets though:
```python
    def update_internal_setpoints(self):
        """
        Set the internal setpoints to the measured field values.
        This can be done in case the magnets have been adjusted manually.
        """
        self.__x = magnet_x.field()
        self.__y = magnet_y.field()
        self.__z = magnet_z.field()
```

damazter:
@Rubenknex 
could it be that this line of code is the wrong way around?
(similarly for `y` and `z`)
```python
        elif np.abs(self._magnet_x.field()) < np.abs(x):
```
(line: 697)

Rubenknex:
@damazter 
Ah that is indeed the case, I think I mixed up 'less than' and 'decreasing' or something.

damazter:
@Rubenknex 
can I propose to change
`phi, theta, field = values`
into
`field, theta, phi = values`
First of all, it would be nice to have a right-handed coordinate system.  and I think it is most natural to have r as the first coordinate.
Then I think the natural extension to cylindrical coordinates is
`rho, phi, z = values`
instead of
`phi, rho, z = values`

damazter:
@Rubenknex 
We had some problem with the order of the coordinates in spherical coordinates. I think that the order is different for setting and getting the spherical magnetic field, right?

Rubenknex:
@damazter Yes there were some unchanged lines from your last comment about changing the order. It should be fixed now.

giulioungaretti:
I suggest to, merge this. And open a new PR with things missing, fixes and so on!
It's a big PR with many comments, and possibly really hard to read in the future!
Let me know what you think!

damazter:
@giulioungaretti 
I would be all for it,

@Rubenknex 
I noticed that the ramprates are not set in the init of this driver. I think it would be wise to check if the ramprate is within the safe limits however, such that you are not accidnetly stuck with a very wrong ramping speed.

giulioungaretti:
@Rubenknex merge and open new ?

Rubenknex:
@giulioungaretti Yes that's fine by me, this PR is getting quite long haha

PR #278, opened 2016-07-22
do not load notebook elements when no notebook is running

peendebak:
When the user is not running a jupyter notebook as a frontend then the graphical elements based on the notebook should not be loaded.

It is not possible to detemine from python itself whether the user is running a notebook (or just an ipython terminal), so the function `frontend` uses an enviroment variable to overrule the default setting. The default setting is still notebook.

- Add `frontend()` function
- Do not load HiddenUpdateWidget when the frontend is not `notebook`

### DIFF
--- a/qcodes/__init__.py
+++ b/qcodes/__init__.py
@@ -33,7 +33,6 @@ if in_notebook():  # pragma: no cover
 
     # only load the show_subprocess_widget when running jupyter notebook
     if frontend()=='notebook':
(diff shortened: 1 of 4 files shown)


giulioungaretti:
@MerlinSmiles @eendebakpt  this is superseed by the new config options.

giulioungaretti:
@MerlinSmiles @eendebakpt this is superseed by the new config options.

PR #288, opened 2016-07-27
Ivvi/termchar

### LINKED ISSUE #276: pyvisa does not handle termination characters
Several instruments (for example the IVVI and the FPGA used at Delft) use a protocol without termination characters. The drivers try to handle this by setting VI_ATTR_TERMCHAR_EN to 0 (False), but this feature does not work. The results is that the instrument drivers fail at random times (e.g. when the measured data contains the termination character \n by coincidence).

The bug is not in qcodes, but in the visa lib itself. The issue is reported here for people who run into similar problems.

We have worked around the issue for now by modifying the drivers to read data one character at a time. This is less efficient, so we have not pushed the code to github.

The bug is also reported at the pyvisa github https://github.com/hgrecco/pyvisa/issues/225

eendebakpt:
Fixes #276

- Explicitly set `VI_ATTR_ASRL_END_IN` to 0 (false)
- Use multiple block read as a safety fallback

@AdriaanRol @giulioungaretti @alexcjohnson

### DIFF
--- a/qcodes/instrument_drivers/QuTech/IVVI.py
+++ b/qcodes/instrument_drivers/QuTech/IVVI.py
@@ -90,7 +90,7 @@ class IVVI(VisaInstrument):
 
         t1 = time.time()
 
-#         basic test to confirm we are properly connected
+        # basic test to confirm we are properly connected
         try:
             self.get_all()
         except Exception as ex:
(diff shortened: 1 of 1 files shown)


giulioungaretti:
Closing this as merged in #320 !

PR #320, opened 2016-08-30
Improved IVVI driver speed

Rubenknex:
- Removed checking for same DAC values in the `_set_dac`. At least for my machine the setting is fast enough that a check only causes an extra delay.
- Removed 50ms sleep in `_set_dac` as there seems to be no reason to sleep here
- Reduced the interval for checking the buffer in `read` from 50ms to 25ms as this (again on my machine) seems to always be completed in under 20ms.
- [ ] Test in an actual measurement

@alexcjohnson


giulioungaretti:
So @Rubenknex  @elrama- @AdriaanRol  @eendebakpt  this has 288  and it's ready for review and merge, correct ?

Rubenknex:
@giulioungaretti Yes, in principle it is. I just made some small additions but cannot commit them as I don't have access to this repository yet.

Also maybe @eendebakpt and/or others want to test this before merging. I tested it myself and everything worked under my test circumstances.

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -94,13 +110,24 @@ def __init__(self, name, address, reset=False, numdacs=16, dac_step=10,
...
+
+        # make sure we igonore termination characters
+        # http://www.ni.com/tutorial/4256/en/#toc2 on Termination Character
+        # Enabled
+        v.set_visa_attribute(visa.constants.VI_ATTR_TERMCHAR_EN, 0)
+        v.set_visa_attribute(visa.constants.VI_ATTR_ASRL_END_IN, 0)
+
         print('Initialized IVVI-rack in %.2fs' % (t1-t0))
 
     def get_idn(self):
```
AdriaanRol: I think it is not a good idea to remove this feature (which is part of the recent IVVI's in the base class. I think it is better to create a child class that has this overwritten specifically for if you have a legacy version IVVI. 

Would it be possible to change this before merging?

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -58,10 +60,24 @@ def __init__(self, name, address, reset=False, numdacs=16, dac_step=10,
...
                            get_cmd=self._get_version)
-        
+
+        self.add_parameter('check_setpoints',
+                           parameter_class=ManualParameter,
+                           initial_value=False,
+                           label='Check setpoints',
+                           vals=Bool())
+
+        self.add_parameter('dac_set_sleep',
```
AdriaanRol: This parameter has no units (I assume 's' but it could also be 'ms') 

I don't know if this has been implemented yet but at some point we discussed having some extra input for a docstring in the parameter @giulioungaretti has this been implemented? If so that would also be a good addition.

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -260,17 +338,15 @@ def read(self, message_len=None):
 
         while bytes_in_buffer < message_len:
             t1 = time.time()
-            time.sleep(.05)
+            time.sleep(.025)
```
AdriaanRol: Not a fan of hardcoded sleep times but I believe this is necessary for the IVVI.

qcodes/instrument_drivers/QuTech/IVVI.py
peendebak: Are there any IVVIs that use termination characters? If not, then let's keep these lines in the main driver

AdriaanRol: @peendebak , my comment relates to overwriting the `get_idn` function. AFAIK there are no IVVI's that use termination characters. That part should definitely remain in the main driver.

qcodes/instrument_drivers/QuTech/IVVI.py
peendebak: It is implemented, see for example the PR #335

peendebak:
@CJvanDiepen and I will test this as soon as we can. We have some serious issues with the driver related to threading, but are getting close to fixing these.

qcodes/instrument_drivers/QuTech/IVVI.py
Rubenknex: I already changed this but haven't been able to commit it to the new repo yet, both the 'set dac' sleep call and this one for waiting for the buffer are manual parameters.

AdriaanRol: > I already changed this but haven't been able to commit it to the new repo ye

@giulioungaretti , I was wondering, these PR's are based on branches but we don't have write access to the the new QCoDeS repo. How can @Rubenknex make changes to this branch in order to get it to merge?

giulioungaretti:
@peendebak @eendebakpt  this is ready to merge unless  there's more from your side ?

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -94,13 +131,24 @@ def __init__(self, name, address, reset=False, numdacs=16, dac_step=10,
...
             print(traceback.format_exc())
 
+        v = self.visa_handle
+
+        # make sure we igonore termination characters
+        # http://www.ni.com/tutorial/4256/en/#toc2 on Termination Character
+        # Enabled
+        v.set_visa_attribute(visa.constants.VI_ATTR_TERMCHAR_EN, 0)
+        v.set_visa_attribute(visa.constants.VI_ATTR_ASRL_END_IN, 0)
+
```
CJvanDiepen: And move the code which does the `get_all` to after the setting of the termination characters. First do the settings and then the communication.

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -247,6 +309,43 @@ def ask(self, message, raw=False):
         message_len = self.write(message, raw=raw)
         return self.read(message_len=message_len)
 
+    def _read_raw_bytes_direct(self, size):
+        """ Read raw data using the visa lib """
+        with(self.visa_handle.ignore_warning(visa.constants.VI_SUCCESS_MAX_CNT)):
+            mes = self.visa_handle.visalib.read(self.visa_handle.session, size)
+
+        return mes[0]
```
giulioungaretti: May want to add why use the item of the response !

giulioungaretti: @peendebak  that it's hard to say why you return mes[0].
The most pythonic way of doing this would be:
```
a_nice_name_that_explains _what_this_variable_is = mes[0] 
```

giulioungaretti: @eendebakpt @eendebakpt  easier to just write here in the comments.

eendebakpt:
@unga A PR with requested changes is #423

@AdriaanRol @Rubenknex I replaced the `_get_version` function with a dummy. Some of the older connection modules cannot handle this command.

brunobuijtendorp:
@unga This PR is closed, but not merged. Will another PR be opened?

giulioungaretti:
@brunobuijtendorp   no, I misread @eendebakpt  comment.

PR #325, opened 2016-09-06
feat: Snapshot value

nulinspiratie:
If a parameter contains a large amount of data, such as a trace acquisition parameter, it is often undesirable to save a full trace of data when creating a snapshot. Therefore I have added a snapshot_value kwarg to the Parameter init that can remove the parameter's 'value' from a snapshot.

Note that this is not achieved when providing the kwarg snapshot_get=False, since this only stops the parameter from performing a get when a snapshot is created.

- Add snapshot_value to the Parameter init kwargs.
- If param.snapshot() is called, and param._snapshot_value==False, remove 'value' from the state dictionary.
- If param.snapshot() is called and snapshot_value==False, do not perform a param.get() command, even if snapshot_get==True (since the value is not added to the snapshot).
- Small update to the snapshot_get description

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -142,6 +142,7 @@ class Parameter(Metadatable, DeferredOperations):
 
         snapshot_get (bool): Update the parameter before a snapshot.
             Set to false if it takes too long to update the parameter.
+            Note that snapshot_get is ignored if snapshot_value=False
 
         snapshot_value (bool): Store the parameter value in a snapshot.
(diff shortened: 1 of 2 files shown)


giulioungaretti:
@nulinspiratie if you fix the conflicts and or this is still relevant do let me know!

nulinspiratie:
@giulioungaretti the branch seems to have been deleted, not sure why. If you can keep this PR for a bit, I can make a new one.

nulinspiratie:
See #599 for an updated version

Issue #257, opened 2016-07-09
MatPlot bug


nataliejpg:
### Steps to reproduce
1. Do a Loop over a parameter (with not many steps)
   eg `data1 = qc.Loop(v1.power[1,5,1]).each(v1.trace())`
2. Plot data using MatPlot
   eg `qc.MatPlot(data1.magnitude)`
### Expected behaviour

Hence I expect a measurements at power = 1,2,3,4 (although I also think it's worth a discussion over whether this is a bit misleading and it should include 5 but I realise that's a deliberate data acquisition decision). The plot should reflect this and have each measurement 'sweep/point' centred on the corresponding power at which it was measured. 
### Actual behaviour

Data acquisition is fine and plots fine in QtPlot but MatPlot shifts the data so that the centre of my power=1 data is at 1.5 etc and then my power=4 data is thrown away. In other words the plot shows data from power = 1-4 but am actually plotting data from 1=3 and shifting it all up by half so on the power axis what is shown between 1 and 2 is my data from my measurment at power=1 and so on.

Here's my actual case: 
![capture](https://cloud.githubusercontent.com/assets/4841569/16709026/04c08c58-4605-11e6-9106-a8e5ae981669.PNG)
### System

T5 Windows (sorry haven't checked version)

ZNB20_2 branch (merged master 8th July morning)

giulioungaretti:
@jenshnielsen we did not fix this, right ?

Not sure it's a bug, rather a design flaw :P

PR #337, opened 2016-09-19
Matplotlib updates

nulinspiratie:
This pull request fixes some of the issues for plotting with MatPlot. One of the fixes solves issue #257.

- add a point to x- and y-axes such that datapoints are centered around their x and y values. This also stops the colorplot from ignoring the edge row and column.
- Add possibility of working with a 2D array of subplots. This can be used with the kwarg subplots=(nrows, ncols) upon instantiation. The subplots are flattened such that you always need a single int to be able to specify any subplot.
- Add default plot_kwargs_1D and plot_kwargs_2D to class. These are default kwargs that are always passed on when plotting, and can be overwritten by custom kwargs.
- Actually pass kwargs as kwargs to function _get_axes. I don't know why this was not done before
- Add nticks kwarg to _pcolormesh, which specifies preferred number of ticks. I did this because I sometimes ran into issues that for small ranges, there was only a single tick with label. Note that matplotlib actually refers to nbins, but nticks seems like a more suitable name in this case
- add use_offset for axes, default to False. When a range is reasonably small, matplotlib sometimes has a fixed offset, and only displays differences w.r.t. that offset as ticks. This can be quite confusing at times, so I have set the default behaviour to off.
- Scale the colors to the minimum and maximum data values.

Solving the x- and y-axes was not trivial, and so better solutions might well be possible. The biggest problem I ran into was that it had to deal with nan values as well. This method does deal with it correctly, but changes any 2D arrays into 1D arrays. There might be a way that would retain the 2D array, but I have not found one. Also, the current way that the function is called, it makes no difference.

### DIFF
--- a/qcodes/plots/qcmatplotlib.py
+++ b/qcodes/plots/qcmatplotlib.py
@@ -13,7 +13,8 @@ from .base import BasePlot
 
 
 class MatPlot(BasePlot):
-    plot_kwargs = {}
+    plot_kwargs_1D = {}
+    plot_kwargs_2D = {}
     """
     Plot x/y lines or x/y/z heatmap data. The first trace may be included
     in the constructor, other traces can be added with MatPlot.add()
@@ -174,7 +175,9 @@ class MatPlot(BasePlot):
         # didn't want to strip it out of kwargs earlier because it should stay
         # part of trace['config'].
         args = [arg for arg in [x, y, fmt] if arg is not None]
-        line, = ax.plot(*args, **kwargs)
+
+        full_kwargs = {**self.plot_kwargs_1D, **kwargs}
+        line, = ax.plot(*args, **full_kwargs)
         return line
 
     def _draw_pcolormesh(self, ax, z, x=None, y=None, subplot=1,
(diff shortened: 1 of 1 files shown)


qcodes/plots/qcmatplotlib.py
```diff
@@ -13,8 +13,8 @@
 
 
 class MatPlot(BasePlot):
-    plot_kwargs_1D = {}
-    plot_kwargs_2D = {}
+    plot_1D_kwargs = {}
```
giulioungaretti: Do we really want to have this a class attributes?

qcodes/plots/qcmatplotlib.py
```diff
@@ -57,8 +59,13 @@ def _init_plot(self, subplots=None, figsize=None, num=None):
...
             self.fig, self.subplots = plt.subplots(*subplots, num=num,
                                                    figsize=figsize)
-        if not hasattr(self.subplots, '__len__'):
-            self.subplots = (self.subplots,)
+
+        # Test if subplots is actually a single axis
+        if not isinstance(self.subplots, np.ndarray):
+            self.subplots = np.array([self.subplots])
+
+        # Flatten subplots in case it is a 2D array
```
giulioungaretti: Not sure I understand why you are flattening ?  :dango:

nulinspiratie:
Closed due to being replaced by #636

PR #347, opened 2016-09-30
Add function to instrument to validate all gettable parameters

peendebak:
This is useful when connecting to an instrument for the first time to check that all values are in the required range.

### DIFF
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -628,3 +628,12 @@ class Instrument(Metadatable, DelegateAttributes, NestedAttrAccess,
             'were trying to use a local instrument (defined with '
             'server_name=None) in a background Loop. Local instruments can '
             'only be used in Loops with background=False.')
+
+    def validate_status(self, verbose=0):
+        """ Validate the values of all gettable parameters """
+        for k, p in self.parameters.items():
+            if p.has_get and p.has_set:
+                value = p.get()
+                if verbose:
+                    print('validate_status: param %s: %s' % (k, value))
+                p.validate(value)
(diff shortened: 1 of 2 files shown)


qcodes/tests/test_instrument.py
```diff
@@ -53,7 +53,8 @@ def setUpClass(cls):
 
         cls.gates = MockGates(model=cls.model, server_name='')
         cls.source = MockSource(model=cls.model, server_name='')
-        cls.meter = MockMeter(model=cls.model, keep_history=False, server_name='')
+        cls.meter = MockMeter(
+            model=cls.model, keep_history=False, server_name='')
```
peendebak: My autopep8 formatter says its fine

qcodes/instrument/base.py
```diff
@@ -628,3 +628,12 @@ def __getstate__(self):
             'were trying to use a local instrument (defined with '
             'server_name=None) in a background Loop. Local instruments can '
             'only be used in Loops with background=False.')
+
+    def validate_status(self, verbose=0):
+        """ Validate the values of all gettable parameters """
```
peendebak: The `verbose` parameter is an int on purpose. In general I use the convetion: 0: no output, 1: normal output, >=2: used for debugging, output not stable

qcodes/tests/test_instrument.py
giulioungaretti: https://www.python.org/dev/peps/pep-0008/#indentation

giulioungaretti: cool

qcodes/instrument/base.py
giulioungaretti: a) Cool,  just write the assumption in the docstring ! 

b) In this function 2 and 1 are the same then, which I don't like, and I don't see it documented or understandable from the code. 
Change to true and false in this  case.

PR #358, opened 2016-10-21
Fix: immediately close socket if connection is refused

Rubenknex:
This is something I ran into when connecting to a python server script using `IPInstrument` and `persistent=False`. When the server script crashes for some reason and a new connection attempt is made, an error is thrown as expected. However when the server script is now up and running again, a new connection cannot be made as the call to `socket.shutdown(...)` in `_connect` throws an error as it is not connected to anything.

By immediately closing the socket, this is prevented. I'm not well versed in socket programming so there might be a better way to go about this.

- Close socket if a connection attempt is refused


giulioungaretti:
@Rubenknex  Can you confirm that if you close the socket manually then you can connect again ?

Rubenknex:
@giulioungaretti For me this fixes the error I had. This was when the server script raised an error during a connection. Now that I say this I see that handling ALL errors in the server script might also be a solution, making sure that connections are always closed gracefully.

PR #366, opened 2016-10-25
only load show_subprocess_widget when in notebook

### LINKED ISSUE #316: show_subprocess_widget causes problems outside the notebook
The `show_subprocess_widget` causes problems with systems not using ipython notebook (command line, spyder). A first approach to solving is was made in PR #278, but no definite conclusion was reached on how to solve it. There are a few options:

* Merge #308 (or some other config system) and then make all code loading the ipython widgets optional (with the default option: load the widgets, so there is no difference with current behaviour)
* Do not load widgets at all by default. This means users have to specifically load the widgets
* Use the PR #278 (with some polishing). In this PR a function `frontend()` is introduced that determined whether the widgets are loaded or not. The function itself check an environment variable to determine whether we are in notebook mode or not
* Remove the `show_subprocess_widget` entirely 

peendebak:
Fixes #316, #278

- Only load show_subprocess_widget when in notebook

@giulioungaretti

### DIFF
--- a/qcodes/__init__.py
+++ b/qcodes/__init__.py
@@ -34,6 +34,7 @@ if in_notebook():  # pragma: no cover
               'try "from qcodes.plots.pyqtgraph import QtPlot" '
               'to see the full error')
 
+if config['gui']['notebook']:
     from qcodes.widgets.widgets import show_subprocess_widget
 
 from qcodes.station import Station


giulioungaretti:
We need a bit more, so I will close this and open another that fixes all of this.

PR #369, opened 2016-10-27
Feat/validate status

peendebak:
New PR for validating the instrument values. (old PR is #347, that one now read-only).

@giulioungaretti


giulioungaretti:
@peendebak @eendebakpt  there seems to be some git-snafu ?
This PR is not using the branch that has this commit ab0a10297a3f3777c27f73025f30fe676321678e ?

Should I fix it ?

PR #375, opened 2016-10-30
Update docs of formatter, write metadata by default, pass arguments

peendebak:
Update from #346 

@giulioungaretti

### DIFF
--- a/qcodes/data/data_set.py
+++ b/qcodes/data/data_set.py
@@ -684,11 +684,14 @@ class DataSet(DelegateAttributes):
             return
         self.formatter.read_metadata(self)
 
-    def write(self):
+    def write(self, write_metadata=False):
         """
         Writes updates to the DataSet to storage.
         N.B. it is recommended to call data_set.finalize() when a DataSet is
         no longer expected to change to ensure files get closed
+
+        Args:
+            write_metadata (bool): write the metadata to disk
         """
         if self.mode != DataMode.LOCAL:
             raise RuntimeError('This object is connected to a DataServer, '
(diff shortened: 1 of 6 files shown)


giulioungaretti:
@AdriaanRol  wrote: 

I included write_metadata in the write function of the hdf5 formatter (this makes sense to me). The current implementation would result in writing twice (which should work correctly) however it introduces some overhead.
Just a suggestion, would it be a cleaner solution to have the .write of the default formatter also include the .write metadata? To me the separation of the metadata from the .write does not make sense. It is part of the data, not a distinct thing that needs to be handled separately

peendebak:
@giulioungaretti @AdriaanRol I can change the code such that the default `.write` always calls the `write_metadata`. Agree that this makes the most sense?

giulioungaretti:
let's hear @alexcjohnson  about this!

alexcjohnson:
> I can change the code such that the default `.write` always calls the `write_metadata`. Agree that this makes the most sense?

I agree that for manual usage this would be nice, but I wouldn't want `.write` to *always* call `write_metadata` - `.write` is meant to be incremental and as fast as possible so it can be called very frequently during acquisition, and metadata is nearly all set at the beginning, with only a little bit (ie the ending time) set at the end, and nothing during the acquisition. So perhaps a kwarg `write_metadata` that defaults to `True` but can be set `False` when you want to keep it separate?

peendebak:
@alexcjohnson @giulioungaretti Writing the metadata is optional now. 
@AdriaanRol Is the writing of the metadata called twice? If so, can you point me where this happens

peendebak:
@AdriaanRol I changed one of the arguments names of `.write_metadata` in the hdf5 formatter to make it consistent with the base class.

PR #379, opened 2016-11-03
Fix SR830

lukabavdaz:
Updated aux input/output and auto_offset to use the proper values
(the values written in the manual)
Added the function to get the ID.

@giulioungaretti

### DIFF
--- a/qcodes/instrument_drivers/stanford_research/SR830.py
+++ b/qcodes/instrument_drivers/stanford_research/SR830.py
@@ -267,8 +267,6 @@ class SR830(VisaInstrument):
 
         self.add_function('disable_front_panel', call_cmd='OVRM 0')
         self.add_function('enable_front_panel', call_cmd='OVRM 1')
-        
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -267,6 +267,8 @@ def parse_offset_get(s):
 
         self.add_function('disable_front_panel', call_cmd='OVRM 0')
         self.add_function('enable_front_panel', call_cmd='OVRM 1')
+        
+        self.add_function('id', call_cmd='*IDN?')
```
giulioungaretti: I guess  you want this just to get the full idn string, instead of the parsed one from get_idn ?

I kind of think it would be nicer if the name was idn, what  do you think ?

lukabavdaz: Actually, I will remove this function entirely.
We used it for testing purposes but it's no longer needed.

Issue #376, opened 2016-10-31
Auto saving plots.


giulioungaretti:
Some people in qdev would like to have this feature, so that they can immediately spot the data they are searching for.
Related to #158, but different because does not require extra stuff.
Not sure which approach I like the best for now.

AdriaanRol:
@giulioungaretti , The location_formatter I use saves datafiles all in their own folder (named with their timestamp). I think it makes sense to save a default plot (by default) in the location next to the corresponding dataset. This makes it very easy to browse around the data and quickly look at analyzed files. 

To clarify. 
I suggest, give a default location to the saveplot method that is next to the dataset.

PR #382, opened 2016-11-07
feature: Add png for jupyter nb and saving

### LINKED ISSUE #360: Save plot
Provide easy way to save plots as images both for qt and matplotlib.

giulioungaretti:
Fixes #360 and hints at an easy solution for #376 .

-  add _repr_png
-  add save method
-  add auto save feature to loop by expanding  ```with_bg_task ```

Refactoring QT in an idoimatic way will take a bit more time, this at least lessens the pain for say @nataliejpg  and some other experimentalist.


qcodes/plots/pyqtgraph.py
```diff
@@ -382,3 +382,19 @@ def _cmap(self, scale):
...
+        """
+        Create a png representation of the current window.
+        """
+        image = self.win.grab()
+        byte_array = self.rpg.QtCore.QByteArray()
+        buffer = self.rpg.QtCore.QBuffer(byte_array)
+        buffer.open(self.rpg.QtCore.QIODevice.ReadWrite)
+        image.save(buffer, 'PNG')
+        buffer.close()
+        return bytes(byte_array._getValue())
```
giulioungaretti: note to self: this is  a gnarly hack IMHO, but then again no other way around.

giulioungaretti:
@alan-geller  can you  look at https://github.com/QCoDeS/Qcodes/pull/382/commits/55eab2485ea5194b3f2938c6c6132267cd883446 and https://github.com/QCoDeS/Qcodes/pull/382/commits/dff292bb73f31f80e79f159a4b6131ab4242750c  , since it touches your code :D 

It just adds the option to run a task at the end of the loop or active loop , which is really useful to automatically save plots and sets the default refresh rate to around 100hz (which is kind of a basic golden standard, IMHO) !

qcodes/loops.py
```diff
@@ -249,22 +250,25 @@ def each(self, *actions):
         return ActiveLoop(self.sweep_values, self.delay, *actions,
                           then_actions=self.then_actions, station=self.station,
                           progress_interval=self.progress_interval,
-                          bg_task=self.bg_task, bg_min_delay=self.bg_min_delay)
+                          bg_task=self.bg_task, end_task=self.end_task, bg_min_delay=self.bg_min_delay)
 
-    def with_bg_task(self, task, min_delay=1):
+    def with_bg_task(self, task, end_task=None, min_delay=1):
```
alan-geller: @giulioungaretti I think this works fine, but I'm not sure the two features (background task and end task) should be joined at the hip like this. If we think of this end_task as specifically finishing whatever was done by the background task, which seems reasonable to me, then linking them is fine, but perhaps a more descriptive name for end_task would help make it clearer to the user. bg_final_task? bg_task_finish? end_bg_task?

qcodes/loops.py
```diff
@@ -263,7 +263,7 @@ def with_bg_task(self, task, end_task=None, min_delay=1):
             end_task: A callable object with no parameters. This object will be
                 invoked at the end  of the measurement loop.
 
-            min_delay (default 1): The minimum number of seconds to wait
+            min_delay (default 0.01): The minimum number of seconds to wait
```
alan-geller: I had originally set this to 1 because anything shorter caused the loop to slow appreciably when using matplot. With qtplot it's not a problem. I'm fine with being more aggressive with the default, but I think in the doc string perhaps we should suggest that users lengthen the min_delay if the background task is doing a lot of processing, such as updating a matplot graph.

qcodes/loops.py
```diff
@@ -249,22 +250,25 @@ def each(self, *actions):
...
+    def with_bg_task(self, task, end_task=None, min_delay=0.01):
         """
         Attaches a background task to this loop.
 
         Args:
             task: A callable object with no parameters. This object will be
                 invoked periodically during the measurement loop.
 
-            min_delay (default 1): The minimum number of seconds to wait
+            end_task: A callable object with no parameters. This object will be
```
alan-geller: I think the doc string here should clarify that end_task is intended to clean up after or otherwise finish the background task work, rather than being for general cleanup.

giulioungaretti: :+1:

qcodes/loops.py
giulioungaretti: That is indeed a problem with matplotlib, yes we should definitely add that to avoid nasty surprises!

PR #383, opened 2016-11-08
fix: Do not load widget if outside nb

giulioungaretti:
- do not load js widgets and css things as a side effect of the import of hiddenupdate widget 

@eendebakpt @peendebak  give it a shot


peendebak:
@giulioungaretti works fine here

PR #398, opened 2016-11-21
add handling of unknown types to JSON stream

peendebak:
@alexcjohnson @giulioungaretti

### DIFF
--- a/qcodes/tests/test_helpers.py
+++ b/qcodes/tests/test_helpers.py
@@ -9,7 +9,7 @@ from qcodes.utils.helpers import (is_sequence, permissive_range, wait_secs,
                                   make_unique, DelegateAttributes,
                                   LogCapture, strip_attrs, full_class,
                                   named_repr, make_sweep, is_sequence_of,
-                                  compare_dictionaries)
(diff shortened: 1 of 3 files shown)


peendebak:
@alexcjohnson Added the fix as suggested.

alexcjohnson:
Removing the `finally` was part of the suggested change (and the most important part, as it could lead to confusing errors). The other part is handling the two branches symmetrically. Right now `try` returns, but `except` sets `s`, which gets returned in the outer scope. Not a huge deal (so 💃 ), but really they should either both return or both set `s` for a later return.

Incidentally, we don't seem to have any tests for `NumpyJSONEncoder` do we? It's probably largely covered by way of metadata tests, but if you wanted to write a short test for it in `test_helpers.py` you would have my undying gratitude 🍻

peendebak:
@alexcjohnson @giulioungaretti Added a basic test

peendebak:
@giulioungaretti  @alexcjohnson  Even tough the :dancer: is there, I cannot merge myself any more

giulioungaretti:
@eendebakpt thanks for the PR, will prettyify the tests and merge!

PR #293, opened 2016-08-02
feature: ats9360 driver

nataliejpg:
Changes proposed in this pull request:
-  generalises ATS.py so it can be used with other alazar cards
-  adds specific driver for ATS9360

TODO:
- [ ] make use of the acquisition controller (currently commented out)
- [ ] check it hasnt broken it for the ATS9870
- [ ] make the aquire into a parameter so it can be used in a loop (?)
- [ ] remove debugging print statements and generally tidy up
- [ ] pep8 stuff (broken laptop -> lab computer and notepad++ editor)
- [ ] check some of the functions which I couldnt get to work (get_board_info and find_boards)
- [ ] call_dll error handling needs improving 

@giulioungaretti @alexcjohnson

PR #400, opened 2016-11-21
Driver/ats9360

nataliejpg:
This should replace [#293](https://github.com/QCoDeS/Qcodes/pull/293)

- make Alazar driver work for ATS9360 as well as ATS9870
- add acquisition controllers for arraylike alazar parameter for use with ATS9360 (but easily adaptable to other alazar cards)
- add example notebook 
- add aux.io functionality to alazar driver

- [x] I've written it such that the get_sample_rate function doesn't work for the ATS9360 and i'm not sure what the best way to implement this such that it works is (the issue is that the byte to value dict  of the sampling rate should be different depending on the value of another parameter which sets how it works out the sample rate, not sure the 'nice' way to write this')
- [x] how should different acq controllers be organised? all in one big file seems messy but also the way I've done it seems overkill, maybe just have two files in the master branch: the existing one by @damazter and one which works for the ATS9360 and returns an array 
- [x] currently the two acq controllers rec_samp and rec_buf don't work, they are designed to return a 2d dataset in one acquisition, I didnt try super hard to get them to work but also not sure why they don't, @alexcjohnson any ideas?
- [x] where should filters live considering that they are a quite general tool, also how do I deal with the whole import scipy thing...

@alexcjohnson @giulioungaretti @WilliamHPNielsen


damazter:
@nataliejpg 
A quick response to your message above
It would be great if this Alazar driver gets extended to work with other alazar cards as well!
I  have not taken the time yet to read all the changes proposed here. (I would like to check the changes to ATS.py)

> I've written it such that the get_sample_rate function doesn't work for the ATS9360 and i'm not sure what the best way to implement this such that it works is (the issue is that the byte to value dict of the sampling rate should be different depending on the value of another parameter which sets how it works out the sample rate, not sure the 'nice' way to write this')

I am not sure that I know what you mean, but with the ATS 9870 you had to take the decimation into account which is done in the `get_sample_rate` function in `ats.py` does this solve your problem or would a similar approach solve your problem?

>how should different acq controllers be organised? all in one big file seems messy but also the way I've done it seems overkill, maybe just have two files in the master branch: the existing one by @damazter and one which works for the ATS9360 and returns an array

and

> currently the two acq controllers rec_samp and rec_buf don't work, they are designed to return a 2d dataset in one acquisition, I didnt try super hard to get them to work but also not sure why they don't, @alexcjohnson any ideas?

My original idea was to not put any acquisition controllers in qcodes as it might by to personal from person to person. But I had to put at least one their to sort of explain the idea of the driver (making sure that ATS_WORKS.ipnb works)
The reason that the example does not return an array is that qcodes loops do not play nicely with these kinds of array parameters yet, so I decided to have an example which is not hacky in any way, but would not be really useful as a practical acquisition_contoller. (It only saves a single point per acquisition which is not that useful).

nataliejpg:
Hi @damazter, thanks for your comments. 

The reason the get_sample_rate function doesn't work for me is that when setting the sample rate the byte to value dict you use in the ATS9870 driver works fine if you aren't using a 10MHz reference signal to generate a sample rate from (which isn't an option for the ATS9870). In the 9360 however if you use the reference signal then you can set any sample rate you want and don't use a byte to value dictionary so unfortunately there can be two ways to map the same sample rate depending on if you use a reference signal or not. Effectively if you use a reference signal you don't want to use a byte to value dict you just send the int rate to the alazar. So no, what you suggest won't solve it but if I sit down and think a bit I can probably come up with a nice way to write it I just didn't have time yet so I left it as a to do. 

In terms of acquisition controllers the arraylike parameter works fine for us so far including int he loop (check out the Rohde Schwarz VNA driver whcih has been around a while by now). There are a few plotting bugs but the acquisition part itself is fine. Hence I think it's definitely worth having at least one acq controller which gets an array in qcodes for people to build off. Maybe just the 'basic_controller' I've written gives the idea and together with yours lets people write their own. Let me know what you think when you've had a chance to look. The ones I was having trouble with were the ones which are meant to return a data structure which is (samples * records) or (records * buffers), the ones which return an array of samples or an array of records are both there :)

damazter:
@nataliejpg 
> In terms of acquisition controllers the arraylike parameter works fine for us so far including int he loop (check out the Rohde Schwarz VNA driver whcih has been around a while by now). [...] The ones I was having trouble with were the ones which are meant to return a data structure which is (samples * records) or (records * buffers), the ones which return an array of samples or an array of records are both there :)

I'll check it out!

> Hence I think it's definitely worth having at least one acq controller which gets an array in qcodes for people to build off. Maybe just the 'basic_controller' I've written gives the idea and together with yours lets people write their own.

Yeah exactly, that one dataextractor example can definitely be improved upon. But having just `basic_acquisition_controller` as you suggested will prevent an entire mess of acquisition controllers which are all similar but not the same.

> The reason the get_sample_rate function doesn't work for me is that when [...] Effectively if you use a reference signal you don't want to use a byte to value dict you just send the int rate to the alazar. 

I will also think about that, maybe the ats9870 has the same option and the same problem. I'd have to look that up

nataliejpg:
@damazter 
acq conroller-wise: 
I think we should have at least 1 which just works out the box for ATS9360 and one for ATS9870 and one (could be one of these) which has an arraylike param. I think we're agreeing I just want to be sure :P 

get_sample_rate-wise:
I did look it up and unless I'm drastically misremembering the ATS9870 doesn't have that function and hence doesn't have the problem.

damazter:
@nataliejpg 
get_sample_rate-wise:
That solves a problem for me ;)

acq conroller-wise:
>I think we're agreeing I just want to be sure :P
I think so, I would just like it to be a single one, working with the 9870, 9360 and return an array like reuslt which works with the loop and dataset, such that the ATS_WORKS examplenotebook still works

nataliejpg:
@damazter 
The only problem with having just one acq controller is that from looking at yours I think the samples are arranged differently in the buffer. Not sure if that's an optional setting which differs between our setups or an ATS9870 feature but it would definitely be nice if people could just run the example notebook on their ATS card without having to edit the acq controller to do a basic test. Would be nicer not to have separate example notebooks but it seemed simpler, I guess I could check out trying to merge them and get yours to work for me (although for the data arranging reasons mentioned above I'd have to use one of my acq controllers rather than yours...). 

I'll leave it all up there for the moment and do a clean up and update when we have some feedback from one of the core devs and probably delete quite a bit of acq controller stuff at least. :)

nataliejpg:
I've left only the basic_controller which works with the ATS9360 and returns a trace of unprocessed data (averaged by records and buffers) and deleted the other acq controllers, maybe worth renaming the acq controllers to make it clear what each of them does, any strong feelings @damazter 

I've done a temporary fix so that get_sample now works for the ATS9360 at the expense of being able to set the clock source to be anything other than 10MHz reference signal. If anyone has an elegant way to fix this I'd be very interested, otherwise it's enough to work for our setup now and shouldn't break anything for the 9870 guys.

nataliejpg:
realised all other questions were specific to the acq controllers I've deleted...

jenshnielsen:
Notes to self:

Should probably revert https://github.com/QCoDeS/Qcodes/pull/400/commits/e8190ec03f5e98b824fe09d4d4d4e1ecbfca33b1 that seems to be unrelated and should if needed go in on a different branch

giulioungaretti:
This should be probably closed ?

jenshnielsen:
Yes closed as replaced by #491

PR #405, opened 2016-12-02
Feat: Jsonencoder addition

nulinspiratie:
Changes proposed in this pull request
- Adds optionality for an object to have a custom JSON encoder. The encoder checks if the object has a method _JSONencoder. If it does, this method is used to generate a JSON dict.

### DIFF
--- a/qcodes/utils/helpers.py
+++ b/qcodes/utils/helpers.py
@@ -34,6 +34,7 @@ class NumpyJSONEncoder(json.JSONEncoder):
                 'im': float(obj.imag)
             }
         elif hasattr(obj, '_JSONEncoder'):
+            # Use object's custom JSON encoder
             return obj._JSONEncoder()
         else:
             try:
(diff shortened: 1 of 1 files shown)


giulioungaretti:
:dango:   tahnks for the PR! Two things, why is this needed ? And why it's a private method you're calling ?

nulinspiratie:
Currently, when on object is JSON encoded for a snapshot, if it does not belong to a hardcoded list of encodable objects, its string representation is saved instead. However, if you want to include a snapshot of a custom object, you may want certain attributes to be saved as well. This is  currently not possible, but by adding the possibility of defining the `_JSON_encoder` function to the object.

As an example, I am using a PulseSequence object, which itself is composed of Pulse objects. I don't want to only save a string representation of the PulseSequence, but also want it to save all the pulses and their attributes, such as amplitude, frequency, start time, etc.

As for a private function, I thought that this function does not generally be executed except by the underlying snapshot function, which is why I made it private. Does this make sense?

PR #406, opened 2016-12-05
make docstring be formatted as rich text

peendebak:
In spyder the formatting of the generated `__doc__` of a `Parameter` is not according to rich text formatting. This commits fixes that, so that the docstring is renderer as rich text (tested with Spyder).

-  Add some newlines in the generated docstrings

@core


peendebak:
PR was merged in #431. I will close the PR.

PR #409, opened 2016-12-07
feat: Write metadata on write

giulioungaretti:
See #346

### DIFF
--- a/qcodes/data/data_set.py
+++ b/qcodes/data/data_set.py
@@ -610,14 +610,16 @@ class DataSet(DelegateAttributes):
                     time.time() > self.last_write + self.write_period):
                 self.write()
                 self.last_write = time.time()
-        else: # in PULL_FROM_SERVER mode; store() isn't legal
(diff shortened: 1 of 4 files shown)


giulioungaretti:
@eendebakpt  see if this works for you :space_invader:

peendebak:
@giulioungaretti I would make the writing of metadata optional in the `DataSet.write`, e.g. here https://github.com/QCoDeS/Qcodes/pull/409/commits/83bada893408d9a2f245ca8e04823712f863d018#diff-86f533fd540dcde5df3c6442c5897414R690

Also, why make a new PR? Is that because you cannot add to my PR easily, or some other reason?

In the docstrings I miss this change: https://github.com/QCoDeS/Qcodes/pull/375/files#diff-ca38cb5feb58d38c1bd62ba111978494R15. The metadata is JSON right now, but this is not specified, so there could be other implementations as well.

Besides these minor points, code looks good.

giulioungaretti:
@eendebakpt daamn, yea I will just use your pr! Let's ignore this one.

PR #410, opened 2016-12-08
Driver/awg5014

WilliamHPNielsen:
Fix: Make AWG5014 driver less of a bloody mess

- Remove some dead methods
- Add docstrings to all methods
- Add an example notebook

Changes planned for the near future:
 
- [x] Expand example notebook to show how to upload and load .awg files
- [x] Make SCPI command syntax uniform (use full command strings)
- [x] Delete more dead methods
- [x] Refactor a few set/get methods to be parameters
- [x] Make helper methods to ease the creation of .awg files

@giulioungaretti


qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -1,22 +1,3 @@
...
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program; if not, write to the Free Software
-# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
-
 import numpy as np
```
giulioungaretti: The imports do not follow pep8 :D

qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -1157,72 +1382,89 @@ def is_awg_ready(self):
...
+        """
+        original_timeout = self.visa_handle.timeout
+        self.visa_handle.timeout = 1000  # 1 second as VISA counts in ms
+        gotexception = False
+        while not gotexception:
+            try:
+                message = self.visa_handle.read()
+                if verbose:
+                    print(message)
+            except:
```
giulioungaretti: It would be nice to catch the right exception here!

qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -1157,72 +1382,89 @@ def is_awg_ready(self):
...
+            wfmname (str): waveform name
+
+        Raises:
+            Exception: if the lengths of w, m1, and m2 don't match
+            TypeError: if the waveform contains values outside (-1, 1)
+            TypeError: if the markers contain values that are not 0 or 1
+        """
         logging.debug(
-            __name__ + ' : Sending waveform %s to instrument' % wfmname)
+            __name__ + ' : Sending waveform {} to instrument'.format(wfmname))
```
giulioungaretti: Suggestion for logging:
use a module level logger as  in here:
https://github.com/QCoDeS/Qcodes/pull/416/files#diff-4c81a52a7bb6d513e5c5cd6c1b1f2c38R11

then all the calls to the logging system will include the module name which allows to tweak logging (say levels and handlers per module ) 👠

qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -787,367 +1055,324 @@ def generate_awg_file(self,
...
-            # logging.warning(__name__  + ' : Verkeerde lengte %s ipv %s'
-            #     %(self._values['files'][name]['numpoints'], self._numpoints))
+        return outdict
 
     def delete_all_waveforms_from_list(self):
+        """
+        Delete all user-defined waveforms in the list in a single
+        action. Note that there is no “UNDO” action once the waveforms
+        are deleted. Use caution before issuing this command.
+
```
giulioungaretti: trailing whitespace :P

qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -144,118 +117,145 @@ class Tektronix_AWG5014(VisaInstrument):
...
+                           get_parser=int)
+
+        # sequence parameter(s)
+        self.add_parameter('sequence_length',
+                           label='Sequence length',
+                           get_cmd='SEQuence:LENGth?',
+                           set_cmd='SEQuence:LENGth ' + '{}',
+                           get_parser=int,
+                           vals=vals.Ints(0, 8000),
+                           docstring=('This command sets the sequence length' +
```
giulioungaretti: We may want to use a single """ """" string here!

qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -341,328 +345,426 @@ def __init__(self, name, setup_folder, address, reset=False,
...
-        return self.ask('mmem:cdir?')
+        """
+        This query returns the current directory of the file system on the
+        arbitrary waveform generator. The current directory for the
+        programmatic interface is different from the currently selected
+        directory in the Windows Explorer on the instrument.
+
+        Returns:
+            str: A ""-encapsulated \n-terminated string with the full path
+                of the current folder.
```
giulioungaretti: This line should be indented a bit more, else the documentation breaks!
You  can check it by running the build doc on your machine as well!

qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -341,328 +345,426 @@ def __init__(self, name, setup_folder, address, reset=False,
...
     def get_sq_mode(self):
-        return self.ask('AWGC:SEQ:TYPE?')
+        """
+        This query returns the type of the arbitrary waveform
+        generator's sequencer. The sequence is executed by the
+        hardware sequencer whenever possible.
+
+        Returns:
+            str: Either 'HARD' or 'SOFT' indicating that the instrument is in
+                either hardware or software sequencer mode.
```
giulioungaretti: Same as above indent more!

qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -733,31 +846,186 @@ def generate_sequence_cfg(self):
...
+        been changed from their default value and put them in a
+        dictionary that can easily be written into an awg file, so as
+        to prevent said awg file from falling back to default values.
+        (See self.generate_awg_file and self.AWG_FILE_FORMAT_CHANNEL)
+        NOTE: This only works for settings changed via the corresponding
+        QCoDeS parameter.
+
+        Returns:
+            dict: A dict with the current setting for each entry in
+                AWG_FILE_FORMAT_HEAD iff this entry applies to the
```
giulioungaretti: Same as above

qcodes/instrument_drivers/tektronix/AWG5014.py
giulioungaretti: 👍

qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -310,29 +337,6 @@ def __init__(self, name, setup_folder, address, reset=False,
                     vals=vals.Numbers(-2.7, 2.7),
                     get_parser=float)
 
-        # # Add functions
```
AdriaanRol: What is the reason for removing the add_functions? As far as I understand this is a QCoDeS concept to easily list all the function (that are note parameters) in an instrument.

qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -341,328 +345,426 @@ def __init__(self, name, setup_folder, address, reset=False,
 
     # Functions
     def get_all(self, update=True):
```
AdriaanRol: Why not remove this (or at least add a depreciation warning) and let people use snapshot(update=True)

qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -341,328 +345,426 @@ def __init__(self, name, setup_folder, address, reset=False,
...
+            str: A comma-seperated string of the folder contents.
+        """
         if print_contents:
             print('Current folder:', self.get_current_folder_name())
-            print(self.ask('MMEM:CAT?')
+            print(self.ask('MMEMory:CATalog?')
                   .replace(',"$', '\n$').replace('","', '\n')
                   .replace(',', '\t'))
-        return self.ask('mmem:cat?')
+        return self.ask('MMEMory:CATalog?')
```
AdriaanRol: Why query twice? Is it not better to ask it once and then print if print_contents=True

qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -787,367 +1055,324 @@ def generate_awg_file(self,
...
-    def _set_setup_filename(self, fname):
-        folder_name = 'C:/' + self._setup_folder + '/' + fname
-        self.set_current_folder_name(folder_name)
-        set_folder_name = self.get_current_folder_name()
-        if not os.path.split(folder_name)[1] == os.path.split(set_folder_name)[1][:-1]:
-            print('Warning, unsuccesfully set AWG file', folder_name)
-        print('Current AWG file set to: ', self.get_current_folder_name())
-        self.write('AWGC:SRES "%s.awg"' % fname)
-
-    def set_setup_filename(self, fname, force_load=False):
```
AdriaanRol: I'm not sure if removing this is a good idea. Knowing about the hardware limitations (writing speed) it is sometimes very convenient to load a sequence that was previously put in memory of the 5014. I think this function can be cleaned up a bit but removing it completely may not be such a good idea.

qcodes/instrument_drivers/tektronix/AWG5014.py
WilliamHPNielsen: No good reason. I left it in case of compatibility issues, but I agree that it is a silly wrapper/renaming for a standard function.

qcodes/instrument_drivers/tektronix/AWG5014.py
WilliamHPNielsen: Yes, that is better. I didn't go through all the methods to see if they were smart, so there are probably (quite?) a few blunders left. I'll certainly fix this one.

CJvanDiepen:
@AdriaanRol @WilliamHPNielsen Just tested if the functionality of the AWG driver that we use (the `send_waveform_to_list` style) still works and it does! Only had to adjust the initialization because the revised driver does not work with a `setup_folder` anymore. Actually we did not use this but I also do not know if the functionality it offered was important, do any of you know?

AdriaanRol:
Hi William, it is quite important to be able to load files from the AWG
memory, as some take quite some time to generate. For this purpose the
setup folder sets what folder to look in. The name may not be the best but
we do use it. In the same line of thought, can we still load a file from a
filename?

qcodes/instrument_drivers/tektronix/AWG5014.py
WilliamHPNielsen: My idea was that it is better to have just one way of doing things. To load an awg file from memory (which is not a stupid thing to do for large files where file transfer times matter), one can/should now use `load_awg_file`. Do you think that `set_setup_filename` does something very different than `load_awg_file`? And if so, I would suggest to incorporate that into `load_awg_file`.

qcodes/instrument_drivers/tektronix/AWG5014.py
AdriaanRol: I was wondering what happened to this, is the add functions a deprecated qcodes concept?

giulioungaretti: @AdriaanRol  commented code was removed, and in many cases the functions where not even defined.  So in all effects this particular change is  a NOP 🎉 

Functions are still there as methods/parameters or as qcodes.functions where it makes sense to @WilliamHPNielsen (and also according to the documentation of the Function class),  see line 143.
I would say that many of the comment lines were pretty good candiate for being parameters.
example
```python 
# self.add_function('set_event_jump_timing')
# self.add_function('get_event_jump_timing')
```
became: 
```python
        self.add_parameter('event_jump_timing',
                           get_cmd='EVENt:JTIMing?',
                           set_cmd='EVENt:JTIMing {}',
                           vals=vals.Enum('SYNC', 'ASYNC'))
 ```
Hope this explains !

nataliejpg:
<3 especially love the example notebook, thanks William.

jenshnielsen:
replaced by #482

PR #412, opened 2016-12-09
R&S driver enhancement to support setting external reference

AdriaanRol:
Fixes no existing issue.

- Support setting the external refernce in the R&S sources

Tested and used on our setups. 

@giulioungaretti (is there a specific core developer responsible for drivers?)


qcodes/instrument_drivers/rohde_schwarz/SGS100A.py
```diff
@@ -2,6 +2,7 @@
 
 
 class RohdeSchwarz_SGS100A(VisaInstrument):
+
```
giulioungaretti: We don't use empty lines  here or after the docstring, your auto pep8 is probably doing it for you @eendebakpt  had the same issue. 

Also since you are here why not switch to proper triple double quotes ?  :heart:

AdriaanRol:
@giulioungaretti , I agree with the quotation marks, the whitelines etc. I'll try to fix it when I have time for it. Most likely somewhere this week.

nataliejpg:
Does anyone have any experience with using the r and s microwave source as an external generator for the network analyser? The newest vna driver I edited does this but I didn't work out how to enable an external reference for the r and s external generator as well as the vna internal generator itself.

nataliejpg:
Also did you guys have any problems with turning on and off the modulation of the r and s microwave source with this driver as we have had some, I haven't looked into it thoroughly though yet?

AdriaanRol:
@giulioungaretti 
Removed the blank line and switched to double quotes. Turns out it was indeed my autopep8. Do you know how to fix this?

giulioungaretti:
@AdriaanRol @eendebakpt // @peendebak  had the same issue! I don't use autopep8 so I am not sure how to help here. But AFAIK it's actually valid PEP8, it's just that our convention is not to put blank lines (which is also PEP8 compliant, yeah standards). 

Will look at it and merge!

giulioungaretti:
@nataliejpg @AdriaanRol  I will merge this! you can keep the conversation going on slack // open an issue for that <3

PR #416, opened 2016-12-12
feat: Switch to logging

giulioungaretti:
No more no mix between print and log!
The idea is to have every module exposing it's own log.
This way it's a lot easier to see, and in the future filter logs!

- [ ] check that no module level logger is missing
- [ ] make sure that nothing broke 
- [ ] add basic gui
- [ ] make it so qdev people can figure out how to use it

This allows to use real logging config:
https://docs.python.org/3/howto/logging.html#configuring-logging


bin/qcodes_logger_server
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python
+
+import logging
+import zmq
+
+
+def broker(frontend_addres="tcp://*:5559", backend_address="tcp://*:5560"):
```
MerlinSmiles: Maybe use a non-standard ports here? Just to avoid interference with other programs?
Ports from qcodes config? (I know this is independent of qc..)

bin/qcodes_logger_server
```diff
@@ -0,0 +1,51 @@
...
+
+    """
+    context = zmq.Context()
+    # Socket facing clients
+    frontend = context.socket(zmq.XSUB)
+    try:
+        frontend.bind(frontend_addres)
+        logging.info("XSUB listening at {}".format(frontend_addres))
+    except zmq.error.ZMQError:
+        logging.debug("Exiting. Broker is already running")
```
MerlinSmiles: Only if we assume this proker is the only one who would ever occupy that port 😮

qcodes/utils/zmqlogger.py
```diff
@@ -0,0 +1,122 @@
...
+        topic_list.append(record.levelname)
+
+        if topic:
+            topic_list.append(topic)
+
+        btopic = b'.'.join(cast_bytes(t) for t in topic_list)
+
+        self.socket.send_multipart([btopic, bmsg])
+
+def check_broker(frontend_addres="tcp://*:5559", backend_address="tcp://*:5560"):
```
MerlinSmiles: Also here, non standard ports?

bin/qcodes_logger_server
giulioungaretti: yes, it's tricky. Not sure there is another way though (or for now ).

bin/qcodes_logger_server
giulioungaretti: yes, still WIP. we can def use the config.

bin/qcodes_logger_server
giulioungaretti: well, coming back to this. It's going to work. Introducing real service discovery would be way to tricky given how dynamics things are.

If the ports are too standard or things breaks, we will fix as we go.

astafan8:
this is not relevant anymore, right? @WilliamHPNielsen

jenshnielsen:
I agree we are shipping logs via App insights

PR #421, opened 2016-12-19
fix: make sync a no-opt on local data

### LINKED ISSUE #380: Matplotlib can't handle more than 499 data points
Bug from Thorvald at qdev.

### Steps to reproduce
```python
# VNA power scan:
v1.spec_mode(0)
v1.bandwidth(500)
v1.avg(5)

a = 6.61064e9;
v1.start(a-20e6)
v1.stop(a+20e6)
v1.npts(101)

loop = qc.Loop(v1.power.sweep(-45,-55,1)).each(v1.trace)
data = loop.get_data_set();
plot=qc.MatPlot(array, subplots=(1,2))
plot.add(array, subplot=2)
loop.with_bg_task(plot.update, 0.0005).run()
```

### Expected behaviour
The plot should show the data

### Actual behaviour
The plot only show the data if a row has less than 499 points

### System
windows 

**qcodes branch**
master

**qcodes commit**
3475e49d95c587384102d754975a460089c5b75b

jenshnielsen:
fixes #380 by avoiding to overwrite data
with partial data from disk

- Make sync a no opt in local mode rather than read data from disc

I don't think that sync has any function in local mode but if that is too intrusive https://github.com/jenshnielsen/Qcodes/commit/fd2b0d3859627d882a9986f7aa453b980e60a9c6 is a less intrusive fix

@giulioungaretti @nataliejpg

### DIFF
--- a/docs/examples/Measure without a Loop.ipynb
+++ b/docs/examples/Measure without a Loop.ipynb
@@ -379,7 +379,8 @@
    },
    "outputs": [],
    "source": [
-    "from toymodel import AModel, MockGates, MockSource, MockMeter, AverageAndRaw, ArrayGetter\n",
+    "from toymodel import AModel, MockGates, MockSource, MockMeter, AverageAndRaw\n",
+    "from qcodes.instrument.mock import ArrayGetter\n",
     "\n",
     "# now create this \"experiment\", note that all these are instruments \n",
     "model = AModel()\n",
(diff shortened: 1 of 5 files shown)


jenshnielsen:
This breaks `test_local_instrument` because it assumes that you can:

* run experiment in background process
* end background process
* sync data over to main process

MerlinSmiles:
@jenshnielsen 

> ...
> run experiment in background process

Does that mean your small change there will make backround loops with the current multiprocessing stuff totally impossible?

jenshnielsen:
This change makes it impossible to read the data into the main process using `sync`. This is probably not the right fix but the current code is fairly convoluted. It's probably better to prevent plot.update from calling sync on default

jenshnielsen:
Test added

PR #423, opened 2016-12-19
updates to IVVI driver

eendebakpt:
- better variable name
- add term character setting;
- replace version function with dummy

@giulioungaretti

### DIFF
--- a/qcodes/instrument_drivers/QuTech/IVVI.py
+++ b/qcodes/instrument_drivers/QuTech/IVVI.py
@@ -31,7 +31,7 @@ class IVVI(VisaInstrument):
     Halfrange = Fullrange / 2
 
     def __init__(self, name, address, reset=False, numdacs=16, dac_step=10,
-                 dac_delay=.1, dac_max_delay=0.2, **kwargs):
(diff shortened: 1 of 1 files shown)


AdriaanRol:
@eendebakpt , Would it make sense to make a child class (in the same file), that does this overwriting of the get version command. Then we can still use it (we use it as a check to see if we are connected).

giulioungaretti:
This looks good to me, but @AdriaanRol and @eendebakpt  you have to agree on the version function! :P

eendebakpt:
@AdriaanRol @giulioungaretti A child class indeed seems like a good idea, or an option in the IVVI constructor. I will make something for it. Note: for the default I will implement the dummy command, as the command really mess up the connection module.

AdriaanRol:
@eendebakpt , I think that dummy indeed makes sense as a default, given how it messes up things. Looking forward to your input.

giulioungaretti:
great! Thanks for the contribution once again!

eendebakpt:
@AdriaanRol @giulioungaretti Here it is.

giulioungaretti:
LGTM!

AdriaanRol:
Let's 💃

PR #427, opened 2016-12-27
Feat: Slack bot integration

nulinspiratie:
This pull request adds a QCoDeS slack bot that provides information about current measurements.
The user can communicate with the Slack bot via instant messaging (see image below).
The slack bot currently has the following functionality:
- Give information about the current measurement (percentage done, print of dataset).
- Upload the most recent plot
- Notify when measurement is done
- Adding custom commands and periodically-executed tasks

A few notes
- I added the class attributes DataSet.latest_dataset and BasePlot.latest_plot. Every time a dataset/plot is instantiated, this attribute points to that object. This allows the slack bot to determine the latest dataset/plot.
- I added to slack bot to qcodes/widgets, but I was not sure if this is the right place. If not, where should it go?
- Periodic calls to `Slack.update()` must be made, and so the Slack bot works best as a widget in multiprocessing.
- The package 'tempfile' is used to determine a temporary store location for any plot file. This isn't currently in the package requirements

@giulioungaretti @AdriaanRol 

slack


giulioungaretti:
lol this is super cool!  May need more massaging though, as dataset.sync() may disappear soon, and same goes for the hiddenupdatewidget ! 

We can definitely work a way to insert the slack class into the code base,   I have the feeling that this should rather be a daemon-like process (such as monitoring) than a widget like thing :D 

Terrific contribution !

nulinspiratie:
@giulioungaretti I wasn't aware that dataset.sync() would disappear, will it not be necessary later on?

And I was initially also thinking of a subprocess, but with the current multiprocessing architecture I ran into problems with retrieving the latest plot. Also, I noticed some cool things you can do when qcodes is a widget instead of a subprocess. For instance, you can create a cell, fill it with a chunk of code, and then execute it. I'm using this to run measurements from Slack atm, although I'm not sure if this functionality should be part of the QCoDeS package

nulinspiratie:
@giulioungaretti I realized this code doesn't need widgets nor multiprocessing. Instantiating it in a different thread is sufficient. Would this PR be useful for qcodes? If so I can make the changes. What folder would now be a good location for the code?

giulioungaretti:
We can  definitely add this!  I would make an extra folder next to utils.

nulinspiratie:
@giulioungaretti I upgraded the Slack class, it now is a subclass of threading.Thread. The main advantage here is that it operates as a separate thread without depending on ipython widgets.

Starting the Slack client is very easy, namely as such:
```from qcodes.utils.slack import Slack
slack = Slack(**optional_commands)```

`optional_commands` are commands that can also be executed aside from the standard ones. For instance, it could be a parameter, which we can then perform get/set operations on via Slack.

Since the Slack class needs access to the latest measurement and plot, I also added some functionality to BasePlot and loops.py. BasePlot now has the class attribute `BasePlot.latest_plot`, which is updated every time a new BasePlot object is created. This way, `latest_plot` always refers to the latest created plot.

For loops I did something similar: I created ActiveLoop.active_loop. Using this, I added the two functions active_loop() and active_data_set(), which calls active_loop(). When running a loop with kwarg set_active=True (True by default), it will set ActiveLoop.active_loop to itself for the duration of the measurement. This also works with Measure. In the cases where you have a Loop in which a complex parameter itself creates a Loop or Measure, you can avoid overwriting active_loop by adding set_active = False.

I noticed that active_loop and active_data_set are also useful in other ways. For instance, you can use the folder of the active dataset as a base folder, and any parameters that create Loops can create subfolders in the base dataset folder. I added the two functions to qcodes init, so they can be accessed easier.

I finally also added slacker as an optional package in the setup

**Comment by giulioungaretti** [comment]
looks good to me! @jenshnielsen @WilliamHPNielsen  ?

**Comment by jenshnielsen** [comment]
Looks good to me

**Comment by giulioungaretti** [comment]
great will review the code and merge//ask for changes  today !

File: qcodes/utils/slack.py [hunk 1]
```diff
@@ -0,0 +1,418 @@
...
+import threading
+import traceback
+
+from qcodes.plots.base import BasePlot
+from qcodes import config as qc_config
+from qcodes.instrument.parameter import _BaseParameter
+from qcodes import active_loop, active_data_set
+
+
+def convert_command(text):
```
giulioungaretti: not sure why this? Isn't there one extra def () ?

qcodes/utils/slack.py
```diff
@@ -0,0 +1,418 @@
...
+    def try_convert_str(string):
+        try:
+            val = int(string)
+            return val
+        except:
+            pass
+        try:
+            val = float(string)
+            return val
+        except:
```
giulioungaretti: just catch:  ValueError here  :D

qcodes/utils/slack.py
```diff
@@ -0,0 +1,418 @@
...
+from qcodes.instrument.parameter import _BaseParameter
+from qcodes import active_loop, active_data_set
+
+
+def convert_command(text):
+    def try_convert_str(string):
+        try:
+            val = int(string)
+            return val
+        except:
```
giulioungaretti: just catch:  ValueError here  :D

qcodes/utils/slack.py
nulinspiratie: Not sure what you mean, the try_convert_str is a function that's only called within convert_command, so I made it a subfunction.

giulioungaretti: @nulinspiratie shut, never mind I totally misread the code.  Just catch the right Value error and then  I can merge!

PR #417, opened 2016-12-13
Update requirements and simlify readme

giulioungaretti:
The install for most people is described in the website.
Developers just need this info to install and run qcodes.
Agree?

PR #432, opened 2017-01-03
Travis: install sphinx for docs build

jenshnielsen:
Currently the docs don't build on travis because sphinx isn't installed. Fix that

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ script:
 
 after_success:
     # install codacy coverage plugin only on traivs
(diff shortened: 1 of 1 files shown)


giulioungaretti:
I think my previous cleanup #417  was to aggressive.

On a second though we'll also need

-sphinx-rtd-theme 
 -jsonschema 
 -sphinxcontrib-jsonschema

jenshnielsen:
Yes adding that now. Once the docs build more cleanly we should imho make failing to build the docs a hard failure on travis

PR #436, opened 2017-01-05
Test for DataSet

peendebak:
- This PR adds a test to make sure the `DataSet` can be used with a custom loop (e.g. without the qocdes `Loop`).
- It also autopeps some of the other test functions.

@unga @core


giulioungaretti:
Given the imment death of the old dataset, I will close it for now! Thanks for the contribution though.

PR #439, opened 2017-01-09
Travis: Ensure coverage is excluded with pytest

jenshnielsen:


jenshnielsen:
Merge as this is just fixing fallout from #434

PR #443, opened 2017-01-09
fix for hdf5 formatter metadata

### LINKED ISSUE #442: hdf5 formatter crashed for simple dataset
A minimal example:
```
import qcodes
import qcodes.tests.data_mocks
from qcodes.data.hdf5_format import HDF5Format
qcodes.DataSet.default_formatter = HDF5Format()

alldata=qcodes.tests.data_mocks.DataSet2D()
    
alldata.metadata={'a': ['hi', 'there'] }
alldata.write(write_metadata=True)
```

peendebak:
Fixes #442 

@unga @AdriaanRol @core

### DIFF
--- a/qcodes/data/hdf5_format.py
+++ b/qcodes/data/hdf5_format.py
@@ -280,6 +280,7 @@ class HDF5Format(Formatter):
                         elif isinstance(item[0], str):
                             dt = h5py.special_dtype(vlen=str)
                             data = np.array(item)
+                            data = data.reshape( (-1,1))
                             ds = entry_point.create_dataset(
                                 key, (len(data), 1), dtype=dt)
                             ds[:] = data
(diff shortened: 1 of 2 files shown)


jenshnielsen:
Can we change the example in #442 into a simple test for this?

peendebak:
@jenshnielsen Feel free to add the test, it should be a simple copy and past of #442.

giulioungaretti:
@peendebak  can I ask for an explanation of the crash and the fix ? It's not clear from the issue what's broken, and the same from the change  :D

peendebak:
A list is converted to an array of shape (k, ). This cannot be assigned to the hdf structure which is of shape (1,k). So we reshape the data so it can fit @unga

jenshnielsen:
Looks good to me

PR #444, opened 2017-01-10
fix argument type of hdf5 formatter write_metadata function

### LINKED ISSUE #440: hdf5 formatter does not implement proper interface for write_metadata
The `write_metadata` function has an argument `read_first`:

https://github.com/QCoDeS/Qcodes/blob/master/qcodes/data/format.py#L107

The HDF5 formatter does not have this argument:

https://github.com/QCoDeS/Qcodes/blob/master/qcodes/data/hdf5_format.py#L238

This generates an error here: 

https://github.com/QCoDeS/Qcodes/blob/master/qcodes/data/data_set.py#L764

A one line fix would be to add the `read_first` argument to the hdf5 formatter, but ignore it.

@giulioungaretti @AdriaanRol

peendebak:
Fixes #440

@unga @core

### DIFF
--- a/qcodes/data/hdf5_format.py
+++ b/qcodes/data/hdf5_format.py
@@ -235,7 +235,7 @@ class HDF5Format(Formatter):
 
         return dset
 
-    def write_metadata(self, data_set, io_manager=None, location=None):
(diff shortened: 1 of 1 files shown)


jenshnielsen:
Looks good, thanks

PR #445, opened 2017-01-12
Feature: hdf5 formatter with json metadata

peendebak:
This PR adds a new formatters that uses the HDF5 formatter as a base. The metadata is written in plain JSON. In this way the data read back from the metadata is identical to the original data (for all common data types).

- Write metadata as JSON
- Add generic test for all formatters

@AdriaanRol @unga @core


peendebak:
@AdriaanRol @jenshnielsen The build system is complaining about the type of the arguments of `write_metadata` in the HDF5Formatter. Here:

https://github.com/VandersypenQutech/Qcodes/blob/feat/hdf5json/qcodes/data/hdf5_format.py#L239

The `io_manager` and `location` have default value `None`, which is not the case in the parent class `Formatter`. Do we want to add `None` as the default value for all the other formatters as well?

jenshnielsen:
IMHO this is not a big issue as I understand it `location` and `io_manager` is required for the json metadata but not for the pure HDF5 formatter so the signatures are "correct" as is. Ways that this could be solved to not raise a style warning


* refactor the common parts of `HDF5FormatMetadata` and `HDF5Format` format into a common base class. 
* Change the inheritance order and make  `HDF5Format` a subclass of `HDF5FormatMetadata` pylint/codeacy seems to think it's fine with a subclass with less arguments

but I fail to see how this archives anything other than make the linter happy

peendebak:
@jenshnielsen I made the linter happy. Long term fix would be to specifiy the arguments of `write_metadata` and make the formatters obey.

PR #387, opened 2016-11-15
Feature/qt_plot_2

MerlinSmiles:
New PR from my fork so that we can continue to work on PR #290.

@peendebak @giulioungaretti

#######

Fixes the Qt part of #266 
Fixes #259 I think
Fixes the Qt part of #217
Fixes #34 since `setGeometry` is now directly available and `_repr_png_` exists
Fixes #289 I implemented it here too...

Changes proposed in this pull request:
- Use pyqtgraph Docks for subplots
- Expose some of the Dock-functionality to `QtPlot.add()`
- Add transpose argument to `QtPlot.add()` (flip x, y axes, and transpose z)
- Fix plotting when x, y, z datasets are defined
- Remove Javascript update widget which is only used for update timing, replaced with native Qt timer
- Add PNG representation of the plot windows
- Add plotting tutorial
- Add get default array title currently the location and array_id, maybe it should be reduced to the `name` and array_id, in the case where name exists. however, location is nice to later relate screenshots to the right dataset.
- Add get_units
- renamped `update` to `update_data` in `base.py` so it wont interfere with `update()` from the `Qwidget`
- removed pyqtgraph remote plotting stuff

I'm leaving for a holiday now, but it would be great if someone could check this over, and give some input

This PR is mostly meant to fix some of the urgent issues I came across, there is still a lot to do but for that we need to refactor the whole plotting system, which seems somewhat complicated right now. Make it more modular and allow for easy additions...

PR #446, opened 2017-01-12
Fix QtPlot

peendebak:
This PR takes elements from #387.

- Rename `windowTitle` argument to `window_title`
- Add setGeometry function

@core @unga @MerlinSmiles


jenshnielsen:
Added a check for the old argument name otherwise +1

PR #450, opened 2017-01-16
fix: Removed unused code

### LINKED ISSUE #447: Unused class FileWrapper
In the io manager module there is a FileWrapper class that seems to be never used anywhere?
Aside from the dubious utility of such a class (maybe if documented one could figure out) , maybe we should just remove it ?

@QCoDeS/core  ?

giulioungaretti:
close #447.

### DIFF
--- a/qcodes/data/io.py
+++ b/qcodes/data/io.py
@@ -28,13 +28,9 @@ The main thing these managers need to implement is the open context manager:
     - iter(): usually return self, but can be any iterator over lines
     - next(): assuming iter() returns self, this yields the next line.
 
-    (note: iter and next can be constructed automatically by FileWrapper
-    if you implement readline.)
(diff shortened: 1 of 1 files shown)


qcodes/data/io.py
```diff
@@ -28,13 +28,9 @@
...
     - next(): assuming iter() returns self, this yields the next line.
 
-    (note: iter and next can be constructed automatically by FileWrapper
-    if you implement readline.)
-
   In write or append mode:
     - write(s): add string s to the end of the file.
-    - writelines(seq): add a sequence of strings (can be constructed
-      automatically if you use FileWrapper)
+    - writelines(seq): add a sequence of stringso
```
jenshnielsen: Typo?

PR #451, opened 2017-01-16
feat: Add driver for QDac

WilliamHPNielsen:
Add a driver for the QDac and an example notebook explaining the parameters.

- A driver for the QDac is added to the driver collection
- An example notebook is added to the examples collection
- 

@giulioungaretti


qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -0,0 +1,473 @@
...
+                               get_cmd=partial(self.read_state, i, 'vrange'),
+                               vals=vals.Enum(0, 1)
+                               )
+            self.add_parameter(name='ch{:02}_irange'.format(i),
+                               set_cmd='cur ' + stri + ' {}',
+                               get_cmd=partial(self.read_state, i, 'irange')
+                               )
+            self.add_parameter(name='ch{:02}_i'.format(i),
+                               label='Current ' + stri,
+                               units='muA',
```
MerlinSmiles: Shouldnt we use Si units, and do any conversion later?

qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -0,0 +1,473 @@
...
+        Args:
+            name (str): The instrument name used by qcodes
+            address (str): The VISA name of the resource
+            num_chans (int): Number of channels to assign. Default: 48
+
+        Returns:
+            QDac object
+        """
+        super().__init__(name, address)
+        handle = self.visa_handle
```
MerlinSmiles: What is the benefit of using visa? cant we use serial communication directly?

qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -0,0 +1,473 @@
...
+                           vals=self.channel_validator)
+        # TO-DO: maybe it's too dangerous to have this settable.
+        # And perhaps ON is a better verbose mode default?
+        self.add_parameter(name='verbose',
+                           set_cmd='ver {}',
+                           val_mapping={True: 1, False: 0})
+
+        self.verbose.set(False)
+        self.connect_message()
+        print('[*] Querying all channels for voltages and currents...')
```
MerlinSmiles: print?

qcodes/instrument_drivers/QDev/QDac.py
WilliamHPNielsen: Uniformity, I guess... No deep reason for or against.

qcodes/instrument_drivers/QDev/QDac.py
WilliamHPNielsen: Yes, since this takes ~0.3 seconds per query and happens during initialisation.

qcodes/instrument_drivers/QDev/QDac.py
WilliamHPNielsen: I suppose we should... Will fix this.

qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -0,0 +1,473 @@
...
+
+
+class QDac(VisaInstrument):
+    """
+    Driver for the QDev digital-analog converter QDac
+
+    Based on "DAC_commands_v_13.pdf"
+    Tested with Software Version: 0.160915
+
+    The driver assumes that the instrument is ALWAYS in verbose mode OFF
```
WilliamHPNielsen: So, I talked to Rikke today, and she already has a firmware upgrade, meaning that we can implement smooth ramping next week. Then the driver can be properly finished and give a huge WARNING when the firmware version is off.

jenshnielsen: Super then lets wait until that is in place

jenshnielsen:
It seems like the notebook is still using the function generator which was removed in 85539f9

qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -0,0 +1,481 @@
+# QCoDeS driver for QDac. Based on Alex Johnson's work.
+
+import time
+import visa
+import logging
+import numpy as np
+
+from datetime import datetime
+from functools import partial
+from time import sleep
```
jenshnielsen: Does not match the only actual use of `sleep`  in` _wait_and_clear`

qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -0,0 +1,484 @@
...
+            raise ValueError('Channel number must be 1-48.')
+
+        if slope == 'Inf':
+            try:
+                sls = self._slopes
+                to_remove = [sls.index(sl) for sl in sls if sl[0]==chan][0]
+                self._slopes.remove(sls[to_remove])
+            # If the value was already 'Inf', the channel was not
+            # in the list and nothing happens
+        except ValueError:
```
jenshnielsen: Wrong indention?

qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -0,0 +1,484 @@
...
+        self.write('vol {} {}'.format(chan, switchint))
+
+        if xor(old, switchint):
+            voltageparam = self.parameters['ch{:02}_v'.format(chan)]
+            oldvoltage = voltageparam.get_latest()
+            newvoltage = {0: 10, 1: 0.1}[switchint]*oldvoltage
+            voltageparam._save_val(newvoltage)
+
+    def _num_verbose(self, s):
+        '''
```
jenshnielsen: Afaik doc strings should normally use `"""`

qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -0,0 +1,484 @@
...
+
+        old = tdict[self.parameters['ch{:02}_vrange'.format(chan)].get_latest()]
+
+        self.write('vol {} {}'.format(chan, switchint))
+
+        if xor(old, switchint):
+            voltageparam = self.parameters['ch{:02}_v'.format(chan)]
+            oldvoltage = voltageparam.get_latest()
+            newvoltage = {0: 10, 1: 0.1}[switchint]*oldvoltage
+            voltageparam._save_val(newvoltage)
```
jenshnielsen: Is there a way to avoid calling a private member of parameter directly?

qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -0,0 +1,484 @@
...
+            }
+
+            chans[chan - 1] = vals_dict
+            for param in vals_dict:
+                parameter = vals_dict[param][0].format(chanstr)
+                value = vals_dict[param][1]
+                if param == 'vrange':
+                    attenuation = 0.1*value
+                if param == 'v':
+                    value *= attenuation
```
jenshnielsen: It's not clear to me that `attenuation` is defined if param is `v` does `vrange` always come before v in vals_dict

qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -0,0 +1,484 @@
...
+            raise ValueError('unrecognized version line: ' + version_line)
+
+        header_line = self.read()
+        headers = header_line.lower().strip('\r\n').split('\t')
+        expected_headers = ['channel', 'out v', '', 'voltage range',
+                            'current range']
+        if headers != expected_headers:
+            raise ValueError('unrecognized header line: ' + header_line)
+
+        chans = [{} for i in self.chan_range]
```
jenshnielsen: Could probably be simplified to `[{}]*self.chan_range`

qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -0,0 +1,484 @@
...
+        if you want to use this response, we put it in self._write_response
+        (but only for the very last write call)
+
+        Note that this procedure makes it very cumbersome to handle the returned
+        messages from concatenated commands, e.g. 'wav 1 1 1 0;fun 2 1 100 1 1'
+        Please don't use concatenated commands
+        """
+        nr_bytes_written, ret_code = self.visa_handle.write(cmd)
+        self.check_error(ret_code)
+        self._write_response = self.visa_handle.read()
```
jenshnielsen: Should perhaps define _write_response as an empty string in `__init__` so it's always defined

qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -0,0 +1,484 @@
...
+
+        Args:
+            chan (int): The channel number (1-48)
+            slope (Union[float, str]): The slope in V/s. Write 'Inf' to allow
+              arbitraryly small rise times.
+        """
+        if chan not in range(1, 49):
+            raise ValueError('Channel number must be 1-48.')
+
+        if slope == 'Inf':
```
jenshnielsen: Should perhaps also allow for the inf float

`if slope == 'Inf' or type(slope) == float and math.isinf(slope)`

We could also only use the float inf but that is a bit tricky since you will have to use numpy as in `numpy.inf`

qcodes/instrument_drivers/QDev/QDac.py
WilliamHPNielsen: Agreed, that is a bit risky. I'll make it an OrderedDict .

jenshnielsen:
Merging @giulioungaretti's comments have been addressed too

PR #456, opened 2017-01-19
Docs: Change how automatic api docs are build.

jenshnielsen:
Some more small tweaks to the docs that I did at some point


jenshnielsen:
This significantly reduces the number of warnings from the docs build

Also fix a few other things here and there

PR #457, opened 2017-01-19
Fix/units deprecated

jenshnielsen:
Update most trivial uses of units to unit following the merge of #403


giulioungaretti:
lgtm

PR #459, opened 2017-01-19
Fix/current voltage parameter

jenshnielsen:
These parameters were missed in the original conversion

Also needed to change the SR560 driver default values to actually make it work


MerlinSmiles:
@jenshnielsen could you instead of changing the default value, add 'DC' to the cutoffs? That was my mistake, the instrument can actually turn off the filter and that is called DC on the instrument.

jenshnielsen:
Ok thanks for the update @MerlinSmiles

jenshnielsen:
@MerlinSmiles done

giulioungaretti:
LGTM

PR #460, opened 2017-01-20
Restore MercuryiPS driver

jenshnielsen:
This fixes the Mercury Driver to work with the new parameters class 

@alexcjohnson This relied on being able to set what is essentially MultiParameters. 
Is there a particular reason why that is not allowed


jenshnielsen:
Failing since there is an explicit test for set not being allowed

alexcjohnson:
Ah cool - didn't realize anyone had made a settable multi-valued parameter but I guess a magnet driver would be the place. Your warning is right: it's just that multi-valued set is incompatible with `Loop` / `Measure`. That's what the TODO immediately above it is about. If you want to alter or remove that test to get this passing, I can work on that TODO.

jenshnielsen:
My understanding is that these limitations are not new but were there already so I think converting it to a warning is ok for now and I will disable the test. Lets wait with the todo until we are a bit more sure about how we want to use it

PR #462, opened 2017-01-23
driver/cryocon26

Rubenknex:
Driver for the Cryocon 26 temperature controller


jenshnielsen:
Following the merge of #403 all standard parameters should use unit rather than units. Will still work but raise warnings. Otherwise looks good to me

PR #463, opened 2017-01-24
add abort function

peendebak:
Fixes issue mentioned on slack.

- Add an `abort_measurements` callback
- Use the callback in the qcodes Loop.

A way to use the callback:
```
#%% Load packages
import qcodes
import redis
import time

# connect to redis server
r = redis.Redis(host='127.0.0.1', port=6379)
r.set('myabort', 0)

# patch the qcodes abort function
def myabort():
    return int(r.get('myabort'))

qcodes.loops.abort_measurements = myabort


# run a loop
instr = DummyInstrument()
for ii in range(100):
    print('loop %d' % ii)
    value = instr.dac1.get()  # get parameter

    time.sleep(1.5)
    if myabort():
        print('abort the loop...')
        break
    
# the loop can be abort from a different process by calling    
# > r.set('myabort', 1)
```

@unga @jenshnielsen

### DIFF
--- a/qcodes/loops.py
+++ b/qcodes/loops.py
@@ -50,6 +50,8 @@ from datetime import datetime
 import logging
 import time
 import numpy as np
+import warnings
+import logging
 
 from qcodes.station import Station
 from qcodes.data.data_set import new_data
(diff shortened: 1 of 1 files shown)


giulioungaretti:
@eendebakpt @peendebak  I will close this let's talk more in #481  and #598.

The actual implementation will be more pythonic than this, allowing for composition instead.

So you will have Loop.withCancel("REDIS") or Loop.withCancel("SIGINT") or more.

peendebak:
@giulioungaretti Adding the notation `.withCancel` to the `Loop` (or any other similar construction) adds a lot of unnecessary code to each measurement done. A typical user will want do initialize the cancellation method only once.

giulioungaretti:
@peendebak that is kind of true, but having it in the source code  means people will have to edit that loops.py file. 

Also I think people will want to use wrappers around the basic building blocks, such that one can easily have very customised loops that require just a few lines of code to execute !

PR #235, opened 2016-06-11
Keithley2400 driver

ThomasKanne:
Changes proposed in this pull request:
- Made a Keithley driver for the 2400 model

@giulioungaretti @alexcjohnson @MerlinSmiles

PR #488, opened 2017-02-10
feat: Add basic driver for Keithley 2400

WilliamHPNielsen:
This is a clean-up of PR #235. I suspect buggy behaviour but have not tested the driver.

- Add a basic driver 
- No example notebook

@jenshnielsen


giulioungaretti:
@WilliamHPNielsen  I propose to merge this in anyway.
What is the bugginess you suspect?

WilliamHPNielsen:
@giulioungaretti Suspect is perhaps a strong word. But I have no way of knowing that the `msg` referenced in the parser functions has the right format (and that it ALWAYS has it) before I do/see a test.

PR #490, opened 2017-02-13
SGS100A Change lower limit of freq allowed to 1 MHz

jenshnielsen:
Cherrypicked from #400 where it doesn't belong @nataliejpg


jenshnielsen:
This matches the specs here https://www.rohde-schwarz.com/in/product/sgs100a-productstartpage_63493-9030.html?rusprivacypolicy=0 so looks correct to me

Issue #57, opened 2016-03-08
Conversion of input/output parameters


MerlinSmiles:
Related to isue #56 there is another problem, internal conversion of values.

Sometimes the data you get or set in a measurement is not what that data actually represents.
For instance:
- One might have a voltage divider at the output of a voltage source. Here I would like a conversion factor somewhere in the instrument that takes care of that. At the end of the day I'd like to say: `bias.set(1e-3)` but the instrument might have to output `1.213V`.
- There could be an analog instrument in from of the instrument you measure. For instance a Current amplifier that translates Amperes to Volts, When I measure the voltage this represents the current. At the end of the day I'd like to say `current.get()` and get something in Amperes.

Both cases deal with manual additions/changes to instruments, or rather the representation of their values.
Should we have, Is there, or is there planned to have a standardized conversion of this?

giulioungaretti:
@MerlinSmiles  can you rephrase this?

MerlinSmiles:
@giulioungaretti Not sure how to rephrase, what does not make sense?
Its kindof related to the signal chan thing from @WilliamHPNielsen

giulioungaretti:
I am confused by this "Sometimes the data you get or set in a measurement is not what that data actually represents. "

MerlinSmiles:
Lets say you have a DMM measuring the output of and current amplifier, you measure voltage but it represents current.
Right now I use i.e. the parameter that is inside the ithaco driver or the parameter that is inside the SR560 diver, but the whole conversion thing is much more general and IMHO we should not write the same thing for each situation.
Right now I use the ithaco thing for a totally different instrument.

jenshnielsen:
I think this is mostly solved by the parameter aliases in the station configurator i.e. bx in this example https://github.com/qdev-dk/qdev-wrappers/blob/master/examples/station_configurator/exampleConfig.yaml 
We plan to move the configurator into qcodes main

astafan8:
... or by [DelegateParameter class](https://github.com/qdev-dk/qdev-wrappers/blob/3bf23dcf5b5fd184e207ff80c1e1050c5c49ebc8/qdev_wrappers/parameters.py#L4) which will also moved to qcodes main soon (with some rework).

WilliamHPNielsen:
The future looks bright 😎

PR #492, opened 2017-02-14
feature: Add voltage divider device

giulioungaretti:


qcodes/instrument_drivers/devices.py
```diff
@@ -0,0 +1,81 @@
...
+        Returns:
+            number: value at which was set at the sample
+        """
+        return self.v1.get() / self.division_value
+
+    def get_instrument_value(self) -> Union[int, float]:
+        """
+        Returns:
+            number: value at which the attached paraemter is (i.e. does not
+                    account for the scaling)
```
jenshnielsen: This should un-indent to the level of number to render correctly

qcodes/instrument_drivers/devices.py
```diff
@@ -0,0 +1,81 @@
+from typing import Union
+
+from qcodes import Parameter, StandardParameter
+
+
+class VoltageDivider(Parameter):
+    """
+    Resitive voltage divider
+
+    To be used when you use a voltage divider to measure a parameter.
```
jenshnielsen: Are you measuring the parameter. The example sounds more like you are setting an output?

Output of dac is X -> Voltage divider -> x/10 to actual Experiment

giulioungaretti: I think it's fine like this, you can also get values.

giulioungaretti:
@QCoDeS/core  moar opinions, else I will just merge and let people try ?

jenshnielsen:
:+1: examples of using it both in input and output would probably make it more clear how it works but that can wait

MerlinSmiles:
Just want to throw two things in here:
https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/stanford_research/SR560.py
and #57 

Maybe its more general then adding devices for all kinds of conversions...?

jenshnielsen:
I think it makes sense to move the VoltageParameter and also the CurrentParameter from https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/ithaco/Ithaco_1211.py#L7 here to this location they dont really seem to belong in a specific instrument driver

giulioungaretti:
@jenshnielsen doc wise, I am not sure I understand. there are both sets and gets in the examples :D 
@MerlinSmiles @jenshnielsen  yeah it may make sense, but let's postpone the dicussion untill we have a working dataset. It's a lot better to repeat some code than to make the wrong abstractions IMHO :unicorn:

qcodes/instrument_drivers/devices.py
```diff
@@ -0,0 +1,85 @@
...
+            label=self.label,
+            unit=self.v1.unit,
+            metadata=self.v1.metadata)
+
+        # extend metadata
+        self._meta_attrs.extend(['v1', 'division_value'])
+
+    def set(self, value: Union[int, float]) -> None:
+        instrument_value = value * self.division_value
+        self._save_val(value)
```
jenshnielsen: what happens if instrument does not have a set function

giulioungaretti: well it's going to raise an exception :D as expected :D

giulioungaretti:
@jenshnielsen  more feedback, I would merge other wise :D

jenshnielsen:
IMHO the current documentation is still not very clear in the distinction between using it as an input or output device. I had to read it several times to understand how it was ment to be used in the two modes and what refers to what but perhaps it's more clear to others. Otherwise 👍

qcodes/instrument_drivers/devices.py
nulinspiratie: How about setting the parameter's has_set to equal that of the original parameter?

giulioungaretti: @nulinspiratie, right now this is the error you 'd get.
```python
/src/Qcodes/qcodes/instrument/parameter.py in no_setter(*args, **kwargs)
    676
    677 def no_setter(*args, **kwargs):
--> 678     raise NotImplementedError('This Parameter has no setter defined.')


NotImplementedError: ('This Parameter has no setter defined.', 'setting yolo_parabola to 0')

```
The difference why using  has_set is that one gets the real error i.e. that the underlying parameter has no set :D 

Does that make sense ?

nulinspiratie: Ah yeah that would give a clearer error message

Issue #494, opened 2017-02-14
Standardize unit


giulioungaretti:
Go over all instrument and set to the "correct value".
Example:  mV should be V and so on.
This includes converting when an instrument gets values in mV .

It's then up to the plotting library do to the scaling.

giulioungaretti:
this also implies removing units from the label

peendebak:
@unga I am not sure I like the idea of converting units to the "correct" value. What is correct then? For me the IVVI units in mV make sense, I dont want them in V.

AdriaanRol:
About a year ago we had a pretty similar discussion on the merits of this in issue #56 . I don't remember exactly what the conclusions were but it may be worth including the lessons from that discussion here.

giulioungaretti:
I think this will make some unhappy, but it's mostly unavoidable.
To have a uniform way of doing scaling and plotting, we need to pick a standard.
And it will be si and derived units !

PR #495, opened 2017-02-14
feature: Use unit in label

giulioungaretti:
Use units for plotting. Pyqtgraph uses units in a clever way! 

Needs #494  to be 100% great.

### DIFF
--- a/qcodes/plots/base.py
+++ b/qcodes/plots/base.py
@@ -81,7 +81,11 @@ class BasePlot:
                 into `x`, `y`, and optionally `z`, these are passed along to
                 self.add_to_plot.
                 To use custom labels and units pass for example:
-                    >>> plot.add(x=set, y=amplitude, xlabel=("set", "s"), ylabel= ("Amplitude", "V"))
+                    plot.add(x=set, y=amplitude, 
+                             xlabel="set"
+                             xunit="V",
+                             ylabel= "Amplitude",
+                             yunit ="V")
 
         Array shapes for 2D plots:
             x:(1D-length m), y:(1D-length n), z: (2D- n*m array)
--- a/qcodes/plots/base.py
+++ b/qcodes/plots/base.py
@@ -79,7 +79,9 @@ class BasePlot:
 
             kwargs: after inserting info found in args and possibly in set_arrays
                 into `x`, `y`, and optionally `z`, these are passed along to
-                self.add_to_plot
+                self.add_to_plot.
+                To use custom labels and units pass for example:
+                    >>> plot.add(x=set, y=amplitude, xlabel=("set", "s"), ylabel= ("Amplitude", "V"))
 
         Array shapes for 2D plots:
             x:(1D-length m), y:(1D-length n), z: (2D- n*m array)
@@ -171,8 +173,10 @@ class BasePlot:
 
         """
         # TODO this should really be a static method
-        return (getattr(data_array, 'label', '') or
+        name = (getattr(data_array, 'label', '') or
                 getattr(data_array, 'name', ''))
+        unit = getattr(data_array, 'unit', '')
+        return  name, unit
 
     def expand_trace(self, args, kwargs):
         """
--- a/qcodes/plots/pyqtgraph.py
+++ b/qcodes/plots/pyqtgraph.py
@@ -348,23 +348,42 @@ class QtPlot(BasePlot):
         """
         Updates x and y labels, by default tries to extract label from
         the DataArray objects located in the trace config. Custom labels
-        can be specified the **kwargs "xlabel" and "ylabel"
+        can be specified the **kwargs "xlabel" and "ylabel". Custom units
+        can be specified using the kwargs xunit, ylabel
         """
         for axletter, side in (('x', 'bottom'), ('y', 'left')):
             ax = subplot_object.getAxis(side)
+            # danger: 🍝
+            # find if any kwarg from plot.add in the base class
+            # matches xlabel or ylabel, signaling a custom label
+            if axletter+'label' in config and not ax._qcodes_label:
+                label = config[axletter+'label']
+            else:
+                label = None
+
+            # find if any kwarg from plot.add in the base class
+            # matches xunit or yunit, signaling a custom unit
+            if axletter+'unit' in config and not ax._qcodes_label:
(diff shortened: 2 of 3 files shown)


giulioungaretti:
@MerlinSmiles  this is what we talked about, opinions ? 
@QCoDeS/core  opinions ? 

:unicorn:

giulioungaretti:
This clearly does not work as intended with multi dimensional parameters.

MerlinSmiles:
@giulioungaretti Yes, thats exactly the thing
and does not need #494 to be perfect, if a driver does not have units it wont show up.
And you will have issues with multi dimensions [volt, current] ... but that is an issue with how you kind of try to plot an datasomethin that has two sets of data.

MerlinSmiles:
I have no idea where the dataset and stuff is now. previously it was that an array in the dataset got the `units` from the parameter, if it was a list it would just have a list. which ofc is wrong.

giulioungaretti:
@MerlinSmiles  yep. it's pretty much still the same spaghetti.

giulioungaretti:
something that we have to keep in mind for #476 !

AdriaanRol:
@giulioungaretti I really love this feature. However, I generally use the QCoDeS plotting independently. Would it be possible to allow specifying custom units using xlabel and ylabel (as per [the docstring](https://github.com/QCoDeS/Qcodes/blob/dfd48abb1b1261ac23d936f47729ce82029bd7b7/qcodes/plots/pyqtgraph.py#L344)) 

I am suggesting something allowing the following syntax 
```python
PyQtPlotObj.add(x=x, y=np.sin(x), xlabel=('Time', 's'), ylabel=('Amplitude', 'V'))

```

From what I can tell this would be a trivial addition to this PR that would greatly improve the plotting API.

giulioungaretti:
@AdriaanRol sure!

giulioungaretti:
@AdriaanRol  ✔️

AdriaanRol:
@giulioungaretti, If I read your changes correctly, this change would not allow using the labels in the "old" simple way, i.e.,:  

`xlabel=('label', 'unit')`, is usable (+1) but it is no longer be possible to use `xlabel='label'`, while the default PyQtGraph API does allow this in both ways. 

Would it be possible to default the unit to `None` if no unit is specified or otherwise just passing the arguments on to PyQtGraph directly? I think this would be good both for keeping the API as simple as possible to use as well as for backwards compatibility. 

In any case I'd be happy to use this either way

giulioungaretti:
@AdriaanRol  
```python 
qc.QtPlot.add(x=data, y=data, labelx=("a", "V"), labely=("b", "V")
```
where one **must**  pass a tuple of  (label, unit).
It does break a bit the old API , but it's a lot less ambiguous than having  either a tuple or a string.
But it's a easy change if people feel like it's better. Feedback  @QCoDeS/core  @QCoDeS/core-contributors   >

MerlinSmiles:
@giulioungaretti im with @AdriaanRol here. In the simple way to plot something with `add(xdata, ydata, 'x-string', `y-string`)` it would be nice if one isn't forced to write tuples and `None` units, but just leave them out.

giulioungaretti:
@MerlinSmiles @AdriaanRol  the best we can do, give the current way this is handled is:

```python
qc.QtPlot.add(x=data, y=data, xlabel="a", xunit= "V", ylabely="b", yunit= "V")
```
everything **must** be a kwarg, not my favourite but that's the way qc plot was give us :D

AdriaanRol:
@giulioungaretti, I'd be fine with that syntax (though there is a funny `ylabel*y*` ). 

That being said, I see no reason why accepting the kwarg to either be a float or a tuple of two floats would not be possible, a simple type check on the value should suffice. (Though at this point I'm happy either way). 

I also agree that the API being kwargs only is not optimal, but that is a completely different topic :)

giulioungaretti:
@AdriaanRol @MerlinSmiles @QCoDeS/core  ok took more that I thought to fix this, the internals are quite interesting. Would love a review!

qcodes/plots/base.py
```diff
@@ -81,7 +81,11 @@ def add(self, *args, updater=None, **kwargs):
                 into `x`, `y`, and optionally `z`, these are passed along to
                 self.add_to_plot.
                 To use custom labels and units pass for example:
-                    >>> plot.add(x=set, y=amplitude, xlabel=("set", "s"), ylabel= ("Amplitude", "V"))
+                    plot.add(x=set, y=amplitude,
```
AdriaanRol: Good docstring :+1: 

I guess it would not be possible to make them explicit keyword arguments of the add function that then always get passed on to the relevant subfunction? I think this would improve usability, as the shift-tab functionality would then show this as default keyword arguments. 

That being said, I consider that an enhancement that goes beyond the scope of this PR, so maybe for a next time.

qcodes/plots/pyqtgraph.py
```diff
@@ -348,23 +348,42 @@ def _update_labels(self, subplot_object, config):
         """
         Updates x and y labels, by default tries to extract label from
         the DataArray objects located in the trace config. Custom labels
-        can be specified the **kwargs "xlabel" and "ylabel"
+        can be specified the **kwargs "xlabel" and "ylabel". Custom units
+        can be specified using the kwargs xunit, ylabel
         """
         for axletter, side in (('x', 'bottom'), ('y', 'left')):
             ax = subplot_object.getAxis(side)
+            # danger: 🍝
```
AdriaanRol: why danger? I'd say that if someone wants to overwrite a label (to e.g., more clearly distinguish two voltages) that is intended behaviour.

giulioungaretti: @AdriaanRol  danger as in enter spaghetti code. (The emoji maybe does not render :D )

qcodes/plots/base.py
giulioungaretti: @AdriaanRol  right now that's unfortunately not possible :(  but it soon will !

giulioungaretti:
@AdriaanRol testing wise, this is tricky as you say. I think we'll live with this for now. Then once we have a leaner api, we can actually test the logic without plotting (i.e. making sure we actually have the right data and metadata  ) and let the plotting  library developers do their testing 🗡 

I will then proceed merging soon!

PR #500, opened 2017-02-20
Correct metadata for MultiParameter units

### LINKED ISSUE #499: Units and is_setpoint is not propagated by `_parameter_arrays`
For instance for a simple multiparameter with one array 

```

Out[8]:
{'__class__': 'qcodes.data.data_array.DataArray',
 'action_indices': (0, 0),
 'array_id': 'MyMockAlazar_rawname',
 'instrument': '__main__.MockAlazar',
 'instrument_name': 'MyMockAlazar',
 'is_setpoint': False,
 'label': 'rawlabel',
 'labels': ('rawlabel',),
 'name': 'rawname',
 'names': ('rawname',),
 'setpoint_labels': (('timelabel',),),
 'setpoint_names': (('timename',),),
 'shape': (629,),
 'unit': None,
 'units': ('v',)}
```

and

```
{'__class__': 'qcodes.data.data_array.DataArray',
 'action_indices': (0,),
 'array_id': 'timename',
 'is_setpoint': False,
 'label': 'timelabel',
 'name': 'timename',
 'shape': (629,),
 'unit': None}
```

Fix coming up in a few minuts

jenshnielsen:
Fixes part of #499

@QCoDeS/core

### DIFF
--- a/qcodes/loops.py
+++ b/qcodes/loops.py
@@ -553,8 +553,15 @@ class ActiveLoop(Metadatable):
             action_indices = ((),)
         else:
             raise ValueError('a gettable parameter must have .name or .names')
-
+        if hasattr(action, 'names') and hasattr(action, 'units'):
+            units = action.units
+        elif hasattr(action, 'unit'):
+            units = (action.unit,)
+        else:
+            units = tuple(['']*len(names))
         num_arrays = len(names)
+        num_units = len(units)
+        assert num_arrays == num_units
         shapes = getattr(action, 'shapes', None)
         sp_vals = getattr(action, 'setpoints', None)
         sp_names = getattr(action, 'setpoint_names', None)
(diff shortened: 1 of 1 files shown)


jenshnielsen:
This should probably have some tests

jenshnielsen:
This is not quite as simple as setting these arrays to setpoint changes their names by appending a _set. I'm not sure we want that?

jenshnielsen:
Decided to simplify this. Now only fixes the missing unit propagation from Multiparameter to the arrays

qcodes/loops.py
```diff
@@ -553,8 +553,15 @@ def _parameter_arrays(self, action):
...
-
+        if hasattr(action, 'names') and hasattr(action, 'units'):
+            units = action.units
+        elif hasattr(action, 'unit'):
+            units = (action.unit,)
+        else:
+            units = tuple(['']*len(names))
         num_arrays = len(names)
+        num_units = len(units)
+        assert num_arrays == num_units
```
giulioungaretti: Not sure why  this ?

jenshnielsen: Happy to get rid of it. In principle that should always be the case as units is constructed as 
`len(names)*" "` (wrapped in the right tuples) if not supplied. All of this extra handling was just to fix a test that explicitly deletes names from the loop and checks for the ValueError above

Issue #286, opened 2016-07-27
Metadata: Add qcodes version, git status, git diff and things


MerlinSmiles:
There are a few points we should save for the metadata:

For qcodes
- Save qcodes version
- Save current git hash
- Save current git branch
- Save git diff if there are changes in the branch (this should be limited to not go crazy on data, notebooks in branches...)

Then there it should be encouraged that a user has all working scripts and stuff in a git repo, and for that one we should save the same information
`station.add_repo('./scripts')`

Then there is the situation where qcodes is a package and installed on a system without git, we need to cover that too, I guess only saving qcodes version then it ok.

I'm not a git genius, which is why it didn't come with the metadata.
How should we do this exactly?

```
~/g/Qcodes git:feature/pyqtgraph_improvements ❯❯❯ git describe --all --dirty --long
heads/feature/pyqtgraph_improvements-0-gb269d73-dirty

~/g/Qcodes git:feature/pyqtgraph_improvements ❯❯❯ git rev-parse  HEAD
b0485b7c5e8448ac5107a43ff05eb730114c284c

~/g/Qcodes git:feature/pyqtgraph_improvements ❯❯❯ git rev-parse --abbrev-ref HEAD
feature/pyqtgraph_improvements
```

PR #501, opened 2017-02-21
Mark array parameter setpoints as such

jenshnielsen:
Second part of fix for #499 which was lifted out of #500 

This has the issue that it changes the name of a large number of arrays but it seems like these should be setpoints per definition

@QCoDeS/core


giulioungaretti:
Should be merged once hash qcodes  commit is in metadata.

giulioungaretti:
see #286 !

Issue #498, opened 2017-02-20
MultiParameter usability issues


jenshnielsen:
Just some notes of various issues that we should improve with respect to MultiParameters

* If you provide setpoints but not setpoint_names the setpoints are not used in plots. But stored in set_array in the dataset. Should perhaps raise
* The documentation claims that setpoints can be a tuple of tuples of various types including numpy arrays. This is currently not true. The code checks that the inner type is among  `nt, DataArray, collections.Sequence, collections.Iterator`which `numpy.ndarray` is not. If you add it to the check it will fail with a type error (see below). I have not tested if DataArrays work correct. Converting to a tuple solved the immediate problem

* The multidimensional setpoints are too complicated to use (1D then 2D the 3D etc)

jenshnielsen:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-3209025a3c3c> in <module>()
----> 1 data = qc.Measure(a.acquisition).run()

/Users/jhn/src/Qcodes/qcodes/measure.py in run(self, use_threads, quiet, data_manager, station, **kwargs)
     76 
     77         data_set = self._dummyLoop.get_data_set(data_manager=data_manager,
---> 78                                                 **kwargs)
     79 
     80         # set the DataSet to local for now so we don't save it, since

/Users/jhn/src/Qcodes/qcodes/loops.py in get_data_set(self, data_manager, *args, **kwargs)
    721 
    722             data_set = new_data(arrays=self.containers(), mode=data_mode,
--> 723                                 data_manager=data_manager, *args, **kwargs)
    724 
    725             self.data_set = data_set

/Users/jhn/src/Qcodes/qcodes/loops.py in containers(self)
    521                 # note that this supports lists (separate output arrays)
    522                 # and arrays (nested in one/each output array) of return values
--> 523                 action_arrays = self._parameter_arrays(action)
    524 
    525             else:

/Users/jhn/src/Qcodes/qcodes/loops.py in _parameter_arrays(self, action)
    594             for j, (vij, nij, lij) in enumerate(zip(sp_vi, sp_ni, sp_li)):
    595                 sp_def = (shape[: 1 + j], j, setpoints, vij, nij, lij)
--> 596                 if sp_def not in all_setpoints:
    597                     all_setpoints[sp_def] = self._make_setpoint_array(*sp_def)
    598                     out.append(all_setpoints[sp_def])

[...]

PR #502, opened 2017-02-21
Add Setpoint units

jenshnielsen:
Fixes part of #498 by adding a setpoint units which .are propagated to units for the setpoint data arrays

Done on top of #500 so that should be merged first

Needs some testing I think


giulioungaretti:
Good idea with testing!

giulioungaretti:
The underlying bits are so spaghetti that having a test will just make it a lot easier to understand what's going on 🗡

jenshnielsen:
Yes especially `_parameter_arrays` is more complicated that it needs to be.

jenshnielsen:
Should remember to add this to meta_attrs

PR #505, opened 2017-02-23
fix: Log exception and exit

### LINKED ISSUE #504: background task hides exceptions.
If you encounter a bug use the following template.
If you have a feature request feel free to freestyle.
### Steps to reproduce
``` python 
import qcodes as qc

d = DummyInstrument("dummy")

def fail():
    raise ValueError


loop = qc.Loop(d.dac1.sweep(0,20,0.1), delay=0.001).each(d.dac2)
data = loop.get_data_set(name='testsweep')


plot = qc.QtPlot()
plot.add(data.dummy_dac2)
_ = loop.with_bg_task(fail, plot.save).run()
``` 

### Expected behaviour
Raise exception and stop loop?

### Actual behaviour
Nothing happens

giulioungaretti:
Closes #504

-  log exception in background task
-  exit loop 

@alan-geller , and @QCoDeS/core  the question here is if we want to exit (early return) or not.
My feeling is that we may as well not exit, technically the data could be still good even if the background task failed, but one never knows.

The clean way is to return, but I am curious about your 💎  feedback    !

### DIFF
--- a/qcodes/loops.py
+++ b/qcodes/loops.py
@@ -47,6 +47,7 @@ Supported commands to .set_measurement or .each are:
 """
 
 from datetime import datetime
+import logging
 import multiprocessing as mp
 import time
 import numpy as np
@@ -64,6 +65,8 @@ from qcodes.utils.metadata import Metadatable
 from .actions import (_actions_snapshot, Task, Wait, _Measure, _Nest,
                       BreakIf, _QcodesBreak)
 
+
+log = logging.getLogger(__name__)
 # Switches off multiprocessing by default, cant' be altered after module
 USE_MP = config.core.legacy_mp
 MP_NAME = 'Measurement'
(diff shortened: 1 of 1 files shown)


alexcjohnson:
The logging is a great idea. I agree with @alan-geller though that we shouldn't exit on background task failure, unless the things people are doing with background tasks have changed substantially - data collection should not be contingent on live plotting, especially since the user can manually quit the loop if they're not seeing what they want.

Originally the code had the `last_task_failed` flag - the idea here was to allow the background task to fail *once* without being removed, following the idea "once is never, twice is always." I seem to recall seeing this with live plotting when the original draw may have had no data in it, but the second draw, when there is some data, would work? Of course that bug could be fixed, but who knows what else will come up particularly with Qt.

giulioungaretti:
@alexcjohnson @alan-geller I am not sure I like the "once is never, twice is always." But let's not change it for now.
So no early return, log the exception and remove the task if fail twice.

PR #509, opened 2017-03-03
Make ATS instrument serializable

### LINKED ISSUE #503: bug
### Steps to reproduce
1. make parameter class that inherits from `Parameter` with `instrument=None`
2. have an attribute in this class called `self.instrument` 

### Expected behaviour
this should work

### Actual behaviour
upon making a json snapshot, an error occurs. this happens because the instrument attribute interferes with
```python
       for attr in set(self._meta_attrs):
           if attr == 'instrument' and self._instrument:
               state.update({
                   'instrument': full_class(self._instrument),
                   'instrument_name': self._instrument.name
               })

           elif hasattr(self, attr):
               val = getattr(self, attr)
               attr_strip = attr.lstrip('_')  # eg _vals - do not include _
               if isinstance(val, Validator):
                   state[attr_strip] = repr(val)
               else:
                   state[attr_strip] = val
```
from
`_baseparameter` line 193
### System
**operating system**
win10
**qcodes branch**
master
**qcodes commit**
commit 6aabf835bc0dcdcb32a46285bd082f2c43d4eff7

damazter:
The AlazarParameter were non serializable because an explicit reference was stored.
The way that baseparameter works, now only the name is stored.

Fixes #503 

@giulioungaretti 
@jenshnielsen

### DIFF
--- a/qcodes/instrument_drivers/AlazarTech/ATS.py
+++ b/qcodes/instrument_drivers/AlazarTech/ATS.py
@@ -793,7 +793,8 @@ class AlazarParameter(Parameter):
                 # TODO(damazter) (S) test this validator
                 vals = validators.Enum(*byte_to_value_dict.values())
 
-        super().__init__(name=name, label=label, units=unit, vals=vals)
+        super().__init__(name=name, label=label, units=unit, vals=vals,
+                         instrument=instrument)
         self.instrument = instrument
         self._byte = None
         self._uptodate_flag = False


jenshnielsen:
:+1: is there any specific reason for keeping the `self.instrument = instrument line` too? I don't think there is any problem with it but it makes that paramenter slightly different from other parameters

damazter:
I cannot remember why that line of code is there. Could we postpone the possible resolution of that to a later moment?

jenshnielsen:
Ok lets merge as is

PR #512, opened 2017-03-07
Multiparameter fixes

jenshnielsen:
@QCoDeS/core 

This replaces #500 #501 and #502 

Simpler to do in one as the tests are strongly coupled


jenshnielsen:
Compared to the other prs this adds tests and removes an unwanted assert otherwise identical

PR #514, opened 2017-03-09
Fix: awg5014 missing units

jenshnielsen:
@WilliamHPNielsen 

One missing units -> unit transition in the awg driver. Spotted during our work with the alazar driver


qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -352,7 +352,7 @@ def __init__(self, name, address, timeout=180, **kwargs):
                 self.add_parameter(
                     'ch{}_m{}_del'.format(i, j),
                     label='Channel {} Marker {} delay'.format(i, j),
-                    units='ns',
+                    unit='ns',
```
giulioungaretti: Also I'd argue this should be seconds !

giulioungaretti: But let's wait for a later PR.

PR #515, opened 2017-03-09
Driver/keithley2400

nataliejpg:
- add a basic keithley 2400 driver
- could be extended/merged with this old one [here](https://github.com/nataliejpg/Qcodes/blob/Keithley2400-driver/qcodes/instrument_drivers/tektronix/Keithley_2400.py)
- needs (better) docstrings (?)

@giulioungaretti


jenshnielsen:
There is also a PR #488 by @WilliamHPNielsen

nataliejpg:
@jenshnielsen good point, didnt spot that. Thanks. It looks much more comprehensive than this one so I'll close this one. Would be fab if the other one could be merged soon though as it seems like it's a pretty useful driver (given that 3 people have now written it). Also are we no longer giving prs a 'driver' tag? I should have checked more carefully anyway but the reason I didnt see #488 is because I looked for 'driver' prs rather than 'feat's

giulioungaretti:
@nataliejpg yes that's the idea to have  drive/foo but being humans we tend to fail sometimes :D 
I will just merge it as proposed, and we'll fix the suspected bugs it may have as we go!

giulioungaretti:
also @nataliejpg  it is under driver/foo 🌹

nataliejpg:
@giulioungaretti cool cool, just wanted to check, it wasn't a moan (although I did double check and it's under 'feat: Add basic driver for Keithley 2400' so I wouldn't have called that 'driver/foo'. It did make me realise that the ZNB20 update I did is under feat as well where it shouldn't be so @WilliamHPNielsen  is by no means the only fallible one)

giulioungaretti:
<3 <3

PR #518, opened 2017-03-09
remove warning for get_idn, add default value

### LINKED ISSUE #517: default Instrument get_idn returns a warning
For any `Instrument` that does not have a `self.ask('*IDN?')` a warning is generated. This is inconvenient for all derived classes that do not have a VISA interface.

The warning is generated here:

https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument/base.py#L130

A minimal example:
```
from qcodes import Instrument

class dummy(Instrument):
    pass

d=dummy(name='test')
idn=d.get_idn()
```
Can we remove the warning (or change it into a `logging.info`)?

@giulioungaretti

peendebak:
Fixes #517 .

@unga @AdriaanRol

### DIFF
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -127,8 +127,8 @@ class Instrument(Metadatable, DelegateAttributes, NestedAttrAccess,
             if len(idparts) < 4:
                 idparts += [None] * (4 - len(idparts))
         except:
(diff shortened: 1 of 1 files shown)


giulioungaretti:
@peendebak  thanks ⚜️

PR #520, opened 2017-03-10
fix: Recursively update the configuration object

### LINKED ISSUE #519: Config  breaks on upstream updates.
If a user had a custom config and we update upstream by adding a new field.
Their config value will remove the new field added upstream, thus breaking the import of qcodes!

The other way around is impossible to prevent though. If we remove a field upstream there's no way to delete it locally ,or even to raise a warning , but that's cool because maybe local uncommitted doe depends on it :D  At worst it's a bit of extra noise!

giulioungaretti:
Closes #519

@eendebakpt does this fix your issues ?

### DIFF
--- a/qcodes/config/config.py
+++ b/qcodes/config/config.py
@@ -1,3 +1,4 @@
+import collections
 import copy
 import json
 import logging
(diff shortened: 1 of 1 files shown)


giulioungaretti:
I will merge this anyhow as it will for ex fix @nataliejpg issues!

PR #529, opened 2017-03-16
Add adjust_parameter_validator function to IVVI instrument

### LINKED ISSUE #301: validate fails with IVVI driver
The IVVI driver internally converts any `set` value to a integer number for setting (12 or 16 bit I think). For example -800 is converted to -800.0305180.
When a validator is set to range [-800, 400] a `set` with value -800 works fine, but then a `set` to -700 fails because -800.0305180 is not in the valid range.
### Steps to reproduce
1. Create an IVVI instrument and set validator range to [-800, 400]
2. Do ivvi.dac1.set(-800)
3. Do iivi.dac1.get()

I am not sure where to solve this one. The actual error is generated in `self.validate(start_value)` (see code snippet below. We could argue that the current value of a parameter is _always_ allowed, but there might be exceptions when multiple parameters use the same instrument.

``` python
    def _sweep_steps(self, value):
        oldest_ok_val = datetime.now() - timedelta(seconds=self._max_val_age)
        state = self._latest()
        if state['ts'] is None or state['ts'] < oldest_ok_val:
            start_value = self.get()
        else:
            start_value = state['value']

        self.validate(start_value)
```

@AdriaanRol @dbouman1 @CJvanDiepen @alexcjohnson

CJvanDiepen:
The dac's of the IVVI have a finite resolution. If the validator range min and max values are not values the dac's can actually have, then it can occur that a set command results in the dac's going to a value just outside the validator range. Adjusting the validators with this function prevents that.

Fixes #301.

- Add `adjust_parameter_validator` function to IVVI instrument.

@giulioungaretti @peendebak


giulioungaretti:
@peendebak @eendebakpt updates on this ? From my side it can be merged.

peendebak:
@cjvanDiepen @unga The PR looks good and is fully backwards compatible.:dancer:

PR #538, opened 2017-03-23
Awg enhancements

### LINKED ISSUE #537: AWG 5014 ref clock bad parameter name /confusion
The AWG 5014 driver has two distinct "parameters" in the hardware, 
- Clock source -> int or ext 
- Reference oscillator -> int or ext
image

The recently updated 5014 driver introduces a 'parameter' ref_clock_source. 
https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/tektronix/AWG5014.py#L160
This refers to using an internal or external clock and and has nothing to do with the reference oscillator, making this a confusing name. 
Additionally there is no parameter that sets the reference oscillator. 

The driver does contain some reference to these parameter under their proper name. 
https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/tektronix/AWG5014.py#L847-L852

I would suggest the following fix. 
- Rename the `ref_clock_source` parameter to `clock_source`. 
- Add a parameter for the `reference_source`
 Both of these names are the names that the AWG uses internally, both in it's GUI as in it's sequencer config. 

@WilliamHPNielsen as author of the recent fix, do you agree with these changes? If so I will be sending a PR shortly.

AdriaanRol:
Fixes #537 

- Added `clock_source` and `ref_source` parameters , removed ref 
- Changed get parser for filter function to use INF instead of 9.9e37 (which causes a rare bug)

A motivation for the first change is in #537 
A motivation for the second change is that INF is more readable (which is nice), it is the same value under the hood, and the value 9.9e37 caused me trouble when sending it to the monitor. I suspect it will prevent more trouble in the future. Additionally, the manual of the AWG says that this INF setting corresponds to open so the 9.9e37 is really a placeholder value on the 5014 side. 


YakBizzarro:
I've noticed that all Keithley/Tektronik instruments signal an out-of-range value with 9.9e37, maybe would be nice to have it has a general converter for all these instruments. 
Maybe to make it more resistant, it would be better to run the check on the converted value, something like

value = float(string)
value = value if value < 9.9e37 else float("inf")

AdriaanRol:
@YakBizzarro , Good point. I made the parser a bit more robust by using the change you proposed.

nataliejpg:
I'll try it out this next week, looks good to me though.

PR #539, opened 2017-03-24
Fix: zi ensure single trigger mode is always set

jenshnielsen:
Fixes garbage data on ZI 

@WilliamHPNielsen @jana-d fixes what we spent most of friday debugging


jenshnielsen:
Need to remember to fix scopemode as hacked in https://github.com/qdev-dk/Qcodes/pull/11

jenshnielsen:
We should probably try to only use one scope instance and then call 

unbind and finish in the end. That seems to be what the ZI gui does. Verify this from the gui log file.

Issue #508, opened 2017-03-03
Ivvi st_dacs_zero too fast


damazter:
@AdriaanRol 
### Steps to reproduce
1. use `ivvi.set_dacs_zero`
2.
3.

### Expected behaviour
All dacs should sweep to zero, no faster than is allowed by their parameter settings 
`dac_step` and `dac_delay`

### Actual behaviour
instead the function `_set_dac` is used which does not listen to `dac_step` and `dac_delay`

###Proposed Solution
change `set_dacs_zero` to (untested):
```python
for i in range(self._numdacs):
    self.parameters['dac{}'.format(i+1)].set(0)
```

### System
**operating system**
Windows 7

branch derived from
**qcodes branch**
master
**qcodes commit**
latest merged qcodes master commit:
c67ced65e36fa38b4369275924ffb5e2df505cdd

AdriaanRol:
@damazter I agree that this is unintended behaviour, and ofcourse I aprove of the proposed solution :+1: Thanks for finding this.

Issue #513, opened 2017-03-07
IVVI communication error does not raise correct error


AdriaanRol:
When using the IVVI driver, if you repeatedly get a communication error it should raise that exception. 

However, because of the line https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/QuTech/IVVI.py#L291 , the actual error is not raised. Instead it will give a very nondescript ``` File "d:\githubrepos\qcodes\qcodes\instrument_drivers\QuTech\IVVI.py", line 291, in _get_dacs
    raise('IVVI Communication error')
TypeError: ('exceptions must derive from BaseException', 'getting IVVI_dac16')```

Easy to fix by raising the original exception instead of a custom one, I think this both solves the bug and would be the intended behaviour.

giulioungaretti:
@AdriaanRol yes, indeed that is just not valid python code :D

WilliamHPNielsen:
This was fixed by #541

PR #541, opened 2017-03-24
Feat: Fix: Support for ivvi dac polarity and minor changes

lukabavdaz:
The main change proposed in this pull request is the functionality to set the polarity of the dacs when the instrument is intitialized. The validators of the dacs are modified to reflect the actual voltage range.

Minor changes include:
- The "(mV)" in the label of the dacs was removed for being out of place (the "unit" string should store this information)
- The function set_dacs_zero() has been modified to ramp the dacs according to the parameter settings by using the regular set function (Fixes issue #508)

@giulioungaretti


AdriaanRol:
Maybe out of scope of this PR, but would it make sense to also solve #513 (a horrible syntax error by me) in this PR and change mV to V in the unit to be consistent with #494 ?

lukabavdaz:
I completely agree with fixing #513. Since it's such a small fix it should still be within the scope of the PR.

However, I'm hesitant to change the unit because it will break backwards compatibility. I suggest doing these kinds of changes in a more visible way (like associating them with a release) so everyone will know that code relying on the units needs to be updated.

PR #542, opened 2017-03-25
A validator for lists

AdriaanRol:
Adds a new validator for lists. Optionally check each element in the list using any other validator. 

Includes test. 

@giulioungaretti


jenshnielsen:
👍 

I am not sure about how to best handle checking of elements. For a 10 element list just checking each element is Anything adds more than a factor of 10 to the runtime compared to just checking for the list type. Perhaps it should be possible to skip the element comparison? 


```python
In [33]: b = Anything()

In [37]: a = list(range(10))

In [38]: %timeit for c in a: b.validate(c)
1000000 loops, best of 3: 1.89 µs per loop

In [40]: %timeit isinstance(a, list)
10000000 loops, best of 3: 119 ns per loop

In [43]: %timeit for i in a: isinstance(i, int)
1000000 loops, best of 3: 1.3 µs per loop
```

AdriaanRol:
@jenshnielsen , I thought about that and thought that I added in the feature to skip the elt_wise validation if the base validator is vals.Anything but I guess I forgot to do that. Not sure why. 

I'll add it.

jenshnielsen:
@AdriaanRol Sounds good to me

AdriaanRol:
@jenshnielsen  I've implemented the change and tested that it works using a list of 30 elts. 

screenshot 2017-04-02 21 44 25

PR #553, opened 2017-04-01
Add locking to IVVI driver

peendebak:
- Add locks to the IVVI driver

The locking is optional, so without enabling them the behaviour of the driver is unchanged. The locks are needed for safety of our systems (otherwise we could have set random values to the DACs).

@giulioungaretti @AdriaanRol @WilliamHPNielsen @jenshnielsen

### DIFF
--- a/qcodes/instrument_drivers/QuTech/IVVI.py
+++ b/qcodes/instrument_drivers/QuTech/IVVI.py
@@ -3,6 +3,7 @@ import logging
 import numpy as np
 import visa  # used for the parity constant
 import traceback
+import threading
 
 from qcodes import VisaInstrument, validators as vals
 from qcodes.instrument.parameter import ManualParameter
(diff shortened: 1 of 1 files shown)


giulioungaretti:
@eendebakpt looks good to me! @AdriaanRol  ?

AdriaanRol:
@eendebakpt I'm a bit confused as to what the lock actually does. I guess it relates to multi threading in some way but I'm not really sure. Could you add a docstring explaining or referencing the appropriate concept? 

and @giulioungaretti , I'm also curious how the merge of this relates to the merge of #541 , which contains some related fixes to the IVVI driver and is also good to merge from what I can tell.

peendebak:
@AdriaanRol @giulioungaretti The IVVI driver is not thread safe. We use threads in our code in several ways: 1) a parameterviewer gui that runs in the background 2) the variable explorer of our IDE uses threads to display variables 3) some of our measurement functions use threads. Without the locking mechanism our IVVI driver will lock up during measurements.

AdriaanRol:
OK, that explains why we don't experience that. 

Thanks for the comment. @giulioungaretti lgtm let's 💃

PR #560, opened 2017-04-05
wip: Nasty fix for monitor

giulioungaretti:
Not sure I like this :D 

The other option is to make the Gui completely separate!


giulioungaretti:
fail.

PR #561, opened 2017-04-07
feature: Signal Chain

WilliamHPNielsen:
It is considered useful to be able to absorb passive elements along a transmission line such as attenuators into a QCoDeS parameter. This PR enables a minimal version of that functionality. We call it a "signal chain" to mimic the notion of a voltage propagating through INSTRUMENT -> ATTENUATOR -> OFFSET -> SAMPLE.  One is usually interested in setting the voltage on SAMPLE directly from the instrument parameter.

Add a signal chain for StandardParameters and a notebook on using them.

Still pending
- [x] Make signal chain elements removable
- [ ] Add more than an offset and a scale factor?

@giulioungaretti

### DIFF
--- a/docs/examples/Signal Chain Example.ipynb
+++ b/docs/examples/Signal Chain Example.ipynb
@@ -21,7 +21,7 @@
    "source": [
     "import qcodes as qc\n",
     "from qcodes.instrument.parameter import StandardParameter\n",
-    "from qcodes.instrument.signalchain import addToSignalChain"
+    "from qcodes.instrument.signalchain import addToSignalChain, removeFromSignalChain"
    ]
   },
   {
@@ -71,11 +71,11 @@
      "output_type": "stream",
      "text": [
       "DataSet:\n",
-      "   location = 'data/2017-04-07/#001_testsweep_10-05-28'\n",
+      "   location = 'data/2017-04-07/#005_testsweep_10-40-42'\n",
       "   <Type>   | <array_id>      | <array.name> | <array.shape>\n",
       "   Setpoint | dac_voltage_set | voltage      | (21,)\n",
       "   Measured | dmm_voltage     | voltage      | (21,)\n",
-      "started at 2017-04-07 10:05:31\n"
+      "started at 2017-04-07 10:40:45\n"
      ]
     }
    ],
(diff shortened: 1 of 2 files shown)


MerlinSmiles:
@WilliamHPNielsen does it make sense to change the actual parameter like that, instead of making a device_voltage parameter or whatever, that has those conversions, the raw data, metadata, and a clear understanding for the user that it is actually a different parameter than the one of the instrument?

I just imagine myself to mess that totally up, especially when restarting the notebook. then suddenly `dac.volt(1)` has a totally different meaning....

WilliamHPNielsen:
@MerlinSmiles Yes, I think so. It is hard to argue about what is confusing and what is not, but the user feedback we got from T10 made it clear that they found the conceptual divide between a DAC output channel parameter and a Voltage Divider is artificial and confusing (especially: it's hard to remember that the DAC paramter is now "dead"). The absorption of passive elements into the parameter is also intuitive to me personally, but I don't sit at a fridge.

Another benefit of this approach is that it easily enables accounting for multiple elements, so that you don't accidentally call VoltageDivider_2 when in fact your Offset_3 is the parameter with the actual voltage-at-target-value and VoltageDivider_2 is just somewhere along the chain.

MerlinSmiles:
@WilliamHPNielsen I do like your idea of chaining those kind of things, putting it in the original parameter however is a step backwards for me. Back to the situation where a parameter changes meaning and functionality and so on. Which was just what lead to the parameter confusions in the first place.

If I have a device i think in 
source current
gate_right
gate_left
gate_plunger
and so on, not dac5 or dac6

Also what happens to the voltage output (and unit) if you add a bias resistor? ;)

But hey, not my call...

YakBizzarro:
I agree with @MerlinSmiles , I find this signal chain interesting, but it would be very nice to have "final" name and unit like a standard parameter. In my case i extended the voltage divider to do account for gain/division/offset, here my 2 cents: https://gist.github.com/YakBizzarro/93289dfd27905e47804c4a216f44e1ec

WilliamHPNielsen:
We'll park this for now. The new parameters also include a scale factor.

PR #562, opened 2017-04-07
Fix: QDac save_val when not ramping voltage

WilliamHPNielsen:
The QDac was not saving UNramping voltages

@jenshnielsen


jenshnielsen:
This should not be needed as std paramers already call _save_val as they are calling the set and get cmds

WilliamHPNielsen:
I believe you are right. I'll take my head out of my ass and look at this again.

WilliamHPNielsen:
Now it's the other way round 😄

giulioungaretti:
@WilliamHPNielsen so this PR is actually fixing a double call to save_val ?

WilliamHPNielsen:
@giulioungaretti Correct. You have to appreciate the irony.

jenshnielsen:
I think this can be merged?

WilliamHPNielsen:
@jenshnielsen Do you want to give a positive review, or should I force merge?

jenshnielsen:
done

PR #566, opened 2017-04-10
add drivers for SRS SIM928 and Advantech PCIE-1751

antsr:
- add a `VisaInstrument` driver for Stanford Research Systems SIM928 module inside a SIM900 mainframe
- add a driver for Advantech PCIE-1751 DIO card that uses `cffi` for calling methods from the DLL


qcodes/instrument_drivers/stanford_research/SIM928.py
```diff
@@ -0,0 +1,264 @@
...
+        Queries the SIM900 mainframe, in which slots there is a module present.
+        :return: a list of slot numbers where a module is present (starting
+                 from 1)
+        """
+        CTCR = self.ask('CTCR?')
+        CTCR = int(CTCR) >> 1
+        modules = []
+        for i in range(1,10):
+            if CTCR & 1 != 0:
+                modules.append(i)
```
nulinspiratie: The code assumes that if there is a module in the slot, it must be a SIM928, while this is not always the case. Having an additional check here for a SIM928 solves this problem. For example:

```python
model = self.get_idn(i)['model']
if model == 'SIM928:
    modules.append(i)
```

qcodes/instrument_drivers/stanford_research/SIM928.py
```diff
@@ -0,0 +1,264 @@
...
+        self.write('SNDT {},"{}"'.format(i, cmd))
+
+    def set_voltage(self, i, voltage):
+        self.write_module(i, 'VOLT {:.3f}'.format(voltage))
+        self.parameters['voltage{}'.format(i)]._save_val(voltage)
+
+    def get_voltage(self, i):
+        return float(self.ask_module(i, 'VOLT?'))
+
+    def set_smooth(self, voltagedict, equitime=False, verbose=False):
```
nulinspiratie: This is a very useful function! We usually give the SIM modules a name, such as `top_gate`, and then forget about the slot number it is in. Would it make sense for the voltagedict to also allow parameter names as keys? It can then retrieve the corresponding slot?

qcodes/instrument_drivers/stanford_research/SIM928.py
```diff
@@ -0,0 +1,264 @@
...
+        self.write('SRST')  # SIM reset (causes 100 ms delay)
+        time.sleep(0.5)
+
+        self.modules = self.find_modules()
+        for i in self.modules:
+            self.write_module(i, 'TERM LF')
+            self.add_parameter('IDN{}'.format(i),
+                               label="IDN of module {}".format(i),
+                               get_cmd=partial(self.get_idn, i))
+            self.add_parameter('voltage{}'.format(i), unit='V',
```
nulinspiratie: We usually give the modules names, such as `top_gate` etc. Perhaps a way to also give custom names to the SIM modules would be useful? For instance, an optional kwarg that is a dict with items of the form `slot: name`?

nulinspiratie:
Hey @antsr thanks a lot for the PR! We also use the SIM928 voltage sources, and this driver would definitely be useful. I had a few comments, some of them I also addressed in the code.

- SIM modules are not necessary SIM928, an additional check such as a `get_idn` would solve this.
- The ability to give custom names to the voltage parameters would be useful
- The `voltagedict` arg in `set_smooth` uses the slots as keys. As we usually think of the modules in terms of their names, having the possibility to provide the parameter names would be useful.
- Once a SIM slot is stuck due to an error, a reset command is needed to unfreeze it
- A way to combine modules into a new parameter would be very useful. We often perform DC scans with these slots which we want to run as fast as possible. If we use the normal `CombinedParameter`, it will iteratively set each of the modules to the desired parameter. Instead, a combined parameter can send the commands to the different modules in a single GPIB command.

If you want to implement any of this and would like some help, let me know!

antsr:
Hi @nulinspiratie, thanks for the feedback. I think the additions you proposed would indeed be very useful. I added the possibility to name the SIM928 modules -- would such implementation suit your needs? If you could tell me the commands you use to reset the modules reliably, I could easily also add it to the driver.

nulinspiratie:
Hey @antsr  the naming of slots looks good! The reset command for a single slot is `SRST {slot_number}`, where `{slot_number}` should be replaced

qcodes/instrument_drivers/Advantech/PCIE_1751.py
```diff
@@ -0,0 +1,258 @@
+import cffi
+from functools import partial
+import collections
+import os
+import logging
```
giulioungaretti: We use  the convetion of

to get nicer logggin per module!

qcodes/instrument_drivers/stanford_research/SIM928.py
```diff
@@ -0,0 +1,413 @@
...
+            errors.append('Voltage Overload.')
+        if BatFault:
+            errors.append('Battery Fault.')
+
+        if raiseexc:
+            if len(errors) != 0:
+                raise Exception(' '.join(errors + warnings))
+        return errors + warnings
+
+    @classmethod
```
giulioungaretti: shouldn't this be a static method  ?

giulioungaretti:
@antsr  looking good, thanks for the contribution.
After the fixes request land it's ready to merge.

antsr:
I left a few unused variables in in the SIM928 driver, where it made sense to have all the status bits named as in the manual. Otherwise I addressed all the change suggestions and codacy issues.

giulioungaretti:
@antsr yes, good call! Clarity wins in this case! 

Merging!

PR #569, opened 2017-04-11
Doc fixes and notebooks

jenshnielsen:
Add example notebooks to docs.

* Convert notebooks to rst as part of docs build and include in docs.
* For the notebooks to be cleanly imported they need exactly one 1 level header so change them all to have that. 
* Mock the Zi so it's driver docs can be build.
* Few other small docs fixes

In general many of these notebooks needs some work. Either deleted, expanded or just checked for errors.

@QCoDeS/core

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,7 @@ coverage.xml
 
 # Sphinx documentation
 docs/_build/
+docs/_notebooks
 docs/api/generated/
 docs/_auto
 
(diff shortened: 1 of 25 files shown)


giulioungaretti:
@jenshnielsen looking good! Let's merge now so at least they are visible, then we can fix the notebooks one at time!

jenshnielsen:
I agree having a big pr with lots of notebook edits is just asking for merge conflict hell

giulioungaretti:
@jenshnielsen great! Merge whenever you deem appropriate !

PR #571, opened 2017-04-11
Travis: test on Ubuntu 14.04

jenshnielsen:
12.04 is soon end of life and 14.04 has a nbconvert supported version of pandoc

@QCoDeS/core


jenshnielsen:
Against this is naturally the fact that 14.04 is still in beta on travis

jenshnielsen:
This has become the default no need to configure

PR #572, opened 2017-04-11
Small changes cherrypicked from qdev-dk version 

jenshnielsen:
Make a couple of methods static and block a broken version of ipykernel


jenshnielsen:
@giulioungaretti as requested in https://github.com/qdev-dk/Qcodes/pull/4

PR #573, opened 2017-04-12
Mercury fixes

jenshnielsen:
Some small issues found while working with the code on T10

@MerlinSmiles @QCoDeS/core

### DIFF
--- a/qcodes/instrument_drivers/oxford/mercuryiPS.py
+++ b/qcodes/instrument_drivers/oxford/mercuryiPS.py
@@ -39,7 +39,7 @@ class MercuryiPS(IPInstrument):
     Args:
         name (str): name of the instrument
         address (str): The IP address or domain name of this instrument
-        port (int): the IP port to communicate on (TODO: what port is normal?)
+        port (int): the IP port to communicate on (default 7020, as in the manual)
 
         axes (List[str], Optional): axes to support, as a list of uppercase
             characters, eg ``['X', 'Y', 'Z']``. If omitted, will ask the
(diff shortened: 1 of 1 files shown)


MerlinSmiles:
@jenshnielsen are you sure that your changes of `[self.axes.index(ax)]` to `[0]` actually work?
the `_ATOB` makes setpoint and field a `len(ax)` array, no?

MerlinSmiles:
while youre at it, maybe we want to change this function:
```
    def _carttosphere(self, field):
        field = np.array(field)
        r = np.sqrt(np.sum(field**2))
        if r == 0:
            theta = 0
            phi = 0
        else:
            theta = np.arccos(field[2] / r)
            phi = np.arctan2(field[1],  field[0])
            if phi<0:
                phi = phi+np.pi*2
        return [r, theta, phi]
```
basically adding:
```
            if phi<0:
                phi = phi+np.pi*2
```
to make sure phi is always positive?

MerlinSmiles:
and maybe we want to define if we run the magnet in current or field mode, and then remove all the `...C` guys

jenshnielsen:
@MerlinSmiles you are probably right but unfortunately it doesn't take that code path for other parameters such as z field

MerlinSmiles:
now I see the issue, yes its only going there for the 'current' version of sending the field. Stupid mistake.

A note on the side, using the current parameters instead of tesla parameters gives you one decimal place higher in resolution on the magnet.

jenshnielsen:
Agreed with @MerlinSmiles in Slack

PR #575, opened 2017-04-18
Feature/signadyne driver

lgnjanssen:
Adds drivers for the Keysight AWG and Digitizer PXIe cards (series M32/M33xxA).

- Add general Keysight SD_Module, SD_AWG and SD_DIG drivers
- Add specific M3201A and M3300A card drivers
- Add test suites for the above drivers

@giulioungaretti @WilliamHPNielsen @maij


giulioungaretti:
@jenshnielsen aka master of sphinx. How do we solve the build issue here?
In other drivers we raise a warning, instead of re raising the exception. Maybe we should to the same here ?

jenshnielsen:
The tests in test_suite.py need to be skipped if the keysight module is not installed.

I would suggest

```
import unittest
try:
    from .M3201A import Keysight_M3201A
    from .M3300A import M3300A_AWG
    from .SD_common.SD_Module import SD_Module
except ImportError:
    pass
    SD_Module = None
    Keysight_M3201A = None
    M3300A_AWG = None
```
changing the import to catch any import errors and then decorating the test classes with 

```
@unittest.skipIf(not SD_Module, "SD_Module tests requires the keysightSD1 module")
```

Alternatively you can catch the ImportError and raise a unittest.SkipTest in the exception handling but then the tests are not counted correctly by pytest. 

To make sure the documentation builds without the keysight module you should add it to autodoc.mock_import here https://github.com/QCoDeS/Qcodes/blob/master/docs/conf.py#L387

giulioungaretti:
@lgnjanssen  see  suggestion from @jenshnielsen (with which I totally agree)

and @lgnjanssen  super contribution! Thanks

PR #576, opened 2017-04-19
Triton fixes

jenshnielsen:
@MerlinSmiles I think this is mainly your driver does this make sense to you?

### DIFF
--- a/qcodes/instrument_drivers/oxford/triton.py
+++ b/qcodes/instrument_drivers/oxford/triton.py
@@ -155,7 +155,7 @@ class Triton(IPInstrument):
         for ch in allchans:
             msg = 'READ:SYS:DR:CHAN:%s' % ch
             rep = self.ask(msg)
-            if 'INVALID' not in rep or 'NONE' not in rep:
+            if 'INVALID' not in rep and 'NONE' not in rep:
                 alias, chan = rep.split(':')[-2:]
                 self.chan_alias[alias] = chan
                 self.add_parameter(name=alias,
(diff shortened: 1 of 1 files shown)


MerlinSmiles:
@jenshnielsen dont i just cross checked on T12, and the channel numbering as i wrote it seems to be correct?
Cant comment on the reg files, you are probably right :)

image

jenshnielsen:
Yes I see that too. But if i Ask the api for T0 I get nonsense such as (1.6e123) If I ask for T1 I get the value that matches `chan[0]` So in the regedit plus default names are chan[0] ... chan[n-1] where as in the api I need to ask for T1 ... TN

jenshnielsen:
This seems to match section 5.11.11 in the triton manual

```
<UID> [T1 ... Tn]
```

MerlinSmiles:
argh i missed that I used a modified version, so I guess you are right. Sorry for that... lets have a quick chat on slack

qcodes/instrument_drivers/oxford/triton.py
```diff
@@ -163,30 +164,43 @@ def _get_named_channels(self):
...
-                self.chan_temps[chan] = {'name': name, 'value': None}
-                self.add_parameter(name=chan,
-                                   unit='K',
-                                   get_cmd='READ:DEV:%s:TEMP:SIG:TEMP' % chan,
-                                   get_parser=self._parse_temp)
+                self.chan_temp_names[chan] = {'name': name, 'value': None}
+
+    def _get_temp_channels(self):
+        self.chan_temps = []
+        for i in range(1, 16):
```
jenshnielsen: This should probably be range(1,17) so that all 16 channels are included. We are not using T16 at Triton 10 but anyway

jenshnielsen:
Discussed with @MerlinSmiles on Slack and made some modifications based on his suggestions. The reg file is now optional and a few bugs fixed

PR #583, opened 2017-04-26
fix: Make driver print connect_message

WilliamHPNielsen:
Make the driver print the connect_message. Also add a bare minimum of docstrings.

- The driver now prints a connect message upon connecting

@giulioungaretti


MerlinSmiles:
That silent switch... I think it would be great to have that in the base driver, (maybe with a different name) or in the config files... :)

giulioungaretti:
@MerlinSmiles do we an issue for that request ?

MerlinSmiles:
@giulioungaretti I dont actually think so, just stumbled upon this and thought it would be nice to have in a consistent way, also thought that a keyword 'silent' could mean whatever, maybe the instrument stops beeping or so...

PR #585, opened 2017-04-27
Mercury iPS make it possible to block on set until the magnet is in hold mode

jenshnielsen:
@MerlinSmiles It seems to me like the easiest way to ensure that the magnet is stable is to wait for it to go into hold more before moving on. Do you see any issues with this

### DIFF
--- a/qcodes/instrument_drivers/oxford/mercuryiPS.py
+++ b/qcodes/instrument_drivers/oxford/mercuryiPS.py
@@ -211,6 +211,11 @@ class MercuryiPS(IPInstrument):
             fld = getattr(self, ax.lower()+'_fld')
             fld.set(setpoint)
 
+    def _ramp_to_setpoint_and_hold(self, ax, cmd, setpoint):
+        self._ramp_to_setpoint(ax, cmd, setpoint)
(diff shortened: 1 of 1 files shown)


MerlinSmiles:
@jenshnielsen yes I do, a sweep will take much longer time. And it will not get more precise than a difference.
If one wants better stability one would have to add delays and also wait for the device to settle, i.e. the sample temperature usually changes when sweeping the field.

In fact in some cases (relativeley fast measurement and slow magnet sweeps) can give you a much cleaner measurement if you continously sweep the field from start to stop (i.e. not stepping the field) and measure along the way. The reason is that the magnet can just slowly ramp instead of starting and stopping... But we cant really do that because we cant just measure for 600 seconds :(

jenshnielsen:
Note that in this proposal it is optional and off by default but I would like to have the option of waiting for the hold. 

I agree it's preferred to sweep the magnet and measure continuously but that can't be done for more complex curves with the current mercury firmware. (This is what we ultimately want but it's unclear if it will be possible any time soon) 

In principle your are right that you can achieve the same by waiting for a fixed delay but I would like to optionally be sure that the magnet is in hold mode before moving on.

MerlinSmiles:
Sorry I didn't look at the code at all if its optional i think its a great addition :)

PR #588, opened 2017-05-02
Add block averaged reads and python headers to the driver

peendebak:
Fixes #issuenumber.

@giulioungaretti @jenshnielsen @jpdehollain


peendebak:
@giulioungaretti Merging is blocked because of some formatting checks. These are in headers files (e.g. `h2py.py`) we have not written, and which I would rather not change because of that reason. 
Can we leave them as is?

giulioungaretti:
@peendebak  let me get back to you. I guess it may make sense to leave them as is.

jenshnielsen:
I think we should leave them as they are but I think there should be a readme/licence documenting the origin and which licence the code is distributed under. The headers are from the main python source I think?

peendebak:
@jenshnielsen @giulioungaretti Added a small README file

qcodes/instrument_drivers/Spectrum/py_header/h2py.py
```diff
@@ -0,0 +1,180 @@
...
+            # gobble up continuation lines
+            while line[-2:] == '\\\n':
+                nextline = fp.readline()
+                if not nextline: break
+                lineno = lineno + 1
+                line = line + nextline
+            name = match.group(1)
+            body = line[match.end():]
+            body = pytify(body)
+            #UE: Python3 kommt mit den l am Ende der Konstanten nicht klar, daher entfernen
```
giulioungaretti: lol

qcodes/instrument_drivers/Spectrum/py_header/h2py.py
```diff
@@ -0,0 +1,180 @@
...
+            body=body.strip()
+            if body[-1:] == 'l':
+                body = body[:-1]
+            #/UE
+            ok = 0
+            stmt = '%s = %s\n' % (name, body)
+            try:
+                exec stmt in env
```
giulioungaretti: double lol

giulioungaretti:
ok, @eendebakpt  the code in header is like mindblowlingly sub-par. But I guess we should just keep it there! If you agree, then I will merge!

peendebak:
@giulioungaretti I agree. Let's merge!

peendebak:
Changing channels with `m4i.enable_channels` takes a long time if the channels that are active are changed. For that reason the `_read_channel` funciton now always sets all channels to active.

@giulioungaretti PR is ready for now

PR #589, opened 2017-05-02
fix: Triton: Don't read existing channel

jenshnielsen:
@MerlinSmiles  As far as I can see this call to `_get_control_channel` would always return the previous channel (since force defaults to True now) We can either do like this or explicitly call `_get_control_channel`  with `Force=False` Does this make sense to you?

### DIFF
--- a/qcodes/instrument_drivers/oxford/triton.py
+++ b/qcodes/instrument_drivers/oxford/triton.py
@@ -5,7 +5,7 @@ import logging
 from traceback import format_exc
 
 from qcodes import IPInstrument
-from qcodes.utils.validators import Enum
+from qcodes.utils.validators import Enum, Ints
 
 
 class Triton(IPInstrument):
@@ -53,7 +53,8 @@ class Triton(IPInstrument):
         self.add_parameter(name='pid_control_channel',
                            label='PID control channel',
                            get_cmd=self._get_control_channel,
-                           set_cmd=self._set_control_channel)
+                           set_cmd=self._set_control_channel,
+                           vals=Ints(1,16))
 
         self.add_parameter(name='pid_mode',
                            label='PID Mode',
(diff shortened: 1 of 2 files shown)


jenshnielsen:
Also added a validator to pid_channel since it seems to always be a channel number between 1 and 16

MerlinSmiles:
@jenshnielsen you are right I think we should set the force to False otherwise every call will do that stupid loop thing.
Or alternatively we could in the loop check the current control channel first to see if it has changed, before doing the whole loop. And maybe set force to false on the `pid_setpoint` parameter, which is the only parameter i'd ever use in a sweep.
Perfect with the validator..

jenshnielsen:
Thanks The specific problem that I am trying to solve here is that is also called with Force True when trying to set the channel which prevents the channel from ever being set to something else as far as I can see

MerlinSmiles:
Sounds like a very useless feature 💯

jenshnielsen:
I have added a short circuit to `_get_control_channel` that always checks the current channel first. I think this should be ready to merge

qcodes/instrument_drivers/oxford/triton.py
```diff
@@ -125,19 +126,25 @@ def get_idn(self):
...
+        if self._control_channel:
+            tempval = self.ask('READ:DEV:T{}:TEMP:LOOP:MODE'.format(self._control_channel))
+            if not tempval.endswith('NOT_FOUND'):
+                return self._control_channel
+
+        # either _control_channel is not set or wrong
+        for i in range(1,17):
+            tempval = self.ask('READ:DEV:T{}:TEMP:LOOP:MODE'.format(i))
+            if not tempval.endswith('NOT_FOUND'):
+                self._control_channel = i
```
MerlinSmiles: Add a break here too? So that we use the first value we find?

jenshnielsen: Yes. Hopefully there will only be one?

MerlinSmiles: Yes. if not the current way would take the last possible... so both would be wrong

MerlinSmiles:
Looks good to me!

PR #590, opened 2017-05-02
Triton Http proxy server

jenshnielsen:
- [x] Basic aiohttp server for setting and getting Triton parameters, with GET and POST
- [x] RemoteTriton driver. Should be a drop in replacement for the current driver to use the proxy. 
- [x] Websockets exposing Json representation of the data
- [x] Endpoint to start/stop ws
- [x] Config for Triton location, ports, wait times etc.
- [x] Testing locally on real Triton computer
- [x] Benchmarking. Requires Local testing
- [ ] Finish reading of Parameter names from regdatabase on local Triton. (low priority)


setup.py
```diff
@@ -60,7 +60,8 @@ def readme():
           'ipython>=4.1.0',
           'ipykernel!=4.6.0', # https://github.com/ipython/ipykernel/issues/240 in 4.6
           'jupyter>=1.0.0',
-          'h5py>=2.6'
+          'h5py>=2.6',
+          'requests>=2.13.0'
```
jenshnielsen: Perhaps. The client uses requests currently so you only really need aiohttp on the Fridge. We could add it to an extra requires thing as Matplotlib and pyqtgraph

giulioungaretti: either is fine, but somewhere :D

qcodes/instrument_drivers/oxford/triton_remote_driver.py
```diff
@@ -0,0 +1,135 @@
...
+                               get_parser=float,
+                               get_cmd='{}?attribute=value'.format(temperature))
+
+    def close(self):
+        self._session.close()
+        super().close()
+
+
+    def ask_raw(self, cmd):
+        response = self._session.get('http://localhost:8000/{}'.format(cmd))
```
jenshnielsen: The address will most likely be the IP of the Triton computer as the remote driver runs on the experiment computer but the server on the triton computer

jenshnielsen: This now takes both from the constructor. I will add a default port from config

setup.py
jenshnielsen: Added as extra requirement

jenshnielsen:
I think this is now ready to be tested locally on a Triton computer. The websockets implementation is done. However as noted in the comments the aiohttp websockets library is now well suited for this kind of operation due to the design of its write coroutine.

giulioungaretti:
@jenshnielsen  yes  that  web socket looks  really nasty. The loop wait time should also be configurable  .

Let's try this on fridge and see, then we can review this properly.

giulioungaretti:
As per discussion  IRL, it's better to have the http server to also be ws client rather than exposing a web socket endpoint ?

jenshnielsen:
First version of that pushed. Seems to work ok, gracefully handles ws-server going offline and coming back. The websockets library is way better than the aiohttp websockets support

giulioungaretti:
@jenshnielsen probably not pushed but yes it is indeed much better!

jenshnielsen:
pushed now. Not sure what happened yesterday thought I pushed

giulioungaretti:
Looks good, need some polishing around but nice!

We may want to consider to add a endpoint to stop/start/restart the web socket client ?

qcodes/instrument_drivers/oxford/fridge_http_server.py
```diff
@@ -1,23 +1,27 @@
...
 from qcodes.instrument_drivers.oxford.triton import Triton
 from qcodes.instrument_drivers.oxford.mock_triton import MockTriton
 
+log = logging.getLogger(__name__)
+
 class FridgeHttpServer:
 
-    def __init__(self, name='triton', tritonaddress='http://localhost', use_mock_triton=True, tritonport=33576, websocket_wait_time=10):
-        self._websocket_wait_time = websocket_wait_time
+    def __init__(self, name='triton', triton_address='http://localhost', use_mock_triton=True, triton_port=33576):
```
giulioungaretti: should we load the defaults from the config at module import ?

qcodes/instrument_drivers/oxford/fridge_http_server.py
```diff
@@ -0,0 +1,231 @@
...
+        """
+        Return a dict that contains the parameter from the Triton to be exported via WS. This includes a read only copy
+        of all parametes exposed by the driver.
+        """
+
+        triton_parameters = []
+        for parametername in triton.parameters:
+            parameter = getattr(triton, parametername)
+            temp = {}
+            temp['value'] = parameter.get()
```
giulioungaretti: shouldn't this maybe a get_latest ?

jenshnielsen: Maybe, not sure. If it's a get_latest and nothing else is reading from the remote driver then it will never be updated?

qcodes/instrument_drivers/oxford/fridge_http_server.py
```diff
@@ -0,0 +1,231 @@
...
+                            await websocket.send(jsondata)
+                        await asyncio.sleep(ws_sleep)
+            except OSError:
+                log.info("Websocket server is offline will retry later")
+                await asyncio.sleep(ws_connect_sleep)
+            except websockets.exceptions.ConnectionClosed:
+                log.info("Websocket closed. Will try to reconnect later")
+                await asyncio.sleep(ws_connect_sleep)
+
+    async def websocket_enable(self):
```
giulioungaretti: Not sure this should be web socket  actually !
Just a simple http endpoint would totally do,  no ?

jenshnielsen: Yes I guess I made it a web socket so that the ws server would only need to communicate over one protocol but it doesn't actually help since it's using a different port anyway

qcodes/instrument_drivers/oxford/fridge_http_server.py
giulioungaretti: true that

qcodes/instrument_drivers/oxford/fridge_http_server.py
jenshnielsen: Changed to a http endpoint now

jenshnielsen:
Benchmarked with @MerlinSmiles on T12

Getting

Existing driver
```
%%time
triton.PT1()
CPU times: user 752 µs, sys: 1.08 ms, total: 1.83 ms
Wall time: 3.09 ms
```
Http server proxy
```
%%time
triton.T13()
CPU times: user 3.92 ms, sys: 916 µs, total: 4.83 ms
Wall time: 11.6 ms
```

Setting

Local
```
%%time
triton.pid_rate(0.001)
CPU times: user 1.04 ms, sys: 1.14 ms, total: 2.18 ms
Wall time: 281 ms
```
Http server proxy
```
%%time
triton.pid_rate(0.002)
CPU times: user 3.34 ms, sys: 1.29 ms, total: 4.63 ms
Wall time: 316 ms
```

jenshnielsen:
So it is quite a bit slower getting values via a http get but the effect on setting them seems to be much smaller

jenshnielsen:
Running benchmark again. T10 while websocket connection is sending every second

```
%%time
triton.T13()
CPU times: user 4.07 ms, sys: 1.26 ms, total: 5.33 ms
Wall time: 6.18 ms
```
Repeated several times seems to be consistent around 6 ms

Faster than above (due to wired network connection ?)

jenshnielsen:
Repeating the same benchmark while another client is getting a result every 1 second seems to give similar results 

```
%%time
triton.T13()
CPU times: user 3.84 ms, sys: 1.21 ms, total: 5.04 ms
Wall time: 5.83 ms
```

jenshnielsen:
Increasing the polling to every 0.1 seconds does not seem to be an issue either.

```
%%time
triton.T13()
CPU times: user 3.32 ms, sys: 1.49 ms, total: 4.81 ms
Wall time: 5.09 ms
```

MerlinSmiles:
could you do those tests continously to see for how long another client blocks a request?

jenshnielsen:
Running one client continuously with no throttle and getting from another one I see

```
%%time
triton.T13()
CPU times: user 4.51 ms, sys: 2.41 ms, total: 6.92 ms
Wall time: 8.72 ms
```

jenshnielsen:
Getting the results locally is still faster (so the current speedup is likely due to wired network)

```
%%time
triton.T1()
CPU times: user 615 µs, sys: 815 µs, total: 1.43 ms
Wall time: 1.49 ms
```

giulioungaretti:
@jenshnielsen let's merge this asap but not before #519 is fixed 🤐

giulioungaretti:
lol sleep deprivation, it's closed already./

PR #591, opened 2017-05-04
Small tweaks to docs build

jenshnielsen:
* Split dependencies out from travis file so they are easier to install locally
* When building from a pr build the actual pr docs rather than the master docs so we can verify these from the travis logs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,8 +37,14 @@ script:
 
 after_success:
     # install dependencies for docs build
-    - pip install -r docs_requirements.txt
+    - pip install -r ../docs_requirements.txt
     - python-codacy-coverage -r coverage.xml
     - cd ..
     # build the docs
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -37,8 +37,14 @@ script:
...
 after_success:
     # install dependencies for docs build
-    - pip install -r docs_requirements.txt
+    - pip install -r ../docs_requirements.txt
     - python-codacy-coverage -r coverage.xml
     - cd ..
     # build the docs
-    - make -f docs/Makefile gh-pages
+    - |
+      if [[ $TRAVIS_REPO_SLUG == "qcodes/qcodes" ]]; then
```
giulioungaretti: This works for PR from forks, but not from branches from this repo, no ?

jenshnielsen: fixed now to also check that we are on the master branch

PR #604, opened 2017-05-08
feat: Add loop index for quick access to actions

nulinspiratie:
Add `__getitem__` for `Loop` to easily access actions. 
This way, `loop[idx]` is equivalent to `loop.actions[idx]`. 
When working with 2D-loops or higher, this makes it easier to access inner actions.

An example for when this is useful is if you want to attach `plot.update` to an inner loop.

@giulioungaretti


giulioungaretti:
@nulinspiratie  not sure I see the benefit but I don't have anything against this either :D

nulinspiratie:
Yeah it's definitely a minor thing. It just makes accessing actions a bit easier, and I think the actions are the natural choices for items of the loop

giulioungaretti:
LGTM!

PR #608, opened 2017-05-09
Added methods to modify results within an in-progress DataSet

alan-geller:
- Added the ability to modify an existing result in a DataSet, as requested on Slack by Adriaan Rol.

@giulioungaretti

### DIFF
--- a/specs/DataSet.rst
+++ b/specs/DataSet.rst
@@ -77,7 +77,8 @@ Writing
 ----------
 
 #. It should be possible to add a single result or a sequence of results to an in-progress DataSet.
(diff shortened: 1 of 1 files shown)


AdriaanRol:
@alan-geller thanks for adding this. I would like it if the modify data would also support setting data (i.e., if there is no data set for a particular index). That way I can use the same method to set the data and don't have to add an extra if statement to my code. However I can see the benefit of enforcing the user to be aware of the fact that they are modifying data. I'm fine with leaving it as is and adapting to the new API.

giulioungaretti:
@AdriaanRol let's start with this, we can maybe later expand !

PR #611, opened 2017-05-11
Fix/do1d flushing

WilliamHPNielsen:
Fixes an issue so far only seen with the SR830: the message queue becomes unsynchronised at non-reproducible times (but seemingly always during loops), causing the data returned by the instrument to be misread and consequently many orders of magnitude wrong.

- Preflush all VISA buffers in do1d

@giulioungaretti


WilliamHPNielsen:
lol nevermind

PR #613, opened 2017-05-15
Travis: fail on failed docs build (and fix issues)

jenshnielsen:


jenshnielsen:
@peendebak I had to make some changes to the spectrum code to make it actually python3 compatible/

* h2py exec from statement to function
* h2py long() to int(). Python 3 does not have a specific long integer type

I am not sure that h2py is actually used anywhere 

* Do a relative import of py_header.regs and spcerr otherwise it would never import the bundled versions 

Is this ok?

jenshnielsen:
@QCoDeS/core Can we merge this to unblock the docs build which is currently broken

giulioungaretti:
@jenshnielsen  we could split this into two as  its seems like the repose time time from Delft is longish ?

jenshnielsen:
Yes perhaps we should just exclude the pyheaders code from sphinx instead of fixing it

jenshnielsen:
Ok changed this around to just exclude the headers from the auto api docs build

giulioungaretti:
@jenshnielsen  great thanks !

PR #615, opened 2017-05-16
Fix: Matplot interpolate x and y coords for pcolormesh

### LINKED ISSUE #257: MatPlot bug

jenshnielsen:
@nataliejpg this should fix #257

Missing
- [ ] I would like to get rid of the for loop for x
- [ ] There should probably be some tests for this

### DIFF
--- a/qcodes/plots/qcmatplotlib.py
+++ b/qcodes/plots/qcmatplotlib.py
@@ -109,6 +109,26 @@ class MatPlot(BasePlot):
             # in case the user has updated title, don't change it anymore
             self.title.set_text(self.get_default_title())
 
+    @staticmethod
+    def _calc_corners(array):
+        """
+        Extrapolate coordinates of between data points from a list of data points.
+        Needed because pcolor and friends take the list of edges as x and y rather 
+        than the center positions. This should handle general non linear axis.
+        """
+        myarray = array.copy()
+        # extrapolate points before and after array
+        # so we can find steps in the end/start
+        myarray_plusone = 2 * myarray[-1] - myarray[-2]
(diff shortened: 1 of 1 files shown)


jenshnielsen:
Sorry I missed that this is also solved by #337 handling NaN's better

nulinspiratie:
@jenshnielsen I can get an updated version out pretty soon

nulinspiratie:
@jenshnielsen can this be closed?

jenshnielsen:
yes thanks

PR #392, opened 2016-11-19
feature: driver znb20 update

nataliejpg:
Changes proposed in this pull request:
- adds spectroscopy options to the ZNB driver
- cleans it up somewhat, adding better docstrings (maybe overkill...)
- extension to example notebook

@WilliamHPNielsen @giulioungaretti @alexcjohnson (not sure who best to mention...)

- [ ] what should it be called given that is can be used with ZNB20 and ZNB8?
- [x] does it handle 'on' and 'off' of spec mode properly
- [ ] do comments and docstrings make sense

PR #618, opened 2017-05-16
driver/rebased feature: znb20 update

jenshnielsen:
Rebased version of #392. As discussed there it will not go in like this but can be used as a starting point for other changes


nataliejpg:
also one more bug I came across is that the get and set parsers for start, stop, centre, span parameters should be float not int

jenshnielsen:
Thanks @nataliejpg

nataliejpg:
also there should be a check that start frequency is less than stop frequency.

PR #568, opened 2017-04-11
feat/channelization: Add channels to instruments

spauka:
Implements #273

This merge implements channelization in much the way that was discussed in #273, and commits two sample instruments that demonstrate the key features. Briefly, the syntax allows the creation of lists of channels within an instrument and manipulation of those channels. Examples of usage appear at the bottom of the commit.

Changes proposed in this pull request:
- Add InstrumentChannel and ChannelList classes to hold channel classes. Addresses #273.
- Create a driver for the Lakeshore Model336 Temperature controller that uses channels
- Modify the driver for the Harvard DecaDAC to use channels. Fixes #478, #546, #563.

For example:
```pycon
>>> import qcodes as qc
>>> from Harvard.Decadac import Decadac
>>> d = Decadac("DAC", "ASRL4::INSTR")
Connected to Harvard DecaDAC (hw ver: 0, serial: 0) in 0.47s
>>> # Set a single channel
>>> d.channels[0].volt.set(-1)
>>> # Get a single channel
>>> d.channels.volt[0].get()
-1.00006103515625
>>> # Get multiple channels simultaneously
>>> d.channels.volt.get()
(1.00006103515625, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0)
>>> # Get a subset of channels
>>> d.channels[0:4].volt.get()
(-1.00006103515625, 0.0, 0.0, 0.0)
>>> # Combining different subsets is also possible
>>> fine = d.channels[0:2] + d.channels[4:6]
>>> fine.volt.get()
(-1.00006103515625, 0.0, 0.0, 0.0)
>>> # This also works with measure and loop
>>> qc.Measure(d.channels[0].volt).run()
DataSet:
   location = 'data/2017-04-11/#001_{name}_11-04-36'
   <Type>   | <array_id>           | <array.name> | <array.shape>
   Setpoint | single_set           | single       | (1,)
   Measured | DAC_Slot0_Chan0_volt | volt         | (1,)
acquired at 2017-04-11 11:04:36
```

Things left
 - [x] Channel metadata is not correctly extracted from an instrument snapshot.
 - [x] Test cases need to be written.

@giulioungaretti

PR #623, opened 2017-05-19
Znb 20 fixes as requested by t10

jenshnielsen:
- [x] Read out S11, S12, S22, S21
- [x] Ability to autoscale
- [x] Center/Span
- [x] Display all outputs on request
- [ ] Markers and peak finders
- [x] Output scale
- [x] default settings on startup


jenshnielsen:
@core This should be ready for review. The markers and peak finders are not implemented but that is less de scoped for the moment. 

@nataliejpg Would be great if you have a chance to look at the changes. The example notebook is probably the easiest place to start

jenshnielsen:
A new version based on channels is available in https://github.com/jenshnielsen/qcodes/tree/znb_with_channels this requires #568 to be merged along with more modifications in prs for that pr

This is a good deal cleaner but still defaults to setting up 4 channels. In the future we we can expand this to dynamically add channels with parameters (z, y snm n,m>2...) I don't think this is a performance problem as we disable output before capturing a channel output 

- [ ] Actually test the new implementation with channels 
- [ ] benchmark performance implications of multiple channels

jenshnielsen:
replaced by #643

PR #625, opened 2017-05-24
fix: Remove useless notebook detection

### LINKED ISSUE #624: notebook // matplotlib confusion
The function in_notebook  is pretty useless, and it  prevents MatPlot to be in the qcodes name space, for no good reasons.
We do not really depend on the notebook anymore (thankfully).

giulioungaretti:
Fixes #624 .
@eendebakpt // @peendebak  this fixes our slack conversation too. 

@QCoDeS/core .

### DIFF
--- a/qcodes/__init__.py
+++ b/qcodes/__init__.py
@@ -9,25 +9,20 @@ from qcodes.config import Config
 config = Config()
 
 from qcodes.version import __version__
-from qcodes.utils.helpers import in_notebook
 
-# code that should only be imported into the main (notebook) thread
-# in particular, importing matplotlib in the side processes takes a long
-# time and spins up other processes in order to try and get a front end
-if in_notebook():  # pragma: no cover
-    try:
-        from qcodes.plots.qcmatplotlib import MatPlot
-    except Exception:
-        print('matplotlib plotting not supported, '
-              'try "from qcodes.plots.qcmatplotlib import MatPlot" '
-              'to see the full error')
+try:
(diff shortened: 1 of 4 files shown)


jenshnielsen:
in_notebook is included in public.rst and should probably be removed from here too

```
Utils & misc
~~~~~~~~~~~~

.. autosummary::
   :toctree: generated/

   qcodes.utils.validators
   qcodes.utils.helpers.in_notebook
```

giulioungaretti:
@jenshnielsen thanks for catching, zapped there too.

qcodes/__init__.py
```diff
@@ -9,12 +9,10 @@
...
 from qcodes.version import __version__
-from qcodes.utils.helpers import in_notebook
 
-# code that should only be imported into the main (notebook) thread
-# in particular, importing matplotlib in the side processes takes a long
-# time and spins up other processes in order to try and get a front end
-if in_notebook():  # pragma: no cover
+
+if config.get("gui.plotlib"):
+    # load both because you never know what scientists actually want
```
MerlinSmiles: I'd like an option for only one, if you want to be scientist proof and give it the default option 'both' with the possibility to only load one of them?

giulioungaretti: cool!

giulioungaretti: although, I bet, plenty will want both.  It feels pretty ugly to have a "Both", or "All" option.

jenshnielsen:
replaced by #699

PR #629, opened 2017-05-30
Fix/buffer preflush

WilliamHPNielsen:
Fixes an outstanding issue with the SR830

- Preflush all VISA buffers before a doNd measurement

@jenshnielsen


WilliamHPNielsen:
Sorry, wrong repo.

PR #635, opened 2017-06-04
Fix: Fix reading amplitude on very old firmware (A06) of Keithley 2000

YakBizzarro:
When asking the voltage amplitude on very old firmware revisions (A06) of Keithley 2000 multimeter, the request fails if the query is in short form (without the SENSE prefix). This commit expand the query to the full form to work with old units.

@WilliamHPNielsen


WilliamHPNielsen:
@YakBizzarro This seems very reasonable. I just had a quick look in the manual, and it seems that `SENSE:DATA:FRESH?` is also a short form, a short form of `:SENSE1:DATA:FRESH?`. Would this form be even more "globally" compatible?

WilliamHPNielsen:
@YakBizzarro I'll give you another 24 hours to answer my question. If I don't hear anything, I'll just merge.

PR #636, opened 2017-06-08
feat: MatPlot updates (Updated version of #337)

### LINKED ISSUE #257: MatPlot bug

nulinspiratie:
Fixes #257, and replaces #337 since that branch was deleted.  Also includes fixes proposed in #615.

- Interpolation of x and y coords in pcolormesh. Previously, MatPlot used to ignore edgepoints.This is because MatPlot treats the x and y values as grid corners instead of grid centers. This also resulted in data points being shifted by half a grid. Now the x and y values are transformed to grid corners (which increases their arrays by 1). Doing this allows the edgepoints to be included as well. If the array only contains one element, the grid size cannot yet be determined, and defaults to 1. This also works with non-equidistant setpoints.
- Args passed in initial instance creation are treated as data that should be plotted on individual subplots. This means that `MatPlot(data.arr1, data.arr2)` will create two subplots, one for each arr. An arg can also consist of multiple args, in which case they will all be plotted on the same subplot (e.g. MatPlot([data.arr1, data.arr2], data.arr3)` will create two subplots: one for arr1 and arr2, the other for arr3).
- Easy access to subplots via subplot index (`plot[k]` is now identical to `plot.subplots[k]`). The `add` method is furthermore added to each subplot, meaning that you can now write `plot[k].add(data)`, which is identical to `plot.add(data, subplot=k)`. A consequence is that the method `_get_axes` is not needed anymore, and is therefore removed.
- Inferring subplot shape from args unless explicitly provided. MatPlot now has a class attribute max_subplot_columns (=3 by default), which indicates the default maximum number of subplot columns. If one passes 3 args and no `subplots` kwarg, it will use `subplots=(1,3)` by default. For 6 args, it will become `subplots=(2,3)`.
- Inferring figsize from subplots unless explicitly provided. The method `default_figsize()` returns a default figsize to use depending on the shape of subplots.
- Scaling of colorbar in pcolormesh
- Add method `tight_layout()`, which performs `fig.tight_layout()` with some extra space at the top for the title. It is called at the end of initialization

@giulioungaretti @jenshnielsen

### DIFF
--- a/qcodes/plots/qcmatplotlib.py
+++ b/qcodes/plots/qcmatplotlib.py
@@ -55,20 +55,19 @@ class MatPlot(BasePlot):
 
         self._init_plot(subplots, figsize, num=num)
 
-        # Add data to plot if passed in args
+        # Add data to plot if passed in args, kwargs are passed to all subplots
         if len(args) > 1:
             # Multiple args passed, add each arg to separate subplot
             for k, arg in enumerate(args):
                 if isinstance(arg, Sequence):
                     # Arg consists of multiple elements, add all to same subplot
                     for subarg in arg:
-                        self[k].add(subarg)
+                        self[k].add(subarg, **kwargs)
                 else:
                     # Arg is single element, add to subplot
-                    self[k].add(arg)
+                    self[k].add(arg, **kwargs)
         elif args:
-            # Single arg, which indicates the data, additional x and y vals
-            # can be passed as kwargs.
+            # Single data arg, which is added to the first subplot
             self.add(*args, **kwargs)
 
         self.tight_layout()
--- a/qcodes/plots/qcmatplotlib.py
+++ b/qcodes/plots/qcmatplotlib.py
@@ -54,21 +54,16 @@ class MatPlot(BasePlot):
             subplots = max(len(args), 1)
 
         self._init_plot(subplots, figsize, num=num)
-
+        
         # Add data to plot if passed in args, kwargs are passed to all subplots
-        if len(args) > 1:
-            # Multiple args passed, add each arg to separate subplot
-            for k, arg in enumerate(args):
-                if isinstance(arg, Sequence):
-                    # Arg consists of multiple elements, add all to same subplot
-                    for subarg in arg:
-                        self[k].add(subarg, **kwargs)
-                else:
-                    # Arg is single element, add to subplot
-                    self[k].add(arg, **kwargs)
-        elif args:
-            # Single data arg, which is added to the first subplot
-            self.add(*args, **kwargs)
+        for k, arg in enumerate(args):
+            if isinstance(arg, Sequence):
+                # Arg consists of multiple elements, add all to same subplot
+                for subarg in arg:
+                    self[k].add(subarg, **kwargs)
+            else:
+                # Arg is single element, add to subplot
+                self[k].add(arg, **kwargs)
 
         self.tight_layout()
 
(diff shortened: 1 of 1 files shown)


qcodes/plots/qcmatplotlib.py
```diff
@@ -75,28 +130,33 @@ def clear(self, subplots=None, figsize=None):
...
-
-            without `z` we draw a scatter/lines plot (ax.plot):
-                `x`, `y`, and `fmt` (if present) are passed as positional args
+        
+        Args:
+            use_offset (bool, Optional): Whether or not ticks can have an offset
+            
+            kwargs: with the following exceptions (mostly the data!), these are
+                passed directly to the matplotlib plotting routine.
+                `subplot`: the 1-based axes number to append to (default 1)
```
jenshnielsen: The default has changed from 1 to 0 and this is now zero based. Is that intentional?

nulinspiratie: Ah sorry I forgot to mention that. I made that change quite a while ago, but yes it was intentional. I couldn't understand why subplots started at 1, since Python is 0-based. Combining it with subplot indices (`plot[idx]`), it made more sense to me to use 0-based axes numbers. Do you agree? If so, I could modify the docstring above

jenshnielsen: In general I agree that it's most logical for all arrays to be 0 indexed. I think the reasoning here is that pyplot.subplot(nml) is 1 indexed, as in pyplot.subplot(211) gives the first subplot of the 2,1 grid. That is an artefact of matplotlibs api being inspired by Matlab. 

The biggest worry is that this will break backwards compatibility, @QCoDeS/core any opinion on this?

nulinspiratie: Ah right good point. In that case we could opt for the subplot kwarg to still be 1-based, while the indexing to be 0-based, similar to `pyplot.add_subplot` being 1-based, but the axes in `fig, axes=pyplot.subplots()` being an array, and hence 0-based. I hope this doesn't add more confusion though

jenshnielsen: Yes that is probably the best even if a bit confusing. It's consistent with matplotlib any way

WilliamHPNielsen: I would also opt for keeping consistency with matplotlib. That way I reckon we confuse the smallest amount of people to the least degree (although confusion is inescapable).

nulinspiratie: Alright just added a commit that changed this

PR #637, opened 2017-06-09
feat: add loop.plot()

nulinspiratie:
Fixes #issuenumber.

Add loop.plot() method to ActiveLoop, which automatically attaches it to the task of the loop. It cleans up the measurement code, as you can now use a single line`loop.plot(data_set.data_vals)` instead of:

```
plot = MatPlot(data_set.data_vals)
loop.attach_bg_task(plot.update)
```

When combined with PR #636 , it becomes especially easy to create a plot with multiple subplots using a single readable line. 

I didn't add it to Loop, but only to ActiveLoop. I'm not sure why we have both, is it also needed in Loop?

Currently only works with MatPlot, as I'm not sure to determine when to use MatPlot, and when to use QTPlot

@giulioungaretti @WilliamHPNielsen @jenshnielsen


astafan8:
@nulinspiratie (also @WilliamHPNielsen @jenshnielsen @giulioungaretti )
What is the status?

astafan8:
is this PR still relevant and useful? :)

WilliamHPNielsen:
I think it's fair to close this PR, since we'll do no further development of the old loop.

jenshnielsen:
Agreed

PR #641, opened 2017-06-16
Feat/channelization further work

jenshnielsen:
A few additional changes on top of the channelization

* Refactor instrument and instrumentchannel to inherit from instrumentbase, to avoid removing methods from channel
*  Convert the channels within a channel list to a named tuple, allowing them to be acesed as `
inst.channels.channelname`
* ~~Add support for allowing channels to be one indexed~~
* make sure that channels are included in print_readable_snapshot 

@spauka @WilliamHPNielsen

### DIFF
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -492,7 +492,7 @@ class Instrument(InstrumentBase):
 
         print(self.name + ':')
         print('{0:<{1}}'.format('\tparameter ', par_field_len) + 'value')
-        print('-'*80)
+        print('-'*max_chars)
         for par in sorted(snapshot['parameters']):
             name = snapshot['parameters'][par]['name']
             msg = '{0:<{1}}:'.format(name, par_field_len)
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -209,14 +209,14 @@ class InstrumentBase(Metadatable, DelegateAttributes):
                 submodule.print_readable_snapshot(update, max_chars)
 
     #
-    # shortcuts to parameters & setters & getters                            #
+    # shortcuts to parameters & setters & getters                           #
     #
     # instrument['someparam'] === instrument.parameters['someparam']        #
     # instrument.someparam === instrument.parameters['someparam']           #
     # instrument.get('someparam') === instrument['someparam'].get()         #
     # etc...                                                                #
     #
-    delegate_attr_dicts = ['parameters', 'functions']
+    delegate_attr_dicts = ['parameters', 'functions', 'submodules']
 
     def __getitem__(self, key):
         """Delegate instrument['name'] to parameter or function 'name'."""
@@ -599,6 +599,3 @@ class Instrument(InstrumentBase):
         raise NotImplementedError(
             'Instrument {} has not defined an ask method'.format(
                 type(self).__name__))
-
-    delegate_attr_dicts = ['parameters', 'functions', 'submodules']
-
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -168,6 +168,14 @@ class InstrumentBase(Metadatable, DelegateAttributes):
                 msg = msg[0:max_chars-3] + '...'
             print(msg)
 
+        for submodule in self.submodules.values():
+            if hasattr(submodule, '_channels'):
+                if submodule._snapshotable:
+                    for channel in submodule._channels:
+                        channel.print_readable_snapshot()
+            else:
+                submodule.print_readable_snapshot(update, max_chars)
+
     #
     # shortcuts to parameters & setters & getters                            #
     #
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -38,6 +38,7 @@ class InstrumentBase(Metadatable, DelegateAttributes):
         self.name = str(name)
         self.parameters = {}
         self.functions = {}
+        self.submodules = {}
         super().__init__(**kwargs)
 
     def add_parameter(self, name, parameter_class=StandardParameter,
@@ -99,6 +100,36 @@ class InstrumentBase(Metadatable, DelegateAttributes):
         func = Function(name=name, instrument=self, **kwargs)
         self.functions[name] = func
 
+    def add_submodule(self, name, submodule):
+        """
+        Bind one submodule to this instrument.
+
+        Instrument subclasses can call this repeatedly in their ``__init__``
+        method for every submodule of the instrument.
+
+        Submodules can effectively be considered as instruments within the main
+        instrument, and should at minimum be snapshottable. For example, they can
+        be used to either store logical groupings of parameters, which may or may
+        not be repeated, or channel lists.
+
+        Args:
+            name (str): how the submodule will be stored within ``instrument.submodules``
+            and also how it can be addressed.
+
+            submodule (Metadatable): The submodule to be stored.
+
+        Raises:
+            KeyError: if this instrument already contains a submodule with this
+                name.
+            TypeError: if the submodule that we are trying to add is not an instance
+                of an Metadatable object.
+        """
+        if name in self.submodules:
+            raise KeyError('Duplicate submodule name {}'.format(name))
+        if not isinstance(submodule, Metadatable):
+            raise TypeError('Submodules must be metadatable.')
+        self.submodules[name] = submodule
+
     def snapshot_base(self, update=False):
         """
         State of the instrument as a JSON-compatible dict.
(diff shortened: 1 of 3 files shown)


qcodes/instrument/channel.py
```diff
@@ -167,9 +140,10 @@ class ChannelList(Metadatable):
 
     """
 
-    def __init__(self, parent, name, chan_type: type, chan_list=None,
+    def __init__(self, parent: Instrument, name, chan_type: type, chan_list=None,
```
WilliamHPNielsen: Either we use type annotations or we don't (and I guess we are moving towards using them), but we should not mix. Also, the `__init__` function raises errors if `chan_type` and `multichan_paramclass` are not something which is more specific than `type`.

qcodes/instrument/channel.py
```diff
@@ -206,9 +181,18 @@ def __getitem__(self, i):
...
+                    raise IndexError("1 indexed channel lists only support positive indexes")
+                i = slice(i.start-1, i.stop-1, i.step)
             return ChannelList(self._parent, self._name, self._chan_type,
                                self._channels[i],
-                               multichan_paramclass=self._paramclass)
+                               multichan_paramclass=self._paramclass,
+                               oneindexed=self._oneindexed)
+        elif self._oneindexed:
+            if i < 1:
+                raise IndexError("1 indexed channel lists only support positive indexes")
```
WilliamHPNielsen: Same as above.

qcodes/instrument/channel.py
```diff
@@ -206,9 +181,18 @@ def __getitem__(self, i):
             i (int/slice): Either a single channel index or a slice of channels to get
         """
         if isinstance(i, slice):
+            if self._oneindexed:
+                if i.start < 1 or i.stop < 1:
+                    raise IndexError("1 indexed channel lists only support positive indexes")
```
WilliamHPNielsen: Merriam-Webster allows for both forms, but I prefer "indices" over "indexes".

qcodes/instrument/channel.py
```diff
@@ -279,7 +264,7 @@ def index(self, obj):
         Args:
             obj(chan_type): The object to find in the channel list.
         """
-        return self._channels.index(obj)
+        return self._channels.index(obj) + self._oneindexed
```
WilliamHPNielsen: Maybe it's just me, but I would find `self._channels.index(obj) + int(self._oneindexed)` to be more obvious.

qcodes/instrument/channel.py
```diff
@@ -296,6 +281,11 @@ def insert(self, index, obj):
             raise TypeError("All items in a channel list must be of the same type."
                             " Adding {} to a list of {}.".format(type(obj).__name__,
                                                                  self._chan_type.__name__))
+        if self._oneindexed:
+            if index < 1:
+                raise IndexError("1 based channel lists only support positive indexes")
```
WilliamHPNielsen: Same as above.

qcodes/instrument/base.py
```diff
@@ -12,7 +12,246 @@
...
+            instrument. Usually populated via ``add_parameter``
+
+        functions (Dict[Function]): All the functions supported by this
+            instrument. Usually populated via ``add_function``
+    """
+    def __init__(self, name, **kwargs):
+        self.name = str(name)
+        self.parameters = {}
+        self.functions = {}
+        self.submodules = {}
```
WilliamHPNielsen: Why does the `InstrumentBase` have submodules? Does it need them? And if so, should it not be in the docstring and there be a `add_submodule` method?

qcodes/instrument/base.py
```diff
@@ -12,7 +12,246 @@
...
+
+        par_lengths = [len(p) for p in snapshot['parameters']]
+
+        # Min of 50 is to prevent a super long parameter name to break this
+        # function
+        par_field_len = min(max(par_lengths)+1, 50)
+
+        print(self.name + ':')
+        print('{0:<{1}}'.format('\tparameter ', par_field_len) + 'value')
+        print('-'*80)
```
WilliamHPNielsen: 80 should probably be `max_chars`.

qcodes/instrument/base.py
```diff
@@ -12,7 +12,246 @@
...
+        for submodule in self.submodules.values():
+            if hasattr(submodule, '_channels'):
+                if submodule._snapshotable:
+                    for channel in submodule._channels:
+                        channel.print_readable_snapshot()
+            else:
+                submodule.print_readable_snapshot(update, max_chars)
+
+    #
+    # shortcuts to parameters & setters & getters                            #
```
WilliamHPNielsen: One space too many 😺

qcodes/instrument/channel.py
jenshnielsen: Why not, it's gradual typing after all ;) That aside it's easy to add them but the main benefit comes from annotating the `chan_type` because the linter has trouble figuring out that we are passing a Class and not an instance of a Class but sure it's easy to add the rest. 

I am not sure that there is a better way of annotating that this is a Class not an instance of the Class. type(multichan_paramclass) is not what you are looking for at least.

qcodes/instrument/base.py
```diff
@@ -522,6 +466,57 @@ def add_submodule(self, name, submodule):
...
+
+        par_lengths = [len(p) for p in snapshot['parameters']]
+
+        # Min of 50 is to prevent a super long parameter name to break this
+        # function
+        par_field_len = min(max(par_lengths)+1, 50)
+
+        print(self.name + ':')
+        print('{0:<{1}}'.format('\tparameter ', par_field_len) + 'value')
+        print('-'*80)
```
WilliamHPNielsen: This 80 character long line should probably be `max_chars` long.

jenshnielsen:
Should check how this works with the QCodes Monitor

spauka:
Just quickly, apologies for not responding earlier, is it worth having a
discussion of whether allowing 1-indexing is a good idea? It created a
syntax that is counter to what a python user would expect.

jenshnielsen:
Im not sure, I don't particularly like it either but I fear that users will be very unhappy with a situation where chan[0] maps to channel 1 on an instrument

WilliamHPNielsen:
Hmm... It's bit hard to be strongly in favour of either.

I don't see porting of code from one driver to the other ever happening easily, so I am not very concerned about that aspect of things. Instruments are in my experience always in a world of their own, and two boxes never quite behave the same way.

Having 1-indexed python lists are not nice, that is for sure. They are not too uncommon in the scientific world, though, with both Matlab/Octave and Mathematica indexing from 1, but I agree that python should be kept with one convention.

My main concern (and Jens' too, I suspect) is that users will immediately complain and/or make their own patching of the driver too be 1-indexed. And just before doing so, they will accidentally fry a sample by setting DAC channel 23 to 10 Volts using `dac.channels[22]`. Is there a good way to avoid that?

nulinspiratie:
Would a dict instead of a list solve this? In this case, there doesn't have
to be a separation between name and index, keys could even be numbers

spauka:
@nulinspiratie there is already a method to address channels by name (i.e. `dac.c0` or now `dac.channels.c0` both work). If the indexing syntax changes to a dict, i.e. `dac.channels['c0']` then slicing no longer makes sense (no way to do `dac.channels[c2:c4]`)

The fact is that adding 1-indexing creates the opposite problem for python users. Someone coming in could very easily set `c[20]` thinking it addresses channel 21!

nulinspiratie:
@spauka Ah right, but why not override slicing in `__getitem__`? As I see it, you ideally want a single way to represent your channel, by its name. Accessing multiple channels through `dac.channels['c2':'c4']` would feel more intuitive than first converting it to indices. One case where this makes sense is instruments that use channel letters instead of numbers. If you need to remember both a letter and a number for a channel, it can quickly get confusing.
I was thinking of something along the lines of 

```
class ChannelList:
    def __getitem__(self, sliced):
        print(f'getting channels {sliced.start} to {sliced.stop}')

channel_list = ChannelList()
channel_list['chA':'chC']
```
which outputs: `getting channels chA to chC`

jenshnielsen:
Agreed with @spauka and @WilliamHPNielsen  to go for the simplest solution now and only support regular python lists/tuples if the need arises we can come up with a more clever solution

jenshnielsen:
With regards to the monitor, Channel parameters are supported as well as regular parameters. 

Namely you cannot pass a MultiChannelParameter to the monitor but that is no different from a regular MultiParameter. Each channel is rendered as a new instrument which is suboptimal but probably ok for now

spauka:
self._submodules` and `add_submodule` should also be part of instrument base, as there is the possibility for nested channel lists, for example in the driver for the Harvard DAC, the list of slots also contains a list of channels, meaning the `add_submodule` function is required in `InstrumentChannel`.

jenshnielsen:
rebased and pushed submodules up to instrumentbase I think this is ready to merge

jenshnielsen:
I changed it around a bit to make it possible to index channels by name without locking the channels, I think this is ready to be merged

qcodes/instrument/base.py
WilliamHPNielsen: I guess we still need the submodules mentioned in the docstring.

WilliamHPNielsen:
If we mention `submodules` in the docstring for `InstrumentBase`, I am happy.

jenshnielsen:
@WilliamHPNielsen done

PR #642, opened 2017-06-16
Add channeled version of TPS

jenshnielsen:
@WilliamHPNielsen


jenshnielsen:
Should probably have a notebook. We may want to just delete the old version?

WilliamHPNielsen:
@jenshnielsen Yes, I think that noone ever used this driver, so we may just replace the un-channelised version with this one.

qcodes/instrument_drivers/tektronix/TPS2012_channels.py
```diff
@@ -0,0 +1,344 @@
...
+                         unit='V ',
+                         setpoint_names=('Time', ),
+                         setpoint_labels=('Time', ),
+                         setpoint_units=('s',),
+                         docstring='holds an array from scope')
+        self.channel = channel
+        self._instrument = instrument
+
+    def calc_set_points(self):
+        message = self._instrument.ask('WFMPre?')
```
WilliamHPNielsen: There are some conditions under which this call does not return the full expected response. We should ensure that these conditions are not met. (Sorry, I don't remember the details, but it's something like an enabling of the channels)

qcodes/instrument_drivers/tektronix/TPS2012_channels.py
```diff
@@ -0,0 +1,344 @@
...
+        # 'SELect?' returns a ';'-separated string of 0s and 1s
+        # denoting state display state of ch1, ch2, ?, ?, ?
+        # (maybe ch1, ch2, math, ref1, ref2 ..?)
+        selected = list(map(int, self.ask('SELect?').split(';')))
+        state = selected[ch - 1]
+        return state
+
+class TPS2012WithChannels(VisaInstrument):
+    """
+    This is the QCoDeS driver for the Tektronix 2012B oscilloscope.
```
WilliamHPNielsen: The class name and docstring should agree; either this is THE QCoDeS driver for TPS2012 or it's the channelised version.

qcodes/instrument_drivers/tektronix/TPS2012_channels.py
```diff
@@ -0,0 +1,344 @@
...
+        self.visa_handle.timeout = 1000  # 1 second as VISA counts in ms
+        gotexception = False
+        while not gotexception:
+            try:
+                message = self.visa_handle.read()
+                if verbose:
+                    print(message)
+            except VisaIOError:
+                gotexception = True
+        self.visa_handle.timeout = original_timeout
```
WilliamHPNielsen: Missing newline.

jenshnielsen:
Removed the old one and fixed som pep8 issues, We still need to fix the preamble issue

PR #650, opened 2017-06-20
Validator for callables as parameter

AdriaanRol:
- A validator for callable objects
- includes tests 

@giulioungaretti

### DIFF
--- a/qcodes/utils/validators.py
+++ b/qcodes/utils/validators.py
@@ -420,3 +420,20 @@ class Lists(Validator):
         if not isinstance(self._elt_validator, Anything):
             for elt in value:
                 self._elt_validator.validate(elt)
+
+
+class Callable(Validator):
+    """
+    Validator for callables such as functions.
+    """
+    def __init__(self):
+        # exists only to overwrite parent class
+        pass
+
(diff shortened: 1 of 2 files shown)


qcodes/utils/validators.py
```diff
@@ -420,3 +420,20 @@ def validate(self, value, context=''):
...
+class Callable(Validator):
+    """
+    Validator for callables such as functions.
+    """
+    def __init__(self):
+        # exists only to overwrite parent class
+        pass
+
+    def validate(self, value, context=''):
+        if not hasattr(value, '__call__'):
```
jenshnielsen: Any reason to not use `if not callable(value)`

jenshnielsen:
Looks good to me apart from a single inline comment

qcodes/utils/validators.py
AdriaanRol: Good suggestion, I'll make the change.

AdriaanRol:
@jenshnielsen , I addressed your comments, let me know what you think.

jenshnielsen:
@AdriaanRol Great looks good to me

Issue #600, opened 2017-05-08
Parameter improvements


nulinspiratie:
Working with parameters, I have noticed there are some features missing that I think could be a useful addition. 
I have listed them below.
The suggested improvements are mainly in the StandardParameter and the CombinedParameter.
Please let me know what suggestions you agree/disagree with, and possible ways (and pitfalls!) of implementing it.

# General
## Add get/set modifier
The `val_mapping` functionality is useful, because it allows the input to be transformed before performing the actual set/get.
The limitation here is that `val_mapping` must be a dict, and an enhancement would be to also accept functions for getting/setting.
This could be especially useful for the CombinedParameter, where it would allow you to use a single value to set a linear combination of several parameters.

## Move delay to BaseParameter
Parameters other than the StandardParameter might need a delay.
It would therefore be good to standardize the `set` method to always perform a delay if the attribute is not set to None.
This could furthermore be used for `inter_delay` (see below).
I'm not sure what would be the best approach for this.
One idea would be to add it to the BaseParameter, which then decorates the set and get commands to include delays etc.
Another way is to create a method such as `_set_with_delay` or `_set_decorated_`, which is triggered by `call()`, and which then calls a `set` and performs delays.

# Parameter
## Default slice step 
Being able to use the slice notation for parameters is very useful, as it is a quick way to create an `np.arange` list.
Currently, the slice notation requires three inputs: `[start:stop:step]`.
However, both the default slice notation and np.arange don't need the `step` argument, and use the default value 1.
To mimick this behaviour, I suggest we also use 1 as a default value for `step`.

## Accept numpy arrays for sweep points
When creating sweep values for a parameter (`param[val_list]`), `val_list` must be a list.
Often one would like to use values through numpy functions, such as `arange` or `linspace`.
Currently, parameters do not accept numpy arrays for sweep values, and so these arrays must first be converted to lists.
I suggest we also allow numpy arrays as sweep values.
As a side question, would this then need to be converted to a list internally?

# StandardParameter

## Replacing methods by dependent properties
Several methods are strongly related to getting/setting attributes (e.g. get_delay, set_delay). 
These can be replaced dependent attributes (via the @property decorator). 
This results in more readible code, and less methods cluttering up the namespace.

## Useless max_delay
From looking at the code, it seems to me that `max_delay` is not doing anything.
I've done some testing, and could not come up with a situation where max_delay actually changes something.
It is also not clear to me from the documentation what its use is.
[...]

Issue #465, opened 2017-01-25
SweepValues, Trobules ? 


giulioungaretti:
Is this the behavior one wants?

```python
from unittest import TestCase
from qcodes.instrument.parameter import ManualParameter

from qcodes.utils.validators import Numbers


class TestSweepValues(TestCase):
    def setUp(self):
        self.c0 = ManualParameter('c0', vals=Numbers(-10, 10))

    def test_slicing(self):
        c0 = self.c0
        self.assertEquals(list(c0[1:2:1]), list(c0.sweep(1,2,1)))
```
Run tests

```python
>>>AssertionError: Lists differ: [1] != [1.0, 2.0]
``` 

- The types are different (it's python so meh, but hella confusing none the less)
- The sweep values one gets  are very different 

@alexcjohnson   maybe it's by design ?

damazter:
I remember that we deliberately kept the slice notation the same as the default python slice behaviour. This is probably at the root of this problem
I am personally not surprised by this behaviour. (I cannot say anything about the types though)

alexcjohnson:
@damazter right - `.sweep` was added because some people (understandably) didn't like the missing endpoint from slice notation, or wanted the ability to specify the number of points rather than the step. But slice notation is well defined in python - and never includes the endpoint - so we don't want to alter that.

We could debate whether coercing always to float or leaving ints where it's possible is better... probably doesn't have any practical effect at the moment but definitely confusing.

giulioungaretti:
Reminds me of this:
![img_0075](https://cloud.githubusercontent.com/assets/2694199/22302383/e150829c-e32e-11e6-9fa8-beed23e1807c.PNG)
](url)

giulioungaretti:
jokes aside. I think it's really too confusing.
I will deprecate and then remove in the next version.

Background:
```python
In [8]: range(0,100)[0:100:1]
Out[8]: range(0, 100)
``` 

If one  thinks in terms of generators then slicing is also not supported.

Lastly while slicing is pythonic, it has to do with existing sequences (that's also why by default one can't really slice a generator as mentioned above).

This syntax creates a sequence by using a slice, and hence the behavior of the syntax is not predictable, unless one has been involved in the discussion. Moreover  it is impossible to document.

I am deeply sorry for those who used this syntax. 
☮

Issue #377, opened 2016-11-01
New "Configuration Parameter" 


AdriaanRol:
Based on a discussion with @damazter I would like to put forward the concept of a configuration parameter. 

What is a configuration parameter? 
* Behaves like a manual parameter but if it gets set changes flag `_configure_required` to `True`. 
* Some other parameters call a `configure()` method whenever `_configure_required==True`. 

What problem does this solve? 
* Some instruments require configuration before a measurement can be started. Examples include the ATS driver and the Signal Hound.  The configure command usually uses a set of parameters to at once upload/set some configuration. 
* The configuration parameter allows interacting with all parameters of the instrument as if they are normal parameters while simultaneously ensuring that the instrument gets reconfigured whenever required. 

I think this is a sufficiently common problem that it warrants a new parameter class. The current instruments that need it both have a different solution. 

The [signal hound driver](https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/signal_hound/USB_SA124B.py) solves this problem by using manual parameters and the requiring the user to call the configure command by hand. The ill documentation of this functionality, combined with the missing of a simple getable parameter that represents a measurement (based on my shoddy port of an existing qtlab driver) is what initiated this discussion. 

The [ATS driver](https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/AlazarTech/ATS.py) solves this problem by introducing an unsetable `AlazarParameter` that can only be changed using either the private method or the configure command. 

I would argue that the concept of a configuration parameter is the prefered solution to this problem. However, I explicitly put this forward as a point of discussion as introducing a new concept must not be done lightly. 

( @giulioungaretti, Probably intentional but I am no longer able to apply labels to issues, would make it easier to structure things a bit.)

P.S. as this issue was motivated by the low quality of a driver that I ported this raises a separate issue as to who will work on improving existing drivers, maybe an interesting subject for a discussion on Slack.

giulioungaretti:
@AdriaanRol  in terms of who will improve existing drivers, I think the answer is pretty boring: those who use them. In general it's hard to improve them without "feeling" the limitations they have.
This does not exclude asking for help / suggestions!

As for the configuration parameter, I kind of see what you mean but need to think about it a bit more!
Somehow I feel like this flag should be at driver level (as in some class attribute)  rather than a parameter.  But idk yet.

AdriaanRol:
[...]

PR #651, opened 2017-06-21
Feature/parameter improvements

nulinspiratie:
Fixes parameter issues raised in #600 (except CombinedParameter), and maybe also #498, #92. See below for list of open questions

Major parameter changes
- [x] Remove ManualParameter (it's behaviour can be captured quite easily in Parameter). Keep as deprecated parameter.
- [x] Remove StandardParameter,, as most of it properties should be used by a wider range of parameters. Its unique feature is connecting a function/VISA command to its get/set operations. This can be incorporated into the Parameter. Keep as deprecated parameter
- [x] Create wrappers for get/set that ensures additionals such as delays are always included

Other changes
- [x] Change behaviour of `parameter.get_latest()`. It now will return the latest value, unless it was obtained longer than `max_val_age` ago, in which case it will perform a new `get()`. This ensures that a value is never used that is older than `max_val_age`. Part of this includes moving `max_val_age` into GetLatest.
- [x] Remove `max_delay`, `delay_tolerance` as they were not doing anything. If anyone knows what they are supposed to  do please let me know
- [x] Modify `get/set_delay` into property
- [x] Change `delay` into `post_delay`, i.e. delay (settling time) after each `get()`
- [x] Add `inter_delay`, i.e. minimum delay between successive `get()` calls
- [x] Move delays to BaseParameter, as they can be useful for other parameter types as well
- [x] Change `get_step/set_step` to property
- [x] Move step to BaseParameter, as it could also be useful for other parameters (such as CombinedParameter or ManualParameter)
- [x] Change `_set_get/_set_set` to `_initialize_get/_initialize_set` because naming was confusing
- [x] Combine `validate_and_set`, `validate_and_sweep` to `set()`
- [x] Add `get_sweep_values` method to BaseParameter, used when performing a .set. It's default behaviour is to simply return the same value, but if step is defined, it will return a list of sweep values (taken from StandardParameter).
- [x] Move validate to BaseParameter, as it later might be used by other parameter types as well.
- [x] Accept numpy arrays as setpoints (Fixes one of the MultiParameter issues in #498, other issues need to be fixed elsewhere)
- [x] Add `scale`, which can scale (convert) its value. This can be handled in the get/set wrappers. Additionally, a `raw_value` should be an attribute which is included in the metadata
- [x] Accept numpy array for sweep points. Not sure if this should be via slicing (might be removed #465), or .sweep, although this might be too long (`param.sweep(vals=sweep_vals)` vs `param[sweep_vals]`. **Required changing SweepFixedValues**
- [x] Update tests, include new tests. 
- [x] Removal of units in (Array)Parameter, as it was already deprecated
- [x] Remove `full_name`, replace by `str(parameter)`. `full_name` is kept with deprecation warning
- [x] Combine `_latest_val`, `_latest_ts` into `_latest` dict attr, remove `_latest` method
- [x] Remove `no_getter/setter`, as they are no longer used
- [x] Only attributes in `meta_attrs` that are not None are included. As there are now more possible attrs that are by default included, this could otherwise lead to significant overhead when obtaining a snapshot of all parameters.

**Open questions**
- Should parameters include logging through the logging module? I'm asking because it might lead to excessive logging (in measurements, parameters can be set hundreds of thousands of times). 
- Should signalling be added to notify parameter event changes (#92)? This could have several useful features. One example is that instruments can connect to a parameter, such that if the parameter changes its value, the instrument can raise a flag that it needs to be setup (#377). Using a package like [blinker](http://pythonhosted.org/blinker/) works well for this purpose, but only in a single process (i.e. it does not by itself support multiprocessing). However, some sort of message passing can probably be used to incorporate this. @giulioungaretti do you think this is worthwhile? It can be easily included in the get/set wrapper
- How can I test for things like delays? Won't including delays in the TestSuite slow down testing?

@giulioungaretti @jenshnielsen @WilliamHPNielsen @MerlinSmiles


giulioungaretti:
hi @nulinspiratie  I will let @jenshnielsen  and @WilliamHPNielsen  take over this :D

jenshnielsen:
Seems sensible at a first look but I am a bit worried about removing Manual and StandardParameter there is a lot of code using it both in QCoDeS and in other modules. I think we would at least need to deprecate them for several months before removing them

AdriaanRol:
@jenshnielsen, @nulinspiratie  , removing ManualParameter and StandardParameter would break a lot of stuff for us. I think it is very important to have a deprecated mode for a while. Otherwise we get diverging branches and we can (potentially ) never update to the latest version. 

Especially since it can be quite hard to keep up to date with all changes.

nulinspiratie:
@jenshnielsen @adriaanrol good point, I will keep them with a
deprecationwarning. On a similar note, do you guys reckon enough time has
passed to remove deprecated units from (array)parameters? I don't think any
instruments have them anymore

nulinspiratie:
At this point, my intended modifications to the Parameter objects are basically done. I still have some open questions (see above), and any input would be welcome.

I had a look at the failing tests, and they are nearly all due to breaking changes (removal of deprecated `units`, etc.). Before changing these tests and adding additional tests, I think it would be good if the code can be reviewed to avoid unnecessary work ( @jenshnielsen @WilliamHPNielsen ).

nulinspiratie:
Could someone please give a review on this? @jenshnielsen @WilliamHPNielsen

jenshnielsen:
Sorry, I will do a review in the weekend

qcodes/instrument/parameter.py
```diff
@@ -94,83 +94,89 @@ class _BaseParameter(Metadatable, DeferredOperations):
...
+                 val_mapping=None, get_parser=None, set_parser=None,
+                 snapshot_value=True, max_val_age=None, vals=None):
         super().__init__(metadata)
-        self._snapshot_get = snapshot_get
         self.name = str(name)
         self._instrument = instrument
+        self._snapshot_get = snapshot_get
         self._snapshot_value = snapshot_value
 
-        self.has_get = hasattr(self, 'get')
```
nulinspiratie: Aside from unit tests which have to be modified, I found one other instance in QCoDeS that uses has_get and has_set. I changed that in the latest commit. 

I can't say if other forks of QCoDeS use has_get/set. To me it looks like an attribute that is not used much, but I could keep it as a deprecated attribute, I'm fine either way. What do you suggest?

qcodes/instrument/parameter.py
```diff
@@ -94,83 +94,89 @@ class _BaseParameter(Metadatable, DeferredOperations):
...
+        self.step = step
+        self.scale = scale
+        self.raw_value = None
+        self.inter_delay = inter_delay
+        self.post_delay = post_delay
 
-        if not (self.has_get or self.has_set):
-            raise AttributeError('A parameter must have either a get or a '
-                                 'set method, or both.')
+        # TODO (nulinspiratie) handle int vs string conversion in val_mapping
```
nulinspiratie: Yeah this is the one last thing I need to implement, I'll do it soon. 

The issue is that a val mapping can map either strings or ints. For instance, a message returned by an instrument could always be a string, even though the string would be either `'0'` or `'1'`. The old val_mapping used to attempt to convert the value into a string and then also check if it had a mapping.

qcodes/instrument/parameter.py
```diff
@@ -94,83 +94,89 @@ class _BaseParameter(Metadatable, DeferredOperations):
...
         # record of latest value and when it was set or measured
         # what exactly this means is different for different subclasses
         # but they all use the same attributes so snapshot is consistent.
-        self._latest_value = None
-        self._latest_ts = None
-        self.get_latest = GetLatest(self)
+        self._latest = {'value': None, 'ts': None}
+        self.get_latest = GetLatest(self, max_val_age=max_val_age)
+
+        if hasattr(self, 'get'):
```
nulinspiratie: The other way I can think of is to use it as a wrapper (e.g. place `@wrap_get` above `def get`). However, the issue here is that anyone who creates a custom parameter must remember to place it before each get and set, otherwise all the attributes that are in the wrapper won't have any effect. 

In case someone explicitly does not want the wrapping functionality, it can always be overridden afterwards by redefining its `get/set` method. (e.g. `self.get = CustomParameter.get`). I thought this was the better solution, as the wrapping is enabled by default. 

Potentially a metaclass could also handle this, but I thought it would add complexity and be unnecessary. But if anyone has another way please let me know

qcodes/instrument/parameter.py
```diff
@@ -203,32 +209,262 @@ def snapshot_base(self, update=False):
...
+        elif step <= 0:
+            raise ValueError('step must be positive')
+        elif isinstance(self.vals, Ints) and not isinstance(step, int):
+            raise TypeError('step must be a positive int for an Ints parameter')
+        else:
+            self._step = step
+
+    @property
+    def post_delay(self):
+        """Property that returns the delay time of this parameter"""
```
nulinspiratie: Yeah I still need to go over the docstrings and update where necessary.

nulinspiratie:
@jenshnielsen Thanks for looking over the PR. It sounds like one of the main concerns is that some of the attributes/methods may still be used. In the main QCoDeS repository, these attributes are not used (except for `max_delay`, see below). That being said, they might still be used in some of the forks. Depending on the attribute, this can be more or less likely. I think asking in Slack would be a good idea. We could then opt to keep only the attributes/methods that are likely still used as deprecated attributes, or we could keep all of them. The latter might cause quite a bit of code overhead though, especially since some of the old functionality (such as `validate_and_sweep`, `no_getter/setter`) do not fit well with how the get/set methods have been rearranged.

I can add logging to the parameters, and once all the deprecations etc have been solved I will go over the tests and docstrings


A list of parameters/attributes/methods that have been deprecated/removed:

**Removed parameters**
- `StandardParameter`: Remains as deprecated parameter, though all its functionality is encompassed in the Parameter. 
- `ManualParameter`: Remains as deprecated parameter, though all its functionality is encompassed in the Parameter. 

**Removed attributes**
- `max_delay`: Still used in QuTech IVVI, though I haven't figured out what `max_delay` actually does. We should ask why they use it.
- `delay_tolerance`: Haven't found a case where it actually does something.
- `delay`: Belonged to StandardParameter, still kept as deprecated attribute.
- `has_get` and `has_set`: With the wrapper this is replaced by `hasattr`. Might still be used in other forks.
- `full_name`: Replaced by ``str(parameter)`. Not sure if this is used, currently kept as deprecated

** Removed methods**
- `no_getter/setter`: These used to be used in the `_get/set` in StandardParameter. They are now unnecessary since the get/set is only defined in the StandardParameter if an explicit `get/set_cmd` is provided. I doubt these are ever used.
- `validate_and_set/sweep`: Functionality is now replaced by get/set_wrapper. Does not seem likely that they are used, since they were low_level functions called by get or set. Nevertheless might be worth asking on Slack
- `get/set_step`: Worth asking if it's used
- `get/set_delay`: Worth asking if it's used

AdriaanRol:
@nulinspiratie @jenshnielsen I'm using the `has_set` method when I am loading settings onto an instrument from a previously saved datafile. I use the has_set method to make sure I'm not trying to set things that cannot be set. 

I would appreciate it if this stays around for a while. If there are strong arguments to deprecate it that is fine and I can update our code accordingly.

nulinspiratie:
@AdriaanRol I'm fine keeping has_set/get as a deprecated attribute, though in the updated parameters it would suffice to check `hasattr(parameter, 'set')`

nataliejpg:
This looks cool, a few thoughts:

We use the get/set_delay/step quite a lot, seems fine to move over to properties though. Could you expand a bit on what the difference is between post_delay and inter_delay? Thanks

Did you remove 'set_validator' or am I missing something? This is pretty useful if you are using a driver for a number of instruments with different limits etc and it is unnecessary to change the driver but you want to set the limits on the fly (or if you just want to set some limits for safety in that particular experiment). Is there a way to do this in this pr?

Also I'm a bit confused about the wrapping order when a get_cmd and/or set_cmd is specified in the initialisation of a Parameter. I thought you intended the wrapping to be done by calling super.__init() but then you don't appear to call that first and do the get_cmd, set_cmd logic afterwards and do the wrapping there. 

I like the idea of removing ManualParameter but it's also something we've used quite a lot so I'd definitely like a bit of time to move things over and check nothing breaks!

nulinspiratie:
Hi @nataliejpg, thanks for looking over the PR

The difference between post_delay and inter_delay is the following.
post_delay can be thought of as a settling time, e.g. after changing a magnetic field you want to wait 1 minute before any further actions are taken. This occurs after each set operation.
inter_delay does not take effect after every set operation, but only if the previous set occured less than this time difference. For example, if you don't want commands to be sent too quickly to the instrument. inter_delay can also be preferred if you don't want a delay to occur every time you set a parameter, but only if it is actually ramping to a value. Does this clarify its use?

The `vals` isn't removed, but moved to the Base parameter, this way it can be used by other Parameter types as well.

As for the wrapping of `get/set_cmd`, they need to be created after calling `super().__init__`, since they depend on attributes initialized there. However, since `get()` and `set()` did not exist when calling `super().__init__`, no wrapping occurred. That's why they need to be explicitly wrapped afterwards.

And `ManualParameter` will stay as a deprecated parameter to help transition to the new Parameter.

AdriaanRol:
@nulinspiratie We use the `manualparameter` a lot. I understand it will be deprecated. I was wondering how it works, is manual paramterness now some implicit behaviour for standard parameters?  Your description says it *can* be quite easily captured, not how it is captured. 

In some sense I liked the explicit nature of the `ManualParameter` as per the python zen, but I can understand that less classes is also nice.

nulinspiratie:
@AdriaanRol In the new parameter, a get_cmd and set_cmd can be passed. These are used if the Parameter does not have an explicit get/set method, and can be set to four things:
- `False`: The Parameter cannot be set if `set_cmd=None` (same for get)
- `None`: The parameter will store to (`set_cmd=None`) or load from (`get_cmd=None`) an attribute. If both `set_cmd` and `get_cmd` are set to None, this would correspond to the ManualParameter
- string: Used as visa command, like its current use in the StandardParameter
- callable: called upon get/set, like its current use in the StandardParameter.

For the parameter, the default kwargs are: `get_cmd=None`, and `set_cmd=False`. For the deprecated ManualParameter and StandardParameter, these are different.

By being able to set the `get/set_cmd` to any of these independently, Parameters can also be created that are part `ManualParameter` and part `StandardParameter`. One thing we ran into a few times is an Instrument whose VISA parameters are settable, but not gettable. In some cases we'd want the `get` method to return the latest set value. This wasn't easily doable before, but is now.

AdriaanRol:
@nulinspiratie Cool, very nice improvement. 

I was wondering, is this reasoning also documented in the get/set cmd explanation? I think that would be useful.

nulinspiratie:
@adriaanrol yeah the documentation still needs to be updated. I first
wanted to have a review round to see if significant modifications were
still necessary. I'm planning on updating the documentation soon though

nataliejpg:
@nulinspiratie  thanks for this explanation, I think I get the point of having two kinds of delay but to clarify:
post_delay: don't do anything for x seconds after setting this parameter
inter_delay: if you have set this parameter recently you aren't allowed to set it again for x seconds (and then only by a difference of step)
Is that right? And for inter_delay is it just setting which you 'wait to do' or would I also not be able to call get or something else in this time? thanks

Also for the life of me I could not fine 'set_validator' in _BaseParameter. I get that vals is still there so you can set the validator on initialisation but I couldn't see a way to change it or set it after that which is something we do quite a bit. Maybe I've missed it and you could tell me where it is or how I would set the Validator after an instance of the Parameter is created.

Otherwise I really like it 👍

nulinspiratie:
@nataliejpg your description of the two delays is correct! By the way the step is independent of the delays, so also with a post_delay it could be maximally changed by a step. It uses a simple time.sleep, so no fancy multithreading. If this is a feature people want we could add it later in an enhancement PR.

Also, vals is an attribute and can be changed by changing the attribute value, i.e. parameter.vals=new_validator

nataliejpg:
@nulinspiratie fab, now I'm with you. Thanks

nulinspiratie:
I've added the tests for the new features of the parameter, and fixed the ones that were broken due to the changes. I've also updated the docstring. Everything seems to be working fine, although I think it definitely still needs a review @jenshnielsen .

If it all looks good, what's the plan after? Since all instruments rely on StandardParameters/ManualParameters, plus all the tests, would it make sense to change all of them in all the files, and then make a PR to this branch before merging this PR?

jenshnielsen:
@WilliamHPNielsen Do you have time to review this too

AdriaanRol:
@nulinspiratie 
>Since all instruments rely on StandardParameters/ManualParameters, plus all the tests, would it make sense to change all of them in all the files, and then make a PR to this branch before merging this PR?

I would really appreciate it if you can create some legacy classes to implement the `ManualParameter` and `StandardParameter`. These should be subclasses of your new parameter that work exactly as the old parameters but with a deprecation warning on instantiation. As we for instance depend heavily on QCoDeS and a change like this would break backwards compatibility big time. 

A grace period to transfer over would be greatly appreciated and would prevent us from being locked out of future version.

jenshnielsen:
@AdriaanRol I think there are already deprecated methods in place for standard and manual parameter. Before merging this I want to make sure they work as expected

WilliamHPNielsen:
Okay, I am working through the `driver_examples` for instruments in our test setup.

qcodes/instrument/parameter.py
```diff
@@ -94,84 +93,132 @@ class _BaseParameter(Metadatable, DeferredOperations):
...
+            after the *start* of each set, whether part of a sweep or not. 
+            Can be set to 0 to go maximum speed with no errors.
+        
+        val_mapping (Optional[dict]): a bidirectional map data/readable values
+            to instrument codes, expressed as a dict:
+            ``{data_val: instrument_code}``
+            For example, if the instrument uses '0' to mean 1V and '1' to mean
+            10V, set val_mapping={1: '0', 10: '1'} and on the user side you
+            only see 1 and 10, never the coded '0' and '1'
+            If vals is omitted, will also construct a matching Enum validator.
```
WilliamHPNielsen: As far as I can see, the "matching `Enum` validator" is never set, and the parameter thus gets the default `Numbers` validator. This breaks (amongst others) a handful of parameters using a `{'ON': 1, 'OFF': 0}` val mapping.

jenshnielsen:
Other than Williams issue I think this is nearly ready to go. 

Could you strip the trailing whitespaces from parameter.py 

There are a few rst related warnings which will be nice to fix

```
/home/travis/build/QCoDeS/Qcodes/qcodes/instrument/parameter.py:docstring of qcodes.Parameter:8: WARNING: Unexpected indentation.
/home/travis/build/QCoDeS/Qcodes/qcodes/instrument/parameter.py:docstring of qcodes.Parameter:12: WARNING: Unexpected indentation.
/home/travis/build/QCoDeS/Qcodes/qcodes/instrument/parameter.py:docstring of qcodes.Parameter:13: WARNING: Block quote ends without a blank line; unexpected unindent.
```

nulinspiratie:
@WilliamHPNielsen Good catch, I'll fix that, along with the whitespaces and
rst warnings

nulinspiratie:
@WilliamHPNielsen the validator is now set by default to Enum if val mapping is provided.
I modified one last thing: the Parameter used to have `vals=Numbers()` by default, which I changed to None. I think there will be too many cases when someone wants to use something else, such as a string. This could save quite some headache.

I also updated the docstrings, removed trailing whitespaces etc. I think I've done all I can do now. @jenshnielsen  how do we proceed?

jenshnielsen:
@nulinspiratie looks good. There is a test for the default validator being number that needs to be changed or removed

nulinspiratie:
@jenshnielsen fixed!

WilliamHPNielsen:
Running through the driver examples again to check if stuff works.

Keithley 2600... Passed with warnings (`ManualParameter` deprecated, `full_name` deprecated)

Keysight 33500B... Passed with warnings (`ManualParameter` deprecated)

QDac Channels... Passed with warnings (`ManualParameter` deprecated, `MultiParameters` do not support `set`). Note that the celebrated channel features do not really work.

SR830... Failed because of `set_validator` in driver.

Tektronix AWG5014C... Passed with warnings (`ManualParameter` is deprecated)

WilliamHPNielsen:
Upshot from the above test: we need to fix the SR830 driver. The channelised QDac warnings are not introduced by this PR, so we should fix them separately.

My verdict: 💃 !

jenshnielsen:
I would prefer not to break any drivers

Is there any problem with restoring `set_validator` as deprecated 

```
def set_validator(self, vals):
    """
        Deprecated Set a validator `vals` for this parameter.
            Args:
                vals (Validator):  validator to set

    """
    warnings.warn(
        "set_validator is deprected use `inst.vals = MyValidator` instead")
    if isinstance(vals, Validator):
        self._vals = vals
    else:
        raise TypeError('vals must be a Validator')
```

I tend to thing that Deprecation warnings should not use the logging module but do warnings.warn

WilliamHPNielsen:
@jenshnielsen, I think it's fine to deprecate as you suggest. Note that `self._vals` has become `self.vals`. D'you wanna push that change?

jenshnielsen:
Yes let me do that now

jenshnielsen:
@WilliamHPNielsen pushed, from my point of view this is ready to land

PR #653, opened 2017-06-21
Driver/buffered sr830

WilliamHPNielsen:
Fixes the fact one can not do buffered acquisition

- Add two array parameters representing the internal data buffer

@jenshnielsen


qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -1,5 +1,103 @@
...
+class ChannelBuffer(ArrayParameter):
+    """
+    Parameter class for the two channel buffers
+
+    Currently always returns the entire buffer
+    TODO (WilliamHPNielsen): Make it possible to query parts of the buffer.
+        The instrument natively supports this in its TRCL call.
+    """
+
+    def __init__(self, name, instrument, channel):
```
jenshnielsen: I suggest that we start moving the typos from the docstring into type annotations. I.e. 

```
def __init__(self, name: srt, instrument: SR830, channel: int):
```
and so on.

 It's imho fine to leave the types in the doc strings too if you prefer.

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -1,5 +1,103 @@
...
+        """
+        Args:
+            name (str): The name of the parameter
+            instrument (SR830): The parent instrument
+            channel (int): The relevant channel (1 or 2). The name should
+                should match this.
+        """
+        if channel not in [1, 2]:
```
jenshnielsen: I would make this an parameter and probably a tuple i.e. `self._valid_channels = (1,2)`

qcodes/instrument_drivers/stanford_research/SR830.py
WilliamHPNielsen: I'd say we get rid of all typos altogether! ;)

I totally agree with regard to the types.

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -1,5 +1,103 @@
...
+        """
+
+        N = self._instrument.buffer_npts()  # problem if this is zero?
+        # TODO (WilliamHPNielsen): what if SR was changed during acquisition?
+        dt = 1/self._instrument.buffer_SR()
+        self.shape = (N,)
+        self.setpoints = (tuple(np.linspace(0, N*dt, N)),)
+
+        params = self._instrument.parameters
+        # YES, it should be: "is not 'none'" NOT "is not None"
```
jenshnielsen: Is there a particular reason why you cant change it to None in the mapping dicts around l 467?

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -235,11 +334,23 @@ def parse_offset_get(s):
...
+                               label='Channel {} display'.format(ch),
+                               get_cmd=partial(self._get_ch_display, ch),
+                               set_cmd=partial(self._set_ch_display, ch),
+                               vals=Strings())
+            self.add_parameter('ch{}_databuffer'.format(ch),
+                               channel=ch,
+                               parameter_class=ChannelBuffer)
 
         # Data transfer
         self.add_parameter('X',
```
jenshnielsen: How do these work, Do they always return from a specific channel?

qcodes/instrument_drivers/stanford_research/SR830.py
WilliamHPNielsen: I think conceptually there is a difference; the `'none'` is the proper response string from the instrument. I would expect the python `None` to mean that the value doesn't exist, whereas `'none'` in this case rather means OFF than "not defined" or "not present".

qcodes/instrument_drivers/stanford_research/SR830.py
WilliamHPNielsen: Yes. The internal buffer is filled simultaneously for both channels, which is why there is only one `buffer_SR` etc., but polling is done from either of the two (but not both).

qcodes/instrument_drivers/stanford_research/SR830.py
WilliamHPNielsen: Fair enough. Seems like better practice.

qcodes/instrument_drivers/stanford_research/SR830.py
jenshnielsen: Fair enough, perhaps we could avoid the confusion by calling it one of those names?

WilliamHPNielsen: `'none'` is what the manual calls it, but I'm fine with `'off'` or even `'identity'`.

jenshnielsen: I don't have any strong opinion. It's probably better to match the manual then

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -1,5 +1,105 @@
...
+class ChannelBuffer(ArrayParameter):
+    """
+    Parameter class for the two channel buffers
+
+    Currently always returns the entire buffer
+    TODO (WilliamHPNielsen): Make it possible to query parts of the buffer.
+        The instrument natively supports this in its TRCL call.
+    """
+
+    def __init__(self, name: str, instrument: SR830, channel: int):
```
jenshnielsen: You might need to do quote i.e. do 'SR830' as a forward declaration because it's defined below the ChannelBuffer and not available when the class is defined

WilliamHPNielsen: Check.

PR #661, opened 2017-06-29
Protect against crashed qt remote processes

jenshnielsen:
This may happen if something crashes the remote process used by pyqt graph, which can be as mundane as an out of bounds access somewhere triggered by a gui action. Previously this required restarting the kernel/interpreter. This fixes that by:

* Automatically restart the process if creating a window fails
* make _init_qt a class method. Now it can be called manually as `QtPlot._init_qt()` before creating the class.

We should also fix any crashes caused by gui actions

@WilliamHPNielsen


jenshnielsen:
merged via #662

PR #664, opened 2017-07-11
Feature/dataset

giulioungaretti:
Introducing  the new dataset and experiment container.

Check out the examples. And let me know.

There are some though areas still:

- [ ]  as of now every time a result is inserted with add_result the db is saved same for add_results. IS THIS THE BEHAVIOUR WE WANT?

- [ ] storing parameters in separate table as an extension (dropping  the column parametenrs would be much nicer

- [ ] metadata split between well known columns and maybe something else is  not such a good idea. The problem is if we allow for specific columns then  how do the user/us know which are metatadata?  I THINK the only sane solution is to store JSON in a column called metadata (see beautiful sketch below)

screen shot 2017-07-11 at 15 54 09

- [ ] we cant have parameters with the same name in the same dataset/run (I think Its not a big deal, but different than what's now).


giulioungaretti:
note:  the subscribing features only work if one is using the https://github.com/QCoDeS/Qcodes/tree/feature/logging branch!

nulinspiratie:
@giulioungaretti From the examples in the notebook it looks like a nice improvement over the current data storage methods. Nicely done!

One thing I didn't understand is if it's now saved in a single database file, or still in one file/folder per measurement, like is being done currently. If it's all saved in one file, it might give problems with backups, as the file will continuously change, and a new backup needs to be made every time

peendebak:
Could someone of the dev team comment on the status of the current PR, and development in the next period? I tested the dataset quickly, but have some issues that need to be addressed before we could consider using it.

- The new dataset works only with python 3.6. Is there a reason for this? We will not convert current (python 3.5) installations

- Can we load old datasets and convert them into the new format? 

- Storing metadata does not work for simple types, e.g. `dataSet.add_metadata('a', [10., 4.])` fails

- How is the experiment container supposed to scale? I created a new experiment with 100 empty datasets. Then from a new console
```
import codes
exp=qcodes.load_last_experiment()
exp.data_set(10)
```
works fine. If I do the same for an experiment with of 1000 datasets the system hangs forever.

- Can a dataset be used without an experiment? 

- How can I save a dataset? It seems to save automatically to the experiment. Is there an option to save a single dataset to disk?

- Can a dataset be serialized? So we can store a dataset in the metadata, or just store a dataset together with other data

giulioungaretti:
@nulinspiratie it's one file, and yes it will be like that if you store it in say dropbox // gdrive. (But it was the same before I think).

BUT and this is also for @eendebakpt one can write their own measurements procedures and save only at special points (say after 100 seconds // xxxx points) which speeds things up.

giulioungaretti:
- The new dataset works only with python 3.6. Is there a reason for this? We will not convert current (python 3.5) installations

There are two reasons, types and f strings. Without the latter writing sql is hard and error prone. 

- Can we load old datasets and convert them into the new format? 

I did not write a converter, but should not take more than 10 lines of code. 

- Storing metadata does not work for simple types, e.g. `dataSet.add_metadata('a', [10., 4.])` fails

That's correct, list are not supported ATM, arrays will work tho. This is left like this on purpose because it was not decied how to store the metadata (i.e. a json string vs a binary blob) .

- How is the experiment container supposed to scale?
It scales as good as sqlite. I can't reproduce your issue on my machine. 
The db file produced with the revision I pushed does not optimise anything (which one could do).

- Can a dataset be used without an experiment? 

Yes, because there will be always one experiment.  There is a note in the code on how to achieve this.

- How can I save a dataset? It seems to save automatically to the experiment. Is there an option to save a single dataset to disk?

see this note https://github.com/QCoDeS/Qcodes/blob/feature/dataset/qcodes/dataset/data_set.py#L26

- Can a dataset be serialized? So we can store a dataset in the metadata, or just store a dataset together with other data

A dataset is serialised already (being part of the sqlite database). 
If the question  is how to share a file that maps to  a single dataset, that is not implemented but is be rather trivial.

nulinspiratie:
@giulioungaretti I see, having the entire database as a single file might lead to issues with data backups/archival though. Since every dataset currently has its own file, this makes data archival easy, since only the new datasets need to be uploaded. If all datasets are stored in a single database file, it seems to me that its no longer possible to only upload new datasets for archival, and instead the whole database would need to be stored every time. Is this correct, or is there a way around this, such as to segment the database into discrete files?

As for your last checkbox:
` we cant have parameters with the same name in the same dataset/run (I think Its not a big deal, but different than what's now).`

Our measurements often use the same parameter multiple times in a single measurement. For example, one of our more complex parameters does a retuning sequence. If our measurement has several stages, we often add this parameter in between to ensure that the system remains tuned between stages. For us it's quite important to not be restricted to using parameters at most only at one point in a measurement

giulioungaretti:
@nulinspiratie no, there is no way around this. The file contains much more info than a collection of single files which is what allows smart things like counters and fast searches. BUT one can definitely do something smart like:  adding features to save just the new things to a different place (if that sound something you'd want to explore: you can use a trigger on an insert on the experiment / runs table that pushes the new stuff into a different database. OR http://www.sqlite.org/c3ref/update_hook.html ).

jenshnielsen:
Does anyone have experience with http://litereplica.io/ sounds like it would be a just what we need. Its opensource but under AGPL licence so we probably need a commercial licence (which is not to expensive)

WilliamHPNielsen:
I think the context manager is done now. By "done" I mean "ready for next round of review". It has tests and a relatively sane API.

quantumkoen:
Hi, sorry to interrupt, but would it be at all possible to avoid having the DataSet having a hard dependency on sqlite? This seems to violate the valued principle of loose coupling.

I'll have to look into it in a bit more detail (only stumbled upon this PR now), but instead of hard-coding  calls to imported non-instance functions in the DataSet object, wouldn't it make more sense to use the dependency-injection design pattern? Ie, give the constructor an object that is responsible for the persistence of the DataSet, instead of relying on non-object methods imported from the sqlite_base module. So basically, have an object that represents the sqlite3 db, and define methods on that object that are now functions in the sqlite_base module.

That way, the whole DataSet class becomes more future-proof and allows for customisation by simply giving the constructor another object with the same interface but for example writing to mongodb, memory, hdf5 or whatever. Doing things this way breaks the dependency of DataSet on sqlite_base, which makes the code more maintainable and testable, and thus more future-proof and stable.

The same I think can be said for Experiment class.

WilliamHPNielsen:
@quantumkoen Quick answer: no, too late. But I think you point is valid and I'd like to have this discussion about decoupling the new developments from any particular backend. Could you please repost you concern as an issue?

sohailc:
While waiting for the data set module to complete, I developed an interim
solution myself which I think looks much more like what Koen in suggesting.
Have a look at this:

https://github.com/sohailc/pysweep2.0/blob/master/pysweep/data_storage/base_storage.py

The class DataSet is meant to be sub-classed, for instance, to save in the
Pandas format as shown here:

https://github.com/sohailc/pysweep2.0/blob/master/pysweep/data_storage/pandas_storage.py

Or as a Spyview/GnuPlot file, as shown here:

https://github.com/sohailc/pysweep2.0/blob/master/pysweep/data_storage/spyview.py

PR #671, opened 2017-07-14
add catch for network timeouts

peendebak:
Sometimes the slack bot generates an error due to a network timeout. This PR prevents this from happening by catching a timeout error.

@giulioungaretti @nulinspiratie


nulinspiratie:
Nice! I ran into the same problem a few times. I think the specific error that is raised is 
`from requests.exceptions import ReadTimeout`
It might be worthwhile to catch that particular error. Also, perhaps sending the exception to a logger is more useful.

peendebak:
@nulinspiratie Thanks for the idea, I changed the code.

PR #675, opened 2017-07-18
add one more exception to the catch in slack module

peendebak:
@WilliamHPNielsen @jenshnielsen @nulinspiratie


nulinspiratie:
@peendebak Good catch, I hadn't had that one before

PR #681, opened 2017-07-24
better warning handling for slack class

peendebak:
@jenshnielsen @nulinspiratie

-    Add a custom SlackTimeoutWarning so one can filter the slack timeout messages
-    Make the default error logging less verbose
-    Add one more type of error to the except


nulinspiratie:
@peendebak Haha I'm starting to regret my comment on catching a specific error, I didn't realize that there were several possible errors. Looks good though!

PR #716, opened 2017-08-21
Foreground qt window

jenshnielsen:
Using pywin32


qcodes/plots/pyqtgraph.py
```diff
@@ -95,6 +98,7 @@ def _init_qt(cls):
         pg.mkQApp()
         cls.proc = pgmp.QtProcess()  # pyqtgraph multiprocessing
         cls.rpg = cls.proc._import('pyqtgraph')
+        cls.qc_helpers = QtPlot.proc._import('qcodes.utils.helpers')
```
Dominik-Vogel: QtPlot should be cls

PR #719, opened 2017-08-25
Fix: ip instrument log empty responce and ensure all data is send

### LINKED ISSUE #748: IPInstrument._send() could not send all its data
I was browsing the instruments code and found out that in the `IPInstrument` class the `socket.send(...)` function is implemented wrongly:

```python
    def _send(self, cmd):
        data = cmd + self._terminator
        self._socket.send(data.encode())
```

This is what the documentation says about `socket.send(...)`:

> socket.send(bytes[, flags])
Send data to the socket. The socket must be connected to a remote socket. The optional flags argument has the same meaning as for recv() above. Returns the number of bytes sent. Applications are responsible for checking that all data has been sent; if only some of the data was transmitted, the application needs to attempt delivery of the remaining data. 
https://docs.python.org/3.6/library/socket.html

At this moment, if `send(...)` fails, only a part of the message is transmitted. Which will create strange bugs.

A better solution is to use `socket.sendall(...)` or as the example shows:
```python
    def mysend(self, msg):
        totalsent = 0
        while totalsent < MSGLEN:
            sent = self.sock.send(msg[totalsent:])
            if sent == 0:
                raise RuntimeError("socket connection broken")
            totalsent = totalsent + sent
```
https://docs.python.org/3.6/howto/sockets.html#socket-howto

jenshnielsen:
As I understand it that is a sign of a closed socket
This should probably be an error in the future

Also use sendall to ensure that we dont drop data. Fixes #748

### DIFF
--- a/qcodes/instrument/ip.py
+++ b/qcodes/instrument/ip.py
@@ -1,8 +1,10 @@
 """Ethernet instrument driver class based on sockets."""
 import socket
+import logging
 
 from .base import Instrument
 
+log = logging.getLogger(__name__)
 
 class IPInstrument(Instrument):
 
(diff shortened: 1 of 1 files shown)


jenshnielsen:
@WilliamHPNielsen Note that this also now includes the fix for #748

WilliamHPNielsen:
Ah, even better.

PR #721, opened 2017-08-31
Cleanup dependencies

jenshnielsen:
- [x] Update setup.py information, maintainer and supported versions
- [x] Remove dependency on Jupyter. QCodes can and is used without Jupyter
- [x] Bump pip installed pyqt to 5.9.1
- [x] Bump numpy to 1.13.1


peendebak:
@jenshnielsen Is it necessary to increase the version number of numpy (and also some of the other packages)? Numpy 1.13 was only released very recently, so not many people will have this installed. I think this will trigger updates for many systems when a new version of qcodes is installed.

jenshnielsen:
The version given in requirements.txt is the one installed when you do `pip install -r requirements.txt` (as travis does) the version installed when pip installing qcodes is defined in setup.py.  It's unchanged from before (numpy>=1.10)

PR #724, opened 2017-09-05
Driver for tektronics awg5200

peendebak:
The driver is copied from the 5014 which is largely compatible. Changed:

- the number of channels in the driver file (which is 8 for the 5200).
- sending the waveform and marker to the device, which is a little bit different from the 5014

Not all new options of the new 5200 are supported, and not all the old options have been tested, but the device seems to work fine.

@christian-volk @jenshnielsen 

Note: we driver could be combined using a shared base class, but this is not something we will implement in the near future.


WilliamHPNielsen:
> Not all new options of the new 5200 are supported, and not all the old options have been tested [...]

That kind of leaves me *not* wanting to put this driver in the code base. 

A quick diff reveals that you only changed the number of channels and made an "integer" into a "real" in the internal AWG file format. Wouldn't a subclassing of the AWG5014 then suffice for now? You'd just need to actually use the `self.num_channels` in the AWG5014 driver (and eventually add a `self.num_markers`) and then override `send_waveform_to_list`.

peendebak:
@WilliamHPNielsen I changed the driver to reuse the 5014 code. Full testing of the driver will not happen soon, we will only use the parts of the driver that we need for our experiments.

PR #725, opened 2017-09-06
moved files from lower to upper case Keysight folder

Dominik-Vogel:
# Problem: separate 'Keysight' and 'keysight' folders in qcodes/instrument_drivers

From @jenshnielsen I learned that on Windows and MacOsX the path queries to the file system are case insensitive, while on Linux they are not. You can see the two folder in the repository as it is presented by the github page.
So on a Windows machine there is only one visible Keysight folder (the capitalized one), that contains the files of both folders.
A possible scenario in which this could have happened, is that a user created locally a folder on a windows machine, and subsequently merged with a new version from main repository, in which the same  folder has been added but with a different case choice.

### Damage
This pull request simply moves files from the lower case keysight folder to the upper case one. This choice minimized the damage by only breaking the code that uses the affected drivers for Linux systems. 

### New Standard?
I propose to discuss the possibility of a standard for the use of capital letters for all new folders in the Qcodes repository.


jenshnielsen:
This has the issue that when I do a fresh checkout on an case insensitive machine the Keysight/__init__.py file is missing because that is really the same file as keysight/__init__.py which is deleted. And you end up in a state where you checkout the branch and you automatically have one change not committed (the deletion) of Keysight/__init__.py

jenshnielsen:
What happened to this? I think we should move all the files from the lower case folder to the upper case one and leave the ```__int__.py``` file there along with a note to put files in the upper case one

Dominik-Vogel:
Sorry, that was my incapacity.. I didn't realize, that the branch was in the PR status and deleted it when I cleaned up.... I have learned this for now :-)

jenshnielsen:
@lgnjanssen This moves your driver from keysight to Keysight folder. On case insensitive os's (OSX 
 and windows) this should not break anything but if you are on linux this may break the import of the driver

jenshnielsen:
@maij Thanks, not sure how it happened. The two folders may have been created in parallel from two different prs

Issue #53, opened 2016-03-04
cannot Loop with a real instrument


alexcjohnson:
The prize goes to @MerlinSmiles for being the first to actually try a data taking loop with a real instrument... and discovering that they can't be pickled to send to the measurement process. In hindsight this makes sense, otherwise you'd be able to make competing connections where only one is allowed.

I was hoping to avoid making a separate process for every instrument, as every new process makes the system more brittle and makes it harder to recover from errors, but at this point that's the only solution I see. I don't think this will be that hard, actually. Here's what I'm thinking:
- On instantiation, we don't make any hardware calls, but we start a new process that connects and then just sits there forever listening to a queue.
- Other methods that _do_ talk to the hardware get a decorator, actually I think we'll need two, something like `@ask_instrument` and `@write_instrument` depending on whether we're supposed to wait for a return or not. These decorators look at an attribute like `self._on_server` - if it's called on the server it executes the function, but if not it passes the function call through the queue
- For some specific cases we may want to short-circuit the decorator logic for a little performance boost. Like I bet we can handle everything in visa instruments just explicitly wiring `.read`, `.write`, and `.ask` to the queue. But that's an optimization we can work out later.
- The server should also hold all instrument saved state (ie calls to `.get_latest` for any `Parameter` that's connected to an instrument) so that different processes will agree on the state.

Sound reasonable? Anyone see a better option?

PR #730, opened 2017-09-08
Faster qdac (#53)

jenshnielsen:
* add support for faster setting of qdac voltage
* add option to qdac to ignore return read on voltage set
* SR830 fail early in buffered readout if butffer is not full as expected

@WilliamHPNielsen

### DIFF
--- a/qcodes/instrument_drivers/QDev/QDac.py
+++ b/qcodes/instrument_drivers/QDev/QDac.py
@@ -156,6 +156,22 @@ class QDac(VisaInstrument):
                            set_cmd='ver {}',
                            val_mapping={True: 1, False: 0})
 
+        self.add_parameter(name='fast_voltage_set',
+                           label='fast voltage set',
+                           parameter_class=ManualParameter,
+                           vals=vals.Bool(),
+                           initial_value=False,
+                           docstring=""""Toggles if DC voltage set should unset any ramp attached to this channel.
+                                     If you enable this you should ensure thay any function generator is unset
+                                     from the channel before setting voltage""")
+        self.add_parameter(name='voltage_set_dont_wait',
+                           label='voltage set dont wait',
+                           parameter_class=ManualParameter,
+                           vals=vals.Bool(),
+                           initial_value=False,
+                           docstring=""""If enabled the voltage set does not wait for a reply from the 
+                           qdac. This is experimental and may break but has the potential to speed up
+                           stepping the qdac""")
         # Initialise the instrument, all channels DC (unbind func. generators)
         for chan in self.chan_range:
             # Note: this call does NOT change the voltage on the channel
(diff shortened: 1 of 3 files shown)


qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -553,7 +570,8 @@ def write(self, cmd):
 
         nr_bytes_written, ret_code = self.visa_handle.write(cmd)
         self.check_error(ret_code)
-        self._write_response = self.visa_handle.read()
+        if not fast:
```
WilliamHPNielsen: What happens to the response if `fast`? Isn't it still in some output queue of the QDac?

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -108,7 +108,8 @@ def get(self):
         # parse it
         realdata = np.fromstring(rawdata, dtype='<i2')
         numbers = realdata[::2]*2.0**(realdata[1::2]-124)
-
+        if self.shape[0] != N:
```
WilliamHPNielsen: This looks like a good precaution (and I guess it's indeed related to fast measurements with the QDac), but shouldn't it be a seperate PR?

jenshnielsen: Maybe it should but the commits have already been squashed because its a backport and I don't really see the point of splitting them

qcodes/instrument_drivers/QDev/QDac.py
jenshnielsen: I think this in general is a bad idea but they wanted it. Hence off by default and the warning in the doc string

qcodes/instrument_drivers/stanford_research/SR830.py
WilliamHPNielsen: Fair enough.

jenshnielsen:
Need to implement this for the channel version too

jenshnielsen:
Should be ready now

qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -553,7 +562,6 @@ def write(self, cmd):
 
         nr_bytes_written, ret_code = self.visa_handle.write(cmd)
         self.check_error(ret_code)
-        self._write_response = self.visa_handle.read()
```
WilliamHPNielsen: Woops! We need to still get the `_write_response`.

jenshnielsen:
We could do this in a smarter way perhaps.

Concatenate the two write commands and then:
Make write take an optional number of newlines to read back and either 

* Read n times at the end of write
* Temporary change the termination to `m * \n` where m is the number of commands written. 

I suspect that the first method may be solver but without benchmarking it's hard to know.

PR #733, opened 2017-09-12
Unit autoscale matplotlib and pyqtgraph graphs

jenshnielsen:
@WilliamHPNielsen 

Add methods to autoscale graphs backported and improved from qcodes-dk

Must currently be called manually


WilliamHPNielsen:
Nice stuff. But when running a standard QCoDeS loop, the `fixUnitScaling` function should be called before the plot is dumped as a .png (and preferably before that). I guess it would make sense to include it in the updater for the plot?

jenshnielsen:
Done. I have put it in the create function and added support for rescaling the limits. I think this is better than the update both because it's waist of time to do it at every update and because it might change the limits during the plot which I doubt is a good idea. Note that rescaling of non array parameters to the correct limits requires you to pass them to the function. This is not done automatically as the plot does not know the limits before the data is acquired

PR #737, opened 2017-09-14
Smarter way of speeding up qdac

jenshnielsen:
As suggested by @WilliamHPNielsen 
Rather than disabling the second write this concatenates the commands and counts the numbers 
of reads to perform afterwards.

Both the new and old way has the same speedup from 32 to 16 ms for a single set command

### DIFF
--- a/qcodes/instrument_drivers/QDev/QDac_channels.py
+++ b/qcodes/instrument_drivers/QDev/QDac_channels.py
@@ -645,7 +645,6 @@ class QDac(VisaInstrument):
         commands. Note that only the response of the last command will be
         available in `_write_response`
 
-        TODO (WilliamHPNielsen): add automatic de-concatenation of commands.
         """
         if self.debugmode:
             log.info('Sending command string: {}'.format(cmd))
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/QDev/QDac_channels.py
```diff
@@ -642,9 +639,11 @@ def write(self, cmd):
...
-        Note that this procedure makes it cumbersome to handle the returned
-        messages from concatenated commands, e.g. 'wav 1 1 1 0;fun 2 1 100 1 1'
-        Please don't use concatenated commands
+        In this method we expect to read one termination char per command. As
+        commands are concatenated by `;` we count the number of concatenated
+        commands as count(';') + 1 e.g. 'wav 1 1 1 0;fun 2 1 100 1 1' is two
+        commands. Note that only the response of the last command will be
+        available in `_write_response`
 
         TODO (WilliamHPNielsen): add automatic de-concatenation of commands.
```
WilliamHPNielsen: This PR just does that: it adds automatic de-concatenation (or at least something equivalent). This `TODO` should thus be removed.

jenshnielsen: Done

PR #738, opened 2017-09-15
Rigol ds4034

sohailc:
I have added a driver for the Rigol DS4034 scope as I needed it for a calibration.


WilliamHPNielsen:
Whoa! Slow down, bro. I don't think all these commits should be in the PR.

sohailc:
Did I do something wrong? I intended to add a single driver but earlier I made a few mistakes with the PR. I have reverted everything other then the Rigol DS4034 driver to its original state, so everything is good to go now, right?

WilliamHPNielsen:
It's fine, we'll squash it when we merge it. And yes, the sum of all these commits is indeed just a single new driver, which I'll review now.

qcodes/instrument_drivers/rigol/DS4034.py
```diff
@@ -0,0 +1,106 @@
...
+class Rigol_DS4035(VisaInstrument):
+    def __init__(self, name, address, **kwargs):
+        super().__init__(name, address, terminator='\n', **kwargs)
+
+        self.add_parameter(
+            "data_format",
+            get_cmd=":wav:form?",
+            set_cmd=":wav:form {}",
+            val_mapping={
+                'asc': 'ascii',
```
WilliamHPNielsen: Is there a reason for these two strings to be abbreviated? Personally, I like long, verbose, telling values for string parameter.

sohailc:
Cheers dude...

jenshnielsen:
I think this has a lot of extra commits because you started out from your ami branch. You can avoid that by starting from the master branch

sohailc:
Yes thats probably it. I will remember that the next time

qcodes/instrument_drivers/rigol/DS4034.py
```diff
@@ -0,0 +1,106 @@
...
+            get_cmd=":syst:aut?"
+        )
+
+        self.add_parameter(
+            "auto_scale",
+            set_cmd=":AUToscale",
+            vals=Bool()
+        )
+
+        for channel_number in range(1, 5):  # It is really unfortunate that a get parameter does not except
```
WilliamHPNielsen: This should be channelised. You can confer with the TPS2012 driver (see `TPS2012Channel`).

qcodes/instrument_drivers/rigol/DS4034.py
```diff
@@ -0,0 +1,106 @@
...
+
+            self.add_parameter(
+                "vertical_scale_channel{}".format(channel_number),
+                get_cmd="chan{}:scale?".format(channel_number),
+                set_cmd="chan{}:scale ".format(channel_number) + "{}",
+                get_parser=float
+            )
+
+        self.add_function(
+            "get_wave_form",
```
WilliamHPNielsen: The wave form should preferrably be an ArrayParameter for integration with the DataSet and Loop. The TPS2012 driver (see `ScopeArray`) has a good example of this.

qcodes/instrument_drivers/rigol/DS4034.py
```diff
@@ -0,0 +1,112 @@
...
+        channels = ChannelList(self, "Channels", Rigol_DS4035_Channel, snapshotable=False)
+
+        for channel_number in range(1, 5):
+            channel = Rigol_DS4035_Channel(self, "ch{}".format(channel_number), channel_number)
+            channels.append(channel)
+
+        channels.lock()
+        self.add_submodule('channels', channels)
+
+    def get_wave_form(self, n_samples, channel):
```
WilliamHPNielsen: I still think that the waveform should be an ArrayParameter. Then it fits nicely into a `Loop`.

YakBizzarro:
Ah crap, I did'n realized that there was already an effort to implement a driver for Rigol DS4000.
Here there is my WIP version: https://github.com/YakBizzarro/Qcodes/commit/0e90c5f0e1be5a382373d094e071b159c2cf5670
The main difference is that I use binary transfer instead of ascii and that I support reading of raw data as well. Maybe we can put together the best of the two.

sohailc:
It would be nice to compare the differences. Also, I have implemented only
the functionality that I needed at the moment. If you have implemented
more, lets go with yours.

I could not get the binary transfer to work, so I am curious how you do
that.

Cheers,

Sohail

YakBizzarro:
You can find my version here: https://github.com/YakBizzarro/Qcodes/commit/0e90c5f0e1be5a382373d094e071b159c2cf5670

Basically I implemented only the waveform readout and few commands, I'm trying to merge this code in your

Best,

Andrea

YakBizzarro:
I merged the two drivers version and make use of ArrayParameter like in TPS2012 driver (lot a copy paste from there): https://github.com/YakBizzarro/Qcodes/blob/c665fa0f90ea94947a200a6b5d35844d6c89df7a/qcodes/instrument_drivers/rigol/DS4000.py 

What do you think? @sohailc  @WilliamHPNielsen 

Regards,

Andrea

sohailc:
Thanks Andrea. On first glance.... good job! I will look at it in depth and
give proper feedback Monday

Cheers,

Sohail

WilliamHPNielsen:
@YakBizzarro I don't quite know what `raw` mode entails and there is a wrong indentation at line 72, but apart from that it looks really fine. Very TPS2012-esque.

YakBizzarro:
@WilliamHPNielsen Thank you for the review. The raw mode return the data acquired in the internal buffer, while the normal one returns the data on the display. I'll add a note to specify. At line 72 the indentation seems me correct: the 'if' clause is inside the 'for' loop and the next 'else' clause is relative to the 'for' loop, not to the 'if'

jenshnielsen:
Closed in favour of #760

PR #739, opened 2017-09-15
Ivvi triggers

sohailc:
I have added a parameter on the QUTech IVVI driver that sends a trigger.


WilliamHPNielsen:
Does it make sense that a trigger is a parameter? Is this not more like a `*TRG` (force trigger) call, which we usually put in as a function?

sohailc:
I guess that makes sense. I will fix this

PR #740, opened 2017-09-15
driver/Minicircuits RCDAT-6000-60 variable attenuator

euchas:
- implemented all of the main commands for the programmable attenuator 

@jenshnielsen


jenshnielsen:
@Dominik-Vogel Has this been implemented in another pr?

astafan8:
@Dominik-Vogel Is this PR still relevant? the build passes, shall it be reviewed and merged? also, shall it better be merged into into contrib drivers?

Dominik-Vogel:
This is a variable attenuator. We do have a driver for a model of the RUDAT series, which to my perception is the same but without the ethernet connection (usb only). The RCDAT has both. I think we should close it and if there is a need for a driver for this specific device we should write a new one based on the minicircuits infrastructure we have.

PR #743, opened 2017-09-15
driver/Stanford Research sr830 Lock-in amplifier

sohailc:
Hi everyone, 

I need some help with this driver. I cannot seem to create a channel array for this because the ChannelBuffer class complains that "SR830_Channel_Array" is not an instance of "SR830". Commenting out the lines raising the exception makes the driver break further down stream: 

'SR830_Channel_Array' object and its delegates have no attribute 'buffer_npts'

I am unsure how to proceed. Can anyone help me out?

### DIFF
--- a/qcodes/instrument_drivers/stanford_research/SR830.py
+++ b/qcodes/instrument_drivers/stanford_research/SR830.py
@@ -45,7 +45,15 @@ class SR830_Channel_Array(InstrumentChannel):
 
         self.add_parameter('databuffer',
                             channel=channel_number,
-                            parameter_class=ChannelBuffer)
+                            parameter_class=ChannelBufferWrapped(parent))
+
+
+class ChannelBufferWrapped:
+    def __init__(self, instrument):
+        self._instrument = instrument
+
+    def __call__(self, name, instrument, channel):
+        return ChannelBuffer(name, self._instrument, channel)
 
 
 class ChannelBuffer(ArrayParameter):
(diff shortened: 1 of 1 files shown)


sohailc:
The problem obviously is that "SR830_Channel_Array" is not a subclass of "SR830" and does not have its attributes and methods. But I cannot make the former inherit from the latter as SR830 is dependent on SR830_Channel_Array. This causes a circular dependency.

sohailc:
Perhaps we have stumbled upon a problem we did not anticipate before?

sohailc:
I got this driver to run without throwing exceptions, but the solution is *very* ugly

jenshnielsen:
I think you just need to refactor it slightly. `ChannelBuffer` needs to be a parameter that lives on the  Channel not on the instrument. Thus it takes a Instrumentchannel and not an instrument as a constructor input. Then for the parameters that live on the instrument and not on the channel that you need access to from the parameter you will have to go trough the `_parent` member of the channel.

As in `self._instrument.channels[self.channel].ratio.get()` should be `self._instrument.ratio.get()` and `self._instrument.buffer_npts()` should be `self._instrument._parent.buffer_npts()` and so on.

It is unfortunate that the constructor argument is an instrument where it can really be both a channel and an instrument but that constructer predates the existence of channels as a QCoDeS concept

jenshnielsen:
@sohailc did my explanation make sense to you

WilliamHPNielsen:
@sohailc, would you like to apply these changes? I can also do it; we have an SR830 in our stack.

sohailc:
Sorry for the late reply. Let me apply these changes next week

jenshnielsen:
replaced by #811

sohailc:
In order to get rid of the commit messages that are unrelated to this PR, I deleted my original clone of the QCoDeS repo and recloned a fresh one. For this reason I had to create a new PR for this issue, which can be found in https://github.com/QCoDeS/Qcodes/pull/811

Please permanently reject this PR

PR #744, opened 2017-09-18
fix: removed newline from val_mapping values + early type check

Dominik-Vogel:
### Problem
The current impelementation of val_mapping requires users to append newlines for the mapping.
val_mapping allows us to specify a dictionary that translates both set and return values for commands to python-typed values.

### Example
The following snippet taken from the Keysight MSOS104A driver adds a boolean parameter to the instrument
```
self.add_parameter('timebase_roll_enabled',
                    get_cmd=':TIMebase:ROLL:ENABLE?',
                    set_cmd=':TIMebase:ROLL:ENABLE {}',
                    val_mapping={True: "1", False: "0"}
                  )
```
This enable a call of "driver_instance.timebase_roll_enabled()" wich returns a boolean. For the generated VISA command:
```
:TIMebase:ROLL:ENABLE?\n
```
the device returns e.g.
```
0\n
```
which is not encountered in the dictionary and will fail the call. The problem can be solved by writing:
```
val_mapping={True: "1\n", False: "0\n"}
```
which is, IMO, an ugly workaround.
In other drivers, eg Tektronix TPS2012, this issue is solved alternatively by
```
val_mapping={'ON': 1, 'OFF': 0}
```
Here the dicitionary has integer-valued keys. This only works because in the current parameter implementation, if value, as string, is not found in the dictionary, it will converted to an integer and the dictionary will be searched again.
This again only works because
```
val = int(val)
```
does not seem to care about trailing newlines.

### Solution
My suggested soulution is to become more stringent with the val_mapping and strip trailing newlines from the visa value. To ensure compatibility, the dictionary strings will also be right-stripped.

### Fail early
To circumvent that for every call of a visa integer value mapped getter, the dictonary is requested twice, I suggest an fail early approach:
On setting the val_mapping all keys get converted to strings and the conversion code will be removed from the getter.
This way it is ensured to have the dictionary in a defined state and any type error will be produced on instrument instantiation, rather than during the experiment for a certain return value. 

### Compatibility issues
The fix should be non-invasive, with the only exception, that in the exotic case, a val_mapping is used with a get preparser that does not return a string value. From my perspective as being new to QCoDeS

### Test
I will test the code on the Keysight MSOS104A driver as soon as there is a free time slot on the test setup.

What do you think @jenshnielsen and @WilliamHPNielsen


nulinspiratie:
@Dominik-Vogel The visa_handle of a VisaInstrument allows you to define a read termination: `VisaInstrument.visa_handle.read_termination = '\n'`. This will strip away any newline characters when returning values.

Dominik-Vogel:
@nulinspiratie ok thanks for the tip about the termination. Anyway this does not solve the issue. Can you see any point in having the termination present as a default?.
Stripping the string manually enforces the val_mapping to work no matter what state the termination is in and the early check of the dictionary is also useful in my opinion, because it will raise an exception upon creation of the parameter and not upon use.

nulinspiratie:
@Dominik-Vogel Are you sure it doesn't solve the issue? We've used it in an instrument with `\n` termination, and it seems to solve it. The stripping happens at a low level so when the Instrument receives the value from the visa .ask, it is already stripped.

It seems that the code would place a limit on val_mapping to only accept strings as values, while a val_mapping should also in principle be allowed for other types (useful in cases unrelated to visa instruments). Also, in the new Parameter, val_mapping is simply an attribute, and so not called upon initialization

I do agree that if there is a simple solution, then it would make sense to make stripping implicit. However, we must be careful that it doesn't have unintended consequences. Also, stripping the termination character can also be useful in cases where a val_mapping is not defined

nulinspiratie:
The QCoDeS VisaInstrument also has the method `VisaInstrument.set_terminator`, which wraps the underlying visa command.

Dominik-Vogel:
@nulinspiratie Sorry, I missed an essential "not" in my post...
I agree that it solves the issue in the sense that things will work with it. But my critique is that in almost all cases one has to remember to pass the terminator='\n' as an argument. I am arguing for this to be the default. I am pretty sure that anyone who is fresh to qcodes will fall into the same trap when trying to use val_mapping for the first time.

Dominik-Vogel:
@nulinspiratie I was also thinking about this limitation to strings and ints.
The only situation I could come up with where you need val_mapping for other types than strings is when you use a get-preparser. But doing so in conjunction with val_mapping seems odd to me. Why would I parse the original string, output some other type and subsequently map that type to yet another type via val_mapping....

nulinspiratie:
At the moment I can't think of a parameter that needs both a get_preparser and a val_mapping other than strings, but I have noticed that Parameters can have many completely different use cases, so keeping it as general as possible is a good idea imo.

Since the situation of a character termination is only relevant for VisaInstruments, wouldn't it make sense to perform the rstrip() there? For example in VisaInstrument.ask_raw?

WilliamHPNielsen:
I would like to remark that the current checks are poorly implemented.

Let's say that we have defined a `val_mapping = {1: 'CH1', 2: 'CH2'}` and not specified the termination character. Then we receive `'CH1\n'` from the instrument, which we try to match against the inverted `val_mapping`. From `parameter.py`:

```
if self.val_mapping is not None:
    if value in self.inverse_val_mapping:
        value = self.inverse_val_mapping[value]
    elif int(value) in self.inverse_val_mapping:
        value = self.inverse_val_mapping[int(value)]
    else:
        raise KeyError("'{}' not in val_mapping".format(value))
```
which will raise the strange and confusing error that `'CH1\n'` can not be cast to an `int`.

astafan8:
@Dominik-Vogel @WilliamHPNielsen @nulinspiratie 
This pull request was trying to fix a problem with the StandardParameter which has been [deprecated some time ago](https://github.com/QCoDeS/Qcodes/blob/416fda49380e5e80ea7df4cf1279eff1cf6ba983/qcodes/instrument/parameter.py#L1534). Hence, please, close this PR, if you think it is solved.

WilliamHPNielsen:
@astafan8, are you certain that the same code did not creep into the new `Parameter`? We deprecated a name, not all the functionality.

astafan8:
@WilliamHPNielsen I had a look at Parameter and _BaseParameter on master branch - it does not contain this logic.

Shall we close this PR?

Dominik-Vogel:
This PR addresses two issues:
- early failing of problematic val_mappings
- termination character stripping

The first point has not been addressed in the current version of qcodes and I consider it still valid.
The second point is possibly not relvant as Serwan pointed out. We will need to look into how the read termination is handled.

astafan8:
@Dominik-Vogel is `early failing of problematic val_mappings` still an issue?

jenshnielsen:
This does no longer have any commits. Should we close ?

astafan8:
> This does no longer have any changes

ah, in this case, trivially and obviously - yes :)

PR #758, opened 2017-09-25
ZNB don't snapshot traces

jenshnielsen:
They are slow and may not even function correctly if the trigger is not set and there is no point of snapshotting array parameters

### DIFF
--- a/qcodes/instrument_drivers/rohde_schwarz/ZNB.py
+++ b/qcodes/instrument_drivers/rohde_schwarz/ZNB.py
@@ -345,6 +345,12 @@ class ZNBChannel(InstrumentChannel):
         self.trace.set_sweep(start, stop, npts)
         self.trace_mag_phase.set_sweep(start, stop, npts)
 
+    def snapshot_base(self, update=False, params_to_skip_update=None):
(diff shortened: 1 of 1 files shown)


jenshnielsen:
@WilliamHPNielsen

AdriaanRol:
@jenshnielsen , I feel that this is a problem that is bound to occur more frequently. Rather than overwritting the snapshot method, would it not be better to add this as a property of the parameter that is set through the `add_parameter` method. Something like `include_in_snapshot=True/False` ?

jenshnielsen:
Perhaps, but that API does not exist and  this one does because of special handling in https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/QDev/QDac_channels.py#L107 that cannot be done with a bool as the point is to intersept the snapshot and call an alternative method the bulk updates all the parameters in one instrument call. 

I will prefer not touch the parameters until #600 lands. Honestly I think we will get a long way by excluding Array and Multiparameters per default there is almost never any good reason to snapshot those.

jenshnielsen:
I added #759 to remind us to improve this in general

PR #761, opened 2017-09-25
Driver/awg70000Aseries

WilliamHPNielsen:
Add a driver for the Tektronix 70000A series AWG.

Still very unfinished.

- Add driver for 70000A series
- Add specific 70002A driver
- Add documentation notebook
- Add benchmark for waveform transfer speeds

Done so far:

- Add function to cast numpy arrays into `.wfmx` files and transfer and load them
- Add basic channels functionality
- Figure out proprietary `.seqx` format
- Make a benchmark notebook
- Support a reasonable subset of channel options

Still pending:

- Run the benchmark notebooks at our test setup
- Fully understand the `.seqx` format ( I guess we have that down now)

Due to time constraints, the handling of the `.seqx` files is not well done. Two bugs that I know of:

- The `.seqx` file is <del>invalid if only one channel is specified</del> cannot reproduce
- The `.seqx` file is <del>invalid if more than 20 waveforms are in it (at least if they have 100k points)</del> that was a bug on our side that has been fixed

And one thing that should be optimised is that the `.seqx`<del> file is stored to disk locally upon creation</del>, this has been fixed now.

UPDATE: The `.seqx` file format is reasonably well understood now, thanks to support from Tektronix and more time spent staring at our code. 

@jenshnielsen


jenshnielsen:
@WilliamHPNielsen Should I review this or do you want to wait till there is time to fix the issues?

WilliamHPNielsen:
@jenshnielsen Let's wait until we actually have a unit to play with.

WilliamHPNielsen:
@core this baby is ready for a review.

qcodes/instrument_drivers/tektronix/AWG70000A.py
```diff
@@ -0,0 +1,936 @@
...
+        # one would expect in DIR mode.
+        self.add_parameter(
+            'awg_amplitude',
+            label='Channel {} AWG peak-to-peak amplitude'.format(channel),
+            set_cmd='SOURCe{}:POWer {{}}'.format(channel),
+            get_cmd='SOURce{}:POWer?'.format(channel),
+            unit='V',
+            get_parser=lambda s: 2*10**((float(s)-10)/20),
+            set_parser=lambda V: 10 + 20 * np.log10(V/2),
+            vals=vals.Numbers(0.250, 0.500))
```
jenshnielsen: How about using [SOURce[n]]:VOLTage[:LEVel][:IMMediate][:AMPLitude] as Carl suggested

qcodes/instrument_drivers/tektronix/AWG70000A.py
```diff
@@ -0,0 +1,936 @@
...
+
+        return xmlstr
+
+    @staticmethod
+    def _makeSMLFile(trig_waits: List[int],
+                     nreps: List[int],
+                     event_jumps: List[int],
+                     event_jump_to: List[int],
+                     go_to: List[int],
+                     wfm_names: List[List[str]],
```
jenshnielsen: Are these modified in place otherwise I would in general suggest using Sequence see http://mypy.readthedocs.io/en/stable/common_issues.html#invariance-vs-covariance

qcodes/instrument_drivers/tektronix/AWG70000A.py
```diff
@@ -0,0 +1,936 @@
...
+
+        return binary_out
+
+    @staticmethod
+    def makeSEQXFile(trig_waits: List[int],
+                     nreps: List[int],
+                     event_jumps: List[int],
+                     event_jump_to: List[int],
+                     go_to: List[int],
+                     wfms: List[List[bytes]],
```
jenshnielsen: Sequence

qcodes/instrument_drivers/tektronix/AWG70000A.py
WilliamHPNielsen: I'm not sure I understand the docs... The above code will fail if I pass something which is a `list` of a subclass of `int` because mypy considers `List` to be invariant? But immutable containers like `Sequence` are covariant. Is this why `Sequence` is preferred?

If so, I would say that it doesn't seem like such a big issue here, since we rarely subclass `int`, but I could be overlooking something crucial.

qcodes/instrument_drivers/tektronix/AWG70000A.py
WilliamHPNielsen: Yes, that seems more straightforward and less errorprone. I couldn't find that command in the manual, but I'll test is this afternoon.

qcodes/instrument_drivers/tektronix/AWG70000A.py
jenshnielsen: So in general Sequence is preferable just to tell the user that it's imitable. The 
above would in principle be wrong if you pass a tuple as go_to

PR #762, opened 2017-09-26
feat: add MatPlot guide

### LINKED ISSUE #757: Subplots with MatPlot
MatPlot seems to have a subplot feature, but it's not obvious to use. Perhaps there could be a short tutorial on how to plot multiple heatmaps with certain figure sizes in a single figure.

nulinspiratie:
Fixes #757 

- Adds comprehensive guide on plotting with MatPlot

This is basically all the tools I use in MatPlot, if anyone has additions please add it.
Also, it doesn't have anything yet on QTPlot, so could someone who uses that please add documentation?

@gzheng29 @WilliamHPNielsen


nulinspiratie:
Accidentally PRed to wrong fork

jenshnielsen:
Looks great: 

Could you add the title of the notebook as a Markdown title to the first cell of the notebook. This ensures that the notebook will be included under that name in the webpage. Currently I think it will be indexed under Matplot as that is the first title in the notebook.

It would be great to include a small section about how to use Matplotlib buildin methods as this is often a question we get. I would pub something like the following in before the the event section:


```
Interfacing with Matplotlib
As Matplot is build directly on top of Matplotlib you can use standard Matplotlib functions which are readily available online in matplotlib documentation as well as StackOverflow and similar sites. Here we first create a figure as above to manipulate
```
```python
loop = qc.Loop(
    p_sweep.sweep(0, 20, step=1), delay=0.5).loop(
        p_sweep2.sweep(0, 10, step=1), delay=0.1).each(
            p_measure)
data = loop.get_data_set('test_plotting_2D_2')
​
# Create plot for measured data
plot = qc.MatPlot([*data.measured_val], data.measured_val)
# Attach updating of plot to loop
loop.with_bg_task(plot.update)
​
loop.run()
```
```
To use the Matplotlib api you need to acess the matplotlib axis and figure object. The axes object can be accessed for each subplot via the subplots list i.e. we can change the title of the first subplot with:
```
```python
ax = plot.subplots[0]
ax.set_title("My figure title")
```
```
See http://matplotlib.org/api/axes_api.html for documentation of the Matplotlib Axes class
The Matplotlib figure object can be accessed via the fig attribute on the QCoDeS Matplot object i.e.
```
```python
fig = plot.fig
fig.tight_layout()
```
```
See http://matplotlib.org/api/figure_api.html for documentation of the Matplotlib Figure class
```

WilliamHPNielsen:
This is terrific! Let's get Jens' suggestions in and merge this baby.

nulinspiratie:
@jenshnielsen Good idea, I added your suggestions, and a bit on using Pyplot. I think it's good to go

PR #763, opened 2017-09-26
feat: add threading to qcodes Loops

nulinspiratie:
This PR adds the possiblity of running a loop in a separate thread. I understand that the final goal is to move towards multiprocessing, but I haven't heard much about that recently, and this could be an intermediate solution.

- Adds possibility to run loop in separate thread.
    Activated by passing `Loop.run(thread=True)`. 
- Add `qc.config.core.loop_thread`, which is False by default. If set to True, loops will automatically start in new thread unless `thread=False` is explicitly specified.
- Add `qc.stop()`, which will raise a `_QCoDeSBreak` in the loop, stopping it after the loop is done with its current iteration. Done by attaching `ActiveLoop._raise_if_stopped` as background task. For multidimensional loops, this method is injected in the inner loops as well.
- Add KillableThread, which is instantiated when a new thread is created for a Loop. This class has `KillableThread.terminate`, which will attempt to force stop the thread. Not to be used in general, but useful as last-case emergency.
- Add UpdaterThread, which can periodically perform actions. Useful in combination with having a loop in a separate thread, for live plotting etc.


jenshnielsen:
Sorry this feel behind. I was looking at it but could not immediately figure out how to use the UpdaterThread in the intended way. Can you provide an example of that?

WilliamHPNielsen:
Also, this PR is not python 3.5 compatible. We should keep that in mind as long as we officially support python 3.5.

nulinspiratie:
@jenshnielsen I renamed UpdaterThread to PeriodicThread, as it more suits its functionality.
It basically repeatedly executes a callable/callables at a specified interval. One example where this can be useful is updating a plot

I also added some documentation for both the PeriodicThread and the KillableThread.

@WilliamHPNielsen I changed the f-string so it should be 3.5 compatible now as well

jenshnielsen:
I suggest closing this. We are not continuing development of the old loop

WilliamHPNielsen:
Agreed.

PR #764, opened 2017-09-26
Feature/mat plot updater

nulinspiratie:
Adds an updater for when a Loop is on a separate thread


nulinspiratie:
Accidentally PRed to wrong branch

PR #772, opened 2017-09-29
driver/R&S HMC804x power supply

YakBizzarro:
Add driver for R&S HMC804x power supply.
Basic commands are supported.

@WilliamHPNielsen, @jenshnielsen


qcodes/instrument_drivers/rohde_schwarz/HMC804x.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+    This is the qcodes driver for the Rohde & Schwarz HMC804x Power Supply
+    """
+
+    _max_currents = {3: 3.0, 2: 5.0, 1: 10.0}
+
+    def __init__(self, name, address, num_channels, **kwargs):
+        super().__init__(name, address, **kwargs)
+
+        self.max_current = RohdeSchwarzHMC804x._max_currents[num_channels]
```
jenshnielsen: should this be self rather than RohdeSchwarzHMC804x?

qcodes/instrument_drivers/rohde_schwarz/HMC804x.py
```diff
@@ -0,0 +1,115 @@
...
+        self.add_parameter('state',
+                           label='Output enabled',
+                           set_cmd='OUTPut:MASTer:STATe {}',
+                           get_cmd='OUTPut:MASTer:STATe?',
+                           val_mapping={'ON': 1, 'OFF': 0},
+                           vals=vals.Enum('ON', 'OFF')
+                           )
+
+        # channel-specific parameters
+        channels = ChannelList(self, "SupplyChannel", RohdeSchwarzHMC804xChannel)
```
WilliamHPNielsen: The ChannelList should not be snapshotable, should it? The channels are also added individually below.

WilliamHPNielsen:
@YakBizzarro, it would be great if you could add an example notebook for this driver (it should be called `qcodes/docs/examples/driver_examples/Qcodes example with Rohde Schwarz HMC804x`). Even for a simple instrument like this, the example notebook serves as a kind of rudimentary test. It is very valuable for us to have.

qcodes/instrument_drivers/rohde_schwarz/HMC804x.py
YakBizzarro: Is a static member of the class, not a variable of the object, so the reference to RohdeSchwarzHMC804x is correct

qcodes/instrument_drivers/rohde_schwarz/HMC804x.py
YakBizzarro: Probably not. How can mark it a non-snapshotable?

YakBizzarro:
> To keep consistency with most of the driver codebase (which is not exactly famous for its internal consistency smiley), it would perhaps be good to split out the specific drivers into separate files. It will make imports more uniform.

Uhm, I not very convinced. The instruments are so similar (the differs only by one line of code) that I don't know if it's warranted

> Does it make sense to have a ChannelList for this instrument? Do you ever iterate over the channels? For the HCM8041 it seems a bit silly to have a list.

I found channels as the easiest way to deal with that. I agree that for the HCM8041 is kind of an overkill, but I meant it for the biggest brother HCM80413

qcodes/instrument_drivers/rohde_schwarz/HMC804x.py
WilliamHPNielsen: Give it the `snapshotable` keyword argument upon init.

From the `ChannelList` docstring:
```
        snapshotable (bool): Optionally disables taking of snapshots
            for a given channel list.  This is used when objects
            stored inside a channel list are accessible in multiple
            ways and should not be repeated in an instrument snapshot.
```

WilliamHPNielsen:
@YakBizzarro 
> Uhm, I not very convinced. The instruments are so similar (the differs only by one line of code) that I don't know if it's warranted

I don't know if this will convince you, but it is convenient to know by a simple look at the code base which instruments we have drivers for. Let's say I have a GenericAWG by SomeCompany model 54. If I see a `genericawg_54.py` file in the `somecompany` folder, I know that we are good. If I have to open `genericawg.py` to read that models 50, 51, 52 (but not 53, since that's a different firmware layout), and 54 are actually covered by that file, it's more likely that I will not find what I am looking for. I personally have to look up this stuff all time when people build new setups. The `genericawg_54.py` file can basically just be a single import line or a three-line class definition that just calls `super().__init__`.

If you are still not convinced, then you have my blessing to do as you see fit.

YakBizzarro:
Ok, no problem. I refactored the driver as you asked, the base driver is in the private file _HMC804x.py
I made the channel list non-snapshotable

YakBizzarro:
Uhm, I can't really understand why the build fails, any suggestion?

jenshnielsen:
I am not sure and I cant reproduce it locally but I suspect that the underscrore in the file name might trip up Sphinx. Could you try to remove that?

YakBizzarro:
@jenshnielsen done, it seems good

YakBizzarro:
yes, your're right, done

jenshnielsen:
Great, I think this is ready to land @WilliamHPNielsen do you agree?

PR #774, opened 2017-10-03
fix:save scaled value, perform initial get

nulinspiratie:
Minor but important Parameter fixes I ran into after merging the latest PR.
These are related to ramping and scaling of values, and also incorporating raw_value.

- Perform Parameter.get if get_latest returns None and a step is used. Without this, it will jump to the target value during the first .set if not .get is called first
- scale parameter value during ramping
- Remove raw_value from meta_attrs, add it to self._latest. 
    This is done for the following reasons:
    - The self.get function (previously in ManualParameter) is no longer equal to self.get_latest. Instead, it will directly get the value from self._latest ('value' if parameter doesn't have scale, 'raw_value' otherwise). This is because get_latest always returns the scaled value, while if the parameter has a scale, we want the get function to return the raw value
    - the raw value will now also be popped if snapshot_value=False

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -188,7 +188,7 @@ class _BaseParameter(Metadatable, DeferredOperations):
 
         # subclasses should extend this list with extra attributes they
         # want automatically included in the snapshot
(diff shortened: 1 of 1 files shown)


WilliamHPNielsen:
This has been superseded by #788.

PR #776, opened 2017-10-03
Fix parameter warnings

jenshnielsen:
Fixes a number of warnings following the merge of #600


jenshnielsen:
@nulinspiratie This now reverts the deprecation of full_name. The current dataset makes use of full_name in ways that make it less safe to change so I suggest that we wait with the deprecation until such times as we deprecate the dataset too

nulinspiratie:
Sounds good!

PR #782, opened 2017-10-04
unpin h5py dependency

Dominik-Vogel:
The h5py dependency was pinned to a fixed version. Thus numpy won't update as it depends on it.

found by
@jenshnielsen


jenshnielsen:
We don't pin this to 2.6 in setup.py but only require >=2.6 so I think the conda env should match that.

Issue #620, opened 2017-05-18
Data truncated when saved to disk


MerlinSmiles:
### Steps to reproduce
1. Measure a parameter with a high resolution (lots of decimals)

### Expected behaviour
Data should be measured and saved with full precision

### Actual behaviour
Data is truncated to 6 decimal places (at least in gnuplot formatter)
You get angry when a time measurement does a staircase thing.

### System
**operating system**
Win 10
**qcodes branch**
Master

@core

jenshnielsen:
Yes, the default gnuplot formatter is "%g" which has 6 digits. I am not sure thats the best. 

You should be able to work around this by creating a GnuPlotFormatter and passing 
a different number_format to it such as `.10g` to get 10 digits after the dot. I am not sure what the default should be be but I suspect we should change it.

See https://github.com/QCoDeS/Qcodes/blob/master/qcodes/data/gnuplot_format.py#L30


```python
In [11]: "%g" %  (numpy.pi)
Out[11]: '3.14159'

In [17]: "%.10g" %  (numpy.pi)
Out[17]: '3.141592654'
```

giulioungaretti:
@MerlinSmiles sorry to hear that. This is indeed embarrassingly bad. 
It happens here: https://github.com/QCoDeS/Qcodes/blob/master/qcodes/data/gnuplot_format.py#L70 .

And I guess the fact that by default it rounds six digits was either not known by the original writer or casually not-documented.

MerlinSmiles:
Yes, I saw that it was happening there, but kind of get unhappy with such a feature :(

jenshnielsen:
I don't think we disagree about that @MerlinSmiles

MerlinSmiles:
I would never doubt on that, just reporting to get attention to it :)

WilliamHPNielsen:
Indeed, #783 fixed this issue.

PR #783, opened 2017-10-04
Increase default gnuplot precision from 6 to 15 digits

jenshnielsen:
Workaround for #620 Will be fixed when we use new dataset


nulinspiratie:
What about increasing it to 15 digits? the `g` formatter won't include any additional zeros anyway, so `'{:.15g}'.format(0.123)` will become `0.123` instead of `0.123000000000000`

jenshnielsen:
Yes that seems reasonable 10 was completely arbitrary

PR #785, opened 2017-10-05
driver: Add GUD for Keysight/Agilent/HP wfm gens

WilliamHPNielsen:
Since these instruments basically have not changed through three vendors (and at least one decade), they might as well have a common driver. This will also allow for seamless exchange of one unit with another.

- Add GUD

I have not yet put in deprecation warnings nor created shim drivers pointing to this one. I'll do so if everyone likes the idea.

@jenshnielsen

### DIFF
--- a/qcodes/instrument_drivers/Keysight/KeysightAgilent_33XXX.py
+++ b/qcodes/instrument_drivers/Keysight/KeysightAgilent_33XXX.py
@@ -50,6 +50,8 @@ class OutputChannel(InstrumentChannel):
 
             return output
 
+        self.model = self._parent.model
+
         self.add_parameter('function_type',
                            label='Channel {} function type'.format(channum),
                            set_cmd='SOURce{}:FUNCtion {{}}'.format(channum),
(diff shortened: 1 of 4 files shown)


qcodes/instrument_drivers/Keysight/KeysightAgilent_33XXX.py
```diff
@@ -0,0 +1,349 @@
...
+                           vals=vals.Enum('CW', 'LIST', 'SWEEP', 'FIXED')
+                           )
+
+        self.add_parameter('frequency',
+                           label='Channel {} frequency'.format(channum),
+                           set_cmd='SOURce{}:FREQuency {{}}'.format(channum),
+                           get_cmd='SOURce{}:FREQuency?'.format(channum),
+                           get_parser=float,
+                           unit='Hz',
+                           # TODO: max. freq. actually really tricky
```
jenshnielsen: Does this need to depend on the model? I don't know why it's really tricky (does it perhaps depend on output type?)

WilliamHPNielsen: It depends on model (I just pushed that change), but also other things which I don't remember. Amplitude, prolly.

jenshnielsen: never mind

WilliamHPNielsen:
@Dominik-Vogel, maybe you'd like to review this?

Dominik-Vogel:
It looks good to me, great that to made this unification. At some point in the future we are going to be really happy about it!

PR #790, opened 2017-10-12
Restore deprecated parameter setters and getters

jenshnielsen:
On top of #788 

Basically add set and get back for delay and step as deprecated functions to unbreak anything that depends on it along with allowing step to be zero as documented in the tutorial.

Also updates the tutorial to use non deprecated properties


jenshnielsen:
@damazter I think the merge of #600 broke some drivers for you. Can you see if this would fix it for you?

damazter:
I put in on my todo list

WilliamHPNielsen:
This does not solve the issue with the `IVVI` driver.

The issue is the following: a parameter is added using the `add_parameter` method with the kwarg `delay`. An error is raised, `
TypeError: __init__() got an unexpected keyword argument 'delay'`

jenshnielsen:
Ok can you merge #788 and I will try to fix that here too

WilliamHPNielsen:
Done.

jenshnielsen:
Ok I think that should have been taken care of now too

WilliamHPNielsen:
As far as I can see, it solves the issue. 👍

PR #793, opened 2017-10-16
Fix keithley no channel

jenshnielsen:
Move definition of self.channel to before its used first time


jenshnielsen:
@WilliamHPNielsen also fixes the second issue

PR #794, opened 2017-10-17
fix double validation when setting parameters

### LINKED ISSUE #754: ManualParameter calls validate twice
The qcodes `ManualParameter` calls validate twice for a single set.
``` python
import qcodes
from qcodes.utils.validators import Validator

class MyValidator(Validator):

    def __init__(self, min_value=-float("inf"), max_value=float("inf")):
        pass

    def validate(self, value, context=''):
        print('validate!')

    is_numeric=True

p = qcodes.ManualParameter('p', initial_value=0, vals=MyValidator())
p.set(10)
```
The output is:
```
validate!
validate!
```
The validator should only be called once.

WilliamHPNielsen:
Fixes #754 .

- Don't double validate when setting

@jenshnielsen


WilliamHPNielsen:
I think I got the double validation right this time.

The following example:
```
import qcodes as qc
from qcodes.utils.validators import Validator

class MyValidator(Validator):

    def __init__(self, min_value=-float("inf"), max_value=float("inf")):
        pass

    def validate(self, value, context=''):
        print('validating value {}'.format(value))

    is_numeric=True

p = qc.Parameter('p', initial_value=0, set_cmd=None, vals=MyValidator())
```
upon doing
```
p.set(10)
p2.step = 1
p2.set(0)
```
Yields
```
validating value 0
validating value 9
validating value 8
validating value 7
validating value 6
validating value 5
validating value 4
validating value 3
validating value 2
validating value 1
```
that is to say, each value gets validated exactly once.
```

jenshnielsen:
@WilliamHPNielsen  Nice could you make that into a test?

WilliamHPNielsen:
Good idea.

PR #797, opened 2017-10-18
Don't wrap _BaseParameter get and set

jenshnielsen:
This breaks multi and array parameters because in unconditionally calls _save_val

Among other issues this resulted in a ~1Gb snapshot because all array parameters were suddenly stored in the snapshot

There are other solutions but I think it's fair not to wrap the getters and setters unconditionally


nulinspiratie:
Isn't this why we have snapshot_value?

jenshnielsen:
I have implemented that as an alternative in #800

WilliamHPNielsen:
We'll close this in favour of #800.

PR #800, opened 2017-10-19
Don't snapshot array and multiparameters per default

jenshnielsen:
This is an alternative to #797 works by disabling snapshot storage for Array and Multiparamters by default

Pro this 

* we retain the wrappers including delays etc for all parameters:

Against

* This is an still an api break. (Array/Multi)Parameters that used to do explicit _saveval will no longer  be snapshotted unless they explicitly pass snapshot_value=True to the super class constructor, I have verified that this is the case for all current QCoDeS drivers.
* If you don't want the wrappers in any _BaseParameter subclass you will have to replace them in your subclass with a noop

@WilliamHPNielsen  @nulinspiratie

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -884,15 +884,28 @@ class ArrayParameter(_BaseParameter):
         snapshot_get (bool): Prevent any update to the parameter, for example
             if it takes too long to update. Default True.
 
+        snapshot_value: Should the value of the parameter be stored in the
+            snapshot. Unlike Parameter this defaults to False as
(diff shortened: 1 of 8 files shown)


WilliamHPNielsen:
Re the two against points:

  * It seems unlikely that anyone is actually using `ArrayParameter` and `MultiParameter` out in the wild. If an API break _seems_ unlikely to be bad, then... (I know, right)
  * This is perhaps a bit unfortunate, but it is currently unclear whether that is a likely scenario.

So... let's just go with this one.

WilliamHPNielsen:
We close this in favour of #800.

Issue #778, opened 2017-10-04
driver/Keysight N5245A/N5230C Spectrum analyzer


Dominik-Vogel:
John Hornibrook, Sydney is working on those drivers.

PR #807, opened 2017-10-21
driver/Keysight N5245A

johnhornibrook:
Driver for Keysight network analyzer N5245A

#778 

@Dominik-Vogel 
Also if @StevenWaddy @jana-d could put a quick eye on it that would be great


WilliamHPNielsen:
@Dominik-Vogel @StevenWaddy Any verdict on this guy?

johnhornibrook:
@spauka  is adding some changes for compatibility with the new dataset. We'll ping when ready.

jenshnielsen:
@johnhornibrook @spauka Whats the status on this one? Would be nice to get it merged

qcodes/instrument_drivers/Keysight/N52xx.py
```diff
@@ -0,0 +1,422 @@
...
+                 setpoints=((0,),),
+                 **kwargs
+                 )
+
+    @property # type: ignore
+    def shape(self) -> Sequence[int]: # type: ignore
+        if self._instrument is None:
+            return (0,)
+        return (self._instrument.root_instrument.points(),)
+    @shape.setter
```
jenshnielsen: Do we need these do nothing setters? If we just leave them out setting the value will be an error. Like this it will just do nothing. Is that what we want?

spauka: If they are not included, then the `super().__init__` call fails. It seemed simpler to create a blank setter than to reimplement the rest of the `__init__` call? Although perhaps it's less clear, and it would be worth reimplementing it anyway...

qcodes/instrument_drivers/Keysight/N52xx.py
```diff
@@ -0,0 +1,422 @@
...
+        self.port = int(port)
+        if self.port < 1 or self.port > 4:
+            raise ValueError("Port must be between 1 and 4.")
+
+        pow_cmd = f"SOUR:POW{self.port}"
+        self.add_parameter("source_power",
+                           label="power",
+                           unit="dBm",
+                           get_cmd=pow_cmd + "?",
+                           set_cmd=pow_cmd + "{}",
```
jenshnielsen: Should there be a whitespace between the  pow_cmd and the value when setting like in the commands below. 
I would personally find it clearer to write this as: `get_cmd=f'{pow_cmd}?'` and `set_cmd=f'{pow_cmd} {{}}'`

spauka:
Agreed with all the above 😄 I've made the formatting changes, I'll put together a sample notebook tomorrow.

qcodes/instrument_drivers/Keysight/N52xx.py
spauka: Done

qcodes/instrument_drivers/Keysight/N52xx.py
```diff
@@ -0,0 +1,422 @@
...
+        """
+        paramspec = paramspec.strip('"')
+        ch, param, trnum = re.findall(r"CH(\d+)_(S\d+)_(\d+)", paramspec)[0]
+        return ch, param, trnum
+
+    def _Sparam(self, paramspec: str) -> str:
+        """
+        Extrace S_parameter from returned PNA format
+        """
+        return self.parse_paramstring(paramspec)[1]
```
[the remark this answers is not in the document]
spauka: Done

qcodes/instrument_drivers/Keysight/N52xx.py
```diff
@@ -0,0 +1,422 @@
...
+                           label='Imaginary',
+                           unit='LinMag',
+                           parameter_class=FormattedSweep)
+
+    def write(self, cmd: str) -> None:
+        """
+        Select correct trace before querying
+        """
+        super().write("CALC:PAR:MNUM {}".format(self.trace))
+        super().write(cmd)
```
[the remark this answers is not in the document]
spauka: Done

qcodes/instrument_drivers/Keysight/N52xx.py
jenshnielsen: Lets not spent to much time on that. But can you include a comment to mention that

spauka:
@jenshnielsen There is now an example notebook available:

https://github.com/johnhornibrook/Qcodes/blob/94176004d26cc6c8cd0e98c97853a961ad5a1681/docs/examples/driver_examples/Qcodes_example_with_Keysight_Network_Analyzer.ipynb

Hopefully that resolves all outstanding issues?

qcodes/instrument_drivers/Keysight/N52xx.py
```diff
@@ -0,0 +1,436 @@
...
+        self._auto_sweep = True
+        self.add_parameter('auto_sweep',
+                           label='Auto Sweep',
+                           set_cmd=self._set_auto_sweep,
+                           get_cmd=lambda: self._auto_sweep,
+                           vals=Bool())
+
+        # Functions
+        # Clear averages
+        self.add_function('reset_averages',
```
astafan8: Could you please convert all the `add_functions` to bound methods of the class which body contains `self.write('SENS:AVER:CLE')`? We are working on removing functions in favor of bound methods because apart from other things it brings clarity. @WilliamHPNielsen

qcodes/instrument_drivers/Keysight/N52xx.py
```diff
@@ -0,0 +1,436 @@
...
+        # Note: These will be accessed through the traces property which updates
+        # the channellist to include only active trace numbers
+        self._traces = ChannelList(self, "PNATraces", PNATrace)
+        self.add_submodule("traces", self._traces)
+        # Add shortcuts to trace 1
+        trace1 = PNATrace(self, "tr1", 1)
+        for param in trace1.parameters.values():
+            self.parameters[param.name] = param
+        # By default we should also pull any following values from this trace
+        self.write("CALC:PAR:MNUM 1")
```
astafan8: similar command is used 2 times in the `PNATrace` class.
- could you extract it as a separate method of `PNATrace` class?
- could you re-use the extracted method here since you've created the `trace1` object?

qcodes/instrument_drivers/Keysight/N52xx.py
```diff
@@ -0,0 +1,436 @@
...
+
+        # Set auto_sweep parameter
+        # If we want to return multiple traces per setpoint without sweeping
+        # multiple times, we should set this to false
+        self._auto_sweep = True
```
astafan8: why do you wrap `_auto_sweep` into a parameter, or why do you have a private attribute AND a parameter? So far as I see, the following will work just fine without the need for `_auto_sweep` and `_set_auto_sweep`:
```python
self.add_parameter('auto_sweep',
    label='Auto Sweep',
    set_cmd=None,
    get_cmd=None,
    vals=Bool(),
    initial_value=True)
```

qcodes/instrument_drivers/Keysight/N52xx.py
```diff
@@ -0,0 +1,436 @@
...
+        # Averages ON
+        self.add_function('averages_on',
+                          call_cmd='SENS:AVER ON')
+        # Averages OFF
+        self.add_function('averages_off',
+                          call_cmd='SENS:AVER OFF')
+
+        # A default output format on initialisation
+        self.write('FORM REAL,32')
+        self.write('FORM:BORD NORM')
```
astafan8: are these two commands not supposed to be called elsewhere/elsewhen? because if they might, please, make each of them a method (or a parameter, perhaps)?

qcodes/instrument_drivers/Keysight/N52xx.py
spauka: Yep, done :)

qcodes/instrument_drivers/Keysight/N52xx.py
spauka: To address your concern, I wrapped this into a parameter, `active_trace`, and am writing this instead.

qcodes/instrument_drivers/Keysight/N52xx.py
spauka: Sure :)

I guess I haven't really seen any docs suggesting the best way to handle local parameters? If this is the preferred way, I'm happy to go with it 👍

qcodes/instrument_drivers/Keysight/N52xx.py
spauka: This just sets the transfer format, I can't see any reason why they need to be called again or changed once set. There is no way to change their values in the UI, so it only needs to happen once when connecting...

qcodes/instrument_drivers/Keysight/N52xx.py
astafan8: great!

well, `Parameter`s are always cached: you can call `get_latest` to get the latest value of the parameter without accessing the instrument.

qcodes/instrument_drivers/Keysight/N52xx.py
astafan8: ok

PR #805, opened 2017-10-20
Added a failing test for PyQT plotting (passes <0.1.7, fails at later versions). 

AdriaanRol:
Includes a self contained test for #769. 
Includes all the changes from #770. 

@jenshnielsen I did not know how to make this PR clean into your branch as there are many forks. 
Very important note, I have tested that this test passes on the 0.1.7 version of QCoDeS but not on the more recent update. 

```
[Adriaan:...arloLab_Repositories/Qcodes]$ py.test qcodes/tests/test_plots.py -v      (4ee1167)
===================================== test session starts ======================================
platform darwin -- Python 3.6.0, pytest-3.0.5, py-1.4.32, pluggy-0.4.0 -- /Users/Adriaan/anaconda/bin/python
cachedir: .cache
rootdir: /Users/Adriaan/GitHubRepos/DiCarloLab_Repositories/Qcodes, inifile:
plugins: cov-2.4.0
collected 3 items

qcodes/tests/test_plots.py::TestQtPlot::test_creation PASSED
qcodes/tests/test_plots.py::TestQtPlot::test_simple_plot FAILED
qcodes/tests/test_plots.py::TestMatPlot::test_creation PASSED

=========================================== FAILURES ===========================================
_________________________________ TestQtPlot.test_simple_plot __________________________________

self = <qcodes.tests.test_plots.TestQtPlot testMethod=test_simple_plot>

    def test_simple_plot(self):
        plotQ = QtPlot(remote=False, show_window=False, interval=0)
        plotQ.add_subplot()

        main_QtPlot = QtPlot(
            window_title='Main plotmon of TestQtPlot',
            figsize=(600, 400))

        x = np.arange(0, 10e-6, 1e-9)
        f = 2e6
        y = np.cos(2*np.pi*f*x)

        for j in range(4):
            main_QtPlot.add(x=x, y=y,
                            xlabel='Time', xunit='s',
                            ylabel='Amplitude', yunit='V',
                            subplot=j+1,
>                           symbol='o', symbolSize=5)

qcodes/tests/test_plots.py:65:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
qcodes/plots/base.py:91: in add
    self.add_to_plot(**kwargs)
qcodes/plots/pyqtgraph.py:149: in add_to_plot
    self.fixUnitScaling()
[...]

PR #808, opened 2017-10-23
Travis run plot tests

jenshnielsen:
As suggested by @AdriaanRol  Currently on top of #805


AdriaanRol:
@jenshnielsen you would also have to remove the @skipif statements from the actual plotting tests.

jenshnielsen:
No that is not needed. I have changed the script to not set the condition that triggers the skip

PR #810, opened 2017-10-23
Small fixes to docs build

jenshnielsen:
* Make clean now removes all autogenerated api docs
* Don't annotate instruments in parameter. They generate circular imports and
   the work around break sphinx due to https://github.com/sphinx-doc/sphinx/issues/4072
* Fix a few rst issues to render better


jenshnielsen:
will do  #810 for now

PR #811, opened 2017-10-23
initial commit, untested

sohailc:
This PR aims to solve the same problem as "driver/Stanford Research sr830 Lock-in amplifier #743" We are opening a new PR because I deleted my original clone of the QCoDeS repo to get rid of unrelated commit messages, 

NB: Please permanently reject PR 743.

### DIFF
--- a/qcodes/instrument_drivers/stanford_research/SR830.py
+++ b/qcodes/instrument_drivers/stanford_research/SR830.py
@@ -4,6 +4,59 @@ import numpy as np
 from qcodes import VisaInstrument
 from qcodes.instrument.parameter import ArrayParameter
 from qcodes.utils.validators import Numbers, Ints, Enum, Strings
+from qcodes.instrument.channel import InstrumentChannel, ChannelList
+
+
+class SR830AuxArray(InstrumentChannel):  # We see that InstrumentChannel, ChannelList are something of a misnomer
+    def __init__(self, parent, name, aux_number):
+        super().__init__(parent, name)
+
+        self.add_parameter('aux_in',
+                           label='Aux input {}'.format(aux_number),
+                           get_cmd='OAUX? {}'.format(aux_number),
+                           get_parser=float,
+                           unit='V')
+
+        self.add_parameter('aux_out',
+                           label='Aux output {}'.format(aux_number),
+                           get_cmd='AUXV? {}'.format(aux_number),
+                           get_parser=float,
+                           set_cmd='AUXV {0}, {{}}'.format(aux_number),
+                           unit='V')
+
+
+class SR830ChannelArray(InstrumentChannel):
+    def __init__(self, parent, name, channel_number):
+        super().__init__(parent, name)
+
+        self._parent = parent
+
+        # detailed validation and mapping performed in set/get functions. Note that we only have
+        # two channels
+        self.add_parameter(
+            'ratio',
+            label='Channel {} ratio'.format(channel_number),
+            get_cmd=partial(parent._get_ch_ratio, channel_number),
+            set_cmd=partial(parent._set_ch_ratio, channel_number),
+            vals=Strings()
+        )
+
+        self.add_parameter(
+            'display',
+            label='Channel {} display'.format(channel_number),
+            get_cmd=partial(parent._get_ch_display, channel_number),
+            set_cmd=partial(parent._set_ch_display, channel_number),
+            vals=Strings()
+        )
+
+        self.add_parameter(
+            'databuffer',
+            channel=channel_number,
+            parameter_class=ChannelBuffer
+        )
+
+        def get_parent():
+            return self._parent
 
 
 class ChannelBuffer(ArrayParameter):
@@ -15,11 +68,11 @@ class ChannelBuffer(ArrayParameter):
     The instrument natively supports this in its TRCL call.
     """
 
-    def __init__(self, name: str, instrument: 'SR830', channel: int):
+    def __init__(self, name: str, channel_array: 'SR830ChannelArray', channel: int):
         """
         Args:
             name (str): The name of the parameter
-            instrument (SR830): The parent instrument
+            channel_array (SR830ChannelArray): The channel array of the parent instrument
             channel (int): The relevant channel (1 or 2). The name should
                 should match this.
         """
@@ -29,9 +82,9 @@ class ChannelBuffer(ArrayParameter):
             raise ValueError('Invalid channel specifier. SR830 only has '
                              'channels 1 and 2.')
 
-        if not isinstance(instrument, SR830):
+        if not isinstance(channel_array, SR830ChannelArray):
             raise ValueError('Invalid parent instrument. ChannelBuffer '
-                             'can only live on an SR830.')
+                             'can only live on a channel buffer of SR830.')
 
         super().__init__(name,
                          shape=(1,),  # dummy initial shape
@@ -43,7 +96,7 @@ class ChannelBuffer(ArrayParameter):
                                    'data buffer of one channel.')
 
         self.channel = channel
-        self._instrument = instrument
+        self._instrument = channel_array.get_parent()
 
     def prepare_buffer_readout(self):
         """
@@ -323,48 +376,22 @@ class SR830(VisaInstrument):
                            get_cmd='OEXP? 3',
                            get_parser=parse_offset_get)
 
+        aux = ChannelList(self, "Aux", SR830AuxArray, snapshotable=False)
+        channels = ChannelList(self, "Channel", SR830ChannelArray, snapshotable=False)
+
         # Aux input/output
-        for i in [1, 2, 3, 4]:
-            self.add_parameter('aux_in{}'.format(i),
-                               label='Aux input {}'.format(i),
-                               get_cmd='OAUX? {}'.format(i),
-                               get_parser=float,
-                               unit='V')
+        for count in [1, 2, 3, 4]:
+            iaux = SR830AuxArray(self, "aux{}".format(count), count)
+            aux.append(iaux)
 
-            self.add_parameter('aux_out{}'.format(i),
-                               label='Aux output {}'.format(i),
-                               get_cmd='AUXV? {}'.format(i),
-                               get_parser=float,
-                               set_cmd='AUXV {0}, {{}}'.format(i),
-                               unit='V')
+            if count < 3:
+                channel = SR830ChannelArray(self, "channel{}".format(count), count)
+                channels.append(channel)
 
-        # Setup
-        self.add_parameter('output_interface',
-                           label='Output interface',
-                           get_cmd='OUTX?',
-                           set_cmd='OUTX {}',
-                           val_mapping={
-                               'RS232': '0\n',
-                               'GPIB': '1\n',
-                           })
-
-        # Channel setup
-        for ch in range(1, 3):
-
-            # detailed validation and mapping performed in set/get functions
-            self.add_parameter('ch{}_ratio'.format(ch),
-                               label='Channel {} ratio'.format(ch),
-                               get_cmd=partial(self._get_ch_ratio, ch),
-                               set_cmd=partial(self._set_ch_ratio, ch),
-                               vals=Strings())
-            self.add_parameter('ch{}_display'.format(ch),
-                               label='Channel {} display'.format(ch),
-                               get_cmd=partial(self._get_ch_display, ch),
-                               set_cmd=partial(self._set_ch_display, ch),
-                               vals=Strings())
-            self.add_parameter('ch{}_databuffer'.format(ch),
-                               channel=ch,
-                               parameter_class=ChannelBuffer)
+        aux.lock()
+        channels.lock()
+        self.add_submodule('channels', channels)
+        self.add_submodule('aux', aux)
 
         # Data transfer
         self.add_parameter('X',


qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -4,6 +4,59 @@
 from qcodes import VisaInstrument
 from qcodes.instrument.parameter import ArrayParameter
 from qcodes.utils.validators import Numbers, Ints, Enum, Strings
+from qcodes.instrument.channel import InstrumentChannel, ChannelList
+
+
+class SR830AuxArray(InstrumentChannel):  # We see that InstrumentChannel, ChannelList are something of a misnomer
```
jenshnielsen: I would call this SR830AuxChannel because thats what it is

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -4,6 +4,59 @@
...
+
+        self.add_parameter('aux_out',
+                           label='Aux output {}'.format(aux_number),
+                           get_cmd='AUXV? {}'.format(aux_number),
+                           get_parser=float,
+                           set_cmd='AUXV {0}, {{}}'.format(aux_number),
+                           unit='V')
+
+
+class SR830ChannelArray(InstrumentChannel):
```
jenshnielsen: same here I would call this SR830Channel

qcodes/instrument_drivers/stanford_research/SR830.py
sohailc: Ok

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -4,6 +4,59 @@
...
+    def __init__(self, parent, name, channel_number):
+        super().__init__(parent, name)
+
+        self._parent = parent
```
jenshnielsen: This should not be needed as its assigned in the super class

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -4,6 +4,59 @@
...
+            vals=Strings()
+        )
+
+        self.add_parameter(
+            'databuffer',
+            channel=channel_number,
+            parameter_class=ChannelBuffer
+        )
+
+        def get_parent():
```
jenshnielsen: A public accessor is probably a good idea. But I would make it a property and add it to the Super class

qcodes/instrument_drivers/stanford_research/SR830.py
sohailc: Oh yeah, I forgot

sohailc:
Should I deprecate the old driver and rename this one? Or are you still going to do that?

jenshnielsen:
@sohailc Yes please rename this one and copy the old one back in place and add a deprecation similar to what is done with other old drivers

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -154,6 +154,10 @@ class SR830(VisaInstrument):
     _N_TO_INPUT_CONFIG = {v: k for k, v in _INPUT_CONFIG_TO_N.items()}
 
     def __init__(self, name, address, **kwargs):
+
+        DeprecationWarning("This is an old driver an will be removed. Please use the the new driver: "
```
jenshnielsen: DeprecationWarnings are hidden by default https://docs.python.org/3.6/library/warnings.html#default-warning-filters so we usually use a RunTimeWarbubg or UserWarning to make sure users actually see it

WilliamHPNielsen: A new kind of bug, the RunTimeWarBug! 😄

jenshnielsen: What where I see nothing ;)

qcodes/instrument_drivers/stanford_research/SR830_channels.py
```diff
@@ -0,0 +1,602 @@
...
+class ChannelBuffer(ArrayParameter):
+    """
+    Parameter class for the two channel buffers
+
+    Currently always returns the entire buffer
+    TODO (WilliamHPNielsen): Make it possible to query parts of the buffer.
+    The instrument natively supports this in its TRCL call.
+    """
+
+    def __init__(self, name: str, channel_array: 'SR830Channel', channel: int):
```
WilliamHPNielsen: This will not work. `channel_array` must be renamed to `instrument`, although that is not a very fitting name in this case. We could consider changing that to `parent` in the parameter class.

qcodes/instrument_drivers/stanford_research/SR830_channels.py
```diff
@@ -0,0 +1,602 @@
...
+            channel (int): The relevant channel (1 or 2). The name should
+                should match this.
+        """
+        self._valid_channels = (1, 2)
+
+        if channel not in self._valid_channels:
+            raise ValueError('Invalid channel specifier. SR830 only has '
+                             'channels 1 and 2.')
+
+        if not isinstance(channel_array, SR830Channel):
```
WilliamHPNielsen: `channel_array` -> `instrument`

qcodes/instrument_drivers/stanford_research/SR830_channels.py
```diff
@@ -0,0 +1,602 @@
...
+                         shape=(1,),  # dummy initial shape
+                         unit='V',  # dummy initial unit
+                         setpoint_names=('Time',),
+                         setpoint_labels=('Time',),
+                         setpoint_units=('s',),
+                         docstring='Holds an acquired (part of the) '
+                                   'data buffer of one channel.')
+
+        self.channel = channel
+        self._instrument = channel_array.parent
```
WilliamHPNielsen: `channel_array` -> `instrument`

qcodes/instrument_drivers/stanford_research/SR830_channels.py
```diff
@@ -0,0 +1,602 @@
+from functools import partial
+import numpy as np
+
+from qcodes import VisaInstrument
+from qcodes.instrument.parameter import ArrayParameter
+from qcodes.utils.validators import Numbers, Ints, Enum, Strings
+from qcodes.instrument.channel import InstrumentChannel, ChannelList
+
+
+class SR830AuxChannel(InstrumentChannel):  # We see that InstrumentChannel, ChannelList are something of a misnomer
```
WilliamHPNielsen: The comment here should be removed.

qcodes/instrument_drivers/stanford_research/SR830_channels.py
```diff
@@ -0,0 +1,602 @@
...
+
+        self.add_parameter('Y_offset',
+                           get_cmd='OEXP? 2',
+                           get_parser=parse_offset_get)
+
+        self.add_parameter('R_offset',
+                           get_cmd='OEXP? 3',
+                           get_parser=parse_offset_get)
+
+        aux = ChannelList(self, "Aux", SR830AuxChannel, snapshotable=False)
```
WilliamHPNielsen: If the `ChannelList` is not snapshotable, then the channel should also be added directly to the instrument as a `submodule`. This will also allow for the intuitive `sr.ch1.ratio()` call.

qcodes/instrument_drivers/stanford_research/SR830_channels.py
```diff
@@ -0,0 +1,602 @@
...
+            self.setpoint_units = ('s',)
+            self.setpoint_names = ('Time',)
+            self.setpoint_labels = ('Time',)
+            self.setpoints = (tuple(np.linspace(0, N*dt, N)),)
+
+        self.shape = (N,)
+
+        params = self._instrument.parameters
+        # YES, it should be: "is not 'none'" NOT "is not None"
+        if params['ch{}_ratio'.format(self.channel)].get() is not 'none':
```
WilliamHPNielsen: Now that the driver uses channels, this line no longer works. It should be something like `self._instrument.submodules['ch{}'.format(self.channel)].ratio.get()`

qcodes/instrument_drivers/stanford_research/SR830_channels.py
```diff
@@ -0,0 +1,602 @@
...
+            self.unit = '%'
+        else:
+            disp = params['ch{}_display'.format(self.channel)].get()
```
WilliamHPNielsen: This line also has to be changed to accomodate the restructuring of the driver.

WilliamHPNielsen:
@sohailc, in addition to the changes requested above, it would be good to have an updated example notebook on how to use the channelised driver. It can be very similar to the un-channelised one.

WilliamHPNielsen:
@sohailc bump, bump, any news on this guy?

sohailc:
Currently there are other priorities. I am asked to deliver a live demo of
the WebUI via skype during Microsoft's quarterly meeting in Redmond which
is taking place next week! I will be spending some of my time on that the
coming days

qcodes/instrument_drivers/stanford_research/SR830_channels.py
sohailc: I am adding the ChannelList as a submodule. See line 388

sohailc:
@WilliamHPNielsen @jenshnielsen Can we have another look at this driver? I have added tests and removed the bug which prevented the initialization of the driver.

WilliamHPNielsen:
@sohailc Somehow the example notebook got "corrupted" so that the docs build fails. You can look into Travis' report for more details (but I think it's difficult to make sense of what's happening). Perhaps it is better to simply make a new example notebook. Perhaps @jenshnielsen can shed more light on this? In any case, that has to be fixed.

I'll do a more thorough test and review now and post the verdict later today.

jenshnielsen:
It's probably a merge gone wrong. The json of the notebook it not vaild. I suggest either re executing it locally if possible. Otherwise you can open the notebook in a text editor and copy the input fields to a new one

sohailc:
Yeah i saw that it has something to do with the notebook. I will perform
another push today or tomorrow.

qcodes/instrument_drivers/stanford_research/SR830_channels.py
```diff
@@ -0,0 +1,628 @@
...
+                self.unit = 'deg'
+            else:
+                self.unit = 'V'
+
+        if self._channel_number == 1:
+            self._instrument._buffer1_ready = True
+        else:
+            self._instrument._buffer2_ready = True
+
+    def get(self) ->np.ndarray:
```
WilliamHPNielsen: This might as well be `get_raw` to spare us a warning upon initialisation.

qcodes/instrument_drivers/stanford_research/SR830_channels.py
```diff
@@ -0,0 +1,628 @@
...
+            self._set_units('V')
+        else:
+            self.sensitivity.set_validator(self._CURR_ENUM)
+            self._set_units('A')
+
+        return mode
+
+    def _set_input_config(self, s: str) ->int:
+        if s in ['a', 'a-b']:
+            self.sensitivity.set_validator(self._VOLT_ENUM)
```
WilliamHPNielsen: `set_validator` is deprecated. We should use `self.sensitivity.vals = ` instead.

qcodes/instrument_drivers/stanford_research/SR830_channels.py
```diff
@@ -0,0 +1,628 @@
...
+        self.add_parameter('sync_filter',
+                           label='Sync filter',
+                           get_cmd='SYNC?',
+                           set_cmd='SYNC {}',
+                           val_mapping={
+                               'off': 0,
+                               'on': 1,
+                           })
+
+        def parse_offset_get(s):
```
WilliamHPNielsen: Normal `Parameters` should always be scalar (or strings). The `OEXP` command controls both offset and expansion (scale factor) and should be mapped to two different `Parameters`, just like on the front panel. I realise that this PR did not introduce this mistake, but it should correct it.

qcodes/instrument_drivers/stanford_research/SR830_channels.py
```diff
@@ -0,0 +1,628 @@
...
+                         unit='V',  # dummy initial unit
+                         setpoint_names=('Time',),
+                         setpoint_labels=('Time',),
+                         setpoint_units=('s',),
+                         docstring='Holds an acquired (part of the) '
+                                   'data buffer of one channel.')
+
+        self._channel_number = channel_number
+        self._instrument = instrument.parent
+        self._channel = None  # This is set by the time we call "prepare_buffer_readout"
```
WilliamHPNielsen: Why is `_channel` not simply assigned here? It confuses both me and `mypy`.

qcodes/instrument_drivers/stanford_research/SR830_channels.py
```diff
@@ -0,0 +1,628 @@
...
+        self.write('DDEF {}, {}, {}'.format(channel, disp, ratio_val))
+        self._buffer_ready = False
+
+    def _set_units(self, unit: str) ->None:
+        # TODO:
+        # make a public parameter function that allows to change the units
+        for param in [self.X, self.Y, self.R, self.sensitivity]:
+            param.unit = unit
+
+    def _get_input_config(self, s: str) ->int:
```
WilliamHPNielsen: Doesn't this actually return a string?

PR #814, opened 2017-10-25
Driver/Mini-Circuits

StevenWaddy:
Ethernet driver for Mini-Circuits RF switch:
* RC-4SPDT
* RC-1SP4T

@Dominik-Vogel, @johnhornibrook 
## Resources
 * Programming manual: https://ww2.minicircuits.com/softwaredownload/Prog_Manual-2-Switch.pdf
 * RC-4SPDT https://ww2.minicircuits.com/pdfs/RC-4SPDT-A18.pdf
 * RC-1SP4t https://ww2.minicircuits.com/pdfs/RC-1SP4T-A18.pdf


johnhornibrook:
Hi, I use these drivers in the lab, and have no problems with either the code or the functionality.

Dominik-Vogel:
Thanks @StevenWaddy!
I briefly looked at the datasheet of the RC-1SP4T and it looks like one can set all the four switches individually. It is probably not relevant for your application, but I think that the feature of connecting the COM signal to e.g. outputs 1 and 2 might be wished for in the future. Because implementing this new feature would slightly break the interface of the driver and would make it inconvenient to upgrade, and on the otherhand implementing it straight away is probably not very time intensive.
Would you have the time to implement the RC-1SP4T analogously to the RC4SPDT with channels? Or alternatively one could use a bitmask...

Dominik-Vogel:
Looking at the programming guide it seems to me, that it would be a good idea to unify the two drivers. At a first glance it looks like we can use the RC4SPDT driver for both the RC4SPDT and the RC-1SP4T without any changes

StevenWaddy:
The drivers should work for:
*RC-1SP4T
*RC-2SP4T
*RC-1SPDT
*RC-2SPDT
*RC-3SPDT
*RC-4SPDT
*RC-8SPDT
for both the -A12 and A18 versions
but has only been tested on the RC-1SP4T-A18 and RC-4SPDT-A18

@Dominik-Vogel , both drivers have now channels and the same interface.

The RC-xSP4T switches can only be switched between common and open or one of 1,2,3 or 4.
It is not possible to connect multiple outputs to the common or to each other the instrument 
will ignore you if you try and it would not be possible to maintain Z=50Ohm if you did.
[pg 96 of the programming manual]

The RC-xSPDT only between common and 1 or 2

@johnhornibrook do you want to double check the drivers still work.

Dominik-Vogel:
@johnhornibrook do you have an update of the status of the driver? does it still work, so that I can merge it?

Dominik-Vogel:
@johnhornibrook tested the driver and it still works, so that it is ready to merge.

jenshnielsen:
@Dominik-Vogel remember to close any issue in stationQ

PR #816, opened 2017-10-25
[WIP] Ats9360 with channels

jenshnielsen:
Driver for ATS9360 and new controller that supports capturing from both physical channels as well as with and without software demodulation. The controller works using the new channels api. Per default it does not define any channels but you can setup channels that will return data in any 0d to 2d shape of buffers, records and samples and with and without sorfware demodulation

@nataliejpg It would be great if you can have a look at the notebook in https://github.com/jenshnielsen/Qcodes/blob/9a3822fb56134c7f2693306f068e95cdab23e58c/docs/examples/driver_examples/Qcodes%20example%20with%20Alazar%20ATS9360.ipynb
to see if it does what you want.

@damazter

This contains some changes to the existing driver in 
`ATS.py`
`ATS9870.py`

These are mainly:

* Changes to casting of a few parameters passed to functions. i.e. a cast of the address in `AlazarPostAsyncBuffer` to  `ctypes.cast(buf.addr, ctypes.c_void_p)` I found that this was
needed to get the code to actually run on some machines (As far as I can tell it depends on the windows version) 
* Cleanup to the Ctypes calling to simplify a bit avoiding to create numpy arrays of one element
* Handle cards with non 8 bit resolution
* Add aux_io_mode and aux_io_param

I would like your input on these before we merge this. I do not have a 9870 card to test with at the moment.


qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -574,83 +614,146 @@ def acquire(self, mode=None, samples_per_record=None,
...
+            # buffer handling from acquisition
+            buffers_completed = 0
+            bytes_transferred = 0
+            buffer_timeout = self.buffer_timeout._get_byte()
+            self.buffer_timeout._set_updated()
+
+            buffer_recycling = (self.buffers_per_acquisition._get_byte() >
+                                self.allocated_buffers._get_byte())
+            done_setup = time.clock()
+            while (buffers_completed < self.buffers_per_acquisition._get_byte()):
```
damazter: the outer brackets here are not PEP8 compatible right?

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -574,83 +614,146 @@ def acquire(self, mode=None, samples_per_record=None,
...
+                # Wait for the buffer at the head of the list of available
+                # buffers to be filled by the board.
+                buf = self.buffer_list[buffers_completed % allocated_buffers]
+                self._ATS_dll.AlazarWaitAsyncBufferComplete.argtypes = [ctypes.c_uint32, ctypes.c_void_p, ctypes.c_uint32]
```
damazter: do you need to update the types for every buffer you acquire. Couldn't this line be called only once at the beginning of the driver?
we want to keep this part of the driver as fast as possible

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -574,83 +614,146 @@ def acquire(self, mode=None, samples_per_record=None,
...
+
+                # TODO(damazter) (C) last series of buffers must be handled
+                # exceptionally
+                # (and I want to test the difference) by changing buffer
+                # recycling for the last series of buffers
+
+                # if buffers must be recycled, extract data and repost them
+                # otherwise continue to next buffer
+                if buffer_recycling:
+                    acquisition_controller.handle_buffer(buf.buffer, buffers_completed)
```
damazter: This is incompatible with the old `handle_buffer` callback. If we want to change it, it should also be changed in `class AcquisitionController(Instrument):` which it is not at this moment

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -574,83 +614,146 @@ def acquire(self, mode=None, samples_per_record=None,
...
+            self._call_dll('AlazarAbortAsyncRead', self._handle)
+        time_done_abort = time.clock()
         # -----cleanup here-----
         # extract data if not yet done
         if not buffer_recycling:
-            for buf in self.buffer_list:
-                acquisition_controller.handle_buffer(buf.buffer)
-
+            for i, buf in enumerate(self.buffer_list):
+                acquisition_controller.handle_buffer(buf.buffer, i)
```
damazter: same point here

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -574,83 +614,146 @@ def acquire(self, mode=None, samples_per_record=None,
...
+                     (buffers_completed, buffers_per_sec))
+        logger.info("Captured %d records (%f records per sec)" %
+                     (records_per_buffer * buffers_completed, records_per_sec))
+        logger.info("Transferred {:g} bytes ({:g} "
+                     "bytes per sec)".format(bytes_transferred, bytes_per_sec))
+        logger.info("Pre capture setup took {}".format(setup_time))
+        logger.info("Capture took {}".format(capture_time))
+        logger.info("abort took {}".format(abort_time))
+        logger.info("handling took {}".format(handling_time))
+        logger.info("free mem took {}".format(free_mem_time))
```
damazter: Would it be sensible to add an option to disable all this timing debug info, we don't always want to delay our measurements by printing this stuff

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -678,18 +781,22 @@ def _call_dll(self, func_name, *args):
...
         update_params = []
         for arg in args:
-            if isinstance(arg,AlazarParameter):
+            if isinstance(arg, AlazarParameter):
                 args_out.append(arg._get_byte())
                 update_params.append(arg)
             else:
                 args_out.append(arg)
-
+        logger.debug("calling dll func {} with args: \n {}".format(func_name, args_out))
```
damazter: does this line delay the code? if so, we might want to disable it

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -810,10 +921,13 @@ def __init__(self, name=None, label=None, unit=None, instrument=None,
 
         self._set(value)
 
-    def get(self):
+    def get_raw(self):
```
damazter: why is this not just `get`?
`get_raw` seems to imply some byte code or something, which it does not do.

qcodes/instrument_drivers/AlazarTech/ATS9870.py
```diff
@@ -150,6 +150,23 @@ def __init__(self, name, **kwargs):
...
                            value=0,
                            vals=validators.Ints(min_value=0))
+        self.add_parameter(name='aux_io_mode',
+                           parameter_class=AlazarParameter,
+                           label='AUX I/O Mode',
+                           unit=None,
+                           value='AUX_IN_AUXILIARY',
+                           byte_to_value_dict={0: 'AUX_OUT_TRIGGER',
+                                               1: 'AUX_IN_TRIGGER_ENABLE',
+                                               13: 'AUX_IN_AUXILIARY'})
```
damazter: is 13 a valid value? we could not find it in the manual

qcodes/instrument_drivers/AlazarTech/ATS9870.py
```diff
@@ -150,6 +150,23 @@ def __init__(self, name, **kwargs):
...
+
+        self.add_parameter(name='aux_io_param',
+                           parameter_class=AlazarParameter,
+                           unit=None,
+                           value='NONE',
+                           byte_to_value_dict={0: 'NONE',
+                                               1: 'TRIG_SLOPE_POSITIVE',
+                                               2: 'TRIG_SLOPE_NEGATIVE'})
```
damazter: is `label` missing here?

qcodes/instrument_drivers/AlazarTech/ATS9870.py
jenshnielsen: For the 9360 it is but maybe that we have to overwrite this in that class

jenshnielsen:
Not relevant anymore, The controllers are in qdev-wrappers

PR #817, opened 2017-10-26
RS SGS100A driver

ThorvaldLarsen:
Added parameter for IQ modulation on/off in SGS100A driver.


jenshnielsen:
@ThorvaldLarsen Thanks, looks good to me. I will just push to remove the `.DS_Store` files, that you probably did not want to commit, it that is ok?

ThorvaldLarsen:
That sounds good to me. Thanks.

jenshnielsen:
@ThorvaldLarsen Looks good, only things I changed are

* Remove the `.DS_Store` files (they are mac os icon cache I think)
* Add DS_Store to `.gitignore` so they cannot easily be committed again
* remove trailing whitespace from a single line. We prefer not to have these as they can easily give conflicts

PR #822, opened 2017-10-27
add debugging and make the final write more robust

### LINKED ISSUE #821: Bug: No data gets saved when a parameter returns (too many?) nans
The DataSet cannot match save ranges if an array consists of nans.

### Steps to reproduce

The following MWE reproduces the unwanted behaviour:

```
import qcodes as qc
from qcodes.instrument.parameter import MultiParameter

# set up the instruments
mydac = DummyInstrument('dac', gates=['ch1', 'ch2'])
mydmm = DummyInstrument('dmm', gates=['v1', 'v2'])

# multiparameter designed to break everything

class CrazyParameter(MultiParameter):
    
    def __init__(self, name, instrument, names=('first', 'second'),
                 shapes=((), ())):
        
        super().__init__(name=name, names=names, shapes=shapes,
                         instrument=instrument)
        
    def get(self):
        return (13,)  # THIS IS THE OFFENSIVE LINE CREATING "WRONG" DATA

mydmm.add_parameter(name='crazy',
                    parameter_class=CrazyParameter)

station = qc.Station(mydac, mydmm)

loop = qc.Loop(mydac.ch1.sweep(0, 1, num=10), delay=0.1).each(mydmm.crazy)
dataset = loop.run()
```

### Expected behaviour
The dataset should get written to disk

### Actual behaviour
The save ranges can never be matched (see `gnuplot_format`'s `write` method), so the arrays are always skipped.

### System
**MAC OSX**

**master**

**qcodes commit**
3d7baef4b7c4e7b1ae815965490ef9dfe7adf2da 

@jenshnielsen @sohailc

WilliamHPNielsen:
Fixes #821 

- Add more `log.debug` to the data saving process
- Make the `qc.Loop`'s final writing of data more robust

@jenshnielsen @sohailc

### DIFF
--- a/qcodes/data/data_set.py
+++ b/qcodes/data/data_set.py
@@ -11,6 +11,8 @@ from .io import DiskIO
 from .location import FormatLocation
 from qcodes.utils.helpers import DelegateAttributes, full_class, deep_update
 
+log = logging.getLogger(__name__)
+
 
 def new_data(location=None, loc_record=None, name=None, overwrite=False,
              io=None, **kwargs):
@@ -387,8 +389,11 @@ class DataSet(DelegateAttributes):
         self.last_store = time.time()
         if (self.write_period is not None and
                 time.time() > self.last_write + self.write_period):
+            log.debug('Attempting to write')
             self.write()
             self.last_write = time.time()
+        else:
+            log.debug('.store method: This is not the right time to write')
 
     def default_parameter_name(self, paramname='amplitude'):
         """ Return name of default parameter for plotting
@@ -464,7 +469,7 @@ class DataSet(DelegateAttributes):
             return
         self.formatter.read_metadata(self)
 
-    def write(self, write_metadata=False):
+    def write(self, write_metadata=False, only_complete=True):
         """
         Writes updates to the DataSet to storage.
         N.B. it is recommended to call data_set.finalize() when a DataSet is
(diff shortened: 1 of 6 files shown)


qcodes/data/data_set.py
```diff
@@ -11,6 +11,8 @@
 from .location import FormatLocation
 from qcodes.utils.helpers import DelegateAttributes, full_class, deep_update
```
jenshnielsen: There is a bunch of old style non named loggers in this file. I think we should convert those too

jenshnielsen:
Looks good :+1: for a test based on #821 

Does finalise gets called if you keyboard interrupt ?

qcodes/tests/test_loop.py
```diff
@@ -3,16 +3,29 @@
...
+
+class NanReturningParameter(MultiParameter):
+
+    def __init__(self, name, instrument, names=('first', 'second'),
+                 shapes=((), ())):
+
+        super().__init__(name=name, names=names, shapes=shapes,
+                         instrument=instrument)
+
+    def get(self):  # this results in a nan-filled DataArray
```
jenshnielsen: might as well make this get_raw to not get the warning?

jenshnielsen:
The abstract write method in `format.py` should probably also be changed to match the 2 concrete implementations

PR #823, opened 2017-10-27
Fix __str__ method in Parameter when there is no parent instrument

YakBizzarro:
If a parameter doesn't have a parent instrument, the __str__ method fails. This commit fix this.

@nulinspiratie @WilliamHPNielsen

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -227,10 +227,10 @@ class _BaseParameter(Metadatable, DeferredOperations):
 
     def __str__(self):
         """Include the instrument name with the Parameter name if possible."""
-        inst_name = getattr(self._instrument, 'name', '')
-        if inst_name:
+        try:
+            inst_name = self._instrument.name
             return '{}_{}'.format(inst_name, self.name)
-        else:
+        except AttributeError:
             return self.name
 
     def __repr__(self):


jenshnielsen:
I can't reproduce this issue

```
In [6]: a= Parameter('foo', get_cmd=None, set_cmd=None)
In [7]: a._instrument
In [8]: a._instrument is None
Out[8]: True
In [9]: str(a)
Out[9]: 'foo'
```

Can you give a bit more details on how this happens

WilliamHPNielsen:
@YakBizzarro No it does not, at least not the part you changed. The `getattr` returns an empty string if the attribute `name` does not exist.

WilliamHPNielsen:
Here's one way to reproduce it:

```
a= Parameter('foo', get_cmd=None, set_cmd=None)
delattr(a, '._instrument')
str(a)
```

But why on Earth would anyone do such a crazy thing? Or are you subclassing Parameters and forgetting to assign `self._instrument = None`, @YakBizzarro?

jenshnielsen:
One potentially issue is that _BaseParameter does not have a default value for Instrument but that should break if you try to create one without an instrument

jenshnielsen:
@YakBizzarro I am still curious how this can happen.

nulinspiratie:
@YakBizzarro @jenshnielsen Since `_instrument` is set in the `__init__` of `BaseParameter`, the only three way I can think of for this to occur is either the attribute is deleted (as shown by @WilliamHPNielsen), or that `__init__`  of BaseParameter is never called, or that `__str__` is called before the call to `BaseParameter.__init__`. All three should be avoided, though I don't see a problem with adding this extra try/except.

As a side note, I think the `instrument` kwarg of BaseParameter should have a default value None.

peendebak:
There is an issue if the '_instrument' field is changed, but not deleted. The commit solves that case, and is fully compatible with all the normal cases.

WilliamHPNielsen:
@peendebak it's not fully compatible as shown by the tests not passing. It's a small change, though, an extra `_` should be prepended to `self.name` in the failing case.

jenshnielsen:
@WilliamHPNielsen The issue is that an extra `_` is prepended in the case where instrument is `None` since that does not trigger a error so it takes the first code path

YakBizzarro:
Sorry for the long silence, other thing to do.
The issue arose when I tried to create an enhanced version of the [VoltageDivider ](http://qcodes.github.io/Qcodes/api/generated/qcodes.instrument_drivers.html#module-qcodes.instrument_drivers.devices) (you can see it [here](https://gist.github.com/YakBizzarro/9b8fa750dabc9656d4540efc55209da0) )

I agree that I should set _instrument to None, but it should not fail in this way if missing

YakBizzarro:
I checked again, and effectively _instrument is always set in the constructor. At the time I wrote this fix, it seems that was not the case. So I drop my fix.

PR #826, opened 2017-10-27
[WIP] Decadac fine voltage 

jenshnielsen:
This probably need some more testing

There is an issue with non linear steps around 0 V


Dominik-Vogel:
The changes in this pull request have been implemented in a customized version of this instrument in wrappers: PR#1

qcodes/instrument_drivers/Harvard/Decadac.py
```diff
@@ -327,7 +362,7 @@ def __init__(self, parent, name, slot, min_val=-5, max_val=5):
                            val_mapping=slot_modes)
 
         # Enable all slots in coarse mode.
-        self.slot_mode.set("Coarse")
+        self.slot_mode.set("Fine")
```
spauka: Just an FYI that this change will cause issues for us on our experiments.

Perhaps we should add an unknown state that is set when the instrument is initialized, given that there is no (known) way of reading out this value?

jenshnielsen: @spauka I know, this is not meant to go in in the current format (hence the [WIP] tag) 

Did you see my notification on Slack, If you are using the Decadac driver and do ramping by setting delay and step as in the tutorial here https://github.com/QCoDeS/Qcodes/blob/master/docs/examples/Tutorial.ipynb
it will most likely jump to wrong value at the beginning of the ramp.  Affected QCoDeS master is between the 16th of October and Yesterday

This does not affect you if you use the ramping build into the decadac driver.

spauka: Of course, sounds good :) Thought it worth mentioning the possible solution of setting an unknown state, I hadn't been using delays/steps so this hasn't affected me as yet, but will keep this in mind.

jenshnielsen: Sure, thanks for bringing it up. Better safe than sorry

jenshnielsen:
It does not seem likely that we are continuing development on the decadac driver

PR #828, opened 2017-10-31
Driver/awg5014 dev

nataliejpg:
Fixes #issuenumber.

Changes proposed in this pull request:
- 
- 
- 

@mention one core developer

PR #832, opened 2017-10-31
set_parser for seq element setting added

jenshnielsen:
@nataliejpg Cleanup of #828 are there any relevant changed that I missed


jenshnielsen:
@WilliamHPNielsen Any inside if this makes sense

Dominik-Vogel:
@nataliejpg do you remember why you added the set parser?

nataliejpg:
there was an error when I tried to set a jump to which is what prompted me to do this but I can't remember exactly what, sorry :/ @ThorvaldLarsen  do you remember? i guess you could play on the test setup but otherwise it seems like there should be something which makes sure people set ints as things other than ints don't make sense here.  I couldn't remember what the best way to do this between validator, val_mapping, set_parser etc was so I just made it so the bug stopped...

ThorvaldLarsen:
the problem with the awg driver was when I wanted to step the sequence number with sequence_pos in do1d(). do1d seems to pass floats while the sequence_pos expected an integer hence throwing an error.

jenshnielsen:
I made this a bit safer by adding a validator that will check if a float is close to an int and pass in such cases. This can be used together with the set_parser. 

I also changed the set parser to int(round(x)) in case what you are trying to set the value to is just under the value. int(x) will always "round" down when it casts

PR #834, opened 2017-10-31
Znb spectroscopy

jenshnielsen:
Cleanup of #830


jenshnielsen:
This now only covers non controversial changes, pep8 and a few extra parameters,

nataliejpg:
also just remembered that 'power' should have float as a get_parser not int

jenshnielsen:
@nataliejpg The readout power that you added in this pr or the regular source power. The manual is a bit unclear but it looks like the source power only takes integers

nataliejpg:
it tried to turn -9.99 into an int at some point when I initialised the other day and it tried to get all the parameters so I think this is one of those rare things. Maybe it should try round or just return the float, whatever we think is less dangerous. But i cleared the queue etc and reinitialised and it was definitely trying to get a power of -9.99.

jenshnielsen:
Well if it can be set to floating point values it should be a float.

nataliejpg:
I've never tried, I don't know if it can but I do know that in this one instance it thought it was one at least (but maybe for the instrument -9.99 IS -10) . That's why I just suggested changing the get_parser and not the set part as I don't know if we want to let it be set to floats if it actually cant handle that but that at least sometimes it thinks it is one

jenshnielsen:
Ok, I should probably borrow one at some point and verify all the parameters but for now lets make the parser int(round(float())) ?

nataliejpg:
sounds good. I think it would be good to check at some point and I guess to see what happens if you try to set a float but I also don't feel like it would clear everything up as actually I used this driver tons of times before getting this error so I think there is a good chance it is hard to reproduce.

PR #836, opened 2017-10-31
Add a try catch to getting the time tuple

jenshnielsen:
@nataliejpg This was committed to the wrappers as part of your post t2 branch. Can you remember why it was needed?

### DIFF
--- a/qcodes/monitor/monitor.py
+++ b/qcodes/monitor/monitor.py
@@ -49,7 +49,10 @@ def _get_metadata(*parameters) -> Dict[float, list]:
         # convert to string
         meta['value'] = str(meta['value'])
         if meta["ts"] is not None:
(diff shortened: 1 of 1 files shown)


nataliejpg:
oh gosh yes I remember this. It was part of the horror of trying to use the alazar branch and the newer version of qdev-dk/qcodes and for some reason the monitor started breaking and it would throw a key error here (I think it was) but whenever I did any digging I couldn't see why that would happen especially as it hadn't happened before so I just made it not care and then the monitor worked fine but this should definitely die as it was a late night; need monitor to work tomorrow change to solve exactly the sort of problems which I hope will go away when I'm not mixing and matching so many versions of things. I would just kill it and I can always redo it and record the error message so it can be fixed in a more sane way if it comes up again but I hope it won't

jenshnielsen:
Closing this then. Will revisit if needed

PR #838, opened 2017-11-01
added slot_mode_default to decadac channel

Dominik-Vogel:
I want to add this class variable so that the implementation of the customized driver (eg. T3) becomes easier.


qcodes/instrument_drivers/Harvard/Decadac.py
```diff
@@ -327,7 +328,7 @@ def __init__(self, parent, name, slot, min_val=-5, max_val=5):
                            val_mapping=slot_modes)
 
         # Enable all slots in coarse mode.
-        self.slot_mode.set("Coarse")
+        self.slot_mode.set(DacSlot.SLOT_MODE_DEFAULT)
```
jenshnielsen: actually this should be self.SLOT_MODE_DEFAULT I would say

PR #841, opened 2017-11-02
update installation instructions

jenshnielsen:
Mainly install all dependencies from Anaconda


sl4pp3r:
Did anyone check this file?  It no longer works.
The spyder line is missing its indentation .. and we all know how python *loves* indentation :)

Also .. the link from http://qcodes.github.io/Qcodes/start/index.html#installation is broken ...   it seems to contain "useful" descriptive text, but the link destination is broken.

PR #844, opened 2017-11-02
make print_readable_snapshot work with value-less parameters

WilliamHPNielsen:
If a parameter does not have a value in the snapshot (e.g. because it is an ArrayParameter), the `print_readable_snapshot` function can not iterate through all parameters and values. This PR fixes that.

- Make `print_readable_snapshot` handle missing values

@jenshnielsen @AdriaanRol

### DIFF
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -228,12 +228,10 @@ class InstrumentBase(Metadatable, DelegateAttributes):
             name = snapshot['parameters'][par]['name']
             msg = '{0:<{1}}:'.format(name, par_field_len)
 
-            # in case of e.g. ArrayParameters, the parameter may not have
+            # in case of e.g. ArrayParameters, that usually have
+            # snapshot_value == False, the parameter may not have
             # a value in the snapshot
-            try:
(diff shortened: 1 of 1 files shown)


qcodes/instrument/base.py
```diff
@@ -227,7 +227,14 @@ def print_readable_snapshot(self, update=False, max_chars=80):
         for par in sorted(snapshot['parameters']):
             name = snapshot['parameters'][par]['name']
             msg = '{0:<{1}}:'.format(name, par_field_len)
-            val = snapshot['parameters'][par]['value']
+
+            # in case of e.g. ArrayParameters, the parameter may not have
```
jenshnielsen: I would change the docstring to say in case snapshot_value is False otherwise 👍

jenshnielsen: Perhaps a slight preference for `snapshot['parameters'][par].get('value', 'Not available')` in stead of try except

WilliamHPNielsen: I wholeheartedly agree with your viewpoint.

PR #845, opened 2017-11-03
Driver/Yokogawa GS200

spauka:
Complete driver for yokogawa GS200 Voltage/Current source. (May interest @sohailc)

- Completed driver for GS200 voltage/current source
- Add a nothing validator that rejects everything (used when trying to set VOLT/CURR in CURR/VOLT mode)
- Add a channel-list validator for parameters that return a channel.

I think this handles almost everything in the GS200, so for all intents and purposes it can be considered complete?

@jenshnielsen


qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -1,5 +1,144 @@
...
-from qcodes.utils.validators import Numbers
+from qcodes import VisaInstrument, InstrumentChannel
+from qcodes.instrument.parameter import ManualParameter
+from qcodes.utils.validators import Numbers, Bool, Enum, Nothing, Ints
+
+def float_int(val):
+    """
+    Parses int that are returned in exponentiated form (i.e. 1E0)
+    """
+    return int(float(val))
```
jenshnielsen: I wonder if we want to do this as int(round(float(val))) if for some reason float(val) evaluates to just below an int e.g. `9.9` int(val) vil round down to 9

spauka: In this instance they are all integers, but point taken as a general rule. Changed...

sohailc:
I have made quite a few changes to this driver as I did not like the overall structure of the original one. The main thing is this: Do not make a parameter and change its attributes, like the command string and validators later. This makes it hard to understand what is going on. This for instance happens with the range parameter, where the validators and commands are changed based on the output mode (current or voltage) and if we are in auto mode or not.  

I am solving this by introducing voltage_range and current_range with fixed validators and commands. The an attribute "range" either points to "voltage_range" or "current_range" based on the output mode. 

This is happening in the "current" and "voltage" parameters as well, there the command string is changed based on whether we are in auto mode or not. Don't do that! Just make a set_voltage/set_current function and let it take care of sending the right command. 

Another thing I would avoid is using the parsers as validators. If validation parameters are dynamic based on output mode of other settings, just make setter/getter functions and do the validation in there.  This makes for much more straight forward code. 

I have tested my code as best I can with an actual instrument. I will put the test notebook in the docs/examples folder with a later commit. I could not test the measurement monitoring bit, as my instrument is missing that functionality. 

Please have a look at my code and comment.

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -1,5 +1,155 @@
...
+
+from qcodes import VisaInstrument, InstrumentChannel
+from qcodes.utils.validators import Numbers, Bool, Enum, Nothing, Ints, Validator
+
+
+def float_round(val):
+    """
+    Rounds a floating number represented as a string
+
+    Parameters
```
jenshnielsen: @sohailc  Qcodes uses [google](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) python doc style and not [numpy doc](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html#example-numpy).

sohailc: Oh begging your pardon. I was under the impression that I was using the google python doc style. I will review the style an correct accordingly

WilliamHPNielsen: On that note, I think that all core devs are supposed to use type hints.

sohailc: I have never used type hints so I still need to learn to use them. I am sure its not rocket science though :-) I will use type hints from now on

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -10,29 +160,310 @@ class GS200(VisaInstrument):
...
+                           get_parser=int)
 
-        self.initialise()
+        # Check if monitor is present, and if so enable measurement
+        monitor_present = '/MON' in self.ask("*OPT?")
+        measure = GS200_Monitor(self, 'measure', monitor_present)
+        self.add_submodule('measure', measure)
+
+        # Reset function
+        self.add_function('reset', call_cmd='*RST')
```
spauka: Why remove on/off functions here?

I personally find them convenient to use, but you prefer not to have them?

sohailc: The "add_function" method is rightly controversial as it does not add anything over just defining a function with "def function(self, ...)". What is the point of: 

self.add_function('on', call_cmd=self.on)
self.add_function('off', call_cmd=self.off)

Calling "on" will simply call the method "on", which it would do anyway even if we did not have the "add_function" calls. So why have them?

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -10,29 +160,310 @@ class GS200(VisaInstrument):
...
+        Parameters
+        ----------
+        output_level: float
+        """
+        auto_enabled = self.auto_range()
+
+        if not auto_enabled:
+            self_range = self.range()
+        else:
+            mode = self.source_mode()
```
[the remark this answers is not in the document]
sohailc: Thats a good point. I have created a cached version of the mode and range so that it does not query the instrument each time.

sohailc:
Hi everyone, can we have a look at this driver again?

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -1,5 +1,168 @@
...
+
+    To measure:
+    `GS200.measure.measure()`
+
+    Args:
+        parent (GS200)
+        name (str): instrument name
+        present (bool):
+    """
+    def __init__(self, parent: GS200, name: str, present: bool):
```
WilliamHPNielsen: `GS200` is not defined yet at this point. I think `'GS200'` should do the trick.

WilliamHPNielsen:
Slightly unrelated, but @sohailc, could you make sure that your commits are in your name? It is nice to have this information down the line.

WilliamHPNielsen:
Perhaps this is helpful: https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/ @sohailc

sohailc:
Oh ok. I will make sure my name appears correctly

spauka:
Made a few small changes to the caching to try and pull values in case of a validation failure, and some spelling/style fixes. 

There were also a few syntax errors in the usage example.

@sohailc @WilliamHPNielsen all the changes look great! Are we quite close to merging this?

sohailc:
Let me have a look tomorrow

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -10,29 +167,343 @@ class GS200(VisaInstrument):
...
+                           set_parser=int)
+
+        self.add_parameter('current_limit',
+                           label='Current Protection Limit',
+                           unit='I',
+                           vals=Numbers(1e-3, 200e-3),
+                           get_cmd=":SOUR:PROT:CURR?",
+                           set_cmd=":SOUR:PROT:CURR {:.3f}",
                            get_parser=float,
-                           vals=Numbers(-10, 10))
```
jenshnielsen: Looks like the validator disappeared from the voltage was that intentional?

ThorvaldLarsen:
This doesn't break anything for us (the exciting driver is very basic). It would indeed be nice to have this driver for the GS200.

nataliejpg:
Looks good to me, haven't tested but as @ThorvaldLarsen says there was very little in the old one so all we used it for was set/get voltage and I'm confident that you guys checked this one can do that ;). Thanks for this

WilliamHPNielsen:
There are currently two (small) codacy issues and I think that they are both relevant to address before merging.

qcodes/instrument_drivers/yokogawa/GS200.py
spauka: Validation is now done in the _set_output function, so this is intentional.

spauka:
@WilliamHPNielsen driver looks good to me 👍  Codacy errors are resolved!

@sohailc depending on your editor you should be able to set it up to automatically strip trailing whitespace. Recommended for python scripting :)

sohailc:
@spauka Hey thanks for the tip :-)

PR #851, opened 2017-11-06
fix: use latest value instead of raw value in ramp

nulinspiratie:
- Use latest value in get_ramp_value as starting value

After switching the order of scaling and ramping (#788), the starting value should no longer be the raw value, but the scaled value.


jenshnielsen:
@nulinspiratie thanks. Is it ok that I write some tests for this and push to your branch?

jenshnielsen:
@nulinspiratie  I added some tests for this. @QCoDeS/core would be great if anyone wants to review the tests

nulinspiratie:
@jenshnielsen I wasn't available the past days, thanks for adding tests to it!

PR #852, opened 2017-11-06
Validators/permissive multiples

### LINKED ISSUE #765: Multiples validator only works for Ints
The `Multiples(Ints)` validator checks if a give value is a multiple of some other value (the `divisor`). 

This validator only works for `Ints` and there is no reason for this. 

A perfectly valid example would be requiring for values to be multiples of 20ns. 

The current value would then be valid `40e-9` as this would return 0.0. Currently this will raise an error as 40e-9 (which is a multiple of 20e-9) raises an error as it is not an Int. I propose dropping the constraint of it having to be an Int.

WilliamHPNielsen:
Fixes #765

- Add a validator for non-integer multiples of a base value
- Add a test (@AdriaanRol, perhaps you can think of more/better corner cases?)

The new validator, `PermissiveMultiples` allows integer multiples of a given base value to slip through. It allows negative multiples as well, and therefore by definition always 0. It is subject to rounding errors, so a user-specificied absolute precision is in principle required.

@jenshnielsen

### DIFF
--- a/qcodes/tests/test_validators.py
+++ b/qcodes/tests/test_validators.py
@@ -4,7 +4,7 @@ import numpy as np
 
 from qcodes.utils.validators import (Validator, Anything, Bool, Strings,
                                      Numbers, Ints, PermissiveInts,
-                                     Enum, MultiType,
+                                     Enum, MultiType, PermissiveMultiples,
                                      Arrays, Multiples, Lists, Callable, Dict)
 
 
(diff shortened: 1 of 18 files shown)


qcodes/tests/test_validators.py
```diff
@@ -451,6 +451,38 @@ def test_divisors(self):
...
+
+    def test_not_passing(self):
+        for divind, div in enumerate(self.divisors):
+            val = PermissiveMultiples(div)
+            for mult in self.not_multiples[divind]:
+                with self.assertRaises(ValueError):
+                    print(div, mult)
+                    val.validate(mult)
+
+    # finally, a quick test that the precision is indeed setable
```
jenshnielsen: Where ?

qcodes/utils/validators.py
```diff
@@ -307,6 +311,64 @@ def validate(self, value, context=''):
...
+            self._mulval.validate(abs(value))
+        else:
+            # floating-point division cannot be trusted, so we try to
+            # multiply our way out of the problem by constructing true
+            # multiples in the relevant range and see if `value` is one
+            # of them (within rounding errors)
+            divs = int(divmod(value, self.divisor)[0])
+            true_vals = np.array([n*self.divisor for n in range(divs, divs+2)])
+            abs_errs = [abs(tv-value) for tv in true_vals]
+            if min(abs_errs) > self.precision:
```
jenshnielsen: Should this have a relative tolerance too?

jenshnielsen:
2 Inline comments otherwise looks good

qcodes/tests/test_validators.py
WilliamHPNielsen: Woops... nice catch.

qcodes/utils/validators.py
WilliamHPNielsen: Mimicking `np.allclose`? I am a bit undecided. I think for most cases it will suffice to say "be precise to within 0.5 nV".

jenshnielsen: Yes as long as people are not using "strange" units its fine, we can always add later if needed

PR #854, opened 2017-11-07
Validators/expose valid vals

WilliamHPNielsen:
For some secret plans that I have regarding instrument mocks (EDIT: no longer secret, #859) I would like to have a way to make a QCoDeS Parameter tell me a valid value it could have. I think it makes sense to get such a value from the Parameter's validator. I therefore propose to have all validators expose a list of at least one valid value through a property called `valid_values`. This has the nice side effect that all validators actually have something to do in their `__init__`, which makes it less silly to have the base class demanding them to implement `__init__`.

- Make all validators expose valid values
- Add tests to check that the valid values are, well, valid

@jenshnielsen @sohailc

### DIFF
--- a/qcodes/utils/validators.py
+++ b/qcodes/utils/validators.py
@@ -1,5 +1,5 @@
 import math
-from typing import Union, Tuple
+from typing import Union, Tuple, cast
 
 import numpy as np
 
@@ -171,7 +171,10 @@ class Numbers(Validator):
         else:
             raise TypeError('min_value must be a number')
 
-        if isinstance(max_value, self.validtypes) and max_value > min_value:
+        valuesok = max_value > min_value  # type: ignore
+        valuesok = cast(bool, valuesok)
+
+        if isinstance(max_value, self.validtypes) and valuesok:
             self._max_value = max_value
         else:
             raise TypeError('max_value must be a number bigger than min_value')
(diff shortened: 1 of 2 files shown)


qcodes/utils/validators.py
```diff
@@ -45,12 +45,20 @@ class Validator:
     base class for all value validators
     each should have its own constructor, and override:
 
+    __init__: here a private attribute, _valid_values, should be set.
```
Dominik-Vogel: I would prefer a different name like `_representative_values` or `_mock_values`, because `_valid_values` implies that it might be a complete list. (Which of course does not make sense on second thought)

WilliamHPNielsen: Since it's a private attribute, it can have any old name. But of course it should have the name of the corresponding property, and I was also a bit in doubt about a good name. `valid_values` kind of breaks tab completion for `validate`, but I don't really see an obviously better alternative. `allowed_values`, `representative_values`, ...?

Note that it is actually a complete list for `Enum` validators.

jenshnielsen:
@WilliamHPNielsen I think this was merged during the simulated instruments pr and can close?

WilliamHPNielsen:
You are correct, this was included in #859.

PR #858, opened 2017-11-08
Better tests for stepping parameters

jenshnielsen:
* Use memory parameter and actually check the raw values in the step too 
* Move to the class where they belong. 

The 3 test cases could be unified


jenshnielsen:
@QCoDeS/core could anyone have a look at these?

jenshnielsen:
@WilliamHPNielsen could you review this

PR #859, opened 2017-11-09
Feature/simulated instruments

WilliamHPNielsen:
For testing purposes, it would be very useful to be able to instantiate drivers connecting to a simulated instrument rather than a real one. No simulation can (reasonably) be expected to perfectly match the behaviour of a real instrument, but this would still be good to have, as simply seeing whether the driver instantiates can be very helpful in the CI process. As an example things like #849 would be caught.

To test the driver code, it is important to leave that code untouched. We should therefore distinguish between a mock driver and a mock (simulated) instrument. Note that we have a single test that already does this (for AMI430), but with a home-made implementation of a mock instrument. This PR suggests using PyVISA-sim (https://pyvisa-sim.readthedocs.io/en/latest/index.html#) to create simulated instruments to communicate with.

PyVISA-sim is the "correct" solution in the sense that it is a part of PyVISA that we are already heavily using. Non-VISA instruments won't work, but we'll be able to write tests for some 80% - 90% of our drivers, which is a lot more than what we currently have. The simulated instruments are just `.yaml` files, so we may be able to auto-generate them from the QCoDeS parameters in some cases.

NB: The current state of this PR is still very much WIP. 

- Add new `sims` folder with a simulated instrument (`dummy.yaml`) and a dummy driver to communicate with it (`dummy.py`)
- Change the `VisaInstrument` class to detect that we are using the `sim` backend
- Add a simulated AMI430 instrument
- Add a dependency-injecting class `IPToVisa` to allow us to use PyVISA-sim with the AMI430 driver

Still pending:

- [x] Make the simulated AMI430 pass the test written for its driver
- [ ] Simulate ALL THE INSTRUMENTS! (deferred)
- [ ] Add tests for simulated instruments (deferred)
- [x] Document how to write and add a simulated instrument under the "How to write a driver" documentation
- [ ] Make a way to auto-generate a simulated instrument (deferred)
- [x] Get feedback from competent people (you!)

@QCoDeS/core


WilliamHPNielsen:
@QCoDeS/core, I think this is a good time for review. There is now a PyVISA-sim based test for the AMI430 that (as far as I can tell) carefully reproduces the existing test. This is, too me, proof-of-concept that we can indeed use PyVISA-sim for this. And the new test is 5 times faster! 🎉 

I know that this PR got kind of really messy, but just disregard all the validation stuff and my OCD-like pep8'ing. The relevant files are: `dummy.yaml`, `dummy.py`, `test_ami430_pyvisasim.py`, `AMI430.yaml`, and to some extent `AMI430_VISA.py`.

qcodes/tests/test_visa.py
```diff
@@ -134,7 +134,14 @@ def open_resource(self, address):
         self.assertEqual(rm_mock.call_args, ((),))
         self.assertEqual(address_opened[0], 'ASRL2')
 
+        # this one raises a warning
```
jenshnielsen: If it raises a warning we should catch that and perhaps also test that it indeed does raise

qcodes/instrument/visa.py
```diff
@@ -102,16 +129,19 @@ def set_terminator(self, terminator):
                 eg. '\r\n'.
         """
         self.visa_handle.read_termination = terminator
+        self.visa_handle.write_termination = terminator
```
jenshnielsen: Is this going to break existing drivers?

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -28,32 +30,42 @@ class AMI430(IPInstrument):
...
-
     mocker_class = MockAMI430
     default_current_ramp_limit = 0.06  # [A/s]
 
     def __init__(self, name, address=None, port=None, persistent_switch=True,
-                 reset=False, current_ramp_limit=None, terminator='\r\n', testing=False, **kwargs):
+                 reset=False, current_ramp_limit=None,
+                 terminator='\r\n', testing=False, **kwargs):
+
+        print('Debug: {}, {}'.format(address, port))
```
jenshnielsen: remove?

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -28,32 +30,42 @@ class AMI430(IPInstrument):
...
 
-        if None in [address, port] and not testing:
-            raise ValueError("The port and address values need to be given if not in testing mode")
+        # if None in [address, port] and not testing:
+        #    raise ValueError("The port and address values need to be"
+        #                     " given if not in testing mode")
```
jenshnielsen: Should this be removed?

qcodes/instrument/sims/AMI430.yaml
```diff
@@ -0,0 +1,106 @@
...
+devices:
+  device 1:
+    eom:
+      GPIB INSTR:
+        q: "\n"
+        r: "\n"
+    error: ERROR
+    dialogues:
+      - q: "*IDN?"
+        r: "QCoDeS, AMI430_simulation, 1337, 0.0.01"
```
Dominik-Vogel: We could think about adding an explicit method to determine whether an instrument is simulated while we keep the idn is identical to the non-simulated instrument. This is in any case necessary for drivers that read the idn to determine their model specific behaviour.

qcodes/instrument_drivers/american_magnetics/AMI430_VISA.py
```diff
@@ -0,0 +1,102 @@
...
+        ##################################################
+        # __init__ of VisaInstrument
+
+        self.add_parameter('timeout',
+                           get_cmd=self._get_visa_timeout,
+                           set_cmd=self._set_visa_timeout,
+                           unit='s',
+                           vals=vals.MultiType(vals.Numbers(min_value=0),
+                                               vals.Enum(None)))
+
```
jenshnielsen: Would it be possible to do this by explicitly calling the `__init__` of the super classes.  Not using super but with hard coding the names to avoid calling __init__ on ip instrument

qcodes/instrument/visa.py
```diff
@@ -50,46 +57,66 @@ def __init__(self, name, address=None, timeout=5, terminator='', device_clear=Tr
                            vals=vals.MultiType(vals.Numbers(min_value=0),
                                                vals.Enum(None)))
 
+        # backwards-compatibility
+        if address and '@' in address:
+            address, visa_library = address.split('@')
+            if visalib:
+                warnings.warn('You have specified the VISA library in two '
```
Dominik-Vogel: maybe even error instead of warning? In this case we have a high chance that the driver does not work as expected.

sohailc:
Awesome! I think its a great idea to make mock instruments.

qcodes/instrument_drivers/american_magnetics/AMI430.py
sohailc: Yeah that should be removed.

qcodes/instrument_drivers/american_magnetics/AMI430.py
sohailc: Yes remove this as well

qcodes/instrument_drivers/american_magnetics/AMI430_VISA.py
```diff
@@ -0,0 +1,102 @@
...
+import qcodes.utils.validators as vals
+from qcodes.instrument.ip import IPInstrument
+from qcodes.instrument.visa import VisaInstrument
+from qcodes.instrument_drivers.american_magnetics.AMI430 import AMI430
+from qcodes.utils.helpers import DelegateAttributes, strip_attrs, full_class
+
+
+class IPToVisa(VisaInstrument, IPInstrument):
+    """
```
sohailc: Awesome stuff!

qcodes/instrument/sims/AMI430.yaml
```diff
@@ -0,0 +1,106 @@
+spec: "1.0"
```
sohailc: How do we generate a simulated instrument from this yaml file? Maybe this file: https://pyvisa-sim.readthedocs.io/en/latest/ is helpful for people.

qcodes/instrument_drivers/american_magnetics/AMI430_VISA.py
```diff
@@ -0,0 +1,102 @@
+# if this actually works...
```
sohailc: I think the IPToVisa class should be placed in a testing utilities module. Then for every driver test that is not a VISA instrument, generate a VISA equivalent using this class, write a yaml file to generate a mock instrument and write your tests by instatiating a driver class.

qcodes/tests/drivers/test_ami430_pyvisasim.py
```diff
@@ -0,0 +1,413 @@
+import io
```
sohailc: Nicely done! Except, I am wondering if the test file should not be called test_ami430 without the "_pyvisasim". This is just the new testing module.

qcodes/tests/drivers/test_ami430_pyvisasim.py
```diff
@@ -0,0 +1,413 @@
...
+import pytest
+from hypothesis import given, settings
+from hypothesis.strategies import floats
+from hypothesis.strategies import tuples
+import logging
+from typing import List, Dict
+
+import qcodes.instrument.sims as sims
+from qcodes.instrument_drivers.american_magnetics.AMI430 import AMI430_3D
+from qcodes.instrument_drivers.american_magnetics.AMI430_VISA import AMI430_VISA
```
sohailc: I think we should make the AMI430_VISA class in the this testing module. The folder /qcodes/instrument_drivers/american_magnetics/ should be reserved for actual drivers only.

qcodes/instrument/sims/AMI430.yaml
WilliamHPNielsen: You can determine whether the instrument is sumlated by checking what VISA backend you are using. And you are absolutely right, we sometimes need to write several different simulated instruments per driver with different firmware and/or model numbers.

qcodes/instrument/visa.py
WilliamHPNielsen: Maybe? 😛 

I'll make sure it only happens for simulated instruments, then we can handle the general case later.

qcodes/instrument_drivers/american_magnetics/AMI430.py
WilliamHPNielsen: I don't know, maybe this should be removed?

qcodes/instrument_drivers/american_magnetics/AMI430.py
WilliamHPNielsen: I'll tear this out as well as all references to `testing` and the old `mocker_class` since this is no longer needed.

qcodes/instrument_drivers/american_magnetics/AMI430_VISA.py
WilliamHPNielsen: Ah... that's clever. I'll check it out.

qcodes/instrument_drivers/american_magnetics/AMI430_VISA.py
WilliamHPNielsen: You are absolutely right, there is nothing AMI430-specific to that class. I'll move it.

qcodes/tests/drivers/test_ami430_pyvisasim.py
WilliamHPNielsen: I was hoping you would say that! 😺

qcodes/tests/drivers/test_ami430_pyvisasim.py
WilliamHPNielsen: Fully agree. I'll move it.

qcodes/instrument_drivers/american_magnetics/AMI430_VISA.py
WilliamHPNielsen: That does indeed work! Thanks. `VisaInstrument.__init__` must still be explicitly reimplemented, but from `Instrument` and further up, `super` is no longer dangerous. See commit 0552837.

qcodes/tests/test_visa.py
WilliamHPNielsen: That is a good idea. I have added that now.

WilliamHPNielsen:
@QCoDeS/core Okay, save for documentation on how to write simulated instruments, all the above issues should be addressed.

qcodes/instrument/visa.py
WilliamHPNielsen: I would agree, but I actually don't think that anyone will ever trigger this. I have kept it this way to change the existing tests for the `VisaInstrument` as little as possible.

Dominik-Vogel:
I have an issue with the device clear function. I get a not implemented errror when I call the constructor of the instrument with the default option device_clear=True.
My instrument driver is: (Model_336.py)
```
class Model_336(VisaInstrument):
    def __init__(self, name, address, **kwargs):
        super().__init__(name, address, **kwargs)
        self.connect_message()
```
The instrument is: (lakeshore_model336.yaml)
```
spec: "1.0"
devices:
  device 1:
    eom:
      GPIB INSTR:
        q: "\n"
        r: "\n"
    error: ERROR
    dialogues:
      - q: "*IDN?"
        r: "QCoDeS, m0d3l, 336, 0.0.01"

resources:
  GPIB::9::INSTR:
    device: device 1
```
and the script is:
```
import qcodes.instrument.sims as sims

from qcodes.instrument_drivers.Lakeshore.Model_336 import Model_336
# path to the .yaml file containing the simulated instrument
visalib = sims.__file__.replace('__init__.py', 'lakeshore_model336.yaml@sim')

ls = Model_336('lakeshore', 'GPIB::9::65535::INSTR', terminator = '\n', visalib=visalib, device_clear=True)
```
with the traceback:
```

  File "<ipython-input-1-71ee8fc820f5>", line 1, in <module>
    runfile('C:/Users/a-dovoge/OneDrive - Microsoft/projects/2017-11-29 Lakeshore/test.py', wdir='C:/Users/a-dovoge/OneDrive - Microsoft/projects/2017-11-29 Lakeshore', post_mortem=True)

  File "C:\Users\a-dovoge\AppData\Local\Continuum\anaconda3\envs\qcodes\lib\site-packages\spyder\utils\site\sitecustomize.py", line 710, in runfile
    execfile(filename, namespace)

  File "C:\Users\a-dovoge\AppData\Local\Continuum\anaconda3\envs\qcodes\lib\site-packages\spyder\utils\site\sitecustomize.py", line 101, in execfile
    exec(compile(f.read(), filename, 'exec'), namespace)

  File "C:/Users/a-dovoge/OneDrive - Microsoft/projects/2017-11-29 Lakeshore/test.py", line 7, in <module>
    ls = Model_336('lakeshore', 'GPIB::9::65535::INSTR', terminator = '\n', visalib=visalib, device_clear=True)

  File "c:\users\a-dovoge\qcodes\qcodes\instrument_drivers\Lakeshore\Model_336.py", line 38, in __init__
    super().__init__(name, address, **kwargs)

  File "c:\users\a-dovoge\qcodes\qcodes\instrument\visa.py", line 81, in __init__
    self.device_clear()

  File "c:\users\a-dovoge\qcodes\qcodes\instrument\visa.py", line 122, in device_clear
    self.visa_handle.clear()

  File "C:\Users\a-dovoge\AppData\Local\Continuum\anaconda3\envs\qcodes\lib\site-packages\pyvisa\resources\resource.py", line 270, in clear
    self.visalib.clear(self.session)

  File "C:\Users\a-dovoge\AppData\Local\Continuum\anaconda3\envs\qcodes\lib\site-packages\pyvisa\highlevel.py", line 442, in clear
    raise NotImplementedError

NotImplementedError
```
I can look into it with more detail if you @WilliamHPNielsen don't know already what is happening

WilliamHPNielsen:
@Dominik-Vogel I just know what you already discovered; `device_clear=True` does not work with the `@sim` library. Since there is never any reason to clear a simulated instrument, which in the nature of things does not exist yet and thus has nothing to clear, I'd propose to either simply remember to always set `device_clear=False` for simulated instruments or to bake an `if visalib == 'sim'` into our `visa.py`. What do you prefer?

qcodes/tests/drivers/test_ami430.py
```diff
@@ -0,0 +1,419 @@
...
+@given(set_target=random_coordinates["spherical"])
+@settings(max_examples=10)
+def test_spherical_setpoints(current_driver, set_target):
+    """
+    Check that the individual x, y, z instruments are getting the set
+    points as intended. This test is very similar to the sanity test, but
+    adds in the FieldVector as well.
+    """
+
+    print(set_target)
```
jenshnielsen: remove?

jenshnielsen: remove

WilliamHPNielsen: Yes I think we should remove this

WilliamHPNielsen: Please remove this

WilliamHPNielsen:
Now there's even documentation. I'll let @Dominik-Vogel answer and then possibly apply the changes he wants. When that's done, I think we should merge this one.

sohailc:
So will we require all new drivers submitted to include tests? I think we should

jenshnielsen:
@sohailc IMHO yes at least for visa instruments

Dominik-Vogel:
@WilliamHPNielsen regarding the `device_clear` issue, I would definitely prefer adding a `if visalib == 'sim'` case, so that writing the tests become simpler.

Dominik-Vogel:
@sohailc I agree that it would be great having a test for every instrument driver. And possibly this could be something that we do as main developers. But requiring it generally, I think, is a little bit too strict. We don't want to loose contributors by requiring test. Test in general are to my experience nobody without a deeper background in programming is familiar with. Bear in mind, that Delft may be exceptional in that way that there are many highly experienced qcodes users.

WilliamHPNielsen:
@QCoDeS/core I think this is ready now.

PR #860, opened 2017-11-09
Ats9360 driver only

jenshnielsen:
This is the driver only part of #816 which is hopefully easier to review.

I still have to address @damazter points from the other pr before this is ready to review


jenshnielsen:
Comments from #816 

- [ ] TODO comment `isn't it the case that multiple boards are already supported?`
- [ ] sample rates external/internal
- [ ] comments about ctypes call annotation
- [ ] Generalised to more than 2 channels
- [ ] pep8
- [ ] Move ctypes annotation
- [ ] Handle buffer API
- [ ] Time
- [ ] Get raw vs get 
- [ ] Valid options for AUX_IN_AUXILIARY
- [ ] Missing label

Dominik-Vogel:
We should remember to include **scipy** in the `environment.yml` file. For your ATS9630 branch this is required.

jenshnielsen:
I am not so sure. I think it should be an optional dependency, and definitely not for this branch as it does not use Scipy at all

nulinspiratie:
@jenshnielsen Could you let know when it is ready for review? We're using the ATS9440, which has some quirks such as 4 channels instead of 2. It also uses ATS.py, and I'd like to double check if the updated file is still compatible

jenshnielsen:
@nulinspiratie @damazter @sohailc I think this is more or less ready for review. I will do one more pep8 pass over it before merging

@nulinspiratie the only major api break that I know about is the change to the signature of handle parameter which now needs to take another parameter.

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -864,11 +857,17 @@ def get_sample_rate(self):
         Returns:
             the number of samples (per channel) per second
         """
-        if self.sample_rate.get() == 'EXTERNAL_CLOCK':
+        if (self.clock_source.get() == 'EXTERNAL_CLOCK_10MHz_REF'
```
damazter: where did the decimation part of this function go?

damazter: furthermore, is th elogic here correct? when there is an external clock, then the sample rate is still determined by the internal clock setting I think

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -396,10 +431,26 @@ def config(self, clock_source=None, sample_rate=None, clock_edge=None,
...
+        self._set_if_present('aux_io_mode', aux_io_mode)
+        self._set_if_present('aux_io_param', aux_io_param)
         # endregion
 
+        # handle that external clock and internal clock uses
+        # two different ways of setting the sample rate.
+        # We use the matching one and mark the order one
+        # as up to date since it's not being pushed to
+        # the instrument at any time and is never used
+        if clock_source == 'EXTERNAL_CLOCK_10MHz_REF':
```
damazter: why do you not put a call to `self.get_sample_rate` here?

qcodes/instrument_drivers/AlazarTech/ATS.py
jenshnielsen: The decimation is handled below like its always been.

If clock_source == 'EXTERNAL_CLOCK_10MHz_REF' and there is a external_sample_rate parameter it will use that. Alternatively it will fall back to the internal sample rate if that parameter is not defined. That is to accommodate for the 2 different Alazar Cards. One has a freely settable sample rate with a 10 MHZ external clock and the other one only has one possible rate in that case.

If there is a pure external_clock there is still no automatic way to infer the sample rate.

qcodes/instrument_drivers/AlazarTech/ATS.py
jenshnielsen: Because that would include the decimation which we don't want here but I have changed the code to make the decimation correction optional instead.

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -408,9 +459,10 @@ def config(self, clock_source=None, sample_rate=None, clock_edge=None,
                            self.parameters['coupling' + str(i)],
```
nulinspiratie: handle should be binary, i.e. `2**(i-1)`. This should only matter for 4-channel ATS, like the ATS9440

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -408,9 +459,10 @@ def config(self, clock_source=None, sample_rate=None, clock_edge=None,
...
                            self.parameters['channel_range' + str(i)],
                            self.parameters['impedance' + str(i)])
-            self._call_dll('AlazarSetBWLimit',
-                           self._handle, i,
-                           self.parameters['bwlimit' + str(i)])
+            if bwlimit is not None:
+                self._call_dll('AlazarSetBWLimit',
+                               self._handle, i,
```
nulinspiratie: not sure if handle should be binary here as well, the ATS9440 doesn't have bw limit

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -574,83 +620,140 @@ def acquire(self, mode=None, samples_per_record=None,
...
-            # recycling for the last series of buffers
-
-            # if buffers must be recycled, extract data and repost them
-            # otherwise continue to next buffer
-
-            if buffer_recycling:
-                acquisition_controller.handle_buffer(buf.buffer)
+        try:
+            for buf in self.buffer_list:
                 self._call_dll('AlazarPostAsyncBuffer',
```
nulinspiratie: When performing long acquisitions, AlazarPostAsyncBuffer has sometimes return an ApiFailed error. I haven't figured out why yet. Anyone else experienced this as well? If so, does this PR fix this issue? Otherwise, would it make sense to add a try except statement here that frees the buffer memory?

PR #861, opened 2017-11-09
Feature/parameter logging

nulinspiratie:
- Add logging of parameter during set
- Each parameter has it's own log, whose name is the parameter's full name
- If a scale/parser/mapping is used, this is also added to the log note

Haven't added logging for a .get, not sure if this is needed...


jenshnielsen:
Looks good. I think for performance it would be good to ensure that all the string interpolation is not done if no logging is enabled. Can you do something similar to what is done here 

https://docs.python.org/3.6/howto/logging.html#optimization
and wrap everything in `if logger.isEnabledFor(logging.DEBUG):`

WilliamHPNielsen:
@nulinspiratie, and remember that we still officially support python 3.5, so nice and beautiful string formatting has to wait a bit before we allow it. But hopefully that will happen before too long.

nulinspiratie:
@jenshnielsen Added optimization check
@WilliamHPNielsen I keep forgetting it, I'm getting too used to them. I've converted them back

qcodes/instrument/parameter.py
```diff
@@ -402,6 +404,17 @@ def set_wrapper(value, **kwargs):
                                              not(step_index == len(steps)-1 or
                                                  len(steps) == 1)))
 
+                    if self._snapshot_value and self.log.isEnabledFor(logging.DEBUG):
+                        # Add to log
+                        log_msg = 'parameter set to {}'.format(val_step)
+                        if mapped_value != val_step:
```
WilliamHPNielsen: As Travis taught us, this will fail for multi-valued parameters.

WilliamHPNielsen: @sohailc Perhaps you, as the author of the AMI430 driver, would like to have a look at this? What do you think, is this the fault of the driver or the logging added?

nulinspiratie: @WilliamHPNielsen Sounds like the issue arises when comparing numpy arrays, as these cannot be compared with one another. A solution would be to use `numpy.array_equal`, although this fails when values are not the same to within machine precision. A better solution is probably `numpy.is_close`, as this lets you specify relative tolerances. In any case, I think we'll have to perform a check for numpy arrays

jenshnielsen: @nulinspiratie that is correct but IMHO parameters that are arrays or other non scalar types should be a subclass of array or multiparameter and not hit this code path at all

WilliamHPNielsen: @jenshnielsen, do you then think we should rewrite the AMI430 driver? @sohailc, what's your opinion?

And if so, is there any way that we can enforce that a `Parameter` is never used to hold more than one value. Like checking for `__iter__` or `__getitem__` or what do I know?

nulinspiratie: @jenshnielsen Do you mean they should not have the same get/set wrapper? I think we can get away with only doing a check that excludes Array and MultiParameter, either `isinstace(self, Parameter)`, or checking for attribute `shape/shapes` should do the trick

WilliamHPNielsen: @nulinspiratie I think that sounds sane, as we will need to handle settable multiparameters for the magnet controls. Although not directly related to this issue, we could perhaps also change the `step` setter to fail if being called from another class than `Parameter`. And also think about how to avoid ending up with `Parameters` that are not scalar.

PR #862, opened 2017-11-10
fix: rescale also works for a dataset that is not completed

nulinspiratie:
Currently, MatPlot's `rescale_axis` does not work if the data array contains nan's. Therefore, it doesn't work if you prematurely stop a loop.

- Use `np.nanmax` instead of `np.array.max`, which ignores nans


jenshnielsen:
I think this is all contained within #875 so closing this

PR #864, opened 2017-11-10
fix imports for M4i driver

peendebak:
As discussed on slack. @WilliamHPNielsen


peendebak:
@WilliamHPNielsen The tests import the driver which triggers importing a library which is not included in the qcodes driver. Can you have a look at this?

qcodes/instrument_drivers/Spectrum/M4i.py
```diff
@@ -11,15 +11,23 @@
...
 import ctypes as ct
 from functools import partial
 from qcodes.utils.validators import Enum, Numbers, Anything
 from qcodes.instrument.base import Instrument
+
+
 try:
+    # add the location of the pyspcm header file manually
+    header_dir = os.path.split(__file__)[0]
+    sys.path.append(header_dir)
```
WilliamHPNielsen: I think it's good practice to only append it if it's not already there.

WilliamHPNielsen:
Lemme see if I can make any sense of the importing...

qcodes/instrument_drivers/Spectrum/M4i.py
peendebak: Ok, done

jenshnielsen:
@peendebak The docs build needs to import every driver to build the api docs for it. Due to your changes it no longer throws an ImportError when failing to import pyscm but rather a OSError when not finding the shared object. I think all you need to do is catch the OSError along with the import error. 

The code in `h2py.py` is not python 3 compatible I am not sure why we carry this around as it does not seem to be used.

peendebak:
@jenshnielsen I removed the `h2py.py`. The `OSError` is now handled similar to the `ImportError`.

Issue #750, opened 2017-09-20
Fix dependency on json-schema


jenshnielsen:
Seems to be missing from setup.py

PR #867, opened 2017-11-13
Requirements cleanup

### LINKED ISSUE #863: missing dependency`jsonschema` in requirements.txt
The latest pip installable version of QCoDeS does not list jsonschema as a dependency but requires it. 

This problem came to light when running tests on a project that depeneds on QCoDeS. Part of my build script installs qcodes (pip install qcodes). Importing qcodes then raises an exception because jsonschema is missing.

jenshnielsen:
Fixes #863 #750 and bumps numpy and pyqt/sip to latest bug fix releases @QCoDeS/core

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
-numpy==1.13.1
+numpy==1.13.3
 matplotlib==2.0.2
 pyqtgraph==0.10.0
 PyVISA==1.8
-PyQt5==5.9
-sip==4.19.3
+PyQt5==5.9.1
+sip==4.19.5
 QtPy==1.3.1
 h5py==2.7.1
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,8 +1,9 @@
-numpy==1.13.1
+numpy==1.13.3
 matplotlib==2.0.2
 pyqtgraph==0.10.0
 PyVISA==1.8
-PyQt5==5.9
```
WilliamHPNielsen: Travis can't install this. Does that not mean that our CI will stop working?

jenshnielsen: I know its because that version is only out for win and OSx not linux. I have reverted this

PR #869, opened 2017-11-13
Decadac

ThorvaldLarsen:
- Changed number of slots to 5 (slot 6 doesn't exist).
- Changed voltage label to refer to channel number rather than 'Voltage'.


Dominik-Vogel:
Great you could make this PR so quickly! Looks perfect to me :-)

PR #870, opened 2017-11-14
Add simple demodulator signals to Zi UHFLI 

jenshnielsen:
@QCoDeS/core all in the title


qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -711,6 +710,22 @@ def __init__(self, name, device_ID, **kwargs):
...
+            self.add_parameter('demod{}_sample'.format(demod),
+                               label='Demod sample',
+                               get_cmd=partial(self._getter, 'demods',
+                                               demod - 1, 2, 'sample'),
+                               snapshot_value=False
+                               )
+
+            for demod_param in ['x', 'y', 'R', 'phi']:
+
+                self.add_parameter('demod{}_{}'.format(demod, demod_param),
```
WilliamHPNielsen: Ideally, this should have a unit too. In the notebook, units are shown for these parameters (with a lowercase "v" instead of "V'), but where are they assigned?

jenshnielsen: I must have forgotten to add the commit that does the units

jenshnielsen:
@WilliamHPNielsen  pushed the units commit now

PR #871, opened 2017-11-16
spi D5a driver:  validator has to be updated on span change 

dpfranke:


jenshnielsen:
Looks good to me but we don't have these modules. @eendebakpt Does this look ok to you as the original writer of the driver?

peendebak:
@jenshnielsen I added some docs, the PR looks fine to me.

PR #875, opened 2017-11-21
fix: MatPlot.rescale_axis fixes

nulinspiratie:
- `MatPlot.rescale_axis` now also works if multiple traces have been added to a subplot
- Ignore rescaling if trace is not a DataArray (e.g. a numpy array)


qcodes/plots/qcmatplotlib.py
```diff
@@ -14,6 +14,8 @@
 
 from .base import BasePlot
 import qcodes.config
+from qcodes import DataArray
```
jenshnielsen: Please use the fully qualified import `from qcodes.data.data_array import DataArray` otherwise you end up creating circular references that prevent Matplot from being imported into the top level qcodes namespace

nulinspiratie: @jenshnielsen fixed

PR #887, opened 2017-11-28
Rohde Schwarz ZNB /VNA refactor common code

Dominik-Vogel:
In this pull request the common code from the two sweep classes gets refactored out.
The changes have not been tested with the instrument yet.

Additionally the channel class is abstracted into a class variable in the same way as in the decadac #876 to enable extention.
I also added a feature to automatically select the trace for the sweep when get is being called and restore the previous state on finalization.


Dominik-Vogel:
This way of subclassing a parameter and overriding a method does not support the functionality of get_latest(). So we would need to add a _save_val/supply the get function via a kwarg or implement the warpper manually

Dominik-Vogel:
Okay I jsut learned, that BaseParameter actually does the wrapping....

jenshnielsen:
@Dominik-Vogel Looks like a nice cleanup. I would very much like this to be tested on an actual instrument before merging. @ThorvaldLarsen are you using the vna

jenshnielsen:
You might want to rename get -> get_raw to get rid of the deprecation warning

Dominik-Vogel:
Thanks for the tip with `get_raw`! Yes surly do I want to test it before merging it.

ThorvaldLarsen:
I can help with testing tomorrow afternoon if it is ready for testing by then.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -68,7 +68,7 @@ def __init__(self, name, instrument, start, stop, npts, channel):
                          label='{} magnitude'.format(
                              instrument._vna_parameter),
                          setpoint_units=('Hz',),
-                         setpoint_names=('{}_frequency'.format(instrument._vna_parameter),))
```
ThorvaldLarsen: This is a mistake. Should be {}_frequency with the '_'.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -173,6 +103,17 @@ def __init__(self, parent, name, channel):
                                                        self._tracename,
                                                        self._vna_parameter))
 
+        # source power is dependent on model, but not well documented.
+        # here we assume -60 dBm for ZNB20, the others are set,
+        # due to lack of knowledge, to -80 dBm as of before the edit
+        model = self._instrument.get_idn()['model'].split('-')[0]
```
ThorvaldLarsen: model = self._parent.get_idn()['model'].split('-')[0]

Dominik-Vogel: good point, thanks 👍

ThorvaldLarsen:
@Dominik-Vogel I just remembered a somewhat minor thing but now that we are redoing the driver we might as well add it.
The bandwidth can only be set in steps 1/1.5/2/3/5/7, i.e. 1kHz, 700Hz, 15Hz etc. That is 280Hz will be set to the nearest step in this case 300Hz.
Is it possible to make a parser that checks if the number is allowed and if not reports back the actual value being set on the VNA?

See page 4 in https://cdn.rohde-schwarz.com/pws/dl_downloads/dl_common_library/dl_brochures_and_datasheets/pdf_1/service_support_30/ZNB_dat-sw_en_5214-5384-22_v0900_96dp.pdf

WilliamHPNielsen:
@ThorvaldLarsen Sorry to jump in in the middle of a discussion, but I just caught a glimpse of something that sounds dangerous. Behaviour like
```python
my_inst.my_param(100)  # actually sets it to 120
```
is in my opinion very bad and should be avoided. Either you input a correct value and have that exact value (modulo rounding errors and noise) realised on the instrument or you get an error and *nothing* happens on the insturment. As you most likely know, this is handled by validators.

ThorvaldLarsen:
@WilliamHPNielsen I have no clue what the correct way to handle this is in Qcodes. I am just pointing out that the behaviour of the ZNB instrument when setting bandwidth is pretty unintuitive. If you try to set the bandwidth to 120 Hz it will silently round it off to nearest step being 100 Hz.

WilliamHPNielsen:
@ThorvaldLarsen I see. It is very good that you point this is out. Then the correct way is indeed to make any attempt of setting a value that can not be realised fail using an appropriate validator. We are paranoid about machines silently doing things behind our backs.

WilliamHPNielsen:
@ThorvaldLarsen If you need help writing a custom validator for this case, just ping me.

Dominik-Vogel:
@ThorvaldLarsen, @WilliamHPNielsen  I'll add the validator.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -42,11 +21,12 @@ def __init__(self, name, instrument, start, stop, npts, channel):
         self._channel = channel
         self.names = ('magnitude',
                       'phase')
-        self.labels = ('{} magnitude'.format(instrument._vna_parameter),
-                       '{} phase'.format(instrument._vna_parameter))
+        self.labels = ('{} magnitude'.format(instrument.short_name),
```
jenshnielsen: Can you remind me of name, full_name vs short name

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -127,43 +81,27 @@ def set_sweep(self, start, stop, npts):
...
-                        "values. Will discard the imaginary part.")
-            data = data[0::2] + 1j * data[1::2]
-        self._instrument._parent.cont_meas_on()
-        return data
-
+    def get_raw(self):
+       data = self._instrument._get_sweep_data()
+       if self._instrument.format() in ['Polar', 'Complex',
+                            'Smith', 'Inverse Smith']:
+           log.warning("QCoDeS Dataset does not currently support Complex "
```
jenshnielsen: Perhaps this warning should point to the other parameter

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -127,43 +81,27 @@ def set_sweep(self, start, stop, npts):
...
+                       "values. Will discard the imaginary part.")
+       return data
 
 class ZNBChannel(InstrumentChannel):
 
-    def __init__(self, parent, name, channel):
+    def __init__(self, parent, name, channel, vna_parameter: str=None):
+        """
+        Args:
+            vna_parameter: if left empty name is used
```
jenshnielsen: I would write a bit more here.
```
Args:
    parent: Instrument that this channel is bound to.
    name: Name to use for this parameter
    vna_parameter: Name of parameter on the vna that this should measure such as S11. If left
               empty this will fall back to `name` 
```

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
ThorvaldLarsen: full name is 'Instrument name'_'Channel name' and short_name is 'Channel name'.
In my case full name is VNA_S21 for and short_name is S21.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
ThorvaldLarsen: full name is 'Instrument name'_'Channel name' and short_name is 'Channel name'.
In my case full name is VNA_S21 for and short_name is S21.

jenshnielsen: Ok thats cool

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
Dominik-Vogel: good point

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
Dominik-Vogel: @ThorvaldLarsen could you write this? I think you can find more concise words than me ;-)

ThorvaldLarsen: Might just want to change a word or two for clarity. But the part that Jens wrote looks pretty good to me.
Args:
    parent: Instrument that this channel is bound to.
    name: Name to use for this channel.
    vna_parameter: Name of parameter on the vna that this should measure such as S12. If left
               empty this will fall back to `name`.

Dominik-Vogel: done. Ready to merge?

PR #888, opened 2017-11-28
add update for D5mux

peendebak:
The D5mux module is a variation of the D5a module. It has 8 normal DACs (to be controlled with this driver) and some other specialized output channels (not addressed in this driver). The PR is backwards compatible.

@layeoh @mtiggelman @jenshnielsen


layeoh:
I've checked the changes with the modified control module for the cryomultiplexer - it works fine.
Thanks again!

PR #891, opened 2017-11-30
Minicircuits RF switch USB-4SPDT-A18

ThorvaldLarsen:
Driver for Minicircuits USB-4SPDT-A18 RF switch box. Possibly easily adaptable to other models of USB switch boxes from Minicircuits.

Requires the Pythonnet package to load the .dll file. Also every time I have moved to a new computer I have to manually unblock the .dll file in Windows to load the file into python.

@Dominik-Vogel I made a few changes to the original version fixing some bugs with the use of addresses.


nataliejpg:
works fine, just IDN doesn't return stuff, is it possible to remove the function? Is there a guideline for this @Dominik-Vogel ? Seems weird to be able to IDN and get no information

nataliejpg:
also needs an __init__.py file @ThorvaldLarsen

Dominik-Vogel:
I'll make the driver conform with #814, which is the ethernet version of the same device. I'll also try to take care of the IDN here.

Dominik-Vogel:
There is a channelized version of the driver now. There are probably still bugs in it as I could not test it here. If someone would like to test it, great, otherwise I'll do it once I get some time at a free station.

Dominik-Vogel:
I also had to remove the .dll because the licensing is not clear to me. We cannot put any license restricted  code in the repository.

ThorvaldLarsen:
The dll file can be downloaded from here (the .NET dll):
https://ww2.minicircuits.com/softwaredownload/rfswitchcontroller.html
It doesn't say anything about a license for the dll - also not in the Readme file that is downloaded with the dll file.

Not sure if that matters or if we still want to have people get the dll themselves.

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
```diff
@@ -0,0 +1,88 @@
...
+
+        for c in _chanlist:
+            channel = SwitchChannelUSB(self, 'channel_{}'.format(c), c)
+            channels.append(channel)
+            self.add_submodule('channel_{}'.format(c), channel)
+        channels.lock()
+        self.add_submodule('channels', channels)
+
+    def get_idn(self):
+        fw = self.switch.GetFirmware()
```
ThorvaldLarsen: Both commands exists in the dll but throws an error. I think they need some kind of argument but not sure...
Read_ModelName()
Read_SN()

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
```diff
@@ -0,0 +1,88 @@
+from qcodes import Instrument
+from qcodes.utils import validators as vals
+from qcodes.instrument.channel import InstrumentChannel, ChannelList
+
```
ThorvaldLarsen: we need to import the clr (Pythonnet). Would be good to keep the import error message also.

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
```diff
@@ -0,0 +1,88 @@
...
+                           get_cmd=self._get_switch,
+                           vals=vals.Ints(1, 2)
+                           )
+
+    def _set_switch(self, switch): 
+        self._parent.switch.Set_Switch(self.channel_letter,switch-1)
+
+    def _get_switch(self):
+        status = self._parent.switch.GetSwitchesStatus(self._parent.address)[1]
+        return int("{0:04b}".format(status[-1-self.channel_number]))+1
```
ThorvaldLarsen: Status should be cast to binary form before picking the number connected to the channel (moving one parenthesis):
int("{0:04b}".format(status)[-1-self.channel_number])+1

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
ThorvaldLarsen: Just figured out the right syntax. The address has to passed as a string:
Read_ModelName(str(self.address))
Read_SN(str(self.address))

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
```diff
@@ -0,0 +1,88 @@
...
+                               snapshotable=False)
+
+        _chanlist = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h']
+        _max_channel_number = int(self.IDN()['model'][3])
+        _chanlist = _chanlist[0:_max_channel_number]
+
+        for c in _chanlist:
+            channel = SwitchChannelUSB(self, 'channel_{}'.format(c), c)
+            channels.append(channel)
+            self.add_submodule('channel_{}'.format(c), channel)
```
ThorvaldLarsen: Can we add a line adding self.portA = channel[0].switch?
I don't like that the letters are completely hidden in the driver as those are what is shown on the box.

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
```diff
@@ -0,0 +1,88 @@
...
+            channel_letter (str): channel letter ['a', 'b', 'c' or 'd'])
+        """
+
+        super().__init__(parent, name)
+        self.channel_letter = channel_letter.upper()
+        _chanlist = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h']
+        self.channel_number = _chanlist.index(channel_letter)
+
+        self.add_parameter('switch',
+                           label='switch',
```
ThorvaldLarsen: can we change label to 'switch {}'.self.channel_letter?
This will make the metadata a lot more readable.

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
Dominik-Vogel: that is a good point, I totally forgot the import.

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
Dominik-Vogel: that must have been a typo...

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
Dominik-Vogel: is the idn now working with these modifications?

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
Dominik-Vogel: From my understanding what you can do now is:
```
switch = USB_SPDT('myswitch')
switch.channel_a.switch(1)
```
I think we could add a convenience function that does
```
switch('a',1)
```

Dominik-Vogel: If this is what you meant...

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
Dominik-Vogel: I think that is a good Idea

Dominik-Vogel:
@ThorvaldLarsen could you commit your cahnges to the PR? I just got a request from @johnhornibrook  in Sydney that they need the same driver. If you say that it works with your proposed changes we can merge it so that it will be available for everybody.

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
```diff
@@ -1,7 +1,12 @@
 from qcodes import Instrument
 from qcodes.utils import validators as vals
 from qcodes.instrument.channel import InstrumentChannel, ChannelList
-
+try:
+    import clr
+except ImportError:
+    raise ImportError("""Module clr not found. Please obtain it by
+                         running 'pip install -i https://pypi.anaconda.org/pythonnet/simple pythonnet'
```
jenshnielsen: why this extra index. The package seems to be on pypi?

jenshnielsen: And if this is only going to work on windows we probably should check for that first

ThorvaldLarsen: extra index? I am not too familiar with installing python packages so I have just put what worked for me.

ThorvaldLarsen: I dont know if this is a windows only package...

jenshnielsen: ` -i https://pypi.anaconda.org/pythonnet/simple` means instead of looking in the normal python packages index go and look in this specific location. That is very unusual and I would not recommend doing anything like that unless you absolutely have too. The package is here https://pypi.python.org/pypi/pythonnet/2.3.0 so doing `pip install pythonnet` should work too. 

The dll is probably windows only so this is unlikely to work on any other platform

ThorvaldLarsen: thanks. that does indeed seem nicer if it works.

ThorvaldLarsen:
I have tested this driver and it works with a couple of corrections just commited. A couple of requests for the driver:
- a way to call switch.a(val) instead of switch.channel_a.switch(val)
- a function that sets all channels to a specific value i.e. switch.set_all(val).

jenshnielsen:
@ThorvaldLarsen This fails to build the documentation because `clr` is not installed on travis. To solve this we usually mock (as in fake the install of) when building the docs. Can you append `clr` to `autodoc_mock_imports` towards the end of https://github.com/QCoDeS/Qcodes/blob/master/docs/conf.py as done for `pyspcm` and others

Dominik-Vogel:
@ThorvaldLarsen have you found that string argument you are passing to the dll calls in a documentation? Do you know what it is supposed to be? I am just wondering if, by passing an empty string, we are actually using a special case here, or if maybe one needs to specify each time, which of the switches (devices) one wants to control....

ThorvaldLarsen:
@Dominik-Vogel I first tried with the address in string format but turns out that any string can be passed and you get the same response. I haven't looked at documentation for it.

jenshnielsen:
@ThorvaldLarsen could you move the Windows check into the `__init__` function of the driver. That way the file can be imported when the documentation is build but we still get a meaningful error message when trying to run the driver on a non windows platform

Dominik-Vogel:
I made some changes to the driver:
 * Abstract base, so that the common functions of this one and the RC/ethernet one only need to be implemented in one place
 * added channel aliases and made them callable: `device.a(2)` setting line a to input 2
 * changed the initialization to the serial number as identification as it can be read from the label on the switch, instead of the address which needs to be read out
I have tested all changes successfully and think it is good to be merged to move the RC version on top of the same base in another PR. Feel free to double check.

PR #897, opened 2017-12-04
Driver/Mercury iPS with VISA

### LINKED ISSUE #543: Sweeping magnet in spherical coordinates can fail
I have an issue with a 3-axis magnet (a driver not online yet, but basically using the same spherical to cartesian conversion as the Mercury IPS)

When i sweep a value in the spherical coordinates, lets say the radius, the driver will use \theta and \phi values from cache, and calculate the xyz field. That works fine, unless the t and p values are updated. That happens i.e. when the user actually reads the spherical coordinates, if a `magnet.rtp.get()` is called at every sweep point, the internal r,t,p values will change eventually (right away, or after going through zero, or...).

The issue is basically that we have many representations of one xyz point in the spherical system. And I have no idea how to ensure that the other two variables will not change during a given sweep.

The only way to make this not happen (that i can come up with) would be to just not ask for spherical values during a sweep.

Anyone with any ideas?
@core @AdriaanRol @peendebak <- just tagging because of some reason...

WilliamHPNielsen:
Add a VISA driver for the MercuryiPS using the `FieldVector` setpoint technology already in use in our AMI430 driver. 

Fixes #543

- Rewrite the MercuryiPS driver to use VISA
- Make the driver use the FieldVector to deal with singularities

Work items:

- [x] Add all relevant parameters
- [x] Use `FieldVector`
- [x] (once #859 is merged) Add a simulated instrument and a thorough test

@jenshnielsen


WilliamHPNielsen:
Note: development of this driver is paused until an agreement on the desired magnet ramping behaviour has been reached.

WilliamHPNielsen:
From talking to scientists, it seems that 
  * Using the "first-down-then-up" approach of the AMI430 driver
  * Adding an optional `ramp_resolution` parameter for staircasing non-cartesian ramps
will suit the needs of everyone. 

Development resumed!

qcodes/tests/drivers/test_MercuryiPS.py
```diff
@@ -53,6 +56,37 @@ def cylindrical_limits(x, y, z):
...
+logger = logging.getLogger('qcodes.instrument.visa')
+logger.setLevel(logging.DEBUG)
+formatter = logging.Formatter('%(created)s - %(message)s')
+lh = logging.StreamHandler(iostream)
+logger.addHandler(lh)
+lh.setLevel(logging.DEBUG)
+lh.setFormatter(formatter)
+
+
+def get_messages(iostream: io.StringIO) -> List[str]:
```
jenshnielsen: Could we use pytests capture log for this i.e. https://docs.pytest.org/en/latest/logging.html#caplog-fixture ?

WilliamHPNielsen: LOL! ^^

WilliamHPNielsen:
There are many extra features we can think of adding (one example: rotated coordinate systems to account for sample misalignment), but I think the driver is now in a functional state.

@QCoDeS/core

qcodes/instrument/sims/MercuryiPS.yaml
```diff
@@ -10,6 +10,81 @@ devices:
       - q: "*IDN?"
         r: "IDN:OXFORD INSTRUMENTS:SIMULATED MERCURY iPS:000:000"
 
+    properties:
+
+      x field target:
+        default: 0
+        getter:
+          q: "READ:DEV:GRPX:PSU:SIG:FSET"
+          r: "{}"  # should there be a unit appended? How does the instrument do it?
```
astafan8: is the answer to this question known now?

qcodes/instrument_drivers/oxford/MercuryiPS_VISA.py
```diff
@@ -190,14 +243,73 @@ def __init__(self, name: str, address: str, visalib=None,
         # to ensure a correct snapshot, we must wrap the get function
         self.IDN.get = self.IDN._wrap_get(self._idn_getter)
 
-        # TODO: Query instrument to ensure which PSUs we have
+        # TODO: Query instrument to ensure which PSUs are actually present
```
astafan8: is this an "issue" or "nice-to-have"? in other words, will it be done in this PR? If not, perhaps, make an issue out of it?

qcodes/instrument_drivers/oxford/MercuryiPS_VISA.py
```diff
@@ -190,14 +243,73 @@ def __init__(self, name: str, address: str, visalib=None,
...
             self.add_submodule(psu_name, psu)
 
+        self._field_limits = (field_limits if field_limits else
+                              lambda x, y, z: True)
+
+        self._target_vector = FieldVector(x=self.GRPX.field(),
+                                          y=self.GRPY.field(),
+                                          z=self.GRPZ.field())
+
+        for coord in ['x', 'y', 'z', 'r', 'theta', 'phi', 'rho']:
```
astafan8: can this list be referred to from `FieldVector` class or similar?

WilliamHPNielsen:
Status: I am currently waiting for feedback from a scientist who kindly volunteered to try out this new driver.

WilliamHPNielsen:
@QCoDeS/core, this driver version has now been tested at a *real* lab station and allegedly worked fine and solved two problems with the older driver (the problem described in #543 and a problem with wrong metadata, both basically solved by using `FieldVector`).

So I'd say we merge and actually also deprecate the older driver.

WilliamHPNielsen:
I'll merge this now, but not deprecate the old driver just yet. The reason is that someone is using the old driver with two axes being MercuryPS Slaves and the third one being a less noisy power supply. I will add that functionality in a separate PR and then we can finally get rid of the old, untrusted IPInstrument.

PR #906, opened 2017-12-07
Correct dac validators

dpfranke:
use the correct validator ranges (-4 to 4 and -2 to 2, cf. d5a driver) and adjust naming

@peendebak @jenshnielsen


WilliamHPNielsen:
Neither @jenshnielsen nor I can judge whether this is correct. Code wise it's fine (obviously), but I think we need @peendebak to agree that the the instrument indeed does work this way.

peendebak:
@WilliamHPNielsen PR is fine with me

PR #907, opened 2017-12-07
driver/ RTO-1000 HDEF bandwidth issue

### LINKED ISSUE #856: RTO1024/RTO1000 scope driver issues
I have found several problems with the RTO scope driver. 

1. The when using the HD mode on an RTO1024 the bandwidht get's set to 1MHz and cannot be changed from there. 
This would not be problematic if this mode was not needed but it is also used to set the dataformat which uses the same parameter. 

Observe the following difference and not the (spurious) staircase due to digitization in the left figure. 

image

Both are the same trace measured. The one on the right is measured by a little hack in the driver. 
The second trace contains a scaling error on the y-axis but I'm sure that can be fixed quite easilty. 
image

I added a screenshot from the scope itself as reference. 
image

2. Recent QCoDeS updates make this driver give a warning upon instantiation. IMO this should be fixed. 
image

WilliamHPNielsen:
Fixes #856

- Add a parameter controlling the high def mode bandwidth
- Document the use of this in the example notebook
- Add a simulated instrument and an initialisation test

@jenshnielsen @AdriaanRol


AdriaanRol:
@WilliamHPNielsen this does indeed solve the unusability of the HD mode as we can now use full the proper bandwidth. 

It does not address all of issue #856 as the resolution (when not using HD mode) is still quantized at a too low bit value. 

The following snippet is my current workaround that also makes the non-HD mode usable. 

This is tested and working on our setup (rto 1024 scope). I expect the rto 1044 to behave in the same way. 
```python 
        # we must have this nesting, since the parameter high_definition_state
        # only exists if HD
        # if self.channel._parent.HD:
        #     if self.channel._parent.high_definition_state() == 'ON':
        #         dataformat = 'INT,16'
        #     else:
        #         dataformat = 'INT,8'
        # else:
        #     dataformat = 'INT,8'

        # Hack to always use the high definition data transfer
        dataformat = 'INT,16'
```

AdriaanRol:
Other than that :+1:

WilliamHPNielsen:
@AdriaanRol, I asked you about this over in #856, but let's try again 🙂 

Do you know what the internally used resolution is when the instrument is not in high definition mode? When writing the driver, I assumed it to be 8 bit, but you seem to suggest that it is higher than that. I'll also search the manual. We can of course use 16 bit ints for data transfer always, but it's sort of pointless if the data is really only 8 bit.

AdriaanRol:
@WilliamHPNielsen sorry for the confusion. 
I can not directly answer the question as I was unable to find this in the manual. However, by reading out the data assuming it is 16 bits, the blockyness of the data (as shown in the screenshots) goes away without switching to HD mode (as can also be seen by looking at the scope display). 

As such my experimental observation suggests that even when not using HD mode the data is transferred in more than 8 bits. I can not exclude that it is lower than 16 bits but the protocol at least seems consistent with that. Note that I also get a quantitative agreement between the data I read out and the data I see on the display of the scope when using this workaround. 

Does this help?

WilliamHPNielsen:
@AdriaanRol no problem. Below is the part of the manual I (think I) based my assumption on (page 460 in my version, 1332.9725.02 ─ 10, look under "Data and File Management"):

screen shot 2017-12-07 at 17 02 44

...but it's quite ambiguous, at least to me. Can you do an extra quick test to convince me: count the number of unique values in your trace acquired using non-high definition mode but exported as 16 bit integers. I agree that you basically already proved that it's more than 8 bit, but now I'm curious. 

In any case, l'll upgrade the driver to always use 16 bit integers for data export.

AdriaanRol:
@WilliamHPNielsen I don't think I fully understand the test you want me to do. My signal does not cover the entire range of possible values of the scope so I guess counting unique values would be pointless. Do you instead want me to look at the smallest difference in values or something like that?

AdriaanRol:
@WilliamHPNielsen I just ran into another bug that should be quite trivial to fix. 


On lines 432-327 of the driver the number of averages can be set. However the get_cmd does not work. 
This is because of an extra space between `ACQuire:COUNt-?` I have added a hyphen to denote where the space is now. By removing this space the asking for the number of acquisitions actually works. 

```python 
        self.add_parameter('num_acquisitions',
                           label='Number of single acquisitions',
                           get_cmd='ACQuire:COUNt ?',
                           set_cmd='ACQuire:COUNt {}',
                           vals=vals.Ints(1, 16777215),
                           get_parser=int)
```

WilliamHPNielsen:
@AdriaanRol, it sounds like you totally understood the test 🙂  It was just in case you had a recorded signal covering the entire range "lying around", you could quickly count unique values. But I guess the minimal difference should tell us the same thing. I will in any case return to this PR this afternoon and then also play a bit with trying to figure out the bit depth.

Re bug 2: thanks for catching, I'll make sure to delete the space.

AdriaanRol:
@WilliamHPNielsen 

We ran into quite some problems with the "arithmethics" of the RTO acquistion mode not working correctly if a single run has not been completed. The snippet below implements checking for progress and should resolve this issue. 

I think this should be included in the `.trace()` method that acquires a trace with potentially an option to disable this if fast acquisiton is required. 

```python 
    rto.ch2.arithmetics('AVERAGE')  # other options: 'ENVELOPE' and 'OFF'
    rto.ch2.trace.prepare_trace()
    rto.run_single()  # and perform a single run.

    # This snippet should be included in the get_trace method
    t0 = time.time()
    timeout = 60
    num_avg = int(rto.ask('ACQ:COUNT?'))
    curr_acq =0
    while curr_acq < num_avg:
        curr_acq = int(rto.ask('ACQ:CURR?'))
        t1 = time.time()
        dt = t1-t0
        # prints the progress in a friendly manner :)
        print('\rAt {}/{} averages, acquiring for {:.2f}s'.format(
            curr_acq, num_avg, dt), end='')
        if dt> timeout:
            raise TimeoutError('More than {}s expired'.format(timeout))

    y = rto.ch2.trace()
```

Issue #908, opened 2017-12-08
fix AMI430_3D driver: Parameters should not return arrays 


sohailc:
The AMI430_3D driver has a number of parameters which return arrays, e.g.: 

"cartesian_measured", "cartesian", "spherical_measured", etc. 

As these are regular parameters instead of ArrayParameters, this is wrong. 

The proper fix is to make these parameters, ArrayParameters.

sohailc:
ArrayParameter does not have a set attribute, which means we cannot simply use the ArrayParameter class. There are three solutions: 

1) Use the normal Parameter class but store the value as a single string. 
2) Don't use parameters at all for saving/accessing Cartesian/spherical/cylindrical set point but use class methods. The down side of that is that we need separate get/set functions for  cartesian/spherical/cylindrical. 
3) Add a set method on the ArrayParameter class. This is by far the best way, but seeing that this has not been done yet, makes me suspect that this isn't trivial.

WilliamHPNielsen:
@sohailc As discussed on slack, you will go with solution 3., right?

sohailc:
Yes, I will go with solution 3

PR #910, opened 2017-12-08
Ami430 no arrayed params

sohailc:
Fixes https://github.com/QCoDeS/Qcodes/issues/908

I still  have some things to do: 
1) Proper doc strings 
2) type hints

But perhaps you guys can already give comments.

### DIFF
--- a/qcodes/instrument_drivers/american_magnetics/AMI430.py
+++ b/qcodes/instrument_drivers/american_magnetics/AMI430.py
@@ -2,14 +2,51 @@ import collections
 import logging
 import time
 from functools import partial
+import warnings
 
 import numpy as np
 
 from qcodes import Instrument, IPInstrument
 from qcodes.math.field_vector import FieldVector
-from qcodes.utils.validators import Numbers, Anything
+from qcodes.utils.validators import Numbers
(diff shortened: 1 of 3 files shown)


qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -185,14 +233,23 @@ def _can_start_ramping(self):
 
         return False
 
-    def set_field(self, value, *, perform_safety_check=True):
+    def set_field(self, value: float, perform_safety_check: bool=True) ->None:
```
jenshnielsen: Did you intentionally remove the * that makes the following parameters keyword only?

jenshnielsen:
Travis fails due to RST issues in the documentation build

qcodes/instrument_drivers/american_magnetics/AMI430.py
sohailc: Yeah I did, as I do not get the point of the practice.

jenshnielsen: Here it's imho not that important but it makes sense for functions with many arguments and especially more than one non-keyword argument

PR #912, opened 2017-12-11
Hypothesis timings

jenshnielsen:
@WilliamHPNielsen Seems like 1000 ms is not always enough on travis so increase it here.


jenshnielsen:
It failed here https://travis-ci.org/QCoDeS/Qcodes/jobs/314756807 on master

jenshnielsen:
Travis is really slow for some reason. The test is apparently 10 times slower on Travis than when I run it locally

WilliamHPNielsen:
Should we just disable deadlines for that test then?

jenshnielsen:
Maybe. Now another one was slow. I am a bit worried that something funny is going on but perhaps its just sharing the travis service with other users

PR #913, opened 2017-12-12
Use Codecov for coverage measurement

jenshnielsen:
because codeacy doesnt work for prs


jenshnielsen:
Yes that is a divide by zero because there is no coverage of master. We might want to tweak the report/failure  levels down the line

PR #916, opened 2017-12-14
remove f-strings

WilliamHPNielsen:
@nulinspiratie, how many times do I have to tell you this?! 😛 

Merging PR #723 broke everything. Travis is not using IPython, so he didn't notice. Perhaps we should have a dedicated IPython test?

- Remove python 3.6 f-strings from `magic.py` 

@QCoDeS/core

### DIFF
--- a/qcodes/utils/magic.py
+++ b/qcodes/utils/magic.py
@@ -4,6 +4,7 @@ from IPython.core.magic import Magics, magics_class, line_cell_magic
 if sys.version_info < (3, 6):
     raise RuntimeError('Magic only supported for Python version 3.6 and up')
 
+
 @magics_class
 class QCoDeSMagic(Magics):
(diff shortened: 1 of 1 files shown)


jenshnielsen:
Looks good. An alternative would be to move the check for version number into the conditional in init so that code is not loaded for python 3.5

nulinspiratie:
@WilliamHPNielsen @jenshnielsen Ah man, sorry about that. I thought I removed it all. I'm not sure why Travis didn't complain though...

WilliamHPNielsen:
@nulinspiratie It's fine, it's a good reminder for us to move forwards in terms of python versions. I also wanna use f-strings!

I think Travis didn't catch it since he doesn't run ipython and the `magic.py` only gets imported if `get_ipython()` doesn't raise an exception.

PR #918, opened 2017-12-14
docs/Update or kill obsolete documentation

### LINKED ISSUE #839: Update documentation examples
Currently there are still a bunch of examples around that uses the old toymodel.py and should be updated. At least 

* docs/index.rst
* Qcodes location-format example.ipynb
* Metadata.ipynb

These should be rewritten

WilliamHPNielsen:
Fixes #839

- Update the old and broken documentation

Documentation completed:

- [x] Metadata.ipynb
- [x] Metadata with instruments.ipynb
- [x] Qcodes location-format example.ipynb
- [x] index.rst 

@QCoDeS/core


WilliamHPNielsen:
So I had some problems with the docs build. In markdown, you can have a line break by adding two whitespaces, so that this: `Line␣one␣␣Line␣two` turns into 
```
Line one
Line two
```
which I used in a notebook, but that apparently leads to invalid .rst files. Shooks.

WilliamHPNielsen:
This one is ready for review now.

jenshnielsen:
@WilliamHPNielsen The annoying thing is that travis runs Ubuntu 14.04 which has a quite old version of Pandoc with lots of bugs that are fixed in newer versions (as the one that you are likely to have from Homebrew or any other source of install on your own computer)

WilliamHPNielsen:
@jenshnielsen I see. I did indeed have some problems reproducing the results locally. That's also why I "reported" the problem here, just in case anyone else runs into it.

PR #919, opened 2017-12-14
change delay to inter_delay

eendebakpt:


jenshnielsen:
@WilliamHPNielsen Initially delay was removed as part of the parameter refactor but we added it back as a deprecated option. So I think this is fine

PR #923, opened 2017-12-16
Dont store a max lenght string

### LINKED ISSUE #921: internal data of String validator breaks AWG5014
In PR #859 new variables `._valid_values` are added to the `String` validator. The default validator object `qcodes.utils.validators.String()` now has a size of 1 GB.

The `AWG5014` driver uses the `String` validator four times, so the driver now is 4GB. If one reloads the driver 8 GB is occupied, which crashes a system with limited amount of RAM.

https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/tektronix/AWG5014.py#L326

I am not sure how the `._valid_values` field is used, but I think it does not make a lot of sens for a `String`. Can we set the `_valid_values` field to `None` as a solution?

@WilliamHPNielsen @AdriaanRol @jenshnielsen @lucblom

jenshnielsen:
Fixes #921 There is no good reason to store a string of the length of the max string

### DIFF
--- a/qcodes/utils/validators.py
+++ b/qcodes/utils/validators.py
@@ -147,7 +147,7 @@ class Strings(Validator):
         else:
             raise TypeError('max_length must be a positive integer '
                             'no smaller than min_length')
-        self._valid_values = ['.'*min_length, '.'*max_length]
+        self._valid_values = ['.'*min_length]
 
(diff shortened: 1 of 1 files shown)


jenshnielsen:
@WilliamHPNielsen

PR #824, opened 2017-10-27
Run mypy type checking on the codebase with Travis

jenshnielsen:

PR #927, opened 2017-12-18
ZIUHFLI: add missing import

jenshnielsen:
@WilliamHPNielsen caught by a new version of mypy in #824


jenshnielsen:
Yes but you are probably giving me credit for being more clever than i am. In this case it is only used for sqrt(2) math seems like the obvious choice

Issue #773, opened 2017-09-30
Node-tree based parameter structure in instruments 


AdriaanRol:
@jenshnielsen @WilliamHPNielsen 
As discussed with @alan-geller and @sohailc . 
Any work on this requires #600 to be merged (@nulinspiratie) . 
Can be seen as a generalization of channel parameters (#640 #641 ) @spauka . 
Relevant for future DCL QuTech drivers @nyxus . 

I would like to propose supporting a node/tree based structure for parameters in an instrument. 
This is motivated by the need to group certain parameters on a more generic level than channels. 

## Motivation
A concrete example would an instrument like the `AWG5014` that has a parameter `ch1_m1_high` (channel 1 marker 1 high).  It would be convenient to group parameters in a node structure. This would mean that instead of using underscores in the name of the parameter the instrument would have nodes with attributes, so `ch1_m1_high` would become `ch1.m1.high`. 

Such a grouping of parameters would have several advantages. 
- Autocomplete of large instruments would be more managable and would immediately show the relevant groups of parameters to a user. 
- Possible to natively match structure that exists in many instruments (see e.g., colon separated commands in VISA protocols and  the Zurich Instruments python API). 
- Possible to add documentation to `nodes` or groups of parameters. 
- Provides structure to datafiles (where snapshots are saved). 
- Provides structure that can be used in instrument monitoring capabilities. 
- I think this is simpler to understand than the Channel parameter (that does not work as per comment in #600 ). This argument requires reading the section below. 

## Proposed implementation/required changes
To make the node based parameters work changes are required to the following methods 
- `add_parameter`
- `snapshot`
- `print_readable_snapshot` 
- datasaving/loading -> especially urgent as that is currently being redesigned. 
- A new `parameter_node` object. 
- `Instrument.get` and `Instrument.set`

I will now go over these points individually. 

### add_parameter
The `Instrument.add_parameter` method would need to support adding parameters to nodes in a simple way. I would propose only changing the way the `name` argument of the `add_parameter` method works. 
The name would be split based on the period `.` character (as conventional in python), the last part of the name would be the name of the parameter. Anything before that corresponds to a node. 
The example of pseudo code below shows how it should work

```
ins.add_parameter(name='ch1.m1.high', **kwargs) 
 -> separate string based on period "." character

 -> check if ch1 exists as a parameter in the instrument. 
           -> if ch1 does not exist, create a "node" 
           -> if ch1 is a regular parameter raise exception 
[...]

Issue #767, opened 2017-09-28
Change parameters to be type-specific


nulinspiratie:
I've been using parameters a lot as object attributes, both for instrument drivers, and also for objects that are completely unrelated to physical instruments (meta-instruments).
Doing so has some clear advantages:
- Integration with QCoDeS loops (makes them sweepable and measurable),
- Allows a standard way to add attribute metadata (e.g. docstring, labels, validators, etc.),
- Integration with extra tools, such as a monitor.

However, one of the things that I've struggled with is when exactly to use a parameter, and when to use straightforward object attributes (`self.attr = some_value`). This becomes especially relevant for meta-instruments, as there not every parameter corresponds to an instrument instruction. The root of the problem seems to be that parameters don't really behave the same way as attributes. For example, getting/setting a parameter is done by calling it, instead of simply assigning a value. (`param(new_val)` instead of `param = new_val`). This leads to the annoying situation where I have to think for every attribute if it actually is a parameter or not. 

Ideally, we would want parameters to behave more like attributes, in which case we wouldn't have to worry about if an attribute is actually a parameter or not. any attribute relevant for QCoDeS could be a parameter (or even all attributes, although this might cause unnecessary overhead). I've been thinking about a way to tackle this problem, and my proposed solution is to have type-specific descriptor parameters. As an example, say we have a ListParameter:

```python
class ListParameter(list, Parameter)
    def __get__(self, instance, owner):
        return self
    def __set__(self, instance, val):
        self(val)

list_parameter = ListParameter()
```

By having the ListParameter inherit from both `list` and `Parameter`, we allow it to behave as a list, and get the additional functionality of a parameter. We can therefore treat it exactly like a list (e.g. `list_parameter.append(val)`, while it also has all the Parameter features (e.g. `list_parameter.unit`).

Furthermore, if we attach the ListParameter to an instrument (`instr.list_parameter`), the descriptor methods (`__get__` and `__set__`) allow simple getting/setting of values (e.g. `instr.list_parameter = new_val`). Note that this only works when the ListParameter is an attribute of another object, otherwise we would still need to use `list_parameter(new_val)`. This could also solve a few other issues. A MultiParameter would simply be a ListParameter, validators would usually be implicit, but could still be overridden for additional conditions such as being a multiple of a value. 

[...]

PR #935, opened 2017-12-20
Feature/parameter node, change parameter get/set behaviour

nulinspiratie:
This PR proposes to add the ParameterNode, as raised by @AdriaanRol in #773. My main goal was to have a different way of setting/getting parameters instead of using brackets (raised in #767). I've added a tutorial on how to use the ParameterNode in the PR.

- Add ParameterNode
- Change get/set behaviour of parameters inside a node
    The proposed new method is:
    - Set: `parameter_node.parameter = 42`
    - Get: `parameter_node.parameter` (returns 42)
  These commands set/get the value of the parameter, and the underlying parameter is accessed differently (see below for details).
    Note that the parameter get/set itself isn't changed, if you have a parameter object, you should still use brackets.
- Replace InstrumentBase with ParameterNode
- Add ParameterNode tutorial

I think the main advantage of the new get/set method is that one doesn't need to think if an attribute is actually a Parameter or not. This allows you to replace attributes with parameters as you see fit, while also being able to keep other attributes the way they are. Having parameters behave as attributes also feels more pythonic IMHO.

One potential issue with this new method is that any time a parameter attribute is accessed, this will cause a `.get()`, therefore, `hasattr(parameter_node, 'parameter')` will call its `.get` method. I'm not sure if this is an issue, as there aren't many cases where this happens.

Smaller changes
- Replace `print_readable_snapshot` with `print_snapshot`,  the snapshot being readable should be a given.
- add __dir__ of DotDict to include its keys
- Make name and Instrument of BaseParameter optional

I haven't implemented everything discussed in #773, I first wanted to get some thoughts on the new proposed method

Below is a summary of the ParameterNode, this is explained in more detail in the added tutorial notebook

## Creating a ParameterNode
Parameter nodes are instantiated via: `parameter_node = ParameterNode()`

## Adding parameters
parameters can be added by setting an attribute:
`parameter_node.new_parameter = Parameter()`
Note that it is no longer necessary to specify a name in the Parameter. If it is not explicitly specified, it will be set to the attribute name

## Getting/setting parameter values
Getting and setting of parameters in a parameter node is now done same as you would any other attribute:
- Setting a value: `parameter_node.new_parameter = 42`
- Getting a value: `parameter_node.new_parameter` (returns 42)

## Accessing parameter object
The actual parameter can be accessed in two ways: 
1. `parameter_node.parameters.new_parameter` (`parameter_node.parameters` is a `DotDict`).
2. `parameter_node().new_parameter`, the call `()` will return `parameter_node.parameters`. 

The second method is more succinct, but does have one drawback, namely that autocomplete won't work with it. I haven't figured out another way of easily accessing parameters that does have autocomplete, or a way to add autocompletion to the call method.

## Nesting nodes
Parameter nodes can also be nested, similarly to how parameters are added:
`parameter_node.nested_parameter_node = ParameterNode()`.
This sets the nested parameter node's name to the attribute.

@AdriaanRol @jenshnielsen @WilliamHPNielsen @giulioungaretti @alexcjohnson @alan-geller @spauka


WilliamHPNielsen:
@nulinspiratie, on behalf of the Copenhagen devs:

We certainly appreciate your effort, but need to raise a few words of caution.

Even without having looked at the code in any detail, learning from our experience with our last big change to the parameters, where we had to clean up and hotfix for months after merging, there is no way such a fundamentally changing PR can be merged in one go.

Changing the API of `Instrument` is like a major version revision, something that we can't really bump without a long and careful discussion. Perhaps we even need a PEP/Spec-like system for these big changes (we could call them QEP's?). AS you are of course well aware, there has already been some discussion about parameter nodes, but it came to a halt before converging.

The above entails that the new `Instrument` would only be merged into QCoDeS via a fully independent alternative to the current `Instrument` that does not touch the existing Instrument and Channel classes. 

The work flow would be something like:

* Add parameter node
* Port existing drivers in qcodes over slowly one by one. This will take month if not years.
* Then deprecate the existing Instrument/Channel classes.
* And perhaps eventually remove it fully. That would be long time into the future. 

This may seem like an unfulfilling snail pace to you, but we have a large user groups whose day-to-day work depends on QCoDeS.

nulinspiratie:
@WilliamHPNielsen Thanks for your reply. I definitely agree that this would be a significant change, and would break compatibility, which is a big issue affecting many users. I wasn't expecting this to be merged any time soon, if at all. However, It's something our fork will probably switch to, and I thought I'd place it here as a potential future direction for parameters (QEP does sound nice!).

AdriaanRol:
Hi Serwan, cool that you guys are still innovating on your QCoDeS fork!

I personally do not like the change in the get/set behaviour. I think there were good reasons why it is implemented the way it is and to be honest I personally think the way that part works is really solid and should not be changed. 

Another item (as mentioned by @WilliamHPNielsen ) is the potentially breaking backwards compatibility . I am paranoid about those and for me that is the chief reason that sometimes delays updates and creates diverging branches/forks, something which should be avoided in my opinion. 

I do think a first implementation of the parameter node is cool and potentially very useful, though I do not understand why you changed the inheritance for the base instrument. I also like your example notebook but I would like to have some more detail in there, mostly to see how it support the `print_readable_snapshot` (without the name changed) and what regular snapshots look like. 

I am afraid that in it's current form the node-tree functionality get's buried in the alternative get/set behaviour. I do think it is great as a conversation starter though :).

nulinspiratie:
Hey @AdriaanRol thanks for the useful comments. I definitely see your point that the get/set behaviour works well for Instruments, especially because it makes explicit when VISA commands are communicated with the physical instrument. I've made some changes to the ParameterNode that keeps this behaviour and may be a good middle way (see below). The reason why I removed the instrument inheritance from `InstrumentBase` to `ParameterNode` is that the `InstrumentBase` already functioned as a pretty good `ParameterNode`, and so I simply added a few functionalities and replaced the name. I think `ParameterNode` is better suited, since it can be used for things completely unrelated to instruments.

@WilliamHPNielsen Breaking backwards compatibility is definitely an issue, and I've been thinking how to deal with it. I think I've found a solution that works, namely by adding the flag `use_as_attributes` during the `ParameterNode` initialization. In short, it works as follows:

#### Setting `use_as_attributes=False`, which is the default
```
parameter_node = ParameterNode() # use_as_attributes is False by default
parameter_node.new_parameter = Parameter(set_cmd=None)
parameter_node.new_parameter = 42 # Set parameter value works as above
parameter_node.new_parameter(42) # Set parameter value can also be done via brackets like before
parameter_node.new_parameter # Returns Parameter instance.
parameter_node.new_parameter() # Returns Parameter value
```
This is the default behaviour of the ParameterNode, and pretty much behaves the same way as before. Accessing the Parameter attribute of a ParameterNode returns the Parameter, and not its value. 
In this case there are three differences compared to before, which all shouldn't break compatibility:
1. Setting a parameter value can also be done via `parameter_node.new_parameter = 42`. Currently we never set a parameter attribute, and so shouldn't break anything. We can also remove this behaviour if its unwanted.
2. The Parameter name doesn't need to be specified. If unspecified, it will set its name to that of the attribute.
3. Parameters can be added to the ParameterNode (and hence also Instruments) by simply setting its attribute. I also kept `add_parameter` to retain compatibility.

Pretty much all the tests still work, and I don't see why this wouldn't be backwards compatible. Would this be a viable solution? 

#### Setting `use_as_attributes=True`
```
parameter_node = ParameterNode(use_as_attributes=True)
parameter_node.new_parameter = Parameter(set_cmd=None)
parameter_node.new_parameter = 42 # Set parameter value
parameter_node.new_parameter # Returns 42
```
In this case, the `Parameter` attributes in the ParameterNode behave like the values they represent. This is best suited for more abstract classes (meta-instruments) that still need parameter functionality. An example is a Pulse object, whose parameter `Pulse.duration` can be varied in a loop, which in turn will program the instruments to output the pulse for a specific duration. In this case, it is more convenient to treat the parameter as an attribute, as the `Pulse` may have many more attributes, some of which aren't parameters (e.g. `@property` attributes). If getting/setting parameters happens different than other attributes, you would have to remember each time if an attribute is actually a parameter or not. For our group, the ability to access parameters as attributes definitely improves things.

nulinspiratie:
@WilliamHPNielsen @jenshnielsen Do you think this PR has a shot at being accepted? I understand the hesitancy to change core functionalities, but I do think the change wouldn't cause breaking changes. I've purposely left the old usage intact, and added functionality on top of it. This means that all the drivers etc. should still work as is. The tests are basically passing (one sometimes fails, but seems unrelated to this PR), and I think it would add an avenue to integrate Parameters more easily into classes that are unrelated to Instruments. If not, then we can close this PR.

PR #937, opened 2017-12-21
Add tests for Keysight_33XXX 

jenshnielsen:
Mainly because I wanted to experiment with the pyvisa sim backend

This also changes the blacklisting of driver from code coverage to not include the drivers that actually have tests

### DIFF
--- a/qcodes/.coveragerc
+++ b/qcodes/.coveragerc
@@ -4,6 +4,41 @@ omit =
     */__init__.py
     test.py
     tests/*
-    instrument_drivers/*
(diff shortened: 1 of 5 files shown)


WilliamHPNielsen:
Looks good. As an open question I'll ask whether it should be a goal for us to have 100% test coverage of drivers? I'm thinking that there will be functions so instrument-specific that it is not worthwhile or possibile to test them.

Even if we don't want to strive to achieve full coverage of drivers, I think it's good to include tested drivers in the coverage count.

PR #946, opened 2018-01-23
Adding averaging to Infiniium Oscilloscopes

Dominik-Vogel:
This is WIP and needs to be tested.


WilliamHPNielsen:
I'll try this out on a real scope.

WilliamHPNielsen:
I've tested this on an MSO8104a, the predecessor of the MSOS104a. Not surprisingly, there are features in the driver that do not work with the older oscilloscope model. Let's talk about making this driver working for both models, either by subclassing or a few `if self.model == XXX` statements.

Dominik-Vogel:
I have implemented the changes and tested the driver once more. It should be ready to be merged.

qcodes/instrument_drivers/Keysight/Infiniium.py
```diff
@@ -225,11 +267,27 @@ def __init__(self, name, address, timeout=20, **kwargs):
...
+        allowed_acquisition_rates = {
+            'MSO8104A': vals.Enum(*np.sort(
+                np.append([5e7, 1e8, 1.25e8, 2e8, 2.5e8, 5e8,
+                           1e9, 2e9, 4e9],
+                          np.kron([0.5, 1, 2, 2.5, 4],
+                                  10**np.arange(8))))),
+        }
+        self._max_acquisition_rate = max_acquisition_rate.get(model, inf)
+        self._allowed_acquisition_rates = allowed_acquisition_rates.get(
+            model, Numbers(0, self._max_acquisition_rate))
```
Dominik-Vogel: @WilliamHPNielsen it is a validator though... I confess that it is a slightly confusing choice of name from my side. I just didn't want to make the name any longer... I will call it `self._acquisition_rates_validator` instead to avoid confusion.

Issue #953, opened 2018-01-25
Unable to set current_ramp_limit above an arbitrary default specified in the driver


spauka:
Currently the driver for the AMI430 magnet controller throws an exception if the user attempts to set the current_ramp_limit above 0.6 A/s, preventing the instrument from initializing. This should be a warning rather than an exception.

### Steps to reproduce
1. ami = AMI430("ami", ip, port, current_ramp_limit=1.0)

### Expected behaviour
Should show a warning, but otherwise allow the instrument to initialize.

### Actual behaviour
Throws an exception, albeit one that is of type `Warning`

### System
Issue as at 85b3958d177f4770224be0739c97f5b5ff202cef

@sohailc

spauka:
Fixed by #954

PR #905, opened 2017-12-06
Hypothesis warnings

jenshnielsen:
Try to remove hypotesis warnings in tests

Also reduced the debug output a bit to not flood Travis completely

PR #954, opened 2018-01-25
Driver/ami430 allow all units & fix warnings as exceptions

### LINKED ISSUE #952: AMI430 Driver support for kGauss, minute units during operation
The AMI430 magnet controller supports multiple units that can be switched between, however the driver at the moment mandates Tesla+seconds. It would be good to add support for all units supported by the instrument.

spauka:
Fixes #952, #953.
In addition, adds a switch heater submodule that allows for setting of SwitchHeater parameters.

- Allows the AMI430 to be operated with ramp rates specified in minutes, allows field units in kiloGauss.
- Adds an instrument specific exception for instrument errors (i.e. allows us to differentiate why ramps may have failed between validation errors and instrument state)
- Fixes bug in current_ramp_limit code that prevents user from setting a limit higher than 0.6A/s
- Allows user to set `field_ramp_limit`, `field_rating` in instrument units, rather than having to convert to Amps.
- Adds `AMI430_SwitchHeater` submodule that allows configuration of switch heater parameters.
- Removes duplicated code between `set_field`, `_set_field` and `ramp_to`. Makes `block` a named parameter.
- Deprecate ramp_to as it duplicates the functionality of ramp_to.
- Don't automatically turn on switch heater before ramping, we may have a persistent current and this could cause a quench.
- Ensure the number of segments in ramps is 1. There is currently no support in the driver for more than 1 ramp segment.

### API changes
There are **no changes** to the public facing API apart from the addition of new features and the deprecation warning on ramp_to.

@sohailc I notice that you're also working on changes to how 3D works in this driver, in PR #910. None of the changes as far as I can tell intersect with yours, but as a result I haven't spent much time adding documentation or extending 3D functionality. Can we chat and discuss how best to add docs and type hints such that we don't conflict?

@jenshnielsen @sohailc

### DIFF
--- a/qcodes/instrument_drivers/american_magnetics/AMI430.py
+++ b/qcodes/instrument_drivers/american_magnetics/AMI430.py
@@ -327,7 +327,7 @@ class AMI430(IPInstrument):
         """ Set the ramp rate of the first segment in Tesla per second """
         if rate > self.field_ramp_limit():
             raise ValueError("{} {} is above the ramp rate limit of {} {}".format(
-                rate, self.ramp_rate.unit, 
+                rate, self.ramp_rate.unit,
                 self.field_ramp_limit(), self.field_ramp_limit.unit))
         self.write('CONF:RAMP:RATE:SEG 1')
         self.write('CONF:RAMP:RATE:FIELD 1,{},0'.format(rate))
--- a/qcodes/instrument_drivers/american_magnetics/AMI430.py
+++ b/qcodes/instrument_drivers/american_magnetics/AMI430.py
@@ -11,6 +11,8 @@ from qcodes.utils.validators import Numbers, Anything
 
 log = logging.getLogger(__name__)
 
+class AMI430Exception(Exception):
+    pass
 
 class AMI430(IPInstrument):
     """
--- a/qcodes/instrument_drivers/american_magnetics/AMI430.py
+++ b/qcodes/instrument_drivers/american_magnetics/AMI430.py
@@ -2,12 +2,13 @@ import collections
 import logging
 import time
 from functools import partial
+from warnings import warn
 
 import numpy as np
 
-from qcodes import Instrument, IPInstrument
+from qcodes import Instrument, IPInstrument, InstrumentChannel
 from qcodes.math.field_vector import FieldVector
-from qcodes.utils.validators import Numbers, Anything
+from qcodes.utils.validators import Bool, Numbers, Ints, Anything
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -2,15 +2,97 @@
...
 
-from qcodes import Instrument, IPInstrument
+from qcodes import Instrument, IPInstrument, InstrumentChannel
 from qcodes.math.field_vector import FieldVector
-from qcodes.utils.validators import Numbers, Anything
+from qcodes.utils.validators import Bool, Numbers, Ints, Anything
 
 log = logging.getLogger(__name__)
 
+class AMI430Exception(Exception):
```
sohailc: I do not quite understand why we need to make a custom exception. Can you explain?

sohailc:
This seems like a good modification to me. I have one comment if you can address that, please. Other then that, good job

qcodes/instrument_drivers/american_magnetics/AMI430.py
spauka: Just so that errors that are specific to the instrument state are differentiated from user errors (like `ValueError`s) or python errors. I suppose they can be replaced by `RuntimeError`s, but this seems like a poor fit given that they are used everywhere, and the error type suggests the error occured within python.

In particular, I imagined using it to detect a failure to sweep field in my own code, then automatically taking action.

sohailc: Yeah I see your point.

spauka:
Awesome :)

I can't merge, but if you guys are happy can we merge this?

sohailc:
I think before we can merge we need to make sure all checks pass. It should be quite trivial to make this happen :-)

WilliamHPNielsen:
@sohailc But the tests are passing? The few issues from Codacy are actually non-issues. See here: https://github.com/PyCQA/pylint/issues/1609. The `lambdas` were there before this PR.

sohailc:
Even though the lambda's are from before this PR, I think we should take every chance we get to improve code quality. @spauka can you please review the code quality suggestions and make improvements accordingly? They won't take more than a couple of minutes. 

@WilliamHPNielsen Lets ignore the Codacy issues.

sohailc:
Oh and obviously please make sure that this branch is up to date with the base branch.

spauka:
Hey @sohailc. Regarding Codacy issues:
- As @WilliamHPNielsen, unused __class__ variables are not true, both `AMI430SwitchHeater` and `_Decorators` are used.
- Regarding `lambda`s, there isn't an easy way around this. Both `get_cmd` and `set_cmd` wrap the argument as a `Command`, which checks `qcodes.utils.deferred_operations.is_function`, which returns `False` for a parameter. Either the behavior of `Command` is changed, or it has to be wrapped in a function or `lambda`.

Regarding code coverage, I will aim to add tests for the switch heater tomorrow.

sohailc:
@spauka I don't understand. Why can you not change this: 
```
get_cmd=lambda: self.current_ramp_limit()
```

to 
```
get_cmd=self.current_ramp_limit
```
?

spauka:
@sohailc See the above command. `self.current_ramp_limit` is a parameter, and `is_function` does not detect a `Parameter` as a function with the correct number of parameters.

sohailc:
You are right. I will merge this branch now after updating it

sohailc:
Hang on, the CI tests have failed this time. @spauka can you re-run the test by making a trivial change and pushing your code again?

spauka:
@sohailc Should be all good :)

sohailc:
Merged :-)

WilliamHPNielsen:
@sohailc When the tests fail, you (unfortunately) have to examine why. Travis is *very* slow at running tests, and a test counts as failed if it exceeds the hypothesis deadline. That puts us in a situation where tests fail not due to problems with the code, but heavy load on Travis' servers. It should be pretty easy to deduce from the Travis log what happened.

It is not optimal this way, and we should probably increase/disable deadlines if this keeps happening. @jenshnielsen already did this a few times, see #912 and #905.

jenshnielsen:
We should probably just disable the hypothesis deadline on those tests

PR #960, opened 2018-02-05
Use beta release of sphinx rather than a custom tag

jenshnielsen:
We need 1.7 to fix issues with forward references


jenshnielsen:
This changes the coverage appently because the tests are no longer correctly excluded. I dont think this is related. Rather it's caused by a new release of coverage which is pulled in here

jenshnielsen:
We can actually get rid of that line completely, The release version is installed from doc_requirements.txt

WilliamHPNielsen:
Good point.

PR #962, opened 2018-02-05
added kwargs to init of QDac

Dominik-Vogel:
All drivers should pass on kwargs to the instrument/VisaInstrument class.
For example if one wants to use the simulation backend one has to use the visalib kwarg. As of now this is not possible for the QDac


jenshnielsen:
The change in coverage is due to the same issue as noted in #960

jenshnielsen:
👍  Yes agreed the kwargs should be forwarded

PR #963, opened 2018-02-06
Decadac fix Min/Max_values

### LINKED ISSUE #904: driver/Decadac Voltage scaling
### Problems
The Decadac itself, the slots and the channels initialize with minimum and maximum values for every channel at +-5V:
```
class Decadac(VisaInstrument, DacReader):
    def __init__(self, name, address, min_val=-5, max_val=5, **kwargs):
...
class DacSlot(InstrumentChannel, DacReader):
    def __init__(self, parent, name, slot, min_val=-5, max_val=5):
...
class DacChannel(InstrumentChannel, DacReader):
    def __init__(self, parent, name, channel, min_val=-5, max_val=5):
```
Problems with this are:
 - the actual supported hardware limits are [-10,0], [-10, 10] and [0,10]. [-5, 5] is not valid
 - it is complicated to set different ranges for different channels

### Hardware limits
The `min_val` and `max_val` attributes have two purposes:
* The Decadac driver scales the supplied voltage from float to an integer according to their values
* they serve as limits for value validation
On the front panel of the decadac is a three state switch, that allows to set the ranges as described above. So any other setting than one of those does not make any sense in terms of conversion. If other ranges are wished for in validation, this should be implemented separately.
-> we should implement hardware limist in three modes -1, 0, 1 corresponding to the three ranges
-> limits for validation should be added independently (but of course compatible)

### Per channel setting
The `min_val` and `max_val` attributes can simply be set per channel to achieve the right scaling. But this leads to problems with validators and ramps that will not update
-> we should add the option to provide `min_val` and `max_val` as dictionary to the constructor and maybe supply a method to change those.

Once there is time I can make a PR. For now I just wanted to share the issue and get feedback, especially from @jenshnielsen and @WilliamHPNielsen who wrote the driver.

Dominik-Vogel:
Fixes #904.


spauka:
This doesn't cover DecaDac's with a +5/-5V range (which is like the ones we have in Sydney)? 

Actually looking at #904, I initially wrote the driver with +5/-5 since those ARE the hardware limits that our DAC's use, so in fact min_val and max_val were intended to be hardware, not soft limits...

jenshnielsen:
That's interesting. None of the Copenhagen decadacs have any 5 V limits. There are clearly more variations out there than we know about

Dominik-Vogel:
Thank you @spauka ! It was really a mystery for us how the +-5V came along. What do you guys think about rewriting the driver such that it requires setting the hardware limits explicitly for it to work?  Or do you @spauka know how to read out this piece of information from the device?

spauka:
@Dominik-Vogel I'm not aware of a way of reading out the range unfortunately? 

Actually I was unaware that it was possible for different cards within a single DAC unit to be set differently, although I was aware that there were different ranges floating around, so setting min/max_value when initializing the instrument was my solution.

Rewriting the driver to explicitly set hardware limits sounds good to me :)

astafan8:
@Dominik-Vogel 
- [ ] Rewriting the driver to explicitly set hardware limits

PR #964, opened 2018-02-06
Setting _VERSA_EEPROM_available to False by default

Dominik-Vogel:
The current Decadac driver does not specify the value of _VERSA_EEPROM_available by default. It is only set to false if the feature is unavailable but never gets set to True. In this PR I set it to False by default instead of inserting the probably missing assignment of True, in order to prevent inserting a so far untested and also not missed feature.


Dominik-Vogel:
This has been tested and it works, for the test station (T1/T4)

PR #966, opened 2018-02-07
IpInstrument: reraise error if connection refused

Dominik-Vogel:
As of now errors are caught and nothing is reported if the connection is refused. This PR adds a simple raise to reraise the exception


astafan8:
@Dominik-Vogel shall we abandon or merge ? :)

Dominik-Vogel:
I think we should merge. Not sure how this PR got lost on the way. Thank you for catching this @astafan8 !

Issue #621, opened 2017-05-18
Data mingled across columns in dataset


MerlinSmiles:
### Steps to reproduce
1. Measure something with `.run(use_threads=True)`

### Expected behaviour
Get a neat dataset

### Actual behaviour
Data mingles across columns in dataset 💘 

### System
**operating system**
Win 10

**qcodes branch**
Master

Two identical measurements:
![image](https://cloud.githubusercontent.com/assets/8595386/26193146/99b0d318-3bb4-11e7-9e93-473d782fe41b.png)

MerlinSmiles:
just an example data from a different dataset, see the lines marked with love, columns 2 and 3 switch around...

```
-5.994    4.27718    0.0173826    0.0467188
-5.991    4.27718    0.0173826    0.0467227
-5.988    0.0173826    4.27718    0.0467227 # <3
-5.985    4.27718    0.0173826    0.0467227
-5.982    4.27718    0.0173826    0.0467227
-5.979    4.27718    0.0173826    0.0467188
-5.976    4.27718    0.0173826    0.0467227
-5.973    4.27718    0.0173826    0.0467227
-5.97    0.0173826    4.27718    0.0467227 # <3
-5.967    4.27718    0.0173826    0.0467227
-5.964    0.0173826    4.27718    0.0467227
-5.961    4.27718    0.0173826    0.0467265
-5.958    4.27718    0.0173826    0.0467265
-5.955    4.27718    0.0173826    0.0467265
-5.952    4.27718    0.0192095    0.0467303
-5.949    4.27718    0.0192095    0.0467265
-5.946    4.27718    0.0192095    0.0467265
-5.943    4.27718    0.0192095    0.0467265
-5.94    0.0192095    4.27718    0.0467265 # <3
-5.937    4.27718    0.0192095    0.0467303
-5.934    4.27718    0.0192095    0.0467303
-5.931    4.27718    0.0192095    0.0467303
```

jenshnielsen:
Thats really bad. I tried reproducing it with a couple of dummy instruments but didn't see it. Are you measuring the 3 parameters from the same instrument? I am wondering if it's the driver which is not thread safe?

giulioungaretti:
@jenshnielsen  I think that's def a possibility.

MerlinSmiles:
yup, just commented to @giulioungaretti on slack, it happens when I use two parameters on the same instrument, in this case its an SR830 lock-in

[...]

PR #971, opened 2018-02-19
Add read until terminator support for ip instrument

jenshnielsen:
Hopefully fix stability issues that we have been seeing with the AMI 430 magnet

The idea is to ensure that we always recv until we get read termination char. The current implementation assumes that one recv is always enough to get all the data. That is probably true in almost all cases but might be wrong under heavily congested network traffic.

Still need to check the Terminator for the Mercury. Judging from the manual there may not be one.
Currently it is set to None


qcodes/instrument/ip.py
```diff
@@ -142,10 +166,18 @@ def _send(self, cmd):
...
+        got_read_terminator = False
+        rtl = len(self.read_terminator)
+        result = b''
+        while not got_read_terminator:
+            partresult = self._socket.recv(self._buffer_size)
+            result += partresult
+            if partresult == b'':
+                log.warning("Got empty response from Socket recv() "
+                            "Connection broken.")
+            elif (partresult[-rtl:] == self.read_terminator or
```
sohailc: I would use "endswith" here: 
https://www.tutorialspoint.com/python/string_endswith.htm

sohailc:
Should we not make the read_terminator default to terminator? I think we should always call recv until the read terminator has been seen.

qcodes/instrument/ip.py
jenshnielsen: Agreed that is slightly more clean

jenshnielsen:
I agree. I wanted to test with other instruments before doing that but just doing a single recv without any checks is fundamentally wrong

sohailc:
Other things I am concerned about: 
1) How do we prevent getting stuck in an infinite loop? Maybe we should have a timeout? 
2) How do we know the terminator string is at the end of the string we get from recv? 

Maybe the following link can help us: http://code.activestate.com/recipes/408859-socketrecv-three-ways-to-turn-it-into-recvall/

jenshnielsen:
1. We already set a timeout so the first call that times out will raise a TimeoutError. We should probably catch and handle that better. 
2. As long as we send commands one by one I think we should always get a buffer that ends with a terminator but if we send more than one command before reading then you are right it can be anywhere in the stream.

sohailc:
I think we should be careful and not assume we are reading before another command is send. As is done in the example code I linked to:

1) We should see if the terminator marker is in the string
2) If the termination marker is more then one character, it can be broken up in two read actions. 
3) If the termination marker is half way down the string, what do we do with the part of the string after the marker? Its part of a response to another querry so it should be stored in a buffer somewhere and appended to the string of the next recv command. 

All of this will complicate our code dramaticaly so I am wondering if we really should do all of this. I am leaning towards a "yes" though as it will make our code more robust.

MatthieuDartiailh:
This kind of issues is also addressed in pyvisa-py (tcpip-socket backend). You may be interested in giving it a look.

jenshnielsen:
In general we should strongly consider if we really need this IP class. I would expect we can and I think we should replace both the magnet drivers with visa drivers that connects to visa over sockets.

sohailc:
I agree! We do not want to duplicate efforts

jenshnielsen:
* This now adds a comment suggesting to use sockets over visa.
* Takes into account that the EOL may be split over more than one recv
* Clips the EOL and any residual text from the recv string
* Warns if any text follows the EOL. If that is the case we are doing something wrong

jenshnielsen:
@WilliamHPNielsen would be useful to look at this in the context of the mercury

WilliamHPNielsen:
@sohailc, do you see any reason to not merge this guy?

sohailc:
Well I still don't like the way partial messages are treated. I think there is a more robust way to do this and one which does not throw away anything. Here is an example of how I would do this: https://gist.github.com/sohailc/a0fcbfff5359f5fc4cc1ad9f2e0707ba

jenshnielsen:
@sohailc  imho this is of little consequence. If we ever read more than one reply it is likely due to a non empty output buffer even before whatever write command we did. Most likely the output of the last recv is the answer that you actually want. In principle we need to prevent this from ever happening as the reply is likely to be wrong no matter what. 

Honestly perhaps we should just raise and close the connection in that case.

sohailc:
Jens wrote:  "If we ever read more than one reply it is likely due to a non empty output buffer even before whatever write command we did."

I'm sorry Jens, but I just don't agree. I think we need to write our code and flexible and general as possible and not include any assumptions like the one you are making. Keep in mind that this modification will not only affect the MercuryiPS driver but all IPInstruments. If we receive jibberish from the Mercury power controller the proper location to raise and close the connection is in the instrument driver, not in the base IPInstrument class which I think we should program as general as possible. 

I you guys disagree feel free to merge this PR, but it is just not the way I would do things.

jenshnielsen:
Ok Sohail I will be more explicit. I think your solution will result in wrong results in all relevant cases for all our IP Instruments. I have identified a real use case where it will result in incorrect results. This applies to all IPInstruments and not specifically the mercury. I cannot see any use cases where it does anything useful. If you can identify such a use case I am happy to reconsider

Try looking at the _recv function and how it is used. It is only used after a send command. Each send command must result in a single reply or no reply. If it does not the whole class is fundamentally broken.

sohailc:
I have not tested my particular implementation with any instruments. I was trying to explain that there should be more robust ways of dealing with instrument replies. I am sure you are right that my solution will result in wrong results, but can you share your reasoning? Which use cases have you identified? 

As for your question of when my solution would be useful, what about if a reply to a _recv command is so long that an instrument sends it in two or more chunks? We have never seen that but we cannot altogether rule it out, especially for older instruments.

sohailc:
To be a little more specific, we write `self._buffer_size = 1400` in our driver, which is a magic number. Why 1400 and 1500? Instead of assuming a buffer size we need to read until we reach the termination character and then decide what we want to do with the data. Please let's avoid magic numbers.

sohailc:
You know what, I think you are right. The case that a reply is send in two chucks is already taken care of in this PR. I am approving this PR

jenshnielsen:
Here is a real world example for an instrument with two parameters volt and curr

The output buffer on the instrument contains a reply to a command that has never been read i.e. "1.0e1V\n". This is not read due to a crash at an earlier run. 
They you ask the instrument `CURR?`
The _recv will then get"1.0e1V\n" or perhaps "1.0e1V\n5.0A\n` following the loop construct.
In the first case we will get the wrong result in any case with both solutions. In the second case you will return `1.0e1V` and keep `5.0A` in the internal buffer and return that on the next command.  
 
It's not about 2 or more chunks that's already handled correctly we are in any case reading until a termination char in this PR. It is only about two or more termination characters within one message. And as far as I can see your solution will still not handle that hypothetical situation correctly. If a command to the instrument returns two partial results separated by a newline char the ask command will still only return the first part and the next one will be left in the internal buffer of the Instrument for the next ask command giving an incorrect result here.

jenshnielsen:
@sohailc Just saw your message glad that we understand each other. This still need real world testing before merging can be considdered. And I think we should look into better handling of Interrupts within the _recv command. I suspect these are a source on the kind of inconsistency that I talk about

jenshnielsen:
Actually my solution does not handle it any better. I stand by what I said earlier. I think we should raise. 
Quoting the zen of Python `In the face of ambiguity, refuse the temptation to guess.`

sohailc:
Also, I am not raising issues dealing with multiprocessing/threading so we know our solution might be thread unsafe, right? I am not objecting because for now QCoDeS is single process anyway. But what if two processes call _recv and the internal instrument buffer contains to responses? 

I suppose we can deal with these kind of issues when we implement multiprocessing in the weeks/months to come?

jenshnielsen:
I suspect that no instrument is thread safe see https://github.com/QCoDeS/Qcodes/issues/621 for an example of what can happen in a Visa instrument. We should have a good system, locks or otherwise as part of a multiprocessing architecture to ensure that an instrument can only be called from one thread

WilliamHPNielsen:
@jenshnielsen said:
>  This still needs real world testing before merging can be considered.

As you saw, I did some preliminary testing (basically: it doesn't crash upon init, and commands are sent and read). I'll be happy to try and reproduce an actual fail case and see how the code lives up. What would you consider sufficient testing with an actual instrument?

jenshnielsen:
The one thing that i suspect still needs improving is the behavior when doing interrupts in the middle of get and set commands. 

I think it's hard to reproduce the issue that this aims to fix in a real world scenario. It is so far pure speculation that a recv may only return a partial command. (But according to the specs it is possible) But I suspect this may happen with high volumes of network traffic perhaps coupled with high cpu utilization

jenshnielsen:
We are generally moving away from the ip instrument so lets drop this

PR #972, opened 2018-02-19
[WIP] Driver/lakeshore336 and lakeshore 372

jenshnielsen:
A driver for Lakeshore Temperature Controllers 336 and 372. From QDev probe station by @Dominik-Vogel.

Includes a great new `GroupParamter`+`Group` classes for commands that set/get more than one parameter at once.

https://www.lakeshore.com/Documents/372_Manual.pdf
https://www.lakeshore.com/Documents/336_Manual.pdf

#### ToDo:
- [x] initial work by William, Jens, and Dominik
- [x] testing by William
- [x] testing by Mikhail
- [x] decide what to do with INTYPE current/voltage ranges - postpone once users need it
- [x] test waiting for set point
- [x] check difference in OUTMODE command between the models and adjust the code accordingly
- [x] add heaters for 336 model
- [x] add docstrings/labels/units
- [x] implement sensor status parameter correctly (as sum of statuses)
- [x] make `_has_pid` a kwarg
- [x] check base class features are supported by both models (esp 336)
- [x] move Parameter Group to some "common" part of instrument_drivers?
- [x] add docstrings
- [x] go thorugh TODOs
- [x] revisit `blocking_t` part
- [x] fix PID comments
- [x] revisit `set_setpoint_and_rangeq: range should be selected according to current temperature, not according to current setpoint. right?
- [x] `wait_until_set_point_reached` simplify and remove debug
- [x] PID or more verbose names?
- [x] `wait_until_set_point_reached` - if t is lower than sensor range, it keeps on waiting - add a note
- [x] model 336 - add heaters 3 and 4 (they are different that 1 and 2, for example, list of modes is different)
- [x] add type annotations
- [x] dont forget to refer to manual when necessary
- [x] make groupparameter inherit from parameter (not _baseparameter)
- [x] add more info to the notebook about the parameters
- [x] add info to the notebook about units of the setpoint
- [x] add simple plot-monitoring cell to the notebook
- [x] fix minor problems from the comments in this PR
- [x] see if [mypy bug](https://github.com/python/mypy/issues/4715) is still relevant
- [ ] final testing of 336 by Jens/William
- [x] final testing of 372 by Mikhail

#### WontDoInThisPR:
- tests: separate and explain what mock and what yaml file is used for
- add test for model 336
- agree on names for channels/inputs/measurement-inputs and sensors/outputs and refactor the code
- what are min/max values of SETP?
- different units for PID parameters between models (units)
- different setting resolutions for PID between heater channels (docstring)
- set `vals` for `excitation_range_number` based on `excitation_mode` (and then remove the big comment in the code)
- model 336 -> sensor -> range - add tables for each sensor type and switch them once sensor type is changed


qcodes/instrument/sims/lakeshore_model336.yaml
```diff
@@ -0,0 +1,42 @@
...
+          r: "{}"
+
+      sensor_name:
+        default: "my name is boring"
+        getter:
+          q: "INNAME? A"
+          r: "{}"
+        setter:
+          q: "INNAME A,\"{}\""
+          r: OK
```
WilliamHPNielsen: This "OK" response is now obsolete (#968).

qcodes/instrument_drivers/Lakeshore/Model_336.py
```diff
@@ -1,14 +1,26 @@
+from enum import Enum
+
 from qcodes import VisaInstrument, InstrumentChannel, ChannelList
-from qcodes.utils.validators import Enum, Strings
+from qcodes.utils.validators import Enum as QCEnum, Strings
+
+class ChannelDescriptor(Enum):
```
WilliamHPNielsen: I may be missing something important, but this seems like an unnecessary complication to add. The cons are: reduced code readability (renaming `Enum` to `QCEnum`, having to look up `ChannelDescriptor`).

What are the pros?

In any case, the naming conflict is easily resolved by either just importing `enum` or importing `qcodes.utils.validators` as `vals` (or both). `QCEnum` is an abonimation.

qcodes/instrument/sims/lakeshore_model336.yaml
Dominik-Vogel: OK

qcodes/instrument_drivers/Lakeshore/Model_336.py
Dominik-Vogel: Agreed, `QCEnum` is a really bad name - the state of the code is really WIP. 
But how is it a complication? I agree that it could also just be a dictionary. But does this make it simpler?

qcodes/instrument/sims/lakeshore_model336.yaml
jenshnielsen: I dont know if there was an actual test file using this file that I missed in the rebase

jenshnielsen:
There are some missing `from typing import Dict` due to the refactor and mypy error we should fix otherwise I guess we should get this in asap

Dominik-Vogel:
The driver is in a usable state now. It will need some more refactoring. The two models are not that similar after all. There are  a few things that have to be moved out of the base class into the 336 like the setpoints.
Furthermore I want to change the parameters such that they can also accept strings instead of `Enum`s so that they can be set from the StationConfiugrator.

qcodes/instrument_drivers/Lakeshore/Model_372.py
```diff
@@ -0,0 +1,158 @@
...
+                               set_cmd=f"PID {heater_index}, {{P}}, {{I}}, {{D}}",
+                               get_cmd=f'PID? {heater_index}')
+        
+        self.add_parameter('range', get_parser=lambda x: Range(int(x)),
+                           vals=vals.Enum(*[i for i in Range]),
+                           set_parser=(lambda enum_value: enum_value.value),
+                           set_cmd=f'RANGE {heater_index}, {{}}',
+                           get_cmd=f'RANGE? {heater_index}')
+
+        self.add_parameter('setpoint', get_parser=lambda x: Range(int(x)),
```
jenshnielsen: This has get parser twice

jenshnielsen:
@Dominik-Vogel This will need a bit of work to get the tests passing before we can review it

qcodes/instrument_drivers/Lakeshore/Model_372.py
Dominik-Vogel: Thanks...

Dominik-Vogel:
It is actually not as bad as it seems... I think it is only the repeated keyword (I simply forgot to delete the first get_cmd in the last commit) and the unused imports. Or do you see other errors?
Currently I am removing the Enums agiain, so that the driver can be used with the StationConfigurator. Additionally it needs some thinking how to get the two lakeshores onto the same basis.

jenshnielsen:
Currently both the tests and mypy fails

```
==================================== ERRORS ====================================
___________ ERROR collecting qcodes/tests/drivers/test_lakeshore.py ____________
tests/drivers/test_lakeshore.py:8: in <module>
    from qcodes.instrument_drivers.Lakeshore.Model_336 import Model_336
instrument_drivers/Lakeshore/Model_336.py:6: in <module>
    class Output_372(BaseOutput):
E   NameError: name 'BaseOutput' is not defined
----------- coverage: platform linux, python 3.6.3-final-0 -----------
Coverage XML written to file coverage.xml
!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
=========================== 1 error in 8.85 seconds ============================
The command "py.test --cov=qcodes --cov-report xml --cov-config=.coveragerc" exited with 2.
20.21s$ cd ..
mypy qcodes --ignore-missing-imports --no-strict-optional
qcodes/instrument_drivers/Lakeshore/lakeshore_base.py:65: error: Need type annotation for 'MODES'
qcodes/instrument_drivers/Lakeshore/lakeshore_base.py:66: error: Need type annotation for 'RANGES'
qcodes/instrument_drivers/Lakeshore/lakeshore_base.py:146: error: The return type of "__init__" must be None
qcodes/instrument_drivers/Lakeshore/Model_372.py:59: error: Name 'Model_372_Channel' is not defined
qcodes/instrument_drivers/Lakeshore/Model_372.py:65: error: Name 'Heater' is not defined
qcodes/instrument_drivers/Lakeshore/Model_372.py:66: error: Name 'Heater' is not defined
qcodes/instrument_drivers/Lakeshore/Model_372.py:67: error: Name 'Heater' is not defined
qcodes/instrument_drivers/Lakeshore/Model_336.py:6: error: Name 'BaseOutput' is not defined
qcodes/instrument_drivers/Lakeshore/Model_336.py:26: error: Name 'BaseSensorChannel' is not defined
qcodes/tests/drivers/test_lakeshore.py:19: error: Too many arguments for "__init__" of "object"
qcodes/tests/drivers/test_lakeshore.py:22: error: Need type annotation for 'cmds'
qcodes/tests/drivers/test_lakeshore.py:24: error: Need type annotation for 'queries'
qcodes/tests/drivers/test_lakeshore.py:100: error: Need type annotation for 'heaters'
```

Dominik-Vogel:
This is not very surprising, my last commit is also labeled WIP.

jenshnielsen:
Yes but the comment before that implied that it was ready for review, hence my message

qcodes/instrument_drivers/Lakeshore/Model_336.py
```diff
@@ -1,57 +1,48 @@
...
+
+    def __init__(self, parent, output_name, output_index):
+        if output_name not in ['A','B']:
+            self._has_pid = False
+        super().__init__(parent, output_name, output_index)
+        self.P.vals = vals.Numbers(0.1, 1000)
+        self.I.vals = vals.Numbers(0.1, 1000)
+        self.D.vals = vals.Numbers(0, 200)
+
+        self.range_limits.vals = validators.Sequence(
```
sohailc: Error: validators is unknown. This should be `vals.Sequence(...`

qcodes/instrument_drivers/Lakeshore/Model_336.py
```diff
@@ -1,57 +1,48 @@
...
-    """
+from typing import ClassVar, Dict
+import logging
+from .lakeshore_base import LakeshoreBase, BaseOutput, BaseSensorChannel
+import qcodes.utils.validators as vals
+
+log = logging.getLogger(__name__)
+
+class Output_336(BaseOutput):
+
```
sohailc: Can we add a docstring describing the usefulness of this class? This is not clear to me.

qcodes/instrument_drivers/Lakeshore/Model_336.py
```diff
@@ -1,57 +1,48 @@
...
+        'closed_loop': 1,
+        'zone': 2,
+        'open_loop': 3,
+        'monitor_out': 4,
+        'warm_up': 5}
+    RANGES: ClassVar[Dict[str, int]] = {
+        'off': 0,
+        'low': 1,
+        'medium': 2,
+        'heigh': 3}
```
sohailc: typo in "heigh". Should be "high"

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,302 @@
...
+            return result
+
+
+class Group():
+    def __init__(self, parameters, set_cmd=None, get_cmd=None,
+                 get_parser=None, separator=',', types=None) -> None:
+        self.parameters = OrderedDict((p.name, p) for p in parameters)
+        self.instrument = parameters[0].root_instrument
+        for p in parameters:
+            p.group = self
```
sohailc: There should be a public method called `set_group` on the `GroupParameter` class to do this. Do not assign attributes directly like this.

sohailc: In fact, if you make a parameter group like so: 

```
self._group_param = Group(
	set_cmd="bla {{a}} {{b}}"
)

self.add_parameter(
	"a", 
	group=self._group_param, 
	parameter_class=GroupParameter
)
```
Then line 19 will not be necessary and we will not need to set the `group` attribute because we pass it to the constructor. I think this is more clear.

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,302 @@
...
+        self.output_index = output_index
+        self.add_parameter('mode',
+                           val_mapping=self.MODES,
+                           parameter_class=GroupParameter)
+        self.add_parameter('input_channel', get_parser=int,
+                           parameter_class=GroupParameter)
+        self.add_parameter('powerup_enable',
+                           val_mapping={True: 1, False: 0},
+                           parameter_class=GroupParameter)
+        self.output_group = Group([self.mode, self.input_channel,
```
sohailc: I quite like the way the group parameter works. I think we should make this available generally to other instruments as well.

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
Dominik-Vogel: There is, as far as I can see, a mutual dependence of the parameter and the group: The parameter has to know the group, and the group has to know the parameter. 
This is because we want to call set/get on the parameter and it and needs the values of the other parameters to accomplish that.
So this double link is created in an init method. This can equally happen in the parameter or in the group. In your example the parameter is aware of the group but the group is not aware of the parameter.
I chose to put this linking into the group because it makes the code a little more compact and readable for my taste... But I can be convinced otherwise if you include the previous aspect in your argument.
concerning the attribute I wrote you a PM.

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,302 @@
...
+        return parser
+
+    def set(self, set_parameter, value):
+        if any((p.get_latest() is None) for p in self.parameters.values()):
+            self.update()
+        calling_dict = {name: p.raw_value
+                        for name, p in self.parameters.items()}
+        calling_dict[set_parameter.name] = value
+        command_str = self.set_cmd.format(**calling_dict)
+        set_parameter.root_instrument.write(command_str)
```
sohailc: why not `self.instrument.write(command_str)`?

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
sohailc: I think you are right about this. I think if you clean this code up a little (e.g. add docstrings, proper type hinting) it can be approved and merged. But I am still unsure if we should have a setter for the group or not. I read the link you send me on PM, but I still think having a setter will make things more clear to fellow programmers.

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,302 @@
...
+            return dict(zip(keys, values))
+        return parser
+
+    def set(self, set_parameter, value):
+        if any((p.get_latest() is None) for p in self.parameters.values()):
+            self.update()
+        calling_dict = {name: p.raw_value
+                        for name, p in self.parameters.items()}
+        calling_dict[set_parameter.name] = value
+        command_str = self.set_cmd.format(**calling_dict)
```
astafan8: this means that the `Group` only supports string commands, right? Shall we generalize it to callables as well?

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,302 @@
...
+            self.get_parser = self._separator_parser(separator, types)
+
+    def _separator_parser(self, separator, types):
+        def parser(ret_str):
+            keys = self.parameters.keys()
+            values = ret_str.split(separator)
+            return dict(zip(keys, values))
+        return parser
+
+    def set(self, set_parameter, value):
```
astafan8: is `get` method missing?

WilliamHPNielsen:
Testing of this driver for some simply/quick measurements has begun on QT3. As you can see, we already found one (rather crucial) bug. If the driver suddenly becomes important, I might take over and fix the issues pointed out by @astafan8 above.

qcodes/tests/drivers/test_lakeshore.py
```diff
@@ -0,0 +1,331 @@
+import pytest
+from typing import Dict, Callable
+import logging
+from functools import wraps
+from contextlib import suppress
+import time
+
+# from qcodes.instrument_drivers.Lakeshore.Model_336 import Model_336
```
astafan8: commented out?

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,302 @@
...
+
+        self.add_parameter('wait_equilibration_time',
+                           set_cmd=None,
+                           vals=validators.Numbers(0,100),
+                           label='Duration during which temperature has to be '
+                                 'within tolerance.', unit='')
+        self.wait_equilibration_time(0.5)
+
+
+        self.add_parameter('blocking_T',
```
sohailc: shouldn't this be a method?

astafan8:
Today I've executed the driver on an actual Lakeshore Model 372 instrument. Outcome:
* the driver seems to work
* `wait_until_set_point_reached` method has a bug in `log.debug` (left a separate comment about it)
* scientists requested the ability to turn the channels on and off. This is available via `INSET` command (but it also has multiple inputs...). The rationale for turning the channels off is the following: when Lakeshore measures values at sensors, it goes through the enabled ones one by one, sets proper ranges in its internal lock-in, etc. The more channels you have enabled, the more time it takes between measurements of values for a given sensor. Since the heater uses only one sensor, the other sensors can be turned off to increase the effective time of the heater's feedback loop "response". A minor detail here is that this command also allows to change the "curve" (temperature-resistance) of the sensor - users either should not be able to change it, or there should be huge warnings and banners and prompts before changing it, because you do not want to suddenly change the curve for which the sensor was calibrated for.

question: who is going to add channel on/off feature?

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,302 @@
...
+                    else:
+                        start_time_in_tolerance_zone = time.monotonic()
+                        is_in_tolerance_zone = True
+                else:
+                    if is_in_tolerance_zone:
+                        is_in_tolerance_zone = False
+                        start_time_in_tolerance_zone = None
+
+            log.debug(f'waiting to reach setpoint: temp at '
+                      f'{t_reading}, delta:{delta}')
```
astafan8: this line throws exception about unknown variable `delta` in case `t_reading` is `None`/`False`/etc. Also, why would we have `if t_reading:` condition in the first place?

qcodes/instrument_drivers/Lakeshore/Model_336.py
astafan8: fixed

qcodes/instrument_drivers/Lakeshore/Model_336.py
astafan8: done

qcodes/instrument_drivers/Lakeshore/Model_336.py
astafan8: fixed

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
astafan8: to keep things simple, i will use `instrument`. also, i will add to the docstring that it is assumed that all parameters in the group belong to the same instrument

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
astafan8: maybe later, but not in this PR

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
astafan8: `param_group.get()` does not seem to be a use case. and indeed, `update` is called every time one of the parameter's get method is called.

qcodes/tests/drivers/test_lakeshore.py
astafan8: fixed

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
astafan8: old Loop needs a parameter, so we end up with a parameter on top of a method instead of a method-only.

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
astafan8: fixed, `if t_reading:` is gone

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,549 @@
...
+        self.output_group = Group([self.mode, self.input_channel,
+                                   self.powerup_enable],
+                                  set_cmd=f'OUTMODE {output_index}, {{mode}}, '
+                                          f'{{input_channel}}, '
+                                          f'{{powerup_enable}}',
+                                  get_cmd=f'OUTMODE? {output_index}')
+
+        # Parameters for Closed Loop PID Parameter Command
+        if self._has_pid:
+            self.add_parameter('P',
```
sohailc: Don't call this parameter just "P". Consider "loop_control_proportional"

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,549 @@
...
+                               label='P value of closed-loop controller',
+                               docstring='The value for control loop '
+                                         'Proportional (gain)',
+                               vals=vals.Numbers(0, 1000),
+                               get_parser=float,
+                               parameter_class=GroupParameter)
+            self.add_parameter('I',
```
sohailc: Consider calling this "loop_control_intergral"

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,549 @@
...
+                               label='I value of closed-loop controller',
+                               docstring='The value for control loop '
+                                         'Integral (reset)',
+                               vals=vals.Numbers(0, 1000),
+                               get_parser=float,
+                               parameter_class=GroupParameter)
+            self.add_parameter('D',
```
sohailc: Consider calling this "loop_control_differential"

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,549 @@
...
+                                  set_cmd=f'OUTMODE {output_index}, {{mode}}, '
+                                          f'{{input_channel}}, '
+                                          f'{{powerup_enable}}',
+                                  get_cmd=f'OUTMODE? {output_index}')
+
+        # Parameters for Closed Loop PID Parameter Command
+        if self._has_pid:
+            self.add_parameter('P',
+                               label='P value of closed-loop controller',
+                               docstring='The value for control loop '
```
sohailc: In the docstring, refer to the manual at the right page.

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
```diff
@@ -0,0 +1,471 @@
...
+        # there is a `+1` because `self.RANGES` includes `'off'` as the first
+        # value.
+        self.output_range(self.INVERSE_RANGES[i+1])
+        return self.output_range()
+
+    def set_setpoint_and_range(self, temperature):
+        self.set_range_from_temperature(temperature)
+        self.setpoint(temperature)
+
+    def wait_until_set_point_reached(self,
```
sohailc: I think this method is overcomplicated. Why not do: 

```python
def while_not_at_setpoint(tolerance, settling_time): 
    
    t = time.time()
    t0 = t
    
    while t - t0 < settling_time: 
        
        t = time.time()
        if abs(temperature() - setpoint()) > tolerance: 
            t0 = t
        
        time.sleep(0.1)
```

sohailc: And whats up with all the verbose logging. Its distracting from what you are trying to accomplish

astafan8: sure, I will simplify and clean it up.

qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
astafan8: as discussed, i'm going to do it whenever necessary. (added to todo list so that i dont forget)

astafan8:
Dear @QCoDeS/core , I would like to wrap up my work on this driver for now. See the description for the list of items that have been done (and more). Please, review and leave your comments.

I propose that after testing the two drivers on the real devices, we merge this PR, and create an issue for the bullet items which are not covered in this PR. This is bad, I know, but this will allow our users to use the new driver (at the moment, they are using it via merging this branch into their other branches).

astafan8:
@QCoDeS/core The 372 driver is in active use now at two stations in Delft, everything works as expected and the users are passing their gratitude to those who worked on the driver :)

So shall we merge it "as is" (see my previous message for details)?

WilliamHPNielsen:
@astafan8, can we make the CI pass first?

Dominik-Vogel:
Sorry guys I didn't have time to get back to this one... Let me see if I can get the ci through and then merge it

Dominik-Vogel:
Now all ci passes. It would be great to get a confirming review ;-) so that I can merge

WilliamHPNielsen:
You don't need that for merging.

PR #973, opened 2018-02-19
Pyqt if remote process is crashed it may throw ConnectionResetError

jenshnielsen:
As seen on qt2 in Copenhagen


WilliamHPNielsen:
This looks good. Is there a way to test this, can we reproducibly "make it crash"?

jenshnielsen:
Probably not without significant work. I tested it manually by forcing a divide by zero error and catching that manually

WilliamHPNielsen:
In that case I think we should get this in.

PR #974, opened 2018-02-19
Keithley2600 add resistance

jenshnielsen:
Taken from Qdev prope station


jenshnielsen:
@QCoDeS/core  any objections to this simple change

PR #975, opened 2018-02-20
Add logging of messages to mercury driver

jenshnielsen:
@ThorvaldLarsen


jenshnielsen:
@QCoDeS/core I think this should be safe to land

WilliamHPNielsen:
I just pushed a small change making the log slightly more readable-at-a-glance. Let's get this merged.

PR #976, opened 2018-02-21
[WIP] Ats9440 driver

jenshnielsen:
Should probably try to reduce code duplication

@nulinspiratie Did you have a local driver for this card too?


nulinspiratie:
@jenshnielsen Yeah I do, compared them and they're nearly the same. Only thing I noticed is that for channel selection we have:
```
                           byte_to_value_dict={1: 'A', 2: 'B', 3: 'AB', 4:'C',
                                               5: 'AC', 6: 'BC',8:'D', 9: 'AD',
                                               10: 'BD', 12: 'CD', 15: 'ABCD'})
```
While in your driver the byte is linearly increasing. I can't test which one is correct, but I think it should be the one above.

astafan8:
@jenshnielsen close/merge/leave alone ? :)

jenshnielsen:
We still have some of these cards. So hopefully I will eventually have time to come back to this

PR #977, opened 2018-02-21
Driver/Tektronix AWG5208

WilliamHPNielsen:
- Expand the AWG70000A parent class to accomodate for 5208's
- Add a 5208 driver

Note that we have an AWG5200 driver that subclasses the 5014 driver. This file should probably be deleted, as the software is very different on those two AWGs, and noone seems to be using that driver (@peendebak, do you confirm/agree?).

Work items:

- [x]  PyVISA tests
- [x] Tests with the instrument

@sohailc @jenshnielsen


WilliamHPNielsen:
I would call this a working driver. @sohailc, what do you say?

sohailc:
Let me have a look at this latet this afternoon

WilliamHPNielsen:
@jenshnielsen Let's deprecate it now. Or even remove it. We have intelligence saying that no one is using that driver.

PR #978, opened 2018-02-22
Zi  UFLI add support for Data acquisition module

jenshnielsen:
Basic support with missing support for multiple rows, grids, Sample interpolation and FFT 

* Also partial support for 'MF' option (adding 8 independent oscillators) 
Still needs to figure out how to set frequencies and mappings
* Added units to setpoints for sweeper


qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -1718,14 +2027,12 @@ def ampdef_valid():
...
-                              'amplitudes/3': amp_valid,
-                              'amplitudes/7': amp_valid,
+                              'amplitudes': amp_valid,
                               'offset': offset_valid}
-
         def update_range_offset_amp():
             range_val = params['signal_output{}_range'.format(number+1)].get()
             offset_val = params['signal_output{}_offset'.format(number+1)].get()
-            amp_val = params['signal_output{}_amplitude'.format(number+1)].get()
+            amp_val = params['signal_output{}_amplitude1'.format(number+1)].get()
```
jenshnielsen: This is wrong

PR #979, opened 2018-02-23
Fix incorrect inferred_from retrieval

WilliamHPNielsen:
As pointed out by @sohailc 

- Make paramspecs have their `inferred_from` parameters retrieved correctly
- Add a test that didn't pass before this fix

@jenshnielsen


sohailc:
Looks good to me

PR #980, opened 2018-02-26
Fix tests for pyvisa19

jenshnielsen:
Hopefully fixes Travis failing 

We change the warning to be ami specific and only capture that
We have to rewrite the checking code because exeinfo will contain data for all warnings
not just the captured ones so use the context manager instead to to the check


jenshnielsen:
@WilliamHPNielsen

jenshnielsen:
Self merging to get travis passing

PR #982, opened 2018-02-27
Unpin coverage

jenshnielsen:
The bug in coverage 4.5.0 should be fixed in 4.5.1, see https://bitbucket.org/ned/coveragepy/issues/638/run-omit-is-ignored-since-45

@QCoDeS/core


jenshnielsen:
@WilliamHPNielsen and more importantly :) so does codecov delta for this pr is exactly zero as expected

PR #983, opened 2018-02-27
Fix visa deprecation with pyvisa 1.9

jenshnielsen:
Change ask to query.

I guess a relevant question is if the qcodes visa api should be adapted to match the pyvisa api

This need testing with both pyvisa 1.8 and 1.9 with real instruments before merging

@QCoDeS/core


jenshnielsen:
@QCoDeS/core ask is literally an alias for query in both 1.8 and 1.9 (we require at least 1.8 already) so the risk of landing this should be ultra low

PR #984, opened 2018-02-28
limit numpy to 1.13.x until h5py is compatible with 1.14

jenshnielsen:
Fixes annoying warning at the start of qcodes sessions


jenshnielsen:
@QCoDeS/core

PR #985, opened 2018-03-01
Oxford IPS120: Fix the return of a few methods for the magnet

CJvanDiepen:
Match the return statement of some magnet functions with other functions in the magnet class calling those functions.


WilliamHPNielsen:
Let me see if I understand this. The point of this PR is to make the `set` and `get` methods of some parameters match, such that if `set` takes an `int` as input, `get` also returns an `int`. Is that correct? If so, I think that is certainly a good change.

But I don't see a reason for explicitly printing the strings. We are usually against unconditional printing in drivers. Could logging not work here? 

Also, the docstrings should be updated, now that the return values are `int`s and not strings.

CJvanDiepen:
@WilliamHPNielsen 

It is indeed about matchting the input and output arguments of the `set` and `get` methods. We explicitly print the strings to the console, because the integers are not informative. Is there a more esthetic way of getting a string output to the ipython console but not using the explicit string? Is there a way in which logging could do that for us?

WilliamHPNielsen:
@CJvanDiepen the "proper" way to get this information out is to use a module level logger (already in place) and put `log.info("Remote control status: blah blah")` in the driver.

But if the ints are not informative, why not use a `val_mapping` with the parameter? Then it can be a string based parameter with meaningful values, so that e.g. the snapshot is also easily readable.

astafan8:
May we close this PR in favor of https://github.com/QCoDeS/Qcodes/pull/1402 ? This PR just adds print statements, while the mentioned PR does some useful refactoring, and uses the new logging from the driver instance instead of the print statmements.

CJvanDiepen:
I am fine with that.

PR #987, opened 2018-03-05
Added a printer function which is dynamic in output.

sohailc:
A simple function that will print inserted data to stdout in such a way that single line on the output prompt is refreshed at every call. This prevents the output prompt from being filled with clutter from previous iterations.

### DIFF
--- a/qcodes/dataset/live_monitor.py
+++ b/qcodes/dataset/live_monitor.py
@@ -1,0 +1,47 @@
+"""
+This module is intended to provide various methods to monitor in real time
+data that is added to the data set
+"""
+
+from typing import Callable
+import sys
+
+
+def printer(names: list) ->Callable:
+    """
+    A simple function that will print inserted data to stdout in such a way
+    that single line on the output prompt is refreshed at every call. This
+    prevents the output prompt from being filled with clutter from previous
+    iterations.
+
+    Example:
+    >>> import numpy as np
+    >>> from qcodes.dataset.data_set import new_data_set, ParamSpec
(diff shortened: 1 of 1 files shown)


WilliamHPNielsen:
I really like the idea of a live_monitoring module. This printer function is also a great idea, but currently I think it's a bit too easy to screw up; any additional printing will interfere badly with it.

MWE of screwing it up:

```python
import numpy as np
from qcodes import new_experiment
from qcodes.dataset.data_set import new_data_set, ParamSpec
from qcodes.dataset.live_monitor import printer
from time import sleep
    
specs=[ParamSpec("x", "numeric", unit='Hz'), ParamSpec("y", "numeric", unit='V')]
new_experiment('test', 'no_sample')
data_set = new_data_set("test", specs=specs)
sub_id = data_set.subscribe(printer(["x", "y"]), state=[])
for x in np.linspace(100, 200, 15):
    y = np.random.randn() 
    data_set.add_result({"x": x, "y": y})
    sleep(0.2)
    print("I am here to wreck havoc")
```

sohailc:
Lol, I get your point. But I think we should not be too concerned with all the ways a user can screw this up. Else we will end up being completely paralyzed. Also, I think this should be seen as a relatively quick and dirty solution to a quite urgent problem. Let's get this quick solution out first and worry about a more durable solution later. 

The reason this is an urgent problem is that users have been used to saving to a text file rather than a database. The fortunate side effect of text files is that live monitoring is less difficult: simply open the text file with your favorite editor and keep refreshing! But this is suddenly not possible anymore. The effect of that is that users demand to save to text files in addition to the sql lite database. This is not what we want!

PR #991, opened 2018-03-07
Ensure name uniqueness

sohailc:
When creating a run, make sure the data set name is unique

### DIFF
--- a/qcodes/dataset/sqlite_base.py
+++ b/qcodes/dataset/sqlite_base.py
@@ -1175,7 +1175,7 @@ def _create_run_table(conn: sqlite3.Connection,
         transaction(conn, query)
 
 
-def get_all_names(conn: sqlite3.Connection):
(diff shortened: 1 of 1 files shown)


WilliamHPNielsen:
I think this looks good. A few comments:

- Perhaps we should do a database version update to get this change in. Although we are not changing table layouts, we somehow are changing the tables. @core?
- Regarding type hints, it would be nice to have the return value of functions written as well. 
- I couldn't see any validation of the given run name, but that should surely be in place.

WilliamHPNielsen:
@sohailc any news on this guy?

sohailc:
I worked around the naming non-uniqueness when implementing convenience functions for the group in Delft. I am ensuring that the experiment name / sample name combination is unique instead when inserting new data. We can merge this branch but if this is going to cause a lot of problems with updating exsisting databases, it is not going to hinder work in Delft anymore. @core do we think this PR is useful for other situations?

PR #994, opened 2018-03-08
 add root_instrument property

Dominik-Vogel:
This PR adds a get_root_instrument method to InstrumentBase, InstrumentChannel, and _BaseParameter.
as of now channels are treated as instruments that have a parent. To find the actual instrument a parameter in a channel belongs to, one has to iterate up until one finds an instrument that is not a channel.
With this PR one simply gets this root experiment with the added function.

### DIFF
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -243,6 +243,8 @@ class InstrumentBase(Metadatable, DelegateAttributes):
             else:
                 submodule.print_readable_snapshot(update, max_chars)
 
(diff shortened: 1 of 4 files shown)


jenshnielsen:
:+1: but I would make it a root instrument property

jenshnielsen:
@Dominik-Vogel Would be good to get this in but I think a `root_instrument` property would be much nicer

jenshnielsen:
@Dominik-Vogel I think this is ready to land

PR #999, opened 2018-03-12
alazar trigger delay must be a multiple of the alignment

jenshnielsen:
As in the title


WilliamHPNielsen:
Is the alignment always 8? What *is* the alignment even?

jenshnielsen:
Did you read the comment?

PR #1000, opened 2018-03-12
Feature/remove DeferredOperations

WilliamHPNielsen:
I think that the mixin class `DeferredOperations` causes much more confusion than its functionality is worth. I therefore propose that we get rid of it. This PR tries out what actually happens when we remove it.

Some arguments:

Without `DeferredOperations`, `Parameters` act the way you expect them to. In particular, they can be compared (see #993). It seems that the main drawback is that `Loop` actions have to be typed slightly more explicitly, e.g.
```
loop.each(self.p1, BreakIf(self.p1 >= 0.1)).run()
```
now has to be typed as
```
loop.each(self.p1, BreakIf(lambda: self.p1.get() >= 0.1)).run)
```
which in my opinion is not really even a drawback, since it is, well, more explicit and thus readable.

I am not sure if I have missed some critical use of `DeferredOperations`, though. And of course, I cannot tell if anyone is making heavy use of it.

What do YOU think?

@QCoDeS/core @QCoDeS/core-contributors


jenshnielsen:
I would prefer to keep DeferredOperations around (but deprecate it) and change the inheritance of Parameter (and others?) to not depend on it

WilliamHPNielsen:
@jenshnielsen, you are right, that is better. I think that is where we are now.

Dominik-Vogel:
I would also love to get rid of the deferred operations. I think Jens has a point in deprecating them first, so that people have a chance to react on it.

Dominik-Vogel:
It can obviously not be printed on every use. And as far as I can see, there is no pretty way to find out if someone has been using them. Would it be to annoying to print the warning as a default on loading qcodes? What alternative channels do we have?

WilliamHPNielsen:
> Would it be to annoying to print the warning as a default on loading qcodes? 

That is how it is currently in place. I think it's somehow a bit unfair to unsuspecting users.

> What alternative channels do we have?

That is indeed the question. I think we should just deprecate with no explicit warning.

jenshnielsen:
A warning.warn will only print once normally. The current warning should print if you import DeferredOperation right? If you move it to inside the ```__init__``` function it will print on creation time

WilliamHPNielsen:
But the `__init__` function is not always called from subclasses. Indeed, `GetLatest` inherited from `DeferredOperations`, but did not call the `__init__`.

jenshnielsen:
Hmn thats wrong not much we can do about it. But leaving it where it is should be fine then? It will trigger on import of DeferredOperation

WilliamHPNielsen:
Triggering on import is fine. It was a problem before because there were other things in the `deferred_operations` module that got imported by e.g. the tests. That is fixed now.

jenshnielsen:
I dont think that was the problem, I think the problem was that that the is_function helper function was actually using DeferredOperation

jenshnielsen:
but moving that function makes sense too

WilliamHPNielsen:
This one is ready for evaluation. @QCoDeS/core

PR #1001, opened 2018-03-13
Potential SQL bug

jenshnielsen:
I guess this method is unused but it looks wrong to me


WilliamHPNielsen:
It's being used in `get_dependents`, but it's probably better to remove `one_column` entirely and just use `many_many`.

PR #1002, opened 2018-03-13
[do not merge] Upgrade database to add GUID

jenshnielsen:
A new column is added to the `runs` table of the database to contain unique identifier of the run. The identifier is a special GUID that is based on timestamp and some settings in the qcodes configuration. Users outside of MS collaboration can safely ignore this feature.

Also replaces https://github.com/QCoDeS/Qcodes/pull/965 against master


qcodes/dataset/sqlite_base.py
```diff
@@ -199,8 +209,42 @@ def connect(name: str, debug: bool = False) -> sqlite3.Connection:
 
     if debug:
         conn.set_trace_callback(print)
+
+    perform_db_upgrade(conn)
```
WilliamHPNielsen: I think this is really nice; automatic upgrading.

jenshnielsen:
I think this should be ok now. @WilliamHPNielsen 

We still need to decide between pymongo/bson and python buildin UUIDs

qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -325,11 +325,25 @@ def test_database_upgrade(empty_temp_db):
...
+    set_user_version(connection, userversion + 1)
+    upgraded_userversion = get_user_version(connection)
+    assert upgraded_userversion == userversion + 1
+
+
+def test_perform_actual_upgrade_0_to_1(dataset):
+    connection = connect(qc.config["core"]["db_location"],
+                 qc.config["core"]["db_debug"])
+    userversion = get_user_version(connection)
+    set_user_version(connection, 0)
```
WilliamHPNielsen: This line and the next effectively performs the downgrade 1 to 0. Would it make sense to provide dedicated downgrading functions in `sqlite_base.py`?

jenshnielsen: I don't think so. I am not sure how to do this best. We don't want users to be able to run with the old version of the db so we unconditionally want to upgrade. On the other hand we want to have the old db so we can test it

WilliamHPNielsen:
Taking over this PR as agreed earlier today. Some extension to the config plus a UUID-generating function will follow.

Dominik-Vogel:
It seems to me that the data in the config file are similar to the functionality of the `CURRENT_EXPERIMENT` in the qdev-wrappers. I wanted to move those into core for a long time, so I am happy about this here. I think we should chose a more general name than `GUID_components`.  
One point we should consider is whether the sample name/id should live in the config, as it is, opposed to the rest of the config, something that is subject to frequent changes.

WilliamHPNielsen:
@Dominik-Vogel, I would actually like to keep this GUID-generating business completely separate from anything that users ever think about, so I don't agree with the `CURRENT_EXPERIMENT` overlap. I am not even sure that the config file is the appropriate place for this GUID information. You are right that it is not really clear yet how the sample code best gets in.

qcodes/dataset/guids.py
```diff
@@ -0,0 +1,70 @@
...
+def parse_guid(guid: str) -> Dict[str, int]:
+    """
+    Parse a guid back to its four constituents
+
+    Args:
+        guid: a valid guid str
+
+    Returns:
+        A dict with keys 'location', 'work_station', 'sample', and 'time'
+          and integer values
```
astafan8: typo: 'as' integer values?

qcodes/tests/dataset/test_guids.py
```diff
@@ -0,0 +1,39 @@
...
+        assert comps['work_station'] == stat
+        assert comps['sample'] == smpl
+        assert comps['time'] - gen_time < 2
+    except Exception as e:
+        raise e
+    finally:
+        # important to leave this vital info untouched!
+        cfg = Config()
+        cfg.current_config = original_config
+        cfg.save_config(cfg.current_config_path)
```
astafan8: newline at the end of file?

qcodes/tests/dataset/test_guids.py
```diff
@@ -0,0 +1,39 @@
...
+        guid = generate_guid()
+        gen_time = int(np.round(time.time()*1000))
+
+        comps = parse_guid(guid)
+
+        assert comps['location'] == loc
+        assert comps['work_station'] == stat
+        assert comps['sample'] == smpl
+        assert comps['time'] - gen_time < 2
+    except Exception as e:
```
astafan8: what is this `except` for?

WilliamHPNielsen:
@QCoDeS/core 

Okay, this PR got a bit out of hand, but that's life sometimes. Three major changes are introduced here:

1. The database upgrade system changed and became "more correct". The `connect` function that creates the database takes an optional version argument. This means that the database is always created at version 0 and then upgrades are applied.

2. The `atomic` context manager now yields itself. This is needed to give nested context managers awareness about who is "on top". This *is* actually an API change, but it also fixes a bug, so I think it is acceptable. We now write 
```python
with atomic(conn) as conn:
    transaction(conn)
```

3. The config system will upon import check the `GUID_components` part of the config and prompt for an input if they are found to be zero.

I think the main lines are good, but we should of course not merge before we have an overview of work stations in place. I'll also put info about the GUID in the relevant places in the documentation.

Reviews are welcome now.

qcodes/config/__init__.py
```diff
@@ -1,2 +1,27 @@
+import json
+
 from .config import Config
 from .config import logger
+from .config import DotDict
```
astafan8: DotDict unused?

qcodes/config/__init__.py
```diff
@@ -1,2 +1,27 @@
...
+
+cfg = Config()
+loc = cfg['GUID_components']['location']
+station = cfg['GUID_components']['work_station']
+
+if station == 0 or loc == 0:
+    print('Welcome to QCoDeS!\nTo use QCoDeS, please provide the system with '
+          'enough information to generate unique dataset identifiers.')
```
astafan8: how non-ms-collab users know what defaults/numbers they can/should use?

qcodes/config/__init__.py
```diff
@@ -1,2 +1,27 @@
...
+
+if loc == 0:
+    while loc == 0:
+        loc = int(input('Please enter an integer as your location code: '))
```
astafan8: is there maximum for the location number? if so, it needs to be mentioned and checked against.

qcodes/config/__init__.py
```diff
@@ -1,2 +1,27 @@
...
+    cfg['GUID_components']['location'] = loc
+
+if station == 0:
+    while station == 0:
+        station = int(input('Please enter an integer as your work'
+                            ' station code: '))
```
astafan8: is there maximum for the station code? if so, it needs to be mentioned and checked against.

environment.yml
```diff
@@ -18,4 +18,6 @@ dependencies:
  - pytest
  - pytest-runner
  - spyder
+ - pymongo
```
jenshnielsen: pymongo can hopefully be dropped (only used for bson)

qcodes/config/__init__.py
jenshnielsen: Could/should we move this into the dataset

.appveyor.yml
```diff
@@ -27,6 +27,7 @@ install:
 
 # Test scripts
 test_script:
+  - python qcodes\tests\setup_config.py
```
jenshnielsen: Does this mean that you always need to run this scrip before running the tests?

setup.py
```diff
@@ -64,7 +66,9 @@ def readme():
           'h5py>=2.6',
           'websockets>=3.2',
           'jsonschema',
-          'pyzmq'
+          'pymongo',
```
jenshnielsen: pymongo be gone

qcodes/config/__init__.py
```diff
@@ -1,2 +1,27 @@
...
+    cfg['GUID_components']['work_station'] = station
+
+with open(cfg.default_file_name, "w") as fp:
```
jenshnielsen: I think this should store in user home.

WilliamHPNielsen: D'oh!

environment.yml
astafan8: ..waiting for it to be dropped....

setup.py
astafan8: ...waiting for it to be gone....

setup.py
```diff
@@ -64,7 +66,9 @@ def readme():
...
+          'pyzmq',
```
astafan8: so `pyzmq` became a setup requirement? :)

environment.yml
```diff
@@ -18,4 +18,6 @@ dependencies:
...
  - pyzmq
+ - wrapt
```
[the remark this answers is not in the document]
astafan8: ok, seems to be fixed.

qcodes/config/__init__.py
astafan8: as discussed, please move to the "dataset"

qcodes/dataset/sqlite_base.py
```diff
@@ -209,10 +232,86 @@ def connect(name: str, debug: bool = False) -> sqlite3.Connection:
 
     if debug:
         conn.set_trace_callback(print)
+
+    init_db(conn)
+    perform_db_upgrade(conn, version=version)
```
astafan8: aha, so as i see, we don't modify the `_runs_table_schema` schema to have `guid` column so that `perform_db_upgrade` is called even for a fresh new database file?

qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -51,7 +57,7 @@ def test_tables_exist(empty_temp_db):
     print(qc.config["core"]["db_location"])
     conn = connect(qc.config["core"]["db_location"], qc.config["core"]["db_debug"])
     cursor = conn.execute("select sql from sqlite_master where type = 'table'")
-    expected_tables = ['experiments', 'runs', 'layouts', 'dependencies']
+    expected_tables = ['experiments', 'runs', 'layouts', 'dependencies', 'uuids']
```
astafan8: what is the `uuids` table?

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: Yes, that is considered to be the best practice; the creation is always of version 0 and then upgrades are applied. That makes testing and partial upgrades (if you encounter some old version in the wild) very easy.

qcodes/tests/dataset/test_dataset_basic.py
WilliamHPNielsen: An obsolete idea: a separate tables for the guids (called uuids back then). This should not still be in the PR, thanks for catching it.

astafan8: ok. i wonder why this hasn't been noticed when CI runs there tests...... could you have look? perhaps the test is not working?

qcodes/dataset/sqlite_base.py
astafan8: i see. makes sense.

qcodes/config/__init__.py
astafan8: also, `Config.save_config` method can be used (it contains the same code)

setup.py
Dominik-Vogel: How did you set up this bot @astafan8 ? ;-)

astafan8:
perhaps something that is missing - shall `DataSet` have a read-only `run_guid` or `guid` property (similar to `run_id`)? I think this belongs to this PR. Having this property will also allow to test that the guid has actually been written to the database. Of course, the actual usages of this property (say, `plot_by_guid`, etc.) are out of scope here.

qcodes/tests/dataset/test_dataset_basic.py
WilliamHPNielsen: So the explanation was that the `uuids` table was indeed no longer in the database, but the test did not catch that, since it did a 
```
    for row, expected_table in zip(cursor, expected_tables):
        assert expected_table in row['sql']
```
so if we had *more* expected tables than tables in the DB (i.e. rows in the cursor), the test wouldn't notice. That has been corrected now.

qcodes/config/qcodesrc.json
```diff
@@ -21,5 +21,10 @@
         "enable_forced_reconnect": false,
         "default_folder": ".",
         "default_file": "instrument_config.yml"
+    },
+    "GUID_components": {
+        "location": 0,
+        "work_station": 0,
+        "sample": 0
```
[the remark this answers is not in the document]
WilliamHPNielsen: We've been going back and forth a bit on the defaults topic. Now consensus is that we'll have invalid defaults and prompt users for location and work station.

WilliamHPNielsen:
@QCoDeS/core, thanks a lot for the reviews and input. I think all issues have been addressed now.

qcodes/dataset/sqlite_base.py
```diff
@@ -209,10 +232,86 @@ def connect(name: str, debug: bool = False) -> sqlite3.Connection:
...
+    Args:
+        version: Which version to upgrade to. We count from 0. -1 means
+          'latest version'
+    """
+
+    upgrade_actions = {1: (perform_db_upgrade_0_to_1,),
+                       -1: (perform_db_upgrade_0_to_1,)}
+
+    current_version = get_user_version(conn)
+    if current_version < 1 and (version > 0 or version == -1):
```
jenshnielsen: This does not seem very general and directly extendable. Perhaps something like this would be simpler

```
upgrade_actions = {1: (perform_db_upgrade_0_to_1,)}
max_target_version = 1
if version == -1:
    version = max_target_version 

current_version = get_user_version(conn)
if current_version < version:
        log.info("Commencing database upgrade")
        for action in upgrade_actions[version]:
            action(conn)

```

WilliamHPNielsen: Sold!

docs/dataset/introduction.rst
```diff
@@ -0,0 +1,69 @@
...
+
+The **name** and the **sample name** of an experiment must be provided by the user at creation time.
+
+The **start time** is automatically added upon creation, and the **end time** is added when (if ever) the user decides to mark the experiment as *completed* after which point no more runs may be added.
+
+For an example notebook showing the useage of the database as a container for experiments, see the :doc:`Experiment Container Notebook <../_notebooks/DataSet/example-creation>`.
+
+The intended use of the ``Experiment`` is to group ``DataSets`` that belong together. For example, an experiment may contain all the measurements necessary to characterise a particular sample (Is it superconducting? What is the critical current?). Once the goodness of the sample has been established, a new experiment containing all the "real" measurements can then commence. The grouping runs this way is offered as a convenience and is as such not essential for any functionality in QCoDeS.
```
astafan8: perhaps, move this above the description of the attributes?

qcodes/dataset/guids.py
```diff
@@ -0,0 +1,125 @@
...
+    components = {}
+    components['sample'] = int(guid[:8], base=16)
+    components['location'] = int(guid[8:10], base=16)
+    components['work_station'] = int(guid[10:16], base=16)
+    components['time'] = int(guid[16:], base=16)
+
+    return components
+
+
+def set_guid_location_code() -> None:
```
astafan8: does it make sense to also add a `set_guid_sample_code`? it will be very handy for users who would like to save the sample code in their config.

qcodes/dataset/sqlite_base.py
```diff
@@ -92,7 +95,25 @@
...
+    us to nest atomic context managers
+    """
+    atomic_in_progress: bool = True
+
+
+SomeConnection = Union[sqlite3.Connection, ConnectionPlus]
+
+
+# this should really return sqlite3.Binary,
+#  but there seems to be a bug in mypy 0.590
```
astafan8: it seems that this comment can now be removed, right?

qcodes/dataset/sqlite_base.py
```diff
@@ -209,10 +236,88 @@ def connect(name: str, debug: bool = False) -> sqlite3.Connection:
...
+
+    upgrade_actions = {0: (lambda x: None,),
+                       1: (perform_db_upgrade_0_to_1,)}
+    newest_version = max(upgrade_actions.keys())
+    version = newest_version if version == -1 else version
+
+    current_version = get_user_version(conn)
+    if current_version < newest_version:
+        log.info("Commencing database upgrade")
+        for action in upgrade_actions[version]:
```
astafan8: the upgrades need to be executed in the correct order, right? do we rely here on the developers to list `upgrade_actions` in correct order from `0`? If so, please, add a comment to `upgrade_actions` about it. Otherwise, we need to sort ascending before looping, right?

qcodes/dataset/sqlite_base.py
```diff
@@ -1363,29 +1507,113 @@ def add_meta_data(conn: sqlite3.Connection,
...
+        work_station = cfg['GUID_components']['work_station']
+    except KeyError:
+        log.warning('No work_station information found in the config. '
+                    'Can not proceed.')
+        return
+
+    if location == 0:
+        log.warning('The location is still set to the default (0). Can not '
+                    'proceed. Please configure the location before updating '
+                    'the GUIDs.')
```
astafan8: missing `return` here?

qcodes/dataset/sqlite_base.py
```diff
@@ -1363,29 +1507,113 @@ def add_meta_data(conn: sqlite3.Connection,
...
+    if work_station == 0:
+        log.warning('The work_station is still set to the default (0). Can not'
+                    ' proceed. Please configure the location before updating '
+                    'the GUIDs.')
```
astafan8: missing `return` here?

qcodes/dataset/sqlite_base.py
```diff
@@ -1363,29 +1507,113 @@ def add_meta_data(conn: sqlite3.Connection,
...
+    try:
+        location = cfg['GUID_components']['location']
+    except KeyError:
+        log.warning('No location information found in the config. '
+                    'Can not proceed.')
+        return
+    try:
+        work_station = cfg['GUID_components']['work_station']
+    except KeyError:
+        log.warning('No work_station information found in the config. '
```
astafan8: if config does not even have `GUID_components`, then the warning message will not be completely correct. Shall it be adjusted to smth like `GUID_components -> work_station`?

astafan8: for `location` (above) as well?

qcodes/dataset/sqlite_base.py
```diff
@@ -1363,29 +1507,113 @@ def add_meta_data(conn: sqlite3.Connection,
...
+    c = atomic_transaction(conn, query)
+    no_of_runs = c.fetchall()[0][0]
+
+    # now, there are four actions we can take
+
+    def _both_nonzero(run_id: int, *args) -> None:
+        log.info(f'Run number {run_id} already has a valid GUID, skipping.')
+
+    def _location_only_zero(run_id: int, *args) -> None:
+        log.warning(f'Run number {run_id} is has a zero (default) location '
```
astafan8: typo: "is has"

qcodes/dataset/sqlite_base.py
```diff
@@ -1363,29 +1507,113 @@ def add_meta_data(conn: sqlite3.Connection,
...
+                    'code, but a non-zero work station code. Please manually '
+                    'resolve this, skipping the run now.')
```
astafan8: what is the way to manually resolve this? Can it be written to the warning message or to the docstring at least?

qcodes/dataset/sqlite_base.py
```diff
@@ -1363,29 +1507,113 @@ def add_meta_data(conn: sqlite3.Connection,
...
+
+    def _workstation_only_zero(run_id: int, *args) -> None:
+        log.warning(f'Run number {run_id} is has a zero (default) work station'
```
astafan8: typo: "is has"

qcodes/dataset/sqlite_base.py
```diff
@@ -1363,29 +1507,113 @@ def add_meta_data(conn: sqlite3.Connection,
...
+                    ' code, but a non-zero location code. Please manually '
+                    'resolve this, skipping the run now.')
```
astafan8: I guess, the way to resolve this manually is the same as above?

qcodes/dataset/sqlite_base.py
```diff
@@ -1363,29 +1507,113 @@ def add_meta_data(conn: sqlite3.Connection,
...
+                   UPDATE runs
+                   SET guid = ?
+                   where run_id == {run_id}
+                   """
+            cur = conn.cursor()
+            cur.execute(sql, (guid_str,))
+
+        log.info(f'Succesfully updated run number {run_id}.')
+
+    actions: Dict[Tuple[bool, bool], Callable]
```
astafan8: `Callable[int, SomeConnection, Dict]`? or there is no reason to add this level of details?

qcodes/dataset/sqlite_base.py
```diff
@@ -1363,29 +1507,113 @@ def add_meta_data(conn: sqlite3.Connection,
...
+    if location == 0:
+        log.warning('The location is still set to the default (0). Can not '
+                    'proceed. Please configure the location before updating '
+                    'the GUIDs.')
+    if work_station == 0:
+        log.warning('The work_station is still set to the default (0). Can not'
+                    ' proceed. Please configure the location before updating '
+                    'the GUIDs.')
+
+    query = f"select MAX(run_id) from runs"
```
astafan8: do i understand correctly that since `run_id` is an integer PRIMARY KEY, there are no missing run IDs between 1 and MAX(run_id)?

qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -358,18 +387,115 @@ def test_dataset_with_no_experiment_raises(empty_temp_db):
...
-    userversion = get_user_version(connection)
-    if userversion != 0:
-        raise RuntimeError("trying to upgrade from version 0"
-                           " but your database is version"
-                           " {}".format(userversion))
-    sql = 'ALTER TABLE "runs" ADD COLUMN "quality"'
+def test_guid(dataset):
+    guid = dataset.guid
+    assert len(guid) == 36
+    parse_guid(guid)
```
astafan8: shall we test here that the `generate_guid` call that was used when dataset was created, used current values of location, sample name, workstation from the config, and the timestamp is "fairly recent/reasonable"?

docs/dataset/introduction.rst
```diff
@@ -0,0 +1,69 @@
...
+
+The **start time** is automatically added upon creation, and the **end time** is added when (if ever) the user decides to mark the run as *completed* after which point no more data points may be added.
+
+The **GUID** is a run identifier that aims to go `beyond` the current database. The motivation for the GUID is that collaboration and data sharing across several machines requires a run to be labelled by something more than its name and number in the local database. The GUID is a standard 36 character string GUID composed from three integer codes and a timestamp. The codes are: location code (1-256), work station code (1-16777216), sample code (1-4294967296). The idea is that the people in a certain collaboration will label their locations, machines, and samples according to a scheme agreed upon `within` the relevant collaboration. That way, each run is uniquely identifiable in a human-readable way. This extra label for runs is offered as a convenience as is as such not essential for any functionality in QCoDeS.
+
```
astafan8: do we need to document `generate_guid`, `parse_guid`, `set_guid_*`, `update_GUIDs` functions anywhere for someone?

i'm just checking. i thing that these should only be known to developers, and should not be used by users.

qcodes/dataset/guids.py
WilliamHPNielsen: I'd rather wait with that one out of fear of "our" people using it. It should be straightforward for any hacker to add.

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: Well, the iteration is over the tuple that is the value of `upgrade_actions` for the key `version`. I can add a comment that `2` must have `(perform_db_upgrade_0_to_1, perform_db_upgrade_1_to_2)` if that is not clear from the context.

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: Thank you!

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: It's not wrong; that information is not in the config. But it can actually never happen, since we ship QCoDeS with those settings.

qcodes/dataset/sqlite_base.py
astafan8: oh, sorry, i missed the tuple part...

then the order within the tuple matters, right? i think this should be documented in the comments at least.

I feel that it is annoying and redundant to add two methods `(perform_db_upgrade_0_to_1, perform_db_upgrade_1_to_2)` for an upgrade to `2`. I also do not expect that for moving from `0` to `2` you'll need the two mentioned methods, but for moving from `0` to `3`, you'll need a different list of update functions, right?

I initially thought that each key has a single function assigned, which makes the upgrade from the *previous* version. This is actually also reflected in how the functions are currently called. The one-key-one-function approach also reflects perfectly the design: that all the upgrades shall be made sequentially, for example, an upgrade from 0 to 2 implies an upgrade from 0 to 1 and then from 1 to 2. What is the idea behind the dict with tuples?

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: Short answer: there is no straight-forward way to do that. If you encounter this message in the field, you better go ask a "grown-up". In our organisation, that would be someone like you or me.

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: `Dict[Tuple[bool, bool], Callable]` is enough to make mypy shut up 😉

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: Correct. If you have missing runs, your database is fouled up. That is not possible to achieve with our API, so we don't consider it an option.

qcodes/tests/dataset/test_dataset_basic.py
WilliamHPNielsen: No, I think it is sufficient to have that tested in `test_guids.py`

.appveyor.yml
jenshnielsen: Is this still needed?

WilliamHPNielsen: I suppose it shouldn't be, no. Let me try to remove it.

PR #1003, opened 2018-03-13
Make alazar drivers more standard compliant

jenshnielsen:
* Use regular parameters
* Deprecate special AlazarParamers
* makes it possible to set all paramers as any other qcodes paramers but add a flag to keep track of if they are up to date with the card.
* Add sync to card method

@Dominik-Vogel


jenshnielsen:
I think this is ready to land, as above it does:

* Use regular parameters
* Deprecate special AlazarParamers
* makes it possible to set all paramers as any other qcodes paramers but add a flag to keep track of if they are up to date with the card.
* Add sync to card method

In addition:

* Make sure that the code still works with Old drivers that may be outside the main repo. Mainly this is to support AlazarParameters in addition to regular parameters
* Add some types

jenshnielsen:
@Dominik-Vogel @sohailc @damazter Any opinion on this

@ThorvaldLarsen If you have time to check this out it would be great

Dominik-Vogel:
Let me try that out in practice. I like the new (inversed) dictionaries, even though it feels like it should have been like this from the beginning without the workaround you implemented.
I am really happy to have qcodes parameters for the Alazar available with this PR!
(`_set_or__set_set_or__set` is the worst function name I have ever encounterd ;-) )

jenshnielsen:
That is wrong, It should actually just have been `_set_or__set`

damazter:
I don't understand how this would work with the normal parameters.
so with `self._get_raw_or_bytes(` it works as before, but with the normal parameters, how do you differentiate between the byte_code and the string you use to identify it?

Secondly, it looks like you only look at parameters synced as a complete group of all parameters, which is not something I like. There are stages in the code where some parameters are already loaded into the board and others are not. The previous implementiation garantueed an error on such occasions.
I don't see how that works now, but maybe I am missing it.
but this code does not give enough security i think:
```python
+        if self._parameters_synced == False:
+            raise RuntimeError("You must sync parameters to Alazar card "
+                               "before calling acquire by calling "
+                               "sync_parameters_to_card")
```
What is wrong with subclassing the parameter to enforce this behaviour?

I didn't get the difference between:
Add sync to card **method**
and the old config method, it feels like there should only be one?

jenshnielsen:
> I don't understand how this would work with the normal parameters.
> so with self._get_raw_or_bytes( it works as before, but with the normal parameters, how do you > differentiate between the byte_code and the string you use to identify it?

Please look again, this gets `raw_value`. For QCoDeS parameters this is the value sent to the instrument where as `get` returns the human readable value.

> Secondly, it looks like you only look at parameters synced as a complete group of all parameters, 
> which is not something I like. There are stages in the code where some parameters are already 
> loaded into the board and others are not. The previous implementiation garantueed an error on such 
> occasions.

> I don't see how that works now, but maybe I am missing it.
> but this code does not give enough security i think:

Care to explain how? The only thing that matters is that parameters are in sync once you call acquire.
If the card is out of date with the driver at any other time it can never be reflected in data.
It practice the old system was *broken*. the code was filled with `_set_update` and to work around the inability to call get.

> What is wrong with subclassing the parameter to enforce this behaviour?

I don't mind subclassing and doing the check there too during the acquire if that makes you feel more secure but doing it on get in general has been found to be very confusing and counter productive.

> I didn't get the difference between:
> Add sync to card method
> and the old config method, it feels like there should only be one

Please read again. The config method calls the sync_settings_to_card. It's just splitting the 2 functions of that thing into two methods so that you can now set parameters in a normal way and then sync them without the pain of the config method. The config method is only retained for backwards compatibility and IMHO it should eventually be removed. I tried writing this so that it still works with the old alazar parameter

damazter:
Thanks for your response, I understand `sync_settings_to_card` better now.
I think most importantly, regarding 

> Care to explain how? The only thing that matters is that parameters are in sync once you call acquire.
If the card is out of date with the driver at any other time it can never be reflected in data.
It practice the old system was broken. the code was filled with _set_update and to work around the inability to call get.

The thing I wanted to protect the user from is getting the value of a parameter which is not synced to the board. This is not only a risk when calling acquire (although an important one) but also when logging results. Or sometimes I read values from the alazar to determine how my acquisition should run. If i forgot to call config, my acquisition is screwed up. Keeping track per parameter and forbidding a get of the parameter until it is synced saved me from some errors in the past

jenshnielsen:
But surely only the values when calling acquire matters for actual data. If you are logging at any point outside acquire and correlating that metadata to actual data it will in general be wrong irrespective of this change.

jenshnielsen:
But we could make acquire sync the parameters if needed

ThorvaldLarsen:
I agree that syncing only matters for acquisition. At any other point in time what is actually on the card is irrelevant as long as we know if they are up to date or not. I like the proposed version here with the flexibility added by using normal parameters.

@damazter I am not sure I understand your concern. Maybe you can elaborate a bit more? 
As I see it if you are asking for parameters to set up a run then it is fine that these are in software only. If you then forget to sync with card before measuring the current driver will catch  that and make you sync before running. So no metadata/logging would be wrong.

jenshnielsen:
> There are two things which I can think of off the top of my head that the read-protection warns you for.
> having a read command and a config command in the wrong order (important if you need to reconfig > during your measurement)

This is already taken care of by having acquire calling sync or just will indeed fall once the global variable is checked

> Secondly, depending on the mode you run in, some parameters are not used. because the red- >protection flag is then not removed, you will automatically see that you are doing something wrong.

Sure we can still implement that by tracing each parameter as I said above. Currently the driver is setting a whole bunch of these flags manually so this is not really true anyway.

> Furthermore, because the acquisition_controller allows for arbitrary code execution in the middle of > the acquire method, code debugging can already be quite hard.

I still don't see how this solves anything at all in that respect.

> So in summary, I don't see the advantage of dropping this feature and it might prevent some weird  errors in the future

Every single user that I have interacted with that have used the Alazar has found the config method confusing and super non intuitive because it's different from every other parameter in qcodes. This is a significant barrier to entry. Furthermore we increasingly rely on setting paramters via instrument data files in Copenhagen and that does just not work well with requiring the config function.  

That being said. It will stay in the code and you can keep using it. But I think we should strongly consider replacing it with a context manager. Something along the lines of 

```python
with alazar.sync_at_exit():
    alazar.samplerate(1_000_000)
    alazar.triggerlevel1(128)
```
Would be much more pythonic and much cleaner

damazter:
> Every single user that I have interacted with that have used the Alazar has found the config method confusing and super non intuitive because it's different from every other parameter in qcodes.

With this I see your point, and changing the setting of these parameters instead of the config method seems like a good idea.
Something should still decide when these values are written to the card however, and you don't want to do that on every parameter update, but this is what the sync method/context manager would be for.

But I think we might have a bit of a miscommunication:
changing config to sync: great idea. was not possible when I wrote the driver, but is now more natural in the new version of qcodes, so I am all for it.
Removing the alazar parameter and removing the syncflag per parameter, is the point that i do not agree with. Keeping this feature does not interfere with changing `config` to `sync`, and the feature does protect from some weird programming errors.

>That being said. It will stay in the code and you can keep using it.

I don't know if this refers to config (in that case, ignore what is below) or the flag_check. The flag check per parameter is removed and there are examples where an error should be thrown, but isn't with the proposed change. Although the examples that i can come up with right now are very constructed and weird

damazter:
> Currently the driver is setting a whole bunch of these flags manually so this is not really true anyway.

True, the way the flags were managed in the driver was not the most beautiful, but because this is in the driver file, it only has to be programmed in that driver, and doesn't interfere with people using it. I wouldn't say it is manual though, because now we only have to make sure the driver is correctly setting the flags, and then this code will prevent the end-user making mistakes with getting and setting these parameters

jenshnielsen:
@damazter great. As I wrote already in the second comment i have no problem with reintroducing the per parameter up to date flag

jenshnielsen:
The TraceParamter now track it's synced status in the same way as the Alazar parameter. I have added a context manager as a replacement for the (now) deprecated config method

ThorvaldLarsen:
@jenshnielsen this PR works well together with the channelized controllers for the measurements we are doing on T3.

PR #1004, opened 2018-03-14
Driver/awg70000 add subsequences

WilliamHPNielsen:
- Add new functions taking forged sequences from broadbean
- Make seqx files from forged sequences with subsequence support

https://github.com/QCoDeS/broadbean/pull/60 should be merged in before this PR makes much sense.

Note that this PR will cause QCoDeS (at least the AWG70000A driver) to depend on broadbean.

@Dominik-Vogel @jenshnielsen


WilliamHPNielsen:
I did some preliminary testing with an actual instrument, and it works nicely. I'd like to add a notebook with a complicated sequence, but I am not sure where that notebook belongs. Here or in broadbean's `docs` folder?

WilliamHPNielsen:
Okay, I added the notebook here. This PR is now ready for review.

WilliamHPNielsen:
@QCoDeS/core 👍 / 👎 ❓

Dominik-Vogel:
There are many 🐫 💼  that ran from bb to qcodes.

sohailc:
To be very honest, I would not have let the driver depend on broadbeans. Instead, I would have transferred the checking/assertion functionality from broadbeans to the driver.

WilliamHPNielsen:
@sohailc, is that not what we have done? No broadbean object is used in the driver, only the `fs_schema` for validation. The dependence is in an installation/import sense. Do you have something else in mind?

@Dominik-Vogel, true, I'll see to that.

sohailc:
@WilliamHPNielsen Let me be more explicit what I mean: make a class method in the AWG70000A driver class called "validate_schema" accepting a dict as input. In broadbeans call this class method in the "_plotSequence" method of the "Sequence" class. This will make broadbeans dependent on QCoDeS, which I think is more desirable than the other way around.

WilliamHPNielsen:
@sohailc As discussed on skype, we are going with a layout where broadbean is supposed to be hardware-agnostic. It thus makes most sense to make the QCoDeS drivers anticipate the output of broadbean than vice versa. Also, we plan to promote broadbean to be **the** pulse builder for QCoDeS, so that should be okay.

PR #1006, opened 2018-03-16
make plot by id take a axes argument for easier embedding

jenshnielsen:
@WilliamHPNielsen @jana-d  

This also adds support for more than one dataset per runid


WilliamHPNielsen:
Looks like a good improvement. Looking forward to seeing the updated examples.

jenshnielsen:
@WilliamHPNielsen refreshed the notebooks

qcodes/dataset/plotting.py
```diff
@@ -45,22 +48,31 @@ def set_axis_labels(ax, data):
...
+        axes = [axes]
+
+    if axes is None:
+        axes = []
+        for i in range(nplots):
+            fig, ax = plt.subplots(1,1)
+            axes.append(ax)
+    else:
+        if len(axes) != nplots:
+            raise RuntimeError(f"Trying to plot {nplots} but "
```
WilliamHPNielsen: Consider the rewording `f"Trying to make {nplots} plots, but received {len(axes)} {ax_word}."`
Where `ax_word = 'axis' if len(axes) == 1 else 'axes' :)

jenshnielsen: Agreed apart from the last part. In matplotlib  API an Axes object is the thing that holds both the x and y axis, ax axis is either the x or y axis. That is confusing but the way it works

PR #986, opened 2018-03-01
Simplify monitor websocket use

jenshnielsen:
Should make it compatible with the latest release of websockets python package and fixes issues with restarting the monitor

@Dominik-Vogel

PR #1007, opened 2018-03-16
we now support websockets 4 too

jenshnielsen:
@QCoDeS/core


jenshnielsen:
because #986 has landed

Issue #1008, opened 2018-03-19
Monitor does not close properly when run via ipython runfile (as in spyder)


Dominik-Vogel:
Running:
```
from qcodes.monitor.monitor import Monitor
Monitor()
```
via runfile/F5 twice in spyder produces:
```
Traceback (most recent call last):
  File "C:\Users\a-dovoge\anaconda3\lib\asyncio\base_events.py", line 1044, in create_server
    sock.bind(sa)
OSError: [WinError 10048] Only one usage of each socket address (protocol/network address/port) is normally permitted

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\a-dovoge\anaconda3\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "c:\users\a-dovoge\qcodes\qcodes\monitor\monitor.py", line 124, in run
    self.server = self.loop.run_until_complete(server_start)
  File "C:\Users\a-dovoge\anaconda3\lib\asyncio\base_events.py", line 467, in run_until_complete
    return future.result()
  File "C:\Users\a-dovoge\anaconda3\lib\site-packages\websockets\server.py", line 365, in serve
    server = yield from loop.create_server(factory, host, port, **kwds)
  File "C:\Users\a-dovoge\anaconda3\lib\asyncio\base_events.py", line 1048, in create_server
    % (sa, err.strerror.lower()))
OSError: [Errno 10048] error while attempting to bind on address ('127.0.0.1', 5678): only one usage of each socket address (protocol/network address/port) is normally permitted
```

jenshnielsen:
Closed by #1009

PR #1009, opened 2018-03-19
Spyder exclude qcodes from umr

jenshnielsen:
@Dominik-Vogel closes https://github.com/QCoDeS/Qcodes/issues/1008


jenshnielsen:
Used the solution in https://stackoverflow.com/questions/17728395/detect-where-python-code-is-running-e-g-in-spyder-interpreter-vs-idle-vs-cm to detect spyder not great but should be better

WilliamHPNielsen:
Why is this solution "not great"? Because other environment variables might be called 'SPYDER_lalala'?

jenshnielsen:
Yes or SPYDER may stop using the env variable name. Spyder basically uses env variables for lightweight Inter process communication between the gui and the ipython kernel

jenshnielsen:
If i am a good open source citizen I should really submit a patch to Spyder to add an api for this

WilliamHPNielsen:
I see, thanks for explaining that. I don't think anyone would mind you being a good citizen.

But I guess this solution is OK for now.

jenshnielsen:
This does not seem to be an issue but rather that stuff that has not actually been imported will be marked as reloaded for some reason

PR #1010, opened 2018-03-19
add missing channel to keithley 2600 resistance 

jenshnielsen:
This was missed when converting from non channel version that injects this in  a different way. Matches V and I


jenshnielsen:
@QCoDeS/core

PR #1012, opened 2018-03-20
raise warning level for parameters that cannot be updated

jenshnielsen:
This makes it easier to spot issues such as #1010 (and another one that I will push a fix for shortly)


jenshnielsen:
This would also have worked for #1013

jenshnielsen:
I was convinced it was this ◀️  idiot but anyway

jenshnielsen:
I think it is correct to catch all because we don't want to stop qcodes due to any failed snapshot update but we may want to handle different exceptions in a different way

jenshnielsen:
@WilliamHPNielsen Changed it a bit to add the exception info to the file based logger

jenshnielsen:
@WilliamHPNielsen as requested

WilliamHPNielsen:
Looks good.

PR #1013, opened 2018-03-20
SR860: Add snapshot_base to prevent update of parameter that may not be defined

jenshnielsen:
This parameter only works after a complete capture so any automatic update will most likely fail. This cuts 5 sec off the time needed to add this to the station


qcodes/instrument_drivers/stanford_research/SR86x.py
```diff
@@ -142,6 +143,19 @@ def __init__(self, parent: 'SR86x', name: str) ->None:
         self.bytes_per_sample = 4
         self._capture_data = dict()
 
+    def snapshot_base(self, update: bool = False,
+                      params_to_skip_update: Sequence[str] = None) -> Dict:
+        if params_to_skip_update is None:
+            params_to_skip_update = []
+        # we omit count_capture_kilobytes from the snapshot because
+        # it can only be read after a completed capture and will
+        # timeout otherwise when the snapshot is updated, I.e at
```
WilliamHPNielsen: i.e. -> e.g. (I would say)

jenshnielsen: Done

PR #1014, opened 2018-03-20
Remote plot fail

mltls:
Spawn a new QT process if the old one died

@jenshnielsen


jenshnielsen:
This is really great. These is one problem that I didn't think about, Until now pyqtgraph has been an optional dependency of qcodes. That means that we must either wrap all of this in conditionals or make it a required dependency. @QCoDeS/core any opinion

WilliamHPNielsen:
I cast my vote in the "make it a required depency" urn.

qcodes/loops.py
```diff
@@ -891,11 +894,20 @@ def _run_loop(self, first_delay=0, action_indices=(),
...
                         self.bg_task()
-                    except Exception:
+                        self.last_task_failed = False
+                    except (ClosedError, ConnectionResetError) as err:
+                        # old qt process dead, we spawn a new
+                        log.error("Remote QT process "
+                                  "crashed, restarting:")
+                        log.exception(err)
+                        if isinstance(self.bg_task.__self__, QtPlot):
+                            self.bg_task.__self__._init_qt()
```
jenshnielsen: We should probably add an else clause here that will re throw the error so that it's handled by  the usual error handler if its not a QTPlot

WilliamHPNielsen:
@mltls you had a local test that failed 1/10 of the times it plotted and was caught by the code you added here. Would it be a hassle to add that test to the test suite?

jenshnielsen:
@WilliamHPNielsen we could probably do that by adding it as another background task or regular qcodes task

jenshnielsen:
@mltls I think the reason this doesn't work is that we only replace the qt process but not the window (self.win) which is attached to the process. In the other restart path https://github.com/QCoDeS/Qcodes/blob/master/qcodes/plots/pyqtgraph.py#L98 we also replace the window

jenshnielsen:
The qt plot is considered legacy at this stage and we do not expect to do further development on this

PR #1015, opened 2018-03-20
feat/file logging level

Dominik-Vogel:
Introduce file logging level into qcodesrc. This value is used in the qdev_wrappers repository for automized logging and might be migrated into core at some point.


qcodes/config/qcodesrc_schema.json
```diff
@@ -24,6 +24,18 @@
                         "DEBUG"
                     ]
                 },
+                "file_loglevel" :{
+                    "type" : "string",
+                    "description": "control file logging  level",
```
jenshnielsen: Just to be super nitpick there is an extra space here

Dominik-Vogel: That is a reminder of the incredible fast coding from Giulio ;-). I just checked, it has been copy and pasted to almost every entry of the config, because nobody just writes jason by hand....

jenshnielsen: :) Who is this Jason and does he have the a golden fleece

PR #1016, opened 2018-03-20
Feature/Make database file dynamically selectable

WilliamHPNielsen:
<del>We need some sort of explanation of the experiment container. Here's a brief one.</del>

Writing that notebook, I realised that we probably want to not settle on a database file at import time. Therefore, this PR now aims to make the database file more dynamically selectable, i.e. you can now select different database files during a single session. 

Still pending:

- [ ] Update the tests to reflect this

@jana-d @QCoDeS/core


sohailc:
Nice notebook! I would however also add how to add data to a dataset. I am sure you can find that elsewhere, but it would be nice to have that in this notebook as well

WilliamHPNielsen:
@sohail I think that we should try to sync our notebooks once we have the sweep module in (especially regarding how to add data, where we'll have three layers of API), but that is somewhat beyond the scope of this PR, so I just added a lazy reference to another notebook about how to add data.

@QCoDeS/core now this PR took a turn into a different direction, I'd like opinions. I think the tests got nicer.

qcodes/tests/dataset/test_datasaver.py
```diff
@@ -4,26 +4,21 @@
...
 from qcodes.dataset.measurements import DataSaver
 from qcodes.dataset.sqlite_base import connect, init_db
+from qcodes.dataset.database import initialise_database
 
 CALLBACK_COUNT = 0
 CALLBACK_RUN_ID = None
 CALLBACK_SNAPSHOT = None
 
-# These fixture can't be imported from test_dataset_basic because Codacy/PR Quality Review will think they are unused.
+# These fixture can't be imported from test_dataset_basic because
```
jenshnielsen: This is a very bad argument. We should never write uglier code to make a linter happy

jenshnielsen: But I know its not related to this pr

jenshnielsen:
@WilliamHPNielsen  Can't seem to push to this branch so I made a pr against it

WilliamHPNielsen:
@jenshnielsen, so, all good now?

PR #1019, opened 2018-03-21
Make fgen_signalpath model dependent

WilliamHPNielsen:
- introduce a dictionary (probably the first of many) with val mappings for the `fgen_signalpath` parameter

Tested with an actual AWG5208 using `print_readable_snapshot(update=True)`.

@jenshnielsen

### DIFF
--- a/qcodes/instrument_drivers/tektronix/AWG70000A.py
+++ b/qcodes/instrument_drivers/tektronix/AWG70000A.py
@@ -16,6 +16,17 @@ from qcodes.utils.validators import Validator
 
 log = logging.getLogger(__name__)
 
+# some settings differ between series
+_fg_path_val_map = {'5208': {'DC High BW': "DCHB",
+                             'DC High Voltage': "DCHV",
+                             'AC Direct': "ACD"},
+                    '70001A': {'direct': 'DIR',
+                               'DCamplified': 'DCAM',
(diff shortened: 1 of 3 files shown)


qcodes/instrument_drivers/tektronix/AWG70000A.py
```diff
@@ -122,10 +133,7 @@ def __init__(self,  parent: Instrument, name: str, channel: int) -> None:
                            label='Channel {} {} signal path'.format(channel, fg),
                            set_cmd='FGEN:CHANnel{}:PATH {{}}'.format(channel),
                            get_cmd='FGEN:CHANnel{}:PATH?'.format(channel),
-                           val_mapping={'direct': 'DIR',
-                                        'DCamplified': 'DCAM',
-                                        'AC': 'AC'}
-                           )
+                           val_mapping=_fg_path_val_map[self._parent.model])
```
jenshnielsen: self._parent can now be replaced by the root instrument property but that it not essential

WilliamHPNielsen: Oh yes, that's right! Let's start using it.

WilliamHPNielsen:
Now I replaced `_parent` everywhere.

PR #1020, opened 2018-03-21
Build with latest Sphinx

jenshnielsen:
Now that the bug is fixed in Sphinx we no longer need to use a beta version


jenshnielsen:
@QCoDeS/core

PR #1022, opened 2018-03-22
dont depend on ZMQ in general

jenshnielsen:
Avoid pulling in zmq into qcodes core


WilliamHPNielsen:
Seems like the failing notebook isn't really using the Publisher, so that should be fine.

But why not depend on ZMQ? Just to have as few dependencies as possible?

jenshnielsen:
Yes that is the other obvious solution, I don't have a strong preference either way

jenshnielsen:
closed in favor of #1028

PR #1026, opened 2018-03-28
Change default warning level for stream output

jenshnielsen:
This should not be verbose by default

@Dominik-Vogel this may be why users have had the need to change this locally


jenshnielsen:
Also pip pyvisa in env by default to not get annoying warning

PR #1021, opened 2018-03-22
Add colorbar support to dataset

jenshnielsen:
@WilliamHPNielsen @jana-d

PR #1031, opened 2018-04-06
fix simple typo in #1021

jenshnielsen:


jenshnielsen:
And a line length < 79 too

PR #1032, opened 2018-04-06
Make AWG5208 support four markers

WilliamHPNielsen:
Quick repair of AWG5208 driver

- Increase the number of markers per channel to 4
- Change the allowed channel resolutions from (8, 9, 10) to (12, 13, 14, 15, 16)

I suspect that the `.wfmx` and `.seqx` file generation may still have to be updated.

@wpfff


sohailc:
Looks good to me!

PR #1037, opened 2018-04-10
CI/run CI on Windows machines with AppVeyor

WilliamHPNielsen:
It is a good idea to run CI on all major platforms. And all experiments run on Windows machines, so we should definitely test on Windows.

- Add .appveyor.yml file
- Add notebook execution script

@QCoDeS/core


jenshnielsen:
Looks good but I cant see the report from AppVeyor

sohailc:
I suggest that we merge PR 1038 first as it will solve the problem of failing dataset tests you are having now.

WilliamHPNielsen:
@QCoDeS/core, look, the tests don't fail! I wrote a yaml file in only 15 commits 😛

PR #1038, opened 2018-04-10
the test_measurement_context_manager now passes

sohailc:
This fixes the test: test_measurement_context_manager


sohailc:
This PR is intended to make the dataset tests pass on windows. Specifically, an issue on windows is that closing the experiment connection to the database will not close all connections to the database, as the dataset instance has an independent connection. Since on Windows we cannot remove a file which is still accessed by a process, this connection will cause tests which use the "experiment" fixture to potentially fail. This is because calling "e.conn.close()" will not close *all* connections. 

A solution is to make the dataset and the experiment use the same connection.

sohailc:
I don't understand why I am seeing AppVeyor here. Should that not get merged in with PR 1037?

jenshnielsen:
Appveyor has been activated and will run regardless of the state of the Appveyor config file. But will naturally fail. There is a `Skip branches without appveyor.yml` config option but its not activated by default

jenshnielsen:
I have activated it now

qcodes/dataset/data_set.py
```diff
@@ -637,7 +641,7 @@ def load_by_counter(counter, exp_id):
     return d
 
 
-def new_data_set(name, exp_id: Optional[int] = None,
+def new_data_set(name, exp_id: Optional[int] = None, conn=None,
```
WilliamHPNielsen: I guess any new keyword argument as a good rule of thumb should be added after the already existing ones.

WilliamHPNielsen:
I think this is a sane, non-intrusive fix. If I understand things correctly, we could also explicitly close the connections in the tests, but I don't think it hurts having the option to force a DataSet to use a specific connection.

I left one inline comment, apart from that I vote: 👍

PR #1039, opened 2018-04-11
feat: Add Keysight B2962A instrument driver

neildick:
This adds a instrument driver for the Keysight B2962A power source.
It supports setting voltages, currents, limits and performing
measurements on both channels.

- Adds a new instrument driver 

@jenshnielsen


neildick:
Hi @jenshnielsen thanks for having a look :)
I've been doing a little bit of testing with it here in Sydney and it works with the instrument we have here.

PR #1044, opened 2018-04-11
Make it possible to run tests from python

jenshnielsen:
* Add top level function to run tests
* Bundle various files needed for tests in the installer
* Make simulations reset the state of instruments (pyvisa-sim remembers the state across closing of the qcodes instrument)
* Add a missing function to the simulated awg (which previously would generate an error on the second run)
* Close instrument in a different test after use
* Add init files so dataset and simulated instrument tests are correctly bundled into the package

@QCoDeS/core


jenshnielsen:
@QCoDeS/core I think this is ready to land

PR #1045, opened 2018-04-11
Add option to remove channels from ChannelList

Dominik-Vogel:
Tested and seems to not cause problems - travis test fails... It seems to be a mypy issue. Cannot know if it is a tuple or not.


Dominik-Vogel:
Hang on I just noticed the test that is not passing here...

Dominik-Vogel:
Thank you @jenshnielsen, this is a great suggestion!

jenshnielsen:
By casting to an array l mean something like
```
channels = cast(List[InstrumentChannel], self._channels)
``` 
Which does not mention array at all, sorry for the confusion

qcodes/instrument/channel.py
```diff
@@ -263,6 +263,18 @@ def append(self, obj: InstrumentChannel):
         self._channel_mapping[obj.short_name] = obj
         return self._channels.append(obj)
 
+    def remove(self, obj: InstrumentChannel):
```
astafan8: no testing?

jenshnielsen: It should probably have a test but was not ready to merge anyway. Somehow got pushed to master anyway by accident with failing test

Dominik-Vogel: I added extensive testing and this should be good to go now, @jenshnielsen, @astafan8

astafan8: 👍

PR #1049, opened 2018-04-13
Dataset allow 0d np arrays to be inserted with context manager

jenshnielsen:
@WilliamHPNielsen @jana-d


jenshnielsen:
The idea is to allow numpy arrays with zero dimensions such as `np.array(1)` to be added by casting them to 1D

jenshnielsen:
Appveyor bug fixed by #1062

PR #1056, opened 2018-04-16
offset option for parameter

Dominik-Vogel:
This PR adds the option to add an offset to a parameter in the same way as one can add a scale.
Any idea why codacy does not recognize the argument?


jenshnielsen:
Seems reasonable, I would like to keep the coverage as high as possible for that code since its somewhat convoluted

Issue #1057, opened 2018-04-17
Dataset Context Manager notebook fails when executed


WilliamHPNielsen:
The `Dataset Context Manager.ipynb` notebook randomly fails on Windows when executed from the command line. 

### Steps to reproduce
1. Run `jupyter nbconvert --to notebook --ExecutePreprocessor.timeout=600 --execute "Dataset Context Manager.ipynb"`
2. Repeat ~ 20 times

### Expected behaviour
The notebook should execute and the above command should exit with return code 0.

#### Locally 
When running the following script on my local Windows machine:

```python
from subprocess import run, PIPE

cmd = ['jupyter-nbconvert', '--to', 'notebook',
       '--execute', 'Dataset Context Manager.ipynb']

successes = []
fails = []
N = 100

for n in range(N):
    print(f'[*] Executing notebook, run {n}:{N-1}')
    
    result = run(cmd, stdout=PIPE, stderr=PIPE)
    if result.returncode == 0:
        print('[+] Succes')
        successes.append(result)
    else:
        print(f'[-] Fail, return code {result.returncode}')
        fails.append(result)
    print(f'Current score: {len(fails)} fails, {len(successes)} successes.')
    print('='*25)
```
I get the following error roughly 1 out of 11 times:
```
[NbConvertApp] Converting notebook Dataset Context Manager.ipynb to notebook
[NbConvertApp] Executing notebook with kernel: python3
[NbConvertApp] ERROR | Error while converting 'Dataset Context Manager.ipynb'
Traceback (most recent call last):
  File "C:\Users\qcodes-william\Anaconda3\envs\qcodesqcodes\lib\site-packages\nbconvert\nbconvertapp.py", line 381, in export_single_notebook
    output, resources = self.exporter.from_filename(notebook_filename, resources=resources)
  File "C:\Users\qcodes-william\Anaconda3\envs\qcodesqcodes\lib\site-packages\nbconvert\exporters\exporter.py", line 172, in from_filename
    return self.from_file(f, resources=resources, **kw)
  File "C:\Users\qcodes-william\Anaconda3\envs\qcodesqcodes\lib\site-packages\nbconvert\exporters\exporter.py", line 190, in from_file
    return self.from_notebook_node(nbformat.read(file_stream, as_version=4), resources=resources, **kw)
[...]

PR #1062, opened 2018-04-17
Use uuid over homegrown function

jenshnielsen:
Fixes https://github.com/QCoDeS/Qcodes/issues/1057

the hash function may return the same hash for two subs created just after each other due to the low resolution of time.time. Rather than fixing this by adding extra entropy from a random number i suggest using the python uuid module 
@WilliamHPNielsen

### DIFF
--- a/docs/examples/DataSet/subscriber json exporter.ipynb
+++ b/docs/examples/DataSet/subscriber json exporter.ipynb
@@ -27,7 +27,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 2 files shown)


WilliamHPNielsen:
I guess the haslib import and the definition of `hash_from_parts` should go too.

jenshnielsen:
@WilliamHPNielsen It is indeed unused and I was debating with my self if we want to remove it or just deprecate it for now

jenshnielsen:
@WilliamHPNielsen  Ready to land. Coverage drops because hash_by_parts is now uncovered which is fine

PR #1063, opened 2018-04-18
[Do not Merge] Check timers

jenshnielsen:
Mainly to investigate different timers on CI platforms
@WilliamHPNielsen related to #1057


jenshnielsen:
Judging from this the best is to use `time.perf_counter` which seems to be monotonic too on windows at least

PR #1066, opened 2018-04-19
refactor awg file handling in awg5014 driver and introduce QCoDeS @deprecated decorator

nataliejpg:
The existing driver had 'make_and_save' and 'make_send_and_load' which means that the code for packing a waveform list is copied in two places and also that the sequence 'make', 'save', 'send' and 'load' is not possible without making twice. I pulled out these functions and left the original two to call them so it shouldn't break it for anyone using the old ones. Thoughts welcome :)

- [x] test it

@WilliamHPNielsen


WilliamHPNielsen:
@nataliejpg, you are absolutely right. I put one of my best men on this :)

Dominik-Vogel:
I added a deprecation warning for the `generate_awg_file` method, to remove some of the confusion the user experiences when seeing this function and `make_awg_file`. I tested it with the sample notebooks and it seems to work without any problems.
I think we can merge this for now and do any further refactoring in a new pr.

nataliejpg:
sounds great. Thanks @Dominik-Vogel.

qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -903,7 +906,7 @@ def generate_channel_cfg(self):
         been changed from their default value and put them in a
         dictionary that can easily be written into an awg file, so as
         to prevent said awg file from falling back to default values.
-        (See self.generate_awg_file and self.AWG_FILE_FORMAT_CHANNEL)
+        (See self.make_awg_file and self.AWG_FILE_FORMAT_CHANNEL)
```
jenshnielsen: You could use actual sphinx links here. https://kev.inburke.com/kevin/sphinx-interlinks/

PR #1070, opened 2018-04-24
Update of R&S SGS100A driver

ThorvaldLarsen:
- Adding several more settings to the driver
- Allowing all intuitive input for on/off parameters
- Making driver slightly more consistent

Only tested with SGS100A but strongly suspect it will still work with SMB100A.


qcodes/instrument_drivers/rohde_schwarz/SGS100A.py
```diff
@@ -21,116 +21,155 @@ class RohdeSchwarz_SGS100A(VisaInstrument):
...
+                           label='IQ Modulation',
                            get_cmd=':IQ:STAT?',
-                           set_cmd=self.set_IQ_state,
-                           get_parser=self.parse_on_off,
-                           vals=vals.Strings())
+                           set_cmd=':IQ:STAT {}',
+                           get_parser=self.get_parser_on_off,
+                           set_parser=self.set_parser_on_off)
         self.add_parameter('pulsemod_state',
+                           label='Pulse Modulatuion',
```
WilliamHPNielsen: Small typo

WilliamHPNielsen:
I left one inline comment. Else this looks good, and if you've tested it, I think we should get it in.

ThorvaldLarsen:
Well spotted. Should be fixed now.

One thing about this driver: Not all SGS100A's have all options (either hardware and software options) which means some parameters in this driver does not apply to all instruments. If one tries to set a parameter that is not available it will silently fail. However, this is also the case without this PR.

One could ask the instrument for the available options in the init part and then add parameters according to the option. This seems a bit more involved though than I wanted for this PR.

WilliamHPNielsen:
@ThorvaldLarsen if there's a moderately easy way of doing that, I think we should give it priority to get that in, perhaps in a new PR. Do you think it is easy as in the Keysight DMM driver? See:
https://github.com/QCoDeS/Qcodes/blob/6049ee48329f1899506859404b33c8adacda83ef/qcodes/instrument_drivers/Keysight/Keysight_34465A.py#L116
and
https://github.com/QCoDeS/Qcodes/blob/6049ee48329f1899506859404b33c8adacda83ef/qcodes/instrument_drivers/Keysight/Keysight_34465A.py#L150

ThorvaldLarsen:
It would be rather easy I think. There is a command to ask the instrument what options are available both hardware and software. Depending on available option one can add relevant parameters and validators (eg. hardware option changes the ranges for power).

I briefly thought about this but decided it should be done correctly the first time around instead of trying to do it in this PR. One would have to go through the manual and find specifically what the options allow doesn't allow for the implemented parameters.

Maybe we can open and issue and leave it for a future PR?

WilliamHPNielsen:
> Maybe we can open and issue and leave it for a future PR?

I think that is reasonable. This PR is still an improvement without the option discrimination.

WilliamHPNielsen:
@ThorvaldLarsen, could you be bothered to open the issue? You can put (a sketch of) your proposed solution there.

PR #813, opened 2017-10-24
Driver/Minicircuits variable attenuators

Dominik-Vogel:
This is the driver for the Minicircuits RCDAT-6000-60 variable attenuator. The PR includes the implementation over Telnet(Ethernet) by @euchas and my additional implementation of a version that communicates via USB.
The USB part is general and is supposed to build the basis for other Minicircuit devices that do not have an ethernet interface such as the  RUDAT-13G-90 variable attenuator (#703)

@jenshnielsen , @WilliamHPNielsen:
 - may I assume that no one is using the telnet version yet and chnage the name of the class?
 - What is your opinion for a location of the USBMixin within the qcodes source tree?

Fixes #703.

Todos
 * implement incremental validator
 * change name, to something more general
 * make example notebook
 * test with RUDAT-13G-90

PR #1073, opened 2018-04-25
driver / RUDAT 13G 90 

sohailc:
Added a driver for the microcircuits RUDAT 13G 90 attenutator


Dominik-Vogel:
Looking at it again I think you should build on the old PR. You can simply push to that branch.

sohailc:
Actually, @Dominik-Vogel , I think I should keep the new PR. Your PR (#813) adds a driver for the rcdat_6000_60, which I do not have. I would either have to remove your driver or I'd have to push untested code, neither of which is desirable. I think we should merge this PR (once it is approved) and than in your branch you can use my mixin class.

sohailc:
Using a Mixin class did not work well with mypy. A similar situation is described here: 
https://github.com/python/mypy/issues/4001

I think the only solution is to make a subclass USBHIDInstrument and inherit from that.

Dominik-Vogel:
Dear @sohailc, I would prefer making a mixin and make an exception for mypy until the issue gets resolved. The reason for this is, that mypy is checking, a commodity feature, and the mixin is required to not repeat the same code: For minicircuits there exist quite a few devices that are ethernet and usb connectable. If you want to make a consistent driver you can implement all the same commands and then only add a mixin for the usb case.

sohailc:
Ok, I am using a Mixin, have another look, @Dominik-Vogel

PR #1075, opened 2018-04-26
test / Add simulated instrument + test for R&S HMC804

WilliamHPNielsen:
Ideally, we should have such tests for all instruments. Well, here's one.

@QCoDeS/core


sohailc:
Looks good to me as well

PR #1076, opened 2018-04-27
add support for writing mixed type lists and tuples in HDF5Formatter

peendebak:
In this PR three things are done:

* The existing HDF5 formatter is improved to support also lists and tuples with mixed type elements and numpy.int32 type elements.
* Another formatter is added based on serialization to hdf5 by https://github.com/telegraphic/hickle
* In the hdf5 data a special field is written 
* In the data a special field `__format_tag` is written to describe the format (e.g. make the distinction between the plan hdf5 formatter and the hickle version)

@WilliamHPNielsen @jenshnielsen @AdriaanRol


sohailc:
@eendebakpt I would like to see typing. Docstring would be nice, for example in the function _read_list_group, but this is not strictly necessary for private methods

sohailc:
On the other hand, a method like "write" needs proper docstrings. I see that you were not the author of that method, but now you are enhancing this module anyway, I will like to see docstrings.

sohailc:
I see that "write" does have proper docstrings. However, "read" does not. At least, it is not formatted properly. We cannot infer the input types for example

peendebak:
@sohailc I improved the docstrings for some of the functions (including all the public functions that have been added). 

What do you mean exactly with " infer the input types". Are there some tests from the CI I can check?

sohailc:
All I meant was that the docstring originally did not mention the input type.

sohailc:
I don't have further comments, but as I have not been deeply involved with this module or how writing hdf5 files work, I might have missed something. @WilliamHPNielsen @jenshnielsen, do you have additional comments?

WilliamHPNielsen:
@peendebak you can run `mypy` to check that all type annotations are consistent. In general, we like having type annotations, but we don't force contributors to use them (yet).

I think we can merge this PR now.

WilliamHPNielsen:
Trailing whitespace...

PR #1077, opened 2018-05-01
add _latest_datasets to DataSet

peendebak:
The latest created datasets are stored in a `deque` of finite length in the `DataSet`.

@WilliamHPNielsen


WilliamHPNielsen:
This behaviour is not quite the same as what the SQLite backend provides, is it? This way, the latest datasets are only defined in the current session, whereas the SQLite counts more "globally".

@peendebak, perhaps you can shed some light on what you intend to use this for?

peendebak:
@WilliamHPNielsen I don't know what the SQlite backend does, this is for the `qcodes.data.DataSet` object. It is used for providing a `get_latest_dataset()` function.

It is only for the current session, which is fine.

spauka:
Just to clarify the intended behavior of this PR?

- Currently this would return a list of the last instantiated instance of the DataSet class, which in its current form would be either the last created or the last loaded dataset.
- By the name, I would expect this to return the latest created dataset?
- Is there intended to be a public interface for accessing this data?

peendebak:
@spauka Currently only the last dataset is retained. This is configurable with the `size_latest_dataset` option. (storing a longer lists requires more memory of course)

If wanted I can create the public interface as a method `qcodes.DataSet.get_latest_dataset()` and
`qcodes.DataSet.get_latest_datasets()`.

astafan8:
@peendebak Are you going to add `qcodes.DataSet.get_latest_dataset()` method or not? I do not understand how this new `_latest_datasets` attribute is going to be used (e.g. which functions/classes/methods use it?).

astafan8:
@peendebak What shall we do with this PR? Is it still WIP?

astafan8:
@peendebak Seems that this PR is of no use, so I will close it next week unless you say otherwise.

PR #1078, opened 2018-05-03
Change logging of written commands

WilliamHPNielsen:
Make the QDac log what it writes similarly to the visa instruments.

@jenshnielsen


jenshnielsen:
Failures were due to mypy issue fixed in #1081

PR #1079, opened 2018-05-03
bugfix qtplot png representation

dpfranke:
I haven't looked into the details, but this fixes the png representation on my machine. Probably a change in pyqtgraph?


WilliamHPNielsen:
@dpfranke, two questions: what problem does this PR solve? What versions of packages are you using?

When I play around with
```
PyQt5 (5.9)
pyqtgraph (0.10.0)
QtAwesome (0.4.4)
qtconsole (4.3.1)
QtPy (1.3.1)
```

I get a `UnicodeDecodeError` when using this PR, but a (seemingly) nice bytestring when using the current master version.

dpfranke:
Ok, I get
`AttributeError: 'QByteArray' object has no attribute '_getValue'`
with the master version. This seems to be connected to the remote argument, working / failing example for IPython:
`from qcodes.plots.pyqtgraph import QtPlot`
`plot = QtPlot(remote=False)`
`plot`

WilliamHPNielsen:
@dpfranke, I see, thanks for clarifying. I think the way to go is then indeed as you suggest to only apply the fix from this PR if `remote == False`. That would at least be the quickfix.

dpfranke:
couldn't find that option, now using `hasattr`

Could you check whether that work around works for you, @WilliamHPNielsen

astafan8:
do i understand correctly that this png/plotting functionality is not and **cannot be** automatically tested, hence the complaints about code coverage? If so, then let's merge, because I see that the solution works for both @dpfranke and @WilliamHPNielsen ?

dpfranke:
@WilliamHPNielsen we use it as part of our (qt-based) data viewer / browser

PR #1080, opened 2018-05-03
driver / Update cryogenic magnet driver

mwilmer:
Fixes feature request for operation at ScapucciLab (Leiden).

- Add function to go to field specified by positive and negative numbers (bidirectional) instead of setting absolute value and sign (direction) separately
- Add convenience function to block until requested field is reached
- Change driver to use RAMP ZERO when requesting a to go to exactly 0 Tesla
- Extra safety check for switch heater

@WilliamHPNielsen


mwilmer:
@WilliamHPNielsen @peendebak 
Codacy passes cleanly, but CI gives warnings on parts of the code that I haven't touched.

WilliamHPNielsen:
@mwilmer I think that was fixed by #1081

PR #1081, opened 2018-05-03
MyPy: We don't cleanly pass strict optional so disable that for now

jenshnielsen:
This is an issue with the new MyPy 0.600


jenshnielsen:
Self merge to unbreak master

PR #1083, opened 2018-05-05
Driver / AWG5208 fixes

### LINKED ISSUE #1060: Driver/AWG5208 minor issues
The current Tektronix 5208 driver has a few quirks.

- The validator for channel amplitudes is wrong
- A corrupt `.seqx` file is **not** overwritten by `sendSEQXFile` leading to a lot of hair pulling ("but this code generating a file worked a minute ago!")
- A sequence loaded into memory is not overwritten when loaded a second time even though the `.seqx` file changed
- The marker stopped value is missing
- The marker range is wrong, should be +/- 1.75 V

WilliamHPNielsen:
Fixes #1060 

The sub-issues of #1060 are addressed/fixed in the following ways

1. Validators have been updated
2. Upload of files now by default overwrites (by first deleting the old file if it is present)
3. Marker stopped value has been added
4. A method to delete a sequence from sequence list has been added. There is unfortunately no (easy) way to know what sequences a `.seqx` file contains, so the user should manually delete the relevant sequence from the sequence list.
5. `play` now by default waits for the instrument to be ready. 

@ThorvaldLarsen @jenshnielsen 

@ThorvaldLarsen try to see if `awg.play(); awg.run_state()` always does what you think it should.


ThorvaldLarsen:
@WilliamHPNielsen I had to change 2 lines to get upload to work:
`1090              if lstlens[0] != np.shape(elem_names)[1]`
and
`1282             for assetname in np.array(elem_names)[:, n-1]`

qcodes/instrument_drivers/tektronix/AWG70000A.py
```diff
@@ -555,6 +620,14 @@ def _sendBinaryFile(self, binfile: bytes, filename: str,
 
         self.current_directory(path)
 
+        if overwrite:
+            log.debug(f'Pre-deleting file {filename} at {path}')
+            self.visa_handle.write(f'MMEMory:DELete {filename}')
```
ThorvaldLarsen: Missing quotation marks around {filename}:
self.visa_handle.write(f'MMEMory:DELete "{filename}"'

WilliamHPNielsen: Good one! Thanks.

ThorvaldLarsen:
@WilliamHPNielsen Plans got changed so I will be setting this up on T11 also tomorrow. Any chance that this one will be merged in the near future so I don't have to mess to much around with a local file across multiple computers?

If this is still some time out could I ask you to put the driver in wrappers repository? Then I can use the driver from there until this one is done.

ThorvaldLarsen:
Yes, carrying around a local file between two computers atm.

WilliamHPNielsen:
Gotcha.

WilliamHPNielsen:
@ThorvaldLarsen, I think the current version addresses all of the points you raised in this conversation. Can you confirm, then I'll merge it in.

ThorvaldLarsen:
Yes this looks good. Just tested it and my upload works with this version.
Thanks!

WilliamHPNielsen:
Great, thanks! Then I just need to make the tests pass 😛

PR #1085, opened 2018-05-07
added config needed for wrappers to main qcodes

Dominik-Vogel:
I think we should generally have all the available options in the main qcodes config whether they are needed or not to avoid a copy pasting of files upon every install. Alternatively we should make better error messages.


jenshnielsen:
I don't really like having unused config objects. At the least the doc string should clarify that it's not used inside qcodes

Dominik-Vogel:
Good point to add a comment!

Dominik-Vogel:
Unfortunately there seems to be no good way to add comments to json files...

jenshnielsen:
No but you can add it to the description in the schema

PR #1091, opened 2018-05-09
Fix / dataset / add lists and tuples via add_result

### LINKED ISSUE #1084: DataSaver.add_result should fail with an error if unknown type is passed
Currently, if as part of a sweep, the result of an `ArrayParameter` is passed to `DataSaver.add_result` as a python list or tuple, as opposed to an `ndarray`, the data will silently be truncated to the first element. This is not made clear to the user and hence data is easily lost.

### Expected behaviour
Either an error should be thrown for an unknown type, or python sequence types should be handled.

### Actual behaviour
Data is silently truncated to the first item in the array.

### System
Windows on QCodes Master (940b46f6)

WilliamHPNielsen:
Fixes #1084 

Although it is CLEARLY insinuated in the `ArrayParameter` docstring that an `ArrayParameter` must return a numpy array, some `ArrayParameter` subclasses apparently return lists or tuples. Since there is no good reason *not* to accept such an input, this PR makes the measurement context manager do just that.

- Handle any type of input in `add_result`, allowing for lists and tuples

@spauka, I probably missed some corner case, so I'd be happy to have your eyes on this.

@QCoDeS/core


qcodes/dataset/measurements.py
```diff
@@ -101,7 +104,7 @@ def add_result(self,
                 raise ValueError(f'Can not add a result for {paramstr}, no '
                                  'such parameter registered in this '
                                  'measurement.')
-            if isinstance(value, np.ndarray):
+            if sum([isinstance(value, typ) for typ in array_like_types]) > 0:
```
spauka: Is the regular method of achieving this sort of comparison normally the `any` function?
```py3
any(isinstance(value, typ) for typ in array_like_types):
```

WilliamHPNielsen: Haha, it sure is. Thanks.

WilliamHPNielsen:
@spauka, I think we can be forgiving. I mean, why not? We may enforce that `ArrayParameters` always return numpy arrays, but I really don't see a reason why a custom parameter shouldn't be allowed to be a list or a tuple.

qcodes/dataset/measurements.py
```diff
@@ -18,6 +18,9 @@
 
 log = logging.getLogger(__name__)
 
+array_like_types = (tuple, list, np.ndarray)
```
sohailc: Maybe its better to do something like: 

```python
import collections

if isinstance(value, collections.Iterable): 
   ....
```

qcodes/dataset/measurements.py
```diff
@@ -111,6 +114,14 @@ def add_result(self,
                                      f'and {array_size}')
                 else:
                     input_size = array_size
+            elif any(isinstance(value, t) for t in non_array_like_types):
```
sohailc: Why not just 

```python 
elif not any(isinstance(value, t) for t in non_array_like_typs): 
    raise ValueError(...)
```

qcodes/dataset/measurements.py
jenshnielsen: Note that that will also include strings which we probably dont want to consider arrays of chars

qcodes/dataset/measurements.py
WilliamHPNielsen: yes that does the same

spauka:
Is this one ready to merge?

WilliamHPNielsen:
@spauka, sure thing.

PR #1095, opened 2018-05-16
Dataset: pass colorbar to plot

jenshnielsen:
This may be needed in case you want to reuse your axes for additional plots.
In the current code matplotlib will generate a new colorbar each time you replot gradually shrinking the figure. Just deleting the colorbar axes from the figure is not enough as the old one is not removed from the subplotspec and the space never reclaimed

@QCoDeS/core  and @jana-d


jenshnielsen:
@WilliamHPNielsen I think I have now cleaned this up and it should be ready to merge

WilliamHPNielsen:
Thanks, buddy!

PR #1100, opened 2018-05-17
Driver GS200

johnhornibrook:
Yokogawa GS200 no longer changes state (output off, voltage mode) on initialisation. 

Reasoning is that connections should be made quietly. In this particular instrument, we may wish to reconnect but have a current persist in a coil, device, etc.

As discussed with @jenshnielsen. Also @sohailc, did you see any problem with removing this?


jenshnielsen:
@nataliejpg and @ThorvaldLarsen I think you are using the Yokogawa too. Do you see any issue with this change

Dominik-Vogel:
I think this is a great idea. It should be the general rule, that instantiation does not change the state of an instrument. I would merge it if nobody objects.
We should keep in mind that It might be slightly dangerous for people who expect it to be set to zero.

ThorvaldLarsen:
@Dominik-Vogel Yes, this looks good. If anyone wants to have specific settings like these they can easily be added in the stationconfigurator (if/when it is part of main qcodes).

Dominik-Vogel:
@ThorvaldLarsen I am thinking more about the immediate danger of existing scripts that assume that the voltage is set to zero.

PR #1101, opened 2018-05-17
[WIP] Added the ability of param specs to add dependent and inferred

sohailc:
As part of the effort to integrate pysweep into QCoDeS, I propose to modify the ParamSpec class slightly. We need to make it possible to add "depends_on" and "inferred_from" after the ParamSpec is created. This will largely get rid of the need to have a separate ParamsTable in the Sweep class. 

The envisioned design is as follows: Let us suppose we have a sweep: 

```python
sweep_object = Nest(
   ParameterSweep(p, lambda: [0, 1, 2, 3]), 
   ParameterWrapper(q)
)
```
That is, we sweep parameter p and measure q at each set point. Each partial sweep object in the nest will create a param spec object with no "inferred_from" or "depends_on" arguments. We will call these param spec objects `spec_p` and `spec_q`. The nesting operation will add `spec_p` as a depends_on parameter on `spec_q`. 

Each sweep object will have an array of ParamSpec objects. Nesting, zipping and chaining sweep objects will have corresponding operations on their ParamSpec objects.


sohailc:
Note: I have not added tests yet. Please do *NOT* merge this

jenshnielsen:
@sohailc From a quick read this looks good. Will wait for the tests before doing a more detailed review

sohailc:
@jenshnielsen I have added tests. Can you please have a look?

sohailc:
Thanks Jens. As discussed, I am adding the sweep functionality in small PR's at a time. A follow-up PR is to be found here: https://github.com/QCoDeS/Qcodes/pull/1107. 

This introduces the concept of a parameter table.

sohailc:
@WilliamHPNielsen If you don't have any comments, I will go ahead an merge this in

PR #1110, opened 2018-05-28
Fix/no more experimentsdb everywhere

### LINKED ISSUE #1102: Creation of db files when import qcodes
It's not a bug I guess, but I think its still unwanted. 

### Steps to reproduce
Set any directory to the current working directory
Execute a python script/program that just imports the qcodes module

### Expected behaviour
No side-effects of the file system

### Actual behaviour
An experiment.db file is created in the current working directory. 

I think it's better if measurement databases are only created when we actually call a function that initialized an experiment. 
I use the qcodes module to load my experimental data from a lot of different working directories, like folders where I work on papers, during the analyses, etc...

WilliamHPNielsen:
Fixes #1102

- Stop making all those darn `.db` files

@QCoDeS/core @majoqubitT10


jenshnielsen:
👍 Perhaps we should also change the default config file to store in user home rather than local directory?

WilliamHPNielsen:
@QCoDeS/core, I'd like to merge this. Any objections?

PR #1111, opened 2018-05-28
 Znb read from existing channels

jenshnielsen:
@QCoDeS/core This enables reading channels that are already mapped on the ZNB from a manual user config. This is sort of a stop gap solution because we cant fully setup everything with the driver yet.

For use by @GrigoryanRuben


jenshnielsen:
WIP because the notebook needs cleanup

jenshnielsen:
So I would much prefer to have this done in a proper way but I think this is good enough to be useful. And I have a hart time seeing how we would get enough bandwidth to implement all the functionality needed to setup more advanced channels on the ZNB without relying on using the front panel. @QCoDeS/core What do you think?

astafan8:
Big thanks for reworking the notebook!!!

Minor comments on the notebook:
* perhaps it can be executed with a new db file so that there is no long list of measurements at the top of the notebook))
* in17 - not full_trace but trace_mag_phase if understand it correctly?
* in40 - is the warning there by intention? it does not seem like so.

jenshnielsen:
@astafan8 
* Yes, fixed
* Yes that parameter changed name at some point and the notebook was not updated
* Added the rf on and cleared the warning. I don't have the instrument any more. So I did not rerun

@Dominik-Vogel There is an example of that in [17] as far as I can see

PR #1112, opened 2018-05-29
Sr865 add model A and correct max freq for 865

jenshnielsen:
This is in line with the documentation that I can find online and matches the physical sr865 that we have in the lab

@QCoDeS/core


jenshnielsen:
Still need to test this on an actual device before merging

jenshnielsen:
Tested this, our current sr865 (not A) does not seem to implement all the buffer commands that the driver uses. It is unclear if this is due to fundamental differences or an older version of the firmware. We are running firmware version 1.18 but the sr860 is running 1.47, Awaiting information from SR about firmware versions

jenshnielsen:
@Dominik-Vogel has upgraded the firmware in the 865 and it now matches the other lockins. This is ready to land

PR #1113, opened 2018-05-29
Fix/1109

### LINKED ISSUE #1109: np.nan removed when using the context manager
To reproduce: 
```python
import numpy as np
from numpy import ndarray

from qcodes import ParamSpec, new_data_set, new_experiment
from qcodes.dataset.data_export import get_data_by_id

from qcodes.dataset.experiment_container import new_experiment

new_experiment("nan_types", "nan")

meas.register_custom_parameter(
    name="x", unit="V"
)

meas.register_custom_parameter(
    name="m", unit="I", setpoints=("x",)
)

with meas.run() as datasaver:
    datasaver.add_result(("x", [0, 1, 2]), ("m", np.array([0.0, np.nan, 1.0])))

dataset = get_data_by_id(datasaver.run_id)
```
If we inspect the dataset we will see that the np.nan has been removed entirely. This is unacceptable, especially in 2D datasets as it will prevent us from plotting the data in the usual manner.

sohailc:
Fixes #1109

### DIFF
--- a/qcodes/dataset/sqlite_base.py
+++ b/qcodes/dataset/sqlite_base.py
@@ -117,7 +117,7 @@ def _convert_numeric(value):
     return numeric
 
 
-def _adapt_numeric(fl):
+def _adapt_float(fl):
     if np.isnan(fl):
         return "nan"
     return float(fl)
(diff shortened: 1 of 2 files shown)


Dominik-Vogel:
Forget about it, we are already using NULL for values that are not present.

sohailc:
@Dominik-Vogel I have added typing. Can you approve this now?

sohailc:
@WilliamHPNielsen what do you think?

PR #1115, opened 2018-05-30
Allow SR86x amplitude to be zero.

### LINKED ISSUE #1114: SR86x driver does not allow amplitude set to 0
### Steps to reproduce
1. try runnning `lockin.amplitude(0)`

### Expected behaviour
Lockin amplitude should be set to 0

### Actual behaviour
Validation error is thrown, driver has a minimum amplitude of 10e-9V

spauka:
Fixes #1114 .

- Change validator in `SR86x.py` to allow 0 amplitude.

@jenshnielsen

### DIFF
--- a/qcodes/instrument_drivers/stanford_research/SR86x.py
+++ b/qcodes/instrument_drivers/stanford_research/SR86x.py
@@ -338,7 +338,7 @@ class SR86x(VisaInstrument):
                            get_cmd='SLVL?',
                            set_cmd='SLVL {}',
                            get_parser=float,
-                           vals=Numbers(min_value=10e-9, max_value=2))
+                           vals=Numbers(min_value=0, max_value=2))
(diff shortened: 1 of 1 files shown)


jenshnielsen:
I guess the validator is based on the follwing string from the manual

```
The SLVL v command sets the sine out amplitude to 
v. The value of v will be rounded to 3 digits or 1
nV, whichever is greater.
The amplitude may be programmed from 1 nV to 2.0 V.
``

jenshnielsen:
From my quick testing it does seem to support setting it to zero so I guess this is safe

jenshnielsen:
@spauka I see the same thing as you quite confusing. I think this will work best if we allow to set it to 0?

spauka:
Agree, it's a bit confusing to be able to walk over and set to zero, but not be able to do it from inside qcodes...

Issue #990, opened 2018-03-07
string representation of data base fails when no run_id is present


sohailc:
![image](https://user-images.githubusercontent.com/10454668/37093109-7b04bd30-220e-11e8-8488-26b561729c8d.png)

jenshnielsen:
@sohailc could you in the future copy paste the text instead of screenshots? You can use githubs markdown to format it nicely

PR #1116, opened 2018-05-30
Make loaded datasets similar to new ones

### LINKED ISSUE #1106: dataset.completed does not work
### Steps to reproduce

```
dataset = load_by_id(1)
dataset.completed
```

Output:

> Traceback (most recent call last):
> 
>   File "<ipython-input-178-f869a978e653>", line 1, in <module>
>     dataset.completed
> 
>   File "c:\users\triton5acq\qcodes\qcodes\dataset\data_set.py", line 306, in completed
>     return self._completed
> 
> AttributeError: 'DataSet' object has no attribute '_completed'

### System
**Windows 7**

**qcodes/master**

**19a2016f9c2d999027b3949a646200d2d5b41e7e**

WilliamHPNielsen:
Fixes #1106

- Tie the `DataSet` object somewhat stronger to the sqlite DB
- Make loaded `DataSet`s have the same attributes as newly created ones
- Add a few tests

Also clean up a little in the underlying sql functions (see the `mark_run` function).

@jenshnielsen @sohailc


qcodes/dataset/data_set.py
```diff
@@ -658,16 +677,27 @@ def new_data_set(name, exp_id: Optional[int] = None,
...
-        if len(get_experiments(d.conn)) > 0:
-            exp_id = get_last_experiment(d.conn)
+        if len(get_experiments(conn)) > 0:
+            exp_id = get_last_experiment(conn)
         else:
             raise ValueError("No experiments found."
                              "You can start a new one with:"
                              " new_experiment(name, sample_name)")
+    # This is admittedly a bit weird. We create a dataset, link it to some
+    # run in the DB and then (using _new) change what it's linked to
```
sohailc: Can you also explain why you are doing weird stuff like that?

sohailc:
Your comment "Make loaded DataSets have the same attributes as newly created ones" makes me a bit nervous. A good design principle is that classes should define all attributes in __init__. This way all instances of a class will always have the same attributes. Let's make sure that this convention is followed everywhere.

sohailc:
In fact, come to think of it, the title "Make loaded datasets similar to new ones" makes me really nervous! What do you mean, make it 'similar' to new ones?! Are loaded and new data sets not simply instances of the same class?

sohailc:
Thanks for your comments, William. I will look into 1106 and think about a good solution

WilliamHPNielsen:
@QCoDeS/core while waiting for a "deeper" solution, perhaps we can accept this quickfix that will allow users to load datasets and see whether they are completed?

jenshnielsen:
The best alternative solution that @Dominik-Vogel and I can see is to modify the init method so it can do the work of all the new functions. That admittedly results in a somewhat convoluted init function but perhaps it's still better

WilliamHPNielsen:
I think either way is fine, but I think we should be reasonably quick in applying the fix.

WilliamHPNielsen:
In the meantime, let me just cross-link to #990, since that actually taps in to the same problem.

WilliamHPNielsen:
@astafan8 I think that is a good way to proceed. The AppVeyor business is not directly related to this PR, but feel free to extend the deadlines and push that to this branch if you want. Else I think we can just merge?

astafan8:
@WilliamHPNielsen i had a quick look at the AppVeyor output. I think indeed we can merge. Shall i do it?

PR #1119, opened 2018-06-04
Add support for Matplotlib 2.2

jenshnielsen:
Matplotlib no longer works with nans in the x and y coordinates for pcolormesh. This fills out any missing values by extrapolating based on the step size or the existing data, falling back to a default stepsize if there is only one line. IMHO this also makes the code for shifting coordinates from bin centers to edges slightly simpler. 

@ThorvaldLarsen I think you saw this issue somewhere in the qdev wrappers.

@QCoDeS/core

### DIFF
--- a/qcodes/plots/qcmatplotlib.py
+++ b/qcodes/plots/qcmatplotlib.py
@@ -5,6 +5,7 @@ using the nbagg backend and matplotlib
 from collections import Mapping
 from collections import Sequence
 from functools import partial
+from copy import deepcopy
 
 import matplotlib.pyplot as plt
 from matplotlib import ticker
@@ -242,7 +243,13 @@ class MatPlot(BasePlot):
                     plot_object.remove()
 
                 ax = self[config.get('subplot', 1) - 1]
-                plot_object = self._draw_pcolormesh(ax, **config)
+                kwargs = deepcopy(config)
+                # figsize may be passed in as part of config.
+                # pcolormesh will raise an error if this is passed to it
+                # so strip it here.
+                if 'figsize' in kwargs:
+                    kwargs.pop('figsize')
+                plot_object = self._draw_pcolormesh(ax, **kwargs)
                 trace['plot_object'] = plot_object
 
                 if plot_object:
@@ -302,7 +309,7 @@ class MatPlot(BasePlot):
         # didn't want to strip it out of kwargs earlier because it should stay
         # part of trace['config'].
         args_masked = [masked_invalid(arg) for arg in [x, y, z]
-                      if arg is not None]
+                       if arg is not None]
 
         if np.any([np.all(getmask(arg)) for arg in args_masked]):
             # if the z array is masked, don't draw at all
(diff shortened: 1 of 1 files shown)


ThorvaldLarsen:
I tried running with matplotlib 2.2 in connection with https://github.com/qdev-dk/qdev-wrappers/issues/88 and ran into errors with plotting. However, turned out the problem we were trying to solve was unrelated to matplotlib version.

jenshnielsen:
@ThorvaldLarsen I was referring to this one which I think depend on matplotlib version and which this fixes

> @Dominik-Vogel @jenshnielsen just FYI: qcodes MatPlot fails for interrupted scans with matplotlib version 2.2.2 due to a change in handling of MaskedArrays matplotlib/matplotlib#9723. pcolormesh throws this error: ValueError: x and y arguments to pcolormesh cannot have non-finite values or be of type numpy.ma.core.MaskedArray with masked values when trying to plot the data.

qcodes/plots/qcmatplotlib.py
```diff
@@ -288,6 +295,57 @@ def _draw_plot(self, ax, y, x=None, fmt=None, subplot=1,
...
+                    step_size = np.ma.average(np.ma.diff(arr))
+
+                last_good_value = arr[np.logical_not(arr.mask)][-1]
+                extrapolation_start = last_good_value+step_size
+                n_invalid = np.sum(arr.mask)
+                extrapolation_stop = extrapolation_start+step_size*(n_invalid-1)
+                # numpy (1.14) has a deprecation warning related to shared
+                # masks
+                # lets silence this by making sure that this is not shared
+                # before
```
jenshnielsen: PEP8 is nice and all but IMHO this indentation is worse than too long lines 😛

qcodes/plots/qcmatplotlib.py
```diff
@@ -288,6 +295,57 @@ def _draw_plot(self, ax, y, x=None, fmt=None, subplot=1,
...
+                # modifying the mask
+                arr.unshare_mask()
+                arr[arr.mask] = np.linspace(extrapolation_start,
+                                            extrapolation_stop,
+                                            num=n_invalid)
+
+                # now shift to get edges coordinates rather than center
+                # coordinates
+                # first Add padding on both sides equal to endpoints
```
jenshnielsen: This comment is now wrong. Perhaps we should just use insert, That seems simpler than pad for this

qcodes/plots/qcmatplotlib.py
WilliamHPNielsen: Yeah, I'm still getting used to VSCode. That would just have been `meta+q` in Emacs... sigh...

qcodes/tests/test_qcmatplotlib_functions.py
```diff
@@ -0,0 +1,89 @@
...
+                                     ysign*yrange,
+                                     step=N//2+1)
+
+        args_masked = [np.ma.masked_invalid(arg) for arg in [x, y, z]]
+
+        args = MatPlot._make_args_for_pcolormesh(args_masked, x, y)
+
+        assert len(args[0]) == M + 1
+        assert len(args[1]) == N + 1
+
```
jenshnielsen: Should we not test that the content of x and y is as expected too?

WilliamHPNielsen: Yes... the test can/should be extended a lot. I just got very lazy.

PR #1120, opened 2018-06-04
Bump dependecies

jenshnielsen:
This should probably wait till h5py 2.8 is released on pypi  https://github.com/h5py/h5py/issues/1040
to allow numpy 1.14 without deprecation warnings at import. __Update__: this is now out and the requirements file changed to reflect this.

@WilliamHPNielsen It would be good to let qcodes depend on a released version of broadbean and add it to setup.py. I could not find a reason  for the direct dependency of schema here. That seems to only be an indirect dependency due to broadbean. Is that correct?


WilliamHPNielsen:
>  [scema] seems to only be an indirect dependency due to broadbean. Is that correct?

Yes, that's right.

environment.yml
```diff
@@ -20,5 +21,7 @@ dependencies:
  - pyyaml
  - lxml
  - scipy
+ - ruamel_yaml
+ - pyvisa=1.8
```
astafan8: pyvisa was from pip because Jens told me that conda has weird versions (don't remember exactly the problem). Has it been fixed?

.appveyor.yml
```diff
@@ -18,12 +18,11 @@ install:
...
   - conda config --set always_yes true  # else AppVeyor will hang forever waiting for user ([y]/n)
   - conda update -n base conda
-  - conda env create -q -f environment.yml python=3.6
+  - conda env create -q -f environment.yml
   - activate qcodes
   - pip install -r test_requirements.txt
   - pip install -r docs_requirements.txt
-  - python -m pip install --upgrade pip
-  - pip install .
+  - pip install -e .
```
astafan8: `-e` for appveyor and no `-e` for Travis? why not `-e` for both?

jenshnielsen: We want to test both types of install. As you may remember there are plenty of examples of regular install being broken and editable working. (Because editable just puts in a link and does not care about the install manifest)  But most scientist (at least in Copenhagen) uses an editable on windows so imho it makes the most sense to test these together

jenshnielsen:
I would prefer stopping to use conda-forge because if you are mixing conda channels it's somewhat random which channel a specific dependency is resolved from unless you pin to the specific build number of the conda package. Unfortunately the pyvisa package is named differently in conda and pypi

So an upgrade (`conda env update`) with this file will resolve in the following installed packages.

```
PyVISA                    1.9.1                     <pip>
pyvisa                    1.8                      py36_0    conda-forge
```

.appveyor.yml
astafan8: ok, agree.

WilliamHPNielsen:
@jenshnielsen we haven't received any more crash reports regarding `ipykernel` and `spyder`, but since we now there is an issue, I propose that we play it safe and block `ipykernel`.

jenshnielsen:
@QCoDeS/core I think this is ready to land now

I have not touched the division between travis/pip and appveyor/conda We can revisit that in another pr

For now pyvisa is still installed from conda-forge but the latest version which fixes a bug than requires us to change the tests slightly

I also added a note about keeping the env up to data to the docs and we now depend on a released version of broadbean

astafan8:
great!

one comment - since we are pinning to pyvisa 1.9.1 (and not going back to 1.8), we need to revert some stuff, see https://github.com/QCoDeS/Qcodes/issues/1192 .

docs/start/index.rst
```diff
@@ -81,6 +89,25 @@ If you have installed with git, pull the QCoDeS repository using your
...
+new versions of packages that QCoDeS depends on are released.
+
+To keep the QCoDeS environment upto date you can run
+
+.. code:: bash
+
+   conda env update
+
+from the root of your QCoDeS git repository. Alternatively you can grab a new
+version of the ``environment.yml`` file from above and run the same command from a
```
astafan8: from above? meaning from qcodes git repo?

docs/start/index.rst
```diff
@@ -81,6 +89,25 @@ If you have installed with git, pull the QCoDeS repository using your
 favourite method (git bash, git shell, github desktop, ...). There are
 new commits to the repository daily.
 
+Keeping your environment up to date
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Additional dependencies are periodically added to the QCoDeS environment and
+new versions of packages that QCoDeS depends on are released.
+
+To keep the QCoDeS environment upto date you can run
```
astafan8: typo: up to date (with a space)

docs/start/index.rst
```diff
@@ -81,6 +89,25 @@ If you have installed with git, pull the QCoDeS repository using your
...
+
+.. code:: bash
+
+   conda env update
+
+from the root of your QCoDeS git repository. Alternatively you can grab a new
+version of the ``environment.yml`` file from above and run the same command from a
+directory containing that file. This will update all the QCoDeS dependencies to the
+latest recommended version. Note that if you install packages yourself into the same
```
astafan8: Shall here be a "new paragraph" in order to separate the "Note .." from the above text?

docs/start/index.rst
astafan8: what if we add a reference to `conda env update` command, so that the users can read how to use this command themselves instead of us explaining all the details?

environment.yml
```diff
@@ -3,20 +3,28 @@ channels:
...
- - jsonschema
- - pyvisa=1.8.0
- - QtPy
+ - numpy=1.15
+ - matplotlib=2.2.3
+ - pyqtgraph=0.10.0
+ - pyvisa=1.9.1
+ - h5py=2.8.0
+ - pyqt=5.9
  - sip
```
astafan8: shall we also pin `sip` as it is done in requirements.txt?

astafan8:
@jenshnielsen if you are up to reverting it here, please, do, it's great :) concerning the `attrbiute_set_to`... you can keep it just because it is tested, hence can be used when necessary. i leave it up to you.

docs/start/index.rst
jenshnielsen: Sure but the documentation that I can find is rather bad https://conda.io/docs/commands/env/conda-env-update.html do you have a better link

astafan8: no, i don't)) it's not bad, it's just not formatted properly. i hope they fix it someday...

docs/start/index.rst
jenshnielsen: I only pined it originally because there was a abi version mismatch between the versions of sip and pyqt on pypi. (sip is a direct dependency of pyqt5 to generate the c++ wrappers and only used for that) so in principle we should not need to pin it at all. I don't expect that will ever happen for the conda package

environment.yml
astafan8: https://github.com/QCoDeS/Qcodes/pull/1120#discussion_r218830950

astafan8: ok, then let's unpin it everywhere :)

PR #1122, opened 2018-06-05
made context manager for set

### LINKED ISSUE #1117: Settable parameters as context managers
It might be useful for parameters which can be set to act as context managers, such that user code can be guaranteed that a parameter is returned to its original value after a block of code. For instance:

```python
>>> v = dac.voltage()
>>> with dac.voltage(-1):
...     # Do stuff with the DAC output set to -1 V.
...
>>> assert abs(dac.voltage() - v) <= tolerance
```
@QCoDeS/core

sohailc:
Fixes #1117.


qcodes/tests/test_parameter.py
```diff
@@ -902,3 +902,31 @@ def tearDown(self):
...
+        self.instrument.close()
+        del self.instrument
+
+    def test_none_value(self):
+        with self.instrument.a.set(3):
+            assert self.instrument.a.get() == 3
+        assert self.instrument.a.get() is None
+
+    def test_context(self):
+        self.instrument.a.set(1)
```
astafan8: what is this line needed for?

sohailc: forgot to remove it

sohailc:
@jenshnielsen can you approve this now?

jenshnielsen:
Seems sensible to me. I am still in doubt about is this should bind directly to set. @WilliamHPNielsen any opinion

WilliamHPNielsen:
I would probably prefer to bind this to something called `set_to`, simply because I conjecture that users will find that (slightly) easier to understand. And I also personally find it nice that the line
```
with dac.voltage.set_to(0):
```
kind of reads out like a natural language command.

But I don't feel very strongly about either option.

astafan8:
shall we merge this pull-request or is there anything to do?

astafan8:
Let's poll for the name of the method that can be used as context manager for the "set" functionality:
/polls "existing set() - with dac.voltage.set(1):" "set_to() - with dac.voltage.set_to(1):"

sohailc:
@jenshnielsen @astafan8 @WilliamHPNielsen  Can we have another look, please?

Dominik-Vogel:
Sorry for the delay, I was out last week.
I find the two naming options slightly confusing. For an unexperienced user a `with` statement is something exotic and the prose english interpretation of the two suggested names does not reflect its behavior.
I would prefer something like
```
with dac.voltage.temporarily_at(0)
```

PR #1126, opened 2018-06-07
Nonetype accepted by datasaver

### LINKED ISSUE #1125: Nonetype not accepted as valid type by data saver
Let's consider the situation where we interleave a 1D and 2D sweep. This will look as follows.  

```python
import numpy as np

from qcodes import Parameter

x = Parameter("x", set_cmd=None, get_cmd=None)
y = Parameter("y", set_cmd=None, get_cmd=None)

m = Parameter("m")
m.get = lambda:np.sin(x())

n = Parameter("n")
n.get = lambda:np.sin(x()**2 + y())

meas = Measurement()
meas.register_parameter(x)
meas.register_parameter(y)
meas.register_parameter(m, setpoints=[x])
meas.register_parameter(n, setpoints=[x, y])

with meas.run() as datasaver: 
    for ix in np.linspace(0, 6, 100): 
        x.set(ix)
        datasaver.add_result(("x", ix), ("y", None), ("m", m.get()), ("n", None))

        for iy in np.linspace(0, 6, 100): 
            y.set(iy)
            datasaver.add_result(("x", ix), ("y", iy), ("m", None), ("n", n.get()))
```

Currently, this fails because of `ValueError: Wrong value type received. Got <class 'NoneType'>, but only int, float, str, tuple, list, and np.ndarray is allowed.`.  This is unfortuante as the underlying infrastructure has no problems with None values

sohailc:
Fixes #1125 .

### DIFF
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -360,7 +360,7 @@ class DataSet(Sized):
 
         Args:
             - list of name, value dictionaries  where each
-              dictionary provides the values for of the parameters in
+              dictionary provides the values for the parameters in
               that result. If some parameters are missing the corresponding
               values are assumed to be None
 
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -375,7 +375,7 @@ class DataSet(Sized):
         values = [[d.get(k, None) for k in expected_keys] for d in results]
 
         len_before_add = length(self.conn, self.table_name)
-        insert_many_values(self.conn, self.table_name, expected_keys,
+        insert_many_values(self.conn, self.table_name, list(expected_keys),
                            values)
         # TODO: should this not be made atomic?
         self.conn.commit()
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -360,8 +360,9 @@ class DataSet(Sized):
 
         Args:
             - list of name, value dictionaries  where each
-              dictionary provides the values for all of the parameters in
-              that result.
+              dictionary provides the values for of the parameters in
+              that result. If some parameters are missing the corresponding
+              values are assumed to be None
 
         Returns:
             - the index in the DataSet that the **first** result was stored at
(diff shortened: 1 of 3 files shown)


WilliamHPNielsen:
I got a little confused by the discussion in #1125. Is this ready for review now?

sohailc:
Yes @WilliamHPNielsen it is

qcodes/dataset/data_set.py
```diff
@@ -360,8 +360,9 @@ def add_results(self, results: List[Dict[str, VALUE]]) -> int:
 
         Args:
             - list of name, value dictionaries  where each
-              dictionary provides the values for all of the parameters in
-              that result.
+              dictionary provides the values for of the parameters in
```
WilliamHPNielsen: A small but important word is missing here.

qcodes/dataset/measurements.py
```diff
@@ -116,6 +116,8 @@ def add_result(self,
                     input_size = array_size
             elif any(isinstance(value, t) for t in non_array_like_types):
                 pass
+            elif value is None:
```
WilliamHPNielsen: I guess it's still undecided whether we should allow this (but you actually voted against it, @sohailc, by voting for option 1)

WilliamHPNielsen:
The fix in `data_set.py` is a genuine fix! Great job. An alternative to
```
expected_keys = frozenset.union(*[frozenset(d) for d in results])
```
could be to ask the database what the keys are, but since we don't need to provide all parameters (we don't, right?), I think this is better.

Let us agree on option 1. or 2. from #1125 before merging. I vote 1.

qcodes/dataset/data_set.py
WilliamHPNielsen: ...or there is a word too many.

sohailc:
You are right of course, I voted for option 1. I will remove the offending code. 

As for your comment that we can look for all parameters in the database, I tried that and it does not appear to work. I think the reason is that before the first write to database, there are no parameters yet

qcodes/dataset/measurements.py
sohailc: I will remove this

qcodes/dataset/data_set.py
sohailc: I not sure i'm following....

WilliamHPNielsen: "dictionary provides the values for (all/none/some) of the parameters"
or, alternatively,
"dictionary provides the values for the parameters"

WilliamHPNielsen:
> I think the reason is that before the first write to database, there are no parameters yet.

That is a very good point.

sohailc:
@WilliamHPNielsen Lets merge this if there are no further comments.

qcodes/dataset/data_set.py
WilliamHPNielsen: Just this little thing and then we are set.

PR #1127, opened 2018-06-07
Hotfix/driver/awg5208 small bugs

WilliamHPNielsen:
- Fix some things @ThorvaldLarsen found
- Allow any number of markers (at least up to four)

### DIFF
--- a/qcodes/instrument_drivers/tektronix/AWG70000A.py
+++ b/qcodes/instrument_drivers/tektronix/AWG70000A.py
@@ -940,6 +940,9 @@ class AWG70000A(VisaInstrument):
         # STEP 2:
         # Make all subsequence .sml files
 
+        print('Waveforms done')
+        print(wfmx_filenames)
+
         subseqsml_files: List[str] = []
         subseqsml_filenames: List[str] = []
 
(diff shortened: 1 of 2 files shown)


qcodes/instrument_drivers/tektronix/AWG70000A.py
```diff
@@ -830,12 +830,9 @@ def _makeWFMXFileBinaryData(data: np.ndarray, amplitude: float) -> bytes:
...
-            if M == 2:
-                markers = data[1, :]
-            elif M == 3:
-                m1 = data[1, :]
-                m2 = data[2, :]
-                markers = m1+2*m2  # This is how one byte encodes both markers
+            markers = data[1, :]
+            for i in range(1, M-1):
+                markers += data[i+1, :] * (2**i)
                 markers = markers.astype(int)
```
ThorvaldLarsen: `markers = markers.astype(int)` has to be outside loop. Otherwise it throws an error when it tries to add float number to int.

ThorvaldLarsen:
With the comment above fixed this version runs on T3. Tested marker upload on AWG5208 for up to 4 markers and with upload to 2 channels. Measurement data makes sense after these changes.

PR #1128, opened 2018-06-07
Add set_parameter_bounds to IVVI.py

kNalj:
- updates values of dac upon changing the polarity of that dac

@WilliamHPNielsen


qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -433,8 +433,8 @@ def set_pol_dacrack(self, flag, channels, get_all=True):
         val = flagmap[flag.upper()]
         for ch in channels:
             self.pol_num[ch - 1] = val
-            # self.set_parameter_bounds('dac%d' % (i+1), val, val +
-            # self.Fullrange.0)
+            self.set_parameter_bounds('dac%d' % (i+1), val, val + 
+                                      self.Fullrange.0)
```
jenshnielsen: I don't think this line split is valid python code

jenshnielsen:
@kNalj Can you remove the whitespaces that Codacy noticed then I think this should be ready

PR #1130, opened 2018-06-07
make register_parameter match add_result for array_parameter

jenshnielsen:
@WilliamHPNielsen Fixes an inconsistency between add_parameter and the register_parameter for ArrayParameters. At the one uses the name of the root instrument but the other users the full name. This fails for an arrayparameter inside a channel since one name will contain the name of the channel but the other one does not.


neildick:
👍 Cherry-picking this into master fixes the issue on my end.

astafan8:
do we have a test that covers this bug?

jenshnielsen:
@astafan8 Obviously not :( the coverage remains the same and no test failures

jenshnielsen:
@astafan8 This now has a test. Assuming Appveyor passes this should be ready for review

PR #1131, opened 2018-06-08
qcodes-1124 Add explicit flushing to avoid instability of testing

### LINKED ISSUE #1124: Tests for dataset context manager subscriptions are flaky
@QCoDeS/core know what I'm talking about. It's annoying as hell and might point to some scary underlying problem. We should really try to figure out what is going on.

astafan8:
Partially fixes #1124 

The test for subscriptions for the dataset is not stable, which means that it occasionally passes. This pull-request makes the test stable by actually making it more explicit.

The test essentially checks whether subscribers get executed at the moment the data is flushed to the database. In order to test it, two subscribers are created, data is added in a loop, and assertions are performed in the same loop. Since data is flushed to the database only once within a write_period, the original code uses sleep method next to the call to add_result, where the sleep time is set to 120% of the write_period - this way the test "ensures" that the data is flushed to the database (hence, subscribers are called) and assertions can validly be performed.

The problem of this approach is that if the sleep time is not enough for the data to be flushed to the database at some point, perhaps, due to the fact that some other system processes accidentally get higher priority, then at the moment when an assertion is performed, the expected data may not be in the database yet.

In order to make the test for subscribers more stable and explicit, the sleeping approach is substituted with an explicit call to flush_data_to_database.

@WilliamHPNielsen please have a look at the fix

### DIFF
--- a/qcodes/tests/dataset/test_measurement_context_manager.py
+++ b/qcodes/tests/dataset/test_measurement_context_manager.py
@@ -341,6 +341,10 @@ def test_enter_and_exit_actions(experiment, DAC, words):
 
 
 def test_subscriptions(experiment, DAC, DMM):
+    """
+    Here the following is tested: subscribers should be called at the moment the data is flushed to the database,
+    i.e. after new result is added. For the purpose of testing, flush_data_to_database method is called explicitly.
+    """
 
     def subscriber1(results, length, state):
         """
@@ -383,8 +387,10 @@ def test_subscriptions(experiment, DAC, DMM):
 
             (a, b) = as_and_bs[num]
             expected_list += [c for c in (a, b) if c > 7]
-            sleep(1.2*meas.write_period)
+
             datasaver.add_result((DAC.ch1, a), (DMM.v1, b))
+            datasaver.flush_data_to_database()
+
             assert lt7s == expected_list
             assert list(res_dict.keys()) == [n for n in range(1, num+2)]
 
--- a/qcodes/tests/dataset/test_measurement_context_manager.py
+++ b/qcodes/tests/dataset/test_measurement_context_manager.py
@@ -342,8 +342,15 @@ def test_enter_and_exit_actions(experiment, DAC, words):
 
 def test_subscriptions(experiment, DAC, DMM):
     """
-    Here the following is tested: subscribers should be called at the moment the data is flushed to the database,
-    i.e. after new result is added. For the purpose of testing, flush_data_to_database method is called explicitly.
+    Test that subscribers are called at the moment that data is flushed to database
+
+    Note that for the purpose of this test, flush_data_to_database method is called explicitly instead of waiting for
+    the data to be flushed automatically after the write_period passes after a add_result call.
+
+    Args:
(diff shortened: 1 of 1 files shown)


WilliamHPNielsen:
I may be missing something, but this does not seems like a fix. This just changes the test to not execute the sequence of commands that *ought* to work... Right?

sohailc:
With test subscriptions we are just testing if the subscribers are called
when we flush to the dataset. Flushing was previously being done indirectly
by waiting. Now we call flush directly. The object of the test is being
accomplished so I do not agree that it does not fix the issue; it does.

Having said that, there is indeed still someting wrong. Waiting for a time
before calling add_results *should* have caused the flushing as well. But
that should be the object and concern of a different test.

WilliamHPNielsen:
> With test subscriptions we are just testing if the subscribers are called when we flush to the dataset

The test was originally thought to be more of an integration test (test both that flushing happens and that subscribers get called), but I think it makes sense to split it up like you suggest. Then we need to hammer the flushing with some more testing.

qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -341,6 +341,10 @@ def action(lst, word):
 
 
 def test_subscriptions(experiment, DAC, DMM):
+    """
+    Here the following is tested: subscribers should be called at the moment the data is flushed to the database,
```
WilliamHPNielsen: PEP8 this and let's fly

astafan8: definitely, sorry for that

sohailc:
I agree. We need a test which tests if the database flushing happens at the prescribed write period. Right now this is missing. I will make an issue for that

WilliamHPNielsen:
Doesn't `test_datasaver_scalars` already do that?

sohailc:
Oh I did not see that. I think the test name does not describe what its testing. Also, I think multiple things are tested there again. 

But It's kind of weird that this test passes while "test_subscriptions" was failing. I think we need to keep this issue open, even after merging in PR 1131

qcodes/tests/dataset/test_measurement_context_manager.py
astafan8: sure, will do

qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -341,6 +341,17 @@ def action(lst, word):
...
 
 def test_subscriptions(experiment, DAC, DMM):
+    """
+    Test that subscribers are called at the moment that data is flushed to database
+
+    Note that for the purpose of this test, flush_data_to_database method is called explicitly instead of waiting for
+    the data to be flushed automatically after the write_period passes after a add_result call.
+
+    Args:
+        experiment (qcodes.dataset.experiment_container.Experiment) : qcodes experiment object
```
WilliamHPNielsen: For the main code base we try to move the type annotations out of the docstring and into the call signature, such that `mypy` can read them and tell us that we are being stupid. For tests I don't think it really matters much.

WilliamHPNielsen:
Hooray! Life just got 8% better

PR #1138, opened 2018-06-11
Allow tuple selection for ChannelLists

ThorvaldLarsen:
We are currently calling channellists for measurements with alazar and VNA. Sometimes one wants to take data from only a subset of the channels in the channellists however currently one can only choose slices or a single channel.
This PR allows one to take data from a subset of a channellist defined by a tuple (e.g. alazar_ctrl.channels[2,4,6].data).

- Add functionality for channellist __getitem__ to allow for a tuple in addition to int and slice.

@jenshnielsen

### DIFF
--- a/qcodes/instrument/channel.py
+++ b/qcodes/instrument/channel.py
@@ -187,7 +187,7 @@ class ChannelList(Metadatable):
                 raise TypeError("All items in this channel list must be of "
                                 "type {}.".format(chan_type.__name__))
 
-    def __getitem__(self, i: Union[int, slice]):
+    def __getitem__(self, i: Union[int, slice, tuple]):
         """
         Return either a single channel, or a new ChannelList containing only
         the specified channels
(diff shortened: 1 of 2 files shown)


jenshnielsen:
@ThorvaldLarsen Sounds reasonable, I think this needs  a test or two. Do you have time to write that or should I do it? \

ThorvaldLarsen:
I have never written tests or looked into how it is done so would be great if you can do that part.

jenshnielsen:
No problem.

@spauka I think this is originally your code. Do you have any comments?

jenshnielsen:
@ThorvaldLarsen I have written some tests.
I guess you know that [2,4,6] can be written as [2:7:2] :) but in general that is not always possible so this is still reasonable

ThorvaldLarsen:
Wasn't actually aware of that. Have never really become friends with slices...
Thanks for the tip.

spauka:
Hey, this makes sense, looks good to me :)

PR #1142, opened 2018-06-13
Consistent names

### LINKED ISSUE #1140: Inconsistent naming of parameters and instruments
This is not so much a bug as an annoying inconsistency. I am not sure if there is a good way to fix this whire retaining back wards compatibility but none the less the naming is inconsistent consider the following:

| type       | name                                       | full_name                | short_name        |
|------------|--------------------------------------------|--------------------------|-------------------|
| Instrument | "{name}" e.g. "myinst"                       | N/A                      | N/A               |
| Channel    |  "...{parent_name}_{name}" e.g. "myinst_ch1" | N/A                      | "{name}" e.g. "ch1" |
| Parameter  | "{name}" e.g. "frequency"                  | "{instrument_name}_{name}" e.g. "{myinst_ch1_frequency}" | N/A               |

jenshnielsen:
Fixes #1140 .

This tries to fix #1140 by making instrument, channel and paramters have consistent full_name and short_name. `name` is still behaving different. Adds tests for the behavior too

@QCoDeS/core

### DIFF
--- a/qcodes/instrument/channel.py
+++ b/qcodes/instrument/channel.py
@@ -35,6 +35,7 @@ class InstrumentChannel(InstrumentBase):
 
         self.name = "{}_{}".format(parent.name, str(name))
         self.short_name = str(name)
+        self.full_name = self.name
         self._meta_attrs = ['name']
 
         self._parent = parent
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -45,6 +45,7 @@ class InstrumentBase(Metadatable, DelegateAttributes):
     def __init__(self, name: str,
                  metadata: Optional[Dict]=None, **kwargs) -> None:
         self.name = str(name)
+        self.short_name = str(name)
 
         self.parameters = {} # type: Dict[str, _BaseParameter]
         self.functions = {} # type: Dict[str, Function]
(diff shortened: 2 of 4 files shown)


jenshnielsen:
@Dominik-Vogel I added the name_parts thing that we discussed too

qcodes/instrument/channel.py
```diff
@@ -59,10 +60,24 @@ def ask(self, cmd):
...
 
+    @property
+    def parent(self) -> InstrumentBase:
+        return self._parent
+
     @property
     def root_instrument(self) -> InstrumentBase:
         return self._parent.root_instrument
 
+    @property
```
Dominik-Vogel: Could we not just simply write
```
return self._parent.name_parts().append(short_name)
```
?

jenshnielsen: Good point. This was written inside out not outside in so it wasn't originally possible

Dominik-Vogel:
I think we could make things a little less involved by basing everything of the name_parts:
we can make `full_name` a property that simply concatenates the `name_parts`

Dominik-Vogel:
It would be nice to deprecate `name` but I can see how this entails a lot of changes... maybe we can do this in another PR?

Dominik-Vogel:
I just noticed, that we might also need to update `MultiChannelInstrumentParameter`  and take a look at our beloved multiparameters and array parameters...

jenshnielsen:
We should but honestly I would prefer doing that in another pr. These changes makes it a lot simpler to implement the parameter groups that are needed to replace those things

Dominik-Vogel:
ok

PR #1146, opened 2018-06-14
Change loglevel for send/recv in ip.py

ThorvaldLarsen:
Loglevel `info` for send/recv commands completely floods logfile when using ip instruments. Propose to change level to debug.

- Change loglevel for send/recv of ip instruments from info to debug.

@Dominik-Vogel


WilliamHPNielsen:
The corresponding level in `visa.py` is also `DEBUG`, so that seems reasonable.

PR #1153, opened 2018-06-20
Added support for N5183B Signal Generator

dahlend:
Added support for the Agilent/Keysight N5183B Signal Generator.

Fixed a string type error as well in val_mapping.
The set_cmd has the formatting of a string, so I changed the val_mapping to a string.
```
set_cmd='OUTP:STAT ' + '{:s}'
```

Additionally this device doesn't really match the model number of the previous devices, but it's effectively the same.  I leaned toward adding it despite the name not matching exactly vs creating a whole new driver which would just be a copy of the original with a very slight name change.

### DIFF
--- a/qcodes/instrument_drivers/Keysight/N51x1.py
+++ b/qcodes/instrument_drivers/Keysight/N51x1.py
@@ -42,8 +42,8 @@ class N51x1(VisaInstrument):
 
         self.add_parameter('rf_output',
                            get_cmd='OUTP:STAT?',
-                           set_cmd='OUTP:STAT ' + '{:s}',
-                           val_mapping={'on': '1', 'off': '0'})
+                           set_cmd='OUTP:STAT {}',
+                           val_mapping={'on': 1, 'off': 0})
                             
         self.connect_message()
         
--- a/qcodes/instrument_drivers/Keysight/N51x1.py
+++ b/qcodes/instrument_drivers/Keysight/N51x1.py
@@ -6,7 +6,7 @@ from qcodes.utils.validators import Numbers
 class N51x1(VisaInstrument):
     """
     This is the qcodes driver for Keysight/Agilent scalar RF sources.
-    It has been tested with N5171B, N5181A, N5171B
+    It has been tested with N5171B, N5181A, N5171B, N5183B
     """
 
     def __init__(self, name, address, **kwargs):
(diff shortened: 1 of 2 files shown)


qcodes/instrument_drivers/Keysight/N51x1.py
```diff
@@ -42,7 +43,7 @@ def __init__(self, name, address, **kwargs):
         self.add_parameter('rf_output',
                            get_cmd='OUTP:STAT?',
                            set_cmd='OUTP:STAT ' + '{:s}',
-                           val_mapping={'on': 1, 'off': 0})
```
jenshnielsen: I don't understand the need for this change. The number is automatically converted into a str when you generate the string

jenshnielsen:
Looks good to me. @johnhornibrook as the original author do you have any comments

qcodes/instrument_drivers/Keysight/N51x1.py
jenshnielsen: Imho it should just be `set_cmd='OUTP:STAT {}'`

WilliamHPNielsen:
@dahlend re the naming: I don't think we have any official guidelines on this, but it's unfortunate to have driver file names not matching the actual models (for when we quickly want to look up whether we have a driver for a given instrument). You could consider making a new `Keysight_N5183B.py` file with one line:
```
from qcodes.instrument_drivers.Keysight.N51x1 import N51x1 as N5183B
```

qcodes/instrument_drivers/Keysight/N51x1.py
astafan8: indeed, this is what i've seen in the code for the drivers that I've used so far.

astafan8:
completely agree with @WilliamHPNielsen about the naming. Our users should not be bothered by the fortunate fact of the same driver class working for multiple device models.

dahlend:
Made the additional file as suggested
Changed the string formatting as suggested.

qcodes/instrument_drivers/Keysight/N51x1.py
dahlend: Yeah that's a better solution
My python 3.6 install throws an exception on the original version for some reason.

jenshnielsen:
Agree with @astafan8 codacy may provide insight into issues but it is not a hard requirement for it to pass

PR #1156, opened 2018-06-22
Add a simple Lakeshore 331 driver.

spxtr:
Mostly copied from the 336, although there are some differences. Notably, there are only two channel inputs `A` and `B`, and one of the commands from the 336 does not exist on the 331. Tested on my own 331, let me know if I should do more.

@jenshnielsen


Dominik-Vogel:
There is a major rebuild going on with the Lakeshore driver base. In the PR #972 you will find an almost finished version of the new structure. It would be great if you could base the driver off that new base class.

spxtr:
Sure, no rush. I'll update this once that PR goes in.

PR #1157, opened 2018-06-24
Added Keysight N5230A PNA-L

dahlend:
Adding basic support for Keysight VNA N5230A.
Expect more devices from me over the next few days/weeks.


qcodes/instrument_drivers/Keysight/Keysight_N5230A.py
```diff
@@ -0,0 +1,160 @@
...
+        IDN = self.ask_raw('*IDN?')
+        vendor, model, serial, firmware = map(str.strip, IDN.split(','))
+        IDN = {'vendor': vendor, 'model': model,
+               'serial': serial, 'firmware': firmware}
+        return IDN
+
+    def _get_trace(self, verbose=False):
+
+        self.start_sweep(self.num_averages())
+        while not self.start_sweep():
```
astafan8: Does it make sense to make the interface clearer by implementing smth like `while not self.sweep_is_finished()` where `sweep_is_finished` is a method rather than a parameter, and `start_sweep` is also a method rather than a parameter?

qcodes/instrument_drivers/Keysight/Keysight_N5230A.py
```diff
@@ -0,0 +1,160 @@
...
+    This is the qcodes driver for Keysight/Agilent N5230A PNA-L Network Analyzer.
+    This should be simple to alter to work with other keysight PNAs
+    """
+
+    def __init__(self, name, address, **kwargs):
+        super().__init__(name, address, terminator='\n', **kwargs)
+
+        self.add_parameter('get_trace',
+                           label='Get Trace',
+                           get_cmd=self._get_trace,
```
astafan8: it seems more clear to me (and a better design) to implement this as a method rather then as a parameter with only "get" command. What is the advantage of the current implementation apart from being quick-to-write for the developer?

qcodes/instrument_drivers/Keysight/Keysight_N5230A.py
```diff
@@ -0,0 +1,160 @@
...
+                           get_parser=int)
+
+        self.add_parameter('num_averages',
+                           label='Number of Averages of a sweep',
+                           get_cmd=':SENS1:AVER:COUN?',
+                           set_cmd=':SENS1:AVER:COUN {}',
+                           get_parser=int)
+
+        self.add_parameter('averaging',
+                           label='Enable Averaging',
```
astafan8: "Enable averaging" -> "Enable/disable averaging" (since disabling also happens through this parameter)

qcodes/instrument_drivers/Keysight/Keysight_N5230A.py
```diff
@@ -0,0 +1,160 @@
...
+                                    ':SENS:SWE:GRO:COUN {};'
+                                    ':SENS:SWE:MODE GRO;'
+                                    '*OPC'),
+                           get_parser=lambda x: int(x) & 32 == 32
+                           )
+
+        self.connect_message()
+
+    def get_idn(self):
+        IDN = self.ask_raw('*IDN?')
```
astafan8: please, do not use capital letters in the name of a variable (PEP8). "IDN" -> "idn"

qcodes/instrument_drivers/Keysight/Keysight_N5230A.py
```diff
@@ -0,0 +1,160 @@
...
+        """This gets the current trace on the VNA without waiting for a sweep to finish"""
+        # Get the current viewing format
+        current_format = self.ask('CALC:FORM?')
+
+        # Change format to smith, and return the current full trace
+        data = self.visa_handle.query_binary_values('CALC:FORM SMIT;:FORM:DATA REAL,64;:CALC1:DATA? FDATA',
+                                                    container=np.array,
+                                                    datatype='d',
+                                                    is_big_endian=True)
+        # Format it correctly
```
astafan8: instead of just saying "formatting it correctly", please, write in the comment explicitly that the returned data array actually contains alternating real and imaginary parts of the values, hence the need for "formatting it correctly"

qcodes/instrument_drivers/Keysight/Keysight_N5230A.py
```diff
@@ -0,0 +1,160 @@
...
+                           get_cmd=':SENS:FREQ:SPAN?',
+                           set_cmd=':SENS:FREQ:SPAN {}',
+                           get_parser=float,
+                           unit='Hz')
+
+        self.add_parameter('s_parameter',
+                           label='S parameter',
+                           docstring='Set the measured S parameter, allowed values are "S11", "S12", "S21", and "S22"',
+                           get_cmd=':CALC:PAR:CAT?',
+                           set_cmd=':CALC:PAR:MOD  {}',
```
astafan8: unnecessary extra space in front of `{}`?

qcodes/instrument_drivers/Keysight/Keysight_N5230A.py
```diff
@@ -0,0 +1,160 @@
...
+                           get_parser=lambda x: x[1:-1].split(',')[1])
+
+        self.add_parameter('num_points',
+                           label='Number of points',
+                           get_cmd=':SENS:SWE:POIN?',
+                           set_cmd=':SENS:SWE:POIN  {}',
```
astafan8: unnecessary extra space in front of `{}`?

qcodes/instrument_drivers/Keysight/Keysight_N5230A.py
```diff
@@ -0,0 +1,160 @@
...
+                           set_cmd=':SENS:FREQ:SPAN {}',
+                           get_parser=float,
+                           unit='Hz')
+
+        self.add_parameter('s_parameter',
+                           label='S parameter',
+                           docstring='Set the measured S parameter, allowed values are "S11", "S12", "S21", and "S22"',
+                           get_cmd=':CALC:PAR:CAT?',
+                           set_cmd=':CALC:PAR:MOD  {}',
+                           get_parser=lambda x: x[1:-1].split(',')[1])
```
astafan8: should there be an explanation for this parser so that other developers know the reasoning behind it?

qcodes/instrument_drivers/Keysight/Keysight_N5230A.py
```diff
@@ -0,0 +1,160 @@
...
+        self.add_parameter('span',
+                           label='Span',
+                           get_cmd=':SENS:FREQ:SPAN?',
+                           set_cmd=':SENS:FREQ:SPAN {}',
+                           get_parser=float,
+                           unit='Hz')
+
+        self.add_parameter('s_parameter',
+                           label='S parameter',
+                           docstring='Set the measured S parameter, allowed values are "S11", "S12", "S21", and "S22"',
```
astafan8: is it possible to use `val_mapping` for the allowed values here?

qcodes/instrument_drivers/Keysight/Keysight_N5230A.py
```diff
@@ -0,0 +1,160 @@
...
+import numpy as np
+from qcodes.utils.validators import Numbers
+import time
+"""Driver for the Keysight PNA N5230A"""
+
+
+class N5230A(VisaInstrument):
+    """
+    This is the qcodes driver for Keysight/Agilent N5230A PNA-L Network Analyzer.
+    This should be simple to alter to work with other keysight PNAs
```
astafan8: do you mean that this driver is not N5230A-specific? either way, this line does not add value because "other keysight PNAs" instruments are not explicitly mentioned. In case this driver work for a different model, I would rather recommend to mention that other device explicitly (preferably via creating another class that inherits from a common class, and make this driver inherit from the common one as well).

qcodes/instrument_drivers/Keysight/Keysight_N5230A.py
jenshnielsen: Or an Enum validator

qcodes/instrument_drivers/Keysight/Keysight_N5230A.py
jenshnielsen: I agree , I don't think the get and set methods are a good match for each other. Such a complex set command also makes me slightly nervous. it seems very tailored to a specific use case. But I could be wrong

qcodes/instrument_drivers/Keysight/Keysight_N5230A.py
jenshnielsen: The current design (Loop etc. ) expect everything that returns data to be a parameter. But as it is here that is not really useful for that purpose unless it is expressed as an array or multi parameter. But the name of the parameter should probably not be get_trace but just trace if this is expressed as a parameter. It is odd to be able to do get_trace.get() imho

qcodes/instrument_drivers/Keysight/Keysight_N5230A.py
WilliamHPNielsen: I guess you could argue that this is a constant. But then you'd be arguing about PEP8, and that very quickly escalates to a full-scale war.

astafan8:
@dahlend any chance you will continue work on this PR?

dahlend:
Ah im sorry, I totally forgot about this, I no longer work with any hardware. I will close this.

PR #1159, opened 2018-06-25
Fix flaky subscriber tests

### LINKED ISSUE #1124: Tests for dataset context manager subscriptions are flaky

astafan8:
Fixes #1124 - flaky tests for subscribers. Now with much more assurance than in previous PR.

TL;DR - The reason for the tests being flaky in short is: upon exit of the Runner context manager, the subscribers may get unsubscribed *before* their internal queue is exhausted (e.g. callbacks are executed on *all* the saved data points). This happens when the CPU is too busy.

I was able to reproduce the failure when I was running the flaky test `test_subscriptions_getting_all_points` and watching a 4K video on YouTube while adjusting its playback speed (to make CPU even busier). Then I analyzed the logs which SQL debugging spits out, and I found that the subscriber's callback is not called on the very last data point.

The reason for that is that the subscriber may get unsubscribed *before* the subscribers callback is called on the very last data point (by default the minimum size of the subscribers queue is 1). So, effectively, the subscriber thread would be stopped before his queue containing the last data point is exhausted. Hence, the "state" object, that is associated with a callback of the subscriber, would not be updated with that last data point, which in turn would break the assertion in the test.

I have to note that this pull-request might not solve the flakiness up to 100%. In order to make this test deterministic, we need to ensure that just enough time has passed between the moment the data is flushed to database and the "state" object (that is passed to subscriber constructor) has been updated by the corresponding subscriber's callback function. At the moment, there is no robust way to ensure this. The reason for this is that the subscribers have internal queue which is populated via a trigger call from the SQL database. Hence, from the test's "main" thread, it is difficult to know whether the queue is empty because the subscriber callbacks have already been executed, or because the triggers of the SQL database has not been executed yet.

In general, this problem is caused by both - bad design of the test, and bad design of the subscribers infrastructure. In order to save time, I propose this PR, and in case we are still bothered with the flakiness, I will propose to rewrite the Subscriber and, hence, its tests.

Main change proposed in this pull request:
- [first mark dataset as complete, and only then unsubscribe the subscribers](https://github.com/QCoDeS/Qcodes/compare/master...astafan8:qcodes-1124-subscriber-test-flaky#diff-034a6ccc5c62d7f10514f3c4e44fed2e) in `Runner.__exit__` (qcodes/dataset/measurements.py).

Smaller changes made along the way:
- Refactor some of the subscriber tests for readability
- Refactor Subscriber class for readability
- Fix a typo bug that prevented removing triggers for subscribers from the SQL database + a test for it

@WilliamHPNielsen @jenshnielsen @sohailc @Dominik-Vogel 

(sorry that the commit content is not very clean, next time it will be better because I won't be messing around with unfamiliar features of PyCharm, and will be using good old comfortable git bash)

### DIFF
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -601,7 +601,7 @@ class DataSet(Sized):
             del self.subscribers[uuid]
 
     def _remove_trigger(self, name):
-        transaction(self.conn, f"DROP TRIGGER IF EXISTS name;")
+        transaction(self.conn, f"DROP TRIGGER IF EXISTS {name};")
 
     def unsubscribe_all(self):
         """
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -78,7 +78,7 @@ class _Subscriber(Thread):
                  id_: str,
                  callback: Callable[..., None],
                  state: Optional[Any] = None,
-                 loop_sleep_time: int = 0,  # in seconds
+                 loop_sleep_time: int = 0,  # in milliseconds
                  min_queue_length: int = 1,
                  callback_kwargs: Optional[Dict[str, Any]]=None
                  ) -> None:
@@ -95,7 +95,7 @@ class _Subscriber(Thread):
         self.data_queue: Queue = Queue()
         self._queue_length: int = 0
         self._stop_signal: bool = False
-        self._loop_sleep_time = loop_sleep_time / 1000  # convert seconds to minutes
+        self._loop_sleep_time = loop_sleep_time / 1000  # convert milliseconds to seconds
         self.min_queue_length = min_queue_length
 
         if callback_kwargs is None:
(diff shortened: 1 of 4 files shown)


qcodes/dataset/data_set.py
```diff
@@ -60,66 +60,79 @@ class CompletedError(RuntimeError):
...
 
         self.state = state
-        self.min_wait = min_wait
-        self.min_count = min_count
-        self._send_queue: int = 0
+
+        self.data_queue: Queue = Queue()
+        self._queue_length: int = 0
+        self._stop_signal: bool = False
+        self._loop_sleep_time = loop_sleep_time / 1000  # convert seconds to minutes
```
sohailc: I think you mean convert milli-seconds to seconds

astafan8: absolutely correct, will fix it

astafan8: done

PR #1160, opened 2018-06-25
Standford Research SR860 lock-in amplifier driver improvements

sohailc:
An update for the driver for Standford Research lock-in amplifiers SR86x and SR860 in particular. The work is done together with @astafan8.

### BUG FIXES:
- Fix maximum frequency for SRS860 - it is 500 kHz, not 500 Hz
- get_capture_data uses the above functions; it does not have the 64kB limit anymore, instead it's limit is the current capture length (via _get_raw_capture_data function)
- sine_outdc can set the value below 1 mV, the instrument can handle that

### NEW DRIVER METHODS:
- add caprture_one_sample_per_trigger that captures a given number of samples per each trigger signal; a callable argument starts the trigger pulses train
- add start_capture_at_trigger that captures a given number of samples once it receives a trigger signal; a callable argument initiates sending a trigger signal
- add set_capture_length_to_fit_samples to conveniently set the capture length for the number of sample (not kilobytes)
- add wait_until_samples_captured to conveniently wait until the given number of samples is captured

### NEW DRIVER PARAMETERS:
- add parameter for the source of the trigger signal
- add parameter for the source of the reference signal
- _get_raw_capture_data_block implements the direct low-level call to CAPTUREGET command of the instrument
- _get_raw_capture_data allows to get data from the beginning of the buffer avoiding the instrument's 64kB-reading-per-time limit

### EXAMPLE NOTEBOOK UPDATED
- more descriptive
- covers capture modes

### Code improvements related to capturing data via the buffer
- add min_capture_length_in_kb as a property
- add max_capture_length_in_kb as a property
- make _set_capture_len_parser object method from static method
- make _set_capture_len_parser use new capture-length-related properties
- make _calc_capture_size_in_kb reuse _set_capture_len_parser
- add convenient set_capture_rate_to_maximum method
- add convenient _get_list_of_capture_variable_names and reuse it
- add convenient _get_number_of_capture_variables and reuse it

@jenshnielsen @WilliamHPNielsen @Dominik-Vogel


WilliamHPNielsen:
Not-so-useful comment: It would make my editor happy if we could stick to a maximum line length of 79.

qcodes/instrument_drivers/stanford_research/SR86x.py
```diff
@@ -156,17 +158,53 @@ def snapshot_base(self, update: bool = False,
...
+        Parse the capture length in kB according to the way buffer treats it. This method checks for the range
+        of the input value, and throws an error if the range is not satisfied. Additionally, if the input value
+        is odd, it is made even (and a warning is issued, if desired) since the capture length can only be set
+        in even numbers.
+
+        This method intended for use in other methods inside this class for calculation purposes. Setting issue_warning
+        argument to True allows this method to be used as well for set commands of the instrument parameters.
+
+        Args:
+            capture_length_in_kb (int): The desired capture length in kB.
```
WilliamHPNielsen: I think it is superfluous to write the type down here again. (same for subsequent docstrings)

qcodes/instrument_drivers/stanford_research/SR86x.py
```diff
@@ -156,17 +158,53 @@ def snapshot_base(self, update: bool = False,
...
+            issue_warning (bool): If True, a warning is issued when setting capture length to an odd number
+        Returns:
+            capture_len_in_kb (int)
+        """
+        if capture_length_in_kb % 2:
+            capture_length_in_kb += 1
```
WilliamHPNielsen: I don't think it is a good idea to give the user anything else than either exactly what they ask for or a big fat exception. I realise that this is a private method, but I think that principle is still valid.

WilliamHPNielsen:
I left a few minor comments. Else it looks sensible. Have you tested it on real hardware, @sohailc and @astafan8?

qcodes/instrument_drivers/stanford_research/SR86x.py
```diff
@@ -156,17 +158,53 @@ def snapshot_base(self, update: bool = False,
...
+            capture_len_in_kb (int): The desired capture length in kB.
 
-        if not 1 <= value <= 4096:
-            raise ValueError("the capture length should be between 1 and 4096")
+        Returns:
+            capture_len_in_kb (int)
+        """
+        return self._parse_capture_length(capture_len_in_kb, issue_warning=True)
+
+    def _parse_capture_length(self, capture_length_in_kb: int, issue_warning: bool=False) -> int:
```
sohailc: I think this function has a wrong name; it does not really parse anything. Additionally, I think maybe William is right when he says that the function should return exactly what the user asked (see his comment below). In this case, I think the function should be "validate_capture_length" and should not return anything

sohailc: @WilliamHPNielsen Yes, this has all been tested on real hardware

qcodes/instrument_drivers/stanford_research/SR86x.py
sohailc: agree. i will remove it.

qcodes/instrument_drivers/stanford_research/SR86x.py
```diff
@@ -1,25 +1,26 @@
...
-        instrument (SR86x): This argument is unused, but needed because the add_parameter method of the Instrument
-                            base class adds this as a kwarg.
+    The parameter array that holds read out data. We need this to be compatible
+    with qcodes.Measure
+
+    Args:
+        name
+        instrument
+            This argument is unused, but needed because the add_parameter
+            method of the Instrument base class adds this as a kwarg.
```
jenshnielsen: I don't understand why the instrument is just not passed to the super method so that the paramter is bound to the instrument. (I realize this is nothing new)

astafan8: indeed. smth to fix later

Issue #1143, opened 2018-06-13
driver / AMI430 handles invalid values incorrectly


WilliamHPNielsen:
The AMI430 driver does not handle invalid target field values correctly. I could only test this with the simulated instrument, but as far as I can see from the source, the real instrument will have the same behaviour.

### Steps to reproduce

```python
import numpy as np

import qcodes.instrument.sims as sims
from qcodes.instrument_drivers.american_magnetics.AMI430 import AMI430_3D, AMI430Warning
from qcodes.instrument.ip_to_visa import AMI430_VISA
from qcodes.math.field_vector import FieldVector

field_limit = [
    lambda x, y, z: x == 0 and y == 0 and z < 3,
    lambda x, y, z: np.linalg.norm([x, y, z]) < 2
]

# path to the .yaml file containing the simulated instrument
visalib = sims.__file__.replace('__init__.py', 'AMI430.yaml@sim')


mag_x = AMI430_VISA('x', address='GPIB::1::65535::INSTR', visalib=visalib,
                    terminator='\n', port=1)
mag_y = AMI430_VISA('y', address='GPIB::2::65535::INSTR', visalib=visalib,
                    terminator='\n', port=1)
mag_z = AMI430_VISA('z', address='GPIB::3::65535::INSTR', visalib=visalib,
                    terminator='\n', port=1)

driver = AMI430_3D("AMI430-3D", mag_x, mag_y, mag_z, field_limit)

#######################################################################
# Example starts here

print(f'x is set to {driver.x()}')
try:
    invalid_value = 999
    print(f'Trying to set x to {invalid_value}')
    driver.x(invalid_value)
except ValueError:
    print('Did NOT set x')
print(f'x is set to {driver.x()}')
```

### Expected behaviour
The set point should not update when the value is invalid.

### Actual behaviour
The set point *does* update.

@QCoDeS/core

WilliamHPNielsen:
@sohailc, perhaps it is easiest for you, as the original author, to look into this one?

sohailc:
@WilliamHPNielsen I am on it

sohailc:
https://github.com/sohailc/Qcodes/tree/fix/ami_invalid_field_limits_set

sohailc:
https://github.com/QCoDeS/Qcodes/pull/1162

jenshnielsen:
Fixed by #1162

PR #1162, opened 2018-06-26
1) the routine setting set points will first adjust individual instru…

sohailc:
Solves issue 1143

The problem was that the set point vector was adjusted before testing that the set point was safe.

### DIFF
--- a/qcodes/instrument_drivers/american_magnetics/AMI430.py
+++ b/qcodes/instrument_drivers/american_magnetics/AMI430.py
@@ -570,32 +570,32 @@ class AMI430_3D(Instrument):
         # Get and set parameters for the set points of the coordinates
         self.add_parameter(
             'cartesian',
-            get_cmd=partial(self._get_setpoints, 'x', 'y', 'z'),
-            set_cmd=self._set_cartesian,
+            get_cmd=partial(self._get_setpoints, ['x', 'y', 'z']),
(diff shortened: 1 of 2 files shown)


jenshnielsen:
fixes #1143

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -570,90 +570,88 @@ def __init__(self, name, instrument_x, instrument_y,
         # Get and set parameters for the set points of the coordinates
         self.add_parameter(
             'cartesian',
-            get_cmd=partial(self._get_setpoints, 'x', 'y', 'z'),
-            set_cmd=self._set_cartesian,
+            get_cmd=partial(self._get_setpoints, ['x', 'y', 'z']),
```
jenshnielsen: Mutable default arguments to functions always makes me slightly nervous. These are never modified so you could make them tuples right?

sohailc: mutable defaults? Where do you see that? I guess I can change that to a tuple without problems

sohailc: Or will partial create a function with a list as default arguments? I did not think of that... :-)

jenshnielsen: @sohailc yes that's my thinking, It's a small issue for sure since this never modified the input arg

PR #1164, opened 2018-06-27
Fix maximum frequency for SRS860 - it is 500 kHz

jenshnielsen:
Cherrypicked from #1160


jenshnielsen:
Appveyer failure is due to a timeout not related and fixed by https://github.com/QCoDeS/Qcodes/pull/1166

PR #1168, opened 2018-06-29
Add a wait_till_complete() function to the VisaInstrument class

rohit19081995:
Prior to this change, the instrument timeout needed to be changed for long operations.
This is not ideal since a large timeout may hinder debugging.

This change adds a function which checks the ESR to make sure the operation is complete.

@jenshnielsen

### DIFF
--- a/qcodes/instrument/visa.py
+++ b/qcodes/instrument/visa.py
@@ -2,6 +2,7 @@
 from typing import Sequence
 import warnings
 import logging
+import time
 
 import visa
 import pyvisa.constants as vi_const
@@ -231,3 +232,26 @@ class VisaInstrument(Instrument):
         snap['timeout'] = self.timeout.get()
 
         return snap
+
+    def wait_till_complete(self, check_after = 5):
+        """
+        WARNING: This function repeatedly clears the Event Status Register (ESR).
+
+        The *OPC? command is used to query the machine when it completes an operation.
+        However, in some cases the operation is too long. There are 2 ways of dealing with this.
+
+        1. Change the timeout to an appropriate value (if you can determine how long the operation will take.)
+           This may cause issues while debugging.
+
+        2. Use the following function which asynchronously queries the Event Status Register and checks if the last bit (the OPC bit) is set.
+        """
+
+        try:
+            log.debug("Clearing ESR")
+            self.ask_raw("*ESR?")
+            self.write_raw("*OPC")
+            time.sleep(check_after)
+            while int(self.ask_raw("*ESR?").strip())%2==0:
+                    time.sleep(0.1)
+        except KeyboardInterrupt:
+            raise Exception("{} was interrupted during operation".format(self.name))


jenshnielsen:
This seems sensible to me. In my experience there are several instruments that do not implement the opc and esr correctly e.g. `*OPC?` returns immediately regardless of the status of the device presumably this will have the same issue but looks useful for correctly implemented instruments

WilliamHPNielsen:
Just for completeness, I'd like to add that the timeout can be dynamically changed. Some drivers do this when e.g. capturing long traces. Here is an example of that: https://github.com/QCoDeS/Qcodes/blob/dadc1aba6c8fa4f6b2921114719d8e530c973583/qcodes/instrument_drivers/Keysight/Keysight_34465A.py#L71

qcodes/instrument/visa.py
```diff
@@ -231,3 +232,26 @@ def snapshot_base(self, update: bool=False,
         snap['timeout'] = self.timeout.get()
 
         return snap
+
+    def wait_till_complete(self, check_after = 5):
```
astafan8: shall we use plain English, `wait_until_complete`?

qcodes/instrument/visa.py
```diff
@@ -231,3 +232,26 @@ def snapshot_base(self, update: bool=False,
...
+        """
```
astafan8: please, PEP8 the docstring

qcodes/instrument/visa.py
```diff
@@ -231,3 +232,26 @@ def snapshot_base(self, update: bool=False,
...
+        2. Use the following function which asynchronously queries the Event Status Register and checks if the last bit (the OPC bit) is set.
+        """
+
+        try:
+            log.debug("Clearing ESR")
+            self.ask_raw("*ESR?")
+            self.write_raw("*OPC")
+            time.sleep(check_after)
+            while int(self.ask_raw("*ESR?").strip())%2==0:
+                    time.sleep(0.1)
```
astafan8: could you make a kwarg out of this value for flexibility?

qcodes/instrument/visa.py
```diff
@@ -231,3 +232,26 @@ def snapshot_base(self, update: bool=False,
...
+        except KeyboardInterrupt:
```
astafan8: what is this `try-except` needed for? Wouldn't it be obvious from the call stack (that is displayed upon exception raise) that the instrument has been interrupted while in `wait_till_complete` method?

qcodes/instrument/visa.py
```diff
@@ -231,3 +232,26 @@ def snapshot_base(self, update: bool=False,
...
+
+    def wait_till_complete(self, check_after = 5):
+        """
+        WARNING: This function repeatedly clears the Event Status Register (ESR).
+
+        The *OPC? command is used to query the machine when it completes an operation.
+        However, in some cases the operation is too long. There are 2 ways of dealing with this.
+
+        1. Change the timeout to an appropriate value (if you can determine how long the operation will take.)
+           This may cause issues while debugging.
```
astafan8: what issues may it cause exactly? I'm not an expert (yet) in visa instruments, but so far as I see, the inconvenience of this "way" is that you are changing the timeout value for all the commands, while what you want is just make it longer for a very particular command. With that said, I think the docstring can be improved to be more explicit.

astafan8:
@rohit19081995 any chance you'll finish this PR? do you need help/assistance in completing it?

WilliamHPNielsen:
While we wait for that, here's a relevant document from R&S recommending the practice introduced in this PR: https://cdn.rohde-schwarz.com/pws/dl_downloads/dl_application/application_notes/1gp79/1GP79_1E_SCPI_Programming_Guide_SigGens.pdf

PR #1169, opened 2018-07-02
Refactor Keysight 344xxA code and small fixes

YakBizzarro:
Fix code for Keysight 34461A and split the driver depending on the model, like in #772 
Also it automatically detect if the digitizer option is installed.


qcodes/instrument_drivers/Keysight/private/Keysight_344xxA.py
```diff
@@ -0,0 +1,456 @@
...
+        self._instrument.display_clear()
+
+        return numvals
+
+
+class _Keysight_344xxA(VisaInstrument):
+    """
+    Instrument class for Keysight 34460A, 34461A, 34465A and 34470A multimeters.
+
+    Tested with: 34461A, 34465A.
```
astafan8: is it possible to test the driver on 34460A and 34470A as well?

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA.py
```diff
@@ -0,0 +1,456 @@
...
+
+        self.add_parameter('sample_timer_minimum',
+                           label='Minimal sample time',
+                           get_cmd='SAMPle:TIMer? MIN',
+                           get_parser=float,
+                           unit='s')
+
+        # SYSTEM
+        self.add_parameter('error',
+                           label='Error message',
```
astafan8: according to `errorparser` implementation, this "Parameter" returns not only the message, but also the error code. In order to avoid confusion with the usage of this Parameter, and to avoid abuse of Parameter class as such, could you please make this a method of this class (basically making `errorparser` a method)?

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA.py
YakBizzarro: No, I have only a 34461A. The 34465A was tested by the original author.
Anyway they should work, they are very similar

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA.py
YakBizzarro: No problem

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA.py
```diff
@@ -186,6 +170,12 @@ def __init__(self, name, address, utility_freq=50, silent=False,
         ####################################
         # PARAMETERS
 
+        self.add_parameter('line_frequency',
```
astafan8: should the values of this parameter be limited to `vals=vals.Enum(50, 60)`?

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA.py
```diff
@@ -0,0 +1,455 @@
...
+        self.ranges = ranges[self.model]
+        self.NPLC_list = PLCs[self.model]
+
+        if DIG:
+            self._apt_times = apt_times[self.model]
+
+        ####################################
+        # PARAMETERS
+
+        self.add_parameter('line_frequency',
```
astafan8: Shall the values of this parameter be limited to `vals=vals.Enum(50, 60)`?

YakBizzarro: I don't think, it has only a get, you can't set it. Do we need validator in this case?

astafan8: oh, no, but then please add `set_cmd=False` to the `add_parameter` call - this way you show that setting is indeed not possible. You could also add a `dosctring=` argument where you explain that "the instrument returns the frequency of the grid that it is plugged into" or smth.

astafan8: btw, sorry for misinformation, set_cmd is False by default, so no need for explicitly adding it.

YakBizzarro: True, I added it.

YakBizzarro: Something else for merging?

WilliamHPNielsen:
@YakBizzarro, no more needed, I think we're good to merge. I'll just wait for CI.

YakBizzarro:
Great, thank you!

PR #1171, opened 2018-07-04
Keithley 2400 protect read commands

### LINKED ISSUE #1149: Keithley 2400 does not get added to the station cleanly
The ":read:" command and possibly others does not work when output is off but fails with an error. This is called when getting volt and current are snapshotted 

We should wrap these calls in checking that output is off

jenshnielsen:
In the Keithley 2400 doing a read with output disabled is an error that leaves the instrument in a state where no commands can be sent to it without restart

To overcome this we check that output is on before doing a read. If not we return 'nan' as a result
Fixes #1149

### DIFF
--- a/qcodes/instrument_drivers/tektronix/Keithley_2400.py
+++ b/qcodes/instrument_drivers/tektronix/Keithley_2400.py
@@ -38,14 +38,27 @@ class Keithley_2400(VisaInstrument):
                            get_parser=self._volt_parser,
                            set_cmd=':SOUR:VOLT:LEV {:.8f}',
                            label='Voltage',
-                           unit='V')
+                           unit='V',
+                           docstring="Sets voltage in 'VOLT' mode. "
+                                     "Get returns measured voltage if "
+                                     "sensing 'VOLT' otherwise it returns "
+                                     "setpoint value. "
+                                     "Note that it is an error to read voltage with "
+                                     "output off")
 
         self.add_parameter('curr',
                            get_cmd=self._get_read_output_protected,
                            get_parser=self._curr_parser,
                            set_cmd=':SOUR:CURR:LEV {:.8f}',
                            label='Current',
-                           unit='A')
+                           unit='A',
+                           docstring = "Sets current in 'CURR' mode. "
+                                        "Get returns measured current if "
+                                        "sensing 'CURR' otherwise it returns "
+                                        "setpoint value. "
+                                        "Note that it is an error to read current with "
+                                        "output off")
+
 
         self.add_parameter('mode',
                            vals=Enum('VOLT', 'CURR'),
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/tektronix/Keithley_2400.py
```diff
@@ -77,11 +77,30 @@ def __init__(self, name, address, **kwargs):
                            label='Current integration time')
 
         self.add_parameter('resistance',
-                           get_cmd=':READ?',
+                           get_cmd=self._get_volt_and_curr,
                            get_parser=self._resistance_parser,
                            label='Resistance',
                            unit='Ohm')
 
+    def _get_volt_and_curr(self) -> str:
```
astafan8: Why is the function called “volt and curr “ while READ returns 4 values? Should the name of this function be more generic?

jenshnielsen:
There is something unusual in the implementation of this visa instrument not specific to QCoDeS or pyvisa. If you do a visa read command with no preceding write command you get the result of `:READ?` with output enabled an an error with output enabled. This looks like the same output as `:READ?`. Normally I would expect a timeout for such a read i.e. when the output buffer is empty

The error is as expected that the command is not supported with output off but it still locks up the system. (The error can be seen after manually disconnecting from the instrument front panel and reconnecting )

qcodes/instrument_drivers/tektronix/Keithley_2400.py
```diff
@@ -77,11 +77,30 @@ def __init__(self, name, address, **kwargs):
...
+        output = self.output.get_latest()
+        if output is None:
+            # if get_latest returns None we have
+            # to ask the instrument for the status of output
+            output = self.output.get()
+
+        if output == 1:
+            msg = self.ask(':READ?')
+        else:
+            msg = "nan,nan,nan,nan,nan"
```
spxtr: Why not raise an exception here?

jenshnielsen: Yes I thought about that. My main motivation for not doing that was to prevent adding it to the station from triggering an exception

spxtr: My worry is that if you add it to a station and do some measurements you might miss the fact that it's not actually recording any data. I don't think there's a perfect solution, though I would prefer an exception.

astafan8: @jenshnielsen, sorry, I did not understand what you meant. Despite that I tend to agree with @spxtr.

Do you think it is enough to add a note to the driver docstring (or docstring of each affected parameter) that "output" should be on before reading values, otherwise an exception will be thrown?

jenshnielsen: To be more clear. I think it's fine for the get to throw when called directly, and yes documenting is a good idea. 

I was more concerned with the automatic get that happens at setup time but the snapshot will catch that exception anyway so either way should be fine.

astafan8:
nice! if everyone agrees now, please merge @jenshnielsen

PR #1176, opened 2018-07-05
Fix validation of steps

### LINKED ISSUE #1174: Parameter sets value when value is out of range
### Steps to reproduce
```
from qcodes import Parameter
from qcodes.utils.validators import Numbers

the_value = -10

def set_function(value):
    global the_value
    the_value = value
    print('set {0}'.format(the_value))

def get_function():
    print('get {0}'.format(the_value))
    return the_value

a = Parameter('test', set_cmd=set_function, get_cmd=get_function, vals=Numbers(0, 100), step=5)
a.set(10)
a.get()
```

### Expected behaviour
```
a.set(10) --> ValueError('-10 is invalid')
a.get() --> -10
```
### Actual behavior
```
a.set(10) --> ValueError('-5 is invalid')
a.get() --> -5
```
### System
**operating system**
Windows 10 x64, miniconda v 4.5.4 (venv)

**qcodes branch**
master

**qcodes commit**
8970a083c2b582e4d966da082f039fef49f2089a

@dpfranke, @eendebakpt

jenshnielsen:
Fixes #1174

Ensure that intermediate steps are validated too. In general we cannot assume that a step is valid just because the endpoint is

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -415,6 +415,9 @@ class _BaseParameter(Metadatable):
                 steps = self.get_ramp_values(value, step=self.step)
 
                 for step_index, val_step in enumerate(steps):
+                    # even if the final value is valid we may be generating
+                    # steps that are not so validate them too
+                    self.validate(val_step)
                     if self.val_mapping is not None:
                         # Convert set values using val_mapping dictionary
                         raw_value = self.val_mapping[val_step]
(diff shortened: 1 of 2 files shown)


jenshnielsen:
This failed because the number of validation calls generated have changed. Intermediate steps are validated after setting with the following logic.         
```python
self._save_val(val_step,
    validate=(self.val_mapping is None and
                   self.set_parser is None and
                   not(step_index == len(steps)-1 or
                   len(steps) == 1)))
```
I am not sure I see the logic of that if we validate them up front too. Perhaps we should get rid of it

PR #1179, opened 2018-07-06
[WIP] Yokogawa GS200 cleanup

jenshnielsen:
Remove internal cache. The parameters already implement this so there is no need for duplicated effort

In addition the hard coding of mode at startup is wrong now after the merge of #1100 

WIP because I need an actual instrument to test with before this is ready for review


jenshnielsen:
replaced by #2143

PR #1183, opened 2018-07-09
[WIP] Alazar 9870 use regular parameter

jenshnielsen:
@wpfff This is not tested but should be a useful starting point for getting the 9870 working the same way an the Alazar 9360


jenshnielsen:
on top of #1182

jenshnielsen:
replaced by #1290

PR #1184, opened 2018-07-09
Alazar remove AlazarParameter support

jenshnielsen:
This is only to be merged once we are happy with #1183 and sure that we do not need to still support the Alazar paramter. This still keeps the deprecated Config function


qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -753,7 +697,7 @@ def acquire(self, mode=None, samples_per_record=None,
         bytes_per_record = bytes_per_sample * samples_per_record
 
         # channels
-        channels_binrep = self._get_raw_or_bytes(self.channel_selection)
+        channels_binrep = self.channel_selection.raw_value
         number_of_channels = self.get_num_channels(channels_binrep)
```
ThorvaldLarsen: There is some repeating code above and below here. Maybe we can fix that in the same round?

jenshnielsen: Sure

ThorvaldLarsen: Thanks!

jenshnielsen:
@astafan8 I have tested this with the Alazar 9360 driver. I updated the Alazar 9360 notebook slightly and removed the out of date notebook. There is certainly a lot more work that can be done on the Alazar drivers. I would prefer to do that in the next pr. It would be great if you can test this with the other cards. There is a 9870 notebook which is still out of date. That one should be deleted or updated to use the current api as part of this pr I think

astafan8:
@jenshnielsen tested successfully on 9870 and 9373, haven't noticed anything wrong. i think, it's ready for merging.

astafan8:
what is `Qcodes example ATS_ONWORK` notebook for?

jenshnielsen:
It's leftover. It contained nothing useful that was not in the other notebook so I deleted it

PR #1185, opened 2018-07-09
Update Alazar notebooks

jenshnielsen:
Update Alazar 9360 notebook with newest api and remove the ONWORK which is nearly identical and is out of date


jenshnielsen:
This is now done as part of #1184

PR #1186, opened 2018-07-10
Fix pyvisa 1.9.0 bug with query_binary_values timing out

astafan8:
For pyvisa 1.9.0, query_binary_values calls time out. pyvisa 1.8 does not have this problem. This PR fixes this. See more info in [pyvisa github, issue 348](https://github.com/pyvisa/pyvisa/issues/348)

Once the pyvisa is fixed, this commit can be reverted.

(note, the implementation is sort of dirty because of time constraints)

### DIFF
--- a/qcodes/instrument_drivers/Keysight/Infiniium.py
+++ b/qcodes/instrument_drivers/Keysight/Infiniium.py
@@ -9,6 +9,10 @@ from qcodes import InstrumentChannel, ChannelList
 from qcodes import ArrayParameter
 from qcodes.utils.validators import Enum, Numbers
 
+import visa
+
+
+need_to_fix_visa_bug = visa.__version__ == '1.9.0'
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


Dominik-Vogel:
I think it would be nice to define the workaround in some place and import it in the two relevant drivers, so that we don't have two copies of it floating around.
The problem really screams for a context manager, could you try if something like this works:
```
from contextlib import contextmanager

need_to_fix_visa_bug = visa.__version__ == '1.9.0'

@contextmanager
def pyvisa_query_binary_workaround(visa_handle):
    if need_to_fix_visa_bug:
        original = visa_handle._read_termination
        visa_handle._read_termination = None
        yield
        visa_handle._read_termination =original
    else:
        yield
```
and then in the instruments:
````
from somewhere import pyvisa_query_binary_workaround
...
with pyvisa_query_binary_workaround(self._parent.visa_handle):
    data = instr._parent.visa_handle.query_binary_values(
             'WAV:DATA?', datatype='h', is_big_endian=False)
             'WAV:DATA?', datatype='h', is_big_endian=False)
...
````

astafan8:
@Dominik-Vogel I've implemented the fix using a context manager. It looks similar to the one that you've suggested, the only difference is that mine is based on a generic one. Both context managers are covered with tests. If you approve this, I'll merge.

PR #1188, opened 2018-07-16
add M2j driver

peendebak:
@jenshnielsen @WilliamHPNielsen @Christian-Volk


qcodes/instrument_drivers/QuTech/M2j.py
```diff
@@ -0,0 +1,83 @@
...
+        Args:
+            name (str): name of the instrument.
+
+            spi_rack (SPI_rack): instance of the SPI_rack class as defined in
+                the spirack package. This class manages communication with the
+                individual modules.
+
+            module (int): module number as set on the hardware.
+
+        The gain and RF_level can only be set to the device, but not read from the device.
```
astafan8: according to the code, `RF_level` cannot be "set", only "get" - who's lying, code or this comment? :)

qcodes/instrument_drivers/QuTech/M2j.py
```diff
@@ -0,0 +1,83 @@
...
+
+        Parameters:
+            gain: Amplifier gain in dB, range 32 to 55 dB
```
astafan8: I recommend that you add these descriptions as `docstring` argument of the `add_parameter` method.

qcodes/instrument_drivers/QuTech/M2j.py
```diff
@@ -0,0 +1,83 @@
...
+                          call_cmd=self.m2j.rf_clipped)
+
+    def _set_gain(self, dB):
+        if dB < 32:
+            print(
+                "notice, lower limit for the gain is 32dBm. The limit will be the set value.")
+            dB = 32
+        if dB > 55:
+            print(
+                "notice, upper limit for the gain is 55dBm. The limit will be the set value.")
```
astafan8: hmm, the print statement says "the limit will be the set value" but `dB = 55` assignment is not there, while it is for the `dB = 32` case. What should be here?

qcodes/instrument_drivers/QuTech/M2j.py
```diff
@@ -0,0 +1,83 @@
...
+
+        self.add_function('clear_rf_clip',
+                          call_cmd=self.m2j.clear_rf_clip)
+        self.add_function('is_rf_clipped',
+                          call_cmd=self.m2j.rf_clipped)
+
+    def _set_gain(self, dB):
+        if dB < 32:
+            print(
+                "notice, lower limit for the gain is 32dBm. The limit will be the set value.")
```
astafan8: what do you mean with "the limit will be the set value"? So far as I read it, it means that the limit is going to be changed, which is not true. I would rephrase this to `f"An attempt is made to set the gain to {dB} which is below 32dBm. This is not possible. The gain will be set to 32 dBm."`

Also, a better design is to add the limits to the `add_parameter` function via `vals` argument like `vals=Numbers(min_value=32, max_value=55)` ([see for example this driver for an example](https://github.com/QCoDeS/Qcodes/blob/fb574ef1cf2b7a76cdab5d6591cb4f7a5eced376/qcodes/instrument_drivers/stanford_research/SR86x.py#L619))

qcodes/instrument_drivers/QuTech/M2j.py
```diff
@@ -0,0 +1,83 @@
...
+                           unit='dBm')
+
+        self.add_function('clear_rf_clip',
+                          call_cmd=self.m2j.clear_rf_clip)
+        self.add_function('is_rf_clipped',
+                          call_cmd=self.m2j.rf_clipped)
+
+    def _set_gain(self, dB):
+        if dB < 32:
+            print(
```
astafan8: if you insist on keeping this logic, please use `warning.warn` instead of `print`.

qcodes/instrument_drivers/QuTech/M2j.py
```diff
@@ -0,0 +1,83 @@
...
+            ref_scale = 4095
+
+        self.m2j.set_gain(ref_scale)
+
+    def _meas_power(self):
+        """ Measure the power and convert it to dBm. Calibrated using an R&S
+         SMA100 source. Linear relation between set power and measured data.
+         Measurement range -80 to -40 dBm.
+        """
+        x = self.m2j.get_level()
```
astafan8: could you please extract this line out of this method so that the method does only "conversion to dBm"? so, i'd implement it like this:
```python
def _meas_power(self):
    return self._convert_power_to_dbm(self.m2j.get_level())

def _convert_power_to_dbm(self, power_level):
    # the relation is liner as calibrate with blabla in range blabla....
    return power_level
```

Actually this can be even better - now you can get rid of `_meas_power` and add `self._convert_power_to_dbm(self.m2j.get_level())` directly as "get" command to `add_parameter` for `RF_level` parameter.

qcodes/instrument_drivers/QuTech/M2j.py
```diff
@@ -0,0 +1,83 @@
...
+
+        self.m2j = M2j_module.M2j_module(spi_rack, module)
+
+        self.add_parameter('gain',
+                           label='gain',
+                           set_cmd=self._set_gain,
+                           unit='dB',
+                           vals=Numbers())
+
+        self.add_parameter('RF_level',
```
astafan8: Which of the two names reflects the value returned by the driver: "RF_level" or "power"? I have a feeling that the answer is "RF_level". If so, why do you call the method `_meas_power` instead of `_meas_rf_level`?

qcodes/tests/drivers/test_m2j.py
```diff
@@ -0,0 +1,35 @@
+import sys
+import unittest
+from unittest.mock import patch, MagicMock, call
+
+sys.modules['spirack'] = MagicMock(name='spirack')
+from spirack import M2j_module, SPI_rack
+from qcodes.instrument_drivers.QuTech.M2j import M2j
+
+
+class TestZIHDAWG8(unittest.TestCase):
```
peendebak:
```suggestion
class TestM2j(unittest.TestCase):
```

qcodes/tests/drivers/test_m2j.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def test_gain(self):
+        spi_rack = MagicMock()
+        m2j = M2j('test', spi_rack, 42)
+
+        with self.assertRaises(ValueError):
+            m2j.gain.set(32)
```
peendebak:
```suggestion
            gain_too_low = 32
            m2j.gain.set(gain_too_low )
```

qcodes/tests/drivers/test_m2j.py
```diff
@@ -0,0 +1,35 @@
...
+        with self.assertRaises(ValueError):
+            m2j.gain.set(56)
+
+        for gain in range(33, 56):
+            m2j.gain.set(gain)
```
peendebak: If you test many values to be set, then also set at least 1 float instead of only integers.

qcodes/instrument_drivers/QuTech/M2j.py
```diff
@@ -0,0 +1,71 @@
...
+                           docstring='Measured RF power after amplification '
+                                     '(not calibrated)')
+
+        self.add_function('clear_rf_clip',
+                          call_cmd=self.m2j.clear_rf_clip)
+        self.add_function('is_rf_clipped',
+                          call_cmd=self.m2j.rf_clipped)
+
+    def _set_gain(self, gain):
+        a = 1024.45
```
astafan8: does it make sense to make these constants private attributes of the class so that they can be easily modified when needed (the same comment for the magic numbers `0` and `4095` below)?

peendebak:
@astafan8 Driver has been tested on hardware. We also updated the constants to be private variables.

PR #1190, opened 2018-07-16
update S5i driver

peendebak:
- Improve docs
- Add conversion functions

@jenshnielsen @WilliamHPNielsen @Dominik-Vogel @Christian-Volk

### DIFF
--- a/qcodes/instrument_drivers/QuTech/S5i.py
+++ b/qcodes/instrument_drivers/QuTech/S5i.py
@@ -20,34 +20,90 @@ class S5i(Instrument):
             individual modules.
 
         module (int): module number as set on the hardware.
+
+    Parameters:
+        enabled: switched output on/off
+        frequency: Frequency on Hz, range 40 MHz - 4 GHz, stepsize 1 MHz
+        freq_fine: Frequency on Hz, range 40 MHz - 4 GHz, stepsize 10 kHz.
+        Difference between frequency and freq_fine unclear!?
+        power: power in dBm, range -20 to +15 dBm
+
     """
 
     def __init__(self, name, spi_rack, module, **kwargs):
         super().__init__(name, **kwargs)
 
-        self.s5i = S5i_module(spi_rack, module)
-
-        self.add_parameter('use_external_reference',
-                           label='Use external reference',
-                           get_cmd=self._use_external_reference,
-                           set_cmd=self.s5i.use_external_reference,
+        self.s5i = S5i_module(spi_rack, module, frequency=41e6,
+                              enable_output=1, output_level=0)
+        # Put low freq as it is unlikely it will be resonant with something.
+        self.add_parameter('enabled',
+                           label='RF_on/off',
+                           set_cmd=self.s5i.enable_output_soft,
                            vals=Bool())
-
-        self.add_parameter('frequency_stepsize',
-                           label='Frequency stepsize',
-                           get_cmd=self._get_stepsize,
-                           set_cmd=self.s5i.set_stepsize,
-                           units='Hz',
-                           vals=Numbers())
 
         self.add_parameter('frequency',
                            label='Frequency',
                            get_cmd=self._get_rf_frequency,
                            set_cmd=self.s5i.set_frequency,
-                           units='Hz',
+                           unit='Hz',
                            vals=Numbers())
 
-        self.add_function('optimize_for_frequency', call_cmd=self._optimize)
+        self.add_parameter('freq_fine',
+                           label='Frequency fine',
+                           get_cmd=self._get_rf_frequency,
+                           set_cmd=self.s5i.set_frequency_optimally,
+                           unit='Hz',
+                           vals=Numbers())
+
+        self.add_parameter('power',
+                           label='Power',
+                           unit='dBm',
+                           set_cmd=self._set_power,
+                           vals=Numbers())
+
+    def _set_power(self, dbm):
+        self.s5i.set_output_power(self.convert_dbm_to_ref_scale(dbm))
+
+    @staticmethod
+    def convert_ref_scale_to_dbm(ref_scale):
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -20,34 +20,88 @@ class S5i(Instrument):
             individual modules.
 
         module (int): module number as set on the hardware.
+
+    Parameters:
+        enabled: switched output on/off
+        frequency: Frequency on Hz, range 40 MHz - 4 GHz, stepsize 1 MHz
+        freq_fine: Frequency on Hz, range 40 MHz - 4 GHz, stepsize 10 kHz.
+        Difference between frequency and freq_fine unclear!?
```
astafan8: is the difference still unclear? So far as i understand, the difference is in the step size - one is smaller than the other one. So, if the difference is clear, please, remove the line.

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -20,34 +20,88 @@ class S5i(Instrument):
...
     def __init__(self, name, spi_rack, module, **kwargs):
         super().__init__(name, **kwargs)
 
-        self.s5i = S5i_module(spi_rack, module)
-
-        self.add_parameter('use_external_reference',
-                           label='Use external reference',
-                           get_cmd=self._use_external_reference,
-                           set_cmd=self.s5i.use_external_reference,
+        self.s5i = S5i_module(spi_rack, module, frequency=41e6,
```
astafan8: is this change being introduced because a new version of `spirack` has been released? if so, shall we make it clear somewhere (perhaps in requirements.txt or environment.yml, or just in the docstring of this driver) that a particular version of `spirack` is required?

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -20,34 +20,88 @@ class S5i(Instrument):
...
+        self.add_parameter('power',
+                           label='Power',
+                           unit='dBm',
+                           set_cmd=self._set_power,
+                           vals=Numbers())
+
+    def _set_power(self, dbm):
+        self.s5i.set_output_power(self.convert_dbm_to_ref_scale(dbm))
+
+    def convert_ref_scale_to_dbm(self, ref_scale):
```
astafan8: why is this method not used? does it make sense for the `power` parameter to have a `_get_power` method that uses `convert_ref_scale_to_dbm`?

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -20,34 +20,88 @@ class S5i(Instrument):
...
-                           set_cmd=self.s5i.use_external_reference,
+        self.s5i = S5i_module(spi_rack, module, frequency=41e6,
+                              enable_output=1, output_level=0)
+        # Put low freq as it is unlikely it will be resonant with something.
+        self.add_parameter('enabled',
+                           label='RF_on/off',
+                           set_cmd=self.s5i.enable_output_soft,
                            vals=Bool())
 
-        self.add_parameter('frequency_stepsize',
```
astafan8: if this parameter is removed (actually, is it removed because there is not need for setting the step size separately?), shall the `_get_stepsize` method also be removed?

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -20,34 +20,88 @@ class S5i(Instrument):
...
+        power: power in dBm, range -20 to +15 dBm
+
     """
 
     def __init__(self, name, spi_rack, module, **kwargs):
         super().__init__(name, **kwargs)
 
-        self.s5i = S5i_module(spi_rack, module)
-
-        self.add_parameter('use_external_reference',
```
astafan8: if this parameter is removed (actually, why is it removed?), shall the `_use_external_reference` method also be removed?

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -20,34 +20,88 @@ class S5i(Instrument):
...
                            set_cmd=self.s5i.set_frequency,
-                           units='Hz',
+                           unit='Hz',
                            vals=Numbers())
 
-        self.add_function('optimize_for_frequency', call_cmd=self._optimize)
+        self.add_parameter('freq_fine',
+                           label='Frequency fine',
+                           get_cmd=self._get_rf_frequency,
+                           set_cmd=self.s5i.set_frequency_optimally,
```
astafan8: just to confirm that my understanding is correct, is `s5i.set_frequency_optimally` just capable of setting the frequency with smaller steps or is there indeed some "optimization" involved? If the "optimization" is indeed involved, then I would recommend that the parameter be called differently (e.g. smth with "optimal"). 

Or, better, since the `frequency` parameter and `freq_fine` have the same "get" command, I would make just a public method out of `freq_fine` called `set_frequency_optimally` (or `set_frequency_finer`, or smth better) which will avoid the confusion about `freq_fine` being a separate/independent parameter which it is not.

qcodes/instrument_drivers/QuTech/S5i.py
peendebak: No, the spirack driver is the same. But we choose to have different default settings. The `spirack` is not yet in the `requirements.txt`, since the spirack is optional. Is there another way to specify version requirements?

astafan8: well, i do not think we have a good approach to this. Ideally every python module has its setup.py file where the required packages and their versions are listed. But at the moment all drivers are part of qcodes, hence they don't have their independent setup.py files.

perhaps, in the exception that is thrown in case `spirack` is not installed, you can mention the version requirements like "install spirack >=1.2.3".

PR #1191, opened 2018-07-16
Quick fix for mypy 0.620 behavior

astafan8:
With the new release of mypy 0.620, the mypy part of the test started failing ([an example here](https://travis-ci.org/QCoDeS/Qcodes/builds/404452622?utm_source=github_status&utm_medium=notification)):
```
qcodes/instrument/parameter.py:1046: error: The type alias is invalid in runtime context
qcodes/instrument/parameter.py:1052: error: The type alias is invalid in runtime context
qcodes/instrument/parameter.py:1055: error: The type alias is invalid in runtime context
qcodes/instrument/parameter.py:1059: error: The type alias is invalid in runtime context
qcodes/instrument/parameter.py:1224: error: The type alias is invalid in runtime context
qcodes/instrument/parameter.py:1229: error: The type alias is invalid in runtime context
qcodes/instrument/parameter.py:1233: error: The type alias is invalid in runtime context
qcodes/instrument/parameter.py:1237: error: The type alias is invalid in runtime context
```

~I did not want to spend a time on figuring out why, so here is the fix that disables this new version.~

See more on mypy github https://github.com/python/mypy/issues/5354 - it might be related.

The edits in the code make the type objects be type-annotated which makes the new mypy 0.620 happy 😄 👍 🥇

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -59,7 +59,8 @@ import logging
 import os
 import collections
 import warnings
-from typing import Optional, Sequence, TYPE_CHECKING, Union, Callable, List, Dict, Any, Sized, Iterable
+from typing import Optional, Sequence, TYPE_CHECKING, Union, Callable, List, \
+    Dict, Any, Sized, Iterable, Type
 from functools import partial, wraps
 import numpy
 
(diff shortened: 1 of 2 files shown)


astafan8:
@jenshnielsen , you're right ...

PR #1195, opened 2018-07-18
Make a folder for unsupported code

WilliamHPNielsen:
Okay, so let's talk a bit about restructuring the code base.

This PR introduces a new folder, `qcodes/unsupported` which is meant for code that the core team will not maintain. Whereas the rest of QCoDeS comes with some promise of not being intentionally broken and having known bugs fixed as soon as the core team can get around to it, the code in `unsupported` does not. 

What I envision is that drivers for instruments that we (the core team) do not have access to should go there, because we can not give any guarantees about whether they work or not. Also, I would like to see the `Loop` and its associated (old) dataset go in there, since that will give us the much needed freedom to change how `Parameters` work.

I think perhaps it's good to have some discussion about this?

@QCoDeS/core-contributors 

The way I see it, we can can not make promises about the drivers below (just listed by vendor for easy overview). I am not too sure about all of them and I may have missed some. What do you think?

- Advantech
- QuTech
- Spectrum
- cryocon
- cryogenic
- signal_hound

### DIFF
--- a/qcodes/unsupported/README.rst
+++ b/qcodes/unsupported/README.rst
@@ -1,0 +1,12 @@
+Unsupported
+===========
+
+Welcome to the unsupported part of QCoDeS. Entrance at own risk.
+
+This folder - which might eventually become its own repository - is for user-contributed code that
+the core developer team can not support or maintain.
+
+Examples of code that belongs in here include (but are not limited to):
+- Instrument drivers for instruments that the core team does not have access to
+- Legacy code that is incompatible with newer versions of QCoDeS
+


WilliamHPNielsen:
@astafan8 my preference would actually be 
`from qcodes.unsupported.instrument_drivers.mylab.mydriver import Driver`

 We should of course not move existing unsupported drivers without leaving a deprecation warning behind.

But what do you see as undesirable in simply moving and deprecating?

astafan8:
@WilliamHPNielsen yeah, you're right, a deprecation message will be enough.

nataliejpg:
agree with @astafan8 in terms of making the user notice as little difference as possible. Otherwise I'm fine with this. My only concern is how code get's into this and if by 'unsupported' you also mean 'unmanaged'? I just dont want to end up with there being multiple drivers for the same instrument or really obviously problematic prs being oked because it's going into the 'unsupported' part because that will render it basically unuseable and lost of the good code that's in there will also be avoided. I recognise that with the old loop and dataset it will become pretty much 'use at your own risk', the instrument drivers however it would be a shame to make less useable. In general i feel positively about this though (i realise it doesn't sound like that).

peendebak:
@WilliamHPNielsen A directory for unsupported code is fine I think. I have two remarks:

- Availability of drivers to the core team will change over time (core devs change, hardware will change). We do not want to move drivers between the normal `instrument_drivers` and `unsupported` directories.  (this will break code and confuse people).
- I expected the `qcodes.data.DataSet` to move to unsupported (I do not agree, since there is no proper replacement yet). However, what might be changed in the parameters? There is a lot of code depending on the parameters, so any change has to be done with care.

sohailc:
I think the `unsupported` drivers folder should be a temporary solution. Eventually, they should live in independent repo's like @astafan8 suggests. These drivers are not really unsupported, they are just unsupported by the qcodes core team. If another team wants to support them, that's fine! I understand that they need to be in the repo for the time being so we can make a deprecation message.

sohailc:
Before making any real changes, can you make a list of drivers you intend to place in the deprecation folder?

Dominik-Vogel:
@WilliamHPNielsen I just read this issue and got quite confused in the beginning and it seems like Sohail had the same confusion. The reason was the word 'unsupported'. An unsupported driver is too strongly associated with something that does not work.
'unmaintained' also sounds bad I think. 'community' is maybe an option?

We are however held in charge of maintaining the signalhound.

jenshnielsen:
@WilliamHPNielsen I think we can close this as all unsupported drivres have moved to qcodes-contrib-drivers

PR #1196, opened 2018-07-20
Improve MyPy support

jenshnielsen:
* Set setting in ini file
* Use warn_unused_ignores mad warn_redundant_casts to remove no longer needed workarounds
* Unpin Hypothesis which is now fixed. Fixes #1163 
* Rather than modifiying pyspcm source code ignore it from mypy

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -31,6 +31,6 @@ test_script:
   - cd tests
   - pytest
   - cd ..\..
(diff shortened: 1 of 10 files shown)


jenshnielsen:
They are all the places that mypy detects. With the setting in this ini file it should raise if there are unneeded casts or ignores

PR #1198, opened 2018-07-20
Remove functions from Instruments

### LINKED ISSUE #345: add_function docs unclear
The docstring for `instrument.base.add_function()` is very unclear. 

In particular the Args are unclear and the Returns is unclear. 
Args could be improved by adding `call_cmd` as an explicit (required) argument and explaining what it does. 
It is not clear when the return of this function is ever used. 

The docs of `instrument.function.function` are much clearer and seem to refer to the same thing. Exactly how this works/gets called is very opaque at the moment. 

Similar things can be said about `instrument.base.add_parameter()` however I have not looked into that with as much detail. 

Suggestions for improvement. 
1. explicit example
2. reference to how and when it should be used 
3. documentation in only one place, either in the `instrument.add_function` or in the `function.function`
4. reference the one place where the good docs are from the other place where people will look for it.

### LINKED ISSUE #741: Deprecate/remove instrument.base.add_function ?
Possibly related to #345 . 

Similar to the add_parameter, the instrument has a method `add_function`. To me the use case of this is very unclear and IMO opinion using regular python methods is preferred over using the add_function method to add a function as a method to an instrument. 

Advantages of using regular methods are
- more readable in the code/simpler 
- docstring generated are correct

As such I would propose to deprecate the use of the add_function method. 

(also now that Giulio left, who are all the remaining maintainers?)

### LINKED ISSUE #1197: Functions silently overwritten in Instrument
### Steps to reproduce

```python
from qcodes.instrument.function import Function


class MyInst(Instrument):
    
    def __init__(self, name):
        super().__init__(name)
        
        self.add_parameter('start', get_cmd="I'm a parameter")
        self.add_function('start', call_cmd="I'm a function")
        
    def write(self, cmd):
        print(cmd)
    
    def ask(self, cmd):
        print(cmd)
        
myinst = MyInst('myinst')

myinst.start()
```

### Expected behaviour
Since both functions and parameters get delegated as attributes in the same way, it should not be possible to add a parameter with the name of an existing function of vice versa.

### Actual behaviour
The function gets ignored in favour of the parameter. I found a comment indicating that this is by design https://github.com/QCoDeS/Qcodes/blob/cfc97ec2090ed3620634affa3177419dddb6abae/qcodes/tests/test_instrument.py#L75. I don't think this is such a good design decision.

### Proposed solution

Maybe just get rid of the functions altogether? 

### System
**operating system**
Windows 10

**qcodes branch**
master

**qcodes commit**
 cfc97ec2090ed3620634affa3177419dddb6abae

WilliamHPNielsen:
Fixes #1197, fixes #741, fixes #345

- Remove the entire notion of `functions` as attributes of an instrument

As the issues above show, functions are not implemented that well and are confusing. I don't think there is any real benefit to having them be a part of the snapshot either, since they don't have state that we want to record. Therefore, for the sake of having fewer features and less code, I propose that we sack them.

Work still pending:
- Go through drivers and put in good old bound methods instead of `add_function` calls.

Drivers in need of cleansing:
- [x] Newport AG_UC8
- [x] QuTech F1d
- [x] QuTech M2j
- [x] Rigol DG4000
- [x] Rohde&Schwarz ZNB
- [x] SRS SR830
- [x] SRS SR86x

@QCoDeS/core


sohailc:
There is one benefit of having functions: It creates a clear interface which will greatly aid when attempting to mock drivers. If we have a list of all functions which are publicly exposed to users, then creating a mock instrument driver through automated duck-typing becomes much easier.

astafan8:
also TODO: adjust the "how to write a driver" doc accordingly.

astafan8:
adding to @sohailc point, would anything break from the `yml` `VISA` `..@sim` side or would some things become impossible or inconvenient?

WilliamHPNielsen:
@sohailc I am not sure I understand. The current state is that we don't have a list of all functions that are exposed, only the few ones that the driver author (for whatever reason) chose to add as functions and not bound methods. So part of the "functions" of a given driver are available from `inst.functions` and the rest isn't. To make things worse, more complicated functions can't easily be squeezed into `inst.functions` (see the #741 discussion), so it's not a situation we can do much to remedy as things are now.
Do you then propose to go in the opposite direction of this PR and *extend* the `functions` framework to also handle complex functions? 

@astafan8 as far as I can see, there is no problem. The tests are also passing with no bother. I think this also shows that the `functions` don't really offer anything (except for more stuff to be confused about).

qcodes/instrument/channel.py
```diff
@@ -448,15 +445,6 @@ def __getattr__(self, name: str):
                                      setpoint_labels=setpoint_labels)
             return param
 
-        # Check if this is a valid function
-        if name in self._channels[0].functions:
```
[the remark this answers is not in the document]
WilliamHPNielsen: Yeah, that's true... Guess I got carried away by the sheer joy of deleting lines.

That's a bit of a bummer, actually. I wonder if anyone uses this, though...

qcodes/instrument/base.py
```diff
@@ -97,20 +94,22 @@ def add_function(self, name: str, **kwargs) -> None:
...
             KeyError: if this instrument already has a function with this
                 name.
         """
-        if name in self.functions:
+
+        warnings.warn("The use of 'add_function' has been deprecated. Please "
+                      "use a normal bound method instead.")
+
+        if name in self.parameters:
             raise KeyError('Duplicate function name {}'.format(name))
```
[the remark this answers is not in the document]
WilliamHPNielsen: True dat.

astafan8:
looks good. will be waiting for the drivers cleanup...

astafan8:
@WilliamHPNielsen perhaps another to-do item - remove this file https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument/function.py ?

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -225,8 +225,8 @@ def __init__(self, parent, name, channel, vna_parameter: str=None) -> None:
                            channel=n,
                            parameter_class=FrequencySweep)
 
-        self.add_function('autoscale',
-                          call_cmd='DISPlay:TRACe1:Y:SCALe:AUTO ONCE, "{}"'.format(self._tracename))
+    def autoscale(self) -> None:
+        self.write(f'DISPlay:TRACe1:Y:SCALe:AUTO ONCE, "{self.tracename}"')
```
astafan8: `self.tracename` - private one (with `_`) or public one?

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -451,6 +452,10 @@ def __init__(self, name: str, address: str, init_s_params: bool=True, **kwargs)
         self.rf_off()
         self.connect_message()
 
+    def reset(self) -> None:
+        self.write('*RST')
+
```
astafan8: double "new line"?

astafan8:
great job so far. i went through it in order to spot typos, but it looks ok.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
WilliamHPNielsen: Good catch, thanks.

WilliamHPNielsen:
@QCoDeS/core, any ideas about how to handle the functions that should have been parameters? I can take care of things in the SR8xx drivers, but the ZNB driver and especially the DG4000 are facing some pretty time-consuming refactoring.

astafan8:
But what is more time consuming: refactoring itself or testing with actual instruments that nothing got broken?

Dirty approach: for the troublesome drivers, you could extract the old `add_function` to `workarounds` module, say `_add_function__deprecated`, and update these drivers to use it, and make an issue to clean the drivers up properly later.

qcodes/instrument_drivers/stanford_research/SR86x.py
```diff
@@ -969,18 +964,24 @@ def __init__(self, name, address, max_frequency, reset=False, **kwargs):
         data_channels.lock()
         self.add_submodule("data_channels", data_channels)
 
-        # Interface
-        self.add_function('reset', call_cmd='*RST')
-
-        self.add_function('disable_front_panel', call_cmd='OVRM 0')
-        self.add_function('enable_front_panel', call_cmd='OVRM 1')
```
astafan8: no bound methods for these ones? :(

WilliamHPNielsen: Bro, read the commit message :)

The underlying API does not work. I even have an email from SRS confirming this.

astafan8: oh, thanks, sorry )

WilliamHPNielsen:
Now the Rigol DG4000 has been completely stripped of its `add_function` calls. Since I don't have access to such an instrument, it's not easy for me to verify that I did not break anything. @Rubenknex, as far as I can see, you are the original author of the driver. Would you mind checking that it still works? I'd really appreciate it.

Also, that driver predates the concept of `Channels` in QCoDeS, but now that we do have channels, it would not be a bad idea to "channelise" that driver. But that should be done by someone who actually uses the instrument.

WilliamHPNielsen:
@QCoDeS/core this is ready for review now.

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA.py
```diff
@@ -340,14 +340,30 @@ def __init__(self, name, address, silent=False,
...
+    def init_measurement(self) -> None:
+        """
+        Change the state of the triggering system from "idle" to
+        "wait-for-trigger", and clear the previous set of measurements from
+        reading memory
+        """
+        self.write('INIT')
+
+    def reset(self) -> None:
+        self.write('reset')
```
astafan8: not `*RST`?

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -114,10 +114,7 @@ def __init__(self, name, address, reset=False, numdacs=16, dac_step=10,
                            label='Dac voltages',
                            get_cmd=self._get_dacs)
 
-        self.add_function(
-            'trigger',
-            call_cmd=self._send_trigger
-        )
+        self.trigger = self._send_trigger
```
astafan8: isn't it better to do it on the class level? say, just add `trigger = _send_trigger` after `def _send_trigger(..):.....`?

qcodes/instrument_drivers/rigol/DG4000.py
```diff
@@ -77,7 +77,7 @@ def __init__(self, name, address, reset=False, **kwargs):
             ramp_freq = [5e6, 4e6, 3e6, 1e6][i]
             pulse_freq = [50e6, 40e6, 25e6, 15e6][i]
             harmonic_freq = [100e6, 80e6, 50e6, 30e6][i]
-            arb_freq = [50e6, 40e6, 25e6, 15e6][i]
+            self.arb_freq = [50e6, 40e6, 25e6, 15e6][i]
```
astafan8: `self.` should be added to all of these `*_freq` then, because they are all (?) used below in `_source_apply` calls.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -417,22 +423,37 @@ def __init__(self, name: str, address: str, init_s_params: bool=True, **kwargs)
...
-        self.add_function('display_dual_window',
-                          call_cmd='DISP:LAY GRID;:DISP:LAY:GRID 2,1')
-        self.add_function('rf_off', call_cmd='OUTP1 OFF')
-        self.add_function('rf_on', call_cmd='OUTP1 ON')
+
+        self.add_parameter(name='tooltip',
+                           set_cmd='SYSTem:ERRor:DISPlay {}',
+                           get_cmd='SYSTem:ERRor:DISPlay?',
+                           get_parser=int,
+                           val_mapping={"ON": 1, "OFF": 0})
```
astafan8: do i understand correctly that here both numbers and strings work for this SCPI command?

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -447,10 +468,80 @@ def __init__(self, name: str, address: str, init_s_params: bool=True, **kwargs)
...
             self.channels.autoscale()
 
-        self.update_display_on()
-        self.rf_off()
+        self.display_updates('ON')
+        self.rf_power(False)
         self.connect_message()
 
+    def rf_on(self) -> None:
+        warn("The function rf_on has been deprecated. Please use the "
```
astafan8: why didn't you used the `@deprecate` decorator from qcodes?

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -417,22 +423,37 @@ def __init__(self, name: str, address: str, init_s_params: bool=True, **kwargs)
...
+
+        self.add_parameter(name='display_layout',
+                           label='Instrument display layout mode',
+                           set_cmd='DISPlay:LAYout {}',
+                           get_cmd='DISPlay:LAYout?',
+                           val_mapping={'LINEUP': 'LIN\n',
+                                        'STACK': 'STAC\n',
+                                        'HORIZONTAL': 'HOR\n',
+                                        'VERTICAL': 'VER\n',
+                                        'GRID': 'GRID\n'})
```
astafan8: really? `\n` here? the instrument manufacturer is crazy :)

qcodes/tests/drivers/test_rigol_DG4000.py
```diff
@@ -0,0 +1,89 @@
...
+    yield kw_sim
+
+    kw_sim.close()
+
+
+def test_init(driver):
+    assert driver.IDN()['vendor'] == 'QCoDeS'
+
+    # test that what used to be implemented via add_function is still
+    # accessible after having been turned into bound methods
```
astafan8: <3 :)

qcodes/tests/drivers/test_rigol_DG4000.py
```diff
@@ -0,0 +1,89 @@
...
+                          visalib=visalib)
+    yield kw_sim
+
+    kw_sim.close()
+
+
+def test_init(driver):
+    assert driver.IDN()['vendor'] == 'QCoDeS'
+
+    # test that what used to be implemented via add_function is still
```
astafan8: is this going to remain, the only one notion of `add_function` in qcodes? :)

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -382,9 +383,14 @@ class ZNB(VisaInstrument):
...
 
 
-    def __init__(self, name: str, address: str, init_s_params: bool=True, **kwargs) -> None:
+    def __init__(self, name: str, address: str,
+                 init_s_params: bool=True, **kwargs) -> None:
 
-        super().__init__(name=name, address=address, **kwargs)
+        # NOTE(WilliamHPNielsen): there is an implicit assumption in the
+        # val_mappings of this driver that the terminator is '', i.e. that
+        # a '\n' is also returned
```
jenshnielsen: This sounds like the original author did something wrong? :)

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA.py
WilliamHPNielsen: Nice catch!

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
WilliamHPNielsen: I think I'd rather say that both strings and integers work with python's string formatter.

It goes like this: The user inputs `"ON"`. The `val_mapping` converts that into `1`. That value is passed to the `set_cmd` that calls `.format(1)`. The result, `'SYSTem:ERRor:DISPlay 1'` is sent via VISA to the instrument. 

Now, upon `get`, the string `"1\n"` is obtained. It goes through the `get_parser` and becomes `1`. The `val_mapping` then turns that into `"ON"` and the cycle is complete.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
WilliamHPNielsen: I totes forgot about it until I was halfway done... Do you think it matters now?

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
WilliamHPNielsen: No, it all depends on what you set your termination character to. If it's set to `""`, then you manually deal with the newlines. You can also set it to `"\n"` and never see a newline in your life. I just followed the directions already given in the driver. (but I might change those directions)

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
WilliamHPNielsen: I would be tempted to say yes, but perhaps he had some binary data transfer in mind when he did it? I am at least a little worried about breaking the trace acquisition if we change the termination.

astafan8: let's leave it like this)

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
astafan8: ok, i leave it up to you. this PR is about removing the functions, not about improving this driver.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
astafan8: it almost doesn't matter but there's this little thing that `@deprecate` uses `DeprecationWarning`: `warnings.warn(msg, DeprecationWarning, ...)`. I think it's better to use `DeprecationWarning`. And for that matter, changing to using `@deprecate` is probably the same amount of effort.

qcodes/instrument_drivers/QuTech/IVVI.py
astafan8: in our context, indeed, that doesn't add much value. With my suggestion, `trigger` method would've been visible on the class level, as in without the need for instantiating an instance to see that such a method exists. I guess, it's just probably a "better" programming practice.

WilliamHPNielsen:
@astafan8, I think both remaining suggestions are a "wontfix" from me. If you feel strongly about it, I am not going to object to any changes you push. Else, I'd say let's not spend more time on this dull PR.

jenshnielsen:
Just did a close/reopen to ensure the tests actually run on Azure too

PR #1203, opened 2018-07-24
Driver/infiniium additional parameters

WilliamHPNielsen:
- Add all the measurement parameters in a submodule
- Improve driver safety slightly by switching off headers in `__init__`. If headers are on, instrument responses ping back the query as well as the value, and our `get_parser`s can't handle that

Question: should the measurement submodule be snapshotted?

@QCoDeS/core


WilliamHPNielsen:
> I do not see a big point since it is a *measurement* submodule, and it does not seem to have "settings"

That's also what I am thinking. They describe the state of the signal, not the state of the instrument.

qcodes/instrument_drivers/Keysight/Infiniium.py
```diff
@@ -152,6 +150,122 @@ def get(self):
...
+                           get_parser=float,
+                           unit='V')
+
+        self.add_parameter(name='rms_no_DC',
+                           label='Voltage RMS',
+                           get_cmd=self._meas('VRMS') + ' DISPlay, AC',
+                           get_parser=float,
+                           unit='V')
+
+    def _meas(self, cmd: str) -> str:
```
jenshnielsen: Do you want this to be a static method. Also imho the name is slightly misleading. How about `_generate_meas_str` or some such thing?

qcodes/instrument_drivers/Keysight/Infiniium.py
```diff
@@ -152,6 +150,133 @@ def get(self):
         return channel_data
 
 
+class MeasurementSubsystem(InstrumentChannel):
+    """
+    Submodule containing the measurement subsystem commands and associated
+    parameters
+    """
+    # note: this is not really a channel, but InstrumentChannel does everything
+    # a 'Submodule' class should do
```
spauka: We had a discussion in a bug report about renaming `InstrumentChannel` for this reason (can't find the discussion right now though). Perhaps we should revisit?

astafan8: this can be very easily solved by renaming `InstrumentChannel` class to SubModule` (or `SubInstrument`, whatever) and adding `InstrumentChannel = SubModule` after `SubModule` class definition.

`SubModule` seems like a good name. But what "should this class do"? Contain a piece of functionality with tight reference/coupling to the instrument that it is a part of?

WilliamHPNielsen: I like the name `SubModule` too. I think `InstrumentChannel` already **is** that class. If you guys would like to discuss this more, then let us perhaps open a separate issue.

WilliamHPNielsen:
I'd say this is ready to land.

PR #1206, opened 2018-07-25
Set minimum version of firmware for Rigol DS4000

YakBizzarro:
With firmware version 00.02.02, data transfer is not working correctly when accessing the raw buffer (we receive corrupted data). This commit check that the scope is running at least version 00.02.03 of the firmware.

@astafan8


qcodes/instrument_drivers/rigol/DS4000.py
```diff
@@ -250,3 +252,11 @@ def __init__(self, name, address, timeout=20, **kwargs):
...
         channels.lock()
         self.add_submodule('channels', channels)
+
+    def _check_firmware_version(self):
+        #Require version 00.02.03
+
+        idn = self.get_idn()
+        ver = [int(x) for x in idn['firmware'].split('.')]
+        if ver < [0,2,3]:
+            raise RuntimeError('Firmware version must be at least 00.02.03')
```
astafan8: Please add a clear explanation to the exception message about why the firmware needs to be at least that.

astafan8: also, do you receive corrupted data all the time? as in, is the scope still usable to some extent with `00.02.02` and lower or not? if it is, then please change `raise RuntimeError` to a waning `warnings.warn`, because the users should not get suddenly blocked from using the driver/instrument.

qcodes/instrument_drivers/rigol/DS4000.py
```diff
@@ -250,3 +252,11 @@ def __init__(self, name, address, timeout=20, **kwargs):
 
         channels.lock()
         self.add_submodule('channels', channels)
+
+    def _check_firmware_version(self):
+        #Require version 00.02.03
+
+        idn = self.get_idn()
+        ver = [int(x) for x in idn['firmware'].split('.')]
```
astafan8: is firmware version always a three-number string `#.#.#`?

you do not have to reinvent the wheel here, just use `from distutils.version import LooseVersion` ([example here](https://stackoverflow.com/a/11887885) or [here](https://stackoverflow.com/a/6972866) or [even in QCoDeS](https://github.com/QCoDeS/Qcodes/blob/3ff371052bb90ffdcfef192f9f1a505b6ba50c58/qcodes/instrument_drivers/rohde_schwarz/RTO1000.py#L315)) .

Another thing you could use is `from pkg_resources import parse_version` function ([examples here](https://setuptools.readthedocs.io/en/latest/setuptools.html#specifying-your-project-s-version) and [here](https://stackoverflow.com/a/21065570)) but it is not included in the standard library of python.

qcodes/instrument_drivers/rigol/DS4000.py
astafan8: Also, I just trust you that it is indeed the firmware problem, and not some minor error in the way you are using the instrument. Nevertheless, could you please double check that you are doing things in the right way?

qcodes/instrument_drivers/rigol/DS4000.py
YakBizzarro: Ah good, I was looking for something similar, but I didn't found it, thanks!

qcodes/instrument_drivers/rigol/DS4000.py
YakBizzarro: Sincerely I don't know how extended is the breakage. I just noticed that the function to get the raw buffer was broken and got working after an upgrade. Also now I don't have a way anymore to check with the old version. Since it's easily upgredable, I would not waste time to try to use an old version

astafan8: ok. but still let's keep it safe. So:
- make a warning instead of error
- add more info to the text of the warning explaining what is wrong with an older firmware

YakBizzarro:
@astafan8 updated as requested

qcodes/instrument_drivers/rigol/DS4000.py
```diff
@@ -250,3 +254,11 @@ def __init__(self, name, address, timeout=20, **kwargs):
...
         channels.lock()
         self.add_submodule('channels', channels)
+
+    def _check_firmware_version(self):
+        #Require version 00.02.03
+
+        idn = self.get_idn()
+        ver = LooseVersion(idn['firmware'])
+        if ver < LooseVersion('00.02.03'):
+            warnings.warn('Firmware version should be at least 00.02.03, data transfert may not work correctly')
```
astafan8: typo: `transfer` (without `t`) :) fix it and then I'll merge

astafan8: @YakBizzarro

WilliamHPNielsen: The line is also slightly too long. If we really want to get nitty-gritty (and we do!)

YakBizzarro: Ok, done

WilliamHPNielsen:
@YakBizzarro thank you!

PR #1207, opened 2018-07-25
Dataset Context manager. Allow insertion of Data direcly as arrays

### LINKED ISSUE #1108: register_parameter always assumes 'numeric' type
Register parameter always uses paramtype='numeric' when creating the ParamSpec, regardless of the type of parameter to be registered. This is problematic when we try to register array-valued parameters.

A proposed solution is to check the parameter type. As a side note, using the ArrayParameter is undesirable as the shape of the array needs to be specified beforehand; this is a different, albeit related issue. 

A work around is to use ParamSpec directly and not use 'register_parameter'. We can add the ParamSpec manually to the data set.

jenshnielsen:
This is significantly faster than unrolling and inserting point by point. This in turn uncovered some other issues.

* Correct insertion of multidimensional Arrays in the context manager. Previously it would only 
unroll the first axis resulting in inconsistent results. 
* Adapt the data exporter `get_data_by_id` to return full arrays of setpoints regardless of the way they are inserted.

TODO:

- [x] Add Benchmarks of the difference vs insertion as single points
- [ ] The tests have a fair bit of unneeded code duplication
- [x] Examples of Array Insertion vs single point insertion
- [x] Naively extracting data the way that `DataSet.get_data` does will result in a different result from extracting the data inserted row by row but the `get_data_by_id` exporter works as expected

Edit by William: this PR fixes #1108


astafan8:
Could you expand a bit more in the description what you changed exactly? because I see quite some stuff related to ArrayParameter, and the stuff that is related to tuples `(qcodes_param, numpy_array_of_data)` is lost there.

astafan8:
Also, 👍 (+1) for new `get_data_as_*` functions! I hope there soon will be a PR that makes these also methods of some object that `DataSaver` can return (similar to [Sohail's and Wolfgang's `_DataExtractor`](https://github.com/kouwenhovenlab/pytopo/blob/c0b6676ebe2c4f8defdd4d9ffb08825bfe811ff7/pytopo/qctools/dataset2.py#L57))

qcodes/dataset/measurements.py
```diff
@@ -159,23 +181,34 @@ def add_result(self,
...
-                    if isinstance(value, np.ndarray):
-                        value = np.atleast_1d(value)
-                    res_dict.update({param: value[index]})
-                else:
-                    res_dict.update({param: value})
-
-            self._results.append(res_dict)
+                param_spec = self.parameters[str(partial_result[0])]
+                if param_spec.type == 'array' and index == 0:
+                    res_dict[str(partial_result[0])] = partial_result[1]
```
astafan8: is `res_dict.update(..)` better? (see below)

qcodes/dataset/measurements.py
```diff
@@ -159,23 +181,34 @@ def add_result(self,
...
+                elif param_spec.type != 'array':
+                    param = str(partial_result[0])
+                    value = partial_result[1]
+                    # For compatibility with the old Loop, setpoints are
+                    # tuples of numbers (usually tuple(np.linspace(...))
+                    if hasattr(value, '__len__') and not(isinstance(value,
+                                                                    str)):
+                        value = cast(Union[Sequence, np.ndarray], value)
+                        if isinstance(value, np.ndarray):
+                            value = np.atleast_1d(value).ravel()
```
astafan8: when I was profiling this, I saw indeed that after the thing that I've fixed already, `atleast_1d` takes quite some time together with the `as_float` data adapter (if i call it correctly).

qcodes/dataset/measurements.py
```diff
@@ -159,23 +181,34 @@ def add_result(self,
...
-                        value = np.atleast_1d(value)
-                    res_dict.update({param: value[index]})
-                else:
-                    res_dict.update({param: value})
-
-            self._results.append(res_dict)
+                param_spec = self.parameters[str(partial_result[0])]
+                if param_spec.type == 'array' and index == 0:
+                    res_dict[str(partial_result[0])] = partial_result[1]
+                elif param_spec.type != 'array':
```
astafan8: btw, shall we finally make an enum with these SQLite data types? Isn't there already one in `sqlite` module?

qcodes/dataset/measurements.py
jenshnielsen: Afaik this is only there to handle the case of 0D numpy arrays (i.e. np.array(1)) which is not the same as a numpy scalar but will pass is instance check for array and take this code path. We could rewrite it to make sure that it doesn't

qcodes/dataset/measurements.py
jenshnielsen: It's significantly slower by my benchmark
```
mydict = {'A': 'Foo', 'B': 'Bar'}
%timeit mydict['C'] = 'FooBar'
35.4 ns ± 1.78 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
%timeit dict['C'] = 'FooBar'
36 ns ± 0.393 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
%timeit mydict.update({'C': 'FooBar'})
159 ns ± 3.36 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

astafan8: mmmmmm..... i might not be right, but for the `.update` case, could it be that creation of a new dictionary (`{'C': 'FooBar'}`) takes some time (while in the first two cases, just two strings are created)?

qcodes/dataset/measurements.py
astafan8: omg.... and i guess writing an explicit `if` statement (as opposed to just `isinstance` check) to cover the case is less efficient, hence the `atleast_1d` without any comments around for the rationale.

jenshnielsen: Yes atleast_1d is surprisingly slow especially already 1d and higher arrays. I will replace it with a single reshape if needed

```
a = np.array((0))
a
Out[59]: array(0)
a.shape
Out[60]: ()
%%timeit
if a.ndim == 0:
    a.reshape(1)
565 ns ± 5.2 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
a = np.array((0,1,3))
%%timeit
if a.ndim == 0:
    a.reshape(1)
41 ns ± 1.59 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
a = np.array((0))
%%timeit
np.atleast_1d(a)
1.15 µs ± 10.7 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
a = np.array((0,1,3))
%%timeit
np.atleast_1d(a)
628 ns ± 11.7 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

qcodes/dataset/measurements.py
jenshnielsen: It's not all of it 
```
mydict = {'A': 'Foo', 'B': 'Bar'}
myotherdict = {'C': 'FooBar'}
%timeit mydict['C'] = 'FooBar'
%timeit mydict.update({'C': 'FooBar'})
%timeit mydict.update(myotherdict)
34.9 ns ± 0.977 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
154 ns ± 1.9 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
139 ns ± 4.94 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

I don't think update is the right method here. It's really for merging dicts. We only have one element to insert

qcodes/dataset/measurements.py
jenshnielsen: Yes I would have assumed that atleast_1d was very cheap if dealing with 1d or higher arrays but it looks like it does a lot of other checks too

qcodes/dataset/measurements.py
jenshnielsen: Done by exposing the list in ParamSpec

jenshnielsen:
@astafan8 There is basically no difference between the benchmarks before and after the changes. In both cases I am seeing runtimes between 400 and 500 ms. With the changes to avoid atleast_1d I see a consistent speedup. 

I added more benchmarks using the store as array method demonstrating that its faster along with a notebook showing the speedup.

benchmarking/benchmarks/dataset.py
```diff
@@ -36,9 +37,15 @@ class Adding5Params:
...
-        {'n_values': 10000, 'n_times': 2},
-        {'n_values': 100, 'n_times': 200}
+        {'n_values': 10000, 'n_times': 2, 'paramtype': 'array'},
+        {'n_values': 100, 'n_times': 200, 'paramtype': 'array'},
+        {'n_values': 10000, 'n_times': 2, 'paramtype': 'numeric'},
+        {'n_values': 100, 'n_times': 200, 'paramtype': 'numeric'},
     ]
+    # we are less interested in the cpu time used and more interested in
+    # the wall clock time used to insert the data so use a timer that measures
+    # wallclock time
```
astafan8: wallclock time of _this process_, you mean, right?

qcodes/dataset/data_set.py
```diff
@@ -567,26 +567,32 @@ def get_data(self,
...
         Args:
             - *params: string parameter names, QCoDeS Parameter objects, and
                ParamSpec objects
             - start:
             - end:
 
         Returns:
-            - list of parallel NumPy arrays, one array per parameter
-        per parameter
+            - list of rows of data. Each row will contain a list of columns
```
astafan8: "each row will contain a list of columns"? what do you mean exactly here?

qcodes/dataset/data_set.py
```diff
@@ -567,26 +567,32 @@ def get_data(self,
                  start: Optional[int] = None,
                  end: Optional[int] = None) -> List[List[Any]]:
         """ Returns the values stored in the DataSet for the specified parameters.
-        The values are returned as a list of parallel NumPy arrays, one array
-        per parameter. The data type of each array is based on the data type
-        provided when the DataSet was created. The parameter list may contain
-        a mix of string parameter names, QCoDeS Parameter objects, and
+        The values are returned as a list of lists, rows by columns, e.g.
+        datapoint by parameter. The data type of each element is based on the
```
astafan8: I still did not understand the format. for an sql table like this:

x | y | z
--|---|--
1 | 2 | 3
4 | 5 | 6

is it going to return:
```python
[ [1, 2, 3],
  [4, 5, 6] ]
```
?

qcodes/dataset/measurements.py
```diff
@@ -524,7 +582,7 @@ def register_custom_parameter(
         depends_on, inf_from = self._registration_validation(name, sp_strings,
                                                              bs_strings)
 
-        parspec = ParamSpec(name=name, paramtype='numeric',
+        parspec = ParamSpec(name=name, paramtype=paramtype,
```
astafan8: no checks for `paramtype` here?

astafan8: ah, found it in the `ParamSpec` constructor.

qcodes/dataset/measurements.py
```diff
@@ -407,8 +456,13 @@ def register_parameter(
             basis: The parameters that this parameter is inferred from. If
                 this parameter is not inferred from any other parameters,
                 this should be left blank.
+            paramtype: type of the parameter, i.e. the SQL storage class
         """
         # input validation
+        if paramtype not in ParamSpec.allowed_types:
```
astafan8: is this check necessary here, or, instead, can we wait until `ParamSpec` object is initiated? Seems like it's good that it's here because the error message refers to the fact that a qcpdes parameter is being passed in to the method...

docs/examples/DataSet/Dataset Performance.ipynb
```diff
@@ -0,0 +1,2541 @@
+{
+ "cells": [
```
astafan8: Very nice notebook! three comments:
- could you use log scale for amount of numbers to insert?
- could you add 1.000, 10.000, perhaps 100.000 as well to the benchmarking?
- could you set the database location to smth else so that people's default `.db` files don't get suddenly filled with this test data?

benchmarking/benchmarks/dataset.py
jenshnielsen: No Wallclock time of a process is IMHO a undefined.There is only one wallclock time

docs/examples/DataSet/Dataset Performance.ipynb
jenshnielsen: > - could you use log scale for amount of numbers to insert?
I chose a linear scale because the scaling is linear in the number of points
- could you add 1.000, 10.000, perhaps 100.000 as well to the benchmarking?
I added some more examples to the first example but not to the second one to keep the notebook runtime down
- could you set the database location to smth else so that people's default `.db` files don't get suddenly filled with this test data?
Done

docs/examples/DataSet/Dataset Performance.ipynb
```diff
@@ -0,0 +1,2541 @@
+{
```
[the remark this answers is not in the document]
jenshnielsen: Added example of that to the notebook

WilliamHPNielsen:
The docstring for `add_result` should be updated regarding how arrays are handled. 

Also, although it's all in the tests and performance notebook, I think it would be good with a dedicated tutorial notebook showing an example of storing and retrieving arrays. Perhaps we can extend existing notebooks? (I wouldn't mind doing that)

qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -794,10 +827,11 @@ def test_datasaver_arrayparams_tuples(experiment, SpectrumAnalyzer, DAC, N, M):
     assert datasaver.points_written == N*M
 
 
-@settings(max_examples=5, deadline=None)
+@settings(max_examples=5, deadline=None, use_coverage=False)
```
WilliamHPNielsen: I think `use_coverage` should not be `False`?

qcodes/dataset/data_set.py
```diff
@@ -567,26 +567,33 @@ def get_data(self,
...
         ParamSpec objects. As long as they have a `name` field. If provided,
         the start and end parameters select a range of results by result count
         (index).
         If the range is empty -- that is, if the end is less than or
         equal to the start, or if start is after the current end of the
         DataSet – then a list of empty arrays is returned.
 
+        For a more type independent and easier to work with view of the data
+        you may want to consider using
+        :py:meth:`qcodes.dataset.data_exporter.get_data_by_id`
```
WilliamHPNielsen: I believe there's a typo here, shouldn't it be `qcodes.dataset.data_export.get_data_by_id`? (no ER after export)

WilliamHPNielsen:
I've added a high-level tutorial notebook. I think it's good to have, and if nothing else, writing it convinced that this is a kick-ass PR! There is also a suggestion for an extension of the test in the notebook's version of the `MultiDimSpectrum`.

jenshnielsen:
Addressed @WilliamHPNielsen  comments. I think this is ready to land if the CI passes

PR #1209, opened 2018-07-26
Bug: snapshot fails when instantiating InstrumentBase

astafan8:
The problem occurs when one wants to subclass from `InstrumentBase` (or instantiate it). The reason to do so is to create a "virtual" instrument, and to avoid the overhead of `Instrument` class related to the register of instances.

The problem is that the snapshot functionality of `InstrumentBase` raises an exception saying "the object does not have attribute `_meta_attrs`". This is because `snapshot_base` uses it, but for some reason `_meta_attrs` is only defined in the `Instrument` class. This is not correct, hence this PR.

Changes:
- declare `_meta_attrs` in `InstrumentBase` instead of `Instrument`
- add tests for snapshotting `InstrumentBase`
- modify `Instrument` test to ensure snapshotting still works for it


WilliamHPNielsen:
It's not so clear to me what is gained by instantiating `InstrumentBase` directly, but irrespective of that, this refactor seems sensible to me.

astafan8:
yeah, it is more for subclassing `InstrumentBase`.

PR #1210, opened 2018-07-26
Find or create an instrument without worrying about existing instances

astafan8:
This new function `find_or_create_instrument` avoids the hassle of usage of `Instrument.find_instrument`. In case an instrument does not already exist, the latter will throw an exception, while the new function will just return the reference to that instrument.

If this function is used to create instruments, then it can be reexecuted, because unlike `find_instrument` or just reinstantiation with a constructor, this function does not throw. Moreover, thanks to the `recreate` kwarg, it allows to recreate an instance of the instrument in case one already exists.

Researchers find this convenience function very handy. `pytopo` repository (Delft lab) also has [a variation of this convenience function](https://github.com/kouwenhovenlab/pytopo/blob/11427031eef8be6ce4c56246a92ca90f3833be95/pytopo/qctools/instruments.py#L8). Even I ended up implementing this function for `v0_characterization` [experiment and repository](https://github.com/kouwenhovenlab/v0_characterization/blob/461a622dd0140c4a00eb7ccfb74584086839bedb/v0_utils/qcodes_tools.py#L52). 

@wpfff Let me know if this is good enough (I've decided to avoid the `force_new_instance` feature). Update: aligned on the solution.


astafan8:
just added a test and a note about "same name, different class" case. If necessary, the current behavior can be revisited later.

PR #1211, opened 2018-07-26
Add some minor DataSet conveniences

astafan8:
The features proposed in this PR are coming from Delft experimentalists. Some of these were implemented in one way or another in [`pytopo` repository](https://github.com/kouwenhovenlab/pytopo).

List of changes:
- make `initialise_database` available from `import qcodes`
- add `initialise_or_create_database_at`
- add `load_or_create_experiment`
- add `DataSet.exp_name` property that returns the experiment name
- add `DataSet.sample_name` property that returns the sample name
- add `DataSet.run_timestamp()` method and `DataSet.run_timestamp_raw` property for obtaining the run timestamp in a formatted string and in seconds, respectively
- some very minor refactoring on the way

@wpfff your comments are also welcome


gijs:
Probably another gijs? ;)

qcodes/dataset/data_set.py
```diff
@@ -251,10 +254,30 @@ def paramspecs(self) -> Dict[str, ParamSpec]:
...
+    def exp_name(self) -> str:
+        return get_experiment_name_from_experiment_id(self.conn, self.exp_id)
+
+    @property
+    def sample_name(self) -> str:
+        return get_sample_name_from_experiment_id(self.conn, self.exp_id)
+
+    @property
+    def run_timestamp_raw(self) -> float:
+        """Returns run timestamp as number of seconds since the Epoch"""
```
WilliamHPNielsen: It's probably good to mention (here or elsewhere) when the timestamp is actually made. I guess this one is at the start of the measurement?

astafan8: yes. when the new `run` (which means "a new dataset") is created in the database, a timestamp of that moment is added to the `run` (see the "results" table in the database). That's why the docstring says `run timestamp`. I will methion that it is the "moment when the measurement run started".

WilliamHPNielsen: Great. How about a `completed_timestamp` too?

astafan8: ok, i can add it. anything else? :)

astafan8: added :)

PR #1213, opened 2018-07-30
Hotfix/awg70000 makeseqxfile

WilliamHPNielsen:
Fixes a bug discovered in the lab just now.

- Add a test reproducing the bug
- Fix the failing line

@astafan8


WilliamHPNielsen:
@astafan8 for your curiosity: `wfms` is a list of lists of multidimensional np.arrays (waveform, marker, marker). If the arrays have different lengths, `np.shape` of the lists is undefined.

Something like
```python
np.shape([[np.zeros((3, 4)), np.zeros((3, 3))]])
```
does not work, although it should; it is a single channel with two elements having respectively 4 and 3 points.

astafan8:
aaa., ok, thanks for the explanation. then perhaps numpy expects an array to be like a matrix, while in your case it is not. you basically have to manually account for the fact that your arrays in the lists may have linear-algebra-incompatible shapes. but i'm not sure, whatever.

PR #1214, opened 2018-07-31
[WIP] Feature / dataset merger

WilliamHPNielsen:
- Add a function to merge (very similar) datasets. The idea is that a measurement may have been broken into smaller pieces by instrumental necessity and there should be a way to glue the different pieces together.

Pending:
- [ ] Tests, tests, tests
- [ ] Metadata? How should we capture the fact that this is somehow non-measured experimental data?

@QCoDeS/core


astafan8:
This looks very nice! questions:
* Is `merge` really a good name? Shall we use something more mathematical like `union`?
* What happens if the two datasets that are to be merged, have measured values at the same values of setpoints? For example, if `dataset1` has `{x: 1, y: 2, z: 10}` while `dataset2` has `{x: 1, y: 2, z: 65}`? I guess, this is not a big deal for the dataset itself, right? It's just the worry of the tools which work with data like `plot_by_id`, right?

WilliamHPNielsen:
- The name:  You are right that set theoretical names are more concise in terms of *how* they combine/unite the datasets. But currently `merge` will only work if the datasets are very similar, so whereas you might expect `union` to really just combine everything (and thus works for any two datasets), `merge` (potentially, if you git'ed a lot) offers the user the intuition that not everything can be merged. And the `merge` is a verb, which I like. I suppose `unite` would correspond to `union`? But perhaps it'd be good to already now think about all the combining functions we'd like to have and then name appropriately.

- Setpoints: exactly. If you play around with the notebook, you can see `plot_by_id` going mental if there are overlapping regions and a grid is detected. We should look into that.

astafan8:
Concerning metadata. I think that for now you can combine the metadata into a dictionary by `run_id` and/or `run_name` within a dedicated field, smth like:
```
{
    merged_runs: {
        25: {'snapshot': {...}},
        37: {'snapshot': {...}},
    },
    < other meta data of THIS dataset >
}
```
I think the metadata is going to be a "place for dumping stuff" anyway, hence it is just not easy to say how to perform "operations" on it. So, a simple solution like this is probably fine as long as minimal "separation"/"organization" measures are taken.

WilliamHPNielsen:
I like that metadata solution. Let me see if I can fix the three items today.

qcodes/dataset/data_set_tools.py
```diff
@@ -28,25 +29,32 @@ def merge(datasets: Sequence[Union[int, DataSet]]) -> DataSet:
...
+    for ds in dsets:
         params = ds.get_parameters()
         data = ds.get_data(*params)
         for data_point in data:
             results.append({p.name: val for p, val in zip(params, data_point)})
+        snapshot = ds.get_metadata('snapshot')
+        snapshot = "{}" if snapshot is None else snapshot
+        snapshots[ds.run_id] = json.loads(snapshot)
     new_ds.add_results(results)
+    new_ds.add_metadata('parent_snapshots', json.dumps(snapshots))
```
astafan8: Beware that this way you are adding a new column to the database. `add_metadata` is implemented in such a way that a `tag` is actually a completely new column in the `runs` table. I am a bit concerned about this fact for the reason that the structure of columns should probably be restricted to 'smth' (because it is safer, isn't it; but yes, less flexible; but who needs THIS flexibility...). This is why my proposal was to have what you called `parent_snapshots` inside the `snapshot` part of metadata. What do you think?

WilliamHPNielsen: Yes, I am aware of that, and it has also been bugging me, but at the same time we have no idea what metadata we are eventually going to need, so for now I think dynamically adding columns is acceptable. 

I considering putting this in as a `snapshot`, but I am worried that snapshot-reading tools will break unless the snapshot has a particular format.

WilliamHPNielsen:
This PR should wait for #965 to be merged. That way it will be captured where the data came from.

PR #1215, opened 2018-08-02
[WIP] Feature: convenient snapshot exploration

astafan8:
Users mentioned many times that it is not obvious and easy to work with the snapshot.

This PR tries to solve it in the following way:
- [ ] implement a convenience function that allows to conveniently explore the snapshot dict (similar to [this](https://github.com/qdev-dk/qdev-wrappers/blob/3bf23dcf5b5fd184e207ff80c1e1050c5c49ebc8/qdev_wrappers/show_num.py#L175))

Original list of changes of this PR included these items, but they are implemented in #1322:
- add notebook that explains snapshots
- update `docs/examples/The%20Snapshot.ipynb` to link to the new notebook
- add tests for the snapshot feature of `Measurement`, `Runner`, `DataSaver`, and `DataSet`


qcodes/utils/snapshot_tools.py
```diff
@@ -0,0 +1,133 @@
...
+        'unit',
+        'label',
+        #     'full_name',  # do we care?
+        #     'instrument',  # we should not be strict about this, but rather highlight it when we see it
+        #     'instrument_name',  # do we care?
+    ]
+
+
+def _is_parameter_like(snapshot_part):
+    return True if \
```
Dominik-Vogel: You could make this a simple
```
return all(key in snapshot_part.keys() for key in _parameter_keys)
```
i.e. without the if

astafan8:
This PR was redefined, because its part have moved to #1322 PR.

astafan8:
closing as this has been solved by the experiment_widget and its subwidget for displaying dicts.

PR #1219, opened 2018-08-08
Remove unused function in N52xx driver

spauka:
Remvoes an unused function `_set_auto_sweep` in N52xx driver that slipped through.

@jenshnielsen, @astafan8


spauka:
I couldn't see any :P

PR #1225, opened 2018-08-09
Make data saver accept numpy type floats/ints 

### LINKED ISSUE #1224: numpy types (e.g. int, float) with bit size 8, 16, 32 and 64, etc. not recognized as as non_array_type by data saver
We cannot insert numpy types via the data saver. To reproduce: 

```python
from qcodes import Parameter
from qcodes.dataset.experiment_container import new_experiment

exp = new_experiment("name", "sample")

x = Parameter("x", set_cmd=None, get_cmd=None)

meas = Measurement()
meas.register_parameter(x)

with meas.run() as datasaver: 
    datasaver.add_result((x, np.float16(1.0)))  # Fails with np.int8, np.int16, etc as well. 
```

sohailc:
Fixes #1224

We test if a value is a "non-array-type" by our ability to recast it to a float. This is more robust than having a list of acceptable types as we can always miss types.

### DIFF
--- a/qcodes/tests/dataset/test_datasaver.py
+++ b/qcodes/tests/dataset/test_datasaver.py
@@ -78,7 +78,7 @@ def test_default_callback(experiment):
 
 def test_numpy_types(experiment):
     """
-    Test that we can save numpy types in the dataset
+    Test that we can save numpy types in the data set
     """
 
     p = ParamSpec("p", "numeric")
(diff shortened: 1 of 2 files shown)


sohailc:
@WilliamHPNielsen @astafan8 I have added test. Can we have another look

sohailc:
@WilliamHPNielsen If you approve this, I will merge right away

qcodes/tests/dataset/test_datasaver.py
```diff
@@ -91,3 +91,16 @@ def test_numpy_types(experiment):
...
+    """
+    Test that we can save text in the data set
+    """
+    p = ParamSpec("p", "text")
+
+    test_set = qc.new_data_set("test-dataset")
+    data_saver = DataSaver(
+        dataset=test_set, write_period=0, parameters={"p": p})
+
+    data_saver.add_result(("p", "some text"))
```
astafan8: don't want to be annoying but:
* pep8 required newline at the end of the file
* should the test also assert that `some text` is indeed in the database?

sohailc: I am quite surprised this commit even passed the quality inspection: Doesn't CI check for PEP8? 

Anyway, I will correct the above.

PR #1226, opened 2018-08-09
allow config to update based on specified path

### LINKED ISSUE #1223: Config can only use certain files
It is possible to load a config from a specified file (returning the config in dictionary form) 

`cfg = Config()`
`cfg.load_config('some_path_here')`

and it is possible to update the config (which looks in the cwd, the env and the home directory)

`cfg.update()` 

but the combined version of these where you specify the location of the config file you want to use as current_config doesn't exist. Is there a strong reason for this?

nataliejpg:
Fixes #1223 .

allows a config file path to be specified either in initialisation of the Config object or when running update. This path is then used for future updates.

How's this @jenshnielsen? I haven't tested it yet.

### DIFF
--- a/qcodes/config/config.py
+++ b/qcodes/config/config.py
@@ -125,10 +125,10 @@ class Config():
         self.current_schema = copy.deepcopy(self.defaults_schema)
         self.config_file_path = path or self.config_file_path
         if self.config_file_path is not None:
-            config_file = "{}/{}".format(path, config_file_name)
-            schema_file = config_file.replace(config_file_name,
-                                              schema_file_name)
-            path_config = self.load_config(path)
+            config_file = "{}/{}".format(self.config_file_path, self.config_file_name)
+            schema_file = config_file.replace(self.config_file_name,
+                                              self.schema_file_name)
+            path_config = self.load_config(config_file)
             config = update(config, path_config)
             self.validate(config, self.current_schema,
                           schema_file)
@@ -151,7 +151,7 @@ class Config():
                 self.validate(config, self.current_schema,
                               self.schema_cwd_file_name)
 
-        self._current_config = config
+        self.current_config = config
 
     def validate(self, json_config=None, schema=None, extra_schema_path=None):
         """
--- a/qcodes/config/config.py
+++ b/qcodes/config/config.py
@@ -69,7 +69,7 @@ class Config():
                                                       schema_file_name)
 
     # home dir, os independent
-    home_file_name = expanduser("~/{}".format(config_file_name))
+    home_file_name = expanduser(os.path.join("~", config_file_name))
     schema_home_file_name = home_file_name.replace(config_file_name,
                                                    schema_file_name)
 
@@ -78,7 +78,7 @@ class Config():
     schema_env_file_name = env_file_name.replace(config_file_name,
                                                  schema_file_name)
     # current working dir
-    cwd_file_name = "{}/{}".format(Path.cwd(), config_file_name)
+    cwd_file_name = os.path.join(Path.cwd(), config_file_name)
     schema_cwd_file_name = cwd_file_name.replace(config_file_name,
                                                  schema_file_name)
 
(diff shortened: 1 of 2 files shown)


WilliamHPNielsen:
Judging from the tests, this does not quite work yet.

nataliejpg:
@WilliamHPNielsen the bug is a typo, I think you can get the main points from my comment or looking at the code and I was hoping for some feedback on the concept before I put time into testing

WilliamHPNielsen:
@nataliejpg and I was hoping for actually working code before I put time into reviewing... oh, no, now we're stuck in a deadlock 😄 
I think the idea of passing a kwarg path to force load from that path is A-OK.

nataliejpg:
@WilliamHPNielsen should work now

qcodes/config/config.py
```diff
@@ -121,26 +123,35 @@ def update_config(self):
...
-            config = update(config, env_config)
-            self.validate(config, self.current_schema,
-                          self.schema_env_file_name)
-
-        if os.path.isfile(self.cwd_file_name):
-            cwd_config = self.load_config(self.cwd_file_name)
-            config = update(config, cwd_config)
+        self.config_file_path = path or self.config_file_path
+        if self.config_file_path is not None:
+            config_file = "{}/{}".format(self.config_file_path, self.config_file_name)
```
astafan8: please use `os.path.join` instead of `/` and `format`. let the OS define the file separator.

nataliejpg: yup, that would be better, I just copied the original logic from likes 72 and 81 but will update them all now.

qcodes/config/config.py
```diff
@@ -121,26 +123,35 @@ def update_config(self):
...
-            env_config = self.load_config(self.env_file_name)
-            config = update(config, env_config)
-            self.validate(config, self.current_schema,
-                          self.schema_env_file_name)
-
-        if os.path.isfile(self.cwd_file_name):
-            cwd_config = self.load_config(self.cwd_file_name)
-            config = update(config, cwd_config)
+        self.config_file_path = path or self.config_file_path
+        if self.config_file_path is not None:
```
astafan8: also, what about test? (see `qcodes/tests/test_config.py`)

nataliejpg: Unfortunately beyond my scope. perhaps someone (the author of the existing test?) could advise me what I need to add and I can put it in this pr? Otherwise realistically I won't be able to do this in the near future.

qcodes/config/config.py
```diff
@@ -121,26 +123,35 @@ def update_config(self):
...
+                self.validate(config, self.current_schema,
+                              self.schema_env_file_name)
+
+            if os.path.isfile(self.cwd_file_name):
+                cwd_config = self.load_config(self.cwd_file_name)
+                config = update(config, cwd_config)
+                self.validate(config, self.current_schema,
+                              self.schema_cwd_file_name)
+
+        self.current_config = config
```
WilliamHPNielsen: I suppose we still want this method to return the config?

qcodes/config/config.py
WilliamHPNielsen: The author is long gone, unfortunately. I agree that `test_config.py` is not the most immediately readable piece of code. I can help you extend the test to cover the new functionality introduced in this PR.

qcodes/config/config.py
nataliejpg: why?

astafan8: this depends on how people use it. Unfortunately, `docs/examples/Configuring_QCoDeS.ipynb` does not shed light on it... I leave this decision to you.

qcodes/config/config.py
nataliejpg: thanks @WilliamHPNielsen. How about I test this out when I'm next measuring and then we merge it and @WilliamHPNielsen writes the test in another pr for the actual test?

qcodes/config/config.py
WilliamHPNielsen: The missing return is the reason that the tests fail. I think we need a pretty strong argument for breaking/changing behaviour that we explicitly test for. We don't have visibility on who uses the software and how, so the golden rule is that we don't suddenly change the API. 

But I agree that the old/current API is not very intuitive, so we might deprecate some functions in favour of new ones that act better.

WilliamHPNielsen:
I guess I volunteered to clean this up. I am going to push one or two commits now, fixing the return and extending the tests.

Another issue is improving the documentation...

nataliejpg:
thanks @WilliamHPNielsen

WilliamHPNielsen:
Note to @QCoDeS/core: `Configuring_QCoDeS.ipynb` was last updated two years ago. We might want to have another look at that one.

PR #1227, opened 2018-08-10
[core only] Development/dependencies dsl

WilliamHPNielsen:
This is some internal core dev thing.

### Introduction

Here is my proposal for a first iteration DSL to describe dependencies. 

The idea is simple: let's use what we already have (and then build on that later). What is it that we have? The `ParamSpecs`. The proposal is therefore to simply write down the `ParamSpec` in a yaml file. That is, in all its glory, the DSL.

This PR introduces one object and two functions, an `InterDependencies` object consuming `ParamSpecs` and the functions `interdeps_to_yaml` and `yaml_to_interdeps`.

### Regarding implementation:
If we want to pursue this, the proposal then includes moving all information from the `dependencies` and `layouts` table into a `yaml` file stored in a column in the runs table (just like `snapshot`). I'd further propose that we
- Update the table schema
- Move all validation logic of a set of ParamSpec (are all setpoints present, are there cyclic dependencies, etc.) into the `InterDependencies` object. The `Measurement` class can then use that.
- Refactor the following methods to no longer query `dependencies` and `layout` but read and parse the `yaml` file instead:
  - `get_setpoints`
  - `get_layout`
  - `get_dependents`
  - `get_dependencies`
  - `get_paramspec`
- Extend the `ParamSpec` as extensions are needed. One extension comes to mind:
  - Give the `ParamSpec` a `run_id` (preferably the global one) 

@QCoDeS/core, let us see if this can work for us.


WilliamHPNielsen:
There is now a little notebook showing the basic idea of object -> yaml and yaml -> object.

qcodes/dataset/dependencies.py
```diff
@@ -0,0 +1,61 @@
...
+    # we use a heading, Parameters, since the yaml file might be extended
+    # with more info in the future
+    yaml.dump({'Parameters': idp.paramspecs}, stream=stream)
+    output = stream.getvalue()
+    stream.close()
+    return output
+
+
+def yaml_to_interdeps(yaml_str: str) -> InterDependencies:
+    yaml = YAML()
```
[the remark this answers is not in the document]
WilliamHPNielsen: Good point.

WilliamHPNielsen:
@astafan8, yes, the idea is to make this PR give us an MVP. Sorry, I think I forgot write that. Let me do it now: we have decided to implement the DSL into the database, meaning that the four bullet points will be covered.

qcodes/dataset/dependencies.py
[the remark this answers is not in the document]
WilliamHPNielsen: It does now, via versioning. The code contains an example of a somewhat contrived upgrade, but it establishes the principle.

astafan8:
perhaps a useful note: I was looking at ruamel, and on [their pypi page they state](https://pypi.org/project/ruamel.yaml/) that
> For production systems you should pin the version being used with `ruamel.yaml<=0.15`
Does it impact us in any way?

qcodes/dataset/param_spec.py
```diff
@@ -61,45 +69,82 @@ def depends_on(self):
...
+        if cls.version > ser['version']:
+            for upgrade in upgrades[:cls.version]:
+                upgrade(ser)
+
+        return ParamSpec(name=ser['name'],
+                         paramtype=ser['paramtype'],
+                         label=ser['label'],
+                         unit=ser['unit'],
+                         inferred_from=ser['inferred_from'],
+                         depends_on=ser['depends_on'])
```
astafan8: should here also be `guids=ser['guids']` kwarg?

WilliamHPNielsen: Miss me with those guids in the paramspec

WilliamHPNielsen:
@QCoDeS/core 

Since this PR is becoming rather larger and potentially stuff-breaking, I propose that we for now only introduce the new column with the yaml string (and the objects to parse that). In another PR we can then refactor all the functions that read from `dependencies` and `layouts` to use the `run_description` column instead.

I don't think we'll suffer any tragic consequences from having some redundancy in our database. In fact, that redundancy will allow us to move fast and try out other things (tordensky, plotting information extension of the DSL) while leaving the already existing "infrastructure" still working.

If you agree to splitting up things in the manner described here, I will wrap up this PR today or Monday.

WilliamHPNielsen:
@QCoDeS/core, okay, wrapped up and ready for review before merging.

All the refactoring of methods and validation logic is to follow shortly in separate PRs.

qcodes/dataset/descriptions.py
```diff
@@ -0,0 +1,67 @@
...
+        rundesc = cls(interdeps=idp)
+
+        return rundesc
+
+    def to_yaml(self):
+        """
+        Output the run description as a yaml string
+        """
+        yaml = YAML()
+        stream = io.StringIO()
```
astafan8: is it possible to do `with io.StringIO() as stream ... `?

qcodes/dataset/descriptions.py
```diff
@@ -0,0 +1,67 @@
...
+
+from ruamel.yaml import YAML
+
+from qcodes.dataset.dependencies import InterDependencies
+
+
+class RunDescriber:
+    """
+    The object that holds the description of each run in the database. This
+    object serialises itself to a string and is found under the run_description
```
astafan8: "...serializes itself..." - is `serialize` or `to_yaml` referred to here?

also, the docstring of the class does not have to state where the object is going to be "dumped" to with this precision :)

qcodes/dataset/param_spec.py
```diff
@@ -61,45 +64,79 @@ def depends_on(self):
...
-            self.name,
-            self.type,
-            self.label,
-            self.unit,
-            self._inferred_from,
-            self._depends_on
-        )
+        return ParamSpec(self.name, self.type, self.label, self.unit,
+                         deepcopy(self._inferred_from),
+                         deepcopy(self._depends_on))
```
astafan8: oo, good catch. is it because these are lists, not tuples?

qcodes/dataset/param_spec.py
```diff
@@ -61,45 +64,79 @@ def depends_on(self):
...
+        Write the ParamSpec as a dictionary
+        """
+        output: Dict[str, Any] = {}
+        output['name'] = self.name
+        output['paramtype'] = self.type
+        output['label'] = self.label
+        output['unit'] = self.unit
+        output['inferred_from'] = self._inferred_from
+        output['depends_on'] = self._depends_on
+        output['version'] = self.version
```
astafan8: does yaml has a convention for names of fields that are not describing the contents of an object but rather the object itself? like python has double-undescore for stull like `__class__` and `__path__`. shall we use smth like this for `version`?

qcodes/dataset/param_spec.py
```diff
@@ -1,19 +1,22 @@
...
+from copy import deepcopy
+
 
 # TODO: we should validate type somehow
 # we can't accept everything (or we can but crash at runtime?)
 # we only support the types in VALUES type, that is:
 # str, Number, List, ndarray, bool
 class ParamSpec:
 
+    version = 0  # used when serializing
```
astafan8: Is versioning of  `ParamSpec` enough? What happens when `RunDescription` changes, what version shall be changed then? I'd suggest to version only the top object (in our case, `RunDescription`) so that there is only one version that we want to care about.

qcodes/dataset/sqlite_base.py
```diff
@@ -112,6 +114,47 @@ class ConnectionPlus(wrapt.ObjectProxy):
 SomeConnection = Union[sqlite3.Connection, ConnectionPlus]
 
 
+def upgrader(func: Callable[[SomeConnection], None]):
```
astafan8: great!

qcodes/dataset/sqlite_base.py
```diff
@@ -112,6 +114,47 @@ class ConnectionPlus(wrapt.ObjectProxy):
...
+    Decorator for database version upgrade functions. An upgrade function
+    must have the name `perform_db_upgrade_N_to_M` where N = M-1.
+
+    The decorator takes care of logging about the upgrade and managing the
+    database versioning
+    """
+    name_comps = func.__name__.split('_')
+    if not len(name_comps) == 6:
+        raise NameError('Decorated function not a valid upgrader. '
+                        'Must have name "perform_db_upgrade_N_to_M"')
```
astafan8: "... where M and N are numbers that are.... bla bla"

qcodes/dataset/sqlite_base.py
```diff
@@ -112,6 +114,47 @@ class ConnectionPlus(wrapt.ObjectProxy):
 SomeConnection = Union[sqlite3.Connection, ConnectionPlus]
 
 
+def upgrader(func: Callable[[SomeConnection], None]):
+    """
+    Decorator for database version upgrade functions. An upgrade function
+    must have the name `perform_db_upgrade_N_to_M` where N = M-1.
```
astafan8: can the full interface of the upgrade function be described here? i mean, the fact that it has a single argument which is sql connection object, the fact that it either returns or raises (see a comment in the code), the fact that the return values are ignored, etc? 

i understand that one can look into an existing upgrader function and deduct this info, but that might be time consuming.

qcodes/dataset/sqlite_base.py
```diff
@@ -112,6 +114,47 @@ class ConnectionPlus(wrapt.ObjectProxy):
...
+    """
+    name_comps = func.__name__.split('_')
+    if not len(name_comps) == 6:
+        raise NameError('Decorated function not a valid upgrader. '
+                        'Must have name "perform_db_upgrade_N_to_M"')
+    if not ''.join(name_comps[:3]+[name_comps[4]]) == 'performdbupgradeto':
+        raise NameError('Decorated function not a valid upgrader. '
+                        'Must have name "perform_db_upgrade_N_to_M"')
+    from_version = int(name_comps[3])
+    to_version = int(name_comps[5])
```
astafan8: is the following check missing? :
```
if from_version != to_version - 1:
    raise Exception(f"versions have to be N-1 and N ............")
```

qcodes/dataset/sqlite_base.py
```diff
@@ -301,17 +344,11 @@ def perform_db_upgrade(conn: SomeConnection, version: int=-1) -> None:
             action(conn)
 
 
+@upgrader
 def perform_db_upgrade_0_to_1(conn: SomeConnection) -> None:
```
astafan8: shall all the upgrader infra and the upgraders move to a separate module like `sqlite_base_upgraders`?

qcodes/dataset/sqlite_base.py
```diff
@@ -382,8 +410,109 @@ def perform_db_upgrade_1_to_2(conn: SomeConnection) -> None:
     else:
         raise RuntimeError(f"found {n_run_tables} runs tables expected 1")
 
-    log.info('Succesfully upgraded database version 1 -> 2.')
-    set_user_version(conn, 2)
+
+@upgrader
+def perform_db_upgrade_2_to_3(conn: SomeConnection) -> None:
+    """
+    Perform the upgrade from version 2 to version 3
```
astafan8: does it make sense to list in the docstring what the upgrade consists of? or do we have this info somewhere else? or do we just look at the code?:)

qcodes/dataset/sqlite_base.py
```diff
@@ -382,8 +410,109 @@ def perform_db_upgrade_1_to_2(conn: SomeConnection) -> None:
...
+    """
+
+    no_of_runs_query = "SELECT max(run_id) FROM runs"
+    no_of_runs = one(atomic_transaction(conn, no_of_runs_query), 'max(run_id)')
+    no_of_runs = no_of_runs or 0
+
+    # Insert a new column, "run_description", and fill it out with information
+    # retrieved from the layouts and dependencies tables
+    # If one run fails, we want the whole upgrade to role back, hence the
```
astafan8: __role__ back :)

qcodes/dataset/sqlite_base.py
```diff
@@ -382,8 +410,109 @@ def perform_db_upgrade_1_to_2(conn: SomeConnection) -> None:
...
+
+            sql = f"""
+                   UPDATE runs
+                   SET run_description = ?
+                   WHERE run_id == ?
+                   """
+            cur = conn.cursor()
+            cur.execute(sql, (yaml_str, run_id))
+
+            log.info(f"Upgrade in transition, run number {run_id}: OK")
```
astafan8: what about the final `log.info("Upgrade done sucessfully")`?

qcodes/tests/dataset/test_database_creation_and_upgrading.py
```diff
@@ -151,6 +155,76 @@ def test_perform_actual_upgrade_1_to_2():
         assert len(c.fetchall()) == 2
 
 
+def test_perform_actual_upgrade_2_to_3():
+
+    v2fixpath = os.path.join(fixturepath, 'db_files', 'version2')
+
+    if not os.path.exists(v2fixpath):
+        pytest.skip("No db-file fixtures found. You can generate test db-files"
```
astafan8: could you remind me why we skip instead of fail? i'm asking because we have around 40 tests as skipped, and it seems that nobody looks at it. I wouldn't want to miss it when this test lands in that bucket of skipped tests.

qcodes/tests/dataset/test_database_creation_and_upgrading.py
```diff
@@ -151,6 +155,76 @@ def test_perform_actual_upgrade_1_to_2():
...
+
+    dbname_old = os.path.join(v2fixpath, 'empty.db')
+
+    with temporarily_copied_DB(dbname_old, debug=False, version=2) as conn:
+
+        assert get_user_version(conn) == 2
+
+        desc_query = 'SELECT run_description FROM runs'
+
+        with pytest.raises(RuntimeError):
```
astafan8: are we sure that the `RuntimeError` is related to the fact that `run_description` column does not exist in `runs` table? :) i'd recommend to add `match=` kwarg.

qcodes/tests/dataset/test_database_creation_and_upgrading.py
```diff
@@ -151,6 +155,76 @@ def test_perform_actual_upgrade_1_to_2():
...
+        perform_db_upgrade_2_to_3(conn)
+
+        assert get_user_version(conn) == 3
+
+        c = atomic_transaction(conn, desc_query)
+        assert len(c.fetchall()) == 0
+
+    dbname_old = os.path.join(v2fixpath, 'some_runs.db')
+
+    with temporarily_copied_DB(dbname_old, debug=False, version=2) as conn:
```
astafan8: two tests in one? i'd recommend to split the two - one for an empty db, and the second for "some_runs" db.

qcodes/tests/dataset/test_database_creation_and_upgrading.py
```diff
@@ -151,6 +155,76 @@ def test_perform_actual_upgrade_1_to_2():
...
+
+        desc = RunDescriber.from_yaml(yaml_str)
+        idp = desc.interdeps
+        assert isinstance(idp, InterDependencies)
+
+        # here we verify that the dependencies encoded in
+        # tests/dataset/legacy_DB_generation/generate_version_2.py
+        # are recovered
+
+        p1 = [p for p in idp.paramspecs if p.name == 'p0'][0]
```
astafan8: it's a bit weird that the instance of a parameter with name "p0" is called "p1" :) the same with p5 and p4.

qcodes/tests/dataset/test_descriptions.py
```diff
@@ -0,0 +1,86 @@
...
+    assert desc_1 != desc_2
+    assert desc_3 != desc_2
+
+
+def test_serialization_and_back(some_paramspecs):
+
+    idp = InterDependencies(*some_paramspecs[1].values())
+    desc = RunDescriber(interdeps=idp)
+
+    ser_desc = desc.serialize()
```
astafan8: i think that the contents of the `ser_desc` need to be asserted: what about the "Parameters" field, and "paramspecs" field? If by accident somebody renames the "paramspecs" field to "ParamSpecs_", the tests won't catch it, right?

qcodes/tests/dataset/test_paramspec.py
```diff
@@ -106,18 +123,45 @@ def test_copy(name1, name2, name3):
...
+    assert ser['version'] == p1.version
+
+
+def test_deserialize(version_0_serializations):
+
+    yaml = YAML()
+
+    for ser in version_0_serializations:
+        sdict = {a: b for (a, b) in yaml.load(ser).items()}
+        ParamSpec.deserialize(sdict)
```
astafan8: and no assertions on the return values?

qcodes/tests/dataset/test_paramspec.py
```diff
@@ -106,18 +123,45 @@ def test_copy(name1, name2, name3):
...
+    assert ser['inferred_from'] == p1._inferred_from
+    assert ser['version'] == p1.version
+
+
+def test_deserialize(version_0_serializations):
+
+    yaml = YAML()
+
+    for ser in version_0_serializations:
+        sdict = {a: b for (a, b) in yaml.load(ser).items()}
```
astafan8: what is the reason for supplying fixtures in yaml format? since only the `deserialize` method is tested, the fixtures can just be dicts, right?

qcodes/tests/dataset/test_descriptions.py
```diff
@@ -0,0 +1,86 @@
...
+
+def test_yaml_creation_and_loading(some_paramspecs):
+
+    for group in some_paramspecs.values():
+        paramspecs = group.values()
+        idp = InterDependencies(*paramspecs)
+        desc = RunDescriber(interdeps=idp)
+
+        yaml_str = desc.to_yaml()
+        assert isinstance(yaml_str, str)
```
astafan8: no assertions on the content of the yaml string? shall we at least test that that str is a valid yaml?

qcodes/dataset/descriptions.py
```diff
@@ -0,0 +1,67 @@
...
+    column in the runs table
+
+    Extension of this object is planned for the future, for now it holds the
+    parameter interdependencies
```
astafan8: it seems that the attributes of this class which are included into the serialized version have to conform to a certain interface: having `serialize` and `deserialize` methods that perform the "this class" <-> dict transformation. Can this be mentioned somewhere? if such an interface is not implemented in code as an abstract base class (or any other "code"-way of mentioning this information), then at least shall this information land in some docstring or comment?

qcodes/dataset/param_spec.py
WilliamHPNielsen: Exactly, lists are mutable. I would also really like to get rid of `add_depends_on` and `add_inferred_from`, but I suspect that things in Delft would break then?

astafan8: true. please, keep them for now.

qcodes/dataset/descriptions.py
WilliamHPNielsen: sure thing

qcodes/dataset/descriptions.py
WilliamHPNielsen: > "...serializes itself..." - is serialize or to_yaml referred to here?

The glorious combination of the two.

qcodes/tests/dataset/test_database_creation_and_upgrading.py
astafan8: i see. so you're ok with the fact that the users will test all of the qcodes, except the upgraders? i don't mind.

well, we have quite many tests that are skipped, hence i don't even bother looking up what has been skipped and what is not, hence if the upgrader test will get accidentally skipped, i wouldn't notice, but a possibly broken test might slip through.

qcodes/dataset/dependencies.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    Object containing the ParamSpecs of a given run
+    """
+
+    def __init__(self, *paramspecs: ParamSpec) -> None:
+        self.paramspecs = paramspecs
+
+    def __repr__(self) -> str:
+        output = self.__class__.__name__
+        output += '('
```
jenshnielsen: Could this simply be 

output +="(" + " ".join(self.paramspecs) + ")"

qcodes/dataset/descriptions.py
```diff
@@ -0,0 +1,67 @@
+import io
+from typing import Dict, Any
+
+from ruamel.yaml import YAML
```
jenshnielsen: we might need to handle that the conda package is `ruamel_yaml` e.g. catch the import error and try with this as an alternative.

test_requirements.txt
```diff
@@ -8,4 +8,5 @@ git+https://github.com/QCoDeS/pyvisa-sim.git
 lxml
 codecov
 asv
+ruamel.yaml
```
jenshnielsen: If I understand this correctly this is not only a test requirement but a regular requirement. That means that it should be declared in the setup.py file, unfortunately that would probably install the package under it's pip name on top of the Conda package 😞

qcodes/dataset/sqlite_base.py
```diff
@@ -112,6 +114,47 @@ class ConnectionPlus(wrapt.ObjectProxy):
...
+        raise NameError('Decorated function not a valid upgrader. '
+                        'Must have name "perform_db_upgrade_N_to_M"')
+    from_version = int(name_comps[3])
+    to_version = int(name_comps[5])
+
+    @wraps(func)
+    def do_upgrade(conn: SomeConnection) -> None:
+
+        log.info(f'Starting database upgrade version {from_version} '
+                 'to {to_version}')
```
jenshnielsen: missing 'f' on this line?

qcodes/dataset/sqlite_base.py
```diff
@@ -382,8 +410,109 @@ def perform_db_upgrade_1_to_2(conn: SomeConnection) -> None:
...
+    with atomic(conn) as conn:
+        sql = "ALTER TABLE runs ADD COLUMN run_description TEXT"
+        transaction(conn, sql)
+
+        for run_id in range(1, no_of_runs+1):
+
+            paramspecs: Dict[int, ParamSpec] = {}
+
+            rst_query = ("SELECT result_table_name FROM runs "
+                         f"WHERE run_id={run_id}")
```
jenshnielsen: Should this and below be done with ? replacement?

qcodes/dataset/sqlite_base.py
```diff
@@ -382,8 +410,109 @@ def perform_db_upgrade_1_to_2(conn: SomeConnection) -> None:
     else:
         raise RuntimeError(f"found {n_run_tables} runs tables expected 1")
 
-    log.info('Succesfully upgraded database version 1 -> 2.')
-    set_user_version(conn, 2)
+
+@upgrader
+def perform_db_upgrade_2_to_3(conn: SomeConnection) -> None:
```
jenshnielsen: Would it make sense to break this function up a bit?

qcodes/dataset/sqlite_base.py
```diff
@@ -382,8 +410,109 @@ def perform_db_upgrade_1_to_2(conn: SomeConnection) -> None:
...
+
+            sql = f"""
+                   UPDATE runs
+                   SET run_description = ?
+                   WHERE run_id == ?
+                   """
+            cur = conn.cursor()
+            cur.execute(sql, (yaml_str, run_id))
+
+            log.info(f"Upgrade in transition, run number {run_id}: OK")
```
jenshnielsen: Do we have a benchmark on the time required to do  an actual upgrade of a "big" database. I am slightly concerned that this will be slow and users will get impatient and interrupt the process

qcodes/tests/dataset/test_database_creation_and_upgrading.py
jenshnielsen: I think we should consider the sizes of the db's used. If they are "small" we can simply check in the fixtures after they are generated and avoid running this scrip every time

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: I don't think this error message is the right place to explain how to use the upgrader. It's just there to catch a typo, not to teach users how to write an upgrade function (nor to give them a false sense of understanding of the upgrade process).

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: Good point. Done.

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: Added.

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: What should depend on what, then?

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: That's indeed a fine place to put it. Done.

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: Done.

qcodes/tests/dataset/test_database_creation_and_upgrading.py
WilliamHPNielsen: So a `match` kwarg won't do it, since the `RuntimeError` raises from an `OperationalError`, but the idea is good. Done.

qcodes/tests/dataset/test_database_creation_and_upgrading.py
WilliamHPNielsen: Fixed.

qcodes/tests/dataset/test_database_creation_and_upgrading.py
WilliamHPNielsen: Done.

qcodes/tests/dataset/test_descriptions.py
WilliamHPNielsen: If they accidentally rename it both in `serialize` and `deserialize`, then nobody will notice, that's right.
But nonetheless you are right that it's good to have a test warn you when you change things like that.

qcodes/tests/dataset/test_paramspec.py
WilliamHPNielsen: Fixed.

qcodes/tests/dataset/test_paramspec.py
WilliamHPNielsen: Good point. Only the top-level object, the `RunDescriber` should have to care about the specific output (yaml/JSON/excel sheet), all other objects just serialize to a dict.

qcodes/dataset/descriptions.py
WilliamHPNielsen: Good point. For now, docstring. For the first extension, we can think about refactoring.

qcodes/dataset/dependencies.py
WilliamHPNielsen: With an additional comma, yes. Fixed.

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: Fixed.

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: Yes. Fixed.

qcodes/dataset/sqlite_base.py
astafan8: you're right, this might introduce circular imports.

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: Now that you mention it... I'll do that.

WilliamHPNielsen:
@jenshnielsen and @astafan8, thanks a lot for the reviews. I have addressed the immediate problems. What remains to decide is:

 * How do we want to do versioning of the objects that we serialise and write to disk? Once something has been saved as data, we are obliged to be able to read it back.
 * Should we commit the db files?

And then I should cough up some numbers for the duration of the upgrade.

astafan8:
* versioning of RunDescription: i guess it depends on if we consider its version bound to db version or not. if they are bound, then we don't need to come up with another versioning, but the upgrades will keep taking more and more time. i don't like this idea now. so, i propose to version RunDescription separately, with a private class attribute it should be ok for now.
* yaml: i have no clue.
* if they are small (i don't know, ~10-100kB each), then it should be fine.

jenshnielsen:
With regards to ruamel.yaml

We can add a dependency on it to setup.py requires. That would result in duplicated packages. 
But these packages not only have different names but also different install directories so it will not break anything.

The conda package is complied with the c extension so that one is supposedly more performant but I have no idea if that makes any real difference but perhaps we should do 
```
try:
   from ruamel_yaml import Yaml
except ImportError:
   from ruamel.yaml import Yaml 
```
or some such thing

We are using the ruamel 0.15 api which is marked as unstable and provisional. Do we want this?
This api uses the Yaml class construct whereas the other one uses the load/dump api. Do we want to do this?

astafan8:
( let's use stable yaml api. is it the only yaml package that we can use? does it make sense to use json because of this? )

WilliamHPNielsen:
> does it make sense to use json because of this? 

For the DSL that will be fine, but for the StationConfigurator where humans will write the relevant file, yaml is by far the best choice. So we will have it in our dependencies irrespective of this PR.

WilliamHPNielsen:
@QCoDeS/core if we agree to postpone the discussion about which db files to commit, I think this guy can fly now. (provided CI agrees)

environment.yml
```diff
@@ -25,6 +25,7 @@ dependencies:
  - ruamel_yaml
  - gitpython
  - testpath>=0.4.2 # 0.4.1 is bad due to https://github.com/conda-forge/testpath-feedstock/issues/7
+ - tqdm
```
astafan8: their github page says that it is from conda-forge. are we ok with this?

jenshnielsen: conda search seems to indicate that it's also in the main channel

jenshnielsen: But this should be added to setup.py install_requires or the use made optional

qcodes/dataset/sqlite_base.py
```diff
@@ -112,6 +116,56 @@ class ConnectionPlus(wrapt.ObjectProxy):
...
+    if not ''.join(name_comps[:3]+[name_comps[4]]) == 'performdbupgradeto':
+        raise NameError('Decorated function not a valid upgrader. '
+                        'Must have name "perform_db_upgrade_N_to_M"')
+    from_version = int(name_comps[3])
+    to_version = int(name_comps[5])
+
+    if not to_version == from_version+1:
+        raise ValueError(f'Invalid upgrade versions in function name: '
+                         f'{func.__name__}. Can only upgrade from version N'
+                         'to version N+1')
```
jenshnielsen: does it add value to append something like `this function tried to upgrade from {from_version} to {to_version}`

WilliamHPNielsen: Sure.

qcodes/dataset/dependencies.py
```diff
@@ -0,0 +1,50 @@
...
+                raise ValueError('Got invalid input. All paramspecs must be '
+                                 f'ParamSpecs, but {paramspec} is of type '
+                                 f'{type(paramspec)}.')
+
+        self.paramspecs = paramspecs
+
+    def __repr__(self) -> str:
+        output = self.__class__.__name__
+        tojoin = (str(paramspec) for paramspec in self.paramspecs)
+        output += f'({", ".join(tojoin)})'
```
astafan8: does this mean that there is no space between `output` and `f'({", ".join(tojoin)})'` parts?

qcodes/dataset/descriptions.py
```diff
@@ -0,0 +1,107 @@
...
+                             f'InterDependencies. Got {type(interdeps)}.')
+
+        self.interdeps = interdeps
+
+    def serialize(self) -> Dict[str, Any]:
+        """
+        Serialize this object into a dictionary
+        """
+        ser = {}
+        ser['Parameters'] = self.interdeps.serialize()
```
astafan8: I know, it's most probably a minor detail. Still, `Parameters` is capitalized, while `paramspecs` is not. I have a feeling that how these "names" (or identifiers) are written should have some sense since we are defining a DSL, right? I don't see any sense in `Parameters` being capitalized, and `paramspecs` not. But I can see myself implementing something that uses DSL and making stupid mistakes partially ( ;) ) because of the lack of convention in the DSL.

Another distraction is that `paramspecs` items are `ParamSpec`s while `Parameters` items are `InterDependencies`. But this is acceptable because DSL terms do not necessarily have to map 1-to-1 to class names in qcodes.

My suggestions would be:
* use only lowercase in serialization
* rename InterDependencies, but i don't have good ideas, hence this is not a suggestion

qcodes/dataset/descriptions.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        idp = InterDependencies.deserialize(ser['Parameters'])
+        rundesc = cls(interdeps=idp)
+
+        return rundesc
+
+    @staticmethod
+    def _ruamel_importer():
+        try:
+            from ruamel_yaml import YAML
```
astafan8: the module is loaded for every time `to_yaml`/`from_yaml` is called. does it take a considerable amount of time to load the yaml module? if so, this could be contributing significantly to the overall speed of `to_yaml`/`from_yaml` methods. does it make sense to put the imports out of the class, and define a flag `yaml_exists` or smth (or set YAML to None if it was not possible to load it) and use it in this method in order to improve performance? I also thought that it is generally a good practice to do all the imports on top (unless really necessary).

qcodes/tests/dataset/test_database_creation_and_upgrading.py
```diff
@@ -111,8 +116,10 @@ def test_perform_actual_upgrade_0_to_1():
 
         guid_table_query = "SELECT guid FROM runs"
 
-        with pytest.raises(RuntimeError):
+        try:
             atomic_transaction(conn, guid_table_query)
+        except RuntimeError as err:
+            assert str(err.__cause__) == 'no such column: guid'
```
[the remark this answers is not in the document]
astafan8: shouldn't `error_caused_by` be used here?

qcodes/tests/dataset/test_database_creation_and_upgrading.py
astafan8: ( for some reason, i don't see the fix; i reloaded the github page several times )

WilliamHPNielsen:
Hold up! I forgot to update the `sqlite_base` to actually insert a description when making new runs!

qcodes/dataset/data_set.py
```diff
@@ -457,6 +492,10 @@ def add_result(self, results: Dict[str, VALUE]) -> int:
         the name of a parameter in this DataSet.
         It is an error to add results to a completed DataSet.
         """
+
+        if not self.started:
```
astafan8: this is going to slow things down because `self.number_of_results` is a query :( how about caching the "started" fact?

WilliamHPNielsen: Before checking number of results:
```
=========================================================== 
                                         param1                                     
              ----------------------------------------------------------- ----------
                {'n_values': 10000, 'n_times': 2, 'paramtype': 'array'}    14.9±0ms 
                {'n_values': 100, 'n_times': 200, 'paramtype': 'array'}    219±0ms  
               {'n_values': 10000, 'n_times': 2, 'paramtype': 'numeric'}   631±0ms  
               {'n_values': 100, 'n_times': 200, 'paramtype': 'numeric'}   648±0ms  
              =========================================================== 
```
Now:
```
 =========================================================== 
                                         param1                                     
              ----------------------------------------------------------- ----------
                {'n_values': 10000, 'n_times': 2, 'paramtype': 'array'}    19.2±0ms 
                {'n_values': 100, 'n_times': 200, 'paramtype': 'array'}    227±0ms  
               {'n_values': 10000, 'n_times': 2, 'paramtype': 'numeric'}   688±0ms  
               {'n_values': 100, 'n_times': 200, 'paramtype': 'numeric'}   695±0ms  
              =========================================================== 
```
but I don't know what the natural variation is. I'll look into caching.

qcodes/tests/dataset/legacy_DB_generation/generate_version_2.py
```diff
@@ -47,12 +47,12 @@ def generate_DB_file_with_some_runs():
...
-    from qcodes.dataset.sqlite_base import connect, new_experiment
+    from qcodes.dataset.sqlite_base import connect
     from qcodes.dataset.measurements import Measurement
     from qcodes.dataset.experiment_container import Experiment
     from qcodes import Parameter
 
-    conn = connect(path)
+    connect(path)
     exp = Experiment(path)
     exp._new(name='experiment_1', sample_name='no_sample_1')
```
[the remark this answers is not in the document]
WilliamHPNielsen: Before doing anything drastic, make sure you realise what this line does: https://github.com/QCoDeS/Qcodes/blob/68e38ab71d5c33313534a1fb319bd6349f4fea18/qcodes/tests/dataset/legacy_DB_generation/generate_version_2.py#L146

The whole point is that the code executed here is obsolete. 🙂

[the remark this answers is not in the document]
astafan8: damn, i forgot, sorry :)

qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -491,3 +492,39 @@ def fb(xv, yv):
...
+    prematurely_loaded_ds = DataSet(run_id=1)
+    assert prematurely_loaded_ds.description == RunDescriber(InterDependencies())
+
+    ds.add_result({'ps1': 1, 'ps2': 2})
+
+    loaded_ds = DataSet(run_id=1)
+
+    print(loaded_ds.description)
+    print('+'*10)
+    print(desc)
```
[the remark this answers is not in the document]
WilliamHPNielsen: Whoops, hehe...

qcodes/dataset/sqlite_base.py
```diff
@@ -382,8 +428,265 @@ def perform_db_upgrade_1_to_2(conn: SomeConnection) -> None:
...
+            paramspec = ParamSpec(name=name, paramtype=paramtype,
+                                  label=label, unit=unit,
+                                  inferred_from=inferred_from)
+            paramspecs[layout_id] = paramspec
+
+        # second possibility: this parameter depends on another parameter
+        elif layout_id in deps:
+
+            setpoints = dependencies[layout_id]
+            depends_on = [paramspecs[idp].name for idp in setpoints]
```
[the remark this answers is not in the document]
WilliamHPNielsen: > I'd recommend to not rely on any assumptions unless those are dictated by the db schema.

or perhaps rather by the SQL query. But We resolved it in a more direct way.

PR #1228, opened 2018-08-13
driver / add block_during_ramp parameter to AMI430_3D

emartinez:
@WilliamHPNielsen 

The AMI430_3D magnet does not currently support non-blocking field ramps. A linear field ramp between two safe points is always safe if the ramp rates are safe, so this should be possible.

The new block_during_ramp parameter (default True), if set to False, will call the set_field methods of the children with the argument block=False, so that they get ramped in a non-blocking way.

### DIFF
--- a/qcodes/instrument_drivers/american_magnetics/AMI430.py
+++ b/qcodes/instrument_drivers/american_magnetics/AMI430.py
@@ -330,10 +330,12 @@ class AMI430(IPInstrument):
             return
 
         # Otherwise, wait until no longer ramping
+        log.debug(f'Starting blocking ramp of {self.name} to {value}')
         while self.ramping_state() == 'ramping':
             self._sleep(0.3)
         self._sleep(2.0)
         state = self.ramping_state()
+        log.debug(f'Finished blocking ramp')
         # If we are now holding, it was successful
         if state != 'holding':
             msg = '_set_field({}) failed with state: {}'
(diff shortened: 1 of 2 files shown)


sohailc:
>  A linear field ramp between two safe points is always safe if the ramp rates are safe, so this should be possible.

Not True! This assumes that the safe region is convex. We know that this is not always the case.

WilliamHPNielsen:
@sohailc how was it again..? The sequential ramping keeps you safe only if the safe region is star-shaped around the origin, right? This is more general than convex, but can also in principle be false. But then I guess all bets are off.

In any case, before we merge, let me extend the tests to cover this new parameter.

emartinez:
@sohailc wrote:

> Not True! This assumes that the safe region is convex. We know that this is not always the case.

All the BlueFors AMI magnet specifications in Copenhagen state:

> This magnet can be ramped at any combination of X/Y/Z simultaneous fields starting from zero to a maximum of +/-60, +/-10, +/-10 kG, respectively. The ramp rates of the individual X/Y/Z coils must be adjusted so that the ramp time from zero field to their desired target fields is the same for al coils.

I.e. the safe region is a right rectangular prism, which is convex. We know from the magnet acceptance test that the corners of this prism have been explored and are safe.

If this is not _always_ the case, then I would suggest to subclass the driver so that users of less exotic magnets (at least 5 fridges here) can benefit from reasonable safety assumptions.

WilliamHPNielsen:
@sohailc @emartinez I guess what I am trying to ask is: are there realistic use cases where a linear ramp is not safe but a first-down-then-up approach is? I can think of a few examples, but they are quite contrived. Unless the first-down-then-up actually goes all the way to the origin and then out to the target field value, I guess there generally isn't.

PR #1230, opened 2018-08-15
Add new driver for Newport AG-UC8

jorisvr:
Add new driver for Newport AG-UC8 piezo motion controller.
This device will be used in a new experiment at QuTech, Delft.

We developed this driver with two goals in mind:
1) to use it in our setup;
2) to learn more about QCoDeS.

With regards to 2) any feedback on this driver is very welcome. We hope to learn how QCoDeS can fit in our software development efforts.

The driver has been tested and appears to be fully working.
The code currently fails static type checking because we created a hierarchy of InstrumentChannel classes (instrument has 4 channels, each channel has 2 axes). This works but appears to be forbidden by the InstrumentChannel.__init__() constructor.

We discussed this driver briefly with @astafan8.


qcodes/instrument_drivers/Newport/AG_UC8.py
```diff
@@ -0,0 +1,255 @@
...
+        super().write("RS")
+        time.sleep(self.reset_delay)
+        self.write("MR")
+
+    def get_idn(self) -> Dict[str, Optional[str]]:
+        resp = self.ask("VE")
+        words = resp.strip().split()
+        if len(words) == 2:
+            model = words[0]
+            version = words[1]
```
astafan8: does it mean that the instrument does not return its serial number?

astafan8:
What really missing in this PR is a Jupyter notebook that explains how the driver works! we have those in `docs/examples/driver_examples`, so you can have a look, and write something similar.

astafan8:
by the way, our CI shows that there no problems in type checking, so the nesting of `InstrumentChannel`s worked. How do you get the type check error?

qcodes/instrument_drivers/Newport/AG_UC8.py
```diff
@@ -205,19 +288,32 @@ def __init__(self, name: str, address: str) -> None:
         self.write("MR")
 
     def get_last_error(self) -> int:
+        """Send a TE command (get error of previous command) and return
+        a numerical error code.
+
+        Return value 0 means no error.
+        See
```
astafan8: see what? :)

qcodes/instrument_drivers/Newport/AG_UC8.py
```diff
@@ -171,10 +244,20 @@ class Newport_AG_UC8(VisaInstrument):
...
-    command_delay = 0.002   # wait 2 ms between commands
-    reset_delay = 0.05      # wait 50 ms after reset command
+
+    # After a command which does not generate a response, a short
+    # delay is needed before we can send the following command.
+    command_delay = 0.002
+
+    # After a reset command, a longer delay is needed before
+    # we can send the following command.
+    reset_delay = 0.05
```
astafan8: these four attributes, do you foresee that they will need to be tweaked by users? if so, it might make sense to make them python properties (so that there is a docstring) or even qcodes parameters (with `set_cmd=None, get_cmd=None` and `initial_value=<these numbers>`). and also please explain them in the example notebook, so that the users are aware of these communication-related details.

qcodes/instrument_drivers/Newport/AG_UC8.py
```diff
@@ -171,10 +244,20 @@ class Newport_AG_UC8(VisaInstrument):
...
 
+    # By default, expect response to command within 1 second.
     default_timeout = 1.0
+
+    # Position scan commands (and absolute move commands) can take
+    # up to 2 minutes to complete.
     position_scan_timeout = 120.0
-    command_delay = 0.002   # wait 2 ms between commands
-    reset_delay = 0.05      # wait 50 ms after reset command
+
```
astafan8: in order to not introduce confusion, could you at least mention here that "slow" command is a position scan command (or better, rename the `_slow_command` method to `_position_scan_command`. Just so that we have a single term for this case, not two.

jorisvr:
Thanks for your quick and thorough review. I have rarely received so much constructive feedback. It is great to see that you are applying this level of scrutiny to code contributions.

I was going to respond to your earlier comments but could not find the time until today. I will provide answers now where still applicable.

I will also process your latest comments and make a new revision. However I want to test it on the actual device before committing, that will probably be on monday.

Thanks!

qcodes/instrument_drivers/Newport/AG_UC8.py
```diff
@@ -0,0 +1,255 @@
...
+        self.add_parameter("steps",
+                           label="Accumulated number of steps since last reset of the step counter",
+                           get_cmd="%dTP" % axis,
+                           get_parser=make_get_parser("%dTP" % axis))
+        self.add_parameter("status",
+                           label="Status of the axis (0: ready, 1: stepping, 2: jogging, 3: moving to limit)",
+                           get_cmd="%dTS" % axis,
+                           get_parser=make_get_parser("%dTS" % axis))
+
+        self.add_function("jog",
```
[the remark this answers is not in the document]
jorisvr: That actually simplifies some things for my driver, so great.

qcodes/instrument_drivers/Newport/AG_UC8.py
```diff
@@ -0,0 +1,255 @@
...
+    def write_channel(self, channel: int, cmd: str) -> None:
+        if channel != self._current_channel:
+            super().write("CC%d" % channel)
+            time.sleep(self.command_delay)
+            self._current_channel = channel
+        self.write(cmd)
+
+    def ask_channel(self, channel: int, cmd: str) -> str:
+        if channel != self._current_channel:
+            super().write("CC%d" % channel)
```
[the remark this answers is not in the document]
jorisvr: Oops, you found a bug. Thanks!

qcodes/instrument_drivers/Newport/AG_UC8.py
jorisvr: These are not tunable parameters.
The values deal with the communication behaviour of the device. I would hope that one set of values covers all possible cases. I can not exclude the possibility that the values will have to be tweaked for some combination of USB port and device in the future, but I think such cases should be treated as a bug fix instead of a setup-specific tuning.

astafan8: ok.

astafan8:
You are very welcome! Please take your time, and, indeed, another round of testing on an actual device is very important.

Concerning the type problem with `InstrumentChannel`. The reason `InstrumentChannel` insists on its parent being `Instrument` is that it has its write/ask methods which are forwarded to the parent instrument, which is handy. The reason `InstrumentChannel` is not `Instrument` but `InstrumentBase` is that `InstrumentChannel` does not need to be globally registered as an instrument (which is what happens to the `Instrument` instances). Could you try the following solution: could you try to edit the signature of the `InstrumentChannel` so that it also accepts `InstrumentChannel` like this `def __init__(self, parent: Union[Instrument, InsturmentChannel], name: str, **kwargs) -> None:` (notice the Union)? Let me know if this worked. If it does, than I think it can be part of this PR.

astafan8:
Great! Inform me when you're done with testing one the actual device, and then we can merge.

jorisvr:
> could you try to edit the signature of the InstrumentChannel so that it also accepts InstrumentChannel

Ok I get now why it wants Instrument instead of InstrumentBase. But your suggestion looks good. I tried it and it works: the type error is gone. Thanks!
I included the change in the new revision of this PR.

jorisvr:
The driver still works with the actual device. Tests in the example notebook run without problems.
I think it is ready to merge.

PR #1232, opened 2018-08-16
GroupParameter and Group (for visa commands that set/get more than one parameter at once)

sohailc:
I have created a separate PR for the GroupParameter and added a basic test


Dominik-Vogel:
@QCoDeS/core: This PR is a copy and paste from PR #972 (mostly this commit: https://github.com/QCoDeS/Qcodes/pull/972/commits/b5ddabfde0a2ce0ff94cdd191ccb6c1d06d0b1f5).

It is good practice to use git with its features to enable code traceability. Also, as the author of that code I feel uncomfortable with a 'copy and pasting' of my code without attribution.

Issue #1200, opened 2018-07-23
Feature requests for new dataset


ThorvaldLarsen:
We would like to move from old dataset structure in CPH for cQED measurements. I believe almost everything is in place except for plotting functionalities that is currently implemented in `show_num` and a metadata viewer.
The functions in question are:
- [ ] Subtract average along each row or column
- [ ] Transpose data (swap x and y axis on plot)
- [ ] Add multiple dataset to the same plot
- [ ] Function to extract metadata from a dataset (print/view a readable format of an instrument snapshot).
I only had a brief look at plottr so it is possible that some of these are already implemented but I believe at least the first point is missing. 

Also is it clear from which repository we should be cloning plottr from to have the latest updates?

@jenshnielsen @WilliamHPNielsen (@wpfff adding you to keep you in the loop).

astafan8:
* "Transpose data (swap axes)" is supported in plottr (you always need to select which parameters to plot against which)
* "Subtract average by row/column" has been [added to plottr](https://github.com/kouwenhovenlab/plottr/pull/9)
* "Extracting metadata" from the new DataSet is available: try the following:
```python
# import json # don't forget to import this :)
snapshot_json = dataset.get_metadata('snapshot')
snapshot_dict = json.loads(snapshot_json)
# explore 'snapshot_dict'
```
* "adding multiple dataset to the same plot" - what is meant here is a "union" of datasets; this is being implemented via `merge` function in #1214, so that `plot_by_id` can be called on a merged dataset.

nataliejpg:
High priority
- [ ] alazar multi channel parameter can't be registered to dataset with register_parameter function which means at the moment I can't save the magnitude AND phase data unless I take the data and then save it separately which is baaad
- [ ] loading the values of array parameter from the dataset doesn't work with dataset.get_data(), is there another method (and if so can an example go into the example notebook)? So far I'm unable to check if any of the array parameter datasets are correct :) 

Low(er) priority
- [ ] plot_by_id doesn't always do good axis scaling so I end up with overlapping axes labels
- [ ] plot_by_id doesn't have a title

nataliejpg:
another for lower priority:
- [ ] it is possible for me to save strings as parameter values but not to use them as setpoints (also is it possible for me to measure and save them in the new dataset? there are no examples of this), would be great if I could have strings as setpoints and axes labels (and the spacing just assumed to be linear)

astafan8:
[...]

PR #1235, opened 2018-08-20
Feature / add label to plot_by_id

astafan8:
One of the tiny requests from #1200 was to have a default title on plots created by `plot_by_id`. The title contains run id, experiment name, and sample name.

Contents:
- some code in `plot_by_id`
- updates to example notebooks which call `plot_by_id`


qcodes/dataset/plotting.py
```diff
@@ -5,6 +5,7 @@
 import matplotlib.pyplot as plt
 
 import qcodes as qc
+from qcodes import load_by_id
```
jenshnielsen: I think it's better to use the full path for imports within qcodes. Imports from the toplevel aliases easily lead to circular imports

astafan8: ok, will fix

PR #1236, opened 2018-08-20
Add feature to allow subselection of data with an sqlite where clause

jenshnielsen:
We need better tools for customizing the data selection from the database. A simple way could be to 
allow the user to inject a sqlite where clause into get_data as done here. But it's perhaps not the most pure solution? 

Never the less this allows you to do something like the following:

```python
from qcodes.dataset.data_set import  load_by_id
load_by_id(dataid).get_data('dac_ch1', 'dmm_v1', select_condition='dmm_v1 > 0.5')
```

If we want to go this way this needs:

- [ ] Examples
- [ ] A proper test 

before merge

@nataliejpg

### DIFF
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -565,7 +565,8 @@ class DataSet(Sized):
     def get_data(self,
                  *params: Union[str, ParamSpec, _BaseParameter],
                  start: Optional[int] = None,
-                 end: Optional[int] = None) -> List[List[Any]]:
+                 end: Optional[int] = None,
+                 select_condition: Optional[str] = None) -> List[List[Any]]:
         """ Returns the values stored in the DataSet for the specified parameters.
         The values are returned as a list of lists, SQL rows by SQL columns,
         e.g. datapoints by parameters. The data type of each element is based on
@@ -578,6 +579,13 @@ class DataSet(Sized):
         equal to the start, or if start is after the current end of the
         DataSet – then a list of empty arrays is returned.
 
+
+        You can also use the select_condition to subselect data according
+        to a SQLite where clause See http://www.sqlitetutorial.net/sqlite-where/
+        for examples. Note that as this is automatically combined with the
+        start end end conditions the where keyword is inserted automatically
+        and should be omitted from the string.
+
         For a more type independent and easier to work with view of the data
         you may want to consider using
         :py:meth:`qcodes.dataset.data_export.get_data_by_id`
@@ -588,6 +596,8 @@ class DataSet(Sized):
                ParamSpec objects
             - start:
             - end:
+            - select_condition: SQlite Where clause that is used to
+               subselect data e.g. 'my_column > 10'
 
         Returns:
             - list of lists SQL rows of data by SQL columns. Each SQL row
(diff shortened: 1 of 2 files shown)


WilliamHPNielsen:
I think this sounds like a good way of bringing the query-power of SQL up to the user (otherwise, what's even the point of having it?). How will this work with data saved as arrays?

astafan8:
nice! waiting for examples and tests.

#### Use case discussion

But still, for my understanding, what use case are we tackling here, "select only data that I want because I have too much data to load" or "select only data that I want because don't need the rest"?

I think the second use case can be delegated to the user as he knows what he wants to do with the data. Moreover, pandas and numpy methods are much more known and used among our users (right?) than SQL language. Hence, there is probably no need for this feature, while there is definitely a need for `load_data_as_numpy` or `load_data_as_pandas_dataframe` (btw, where are they?).

The first case seems legit, but then, if there was a looot of data to save, and the data saving speed was at stake, so the user opted for "save as array (aka binary blob)", and in the end the user cannot use this feature to obtain the data that he wants (instead of loading the whole huge blob), it will be sad.

WilliamHPNielsen:
@astafan8, I mainly agree with you, although I think that "most people are probably used to X, so let's not have Y" is a somewhat weak argument. It is, however, a good indicator of priorities.

Regarding having a function that will not always work, conditioned on something basically hidden to the user (how the data are saved), I think you are quite right that that is indeed problematic. I don't really know what the best solution here is... But I would expect that when a run is taken out of a database and put somewhere else, it will be unraveled/de-arrayed. Maybe there ought to be a way to do that on a run in the database?

astafan8:
@WilliamHPNielsen I did not understand, sorry, what do you mean with this?
> But I would expect that when a run is taken out of a database and put somewhere else, it will be unraveled/de-arrayed. Maybe there ought to be a way to do that on a run in the database?

WilliamHPNielsen:
@astafan8, let me see if I can make it more human-readable. 

Basically, I want all data in the database to be of `numeric` type. I recognise that it is sometimes necessary to input data as `arrays`, just to keep up with the data taking in a "live" measurement situation, but I would not consider having `array`s in the database ideal. Therefore, when we e.g. merge a run from database A into database B, I would expect that action to insert the data as `numeric`, irrespective of how it was originally stored (I write "expect", since we have not build that tool yet). But then that same tool could also be used to insert run x from database A back into database A, only now with `numeric` type parameters. What I am mumbling about is thus some kind of two step process: 1. dump data quickly into the database. 2. sanitise the data in the database. When and how step 2 should be executed is not very clear to me at the moment...

Does it make sense?

astafan8:
@WilliamHPNielsen sure, now it does. well, basically expanding the use case :) but i think these sort of things need to be explained in the docs, otherwise the number of people who get to realize that this is possible, is not as big as we would want to expect. To cover this, shall we make an issue (and implement later) a function that takes a run with 'array' values, and re-saves it as a different run as 'numeric'?

anyway, then, i guess, this feature is justified quite well.

Shall we add at least a test when we `get_data(..., "x > 5")` for 'array' values, so that we at least capture the behavior (i guess, there will be some sort of exception)? And a note in the docstring that 'start', 'end', and 'condition' do not work for 'array' values?

nataliejpg:
@WilliamHPNielsen @astafan8 Just wondering if there was a consensus on this or any plan for getting it either cleaned up and merged or closed?

PR #1239, opened 2018-08-20
Feature / rescale ticks and units in plot_by_id

astafan8:
Request along with #1200. This brings the "rescale_axis" feature of plotting functionality of the old dataset to the `plot_by_id` of the new dataset. The feature scale the ticks and the units (on the axes labels) to engineering units (`k`, `M`, `m`, `n`, etc.) so that the plots are more readable. For example, `2.000.000.000` in `V` becomes `2` in `GV` if all the numbers on this axis are not larger than 1.000.000.000.000 volts. The feature can be turned of by using a corresponding kwarg.

ToDo:
- [ ] more tests
- [ ] update "offline plotting tutorial"

### DIFF
--- a/qcodes/dataset/plotting.py
+++ b/qcodes/dataset/plotting.py
@@ -1,8 +1,11 @@
 import logging
+from collections import OrderedDict
+from functools import partial
 from typing import Optional, List, Sequence, Union, Tuple
 import numpy as np
 import matplotlib
 import matplotlib.pyplot as plt
+from matplotlib.ticker import FuncFormatter
 
 import qcodes as qc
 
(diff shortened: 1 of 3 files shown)


qcodes/dataset/plotting.py
```diff
@@ -229,3 +240,118 @@ def plot_on_a_plain_grid(x: np.ndarray, y: np.ndarray,
...
+        a tuple with the ticks formatter and the new label; in case it is not
+        possible to scale, the returned values are None's
+    """
+    ticks_formatter = None
+    new_label = None
+
+    unit = data_dict['unit']
+
+    if unit in _SI_UNITS:
+        maxval = np.nanmax(data_dict['data'])
```
WilliamHPNielsen: Should this be the max of the absolute value?

astafan8: i think it should. but the original implementation did not have it :) I'll change it and add a test for it.

astafan8:
@QCoDeS/core should the rescaling of units and ticks be enabled by default or not?

astafan8:
@QCoDeS/core and @ThorvaldLarsen Another question: should seconds be rescaled? :) `ks` (kiloseconds) looks weird :)

ThorvaldLarsen:
Definitely. We rarely have ks but use ns or us (microsecond) all the time. Thanks!

astafan8:

astafan8:
@WilliamHPNielsen added type annotations. ready for merge now?

PR #1242, opened 2018-08-22
Config cleanup

### LINKED ISSUE #1229: Update docs about QCoDeS config (Configuring_QCoDeS.ipynb)
Why:
* It has been last updated ~2 years ago.
* The notebook does not reflect all the config features (e.g. PR #1226), and does not describe how it is supposed to be used.

jenshnielsen:
A few changes to the config system to hopefully make it more logical

* load_config no longer has the side effect of setting the path to
   the config file as it doesn't actually update the current_config
   but just returns the config loaded from the file
* Make custom path behave like all other configs but load it last
   so it takes preference over the other config files
* Change `__repr__` to include the path of all config files used
* Refactor to reduce code duplication
* Correct a few typos

current_config_file is still there and points to the last config file
loaded as it did before. This is perhaps confusing? Do we want to
remove it altogether.

@WilliamHPNielsen I think this will solve your problem.

@nataliejpg Any issue with the change to path loading logic. The path specified file will still overwrite anything else

EDIT by William: This fixes #1229


astafan8:
Looks very nice!

Two questions:
* no extra tests needed?
* can we update notebook as well (https://github.com/QCoDeS/Qcodes/issues/1229)? or not in this PR?

jenshnielsen:
Not sure about tests. Changes in functionality are very limited. The only user visible change is probably the new `__repr__`   Updating the notebook is probably long overdue so lets do than here too

nataliejpg:
i like this :)

WilliamHPNielsen:
I just pushed a test for cascading. The test fails, but shouldn't it pass? Did I miss how the cascading works?

astafan8:
but where have you specified/assigned `cfg['core']['station_configurator']['default_folder']` before asserting the value?

WilliamHPNielsen:
Just noting that this fixes #1229

WilliamHPNielsen:
@astafan8, I haven't, that's the whole point. It is specified in the default config file that ships with QCoDeS, `qcodes/config/qcodesrc.json`. The idea is that the settings in this file are always applied, *unless* overwritten by another config file. That other config file can thus specify only the few settings that it wants to change.

WilliamHPNielsen:
But there was a stupid error in the assert statement in my test, so luckily the config system is not broken 🙂

PR #1243, opened 2018-08-23
Format ticks for non-SI-unit axis in a more readable way

Dominik-Vogel:
With @astafan8 fantastic PR #1239 the axis in `plot_by_id`  are scaled just like it was the case for the old dataset.
This PR fixes the case of unit less axis, that has not been touched in the named PR. With this PR these get scaled and displayed in the same way as all others.
@nataliejpg


WilliamHPNielsen:
I am also confused 😃
The `else` clause is not specifically for unit-less values, but for anything not `in _UNITS_FOR_RESCALING` (famous example: `e^2/hbar`). Why does that situation call for a not-None `ticks_formatter`?

Dominik-Vogel:
Sorry I overlooked that you @astafan8  were so thorough with your tests ;-).
It is very simple what this PR does: If you have some axis tick at 1e-5 (unitless) it will be rendered as 0.00001 by default. In this PR the same formater gets registered so that it will be rendered as '1e-5'.

But you are right maybe this should live somewhere else, so that we get nice axis exponentials. Let me play around a little and see where to put it.

Dominik-Vogel:
I added standard formatting now where the plots get generated.
This all is just a minor detail. 
The plot looks good now:
capture

ThorvaldLarsen:
@Dominik-Vogel looks great! Concerning the placement of 1e-5 I think it would make sense to place inside the unit brackets as it really is part of the unit after rescaling: Flux (1e-5 e^2/hbar).

Dominik-Vogel:
@ThorvaldLarsen I totally agree. But my impresion is that we should prioritize it lower as the plotting front end will change soon and it requires some extra tweaking. If it is very important to you, then we can certainly change it.

ThorvaldLarsen:
Not really important. Was only if it was a quick change.

astafan8:
@Dominik-Vogel there is some problem in CI when the notebook gets executed for docs.

Dominik-Vogel:
Hopefully the final version... I now use the same logic as is used for the SI units but instead of prepending the unit character (k, M, T...) I prepend the exponent (always going to the closest multiple of 3):

figure_1

(Yes If this works I will update the docstring, just to say before you do it @astafan8 ;-))

astafan8:
please add tests if you have time ;)

qcodes/dataset/plotting.py
```diff
@@ -320,8 +319,9 @@ def _make_rescaled_ticks_and_units(data_dict: Dict[str, Any]) \
     values like "1" instead of "0.000000001" while the units in the axis label
     are changed from "V" to "nV" ('n' is for 'nano').
 
-    The units for which this procedure is performed can be found in
-    `_UNITS_FOR_RESCALING`.
+    The units for which unit prefixes ared added can be found in
+    `_UNITS_FOR_RESCALING`. For all other units an exponential sacling factor
+    is added to the label i.e. `(10^3 x e^2/hbar)`.
```
astafan8: typos in this text :)

qcodes/dataset/plotting.py
```diff
@@ -169,18 +169,17 @@ def _get_label_of_data(data_dict: Dict[str, Any]) -> str:
...
 
-def _get_unit_of_data(data_dict: Dict[str, Any]) -> str:
-    return data_dict['unit'] if data_dict['unit'] != '' else ''
-
-
 def _make_axis_label(label: str, unit: str) -> str:
-    return f'{label} ({unit})'
+    label = f'{label}'
+    if unit != '' and unit is not None:
+       label += f' ({unit})'
```
astafan8: sweet!

qcodes/tests/dataset/test_plotting.py
```diff
@@ -59,35 +60,25 @@ def test_rescaled_ticks_and_units(scale, unit,
...
-        assert f"{param_name} ({expected_prefix}{unit})" == label
+        base_label = param_name
     else:
-        assert f"{param_label} ({expected_prefix}{unit})" == label
+        base_label = param_label
+    postfix = expected_prefix + unit
+    if postfix != '':
+        assert f"{base_label} ({postfix})" == label
+    else:
+        assert f"{base_label}" == label
```
astafan8: the test method became too complex: it basically mirrors the logic of the function under test. Such a test becomes difficult to read and more difficult to modify. Let's be careful next time. As the initial author of the test, i should've also been more careful.

Dominik-Vogel: Good point. Thanks for pointing that out, I got carried away ;-)

jenshnielsen:
Hypothesis prints a warning that we probably need to fix

```
qcodes/tests/dataset/test_plotting.py::test_rescaled_ticks_and_units
  /home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/hypothesis/strategies.py:587: 
HypothesisDeprecationWarning: Cannot sample from {'', 'permille', 'm', 'g', 'T', 'ohm', 'Pa', '%', 'W', 'F', 'Ω', 
'K', 'eV', 'unit', 'kg', 's', 'Wb', 'nW', 'S', 'N', 'rad', 'H', 'mol', 'V', 'A', 'Ohm', 'Hz', 'C', 'J'}, not an ordered 
collection. Hypothesis goes to some length to ensure that the sampled_from strategy has stable results 
between runs. To replay a saved example, the sampled values must have the same iteration order on 
every run - ruling out sets, dicts, etc due to hash randomisation. Most cases can simply use 
`sorted(values)`, but mixed types or special values such as math.nan require careful handling - and note 
that when simplifying an example, Hypothesis treats earlier values as simpler.

```

jenshnielsen:
@Dominik-Vogel Do you want to simplify the tests or should we just merge as is?

Dominik-Vogel:
@jenshnielsen I would love to merge it already.... I think the tests are not great but ok...

PR #1248, opened 2018-08-27
String-valued parameters support

astafan8:
This PR established full support for saving string-valued parameters to the qcodes dataset. `ParamSpec` with sqlite type class `'text'` were already in place, but it was not possible to retrieve the saved data because the default (and until recent times, even hard-coded) sqlite type class was 'numeric', hence our registered adapter would fail. Moreover, convenience plotting function were not aware of the string-valued data.

*Note*: a separate PR will be made for properly implementing 3D plotting. in this PR it does not work for all the cases.

Changes introduced here:
- [x] string-valued parameters can be registered with type `'text'`
- [x] `DataSaver.add_results` raises if an attempt is made to save a string value for a non-`'text'` parameter + test
- [x] `DataSaver.add_results` raises if an attempt is made to save a numeric value into a `'text'` parameter
- [x] add test for saving arrays of strings
- [x] make the 'numeric' adapter smart to return strings, if strings were saved into columns of type 'numeric' (which after this PR cannot longer happen)
- [x] update example notebook about dataset to explain how to save string data
- [x] https://github.com/QCoDeS/Qcodes/pull/1251 is a pre-requisite because it updates matplotlib so that plotting categorical data is supported out of the box
- [x] improve `plot_by_id` to use string-valued parameter values for tick labels for 2D plots
- [x] minor refactoring along the way
- [x] update example notebook for  `plot_by_id`, 2D case
- [x] for 2D plots, sorting by 'x' values is done only for 'line' plots (not for scatter, hence not for plots of string-valued data); one of the examples in the notebook is adjusted to test the case where numeric data for a line plot is not sorted

To-do in other PR:
- improve `plot_by_id` to use string-valued parameter values for tick labels for __3D__ plots + update example notebook for  `plot_by_id`
- rewrite example notebook to use `Measurement`, not `DataSet`
- move type checking against paramtype of ParamSpec into DataSet


astafan8:
@QCoDeS/core dear reviewers :) this PR is close to its completion. I'm inviting you to have a look and leave your comments modulo the following remarks:
* 3D example for the notebook is coming soon (see also other to-do's)
* determining the plot type inside `plot_by_id` is a bit fugly, especially with the handling of string values. Do you have ideas on how to refactor it apart of defining `determinte_plot_type(xvals, yvals, zvals` function that will internally call existing `determinte_plottype_from_setpoints`?
* regarding moving `ParamSpec`-based type checking to `DataSet.add_result(s)`, do you insist on doing it here, or can it be postponed until a different PR? I'd especially like to have https://github.com/QCoDeS/Qcodes/pull/1249 merged before looking to this.

jenshnielsen:
@astafan8  I will give it a review now and see if I can come up with some ideas for the plot_by_id issue. I think it's cool to leave out the moving around of type cheking till after both this and #1249 lands

astafan8:
After a dicsussion with @QCoDeS/core , it has been agreed that this PR is going to be reviewed and merged *without plots working for 3D case*. Hence, the contents of this PR only include necessary changes to data saving and loading, and `plot_by_id` for 2D case.

qcodes/tests/dataset/test_string_data.py
```diff
@@ -0,0 +1,168 @@
...
+
+import qcodes as qc
+from qcodes.dataset.measurements import DataSaver, Measurement
+from qcodes.dataset.param_spec import ParamSpec
+from qcodes.dataset.database import initialise_or_create_database_at
+from qcodes.dataset.data_export import load_by_id
+
+
+# These fixture can't be imported from test_dataset_basic because
+# Codacy/PR Quality Review will think they are unused.
```
jenshnielsen: I am not sure it we shoudn't just ignore that and import them anyway

jenshnielsen:
The tests are failing for some reason that looks related to this pr otherwise the logic seems sound to me.

astafan8:
ok, thanks for the review, @jenshnielsen. yes, i'm working on fixing the failed tests and the example notebook. Once that is done, i'll ask for the final approval.

astafan8:
@QCoDeS/core The PR is ready for final review.

qcodes/dataset/plotting.py
```diff
@@ -242,19 +246,32 @@ def plot_2d_scatterplot(x: np.ndarray, y: np.ndarray, z: np.ndarray,
     Returns:
         The matplotlib axis handles for plot and colorbar
     """
-    mappable = ax.scatter(x=x, y=y, c=z, **kwargs)
+    z_is_string_valued = isinstance(z[0], str)
+
+    if z_is_string_valued:
+        z_int = list(range(len(z)))
```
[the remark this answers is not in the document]
astafan8: you are absolutely right. I say in the example notebook that 3D cases are not properly supported yet in order not to spare the details that are likely to change very soon.

WilliamHPNielsen:
A note about defaults: I think that order of the string-valued data should be the order of insertion, not the alphabetical order. I realise that matplotlib defaults to alphabetical ordering, but I don't like that.

@QCoDeS/core 
Shall we have a quick vote about defaults? "Like" this comment for default insertion order, "dislike" this comment for default alphabetical ordering.

qcodes/dataset/plotting.py
```diff
@@ -131,7 +135,7 @@ def plot_by_id(run_id: int,
             xpoints = data[0]['data'][order]
             ypoints = data[1]['data'][order]
```
astafan8: @WilliamHPNielsen noticed a bug - this sorting should be happening only for the 'line' plots; it definitely should not happen for the string-valued arrays.

astafan8: as discussed with @QCoDeS/core , the solution is to make the sorting only relevant for 'line' plot, since for 'scatter' it is not needed. (the other option was to 'not sort' if 'x' is string-valued.)

astafan8:
2 new commits make the string-valued data appear on the plot axis in the order of its insertion, which means that bug is fixed.

PR #1249, opened 2018-08-27
Dataset support multiparameter

jenshnielsen:
Fixes parts of #1200 

- [x] Add support to Measurement to register and add results from multiparamters. 
- [x] A good deal of refactoring of the logic of add_result and register_parameter to make this possible
- [x] Some new convenience functions for setpoint names
- [x] The setpoint_full_names functions etc should really have individual tests. 
- [x] Examples of this would be good


qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -29,7 +30,7 @@ def empty_temp_db():
     # create a temp database for testing
     with tempfile.TemporaryDirectory() as tmpdirname:
         qc.config["core"]["db_location"] = os.path.join(tmpdirname, 'temp.db')
-        qc.config["core"]["db_debug"] = True
```
jenshnielsen: This should not have been commited

qcodes/dataset/measurements.py
```diff
@@ -90,31 +103,56 @@ def add_result(self,
...
 
         Raises:
             ValueError: if a parameter name not registered in the parent
                 Measurement object is encountered.
             ParameterTypeError: if a parameter is given a value not matching
                 its type.
         """
-        res = list(res_tuple)  # ArrayParameters cause us to mutate the results
+        # ArrayParameters and MultiParameters cause us to mutate
+        # the results.
```
astafan8: is this comment still relevant? you don't seem to mutate results, instead a new "list of resutls" is created now, right? also, does this change impact performance?

qcodes/dataset/measurements.py
```diff
@@ -296,7 +398,8 @@ def __init__(
         self.parameters = parameters
         # here we use 5 s as a sane default, but that value should perhaps
         # be read from some config file
-        self.write_period = write_period if write_period is not None else 5
+        self.write_period = float(write_period) \
+            if write_period is not None else 5.0
```
astafan8: why is this int/float business needed for write_period? also, if it is needed, just keep it in the `write_period.setter` only, there is no reason for other methods to care about this.

jenshnielsen: The existing logic was wrong and only worked because nothing was actually checked by mypy. You cannot use the Number ABC as a type. So I opted for making it float internally. The alternative is just make everything a union[float, int] I don't really care either which way.

astafan8: ok. now, the solution is fine, it's just that i'd rather have `float(..)` calls in one place.

qcodes/dataset/measurements.py
```diff
@@ -485,84 +590,156 @@ def register_parameter(
...
+                                          basis,
+                                          paramtype)
+        elif isinstance(parameter, MultiParameter):
+            self._register_multiparameter(parameter,
+                                          setpoints,
+                                          basis,
+                                          paramtype,
+                                          )
+        elif isinstance(parameter, Parameter):
+            self._register_individual_parameter(name,
```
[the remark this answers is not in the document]
jenshnielsen: sure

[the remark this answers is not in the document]
astafan8: hmmm... on the other hand, this method now does not accept `Parameter` instance.... a! maybe a `_register_parameter` method that just calls `_register_individual_parameter` inside would be a cleaner solution. not sure.

qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -1168,11 +1164,133 @@ def test_datasaver_multidimarrayparameter_as_numeric(experiment,
...
+    assert meas.parameters[
+               'dummy_channel_inst_ChanA_that_setpoint'].depends_on == ''
+
+    with meas.run() as datasaver:
+        datasaver.add_result((param, param()))
+
+    assert datasaver.points_written == 15
+    ds = load_by_id(datasaver.run_id)
+
+    assert ds.get_data('dummy_channel_inst_ChanA_this_setpoint') == [[5],
```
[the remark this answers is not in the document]
jenshnielsen: Yes indeed

qcodes/dataset/measurements.py
```diff
@@ -143,64 +181,40 @@ def add_result(self,
...
                 if not stuffweneed.issubset(stuffwehave):
                     raise ValueError('Can not add this result; missing '
-                                     f'setpoint values for {param}:'
+                                     f'setpoint values for {paramstr}:'
                                      f' {stuffweneed}.'
-                                     f' Values only given for {params}.')
+                                     f' Values only given for'
+                                     f' {found_parameters}.')
+
         if inserting_unrolled_array and inserting_as_arrays:
```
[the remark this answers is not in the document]
jenshnielsen: `inserting_as_arrays` is updated in line 162 and `inserting_unrolled_array` is updated in line 169 
I think this is correct and added a test for mixing numeric and array types that raises here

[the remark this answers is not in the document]
astafan8: indeed, i didn't look into the folded part so fthe code, sorry. Great!

qcodes/dataset/measurements.py
jenshnielsen: Removed the comment. 
My preliminary benchmarks indicate that the difference is small as long as the number of tuples added is small. I therefore think this is worth the performance trade off. The alternative would involve removing paramter tuples from the list for multiparameters. That does not seem like a good idea

jenshnielsen:
I think this addresses @astafan8 comments and now has an example and some more tests. @WilliamHPNielsen and @Dominik-Vogel  any opinions.

@nataliejpg I think this should solve your Multi Parameter issue

qcodes/dataset/measurements.py
```diff
@@ -90,31 +103,56 @@ def add_result(self,
         are not unraveled but stored directly for improved performance.
 
         Args:
-            res: a dictionary with keys that are parameter names and items
-                that are the corresponding values at this measurement point.
+            res_tuple: a tuple with the first element being the parameter name
```
Dominik-Vogel: Maybe we should adjust the doctring to say that one can also pass the parameter itself.

PR #1252, opened 2018-08-29
Pin hypothesis

jenshnielsen:
Due to bug with 3.69.11 and log capture

Should hopefully unbreak master


jenshnielsen:
Issue is reported here https://github.com/HypothesisWorks/hypothesis/issues/1546

PR #1254, opened 2018-08-30
5014c warning

### LINKED ISSUE #1231: AWG 5014c
Warn when using elements shorter than 250 points that software sequencing mode is used which has several limitations including not being able to jump and trigger on elements in the middle of the sequence. This happens silently and is first revealed when trying to turn on the channel. Further more 
remote activating the channel with `awg.ch1_state(1)` or so fails silently 

See Sequencer Mode in the awg5k7.chm help file 

```

Item | Hardware Sequencer Mode | Software Sequencer Mode
-- | -- | --
Minimum Waveform Length
AWG7000 Series | 960 points (Interleave On: 1920 points) | No limitations
AWG5000 Series | 250 points | No limitations
Waveform Length Granularity
AWG7000B/AWG7000C Series | 4 points (Interleave On: 8 points) | No limitations
AWG7000 Series | 64 points (Interleave On: 128 points) | No limitations
AWG5000 Series | 1 point | No limitations
Wait Trigger | No limitations | First element only
Go To | No limitations | Last element only
Event Jump To | No limitations | Unable to use
Command Jump | No limitations | Unable to use
Infinite Repeat | No limitations | Last element only
Restrictions of Waveform Memory | None | Software sequencer uses more waveform memory than hardware sequencer. Expand waveform data into the least common multiple of the  waveform length and the granularity to overcome the restriction of waveform  length granularity.  Repeat is implemented by repeating the waveform data in the  hardware waveform memory.

```

mltls:
Fixes #1231 

@jenshnielsen

### DIFF
--- a/qcodes/instrument_drivers/tektronix/AWG5014.py
+++ b/qcodes/instrument_drivers/tektronix/AWG5014.py
@@ -1256,6 +1256,20 @@ class Tektronix_AWG5014(VisaInstrument):
                 file and will be reset to factory default when the file is
                 loaded. Default: True.
             """
+        limit = 250
+        try:
+            for wf_elems in waveforms:
+                for wf_elem in wf_elems:
+                    wf_len = len(wf_elems)
+                    if wf_len < limit:
+                        raise Exception
+        except:
+            warnings.warn("Length of waveform element under {}. "
+                          "Instrument will use the software sequencer "
+                          "instead of the hardware sequencer. "
+                          "This imposes some limitations on the "
+                          "sequencing.".format(limit))
+
         packed_wfs = {}
         waveform_names = []
         if not isinstance(waveforms[0], list):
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -1256,6 +1256,20 @@ def make_awg_file(self, waveforms, m1s, m2s,
                 file and will be reset to factory default when the file is
                 loaded. Default: True.
             """
+        limit = 250
+        try:
+            for wf_elems in waveforms:
+                for wf_elem in wf_elems:
+                    wf_len = len(wf_elem)
+                    if wf_len < limit:
+                        raise Exception
```
astafan8: isn't it more readable and safer if the trick with an exception is avoided?

WilliamHPNielsen:
I think it would be good have a slightly more "luxurious" version of this check. I'm imagining something like having a function that can determine whether a given sequence is playable using a software sequencer (does it have illegal trigger waits, etc.) and another one that checks if a given sequence will trigger using the software sequencer. By implementing that, we will also be able to give very helpful warnings and error messages, along the lines of `Error: trigger wait on element 3 is incompatible with software sequencer, but length (120 points < 250 points) of element 5 forces instrument to use software sequencer. Cannot proceed`. Ultimately, we want to protect users and ourselves from ever having to figure out this stuff manually again.

PR #1255, opened 2018-09-03
Defaults for Smart color scale

Dominik-Vogel:
This is a commit that got lost in the pipeline to #1253


qcodes/dataset/plotting.py
```diff
@@ -29,7 +30,7 @@ def plot_by_id(run_id: int,
                                    Sequence[
                                        matplotlib.colorbar.Colorbar]]]=None,
                rescale_axes: bool=True,
-               smart_colorscale: bool=False) -> AxesTupleList:
+               smart_colorscale: Union[bool,None]=None) -> AxesTupleList:
```
jenshnielsen: This should be `Optional[bool]` i think. Which is equivalent but less ugly

qcodes/dataset/plotting.py
```diff
@@ -68,6 +69,9 @@ def plot_by_id(run_id: int,
         colorbar axes may be None if no colorbar is created (e.g. for
         1D plots)
     """
+    # defaults
+    smart_colorscale = smart_colorscale or config.gui.smart_colorscale
```
jenshnielsen: Is this logic correct?
Defalt value:
None or True => True
None or False => False
True or False => True
True or True => True 
False or False => True
Seems correct 
but 
False or True  => True
seems wrong to me?

Dominik-Vogel: That is indeed a very good point.

PR #1256, opened 2018-09-03
StationConfigurator

Dominik-Vogel:
Migrating the `StationConfigurator` from qdev-wrappers.
https://github.com/qdev-dk/qdev-wrappers/pull/68

Merging the `StationConfigurator` with the `Station`. The `StationConfigurator` has been a separate object in qdev_wrappers because the `Station` lives in qcodes core, but it conceptionally belongs into the `Station`.
The Station is meant to be a representation of the physical setup. And this description is given by the station config yaml file. 
So far the station has methods for snapshotting:
 - `snapshot_base`
 - `add_component`

and default measurements:
 - `set_measurement`
 - `measure`

and in this pr methods for serialization from a yaml are added:
 - `load_config_file`
 - `load_instrument`

I suggest that we deprecate the default measurement funcitonality in another PR, because as far as I know it is not used and it is only used by the legacy loop. Removing these function would further decouple dependencies.
I also suggest that we accommodate the control over the sample name in the `Station` object. (this would also satisfy @ThorvaldLarsen s request.)

Previous bootstrapping looked like this:
```
-create station (no arguments)
-create StationConfigurator, pass station
-call init_function to specify sample name
```
With this an experiment is bootstrapped by simply
```
-create station with sample name as argument
```

To go in small steps this PR addresses the minimal integration of the station configurator in the station. In late PRs documentation and possible refactorings of the internal structure might follow. Lets think about what we are committing to with this PR.


Dominik-Vogel:
Thank you @ThorvaldLarsen we are currently discussing were to set the sample name.

I think there should not be anything to setup for the folder structure other than some entries in the qcodesrc.

qcodes/config/station_configurator.py
```diff
@@ -0,0 +1,249 @@
...
+        self.filename = filename
+        # a list of method names that got added by :meth:`load_file`
+        self._added_methods: List[str] = []
+
+        self.load_file(self.filename)
+
+    def load_file(self, filename: Optional[str] = None):
+
+        # 1. load config from file
+        if use_pyyaml:
```
ThorvaldLarsen: Is there a reason to support usage of pyyaml? Seems like a good time to choose using ruamel (at least I think it is just better) instead of having two versions of dependencies spread in lab.

Dominik-Vogel: Yes there is unfortunately. There has been an issue with writing back the meta data with ruamel.. I still haven't found out why this happens but I am happy they have pyaml.

qcodes/instrument/parameter.py
```diff
@@ -855,7 +855,8 @@ def __init__(self, name: str,
 
         # Enable set/get methods if get_cmd/set_cmd is given
         # Called first so super().__init__ can wrap get/set methods
```
jenshnielsen: Can you update the doc string to explain how this really works?

Dominik-Vogel: sure thing

qcodes/station.py
```diff
@@ -179,3 +217,202 @@ def __getitem__(self, key):
...
+
+                self._added_methods.append(method_name)
+            else:
+                log.warning(f'Invalid identifier: ' +
+                            f'for the instrument {instrument_name} no ' +
+                            f'lazy loading method {method_name} could be ' +
+                            'created in the StationConfigurator')
+
+
+    def load_instrument(self, identifier: str,
```
astafan8: Some of the functionality of this function is already implemented in [qcodes/instrument/base.py -> find_or_create_instrument](https://github.com/QCoDeS/Qcodes/blob/4a737aa3a7001d5adbff026c107f3d704ebfdea5/qcodes/instrument/base.py#L730). Could you have a look at it and re-use/merge/else, if it makes sense?

qcodes/instrument/parameter.py
```diff
@@ -1798,3 +1801,33 @@ def set_raw(self, value: Union[int, float]) -> None:
 
         self._save_val(value)
         self._wrapped_parameter.set(instrument_value)
+
+
+class DelegateParameter(Parameter):
```
astafan8: Some time ago i implemented a version of this guy [here](https://github.com/kouwenhovenlab/v0_characterization/blob/461a622dd0140c4a00eb7ccfb74584086839bedb/v0_utils/qcodes_tools.py#L146) with support of snapshotting. Could you check it out and, perhaps, use some of the stuff here?

Dominik-Vogel: sure

Dominik-Vogel: Oh, great I was looking for it but could not find it.

qcodes/station.py
Dominik-Vogel: I had exactly that in mind. This is why I added the `revive_instance` kwarg. The reason I am not calling the 'find_or_create_instrument` function is that I need to do stuff in the different cases... (like removing parameters from the monitor) I think I will do a pr on the monitor and another one to refactor these two occurrences of similiar code...

WilliamHPNielsen:
@Dominik-Vogel how does it work when a user sets a scaling value for a parameter in the configuration file, is the original validator kept unchanged? If so, I think that might not be the best behaviour, since it leads to incorrect validation (I discovered this in the lab, where a user could not work in micro V, only mV for this reason). 

I don't know if the best thing is to try and change the validator behind the scenes or to explicitly overwrite the validator in the configuration file. What do you think?

qcodes/utils/helpers.py
```diff
@@ -703,3 +704,23 @@ def abstractmethod(funcobj):
...
+
+def _ruamel_importer():
+    try:
+        from ruamel_yaml import YAML
+    except ImportError:
+        try:
+            from ruamel.yaml import YAML
+        except ImportError:
+            raise ImportError('No ruamel module found. Please install '
+                              'either ruamel.yaml or ruamel_yaml to '
```
astafan8:
```suggestion
                              'either ruamel.yaml or ruamel_yaml.')
```

qcodes/utils/helpers.py
```diff
@@ -703,3 +704,23 @@ def abstractmethod(funcobj):
...
+                              'use the methods to_yaml and from_yaml')
```
astafan8:
```suggestion
```

qcodes/utils/helpers.py
```diff
@@ -703,3 +704,23 @@ def abstractmethod(funcobj):
...
+    return YAML
+
+
+YAML = _ruamel_importer()
+
+
+def get_qcodes_path(*subfolder) -> str:
```
astafan8: i find this name confusing in case the arguments are used, but i can't come up with a better one.

astafan8:
@Dominik-Vogel could you also edit the description of the PR so that we know what has been captured here, and when looking back into history it is easy to understand what has actually been done?

qcodes/station.py
```diff
@@ -217,3 +251,214 @@ def __getitem__(self, key):
...
+
+    def load_config_file(self, filename: Optional[str] = None):
+        def get_config_file_path(
+                filename: Optional[str] = None) -> Optional[str]:
+            filename = filename or get_config_default_file()
+            if filename is None:
+                return None
+            search_list = [filename]
+            if (not os.path.isabs(filename) and
+                get_config_default_folder() is not None):
```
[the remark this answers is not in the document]
Dominik-Vogel: Sorry, that is hard to see in the preview from Github: do you want to align `get_config_default_folder` with `os....`?

qcodes/station.py
```diff
@@ -217,3 +251,214 @@ def __getitem__(self, key):
...
+                    setattr(parameter, attr, val)
+                    # extra attributes that need parsing
+                elif attr == 'limits':
+                    lower, upper = [float(x) for x in val.split(',')]
+                    parameter.vals = validators.Numbers(lower, upper)
+                elif attr == 'monitor' and val is True:
+                    self._monitor_parameters.append(parameter)
+                elif attr == 'alias':
+                    setattr(instr, val, parameter)
+                elif attr == 'initial_value':
```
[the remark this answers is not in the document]
Dominik-Vogel: I have a faint memory that I had some trouble with the `__init__` kwarg. Let me try to remember....

qcodes/tests/test_station.py
```diff
@@ -201,3 +215,306 @@ def test_station_after_instrument_is_closed():
...
+  mock:
+    driver: qcodes.tests.instrument_mocks
+    type: DummyInstrument
+    {f'enable_forced_reconnect: {enable_forced_reconnect}'
+        if enable_forced_reconnect is not None else ''}
+    init:
+      gates: {{"ch1", "ch2"}}
+         """
+
+    def assert_on_reconnect(user_config_val: Union[bool, None],
```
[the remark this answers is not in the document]
Dominik-Vogel: I have used `Union` here on purpose because `None` actually acts as a value more than as a default. Let us try to find out what the intended use of `Optional` is.

qcodes/tests/test_station.py
```diff
@@ -201,3 +215,306 @@ def test_station_after_instrument_is_closed():
...
+                            expect_failure: bool) -> None:
+        qcodes.config["station_configurator"]\
+            ['enable_forced_reconnect'] = user_config_val
+        st = station_from_config_str(
+            get_instrument_config(instrument_config_val))
+        st.load_instrument('mock')
+        if expect_failure:
+            with pytest.raises(KeyError) as excinfo:
+                st.load_instrument('mock')
+                assert ("Another instrument has the name: mock"
```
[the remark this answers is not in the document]
Dominik-Vogel: yes, it should be thank you!

qcodes/utils/helpers.py
Dominik-Vogel: Same here...

qcodes/tests/test_station.py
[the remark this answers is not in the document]
jenshnielsen: From the typing docs `Optional[X] is equivalent to Union[X, None].` so I think those are fully equivalent. https://docs.python.org/3/library/typing.html#typing.Optional

qcodes/station.py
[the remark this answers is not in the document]
astafan8: yes, indeed :)

qcodes/tests/test_station.py
```diff
@@ -201,3 +215,306 @@ def test_station_after_instrument_is_closed():
...
+        Instrument.close_all()
+
+    for user_config_val in [None, True, False]:
+        assert_on_reconnect(user_config_val, False, True)
+        assert_on_reconnect(user_config_val, True, False)
+
+    assert_on_reconnect(True, None, False)
+    assert_on_reconnect(False, None, True)
+    assert_on_reconnect(None, None, True)
+
```
astafan8: could you use argument names in the `assert_on_reconnect` calls in this test to improve readability? reading sequences of True/False/None is not convenient.

qcodes/tests/test_station.py
```diff
@@ -201,3 +215,306 @@ def test_station_after_instrument_is_closed():
...
+    mock = st.load_instrument('mock', gates=["TestGate"])
+    assert "TestGate" in mock.parameters.keys()
+    assert len(mock.parameters) == 2  # there is also IDN
+
+    # special case of `name`
+    mock = st.load_instrument('mock', name='test')
+    assert mock.name == 'test'
+    assert st.components['test'] is mock
+
+    # test address
```
astafan8: please make this a separate test.

qcodes/tests/test_station.py
```diff
@@ -201,3 +215,306 @@ def test_station_after_instrument_is_closed():
...
+    assert_on_reconnect(None, None, True)
+
+
+def test_init_parameters():
+    st = station_from_config_str("""
+instruments:
+  mock:
+    driver: qcodes.tests.instrument_mocks
+    type: DummyInstrument
+    enable_forced_reconnect: true
```
astafan8: perhaps reuse `get_instrument_config` from `test_enable_force_reconnect`? :)

qcodes/tests/test_station.py
```diff
@@ -201,3 +215,306 @@ def test_station_after_instrument_is_closed():
...
+    assert str(mock.ch1.vals) == '<Numbers -10.0<=v<=10.0>'
+    assert mock.ch1() == 5
+    mock.ch1(7)
+    assert p() == 3
+    assert p.raw_value == 7
+    assert mock.ch1.raw_value == 7
+    assert (json.dumps(mock.ch1.snapshot()) ==
+            json.dumps(p.snapshot()['source_parameter']))
+
+
```
astafan8: could you please add an explicit test for `load_instrument` that shows that the instrument gets re-instantiated by default when it's called? and also for the "revive" kwarg as well?

qcodes/tests/test_station.py
```diff
@@ -201,3 +215,306 @@ def test_station_after_instrument_is_closed():
...
+    assert mock.ch1.offset == 0
+    assert isinstance(p.vals, validators.Numbers)
+    assert str(mock.ch1.vals) == '<Numbers -10.0<=v<=10.0>'
+    assert mock.ch1() == 5
+    mock.ch1(7)
+    assert p() == 3
+    assert p.raw_value == 7
+    assert mock.ch1.raw_value == 7
+    assert (json.dumps(mock.ch1.snapshot()) ==
+            json.dumps(p.snapshot()['source_parameter']))
```
astafan8: why comparing JSON dumps? if i remember correctly, pytest should be able to show reasonable introspection on asserting dictionaries.

qcodes/tests/test_station.py
```diff
@@ -201,3 +215,306 @@ def test_station_after_instrument_is_closed():
...
+
+
+def test_channel_instrument():
```
astafan8: Perhaps some extra clarification along the lines of:
```suggestion
def test_channel_instrument():
    """Test that parameters from instrument's submodule also get configured correctly"""
```

qcodes/instrument/parameter.py
```diff
@@ -2012,6 +2015,58 @@ def set_raw(self, value: Union[int, float]) -> None:
         self._wrapped_parameter.set(instrument_value)
 
 
+class DelegateParameter(Parameter):
```
astafan8: could you add some tests for this guy separately, please?

qcodes/tests/test_station.py
Dominik-Vogel: These are not dictionaries but yaml reader objects. I am comparing the json, because that is what eventually will be written to the db.

qcodes/station.py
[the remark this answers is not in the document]
Dominik-Vogel: I cannot use the kwarg, because some parameters have already been created as part of the `Instrument.__init__`

qcodes/tests/test_station.py
astafan8: do you mean that `mock.ch1.snapshot()` is a yaml reader object? how did that happen? so far as i understood, `mock` is a valid qcodes Instrument, isn't it?

> I am comparing the json, because that is what eventually will be written to the db.

hmm. but isn't the intention to ensure that the DelegateParameter's source_parameter is "equal" to the `ch1` parameter? and the taken approach is just to compare the snapshots? I can agree with the convenience of comparing the snapshot dicts, but not with comparing snapshot JSON strings, and my reasoning is that in case something is wrong it's much more convenient to let pytest show where the difference is in a dict instead of a cryptic JSON string. Also, i don't think this test really care about the actual format in which the snapshot is stored in some qcodes DB, it just needs to ensure that the "parameter delegation" is correct. Does this make sense?

qcodes/station.py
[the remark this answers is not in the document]
astafan8: ah! i got you :) For some idiotic reason i got self-confused before. so this helper function is intended for adjusting "properties" of already-created(!) parameters. Hence it is not preparing kwargs for instantiation, it's just altering a parameter instance. And we have both situations - adjusting existing instrument parameters, and creating completely new parameters - and for both of these situations this same function is reused. so, it's ok now :)

qcodes/station.py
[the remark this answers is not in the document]
Dominik-Vogel: why would you prefer it there? (The not is only referring to the `isabs` and not to  `get_config_default_folder`.)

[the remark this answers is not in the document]
astafan8: because of separating the "expression" of the "if" from the "body" of the "if".

please feel free to choose a way to do it, but i'd like to avoid the "blend" or the "if" "expression" and the "if" "body".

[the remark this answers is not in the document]
Dominik-Vogel: Ah, because it coincides with the indent of the body? That is a good point. It would improve readability if it had a different indent. Great! I just found the style guide for it: https://www.python.org/dev/peps/pep-0008/#multiline-if-statements
Thanks for teaching me this Mikhail!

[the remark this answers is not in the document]
astafan8: :)

qcodes/tests/test_station.py
```diff
@@ -201,3 +215,306 @@ def test_station_after_instrument_is_closed():
...
+    return "StationConfigurator" in st.components.keys()
+
+
+@pytest.fixture
+def example_station(example_station_config):
+    return Station(config_file=example_station_config)
+
+
+# instrument loading related tests
+def test_station_config_path_resolution(example_station_config):
```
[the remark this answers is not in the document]
Dominik-Vogel: You are absolutely right. I was feeling pressed in time so I did not add those comments. This is an excellent example of going slow to move fast: without me adding the comments it will take all others a lot longer to review... Thanks again Mikail!

qcodes/station.py
```diff
@@ -217,3 +251,212 @@ def __getitem__(self, key):
...
+    def load_config(self, config: Union[str, IO[AnyStr]]) -> None:
+        def update_station_configuration_snapshot():
+            class ConfigComponent:
+                def __init__(self, data):
+                    self.data = data
+
+                def snapshot(self, update=True):
+                    return self.data
+
+            self.components['StationConfigurator'] = ConfigComponent(
```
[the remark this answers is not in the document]
Dominik-Vogel: I called it `Config`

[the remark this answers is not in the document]
astafan8: good!

but with capital C? or no discussions anymore? :)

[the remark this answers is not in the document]
Dominik-Vogel: sure it is always up for discussions. it is indeed a good point to make it not capitalized...

qcodes/station.py
```diff
@@ -1,17 +1,45 @@
...
+    DelegateParameter)
+import qcodes.utils.validators as validators
+from qcodes.monitor.monitor import Monitor
 
 from qcodes.actions import _actions_snapshot
 
 
+log = logging.getLogger(__name__)
+
+PARAMETER_ATTRIBUTES = ['label', 'unit', 'scale', 'inter_delay', 'delay',
```
astafan8: There's no `delay` in parameter, but there is a post delay.
```suggestion
PARAMETER_ATTRIBUTES = ['label', 'unit', 'scale', 'inter_delay', 'post_delay',
```

qcodes/station.py
```diff
@@ -217,3 +251,212 @@ def __getitem__(self, key):
...
+            p = getattr(instr, name)
+            setup_parameter_from_dict(p, options)
+            # restore source
+            options['source'] = source
+
+        # add the instrument to the station
+        self.add_component(instr)
+
+        # restart Monitor
+        Monitor(*self._monitor_parameters)
```
astafan8: Can this be made optional? I don't think it's fair to fire up a `Monitor` in case the user didn't want it.

Also, we seem to need to incorporate this with the `monitor` argument of the `Station` `__init__`. I do not know what is the intent and history of it, so can't suggest anything. Also `Monitor` is not used in Delft, so I can't suggest what the best user experience should be. What I know for sure is that we need to make this optional. Maybe the following would work in the short term: if there are no parameters to monitor from yaml file, the `Monitor` is not initiated.

Another reason to make this step optional somehow is: what if a user already had a Monitor running (for some reason), and now he loads stuff from yaml station configuration, and all the parameters which he had set before are gone.

qcodes/station.py
Dominik-Vogel: Thanks for catching that. There used to be delay at the time of writing....

qcodes/station.py
Dominik-Vogel: Fair points.... I will think of something to make the monitor part optional.

Dominik-Vogel: ok how about that: we make the monitor init kwarg an optional boolean. When true a monitor is created by the station and gets repopulated with the station `_monitor_parameters`. When false, nothing happens, and when None the value is read from qcodesrc.json.
The situation with the monitor is far from ideal but I don't want to go down that rabbit hole. To start with the parameters are not public and on runtime no parameters can be added...

astafan8: so you are proposing the following kwarg, right?
* `monitor == True` - `Station` creates a monitor when loading instruments from config in case there are parameters to monitor.
* `monitor == False` - `Station` does nothing monitor-related
* `monitor == None` - the True/False is read from qcodesrc.json

I agree with this as a solution for now.

Clarification questions and suggestions:
* what is the default value for `monitor` kwarg? I suggest `None`.
* what is the default value of `monitor` in qcodesrc.json?  I suggest to just not have it in the default qcodesrc.json (or if that's not allowed by the schema, let's make it `False` then)
* do we need to make a settable `monitor` property on the Station for dynamic adjustments of the station's monitor-related behavior? i propose to not do that for now, let's see if we get requests in this direction from users.

astafan8:
@QCoDeS/core I've added some minimal documentation about Station configuration and Station itself. Please feel free to review that as well.

Dominik-Vogel:
Moving this to a new clean PR before final merge.

PR #1258, opened 2018-09-06
plot_by_id: pass kwargs to subplots and plot function

Dominik-Vogel:
This pr enables the use of kwargs in `plot_by_id`. The kwargs applicable to `plt.subplots` are passed upon axis creation. The kwargs that remain are passed to the consequently called plotting function like `plot`, `scatter` or `pcolormesh`.


qcodes/dataset/plotting.py
```diff
@@ -81,9 +94,14 @@ def plot_by_id(run_id: int,
     if axes is None:
         axes = []
         for i in range(nplots):
-            fig, ax = plt.subplots(1, 1)
+            fig, ax = plt.subplots(1, 1, **subplots_kwargs)
             axes.append(ax)
     else:
+        if len(subplots_kwargs) == 0:
```
jenshnielsen: Should this not be !=

Dominik-Vogel:
@jenshnielsen do you approve with the fixed comparison?

qcodes/dataset/plotting.py
```diff
@@ -23,6 +24,13 @@
 AxesTupleList = Tuple[List[matplotlib.axes.Axes],
                       List[Optional[matplotlib.colorbar.Colorbar]]]
 
+# list of kwargs for plotting function, so that kwargs can be passed to
+# :meth:`plot_by_id` and will be distributed to the respective plotting func.
+FIGURE_KWARGS = set(inspect.signature(plt.figure).parameters.keys())
+FIGURE_KWARGS.remove('kwargs')
+SUBPLOTS_KWARGS = set(inspect.signature(plt.subplots).parameters.keys())
+SUBPLOTS_KWARGS.remove('fig_kw')
+SUBPLOTS_KWARGS = FIGURE_KWARGS.union(SUBPLOTS_KWARGS)
```
jenshnielsen: For readability i would prefer if the union was called something else that the individual set

Dominik-Vogel:
I decided against a test because there are not any tests for the plotting yet (except for the labels) and it is not highest prio....

PR #1259, opened 2018-09-06
[WIP] smart color scale: add percentile limits for smart color scale.

Dominik-Vogel:
enable to limit the maximum percentage of datapoints that may be clipped by the smart color scale on either side of the distribution.

### DIFF
--- a/qcodes/utils/plotting.py
+++ b/qcodes/utils/plotting.py
@@ -6,12 +6,16 @@ For the legacy dataset see qcodes.plots
 """
 
 import logging
-from typing import Tuple
+from typing import Tuple, Union
 import numpy as np
 
 log = logging.getLogger(__name__)
 
-def auto_range_iqr(data_array: np.ndarray) -> Tuple[float, float]:
+Number = Union(float, int)
+
+def auto_range_iqr(data_array: np.ndarray,
+                   max_cutoff_percentile: Union[Tuple[Number], Number]=50
+) -> Tuple[float, float]:
     """
     Get the min and max range of the provided array that excludes outliers
     following the IQR rule.
@@ -22,14 +26,23 @@ def auto_range_iqr(data_array: np.ndarray) -> Tuple[float, float]:
     Args:
         data_array: numpy array of arbitrary dimension containing the
             statistical data
+        max_cutoff_percentile: percentile of data that may maximally be clipped
+            on both sides of the distribution.
+            If given a tuple (a,b) the percentile limits will be a and 100-b.
     returns:
         vmin, vmax: region limits [vmin, vmax]
     """
+    if is_instance(max_cutoff_percentile, tuple):
+        t = max_cutoff_percentile[0]
+        b = max_cutoff_percentile[1]
+    else:
+        t = max_cutoff_percentile
+        b = max_cutoff_percentile
     z = data_array.flatten()
     zmax = z.max()
     zmin = z.min()
     zrange = zmax-zmin
-    q3, q1 = np.percentile(z, [75, 25])
+    pmin, q3, q1, pmax = np.percentile(z, [b, 75, 25, 100-t])
     IQR = q3-q1
     # handle corner case of all data zero, such that IQR is zero
     # to counter numerical artifacts do not test IQR == 0, but IQR on its
@@ -42,4 +55,7 @@ def auto_range_iqr(data_array: np.ndarray) -> Tuple[float, float]:
     else:
         vmin = max(q1 - 1.5*IQR, zmin)
         vmax = min(q3 + 1.5*IQR, zmax)
+        # do not clipp more than max_cutoff_percentile:
+        vmin = min(vmin, pmin)
+        vmax = max(vmax, pmax)
     return vmin, vmax


jenshnielsen:
It would be super cool if we marked the clipped areas. The following prototype does that

```
import matplotlib.pyplot as plt
import numpy as np

cmap = plt.cm.get_cmap("viridis")
cmap.set_over('magenta')
cmap.set_under('cyan')
fig, ax = plt.subplots(1,1)
pc = ax.pcolormesh(np.random.rand(100,100), cmap=cmap)
cb =fig.colorbar(pc, extend='both')
pc.set_clim(0.1, 0.9)
```

Dominik-Vogel:
That is great! Thank you for the code, I initially also tried that but couldn't get it to work. I'll include this feature.

qcodes/dataset/plotting.py
```diff
@@ -177,7 +184,8 @@ def plot_by_id(run_id: int,
             if rescale_axes:
                 _rescale_ticks_and_units(ax, data, colorbar)
             if smart_colorscale:
```
jenshnielsen: could we add a info log message here with the limits used and how much is clipped

Dominik-Vogel:
I just remembered the problem: It is all fine to set up the extend as you said, but I need to set it up after the colorbar has been created, and this does not seem to work well....

Dominik-Vogel:
Ok finally found it @jenshnielsen . It seems to be an inconsistency in matplotlib. I will also need to redefine the slice that defines the points inside the colorbar:
```
cb.extend = 'both'
cb._inside = cb._slice_dict[cb.extend]
cb.mappable.set_clim(0, 0.3)
````
Maybe we can make a PR for matplotlib next week...

**Comment by Dominik-Vogel** [comment]
Now I get :-) :
```
axes, cb = plot_by_id(54,smart_colorscale=False)
apply_auto_color_scale(cb[0], cutoff_percentile=(0.5,0.5))
```
image

File: qcodes/dataset/plotting.py [hunk 2]
```diff
@@ -75,21 +77,18 @@ def plot_by_id(run_id: int,
             with standard SI units will be rescaled so that, for example,
             '0.00000005' tick label on 'V' axis are transformed to '50' on 'nV'
             axis ('n' is 'nano')
-        smart_colorscale: if True, the colorscale of heatmap plots will be
-            automatically adjusted to disregard outliers. If False,
-            the adjustment will not be performed. If None, the value from
-            QCoDeS config->"gui"->"smart_colorscale" will determine if the
-            adjustment is going to be performed.
+        auto_color_scale: if True, the colorscale of heatmap plots will be
+            automatically adjusted to disregard outliers.
```
astafan8: description for `cutoff_percentile` is missing

qcodes/tests/test_config.py
```diff
@@ -161,7 +161,7 @@ def default_config():
     qcodes.Config.cwd_file_name = default
     qcodes.Config.schema_cwd_file_name = default_schema
 
-    default_config_obj = qcodes.config
+    default_config_obj = copy.deepcopy(qcodes.config)
```
astafan8: seems like a good catch!

qcodes/utils/plotting.py
```diff
@@ -1,17 +1,33 @@
...
+import qcodes
 
 log = logging.getLogger(__name__)
 
-def auto_range_iqr(data_array: np.ndarray) -> Tuple[float, float]:
+Number = Union[float, int]
+
+
+"""
+General functions (independent of plotting frame work)
```
astafan8: typo(?): framework (without a space)

qcodes/utils/plotting.py
```diff
@@ -42,4 +67,180 @@ def auto_range_iqr(data_array: np.ndarray) -> Tuple[float, float]:
     else:
         vmin = max(q1 - 1.5*IQR, zmin)
         vmax = min(q3 + 1.5*IQR, zmax)
+        # do not clipp more than cutoff_percentile:
```
astafan8: typo: clip

qcodes/utils/plotting.py
```diff
@@ -42,4 +67,180 @@ def auto_range_iqr(data_array: np.ndarray) -> Tuple[float, float]:
...
+                             color_over: Optional[Any]=DEFAULT_COLOR_OVER,
+                             color_under: Optional[Any]=DEFAULT_COLOR_UNDER
+                             ) -> None:
+    """
+    Applies limits to colorscale and updates extend.
+
+    This function applies the limits `new_lim` to the heatmap plot associated
+    with the provided `colorbar`, updates the colorbar limits, and also adds
+    the colorbar clipping indicators in form of small triangles on the top and
+    bottom of the colorbar, according to the where the limits are exceeded.
```
astafan8: typo: no need for "the" in front of "where"

qcodes/utils/plotting.py
```diff
@@ -42,4 +67,180 @@ def auto_range_iqr(data_array: np.ndarray) -> Tuple[float, float]:
...
+    if not isinstance(colorbar.mappable, matplotlib.collections.QuadMesh):
+        raise RuntimeError('Can only scale mesh data, but received '
+                           f'"{type(colorbar.mappable)}" instead')
+    if data_lim is None:
+        if data_array is None:
+            data_array = colorbar.mappable.get_array()
+        data_lim = np.nanmin(data_array), np.nanmax(data_array)
+    else:
+        if data_array is not None:
+            raise RuntimeError('You cannot specify `data_lim` and `data_array` '
```
astafan8: could the message be rephrased a bit, so that it's clear that both cannot be supplised *at the same time*? also, a punctuation before "please" is missing.

qcodes/utils/plotting.py
```diff
@@ -42,4 +67,180 @@ def auto_range_iqr(data_array: np.ndarray) -> Tuple[float, float]:
...
+    if exceeds_min and exceeds_max:
+        extend = 'both'
+    elif exceeds_min:
+        extend = 'min'
+    elif exceeds_max:
+        extend = 'max'
+    else:
+        extend = 'neither'
+    colorbar.extend = extend
+    # the colorbar has no setter method and setting the colorbar extend does
```
astafan8: shall we extract it this as a separate method, so that it does not disturb the main logic of the function? also, when matplotlib fixes it, it will be (yeah, just a tiny bit) easier for use to adapt the code ))

qcodes/utils/plotting.py
```diff
@@ -42,4 +67,180 @@ def auto_range_iqr(data_array: np.ndarray) -> Tuple[float, float]:
...
+                           ) -> None:
+    """
+    Sets the color limits such that outliers are disregarded.
+
+    This method combines the automatic color scaling from
+    :meth:`auto_range_iqr` with the color bar setting from
+    :meth:`apply_color_scale_limits`.
+    If you want to adjust the color scale based on the configuration file
+    `qcodesrc.json`, use :meth:`auto_color_scale_from_config`, which is used
+    In :meth:`plot_by_id` and `show_num` (in qdev-wrappers),
```
astafan8: i don't think we should refer to functions from qdev-wrappers. this reference just does not seem to belong here.

astafan8:
Addition to my review: there are also some typos in the example notebook, so it would be nice if the text can be revisited and those are fixed. but I leave it up to you ))

qcodes/utils/plotting.py
astafan8: Also, i do not think that mentioning `show_num` in the example notebook is appropriate. `show_num` is known by a relatively specific group of users, while qcodes in general is targeting a broader range of users. It seems more fair to me to say smth like "..and custom plotting solutions" instead of mentioning `show_num`.

qcodes/config/qcodesrc_schema.json
```diff
@@ -82,10 +82,33 @@
...
+                            "type": "array",
+                            "items": {
+                                "type": "number",
+                                "type": "number"
+                            },
+                            "default": [0.5, 0.5]
+                        },
+                        "color_over":{
+                            "description": "Matplotlib color representing the datapoints clipped by the upper limit",
+                            "default": "white"
```
astafan8: why are the defaults here different from the defaults in the kwargs of the functions? what does it actually supposed to signify? (for example, that for qcodes users we recommend white/gray, and for users who implement custom plotting solutions that use our autoscaling functions should default to magenta/cyan?)

Dominik-Vogel: So the colors here I chose different on purpose because we also define a default color scale in the config file and the over and under colors have to be matching the color scale. But thank you for asking, as I just realized that the default color map is hot and viridis. So white and black are extremely poor choices here.
Cyan an mangenta are kind of safe colors to use for any color scale ;-)

Dominik-Vogel: I just realized that the colormap from the config is not being used in `plot_by_id` and added this here.

docs/Makefile
```diff
@@ -233,17 +233,15 @@ dummy:
...
-	jupyter-nbconvert --to rst --ExecutePreprocessor.timeout=600 --ExecutePreprocessor.kernel_name=python3 --execute 'examples/DataSet/*.ipynb' --output-dir=_notebooks/DataSet
-	jupyter-nbconvert --to rst 'examples/driver_examples/*.ipynb' --output-dir=_notebooks/driver_examples
-	jupyter-nbconvert --to rst 'examples/benchmarking/*.ipynb' --output-dir=_notebooks/benchmarking
+	jupyter-nbconvert --to rst 'examples/*.ipynb' --output-dir='_notebooks'
+	jupyter-nbconvert --to rst 'examples/DataSet/Real_instruments/*.ipynb' --output-dir='_notebooks/DataSet/Real_instruments'
+	jupyter-nbconvert --to rst --ExecutePreprocessor.timeout=600 --ExecutePreprocessor.kernel_name=python3 --execute 'examples/DataSet/*.ipynb' --output-dir='_notebooks/DataSet'
+	jupyter-nbconvert --to rst 'examples/driver_examples/*.ipynb' --output-dir='_notebooks/driver_examples'
+	jupyter-nbconvert --to rst 'examples/benchmarking/*.ipynb' --output-dir='_notebooks/benchmarking'
+	jupyter-nbconvert --to rst "examples\plotting\*.ipynb" --ExecutePreprocessor.timeout=600 --ExecutePreprocessor.kernel_name=python3 --output-dir="_notebooks\plotting"
+  cp -R "examples/files" "_notebooks/files"
```
astafan8: is indentation ok? (it probably does not matter though... i forgot)

Dominik-Vogel: yeah, that is apparently two spaces vs two tabs...

PR #1260, opened 2018-09-06
Fix deprecation warnings

jenshnielsen:
Pytest 3.8 has been released and this now shows deprecation warnings by default. Turns out we trigger a few of those. This should fix that. This is WIP until all of them have been fixed. Please don't merge just yet

### DIFF
--- a/qcodes/data/format.py
+++ b/qcodes/data/format.py
@@ -13,7 +13,7 @@ class Formatter:
 
     Each Formatter is expected to implement writing methods:
 
-    - ``write``: to write the ``DataArray``\s
+    - ``write``: to write the ``DataArrays``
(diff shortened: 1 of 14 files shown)


jenshnielsen:
I think this now takes care of them all apart from ```<unknown>:251: DeprecationWarning: invalid escape sequence \H``` which i cannot identify where is coming from. It may be triggered in a dependency.

jenshnielsen:
@WilliamHPNielsen Can you have a look at the changes in the AWG drivers. As far as I can tell the strings are equivalent but do not raise warnings now. There are also a few other changes to regexs that i would like a second pair of eyes on. (There were pushed after dominiks review)

WilliamHPNielsen:
Looks kosher to me.

PR #1262, opened 2018-09-07
[wip] Dataset serializer

WilliamHPNielsen:
- Make it possible to serialize a DataSet object and dump it to a string
- Make it possible (under some circumstances) to re-create a DataSet object from said string

This PR uses the `jsonpickle` module for the serialization. A custom handler for the `DataSet` is added (and registered when the `dataset` module is imported).

The usage is straightforward:

```python
import jsonpickle

ds = load_from_id(1)
json_string = jsonpickle.dumps(ds)
new_ds = jsonpickle.loads(json_string)
```

There's still quite a lot to discuss about what exactly should be jsonpickled down (like, does `results_table_name` make _any_ sense?) and how datasets not already in the database should be treated.

@johnazariah

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,11 +1,11 @@
 name: qcodes
-channels: 
+channels:
  - defaults
  - conda-forge
 dependencies:
  - h5py
  - matplotlib=2.2.3
- - pyqtgraph 
+ - pyqtgraph
  - python=3.6
  - jsonschema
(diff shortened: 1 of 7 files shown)


WilliamHPNielsen:
Closing this PR since #1388 and #1415 have developed a similar idea, but taken it much further.

PR #1269, opened 2018-09-14
driver/SignalHound Spectrum analyser refactor to use qcodes parameters

nataliejpg:
Strips and refactors the signal hound USB SA1243 spectum analyser driver so that it uses parameters for setting setting and for getting traces, removes option for setting any modes other than sweeping and taking averages as I could not see that these were actually being set or working 

@jenshnielsen  sorry I though I'd already made this pr, completely feel free to overwrite it, just fyi


jenshnielsen:
I am going to replace this with #1284 which has some additional cleanup on top of this driver

PR #1272, opened 2018-09-17
Remove mypy workarounds

WilliamHPNielsen:
Now that `mypy` version `0.630` is out, we can remove a few workarounds from our code.

@QCoDeS/core


WilliamHPNielsen:
@astafan8 yes, that's a good point, it will complain on `0.620`.

PR #1277, opened 2018-09-19
Enhance run table query performance with indices

geoffroth:
- Adds two indices to the experiments.db runs table, one on the "exp_id" column and one on the "guid" column.
This should enhance performance (especially in large tables) for queries that retrieve data by either column. Indices are DESC based on the assumption that we will more likely be querying higher / more recent exp_ids and guids more often than older data.
- For clarity, indices definitions are grouped directly below the table definitions they modify.
- For convenience, and to hopefully avoid errors, indices are created after any necessary migrations have completed.

@WilliamHPNielsen


jenshnielsen:
I think this fails the tests because it's trying to add the index to the runs table with out a guid colloum in the upgrade 0 to 1 test. I think making this an upgrade too would solve the problem. @WilliamHPNielsen I don't know if that in this case should go in before the dependencies upgrade

WilliamHPNielsen:
@jenshnielsen, I agree that this should be a distinct upgrade, and I think it makes sense that this PR gets database version number 2, since it is already done.

@geoffroth should I make this into an upgrade or will you do that?

WilliamHPNielsen:
Also: great PR, much appreciated! 

If anyone has time for benchmarking, it would be great to see if we do get the expected `O(log(n))` instead of `O(n)` on selected queries.

geoffroth:
@WilliamHPNielsen Since it's just a few lines, would it make sense for you to perhaps simply incorporate this or similar into the db refactor you're currently working on? Otherwise, I'd certainly be happy to rewrite as upgrade 2.

WilliamHPNielsen:
@geoffroth, my refactor is still one or two weeks out in the future, so I would prefer to have this be a separate upgrade that we can put in now. Thanks.

qcodes/tests/dataset/legacy_DB_generation/utils.py
```diff
@@ -1,11 +1,26 @@
...
+# Version 0: the original table schema, runs, experiments, layouts,
+# dependencies, result-tables
+#
+# Version 1: a GUID column is added to the runs table
+
+
+GIT_HASHES: Dict[int, str] = {0: '78d42620fc245a975b5a615ed5e33061baac7846',
+                              1: '056d59627e22fa3ca7aad4c265e9897c343f79cf'}
 
 DB_NAMES: Dict[int, List[str]] = {0: ['']}
```
astafan8: is this used?

qcodes/tests/dataset/legacy_DB_generation/utils.py
```diff
@@ -1,11 +1,26 @@
...
 
-GIT_HASHES: Dict[int, str] = {0: '78d42620fc245a975b5a615ed5e33061baac7846'}
+from git import Repo
+
+# A brief overview of what each version introduces:
+#
+# Version 0: the original table schema, runs, experiments, layouts,
+# dependencies, result-tables
+#
+# Version 1: a GUID column is added to the runs table
```
astafan8: is description of version 2 missing?

qcodes/tests/dataset/legacy_DB_generation/utils.py
WilliamHPNielsen: Nein. Should have been removed.

qcodes/tests/dataset/legacy_DB_generation/utils.py
WilliamHPNielsen: We will add that when we need it, i.e. when we have version 3 on the table.

PR #1282, opened 2018-09-26
Fix decadac docstring

jenshnielsen:
The decadac docstrings were out of date mentioning port and not address and claimed that the driver does not support more than one slot.

This pr corrects that and pep8's the driver. I have also deleted the notebook for now. Once I get a chance test on an actual decadac I will write a new notebook. This has not been tested on an actual instrument but does hopefully not include any actual functional changes


astafan8:
mergeable? :)

jenshnielsen:
Yes, I would have liked to test it on a real instrument but I will do that in the next pr. This has zero functional changes

PR #1288, opened 2018-10-02
Restructure infrastructure for database upgrade testing

WilliamHPNielsen:
Background: when upgrade the database version, our API loses its ability to generate databases in of versions. For testing the upgrade, we need a way to generate those db-files.

This PR proposes to have a folder (in `qcodes/tests/dataset`) with scripts to generate such db-files. The scripts use `gitpython` to roll back the repo, produce the desired files, and roll up to where we were again. The test suite will be set up to simply skip tests that cannot find their db-file on disk.

I apologise in advance that I also refactored the tests in this PR and thus touched 17 files. The new things are introduced in `qcodes/tests/dataset/legacy_DB_generation` and one test in `test_database_creation_and_upgrading.py`. Just looking at 541c50e is probably a good start to get an idea of what is going on. 

@QCoDeS/core


qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -380,7 +285,8 @@ def test_load_by_counter(dataset):
     assert exp.last_counter == 1
 
 
-def test_dataset_with_no_experiment_raises(empty_temp_db):
+@pytest.mark.usefixtures("empty_temp_db")
```
astafan8: do i understand correctly that by using `pytest.mark.usefixtures` we make `mypy` happy, while passing an imported fixture as an argument to `test_*` function makes it complain?

jenshnielsen: I think you can get rid of the unused import warnings from codacy/pylint by adding a ` # pylint: disable=unused-import`  comment to those lines

WilliamHPNielsen: 👍 you totes can. It's done in one place only in the PR right now because I'm a lazy dog, but I can add that everywhere so we can hear codacy again.

astafan8: ok, so unused imports over unused arguments :) (just a note: PyCharm since recent version does not complain about an unused argument if that argument is a pytest.fixture ;) )

jenshnielsen: Yes there are two issues.

* Unused imports that codacy complains about. We can either ignore those or mark up the lines 
* Arguments to functions that are seemingly never passed. Neither the decorator nor the name matching is super clean imho but I agree that the decorator is nicer

WilliamHPNielsen:
@QCoDeS/core, okay, ready for review.

jenshnielsen:
There is still something odd with this pr. Travis finished the tests but does never move on to the next stage in the build

WilliamHPNielsen:
Once Azure Pipelines finishes, I think we can merge this bad boy.

qcodes/tests/dataset/legacy_DB_generation/generate_version_0.py
```diff
@@ -0,0 +1,54 @@
...
+
+import os
+
+from git import Repo
+
+# NB: it's important that we do not import anything from qcodes before we
+# do the git magic (which we do below), hence the relative import here
+import utils as utils
+
+gitrepopath = os.sep.join(os.path.realpath(__file__).split(os.sep)[:-5])
```
jenshnielsen: Would it be useful to assert that this is actually a git repo e.g. check that there is a .git folder

jenshnielsen: Or perhaps that is already handled below by Repo() ?

qcodes/tests/dataset/test_database_creation_and_upgrading.py
```diff
@@ -0,0 +1,204 @@
...
+
+@contextmanager
+def location_and_station_set_to(location: int, work_station: int):
+    cfg = qc.Config()
+    old_cfg = deepcopy(cfg.current_config)
+    cfg['GUID_components']['location'] = location
+    cfg['GUID_components']['work_station'] = work_station
+    cfg.save_to_home()
+
+    yield
```
jenshnielsen: Should this have a try finally around the post actions too

qcodes/tests/dataset/legacy_DB_generation/generate_version_0.py
WilliamHPNielsen: As discussed offline, `Repo()` does that.

qcodes/tests/dataset/test_database_creation_and_upgrading.py
WilliamHPNielsen: Oh, shoot!

PR #1289, opened 2018-10-03
Ensure that post actions are performed in some context managers 

jenshnielsen:
even if an exception is thrown. 

The code after a yield statement in a context manager (the ``__exit__`` code) is not run if an exception is thrown unless wrapped in a try finally construct. In these cases I think it makes sense to always run the `__exit__` code @WilliamHPNielsen

### DIFF
--- a/qcodes/tests/dataset/test_dataset_basic.py
+++ b/qcodes/tests/dataset/test_dataset_basic.py
@@ -65,10 +65,11 @@ def location_and_station_set_to(location: int, work_station: int):
     cfg['GUID_components']['work_station'] = work_station
     cfg.save_to_home()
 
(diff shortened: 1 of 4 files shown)


astafan8:
Indeed, this is from the docstring:
```
"""@contextmanager decorator.

    Typical usage:

        @contextmanager
        def some_generator(<arguments>):
            <setup>
            try:
                yield <value>
            finally:
                <cleanup>

    This makes this:

        with some_generator(<arguments>) as <variable>:
            <body>

    equivalent to this:

        <setup>
        try:
            <variable> = <value>
            <body>
        finally:
            <cleanup>

    """
```

PR #1293, opened 2018-10-04
First attempt at publishing results from Azure pipelines

jenshnielsen:
Should hopefully give a better output


jenshnielsen:
One of the issues that I hit was https://github.com/Microsoft/vsts-tasks/issues/8300

jenshnielsen:
We apparently do not have permission to upload coverage so I am going to remove that for now

PR #1295, opened 2018-10-04
Azure pipelines add coverage

jenshnielsen:
This is WIP until we figure out how to set the right permissions. Taken from #1293


jenshnielsen:
@QCoDeS/core  Using the Azure app rather than then oath this works now See https://github.com/QCoDeS/Qcodes/pull/1295/checks

PR #1298, opened 2018-10-04
Pin testpath to fix conda/pip issue

jenshnielsen:
Discovered as part of #1297 should unbreak a conda install for now


jenshnielsen:
See https://github.com/conda-forge/testpath-feedstock/issues/7 and https://github.com/pypa/pip/issues/5839

It seems like a new version of test path was released on conda forge that contains invalid metadata which in turn makes pip give up and fail on any install

PR #1299, opened 2018-10-05
fix: N52xx various fixes

spauka:
A number of issues became apparent using the PNA driver, that are fixed in this pull request. There is a pending longer term rewrite that adds a number of new features (#1244) but in the mean time these are a couple of quick fixes for the existing driver.

- Trace type is extracted correctly. Previously, the trace type (S11/S21 etc.) was extracted from the trace name however this is not updated if the trace type is changed. The trace type is now extracted correctly from the full trace catalog.
- Added a trigger source parameter.
- Added trigger hints when a user aborts a sweep. Thanks @sohailc for this suggestion.
- Added `center`/`span` parameters.
- Add a link to run_sweep in the base driver.
- Fix average state detection. Previously this was not being correctly queried in `run_sweep`.

@sohailc @jenshnielsen


qcodes/instrument_drivers/Keysight/N52xx.py
```diff
@@ -377,12 +424,15 @@ def traces(self) -> ChannelList:
...
         """
+        active_trace = self.active_trace()
         parlist = self.ask("CALC:PAR:CAT:EXT?").strip('"').split(",")
         self._traces.clear()
-        for trace in parlist[::2]:
-            trnum = PNATrace.parse_paramstring(trace)[2]
-            pna_trace = PNATrace(self, "tr{}".format(trnum), int(trnum))
+        for trace_name in parlist[::2]:
+            self.write(f"CALC:PAR:SEL '{trace_name}'")
+            trace_num = int(self.ask("CALC:PAR:MNUM?"))
```
astafan8: A general comment: could you make separate methods out of `.ask(<CMD>)` and `.write(<CMD>)` in the instrument class instead of explicitly using `.ask()` and `.write()` with explicit VISA commands in the trace subclass? This will allow reuse of these commands, reduce code duplication, and potentially reduce possible errors.

qcodes/instrument_drivers/Keysight/N52xx.py
```diff
@@ -196,30 +218,34 @@ def write(self, cmd: str) -> None:
...
         """
         Select correct trace before querying
         """
-        self.root_instrument.active_trace(self.trace)
+        self.root_instrument.active_trace(self.trace_num)
         return super().ask(cmd)
 
-    @staticmethod
-    def parse_paramstring(paramspec: str) -> Tuple[str, str, str]:
+    def parse_paramstring(self) -> Tuple[str, str]:
```
astafan8: this does not seem to be `parse_...` anymore - it now also gets the `paramspec`. Could you keep the separation for clarity?

qcodes/instrument_drivers/Keysight/N52xx.py
```diff
@@ -196,30 +218,34 @@ def write(self, cmd: str) -> None:
...
-        ch, param, trnum = re.findall(r"CH(\d+)_(S\d+)_(\d+)", paramspec)[0]
-        return ch, param, trnum
+        specs = paramspec.split(',')
+        for spec_ind in range(len(specs)//2):
+            name, param = specs[spec_ind*2:(spec_ind+1)*2]
+            if name == self.trace_name:
+                return name, param
+        raise RuntimeError("Can't find selected trace on the PNA")
 
     def _Sparam(self, paramspec: str) -> str:
```
astafan8: now this method has an unused argument. Please, fix. Also, is there any use for other elements of `self.parse_paramstring()` (not only `[1]` that is used in this method)? And where are they documented?

qcodes/instrument_drivers/Keysight/N52xx.py
```diff
@@ -295,14 +321,28 @@ def __init__(self,
...
+                           get_parser=float,
+                           set_cmd='SENS:FREQ:CENT {}',
+                           unit='Hz',
+                           vals=Numbers(min_value=min_freq,max_value=max_freq))
+        self.add_parameter('span',
+                           label='Frequency Span',
+                           get_cmd='SENS:FREQ:SPAN?',
+                           get_parser=float,
+                           set_cmd='SENS:FREQ:SPAN {}',
+                           unit='Hz',
```
astafan8: neat! :)

qcodes/instrument_drivers/Keysight/N52xx.py
spauka: Agreed, most direct `write`/`ask` calls are now removed.

qcodes/instrument_drivers/Keysight/N52xx.py
spauka: Agreed, I've removed this function and folded the functionality into the `_Sparam` function.

qcodes/instrument_drivers/Keysight/N52xx.py
spauka: The parameter is now removed, as is the `[1]` array index. They were `(name, param)` pairs, so there is nothing more, but this is also clearer now.

spauka:
Any thoughts on this @astafan8 @jenshnielsen?

qcodes/instrument_drivers/Keysight/N52xx.py
```diff
@@ -377,18 +445,53 @@ def traces(self) -> ChannelList:
         """
         Update channel list with active traces and return the new list
         """
-        parlist = self.ask("CALC:PAR:CAT:EXT?").strip('"').split(",")
+        # Keep track of which trace was active before. This command may fail
+        # if no traces were selected.
+        try:
+            active_trace = self.active_trace()
+        except VisaIOError:
```
astafan8: Please check the contents of the exception message because if the reason for the exception is not that "no traces are selected", the driver might do smth wrong (or not do).

astafan8:
Great! waiting for CI and merge.

PR #1300, opened 2018-10-05
Generate fixture dbs on azure

jenshnielsen:
@WilliamHPNielsen

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,7 @@ script:
       pip install -e .
       cd qcodes/tests/dataset/legacy_DB_generation
       python generate_version_0.py
+      python generate_version_1.py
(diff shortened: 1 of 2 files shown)


jenshnielsen:
@WilliamHPNielsen it looks like test_perform_actual_upgrade_1_to_2 is always skipped even on travis?

WilliamHPNielsen:
@jenshnielsen, that makes sense, since `generate_version_1.py` must be run to generate a .db-file for that test to consume. Seems like I forgot to add that before merging #1277 

It would be good to add that here, also for Travis.

jenshnielsen:
@WilliamHPNielsen done and the tests still pass :)

PR #1302, opened 2018-10-08
[discussion] Off main process datasaving with ZMQ PUB-SUB

WilliamHPNielsen:
*DO NOT MERGE THIS*

It is usually helpful to look at some actual code when making plans. 

Here is a minimal working example of / proposal for a scheme where the `Measurement` object spawns a new process to do the actual data writing while pushing the data out via ZMQ.

I encourage you to look at this code and consider the following: is this a good approach for us, or should we follow a different pattern. If we can identify shortcomings of this pattern, we are already one step closer to a good solution. 

One thing that concerns me about this approach is that we could encounter situations were the write process is slower than the data acquisition. What would then happen if we did a large number of consecutive measurements? Perhaps we should wait for a signal back that the writing is complete?

@QCoDeS/core


astafan8:
Thanks for sketching it out. It definitely helps to be on the same page. This is exactly what I was imagining for a draft.

About the concern you raised - sure, there should be a way to tackle that problem, perhaps there is some other way ("protocol"?) for the sender/receiver built in to zmq that we can use? I really hope that we don't have to reinvent any wheels for this, just make a proper use of existing tools.

There are many questions that one might want to ask, and the answers are usually difficult to *quickly* capture in a "sketch". Nevertheless:
- what if writing raises an exception? presumably that has to be propagated to the master process, and have a decent result, right?
- what if the master process encounters an exception, what would the other guy do?
- is json format necessary? dumping and loading might take quite some time for large data arrays.
- a generalization is _already_ being defined here that the "writers" can be basically anything. Do we really want to have this as a requirement from the beginning? I would start by limiting the scope to the following two notions: both processes are python and use same kernel and same virtual environment.
- in the sketch, the zmq part is directly added to the Measurement. Wouldn't it be better if "the way data is communicated" is abstracted away? So that we can pass to the Measurement either `LocalSqliteDatabaseWriter` or `SeparateProcessSqliteDatabaseWriter` (and in the future, anything else that conforms to `add_result` and `flush`/`put_in_queue`/whatever interface)?

So, these and perhaps quite a few more questions need to be discussed and written down in text (because presumably making sketches of these takes too much time instead of describing them or pseudo-coding them).

alan-geller:
A process suggestion, rather than a code suggestion:

- I think supplementing this with a design document would be a good idea. It doesn't have to be formal, but a markdown document that describes the interfaces, primary classes, message formats, etc., would be good.
- We should do some perf testing with ZMQ to make sure it can keep up. We don't want to just swap one bottleneck for another.
- Before we finalize a design, we should make sure we've got an explicit list of the goals and requirements of this work. We've discussed it a bunch, so I think they're "in the air", but written down is better.

Dominik-Vogel:
I agree with the previously said in particular:
 - perf test
 - modlarize zmq out of `Measurement`
Additionally, at some later stage after the perf tests, I would like to discuss the structure of calling the data storing process on a per run basis.

PR #1303, opened 2018-10-09
Newest version of testpath should resolve the pip/conda issue

jenshnielsen:
So lets try to install it


astafan8:
but do we know if it does resolve the issue?

jenshnielsen:
@astafan8 The issue was that 0.4.1. broke Azure which this does not so yes

PR #1305, opened 2018-10-09
Plot_by_id rasterise 2d plots for large datasets

jenshnielsen:
Saving a vector representation of a 2d plot (svg, pdf) such as scatter or pcolormesh, is rather slow and generates a huge file if a lot of points are supplied.  The usefulness of vector graphics with that many points is also IMHO rather limited so the lose from rasterizing is not that big. The axes labels etc are still stored as vector graphics this only influences the data it self. The cutoff at 5000 points is somewhat arbitrary but keeps the time to save a pdf under 1 sec in my benchmarks


ThorvaldLarsen:
I have never had a need for vectorized colormesh. Rather it is annoying as opening large PDFs becomes quite slow (not sure how large 5000 point raster would be). So in my opinion it would be okay to always save rasterized as default.

jenshnielsen:
@ThorvaldLarsen At 5000 points this is still only 100 kB and opens in a matter of a few seconds so I think this is a workable switchover

qcodes/config/qcodesrc.json
```diff
@@ -15,6 +15,7 @@
     },
     "plotting":{
         "default_color_map": "viridis",
+        "rasterize_threshold": 5000,
```
Dominik-Vogel: Thumbs up for putting this here!

PR #1311, opened 2018-10-12
Feature: plot_by_id string-valued data for 2D plots

### LINKED ISSUE #1273: String-valued parameters support: 3D data case in plot_by_id
* Improve `plot_by_id` to use string-valued parameter values for tick labels for __3D plots__.
* Choose good plot types for 3D case (e.g. don't always use "scatter")
* Update example notebook for `plot_by_id` with 3D cases

This "issue" is a continuation of the development that started in https://github.com/QCoDeS/Qcodes/pull/1248.

astafan8:
This PR brings improvements to `plot_by_id` so that 2D plots of data with string-valued components are correct.

Changes:
* `plot_by_id` and related functions from `qcodes.dataset.plotting` and `qcodes.dataset.data_export`
* "offline plotting tutorial" example notebook
* new notebook dedicated to the plotting of string-valued data

The logic in `plot_by_id` is explained in the following table:
`< ??? >`

This PR is work-in-progress.

EDIT by William: fixes #1273

### DIFF
--- a/docs/examples/Creating Instrument Drivers.ipynb
+++ b/docs/examples/Creating Instrument Drivers.ipynb
@@ -2,10 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "source": [
     "# Creating QCoDeS instrument drivers"
    ]
@@ -13,11 +10,7 @@
   {
    "cell_type": "code",
    "execution_count": 1,
-   "metadata": {
-    "collapsed": true,
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "# most of the drivers only need a couple of these... moved all up here for clarity below\n",
@@ -34,16 +27,13 @@
   },
   {
    "cell_type": "markdown",
-   "metadata": {
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "source": [
     "## Base Classes\n",
     "\n",
     "There are 3 available:\n",
     "- `VisaInstrument` - for most instruments that communicate over a text channel (ethernet, GPIB, serial, USB...) that do not have a custom DLL or other driver to manage low-level commands.\n",
-    "- `IPInstrument` - a specialized driver just for ethernet connections. Otherwise works nearly identically to `VisaInstrument`.\n",
+    "- `IPInstrument` - a deprecated driver just for ethernet connections. Do not use this; use `VisaInstrument` instead.\n",
     "- `Instrument` - superclass of both `VisaInstrument` and `IPInstrument`, use this if you do not communicate over a text channel, for example:\n",
     "  - PCI cards with their own DLLs\n",
     "  - Instruments with only manual controls.\n",
@@ -81,10 +71,7 @@
   },
   {
    "cell_type": "markdown",
-   "metadata": {
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "source": [
     "## VisaInstrument: Simple example\n",
     "The Weinschel 8320 driver is about as basic a driver as you can get. It only defines one parameter, \"attenuation\". All the comments here are my additions to describe what's happening."
@@ -93,11 +80,7 @@
   {
    "cell_type": "code",
    "execution_count": 2,
-   "metadata": {
-    "collapsed": false,
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "class Weinschel_8320(VisaInstrument):\n",
@@ -143,10 +126,7 @@
   },
   {
    "cell_type": "markdown",
-   "metadata": {
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "source": [
     "## VisaInstrument: a more involved example\n",
     "\n",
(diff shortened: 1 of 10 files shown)


qcodes/dataset/measurements.py
```diff
@@ -32,8 +32,10 @@ class ParameterTypeError(Exception):
 
 def is_number(thing: Any) -> bool:
     """
-    Test if an object can be converted to a number
+    Test if an object can be converted to a number UNLESS it is a string
     """
+    if isinstance(thing, str):
+        return False
```
astafan8: what about the case of `nan`?

WilliamHPNielsen:
Now the string-valued plotting works, meaning that all combinations of categorical and numerical data are visualised correctly. There is, however, still an open issue regarding the _ordering_ of the categories when plotted. I am not sure what the best solution is, but I don't think we can do anything smarter than allowing a kwarg for ordering. Something like (see the example notebook for the corresponding data)
```python
plot_by_id(run_id, z_order=[useless', 'bad', 'possible', 'good'])
```

What do you guys think? 

And then perhaps we should add some more tests for this?

qcodes/dataset/data_export.py
```diff
@@ -347,24 +369,26 @@ def datatype_from_setpoints_2d(xpoints: np.ndarray,
...
     nx = len(xrow)
     ny = len(yrow)
 
-
     # potentially slow method of filling in the data, should be optimised
     log.debug('Sorting 2D data onto grid')
-    z_to_plot = np.full((ny, nx), np.nan)
+
+    if isinstance(z[0], str):
+        z_to_plot = np.full((ny, nx), ' '*int(str(z.dtype)[2:]))
```
astafan8: Please, add some comments to the `str(z.dtype)[2:]` part. Smth like:
```suggestion
        # strings in numpy have dtype names that look like '<U25';
        # `str(z.dtype)[2:]` extracts that "dtype length" (25 from above),
        # this is needed in order to allocate space for strings in numpy array:
        # once the strings will be assigned, the dtype of specific element will
        # automatically adjust to the string length in that array element
        z_to_plot = np.full((ny, nx), ' '*int(str(z.dtype)[2:]))
```

qcodes/dataset/measurements.py
astafan8: still, what about the case of nan?

qcodes/tests/dataset/test_plot_by_id.py
```diff
@@ -0,0 +1,72 @@
+import numpy as np
```
astafan8: feel free to remove this file because there is the great notebook now.

astafan8:
Regarding the order of string values, it is impossible to deduct what's a sane order without doing some heuristics on the values. I do not think that `plot_by_id` shall do such smart stuff.

astafan8:
now that there is a separate string-valued plotting tutorial, shall it be referenced in the main offline plotting tutorial? moreover, shall the string-valued examples from the main plotting tutorial be removed?

astafan8:
@WilliamHPNielsen thanks for the contribution! Especially the example notebook is great!

I cannot approve this PR since i'm the initiator, so you or @QCoDeS/core need to approve it.

WilliamHPNielsen:
@astafan8, thanks for the review. A question and a comment:

* Ordering: what do you think about about having `plot_by_id` accept the ordering as a kwarg? That is does not imply any smartness from `plot_by_id`. (this is a question for all of @QCoDeS/core)
* Notebooks: yes, one plotting notebook should reference the other. I agree to remove most (if not all) of the categorical plotting from the main plotting tutorial.

I'll look into your inline comments now.

qcodes/dataset/measurements.py
WilliamHPNielsen: Isn't it okay? Do you have anything particular in mind? 

I think it's good that we *don't* have magical strings that are not allowed amongst other strings. Imagine a measurement of some type casting speed benchmark. I have categories `['int', 'float', 'nan', 'none', 'null', 'np.int32']`. I'd really like that to just plot without getting an SQLite error that my `nan` (now no longer `"nan"`) can not be inserted in a text-valued column.

astafan8:
@WilliamHPNielsen ordering: kwarg? kwarg of values in a desired order? don't know why, but it does not sound really right to me. In this case, I'd suggest to wait for our users to tell us that such a feature is needed ;)

qcodes/dataset/measurements.py
astafan8: `"nan"` is not a magical string. The thing is that `float('nan')` returns `nan` (`type(float('nan'))` is `float`). Hence when you are doing `isinstance(thing, str)` before `float(thing)` you are neglecting the case where a string`"nan"` means `nan` floating value. Have a look at converter/adapter functions in sqlite_base - that `"nan"` case is covered there as well.

(additional, not really useful comment: exceptions are cheap in python, so it could be the `try: float(thing)` in case `thing` is a string is faster than `isinstance(thing, str)` :) )

astafan8: and why is this change actually needed?

jenshnielsen:
What is the current behavior? The order that the columns are in the database? I think it's ok to defer anything smarter till we have some real use cases

Dominik-Vogel:
Bearing in mind that this probably a question of taste: I cannot see any simple alternative to having the keyword for the ordering. This solution seems a little inconvenient to me, but it enables the user to get things done as they want it to be.
I think the right solution would be basing off a lot lower level. I think there should be a data type for categorical data. When registering the Paramspec the allowed values should be defined and thereby also their order.

PR #1312, opened 2018-10-12
Make Array validators shape optionally callable

jenshnielsen:
This means that you can use another parameter to define the shape of a parameter with an array validator. This is part of doing #1263 with some more checks for consistency. 

I also did a bit of cleanup on the validators

* Did away with the super init method that cannot be called which is unusual and really doesn't add anything useful. 
* Added typehints to all validators 
* Improved a few docstrings

### DIFF
--- a/qcodes/utils/validators.py
+++ b/qcodes/utils/validators.py
@@ -88,10 +88,7 @@ class Validator:
 
     @property
     def valid_values(self) -> TList[Any]:
-        if hasattr(self, '_valid_values'):
-            return self._valid_values
-        else:
-            raise NotImplementedError
+        return self._valid_values
 
 
 class Anything(Validator):
--- a/qcodes/utils/validators.py
+++ b/qcodes/utils/validators.py
@@ -1,13 +1,19 @@
 import math
-from typing import Union, Optional, Tuple
+from typing import Union, Optional, Tuple, Any, Hashable
+# rename on import since this file implements its own classes
+# with these names.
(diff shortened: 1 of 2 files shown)


qcodes/utils/validators.py
```diff
@@ -58,51 +67,54 @@ class Validator:
...
         raise NotImplementedError
 
     @property
-    def valid_values(self):
-        if not hasattr(self, '_valid_values'):
-            raise NotImplementedError
-        else:
+    def valid_values(self) -> TList[Any]:
+        if hasattr(self, '_valid_values'):
             return self._valid_values
```
astafan8: what is the point of this clause if `_valid_values` exists in the "base" class?

qcodes/utils/validators.py
```diff
@@ -499,14 +516,30 @@ def __init__(self, min_value: Union[float, int]=-float("inf"),
...
+    def shape(self) -> Optional[Tuple[int, ...]]:
+        if self._shape is None:
+            return None
+        shape_array = []
+        for s in self._shape:
+            if callable(s):
+                shape_array.append(s())
+            else:
+                shape_array.append(s)
+        shape = tuple(shape_array)
```
astafan8: nice!

qcodes/utils/validators.py
jenshnielsen: I thin you are right. This was needed before the last refactor but is no longer needed.

PR #1313, opened 2018-10-12
Publish docs as a build artifact

jenshnielsen:
If done correctly this will hopefully make the html docs available on the azure builds tap


jenshnielsen:
This does not work for the same reason as #1295

jenshnielsen:
This seems to be working now. A zip of the docs can be downloaded from https://ms-quantum.visualstudio.com/QubitWorkbench/_build/results?buildId=9891&view=results

WilliamHPNielsen:
This is really nice! One small thing, though: could we call the artifact `qcodes_docs` or something slightly more specific than simply "docs"?

PR #1318, opened 2018-10-15
Signal hound a few more changes

jenshnielsen:
While trying to resolve a usb comm error for @nataliejpg I cleaned up the driver a bit

* Add types to ctypes dll calls matching the api defined in Signal Hounds headers. The only real changes are a few uint -> int that should have no real consequence as we are dealing with small positive numbers 
* Rewrite `_get_sweep_data` to capture data directly into a numpy array saving a loop over the data
* refactor the averageing logic into `_get_sweep_data` removing the avg function. This saves a few redundant dll calls
* remove the retry functionality from `_get_sweep_data`. The api docs are quite clear that there is no point in retrying when the communication is lost. The only way forward is to close the device. All other errors relate to invalid input and will not be fixed by retrying


qcodes/instrument_drivers/signal_hound/USB_SA124B.py
```diff
@@ -178,6 +178,40 @@ def __init__(self, name, dll_path=None, **kwargs):
         self._trace_updated = False
         log.info('Initializing instrument SignalHound USB 124B')
         self.dll = ct.CDLL(dll_path or self.dll_path)
+
+        self.dll.saConfigCenterSpan.argtypes = [ct.c_int,
```
astafan8: does it make sense to extract these into a separate class (or smth) so that they don't pollute the driver constructor?

qcodes/instrument_drivers/signal_hound/USB_SA124B.py
```diff
@@ -549,45 +583,33 @@ def _get_sweep_data(self) -> np.ndarray:
         This function performs a sweep over the configured ranges.
         The result of the sweep is returned along with the sweep points
 
+        Args:
```
astafan8: where is this arg in the signature? (also there is a typo in the description)

jenshnielsen: Removed in the last commit, had the change locally but forgot to commit

qcodes/instrument_drivers/signal_hound/USB_SA124B.py
jenshnielsen: Done

jenshnielsen:
@nataliejpg confirms that this is working

PR #1320, opened 2018-10-15
changes imports of abstract base class collections

### LINKED ISSUE #1319: Using qcodes with Python >=3.7 throws deprecation warnings
When running Qcodes with python >=3.7.0 warnings about deprecation will pop up related to the collections abstract base classes that were moved to `collections.abc` since Python 3.3

Note: Some are related to external libs (h5py, pyqt) and need to be fixed upstream in the respective projects. But I can make a pull request that fixes the Qcodes part.

### Steps to reproduce
  1. Run Unit tests  with Python >=3.7.0 `pytest /Users/stefan/PycharmProjects/Qcodes/qcodes/tests/`
  2. at several places deprecation warnings related to some of the classes in the `collections.abc` module pop up

  e.g. 
  ```
  Qcodes/qcodes/dataset/experiment_container.py:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import Sized
  ```
see attached [DeprecationWarnings.txt](https://github.com/QCoDeS/Qcodes/files/2479312/DeprecationWarnings.txt) for full test log

### System
**operating system**
OSX Mojave (10.14)
Python 3.7.0

**qcodes branch**
Master

**qcodes commit**
58eff051

StefanD986:
Fixes #1319 .

- importing the abstract collections (like Sequences, Iterable, ..) directly from the `collections` module is deprecated. This patch changes the imports to`import from `collections.abc`.
- note that the fix as it is might break things for users who use python <3.3. 

@astafan8

### DIFF
--- a/qcodes/config/config.py
+++ b/qcodes/config/config.py
@@ -444,7 +444,7 @@ class DotDict(dict):
 
 def update(d, u):
     for k, v in u.items():
-        if isinstance(v, collections.Mapping):
+        if isinstance(v, collections.abc.Mapping):
             r = update(d.get(k, {}), v)
             d[k] = r
         else:
--- a/qcodes/data/data_array.py
+++ b/qcodes/data/data_array.py
@@ -268,7 +268,7 @@ class DataArray(DelegateAttributes):
         """
         if data is not None:
             if not isinstance(data, np.ndarray):
-                if isinstance(data, collections.Iterator):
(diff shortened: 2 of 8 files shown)


jenshnielsen:
Yes we are using f-strings and type annotations so we only support 3.6+

PR #1321, opened 2018-10-16
Test with python 3.7 on travis

jenshnielsen:
Should hopefully make it simpler to catch issues such as #1319

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: python
+sudo: required
+dist: xenial
 notifications:
   email: false
     # send all notifcations to slack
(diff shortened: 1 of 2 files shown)


jenshnielsen:
So it turns out that this needs Travis's new Ubuntu 16.04 images in sudo true (vm and not container) mode. I don't think this is an issue. 

The changes to `sudo: true` should be unimportant as Travis is remove the sudo false images anyway https://blog.travis-ci.com/2018-10-04-combining-linux-infrastructures

Testing on a newer linux install is also probably an improvement regardless

jenshnielsen:
This raises a number of new warnings of the type in #1319 but only one of them is coming from qcodes. The rest are from dependencies and will hopefully be fixed there.

PR #1322, opened 2018-10-16
Feature: access snapshot of a DataSet, and docs

astafan8:
What's in:
* add `DataSet.snapshot` and `DataSet.snapshot_raw` (taken from #1262 PR) properties to retrieve snapshot dict and JSON
* add "Working with snapshots" notebook that explains what snapshot is and how to obtain it (original notebook about snapshots has a note that refers to the new notebook)
* tests

Some of the code from this PR was already in #1215 PR, but that one stalled for a bit in WIP mode, hence this new PR.

### DIFF
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -7,6 +7,7 @@
 # Distributed under terms of the MIT license.
 # import json
 import functools
+import json
 from typing import Any, Dict, List, Optional, Union, Sized, Callable
 from threading import Thread
 import time
@@ -240,7 +241,16 @@ class DataSet(Sized):
 
     @property
     def snapshot(self):
-        """Returns snapshot of the run in JSON format (or None)"""
+        """Snapshot of the run as dictionary (or None)"""
+        snapshot_json = self.snapshot_raw
+        if snapshot_json is not None:
+            return json.loads(snapshot_json)
+        else:
+            return None
+
+    @property
+    def snapshot_raw(self):
+        """Snapshot of the run in JSON format (or None)"""
         if column_in_table(self.conn, "runs", "snapshot"):
             return select_one_where(self.conn, "runs", "snapshot",
                                     "run_id", self.run_id)
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -21,6 +21,7 @@ from qcodes.instrument.parameter import _BaseParameter
 from qcodes.dataset.sqlite_base import (atomic, atomic_transaction,
                                         transaction, add_parameter,
                                         connect, create_run, completed,
+                                        column_in_table,
                                         get_parameters,
                                         get_experiments,
                                         get_last_experiment, select_one_where,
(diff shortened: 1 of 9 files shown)


qcodes/dataset/sqlite_base.py
```diff
@@ -485,6 +485,26 @@ def init_db(conn: SomeConnection)->None:
         transaction(conn, _layout_table_schema)
         transaction(conn, _dependencies_table_schema)
 
+
+def column_in_table(conn: SomeConnection, table: str, column: str) -> bool:
```
jenshnielsen: Would this be better named `is_column_in_table` ?

qcodes/dataset/data_set.py
```diff
@@ -262,6 +264,24 @@ def table_name(self):
     def guid(self):
         return get_guid_from_run_id(self.conn, self.run_id)
 
+    @property
+    def snapshot(self):
```
jenshnielsen: Can we add minimal types (Optional[dict] and Optional[str]) to the return of these 2 functions.

qcodes/dataset/data_set.py
```diff
@@ -262,6 +264,24 @@ def table_name(self):
...
+        """Snapshot of the run as dictionary (or None)"""
+        snapshot_json = self.snapshot_raw
+        if snapshot_json is not None:
+            return json.loads(snapshot_json)
+        else:
+            return None
+
+    @property
+    def snapshot_raw(self):
+        """Snapshot of the run in JSON format (or None)"""
```
jenshnielsen: Snapshot of the run as a string in JSON format (or None) ?

qcodes/tests/test_station.py
```diff
@@ -0,0 +1,203 @@
...
+
+def test_station():
+    bob = DummyInstrument('bob', gates=['one'])
+    station = Station(bob)
+
+    assert ['bob'] == list(station.components.keys())
+    assert bob == station.components['bob']
+
+    assert station == station.default
+    assert station == Station.default
```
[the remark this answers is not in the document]
astafan8: first is checking the instance's `default`, the second checks the class's `default`, to confirm that they are the same. it's really just a "just in case" assertion :)

qcodes/dataset/sqlite_base.py
astafan8: sure, that's better!

qcodes/dataset/data_set.py
astafan8: sure!

qcodes/dataset/data_set.py
astafan8: yes. the `snapshot` property return python dict.

or is the formulation confusing or not good english?

jenshnielsen: It's probably very minor and perhaps irrelevant now that the function has a type but I would write `a string in JSON format` to make it clear what the type being returned is

qcodes/dataset/data_set.py
```diff
@@ -265,7 +265,7 @@ def guid(self):
         return get_guid_from_run_id(self.conn, self.run_id)
 
     @property
-    def snapshot(self):
+    def snapshot(self) -> Optional[str]:
```
jenshnielsen: Does this not return a dict?

PR #1327, opened 2018-10-19
Added driver for Basel current preamplifier.

emartinez:
This pull request implements a virtual driver for the Basel preamplifier.

@jenshnielsen


jenshnielsen:
This looks fine as a simpel virtual driver but I wonder if this works with the remote interface https://www.physik.unibas.ch/department/infrastructure-services/electronics-lab/lnhs-i-to-v-converter-remote-control-interface-sp-983a.html that can be controlled from a computer? Is it worth adding support for that too

jenshnielsen:
@emartinez I tried updating the docstrings to explain the diffs between the 2 versions of the device and that we do not use the remote control here. Could you have a look and see if that makes sense to you? Thanks

PR #1328, opened 2018-10-19
Meaningful exception when loading/creating dataset for non-existing run id

astafan8:
Currently when one uses `load_by_id` for non-existing `run_id`, the exception looks like `"Expected one row"`. This PR fixes it with:
* add `run_exists` method to `sqlite_base`
* call `run_exists` in `DataSet` constructor and raise exception with clear message if the `run_id` does not exist
* tests

Note that the used query seems to be quite fast according to [this stackoverflow discussion](https://stackoverflow.com/questions/1676551/best-way-to-test-if-a-row-exists-in-a-mysql-table).

### DIFF
--- a/qcodes/tests/dataset/test_dataset_basic.py
+++ b/qcodes/tests/dataset/test_dataset_basic.py
@@ -48,7 +48,7 @@ def test_has_attributes_after_init():
 
 
 @pytest.mark.usefixtures("experiment")
-@pytest.mark.parametrize("non_existing_run_id", (999999, 'number#42'))
+@pytest.mark.parametrize("non_existing_run_id", (1, 'number#42'))
 def test_create_dataset_from_non_existing_run_id(non_existing_run_id):
     with pytest.raises(ValueError, match=f"Run with run_id "
                                          f"{non_existing_run_id} does not "
--- a/qcodes/tests/dataset/test_dataset_basic.py
+++ b/qcodes/tests/dataset/test_dataset_basic.py
@@ -47,6 +47,15 @@ def test_has_attributes_after_init():
         getattr(ds, attr)
 
 
+@pytest.mark.usefixtures("experiment")
+@pytest.mark.parametrize("non_existing_run_id", (999999, 'number#42'))
+def test_create_dataset_from_non_existing_run_id(non_existing_run_id):
+    with pytest.raises(ValueError, match=f"Run with run_id "
+                                         f"{non_existing_run_id} does not "
+                                         f"exist in the database"):
+        _ = DataSet(run_id=non_existing_run_id)
+
+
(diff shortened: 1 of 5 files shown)


qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -47,6 +47,15 @@ def test_has_attributes_after_init():
         getattr(ds, attr)
 
 
+@pytest.mark.usefixtures("experiment")
+@pytest.mark.parametrize("non_existing_run_id", (999999, 'number#42'))
```
WilliamHPNielsen: When using the `experiment` fixture, you get back `e` where `e = new_experiment("test-experiment", sample_name="test-sample")`. Therefore, `1` should also be a non-existing run_id. That would in some way be a good check of the fixture to use `1` instead of / in addition to `999999` and a string.

qcodes/tests/dataset/test_dataset_loading.py
```diff
@@ -29,6 +29,14 @@ def test_load_by_id():
     assert loaded_ds.completed is False
     assert loaded_ds.exp_id == 1
 
+    # let's take a reasonably arbitrary run number that is quite definitely not
```
WilliamHPNielsen: I think this comment is confusing. Perhaps simply writing "let's take a run number that is not" is better?

qcodes/tests/dataset/test_dataset_loading.py
```diff
@@ -29,6 +29,14 @@ def test_load_by_id():
...
+    # in the temporary test database file
+    non_existing_run_id = run_id + 9999
```
WilliamHPNielsen: Shouldn't `run_id + 1` also do the trick here?

qcodes/tests/dataset/test_sqlite_base.py
```diff
@@ -121,3 +121,8 @@ def test_get_dependents(experiment):
 def test_column_in_table(dataset):
     assert mut.is_column_in_table(dataset.conn, "runs", "run_id")
     assert not mut.is_column_in_table(dataset.conn, "runs", "non-existing-column")
+
+
+def test_run_exist(dataset):
+    assert mut.run_exists(dataset.conn, dataset.run_id)
+    assert not mut.run_exists(dataset.conn, dataset.run_id + 9999)
```
WilliamHPNielsen: Again, I think that `9999` could be `1`. It's not a big deal, but `9999` looks a bit (a little bit) like a magic number.

qcodes/tests/dataset/test_dataset_loading.py
astafan8: sure! and i changed the number to `1` as well.

PR #1329, opened 2018-10-22
Fix JSON-dumping of snapshots with numpy-typed values (and other)

astafan8:
### Example situation
If an instrument parameter is set to a value of `numpy.int32` type, JSON dumping of the snapshot will fail with an exception that "int32 type cannot be serialized to JSON". It's easy to get into such a situation, for example, start an experiment after code like this has been executed: `... for x in np.arange(10, 20, 2): my_instrument.power(x) ...` (explanation: the code leaves the `power` parameter with the value of type `numpy.int32`, so if you start a new measurement afterwards, JSON-ifying the snapshot fails). Note however that using floats in that `arange` call removes the problem: `... np.arange(10.0, 20.0, 2)....`. This is has to do with the interesting fact that `json.dumps(numpy.float64(1))` works while `json.dumps(numpy.float32(1))` or `json.dumps(numpy.float16(1))` (or other) raises `"Object of type <type> is not JSON serializable"`. It seems to be a [python 3 bug/problem, read more here (especially last comments)](https://bugs.python.org/issue24313).

### Reason
`Runner` uses bare `json.dumps` call to serialize the station snapshot. The python built-in `json` does not support `numpy` types (and other stuff as well).

### Solution
`qcodes.utils.helpers` contains `NumpyJSONEncoder` class that allows serialize many things, including `numpy` types. Moreover, it is used for the old qcodes dataset for writing metadata, including station snapshots. Hence, let's re-use it. 

The name of the encoder class is bad, but it is kept for the reasons of not having to spend time on refactoring.

Note that this solution does have a small performance impact (dumping is ~40% faster without the custom encoder class), but it's negligible with respect to the benefits the solution brings. Moreover, snapshot is taken only once before the measurement, and generally takes less than a ms (modulo interaction with actual instruments), so it's really unnoticeable.

### Contents of this PR
* Improve `NumpyJSONEncoder` docstring
* Make `NumpyJSONEncoder` support all `numpy` scalar types via using `numpy.generic` in its `isinstance` call
* Add more tests for `NumpyJSONEncoder` class to capture all of its behavior
* Use `NumpyJSONEncoder` class in `Runner` for converting station snapshot into JSON
* Add a test for `Runner`/`Measurement` for the case where snapshot contains `numpy`-typed values

### DIFF
--- a/qcodes/utils/helpers.py
+++ b/qcodes/utils/helpers.py
@@ -5,7 +5,6 @@ import math
 import numbers
 import time
 import os
-
 from collections.abc import Iterator, Sequence, Mapping
 from copy import deepcopy
 from typing import Dict, List, Any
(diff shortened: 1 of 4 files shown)


qcodes/utils/helpers.py
```diff
@@ -19,19 +18,41 @@
...
 
+
 class NumpyJSONEncoder(json.JSONEncoder):
-    """Return numpy types as standard types."""
-    # http://stackoverflow.com/questions/27050108/convert-numpy-type-to-python
-    # http://stackoverflow.com/questions/9452775/converting-numpy-dtypes-to-native-python-types/11389998#11389998
+    """
+    This JSON encoder adds support for serializing types that the built-in
+    `json` module does not support out-of-the-box. See the docstring of the
+    `default` method for the description of all conversions.
```
jenshnielsen: For future reference I think we should add a link to the cpython issue that you referenced. There may come a day where this is fixed upstream

astafan8: resolving that might not help with supporting __all__ the numpy types though. so, i'm not sure that linking adds value.

jenshnielsen: I still think that it's not intuitive why we need to convert numbers and documenting that is likely to save some time but it should probably be documented along with that specific part i.e. just after https://docs.scipy.org/doc/numpy/reference/arrays.scalars.html

astafan8: agreed offline to add more information to the PR's description.

PR #1330, opened 2018-10-23
Build and execute notebooks with nbsphinx

jenshnielsen:
This tries to simplify the execution of notebooks as part of the build process. This seems to solve
the issue building notebooks on windows and simplifies the logic. Currently all notebooks are
executed unless metadata is added to mark them not executable. The execution can be customized on a per notebook basis to allow longer timeout or cells raising errors. We make use of this in a few examples

Execution of notebooks can be skipped completely by setting an env variable for speed
 
In the process I fixed some broken code in a few notebooks and some broken links.

It is now possible to build the docs in two ways from both windows (with cmd/bat) files and from 
osx/linux with Makefiles.

The target htmlapi will execute all notebooks that can be executed. 
The target htmlfast skips the notebook execution and is better suited for rapid iterations 

TODO:

- [x] Document how this is done. Especially that you need to disable execution of driver notebooks and the way to skip the notebook execution. We should perhaps even have a make target that does this. 
- [x] There still seems to be one failure that I have not tracked down yet


jenshnielsen:
@QCoDeS/core  I think this is ready for review

jenshnielsen:
This is likely to conflict with #1309 so we should decide which one goes in first

Dominik-Vogel:
I will take a look at this PR and will hopefully learn enough to be able to merge it into #1309. So this one can go first.

docs/conf.py
```diff
@@ -351,10 +351,10 @@
 
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
-    'matplotlib': ('http://matplotlib.org/', None),
+    'matplotlib': ('https://matplotlib.org/', None),
     'python': ('https://docs.python.org/3.5', None),
```
astafan8: shouldn't it be `3.6`?

docs/Makefile
```diff
@@ -229,20 +240,16 @@ dummy:
...
 	@echo "Build finished. Dummy builder generates no files."
 
+# this is only left for backwards compatibility. Changed to be consistent with windows
 .PHONY: html-api
-html-api:
+html-api: htmlapi
+
+# generate api docs for instruments automatically
+.PHONY: genapis
+genapis:
```
astafan8: shall this be documented in the `help` above as well or is it not supposed to be called "alone"?

azure-pipelines.yml
```diff
@@ -39,6 +39,7 @@ jobs:
         displayName: "Pytest"
       - script: |
           cd docs
+          set SPHINXOPTS=-W
```
Dominik-Vogel:
```suggestion
          set SPHINXOPTS=-W #Turn warnings into errors
```

docs/examples/DataSet/Working with snapshots.ipynb
```diff
@@ -620,9 +619,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python [conda env:qcodes]",
+   "display_name": "Python 3",
    "language": "python",
-   "name": "conda-env-qcodes-py"
+   "name": "python3"
```
astafan8: i am no sure where to look for it, but is the kernelspec now still ignored or not? when the notebooks are executed, what kernel is used?

docs_requirements.txt
```diff
@@ -1,7 +1,8 @@
 sphinx==1.7.9
 sphinx_rtd_theme
 sphinxcontrib-jsonschema
-nbconvert
+nbsphinx
```
astafan8: so just adding this makes sphinx build rst docs from the notebooks that it finds? :) super cool)

docs/Makefile
```diff
@@ -229,20 +240,16 @@ dummy:
 	@echo
 	@echo "Build finished. Dummy builder generates no files."
 
+# this is only left for backwards compatibility. Changed to be consistent with windows
 .PHONY: html-api
-html-api:
+html-api: htmlapi
```
Dominik-Vogel: This is super confusing to have the target `html-api` that depends on `htmlapi`.

astafan8: well, backwards compatibility :)

Dominik-Vogel: OK now I get it. Its for compatibillity. good job!

docs_requirements.txt
jenshnielsen: And loading the plugin from the conf file but otherwise yes

docs/Makefile
jenshnielsen: Exactly I don't think this is useful to be called alone but we may want to add it to other targets if we want to generate the docs in another format

docs/examples/DataSet/Working with snapshots.ipynb
astafan8: perfect! let's use the solution.

for `jupyter-nbconvert` we used to pass `--ExecutePreprocessor.kernel_name=python3` so that the kernel mentioned in the notebooks is ignored.

note that we will never be able to get rid of the various kernelspecs in the notebooks, because when we edit them, we run them on our own PCs and setups, hence we select a particular kernel for that, and if a developer explicitly does not change it back to `python 3` or whatever, his kernelspec gets saved.

astafan8:
@jenshnielsen ok, let's say that for now the info that you wrote in the driver docs is enough.

PR #1333, opened 2018-10-23
driver / SR830 : add the snap XY command to read X and Y together

GeneralSarsby:
The snap command records the values of X and Y at a single instant. This is a way to query values at the same time. This is important when the time constant is very short, < 100 ms.
 -> tuple (x,y)

- add a single parameter, SNAP_XY


GeneralSarsby:
Hi astafan8,

I've put the changes you recommended. It is now a function and in snake_case.

the "X,Y" is not the only combination, but it the most important of the snap commands.

In brief:
The snap command can return up to six different values at the same time.
This is up to 6, in any order, from 11 different choice values that the lockin can measure, including the phase, Aux1, freq, etc.

This single snap_xy function I think covers the 99% of the use cases. 
However I can see that a more general function would be something like:
lockin.snap('x','y','freq','aux1') -> (x,y,freq,aux1)

What do you think? make it generalized? or simple?

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -386,7 +386,17 @@ def parse_offset_get(s):
                            get_cmd='OUTP? 4',
                            get_parser=float,
                            unit='deg')
-
+        
+        self.add_function('snap_xy',
```
astafan8: I meant class bound method, like `def snap_xy(...): ....`. Sorry that I was not clear enough.

astafan8:
@GeneralSarsby Yes, indeed, since you are implementing this great and useful feature, please, implement it in a flexible way. You can refer to [this implementation in SR860](https://github.com/QCoDeS/Qcodes/blob/a6d4c22a701b8b468342580fa15d40cfbe1d7876/qcodes/instrument_drivers/stanford_research/SR86x.py#L1022) as a reference.

Also, users will be very thankful if you also update the example notebook for SR830 with information on this feature (see the example notebook for SR860 for an example).

I'm sure that there will be users who will fall into the 1% of the cases that `snap_xy` does not cover, and they will be very happy to know that there is a general `snap` (or whatever well-named) method that allows them to utilize the full power of SR830 and return up to 6 __synchronized__ values at once out of 11 available.

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -472,6 +472,60 @@ def parse_offset_get(s):
...
+    def snap(self, *parameter_names: str) -> Tuple[float, ...]:
+            """
+            Gets the values of either 2, 3, 4, 5 or 6 parameters at a single instant.
+            For example, SNAP? is a way to query values of
+            X and Y (or R and θ) which are taken at the same time. This is important
+            when the time constant is very short. Using the OUTP? or OUTR? com-
+            mands will result in time delays, which may be greater than the time con-
+            stant, between reading X and Y (or R and θ)
+            Thus reading  X,Y  OR  R,θ yields  a  coherent  snapshot  of  the  output  signal.
+
```
astafan8: it's great that this corresponds to the description in the manual. However, i think it is also important that the following is also mentioned here:
```
The values of X and Y are recorded at a single instant. The values of R
and θ are also recorded at a single instant. Thus reading X,Y OR R,θ
yields a coherent snapshot of the output signal. If X,Y,R and θ are all
read, then the values of X,Y are recorded approximately 10µs apart from
R,θ. Thus, the values of X and Y may not yield the exact values of R and
θ from a single SNAP? query.
The values of the Aux Inputs may have an uncertainty of up to 32µs. The
frequency is computed only every other period or 40 ms, whichever is
longer. 
The SNAP? command is a query only command. The SNAP? command
is used to
```

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -472,6 +472,60 @@ def parse_offset_get(s):
...
+        'aux 4': '8',  
+         'freq': '9',   
+         'ch 1': '10',
+         'ch 2': '11'  
+    }
+    
+    def snap(self, *parameter_names: str) -> Tuple[float, ...]:
+            """
+            Gets the values of either 2, 3, 4, 5 or 6 parameters at a single instant.
+            For example, SNAP? is a way to query values of
```
astafan8: Please, substitute the notions of exact visa commands which they parameters/methods from the driver. We developer drivers exactly for the reason of hiding all of this visa-level stuff.

astafan8: "for example"? this does not read nicely ((

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -472,6 +472,60 @@ def parse_offset_get(s):
...
+            X and Y (or R and θ) which are taken at the same time. This is important
+            when the time constant is very short. Using the OUTP? or OUTR? com-
+            mands will result in time delays, which may be greater than the time con-
+            stant, between reading X and Y (or R and θ)
+            Thus reading  X,Y  OR  R,θ yields  a  coherent  snapshot  of  the  output  signal.
+            
+            Args:
+                *parameter_names
+                    2 or 3 names of parameters for which the values are
```
astafan8: 2 or 3? it should be "from 2 up to 6", right?

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -472,6 +472,60 @@ def parse_offset_get(s):
...
+                    requested; valid names can be found in `PARAMETER_NAMES`
```
astafan8: typo: `SNAP_PARAMETERS`?

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -472,6 +472,60 @@ def parse_offset_get(s):
...
+            'p': '4',   
+        'phase': '4',  
+           'θ' : '4',
+        'aux 1': '5',  
+        'aux 2': '6',   
+        'aux 3': '7',  
+        'aux 4': '8',  
+         'freq': '9',   
+         'ch 1': '10',
+         'ch 2': '11'
```
astafan8: as a user, can you confirm that a parameter with a "space" is convenient? I, personally, thought that the most convenient for the users is without the space, like `ch1` instead of `ch 1`.

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -472,6 +472,60 @@ def parse_offset_get(s):
...
+         'ch 1': '10',
+         'ch 2': '11'  
+    }
+    
+    def snap(self, *parameter_names: str) -> Tuple[float, ...]:
+            """
+            Gets the values of either 2, 3, 4, 5 or 6 parameters at a single instant.
+            For example, SNAP? is a way to query values of
+            X and Y (or R and θ) which are taken at the same time. This is important
+            when the time constant is very short. Using the OUTP? or OUTR? com-
```
astafan8: within qcodes, we are generally following pep8, which requires the lines of code to not exceed 80 characters. Could you please adjust the code accordingly?

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -472,6 +472,60 @@ def parse_offset_get(s):
...
+                    'It is only possible to request values of 2 to 6 parameters '
+                    'at a time.')
+
+            for name in parameter_names:
+                if name.lower() not in self.SNAP_PARAMETERS:
+                    raise KeyError(f'{name} is not a valid parameter name. Refer '
+                                   f'to `SNAP_PARAMETERS` for a list of valid '
+                                   f'parameter names')
+
+            p_ids = [self.PARAMETER_NAMES[name.lower()] for name in parameter_names]
```
astafan8: typo: perhaps `SNAP_PARAMETERS`?

(have this change been tested after all? :) )

qcodes/instrument_drivers/stanford_research/SR830.py
GeneralSarsby: I will not fight ether way for this. without a space is ok too.
My reasoning way that inside a string is not code, so it should follow English prose rules.
But if without spaces is a better style fit, that's fine.

qcodes/instrument_drivers/stanford_research/SR830.py
GeneralSarsby: Yep. This can be done.

qcodes/instrument_drivers/stanford_research/SR830.py
GeneralSarsby: I've tried to re-write the whole docstring to make things clearer.  Both with what the function does and with the limitations. I think this also covers a previous comment. We can develop this further if needed.

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -472,6 +472,76 @@ def parse_offset_get(s):
...
+         'aux3': '7',  
+         'aux4': '8',  
+         'freq': '9',   
+          'ch1': '10',
+          'ch2': '11'  
+    }
+    
+    def snap(self, *parameters: str) -> Tuple[float, ...]:
+        """
+Get between 2 and 6 parameters at a single instant. This provides a coherent
```
astafan8: so far as i know the text for the docstring shall start from `"""`, like this:
```
    def smth():
        """
        a great docstring
        ...
```

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -472,6 +472,76 @@ def parse_offset_get(s):
...
+         'aux1': '5',  
+         'aux2': '6',   
+         'aux3': '7',  
+         'aux4': '8',  
+         'freq': '9',   
+          'ch1': '10',
+          'ch2': '11'  
+    }
+    
+    def snap(self, *parameters: str) -> Tuple[float, ...]:
```
WilliamHPNielsen: I think `Tuple` was never imported, which, as far as I can tell, is the only reason why the CI is currently failing. Just adding `from typing import Tuple` to the imports should fix that.

qcodes/instrument_drivers/stanford_research/SR830.py
jenshnielsen: In addition our docstrings should be valid restructured text using the Google docs extension see https://qcodes.github.io/Qcodes/examples/Creating%20Instrument%20Drivers.html#Documentation and https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

Specifically the args, returns and example sections should conform to the standard and you are not allowed to create other sections like the Units and Limitations. You probably want to put this in a notes section

astafan8:
@GeneralSarsby Noticed your last commit. Can you confirm that the feature has been tested on the actual instrument so that I can merge?

PR #1334, opened 2018-10-23
qcodes.dataset.* fool-proofing and minor fixes

astafan8:
Major changes proposed in this pull request:
* Raise when creating DataSet with path_to_db and connection both not-None
* Make run_id read-only attribute of DataSet + tests
* Make path_to_db read-only attribute of DataSet + tests
* Type annotation of load_or_create_experiment same as load_experiment
* Adding duplicate parameters: remove confusing todo and improve the test
* load_by_id does not allow None as argument (which would result in creating a new dataset)
* Only create new experiment if exp_id is not None
* load_last_experiment does not create new experiments if there a none in database

Minor changes proposed in this pull request:
* Remove comments that are outdated or useless
* Deprecate DataSet.add_parameters in favor of DataSet.add_parameter
* Experiment.last_data_set raises if there are no runs in the experiment
* Experiment.format_string is made a proper read-only property
* Test load_by_counter properly
* Test get_last_run properly
* Test load_experiment_by_name properly
* Test get_last_experiment properly
* Remove reference to Giulio as author of data_set.py
* Remove deprecated and unused hash_from_parts function
* Some pep8, reformatting, type annotation fixes here and there

@QCoDeS/core Feel free to say what to do else, or what to undo.

### DIFF
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -61,19 +61,19 @@ class CompletedError(RuntimeError):
 
 class _Subscriber(Thread):
     """
-    Class to add a subscriber to a DataSet. The subscriber gets called
-    every time an insert is made to the results_table.
+    Class to add a subscriber to a DataSet. The subscriber gets called every
+    time an insert is made to the results_table.
 
     The _Subscriber is not meant to be instantiated directly, but rather used
     via the 'subscribe' method of the DataSet.
 
     NOTE: A subscriber should be added *after* all parameters have been added.
 
-    NOTE: Special care shall be taken when using the *state* object: it is the user's
-    responsibility to operate with it in a thread-safe way.
+    NOTE: Special care shall be taken when using the *state* object: it is the
+    user's responsibility to operate with it in a thread-safe way.
     """
     def __init__(self,
-                 dataSet,
+                 dataSet: 'DataSet',
                  id_: str,
                  callback: Callable[..., None],
                  state: Optional[Any] = None,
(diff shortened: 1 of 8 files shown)


qcodes/dataset/data_set.py
```diff
@@ -436,14 +428,9 @@ def add_metadata(self, tag: str, metadata: Any):
...
             metadata: actual metadata
-
         """
-        # TODO: this follows the spec but another option:
-        # json_meta_data = json.dumps(metadata)
-        # add_meta_data(self.conn, self.run_id, {"metadata": json_meta_data})
-
         add_meta_data(self.conn, self.run_id, {tag: metadata})
-        # adding meta-data does not commit
+        # `add_meta_data` does not commit, hence we commit here:
```
WilliamHPNielsen: Perhaps better to make `add_meta_data` do an atomic transaction?

WilliamHPNielsen:
These are all improvements as far as I can tell, let's get them in swiftly. `mypy` has a small issue that you can quickly fix. I left one comment, but it does not have to be addressed in this PR, considering that it is carried over from the previous state of the code.

qcodes/dataset/data_set.py
astafan8: this will require a bit more refactoring because it will most probably touch `update_meta_data` and `insert_meta_data` function which i frankly did not want to worry about in this PR.

astafan8:
@QCoDeS/core see some more testing and fixing for `Experiment`-related parts.

PR #1336, opened 2018-10-24
ivvi.linspace

GeneralSarsby:
A replacement of numpy.linspace for use with the ivvi rack dacs.
        
The voltage output is not the same as the set voltage using
ivvi.set_dac1(v) due to 16 bit quantitation of the DAC.
If you only record the requested value, and not the value
from ivvi.get_dac1() the values will be slightly wrong.

This can lead to:
 - accidental oversampling (having the same voltage twice,
 thinking you have different voltages)
 - uneven spacing (changing by 2 bits, then 3, then 2 etc...)
 
This is solved by correctly rounding and picking the correct
start and end values with a fixed spacing with knowledge of
the quantitation.

ivvi.linsweep gives:
 - even spacing in voltages
 - recorded voltages match the set voltages
 - protection from oversampling


qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -183,7 +183,45 @@ def _get_version(self):
...
+            
+        Returns:
+            list of valid voltages for the ivvi rack
+            
+        Examples:
+            #normal dac aligned voltages
+            ivvi.linsweep(-200,200,201) -> [ -199.404, -197.391 ... ] 
+            
+            #protection from oversampling
+            ivvi.linsweep(start=108.8,end=109,points=11) ->
```
astafan8: is this really desired behavior? the caller requests 11 points (`..points=11`) but gets 4? If i was the user, i'd raise an exception in this case. but I am definitely not fully aware how this method is supposed to be used.

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -183,7 +183,45 @@ def _get_version(self):
...
+            ivvi.linsweep(-200,200,201) -> [ -199.404, -197.391 ... ] 
+            
+            #protection from oversampling
+            ivvi.linsweep(start=108.8,end=109,points=11) -> 
+                [108.82734416723888, 108.88838025482566,108.94941634241245, 109.01045242999923]
+        """
+        points = max(round(abs(points)),2)
+        use_reversed = end < start
+        if use_reversed:
+            start,end = end,start
```
astafan8: why is "automatic fixing" favored against over "strictness"? i'd suggest to raise an exception.

either way, this behavior should be documented in the docstring.

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -183,7 +183,45 @@ def _get_version(self):
...
+        dac_quata = 4.0/65535.0
+        byte_start =  int(round(start/1000.0/dac_quata))
+        byte_end = int(round(end/1000.0/dac_quata))
```
astafan8: where does `1000` come from? isn't it somewhere as an attribute of the instrument?

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -183,7 +183,45 @@ def _get_version(self):
...
+            
+            #protection from oversampling
+            ivvi.linsweep(start=108.8,end=109,points=11) -> 
+                [108.82734416723888, 108.88838025482566,108.94941634241245, 109.01045242999923]
+        """
+        points = max(round(abs(points)),2)
+        use_reversed = end < start
+        if use_reversed:
+            start,end = end,start
+        dac_quata = 4.0/65535.0
```
astafan8: this is clearly a magic number that needs to be defined as a constant (in the module, or at least as a class attribute). Are the values of `4.0` (i assume this is some range) and `65535.0` (i assume this is some bit depth/resolution) the same for all the ivvi's? Even if yes, they deserve to be declared as constants so that it is clear what they mean and what role they play in this method.

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -183,7 +183,45 @@ def _get_version(self):
             mes = self.ask(bytes([3, 4]))
             ver = mes[2]
             return ver
-
+    
+    def linsweep(self,start,end,points):
```
astafan8: Please use spaces and type annotations :)
```suggestion
    def linsweep(self, start: float, end: float, points: int) -> List[float]:
```

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -183,7 +183,45 @@ def _get_version(self):
...
+        """
+        A replacement of numpy.linspace for use with the ivvi rack dacs.
+        ivvi.linsweep gives:
+         - even spacing in voltages
+         - voltages match the set voltages
+         - protection from oversampling
+         
+        Args:
+            start (number): starting voltage
```
astafan8: no need for type information in the docstring thanks to python type annotations. (moreover, what is a "number"? "float or int"?)
```suggestion
            start: starting voltage
```

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -183,7 +183,45 @@ def _get_version(self):
...
+        points = max(round(abs(points)),2)
+        use_reversed = end < start
+        if use_reversed:
+            start,end = end,start
+        dac_quata = 4.0/65535.0
+        byte_start =  int(round(start/1000.0/dac_quata))
+        byte_end = int(round(end/1000.0/dac_quata))
+        delta_bytes =  abs(byte_end - byte_start)+1
+        spacing =  max(round(delta_bytes / (points-1)),1)
+        l =  [ el*dac_quata*1000 for el in range(byte_start,byte_end+spacing,spacing)]
```
astafan8: please, use spaces for readability
```suggestion
        l =  [el*dac_quata*1000 for el in range(byte_start, byte_end + spacing, spacing)]
```

qcodes/instrument_drivers/QuTech/IVVI.py
astafan8: is linsweep a good name? wouldn't the name of this function make the impression that it sweeps some voltage due to the word `sweep` in it? perhaps there is a better name that indeed refers to `numpy.linspace` like `dac_linspace` or `linspace_dac` or else?

GeneralSarsby: Taking on board your comments; I've put forward a new version.
I think I have addressed everything. But if I missed something it's just a mistake.

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -25,8 +25,11 @@ class IVVI(VisaInstrument):
     http://qtwork.tudelft.nl/~schouten/ivvi/doc-d5/rs232linkformat.txt
     A copy of this file can be found at the bottom of this file.
     '''
-    Fullrange = 4000
+    
+    Fullrange = 4000.0
     Halfrange = Fullrange / 2
+    resolution = 16
+    dac_quata = fullRange /2**resolution
```
astafan8: `fullRange` or `Fullrange`? :)

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -190,7 +193,101 @@ def get_all(self):
     def set_dacs_zero(self):
         for i in range(self._numdacs):
             self.set('dac{}'.format(i + 1), 0)
-
+    
+    def linspace(self,start,end,samples,flexible = False,bip=True):
```
astafan8:
```suggestion
    def linspace(self, start: float, end: float, samples: int, flexible: bool = False, bip: bool = True):
```

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -190,7 +193,101 @@ def get_all(self):
...
+        '''
+        Creates array of voltages, with correct alignment to the DAC
+        quantisation, in a similar manner to numpy.linspace.
+        This guarrientees an even spacing and no double sampling inside
```
astafan8:
```suggestion
        This guarantees an even spacing and no double sampling inside
```

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -190,7 +193,101 @@ def get_all(self):
...
+            # insufficient resolution
+            linspace(500, 502, 100) -> ValueError: Insufficient resolution
+                        for 100 samples in the range 500 to 502. Maximum :16
+                                     
+            # resolution limited sweep using the flexable option
+            linspace(500, 502, 100, true) -> [500.0991836423285, .. 14 more ..
+                                              , 501.9302662699321]
+            
+            # a too narrow range
+            linspace(0, 0.01, 100, True) #-> ValueError: No DAC values exist
```
astafan8:
```suggestion
            linspace(0, 0.01, 100, true) # -> ValueError: No DAC values exist
```

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -190,7 +193,101 @@ def get_all(self):
...
+                                             in the range 0 : 0.01
+        '''
```
astafan8: that's a killer docstring! great job!

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -190,7 +193,101 @@ def get_all(self):
...
+        This guarrientees an even spacing and no double sampling inside
+        the requested range.
+        
+        Args:
+            start : the start of the voltage range, in millivolts
+            end : the end of the voltage range, in millivolts
+            samples : number of sample voltages
+            flexible : occasionally get a different number of samples if
+                    they can still fit inside the range.
+            bip : if the dac set to bi-polar (-2V to +2V) or
```
astafan8: no examples for `bip` argument?

qcodes/instrument_drivers/QuTech/IVVI.py
```diff
@@ -190,7 +193,101 @@ def get_all(self):
...
+                                     
+            # resolution limited sweep using the flexable option
+            linspace(500, 502, 100, true) -> [500.0991836423285, .. 14 more ..
+                                              , 501.9302662699321]
+            
+            # a too narrow range
+            linspace(0, 0.01, 100, True) #-> ValueError: No DAC values exist
+                                             in the range 0 : 0.01
+        '''
+        import math
```
astafan8: Please move the import to the top of the module, it's generally a good practice because it is visible what is being used.

qcodes/instrument_drivers/QuTech/IVVI.py
astafan8: what is your opinion on this comment?

astafan8:
@GeneralSarsby this PR has been stalled for some time. Are you planning to finish it? will you need help with that? (if so, let me know)

qcodes/instrument_drivers/QuTech/IVVI.py
GeneralSarsby: or full_range?

qcodes/instrument_drivers/QuTech/IVVI.py
GeneralSarsby: This is not a case of automatic fixing. If the user wants decreasing values, they get it. 
It should be possible to the the algorithm generate values similar to numpy.linspace. 
The  algorithm only works forwards (increasing values), so we detect that case and at the end reverse the output to match what is expected.

GeneralSarsby:
I don't know what the requirements for the travis tests to pass.
What is it about  the docstring that causes the build to fail?

qcodes/instrument_drivers/QuTech/IVVI.py
astafan8: ok, justified. but the `this behavior should be documented in the docstring` comment still holds :)

GeneralSarsby:
@astafan8 
I'm only guessing at what is needed for the build to pass, trying to throw stuff against the wall and see what works.  It's something about the docstring. 
The  #QCoDeS CI failed  page says passed, but gives a failed status. 
The Travis CI build failed because of Sphinx complaining about indents on line 22. But I can't see the error.

astafan8:
@GeneralSarsby i'll have a look at it...

PR #1338, opened 2018-10-24
Adapt workaround for spyder UMR to work with newest version of spyder

jenshnielsen:
Yes I know this is horrible but until a fix for disabling this in spyder is available we should try our best to disable it


astafan8:
I'm not familiar with how spyder works under the hood and what this workaround actually does (although i get the general idea of it). and i didn't spend much time on it. So, i just trust you on this :)

PR #1344, opened 2018-10-26
Install python 3.6.7

jenshnielsen:
Matplotlib 3.0.1 has been released with a fix for the python 3.6.7 issue so install that too

It seem that this was actually an issue in matplotlib and not just the conda package but 3.0.1 fixes it


jenshnielsen:
Docs build broke on windows because of the following change in python 3.6.7

In 3.6.6
```python
In [11]: path.abspath('..\\qcodes\\')
Out[11]: 'C:\\Users\\jenielse\\source\\repos\\Qcodes\\qcodes'
```

In 3.6.7
```python
In [13]: path.abspath('..\\qcodes\\')
Out[13]: 'C:\\Users\\jenielse\\source\\repos\\Qcodes\\qcodes\\'
```

jenshnielsen:
A known regression on windows see
https://bugs.python.org/issue31047 and specifically https://bugs.python.org/msg328322

PR #1346, opened 2018-10-28
Update CryogenicSMS driver

lukabavdaz:
- Remove typos
- Hardcode terminator, because the magnet exclusively uses this terminator (note that the changed function signature is the only breaking change in this commit)
- Properly raise an exception when attempting to measure the field in 'AMPS' mode instead of raising a less helpful exception due to unexpected query results
- Make the driver compatible with SMS60C by removing an incorrect value at the start of some replies

@jenshnielsen


qcodes/instrument_drivers/cryogenic/CryogenicSMS120C.py
```diff
@@ -54,10 +55,10 @@ class CryogenicSMS120C(VisaInstrument):
     _re_float_exp = r'[-+]?(\d+(\.\d*)?|\.\d+)([eE][-+]?\d+)?'
 
     def __init__(self, name, address, coil_constant=0.113375, current_rating=105.84,
-                 current_ramp_limit=0.0506, reset=False, timeout=5, terminator='\r\n', **kwargs):
+                 current_ramp_limit=0.0506, reset=False, timeout=5, **kwargs):
 
         log.debug('Initializing instrument')
-        super().__init__(name, address, terminator=terminator, **kwargs)
+        super().__init__(name, address, terminator='\r\n', **kwargs)
```
jenshnielsen: This is slightly dangerous. Even if you were passing the default value this will now result in a `TypeError: __init__() got multiple values for keyword argument 'terminator'`

I would add some lines like the following to the start of the function above
```
if 'terminator' in kwargs.keys():
    kwargs.pop('terminator')
    warnings.warn('Passing terminator to CryogenicSMS is no longer supported and has no effect')
```

lukabavdaz: That's a great way of retaining backwards compatibility!

PR #1350, opened 2018-10-29
Add helper function to get path from conn

WilliamHPNielsen:
This little helper function turns out to be convenient for the copy-pasting of runs from one DB to another. Adding it here for "good practice" purposes.

- Add a helper function to get a path from a connection
- Add a test for said function

@QCoDeS/core


qcodes/dataset/database.py
```diff
@@ -43,3 +44,14 @@ def initialise_or_create_database_at(db_file_with_abs_path: str) -> None:
     """
     qcodes.config.core.db_location = db_file_with_abs_path
     initialise_database()
+
+
+def path_of_connection(conn: SomeConnection) -> str:
+    """
+    Return the path of the connection
```
jenshnielsen: Sorry I am being stupid. Is this the path to the db file?

qcodes/dataset/database.py
```diff
@@ -43,3 +44,14 @@ def initialise_or_create_database_at(db_file_with_abs_path: str) -> None:
     """
     qcodes.config.core.db_location = db_file_with_abs_path
     initialise_database()
+
+
+def path_of_connection(conn: SomeConnection) -> str:
```
astafan8: What about a different name?
```suggestion
def db_file_of_connection(conn: SomeConnection) -> str:
```

Dominik-Vogel: I fully agree to both of your suggestions @astafan8 !

WilliamHPNielsen: Although I acknowledge the confusion, I will stick with `path_of_connection`, since a path is indeed what this function returns. If that is too unclear, we could also go full out with `path_to_db_file_of_connection`.

WilliamHPNielsen:
@astafan8, that's a good idea, but a bit more work than I am willing to put in here. Keep in mind that "our" connection object is called `ConnectionPlus`, but everywhere in the code we accept `SomeConnection`, which is the union of `ConenctionPlus` and `sqlite3.Connection`. It's not just a simple subclass used everywhere. 

But the idea is still good.

qcodes/dataset/database.py
astafan8: `db_path_of_connection` ? :) otherwise, i agree with the "full out explicit" one

PR #1351, opened 2018-10-29
Try to speed up azure by installing the correct python version

jenshnielsen:
Currently is is first installing 3.7 and then installing 3.6 on top of this. Hopefully this will speed things up at least a bit


jenshnielsen:
Compared to https://quarcsw.visualstudio.com/QubitWorkbench/_build/results?buildId=7899&view=logs it looks like this might save ~50 sec, not much but better than nothing

PR #1353, opened 2018-10-30
[WIP] Extension/dynamic module

### LINKED ISSUE #1342: Make it possible to dynamically create/discover/delete sub-modules on instruments
The QCoDeS design allowes drivers to have sub-modules. Often times these are instrument channels and are fixed in number. But sub-modules need not be tied to physical instrument channels. In fact we sometimes encounter instruments with the ability to dynamically create and delete sub-modules. An example of this are measurement traces and windows on the Keysight N52xx precision network analyzers (PNA). Another instrument which can benifit from dynamic sub-moduling is the Keysight M8020 digital pattern generator, whereby each pattern can be defined in its own sub-module. 

We therefore propose the following enhancement to the InstrumentChannel class in QCoDeS: 

1. Change the name to (something similar to) InstrumentSubmodule
2. Develop a way to create/discover/delete submodules dynamically. And implementation to accomplish this for the Keysight precision network analyzers is shown [here](https://github.com/sohailc/Qcodes/blob/doc/PNA_N5222B_example/qcodes/instrument_drivers/Keysight/N52xx/_N52xx_channel_ext.py). 

Additionally, we propose to enhance the ChannelList class by adding an `add` method which will create a new sub-module of the same type. An example of this is likewise shown in the above link. The example file for the N52xx PNA in my branch is also a good resource to understand [this](https://github.com/sohailc/Qcodes/blob/doc/PNA_N5222B_example/docs/examples/driver_examples/Qcodes_example_with_Keysight_Network_Analyzer.ipynb)  

Since a similar design will benefit other instruments as well, I propose to make dynamic sub-moduling available generally

sohailc:
Fixes #1342 

This PR is still work in progress, but I would like to have comments early in the process.


qcodes/instrument/channel.py
```diff
@@ -529,3 +532,183 @@ def validate(self, value, context: str='') -> None:
...
+        super().__init__(
+            parent, name, chan_type, chan_list, snapshotable,
+            multichan_paramclass
+        )
+
+        new_channels = self._chan_type.load_from_instrument(
+            self._parent, channel_list=self, **kwargs)
+
+        for channel in new_channels:
+            # NB: There is a bug in `extend`. TODO: Make a PR!
```
astafan8: what is the bug about? Are you planning to make a PR with a test and a fix? That'd be great to do before landing this PR.

qcodes/instrument/channel.py
```diff
@@ -529,3 +532,183 @@ def validate(self, value, context: str='') -> None:
...
+        """
+        raise NotImplementedError("Please subclass")
+
+    def delete(self) ->None:
+        """
+        Delete the channel from the instrument and remove from channel list
+        """
+        self._delete()
+        if self._channel_list is not None:
+            self._channel_list.remove(self)
```
astafan8: we should not forget to check that `self in self._channel_list`.

qcodes/instrument/channel.py
```diff
@@ -529,3 +532,183 @@ def validate(self, value, context: str='') -> None:
...
+            name: channel name
+            existence (bool): True if the channel exists on the instrument
+            channel_list (N52xxChannelList): If the channel is deleted,
+                delete `self` from the channel list
+        """
+        super().__init__(parent, name=name, **kwargs)
+        self._exists_on_instrument = existence
+        self._channel_list = channel_list
+
+    def create(self) ->None:
```
astafan8: should this do nothing if `self._exists_on_instrument` is True?

qcodes/instrument/channel.py
```diff
@@ -529,3 +532,183 @@ def validate(self, value, context: str='') -> None:
...
+    def _create(self) ->None:
+        """
+        (SCPI) commands needed to create the instrument. Note that we
+        need to use `self.root_instrument.write` to send commands,
+        because self.write will cause _assert_existence to raise a
+        runtime error.
+        """
+        raise NotImplementedError("Please subclass")
+
+    def delete(self) ->None:
```
astafan8: should this do nothing if `self._exists_on_instrument` is False?

qcodes/instrument/channel.py
```diff
@@ -529,3 +532,183 @@ def validate(self, value, context: str='') -> None:
...
+    context does not necessarily mean a physical instrument channel,
+    but rather an instrument sub-module. For some instruments,
+    these sub-modules can be created and deleted at will.
+    """
+
+    @classmethod
+    def load_from_instrument(
+            cls, parent: Instrument, **kwargs) ->List['AutoLoadableInstrumentChannel']:
+        """
+        Discover and load channels from the instrument
```
astafan8: Mentioning `discover` here IMHO makes it a bit confusing.
```suggestion
        Load channels that already exist on the instrument
```

qcodes/tests/test_autoloadable_channels.py
```diff
@@ -0,0 +1,186 @@
...
+        # available
+
+
+def test_channels_list():
+    instrument = DummyInstrumentWithChannelList("instrument2")
+    assert instrument.channels[0].hello() == "Hello from channel 1"
+    assert instrument.channels[1].hello() == "Hello from channel 2"
+    assert instrument.channels[2].hello() == "Hello from channel 4"
+
+    new_channel = instrument.channels.add()
```
astafan8: what happens when the instrument is closed but there are still instances of channels lying around? their methods should be protected against this in order to help in debugging possible weird scenarios.

qcodes/tests/test_autoloadable_channels.py
```diff
@@ -0,0 +1,186 @@
...
+
+class DummyInstrumentWithChannelList(TestInstrument):
+    def __init__(self, name):
+        super().__init__(name)
+
+        channels = AutoLoadableChannelList(self, "channels", SimpleTestChannel)
+        self.add_submodule("channels", channels)
+
+
+def test_sanity():
```
astafan8: thanks for the test! but an example notebook with at least (for now) bare intentions of this infra, would be appreciated. This will also help to (if necessary) improve the design, strip down excessive code, and perhaps even add something. (yes, yes, i know that everything is expressed in the description of the linked issue, but still :) )

qcodes/instrument/channel.py
```diff
@@ -529,3 +532,183 @@ def validate(self, value, context: str='') -> None:
             raise ValueError(
                 '{} is not part of the expected channel list; {}'.format(
                     repr(value), context))
+
+
+class AutoLoadableInstrumentChannel(InstrumentChannel):
```
astafan8: +1 for explicit names :)

qcodes/instrument/channel.py
```diff
@@ -529,3 +532,183 @@ def validate(self, value, context: str='') -> None:
...
+        """
+        new_kwargs = self._chan_type.get_new_instance_kwargs(self._parent, **kwargs)
+
+        new_channel: AutoLoadableInstrumentChannel = self._chan_type(
+            parent=self._parent, channel_list=self, **new_kwargs
+        )
+
+        new_channel.create()
+        self.append(new_channel)
+        return new_channel
```
astafan8: are dropping the intention to rename `channel` to `submodule`? if we do the renaming, it will be important to also rename related variables/attributes in the code...

qcodes/tests/test_autoloadable_channels.py
```diff
@@ -0,0 +1,186 @@
...
+        existing_channels = [int(i) for i in channels_str.split(",")]
+
+        new_channel = 1
+        while new_channel in existing_channels:
+            new_channel += 1
+
+        kwargs = {
+            "name": f"channel{new_channel}",
+            "channel": new_channel
+        }
```
astafan8: in this case, technically `name` and `channel` are not kwargs :) yes, this will work fine, but it leaves a confusion. Can this be improved or at least explained in more detail(ideally, formalized in code :) )?

qcodes/tests/test_autoloadable_channels.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def __init__(self, name):
+        super().__init__(name)
+
+        self._backend = TestInstrumentBackend()
+
+        self.add_parameter(
+            "channel_catalog",
+            get_cmd=":INST:CHN:CAT",
+        )
```
astafan8: to my understanding, this parameter belongs to `DummyInstrumentWithChannelList`, right?

qcodes/instrument/channel.py
```diff
@@ -529,3 +532,183 @@ def validate(self, value, context: str='') -> None:
...
+
+        obj_list = []
+        for new_kwargs in cls._discover_from_instrument(parent, **kwargs):
+            obj = cls(parent, existence=True, **new_kwargs)
+            obj_list.append(obj)
+
+        return obj_list
+
+    @classmethod
+    def _discover_from_instrument(
```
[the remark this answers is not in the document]
sohailc: I don't see how actually. They both to different things.  One reads init kwargs from the instrument and in the other, the programmer can himself apply a strategy for coming up with needed kwargs. For instance, for an argument like `name`, we can apply the strategy that we randomly generate a unique name. That may not be appropriate, but the programmer can choose to do that. In the example provided in the test, I look for the smallest channel number available for the `channel` keyword argument. Alternatively, I could have looked up the largest number already present on the instrument and added plus one for the new channel. So it is up to the  developer how to implement `get_new_instance_kwargs` but `_discover_from_instrument` is fixed.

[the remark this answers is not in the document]
astafan8: I see. What I am worried about is handling of required/significant kwargs vs optional kwargs. ok, let's hope that `__init__` signature will be that single source of truth in the end.

qcodes/instrument/channel.py
```diff
@@ -529,3 +532,183 @@ def validate(self, value, context: str='') -> None:
...
+            cls, parent: Instrument, **kwargs) ->List[dict]:
+        """
+        Discover channels on the instrument and return a list kwargs to create
+        these channels in memory
+
+        Args:
+            parent (Instrument): The instrument through which the instrument
+                channel is accessible
+            **kwargs (dict): Keyword arguments needed to discover the channels
```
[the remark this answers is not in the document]
sohailc: I think there is some confusion. The `existence` keyword argument you are referring to is passed on the initialization of the channel class, while the keyword arguments we are referring to in this comment are passed on to the class method `_discover_from_instrument`, which does not have an `existence` argument. 

The keyword arguments supplied to `_discover_from_instrument` come from `load_from_instrument`. A snippet of code could look like: 

```python
class MyInstrument(VisaInstrument): 
    def __init__(self, name, address): 
        super().__init__(name, address)
         #the `folder="shared:"` keyword argument is passed on to `discover_from_instrument`
         # via  `load_from_instrument`
        channels = AutoLoadableChannelList(self, "channels", MyChannelClass, folder="shared:")
```

The keyword arguments supplied to `_discover_from_instrument` are whatever arguments are needed to load the channels correctly.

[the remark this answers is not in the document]
sohailc: But I will add better docstrings to this comes more clear

qcodes/instrument/channel.py
```diff
@@ -529,3 +532,183 @@ def validate(self, value, context: str='') -> None:
...
+
+        Returns:
+              list: List of keyword arguments to initialize channels
+        """
+        raise NotImplementedError("Please subclass")
+
+    @classmethod
+    def get_new_instance_kwargs(cls, parent: Instrument, **kwargs)->dict:
```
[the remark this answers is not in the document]
sohailc: Sure, we can make `parent` an optional keyword argument. Some instruments indeed do not need the parent instrument, while other do.

qcodes/instrument/channel.py
sohailc: That is indeed correct. We have two options: 
1) We can quietly do nothing if the channel already exists on the instrument 
2) Raise an exception

But I do think we need to check if it exists already; the code as-is is not sufficient.

qcodes/instrument/channel.py
sohailc: If it is not on the instrument (yet), this method will probably raise an exception because `AssertExistance` will fail with a runtime error when we want to write or ask from the instrument. Never the less, as with `create`, we want to assert that the channel is present before deleting and raise a custom exception for this.

qcodes/instrument/channel.py
sohailc: True! I will add that

qcodes/instrument/channel.py
sohailc: Ok, I will make a PR for that as well. Yes, there is a bug I discovered in the `ChannelList` class, but I have not come round to fixing this and making a PR. There is something wrong with the `extend` method in the `ChannelList` class, but I forgot what, exactly :-)

qcodes/instrument/channel.py
sohailc: It's not that I have dropped the intention, but rather I think this should be a PR on its own as it is a big change. Would you not agree?

sohailc: If Iapply the new naming convention partially, I fear that this will cause confusion. Since we are calling it `channel` for now, lets keep that for now.

qcodes/tests/test_autoloadable_channels.py
sohailc: Well, `name` is `channel1` for instance and `channel` is an integer `1`. Why are these not kwargs? I am not sure what additional detail is required.

qcodes/tests/test_autoloadable_channels.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def write_raw(self, cmd: str):
+        return self._backend.send(cmd)
```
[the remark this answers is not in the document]
sohailc: Yes, that is true. I will fix this.

qcodes/tests/test_autoloadable_channels.py
sohailc: I honestly think that adding a driver which uses this infrastructure will be illuminating, perhaps even more so than a notebook as it will show the complexities involved with a real instrument. I will make the PNA driver use the infrastructure which I think will be helpful.

qcodes/tests/test_autoloadable_channels.py
```diff
@@ -0,0 +1,186 @@
...
+
+    with pytest.raises(RuntimeError):
+        new_channel.hello()  # We have only instantiated a channel class, we
+        # have yet to create it on the instrument
+
+    new_channel.create()
+    assert new_channel.hello() == "Hello from channel 3"
+    new_channel.delete()
+
+    with pytest.raises(RuntimeError):
```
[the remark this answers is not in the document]
sohailc: Ok

qcodes/tests/test_autoloadable_channels.py
sohailc: That is indeed a good point, but the same can be said for any submodule, even existing channel classes. What if for the SR860 I do something like: 

```python 
lockin = SR860("lockin", "GPIB0::4::INSTR")
data_channel = lockin.data_channel_3
lockin.close()
```
You will have exactly the same problem. So this is a problem that should be fixed in another PR.

qcodes/tests/test_autoloadable_channels.py
```diff
@@ -0,0 +1,186 @@
...
+    assert instrument.channels[0].hello() == "Hello from channel 1"
+    assert instrument.channels[1].hello() == "Hello from channel 2"
+    assert instrument.channels[2].hello() == "Hello from channel 4"
+
+    new_channel = instrument.channels.add()
+    assert len(instrument.channels) == 4
+    assert new_channel is instrument.channels[-1]
+    assert new_channel.hello() == "Hello from channel 3"
+
+    new_channel.delete()
```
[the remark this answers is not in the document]
sohailc: Ok, I think this is a good idea. The reason why is that when you add a new channel through the `add` method of the `AutoLoadableChannelList` class, it will get new instance kwargs, which may be valid for that particular moment. If you delete a channel the circumstances will have changed (e.g. there is another channel already with the same name), so it is a good idea to clock the re-creation after deletion.

qcodes/instrument/channel.py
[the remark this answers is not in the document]
sohailc: Well we intend for the `AutoLoadableInstrumentChannel` to be sub-classed and every instrument type will have its own required and optional arguments. There is no way to know in the base class which ones are required. The exception is the `name` argument, which must be present in the kwargs dictionary.

qcodes/instrument/channel.py
astafan8: exactly, no need for renaming at the moment.

qcodes/tests/test_autoloadable_channels.py
astafan8: to my understanding it does not, because it is not used in this class. This parameter is only used by the channels that are created in `DummyInstrumentWithChannelList`. `TestInstrument` as an entity can exist perfectly without the notion of the channels.

I made this comment because I was a bit confused when reading the code why you create `DummyInstrumentWithChannelList` __and__ `TestInstrument`. Then I understood that `TestInstrument` was made to hide the "mock" backend. But then having that parameter in `TestInstrument` does not make much sense, but makes a lot of sense in `DummyInstrumentWithChannelList` since channels rely on it.

qcodes/tests/test_autoloadable_channels.py
astafan8: ok, i'll make an issue

qcodes/tests/test_autoloadable_channels.py
```diff
@@ -105,82 +128,94 @@ def __init__(
...
+    assert e1.value.args[0] == "Object does not exist (anymore) on the instrument"
+    # After creating the channel we should be able to talk to the instrument.
     new_channel.create()
     assert new_channel.hello() == "Hello from channel 3"
-    new_channel.delete()
-
-    with pytest.raises(RuntimeError):
+    # Once we remove the channel we should no longer be able to talk to the instrument
+    new_channel.remove()
+    with pytest.raises(RuntimeError) as e2:
```
astafan8: you can use `pytest.raises(Exception, match="Object does not exist (anymore) on the instrument")` instead of working with `e2` explicitly.

sohailc: Oh I did not know that! Thanks!

qcodes/instrument/channel.py
```diff
@@ -529,3 +535,222 @@ def validate(self, value, context: str='') -> None:
...
+        Add a channel to the list
+        """
+        new_channel = self._chan_type.new_instance(  # type: ignore
+            self._parent,
+            create_on_instrument=True,
+            channel_list=self,
+            **kwargs
+        )
+
+        self.append(new_channel)
```
sohailc: Previously, we used a loop here and called `append` at every iteration, instead of `extend` because of a bug in that method. I made PR to fix that bug: https://github.com/QCoDeS/Qcodes/pull/1368/files 

Before merging this PR, we should approve and merge that PR first

qcodes/instrument/channel.py
```diff
@@ -529,3 +535,222 @@ def validate(self, value, context: str='') -> None:
...
+    ) ->None:
+
+        super().__init__(
+            parent, name, chan_type, chan_list, snapshotable,
+            multichan_paramclass
+        )
+        new_channels = self._chan_type.load_from_instrument(  # type: ignore
+            self._parent, channel_list=self, **kwargs)
+
+        self.extend(new_channels)
```
sohailc: Previously, we used a loop here and called append at every iteration, instead of extend because of a bug in that method. I made PR to fix that bug: https://github.com/QCoDeS/Qcodes/pull/1368/files
Before merging this PR, we should approve and merge that PR first

sohailc:
@astafan8 Can you approve this PR if you have no further objections? The tests and CI seem to be passing. 
@Dominik-Vogel Can you have a look at this PR as well?

qcodes/instrument/channel.py
```diff
@@ -532,3 +535,222 @@ def validate(self, value, context: str='') -> None:
...
+
+    def add(self, **kwargs) ->'AutoLoadableInstrumentChannel':
+        """
+        Add a channel to the list
```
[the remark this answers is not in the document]
astafan8: also that the created channel is not only added to the channel list but its instance is also returned.

qcodes/tests/test_autoloadable_channels.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+class MockBackendBase:
+    """
+    A very simple mock backend that contains a dictionary with string keys and
+    callable values. The keys are matched to input commands with regular
+    expressions and on match the corresponding callable is called.
+    """
+    def __init__(self)->None:
+        self._command_dict: dict = {}
```
astafan8:
```suggestion
        self._command_dict: Dict[str, Callable] = {}
```

qcodes/tests/test_autoloadable_channels.py
astafan8: if you look at the channel's `__init__` method, you'll see that technically `name` is not a `kwarg`, but is an `arg` :) it just works because when you give a kwarg it also looks in args for it.

it would be nice if this can be explicitly mentioned in the `_get_new_instance_kwargs` docstring of the parent class.

qcodes/tests/test_autoloadable_channels.py
```diff
@@ -0,0 +1,282 @@
...
+    # conflicting names
+    new_channel_kwargs = SimpleTestChannel._get_new_instance_kwargs(
+        dummy_instrument)
+
+    new_channel = SimpleTestChannel(dummy_instrument, **new_channel_kwargs)
+    # Instrument IO through the newly instantiated channel should raise an
+    # exception before actually creating the channel on the instrument
+    with pytest.raises(
+            RuntimeError,
+            message="Object does not exist (anymore) on the instrument"):
```
astafan8: `message` `kwarg` specifies the message that will be printed in case the code inside the context manager __does not__ raise. `match` `kwarg` specifies a string against which the message of the expected exception should match. The former is convenient for testing, the latter is very important because there may be many `RuntimeError`s while we expect a very specific one - the one with the message that is specified in `match=` `kwarg`.
```suggestion
            match="Object does not exist (anymore) on the instrument"):
```

qcodes/tests/test_autoloadable_channels.py
```diff
@@ -0,0 +1,282 @@
...
+
+        new_channel.hello()
+
+    # After creating the channel we should be able to talk to the instrument.
+    new_channel.create()
+    assert new_channel.hello() == "Hello from channel 3"
```
astafan8: spaces for readability:
```suggestion
    assert new_channel.hello() == "Hello from channel 3"
    
```

qcodes/tests/test_autoloadable_channels.py
```diff
@@ -0,0 +1,282 @@
...
+    # Once we remove the channel we should no longer be able to talk to the
+    # instrument
+    new_channel.remove()
+    with pytest.raises(
+            RuntimeError,
+            message="Object does not exist (anymore) on the instrument"):
```
astafan8:
```suggestion
            match="Object does not exist (anymore) on the instrument"):
```

qcodes/tests/test_autoloadable_channels.py
```diff
@@ -0,0 +1,282 @@
...
+    assert new_channel is dummy_instrument.channels[-1]
+    assert new_channel.hello() == "Hello from channel 3"
+    # Test that we can remove them
+    new_channel.remove()
+    assert len(dummy_instrument.channels) == 3
+    assert new_channel not in dummy_instrument.channels
+    # Once removed we should no longer be able to talk to the channel
+    with pytest.raises(
+            RuntimeError,
+            message="Object does not exist (anymore) on the instrument"):
```
astafan8:
```suggestion
            match="Object does not exist (anymore) on the instrument"):
```

qcodes/tests/test_autoloadable_channels.py
```diff
@@ -0,0 +1,282 @@
...
+    Test the basic functionality of the auto-loadable channels, without using
+    the auto-loadable channels list
+    """
+    # Test that we are able to discover instruments automatically
+    channels = SimpleTestChannel.load_from_instrument(dummy_instrument)
+    assert len(channels) == 4
+    assert channels[0].hello() == "Hello from channel 1"
+    assert channels[1].hello() == "Hello from channel 2"
+    assert channels[2].hello() == "Hello from channel 4"
+    assert channels[3].hello() == "Hello from channel 5"
```
astafan8: why is this channel with `5` here if it was supposed to be skipped thanks to the `channels_to_skip=["5"]`? or am i misunderstanding something?

it would be great if there was an explicit assert/test for the `channels_to_skip=["5"]` part with some explanation in the comments (i appreciate the comments on this in SimpleTestChannel though).

qcodes/instrument/channel.py
```diff
@@ -532,3 +535,222 @@ def validate(self, value, context: str='') -> None:
...
+        """
+        Delete the channel from the instrument and remove from channel list
+        """
+        self._assert_existence()
+        self._remove()
+        if self._channel_list is not None and self in self._channel_list:
+            self._channel_list.remove(self)
+        self._exists_on_instrument = False
+
+    def _remove(self) ->None:
```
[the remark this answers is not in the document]
sohailc: Ok

qcodes/instrument/channel.py
```diff
@@ -532,3 +535,222 @@ def validate(self, value, context: str='') -> None:
...
+    def write(self, cmd: str)->Any:
+        """
+        Only write to the instrument if the channel is present on the instrument
+        """
+        self._assert_existence()
+        return super().write(cmd)
+
+    def ask(self, cmd: str)->None:
+        """
+        Only query to the instrument if the channel is present on the instrument
```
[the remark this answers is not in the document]
sohailc: Ok

qcodes/instrument/channel.py
```diff
@@ -532,3 +535,222 @@ def validate(self, value, context: str='') -> None:
...
+    def exists_on_instrument(self)->bool:
+        return self._exists_on_instrument
+
+
+class AutoLoadableChannelList(ChannelList):
+    """
+    Extends the QCoDeS ChannelList class to add the following features:
+    - Automatically create channel objects on initialization
+    - Make a `add` method to create channel objects
+    """
```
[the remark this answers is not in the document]
sohailc: Yes, that's true. I will put that in the docstring

qcodes/instrument/channel.py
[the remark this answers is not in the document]
sohailc: Ok

qcodes/tests/test_autoloadable_channels.py
sohailc: Ok

qcodes/tests/test_autoloadable_channels.py
sohailc: Ok

qcodes/tests/test_autoloadable_channels.py
sohailc: We the `channels_to_skip=["5"]` code is called when using the ChannelList. In this test, `load_from_instrument` is called without any kwargs. So all channels are present. 

I will make an additional test to see if the kwargs are propagated correctly.

qcodes/tests/test_autoloadable_channels.py
sohailc: The match keyword as you use it here does not work. I will look into how to do this properly.

qcodes/tests/test_autoloadable_channels.py
sohailc: Is it me, or do I not see a difference between the lines.

qcodes/tests/test_autoloadable_channels.py
astafan8: this might be because you need to escape the brackets (if i remember correctly, putting `r` in front of the string solves it), because `match` generally expects a regexp string.

qcodes/tests/test_autoloadable_channels.py
astafan8: extra newline after the assert statement :)

qcodes/tests/test_autoloadable_channels.py
sohailc: On second thought, I added the following to the docstring of the test: 

Please note that the `channels_to_skip` argument in the dummy instrument applies when accessing channels from the channels list. Since we are calling `load_from_instrument` directly in this test without this keyword argument, we will see all channels.

qcodes/tests/test_autoloadable_channels.py
sohailc: Yes, thats it! Thanks Mikhail

PR #1358, opened 2018-10-31
More Documentation cleanup

jenshnielsen:
* Restructure the conf file to be less confusing
* This uncovered a bug that meant that some of the pages that should have been included were not
* Those pages had rst syntax issues that I fixed
* Start adding api docs for the dataset

### DIFF
--- a/docs/api/public.rst
+++ b/docs/api/public.rst
@@ -89,6 +89,7 @@ DataSet
     qcodes.dataset.experiment_container.Experiment
     qcodes.dataset.experiment_container.new_experiment
     qcodes.dataset.experiment_container.load_last_experiment
+    qcodes.dataset.experiment_container.load_experiment_by_name
     qcodes.dataset.experiment_container.load_or_create_experiment
 
     qcodes.dataset.database.initialise_database
--- a/docs/api/public.rst
+++ b/docs/api/public.rst
@@ -89,6 +89,7 @@ DataSet
     qcodes.dataset.experiment_container.Experiment
     qcodes.dataset.experiment_container.new_experiment
     qcodes.dataset.experiment_container.load_last_experiment
+    qcodes.dataset.experiment_container.load_or_create_experiment
 
     qcodes.dataset.database.initialise_database
     qcodes.dataset.database.initialise_or_create_database_at
--- a/docs/api/public.rst
+++ b/docs/api/public.rst
@@ -77,6 +77,15 @@ Data
     GNUPlotFormat
     DiskIO
 
+DataSet
+~~~~~~~
+
+.. autosummary::
+   :toctree: generated/
+
+    qcodes.dataset.measurements.DataSaver
+    qcodes.dataset.measurements.Runner
+    qcodes.dataset.measurements.Measurement
 
 
 Instrument
--- a/docs/api/public.rst
+++ b/docs/api/public.rst
@@ -95,6 +95,8 @@ DataSet
     qcodes.dataset.database.initialise_database
     qcodes.dataset.database.initialise_or_create_database_at
 
+    qcodes.dataset.data_set.load_by_id
+
     qcodes.dataset.plotting.plot_by_id
 
 Instrument
--- a/docs/api/public.rst
+++ b/docs/api/public.rst
@@ -83,10 +83,17 @@ DataSet
 .. autosummary::
    :toctree: generated/
 
-    qcodes.dataset.measurements.DataSaver
-    qcodes.dataset.measurements.Runner
     qcodes.dataset.measurements.Measurement
+    qcodes.dataset.measurements.DataSaver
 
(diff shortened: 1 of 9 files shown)


qcodes/utils/plotting.py
```diff
@@ -217,27 +219,30 @@ def auto_color_scale_from_config(colorbar: matplotlib.pyplot.colorbar,
                                  ) -> None:
     """
     Sets the color limits such that outliers are disregarded, depending on
-    the configuration file `qcodesrc.json`.
-
-    Config:
```
jenshnielsen: @Dominik-Vogel This is not a valid google style header. I have rewritten it more traditionally. Is that ok? Otherwise we can come up with something smarter that tries to match what you had but is still correct

docs/api/public.rst
```diff
@@ -77,6 +77,15 @@ Data
...
     DiskIO
 
+DataSet
+~~~~~~~
+
+.. autosummary::
+   :toctree: generated/
+
+    qcodes.dataset.measurements.DataSaver
+    qcodes.dataset.measurements.Runner
```
astafan8: not sure this is needed as users never interact with this guy.

docs/api/public.rst
```diff
@@ -77,6 +77,15 @@ Data
     GNUPlotFormat
     DiskIO
 
+DataSet
+~~~~~~~
+
+.. autosummary::
+   :toctree: generated/
+
```
astafan8: it is worth it to include DataSet here as well, at least for the time being. Also, we should definitely add all the `load_by_id`, `plot_by_id`, `initialise_database`, etc. stuff. Are they in already or not?

docs/conf.py
```diff
@@ -68,11 +79,7 @@
 copyright = '2016, Giulio Ungaretti, Alex Johnson'
 author = 'Giulio Ungaretti, Alex Johnson'
```
astafan8: the founding fathers :)

jenshnielsen:
Travis revealed that the `keysight.private` module was not installed in a regular (non -e install) adding the ``__init__.py`` file in turn revealed that there was a syntax error in the type annotation of that module

docs/api/public.rst
jenshnielsen: This is really only baby steps. The whole API documentation needs a significant reorg. I was planning to do that gradually. In multiple prs

docs/conf.py
jenshnielsen: @QCoDeS/core Yes, should it include all core devs or should we simply change it to `the QCoDeS developers`?

astafan8: I have no opinion about this :)

but it's probably fair to say that nobody is holding the copyright.

jenshnielsen: Yes I think copyright should just be QCoDeS developers of some such generic thing

qcodes/utils/plotting.py
Dominik-Vogel: I am fine with this. I just did it like this to make it easier to read and considered the non conform part as part of the main doc string...

docs/api/public.rst
```diff
@@ -77,7 +77,23 @@ Data
...
+    qcodes.dataset.measurements.Measurement
+    qcodes.dataset.measurements.DataSaver
+
+    qcodes.dataset.experiment_container.Experiment
+    qcodes.dataset.experiment_container.new_experiment
+    qcodes.dataset.experiment_container.load_last_experiment
```
astafan8:
```suggestion
    qcodes.dataset.experiment_container.load_last_experiment
    qcodes.dataset.experiment_container.load_or_create_experiment
```

docs/api/public.rst
```diff
@@ -77,7 +77,23 @@ Data
...
+
+    qcodes.dataset.database.initialise_database
+    qcodes.dataset.database.initialise_or_create_database_at
 
+    qcodes.dataset.plotting.plot_by_id
```
astafan8:
```suggestion
    qcodes.dataset.data_set.load_by_id

    qcodes.dataset.plotting.plot_by_id
```

docs/api/public.rst
```diff
@@ -77,7 +77,24 @@ Data
...
+.. autosummary::
+   :toctree: generated/
+
+    qcodes.dataset.measurements.Measurement
+    qcodes.dataset.measurements.DataSaver
+
+    qcodes.dataset.experiment_container.Experiment
+    qcodes.dataset.experiment_container.new_experiment
+    qcodes.dataset.experiment_container.load_last_experiment
+    qcodes.dataset.experiment_container.load_or_create_experiment
```
astafan8:
```suggestion
    qcodes.dataset.experiment_container.load_experiment_by_name
    qcodes.dataset.experiment_container.load_or_create_experiment
```

PR #1362, opened 2018-10-31
More consistent table of content for notebooks

jenshnielsen:
To render the notebooks nicely as part of our docs there should only be one 1 level heading at the top of the notebook. This changes a bunch of notebooks that had more.

 In addition

* Level 3 headings should be inside a level 2 section not freestanding. I have not double checked all instances but I have at least fixed on instance where this is not the case. That seems to cause the levels to be revers in the output. Which may be a bug in nbsphinx/nbconvert
* Make the toc for instrument notebooks only one level deep. I think this makes it much easier to get an overview of which example are there. Still to do is to give them more consistent names
* Correct a few typos here and there as i found them


jenshnielsen:
@QCoDeS/core Could someone have a quick look at this? It's quite simple but makes the presentation of notebooks in the docs so much nicer

docs/examples/DataSet/example-creation.ipynb
```diff
@@ -25,9 +24,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Experiment container and dasets\n",
+    "# Experiment container and datasets\n",
```
Dominik-Vogel: What is it about with the double # in the other notebooks?

Dominik-Vogel: ok now I get it sorry.... I was confused by the source view

Dominik-Vogel:
Somehow my PR from yesterady fails here on 3.7, even though it passed in the original pr....

PR #1363, opened 2018-11-01
Snapshot diff and method chaining

cgranade:
- Add new utility functions to quickly compare which parameter settings changed between two runs.
- Make `Measurement` methods follow fluent style to allow for method chaining.

As per discussions with @astafan8 and @WilliamHPNielsen.


qcodes/utils/metadata.py
```diff
@@ -42,3 +49,67 @@ def snapshot_base(self, update: bool=False,
...
+def diff_param_values_by_id(left_id, right_id):
+    # Local import to reduce load time and
+    # avoid circular references.
+    from qcodes.dataset.data_set import load_by_id
+    """
+    Given the IDs of two datasets, returns the differences between
+    parameter values in each of their snapshots.
+    """
+    return diff_param_values(
+        json.loads(load_by_id(left_id).get_metadata('snapshot')),
```
astafan8: `DataSet` has `snapshot` property that returns the dict. it can be used here instead of `json.loads` and `get_metadata` parts.

qcodes/utils/metadata.py
```diff
@@ -42,3 +49,67 @@ def snapshot_base(self, update: bool=False,
...
+        changed={
+            key: (left_params[key], right_params[key])
+            for key in common_keys
+            if left_params[key] != right_params[key]
+        }
+    )
+
+def diff_param_values_by_id(left_id, right_id):
+    # Local import to reduce load time and
+    # avoid circular references.
```
astafan8: where do circular references come from?

qcodes/utils/metadata.py
```diff
@@ -42,3 +49,67 @@ def snapshot_base(self, update: bool=False,
...
+    from qcodes.dataset.data_set import load_by_id
+    """
+    Given the IDs of two datasets, returns the differences between
+    parameter values in each of their snapshots.
+    """
```
astafan8: Is there a reason for the docstring to be above the import?

qcodes/utils/metadata.py
```diff
@@ -42,3 +49,67 @@ def snapshot_base(self, update: bool=False,
...
+    Given a snapshot, returns a dictionary from
+    instrument and parameter names onto parameter values.
+    """
+    parameters = {}
+    for instrument_name, instrument in snapshot['station']['instruments'].items():
+        for param_name, param in instrument['parameters'].items():
+            if 'value' in param:
+                parameters[instrument_name, param_name] = param['value']
+
```
astafan8: a snapshot can contain independent (not bound to instruments) parameters in `snapshot['station']['parameters']`. is it worth covering them here as well?

qcodes/utils/metadata.py
```diff
@@ -1,6 +1,13 @@
+import json
+
 from typing import Sequence
 from .helpers import deep_update
+from functools import partial
 
+from typing import Dict, Tuple, Any, NamedTuple, Generic, TypeVar
+T = TypeVar('T')
+Snapshot = Dict[str, T] # TODO: represent known keys in typing
```
astafan8: that would be exciting! then this type definition can go to `qcodes/station.py` and be reused everywhere :)

qcodes/utils/metadata.py
```diff
@@ -42,3 +49,67 @@ def snapshot_base(self, update: bool=False,
...
+    """
+    parameters = {}
+    for instrument_name, instrument in snapshot['station']['instruments'].items():
+        for param_name, param in instrument['parameters'].items():
+            if 'value' in param:
+                parameters[instrument_name, param_name] = param['value']
+        
+    return parameters
+
+def diff_param_values(left_snapshot: Dict[str, Any],
```
astafan8: great! this is at least something that our users can use for working with snapshots...

qcodes/utils/metadata.py
cgranade: That was just a mistake, thanks for catching. Fixed!

qcodes/utils/metadata.py
cgranade: Changed to use that, thanks!

qcodes/utils/metadata.py
cgranade: I think it could be, yeah. I'll go on and modify accordingly, then.

qcodes/utils/metadata.py
cgranade: What I need to do is figure out how to make a Python type hint that accepts only `dict`s which have at least a given set of keys.

cgranade:
Looks like I've got a few Mypy issues to fix as well. Let me update the PR with those once I figure it out.

qcodes/dataset/measurements.py
```diff
@@ -826,7 +832,7 @@ def unregister_parameter(self,
         self.parameters.pop(param)
         log.info(f'Removed {param} from Measurement.')
```
Dominik-Vogel: @cgranade is it possible that you missed the `return self` here?

cgranade: My thought was that `unregister_parameter` is called in a fairly different way than the others, but I'm happy to add the `return self` here as well.

Dominik-Vogel: Sorry @cgranade, there was a code fold that I overlooked so my comment ended up at the wrong line. I meant to refer to line 802, `register_custom_parameter`. Here the type `T` was added as a return type, but it is not returned.

cgranade: Ah, got it. Yes, that was a mistake I've since fixed.

cgranade:
I have no idea how #1375 got mixed in with this one, sorry. Let me try to fix that.

qcodes/utils/metadata.py
cgranade: I forget, sorry, it was something several steps removed, though.

cgranade:
@astafan8, would you be willing to take another look? Thanks!

qcodes/dataset/measurements.py
```diff
@@ -862,9 +870,11 @@ def add_after_run(self, func: Callable, args: tuple) -> None:
 
         self.exitactions.append((func, args))
 
-    def add_subscriber(self,
+        return self
+
+    def add_subscriber(self : T,
```
astafan8: could you add (or modify existing) a test that uses (all at once) these methods (for no good reason) with the fluidity, just so that this behavior of the methods is captured?

qcodes/utils/metadata.py
astafan8: as discussed offline, let's keep it simple for now until qcodes drops support for python 3.6 and we can make the heavy use of data classes of python 3.7. but please remove the "todo" :)

qcodes/utils/metadata.py
```diff
@@ -42,3 +52,68 @@ def snapshot_base(self, update: bool=False,
...
+    for param_name, param in snapshot['station']['parameters'].items():
+        parameters[param_name] = param['value']
+    for instrument_name, instrument in snapshot['station']['instruments'].items():
+        for param_name, param in instrument['parameters'].items():
+            if 'value' in param:
+                parameters[instrument_name, param_name] = param['value']
+
+    return parameters
+
+def diff_param_values(left_snapshot: Dict[str, Any],
```
astafan8: Should the type hint be used here?
```suggestion
def diff_param_values(left_snapshot: Snapshot[Any],
```

astafan8: coulde you please make a note about these features in the `docs/examples/DataSet/Woking with snapshots.ipynb` notebook, please? :)

qcodes/utils/metadata.py
cgranade: Done!

qcodes/dataset/measurements.py
cgranade: On it.

qcodes/utils/metadata.py
cgranade: Added!

cgranade:
Thanks for the review, @astafan8!

astafan8:
Thanks for the example notebook! once CI passes, i will merge.

PR #1364, opened 2018-11-01
Bug: Return core.loglevel and file_loglevel to config

astafan8:
These are deprecated but still need to be there for backwards compatibility. The were accidentally removed in https://github.com/QCoDeS/Qcodes/pull/1309 that introduced new logging.

Without this, people who use other logging, like logging from qdev-wrappers, will experience exception when they update to the latest qcodes.


astafan8:
@QCoDeS/core Do you know why wasn't this discovered by our CI and/or by the fact that we have schema for our config?

jenshnielsen:
They are not listed as required in the schema so the config file is perfectly valid as far as the schema is concerned

PR #1365, opened 2018-11-01
Add Subscribers from config

Dominik-Vogel:
This PR enables the creation of subscribers for e.g. live plotting by adding them to the config.

There are quite many issue concerning the subscription (i.e. accessing data from different subscriber threads). These are not addressed here :-(.
Take a look at the sample notebook :-)


qcodes/dataset/data_set.py
```diff
@@ -756,6 +758,36 @@ def subscribe(self,
...
+            name: identifier of the subscriber. Equal to the key of the entry
+                in 'qcodesrc.json::subscription.subscribers'.
+        """
+        subscribers = qcodes.config.subscription.subscribers
+        try:
+            subscriber_info = getattr(subscribers, name)
+        except AttributeError:
+            keys = ','.join(subscribers.keys())
+            raise RuntimeError(
+                f'subscribe_from_config: failed to subscribe "{name}" DataSet '
```
astafan8:
```suggestion
                f'subscribe_from_config: failed to subscribe "{name}" to DataSet '
```

jenshnielsen:
Should the plottr link be to https://github.com/kouwenhovenlab/plottr or https://github.com/wpfff/plottr ?

qcodes/tests/dataset/test_subscribing.py
```diff
@@ -4,16 +4,49 @@
...
+
+from qcodes.tests.test_config import default_config
+
+
+log = logging.getLogger(__name__)
 
 VALUE = Union[str, Number, List, ndarray, bool]
 
 
+class TestSubscriber():
```
jenshnielsen: I would suggest not calling this class Test... since it's not actually a test. \

This works fine at the moment because pytest only discovers methods starting with `test` inside such a class and only if there is no __init__ method but it's still potentially confusing

qcodes/tests/test_config.py
```diff
@@ -135,46 +136,55 @@
...
 
+
 @contextmanager
-def default_config():
+def default_config(user_config: Optional[str]=None):
     """
     Context manager to temporarily establish default config settings.
     This is achieved by overwritting the config paths of the user-,
     environment-, and current directory-config files with the path of the
     config file in the qcodes repository.
```
astafan8: do empty strings assigned below to all the file name attributes mean the same as "the path of the config file in the qcodes repository"? I am not sure i'm following what's going on here but at least either the code change is too much or the docstring needs to be adjusted as well.

qcodes/dataset/measurements.py
```diff
@@ -72,6 +73,9 @@ def __init__(self, dataset: DataSet, write_period: numeric_types,
                                     callback_kwargs={'run_id':
                                                          self._dataset.run_id,
                                                      'snapshot': snapshot})
+        default_subscribers = qcodes.config.subscription.default_subscribers
+        for subscriber in default_subscribers:
+            self._dataset.subscribe_from_config(subscriber)
```
astafan8: nice and sneaky :)

qcodes/dataset/data_set.py
```diff
@@ -756,6 +758,36 @@ def subscribe(self,
...
+        Args:
+            name: identifier of the subscriber. Equal to the key of the entry
+                in 'qcodesrc.json::subscription.subscribers'.
+        """
+        subscribers = qcodes.config.subscription.subscribers
+        try:
+            subscriber_info = getattr(subscribers, name)
+        except AttributeError:
+            keys = ','.join(subscribers.keys())
+            raise RuntimeError(
```
astafan8: could you please add a test for this exception? :)

astafan8:
https://github.com/kouwenhovenlab/plottr

qcodes/tests/dataset/test_subscribing.py
Dominik-Vogel: good point!

qcodes/tests/test_config.py
Dominik-Vogel: Maybe these can be `None` but it is all good as long as it is not a filename. I chose the empty string to be on the safe side with typing.

qcodes/dataset/data_set.py
Dominik-Vogel: That was an impressive catch on your side: The dotdict that makes the config system possible does not throw an AttributeError here but a KeyError.

qcodes/tests/test_config.py
astafan8: ok, implementation is fine. My worry is about consistency of the docstring with the code. Now it claims smth that is not happening in the code, namely "overwritting the config paths of the user-, environment-, and current directory-config files __with the path of the config file in the qcodes repository__". What's happening now is "overwritting the config paths of the user-, environment-, and current directory-config files __with empty strings__", right?

qcodes/dataset/data_set.py
astafan8: it's not a catch, it's being (sometimes overly) "pedantic"

astafan8: great, thanks for adding the test! but why are we catching both types of exceptions?

qcodes/tests/dataset/test_subscribing.py
```diff
@@ -50,3 +83,95 @@ def test_basic_subscription(dataset, basic_subscriber):
...
+            }
+        }
+    }
+    """
+    db_location = qcodes.config.core.db_location
+    with default_config(user_config=config):
+        qcodes.config.core.db_location = db_location
+
+        assert 'test_subscriber' not in qcodes.config.subscription.subscribers
+        with pytest.raises(RuntimeError):
```
astafan8: hope that it's fine now, but I recommend next time to also assert the exception message because there are many RuntimeErrors while we are looking for a very specific one.

Dominik-Vogel: I agree. I was running out of time and the test should go in before a certain internal release yesterday.

PR #1366, opened 2018-11-01
MiniCircuits: wrong USB packet size resulting in error on Windows 7

### LINKED ISSUE #703: USB variable attenuator - Minicircuits RUDAT-13G-90
For blue force fridges

### LINKED ISSUE #1361: Minicircuits RUDAT 13G 90 driver does not work on Windows 7
Initiating a `RUDAT_13G_90` driver instance failed on the part where IDN is requested, and the instrument collects values for the parts of the IDN. The exception was: `('Error 1784 when trying to write to HID device: The supplied user buffer is not valid for the requested operation.', "asking ':ATT?' to <RUDAT_13G_90_USB: VA>", 'getting VA_attenuation')`.

The following code changes in the driver resolved the problem:
* change `packet_size` in `__init__` method of `MiniCircuitsHIDMixin` class from `64` to `63` [here](https://github.com/QCoDeS/Qcodes/blob/5398b4f1474106d55ee809153daf02b0dcb587a2/qcodes/instrument_drivers/Minicircuits/USBHIDMixin.py#L158)
* remove subtracting `1` from `packet_size` in `_unpack_string` method of `MiniCircuitsHIDMixin` class [here](https://github.com/QCoDeS/Qcodes/blob/5398b4f1474106d55ee809153daf02b0dcb587a2/qcodes/instrument_drivers/Minicircuits/USBHIDMixin.py#L192)

This means that basically the message size was incorrect.

It is important to mention that the PC where this behavior was observed has `Windows 7`. The original driver code works OK on a PC with `Windows 10` (e.g. __without__ making the 2 adjustments mentioned above). The version of `pywinusb` was the same, `0.4.2`.

Another test is going to be performed soon to find out what this bug really depends on: operating system, or smth else.

Originally discovered and fixed by @damazter.

astafan8:
Fixes #1361 (which was introduced in #1073). Also some cosmetic modifications (see commit messages).

Closes #703. (Shall it, @QCoDeS/core ?)

Note that a more elaborate driver is being developed in #813.


qcodes/instrument_drivers/Minicircuits/USBHIDMixin.py
```diff
@@ -19,32 +19,29 @@
...
             raise ImportError(
                 "pywinusb is not installed. Please install it by typing "
                 "'pip install pywinusb' in a qcodes environment terminal"
             )
 
-    def __init__(self, name, instance_id: str=None, timeout: float=2,
-                 **kwargs) ->None:
-
+    def __init__(self, name: str, instance_id: str=None, timeout: float=2,
+                 **kwargs):
```
Dominik-Vogel: wow you are so up to date. This was only in September:
https://github.com/python/mypy/pull/5677

qcodes/instrument_drivers/Minicircuits/USBHIDMixin.py
```diff
@@ -61,68 +58,83 @@ def __init__(self, name, instance_id: str=None, timeout: float=2,
...
+    def _unpack_string(self, response: bytes) -> str:
         raise NotImplementedError("Please subclass")
 
-    def write_raw(self, cmd: str) ->None:
+    def write_raw(self, cmd: str) -> None:
         """
-        Send binary data to the human interface device
+        Send a string command to the human interface device
+
+        The given command is processed by `_pack_string` method to return a
```
sohailc: I don't think this belongs in a docstring. Only put stuff in here which gives info on how to call a function, what it returns and what it does. The user does not care that the command is processed by a private method called `_pack_string`

qcodes/instrument_drivers/Minicircuits/USBHIDMixin.py
```diff
@@ -61,68 +58,83 @@ def __init__(self, name, instance_id: str=None, timeout: float=2,
...
+            raise RuntimeError(f"Communication with device failed for command "
+                               f"{cmd}")
 
-    def ask_raw(self, cmd: str) ->str:
+    def ask_raw(self, cmd: str) -> str:
         """
-        Send binary data to the human interface device and wait for a reply
+        Send a string command to the human interface device and wait for a reply
+
+        The given command is processed by `_pack_string` method to return a
```
sohailc: See above

sohailc:
Other then a small comment about docstrings, I think this is good to go

Dominik-Vogel:
The fix of the flaky test seems to have worked, all lights are green :-)

qcodes/instrument_drivers/Minicircuits/USBHIDMixin.py
astafan8: in this particular case, the end user of the drivers will not ever directly work with `write_raw`. However, a person who might be working on debugging the driver or adding features to it will look at `write_raw` and he/she will be probably glad to hear this extra info about `_pack_string` (although it is sort of clear from the code).

PR #1373, opened 2018-11-05
Driver dev/switch matrix

StefanD986:
Adds driver and example notebook for Keysight B2200 Femto leakage switch Matrix

- Example Notebook under docs/examples/driver_examples/Qcodes_example_with_Keysight_B2200_femto_leakage_switch_matrix.ipynb
- Driver under qcodes/instrument_drivers/Keysight/keysight_b220x.py
- Tests under qcodes/tests/drivers/test_keysight_b220x.py

@astafan8


qcodes/instrument_drivers/Keysight/keysight_b220x.py
```diff
@@ -0,0 +1,181 @@
...
+                          call_cmd='*CLS',
+                          docstring='Clears status register and error queue of the instrument.')
+
+        self.add_function(name='reset',
+                          call_cmd='*RST',
+                          docstring='Performs an instrument reset. Does not reset error queue!')
+
+        self.add_parameter(name='get_status',
+                           get_cmd='*ESR?',
+                           get_parser=lambda status_str: int(status_str),
```
sohailc: I think `get_parser=int` should work here.

sohailc:
Great work, @StefanD986! I would suggest: 
- Add type hints 
- Check that some lines of code are not overly long. I think the standard is 80 characters.

qcodes/instrument_drivers/Keysight/keysight_b220x.py
StefanD986: Oh yes of course 😅. I’ll change that.

jenshnielsen:
@StefanD986 Thanks for the contribution. There is a few small things you need to do to get the docs to build correctly. For this to work there are some requirements for the structure of the notebook. Please ensure that the notebook starts with a single level 1 heading (A single #) with the title of the notebook. All other titles should be level 2 (##) or 3 (###) taking care that level 3 headings only appear within level2 headings. Could you also rename the notebook file so that it is consistent with the other ones in https://github.com/QCoDeS/Qcodes/tree/master/docs/examples/driver_examples

docs/examples/driver_examples/Qcodes_example_with_Keysight_B2200_femto_leakage_switch_matrix.ipynb
```diff
@@ -0,0 +1,1262 @@
...
+     "text": [
+      "using physical instrument\n"
+     ]
+    },
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "c:\\users\\administrator\\pycharmprojects\\venvs\\qcodes_runtime\\lib\\site-packages\\pyvisa\\resources\\messagebased.py:569: UserWarning: read string doesn't end with termination characters\n",
+      "  return self.read()\n"
```
jenshnielsen: Do you have any idea why this is printed? Would be nice to fix this warning

qcodes/instrument_drivers/Keysight/keysight_b220x.py
```diff
@@ -0,0 +1,181 @@
...
+
+class KeysightB220X(VisaInstrument):
+    """
+    QCodes driver for B2200 / B2201 switch matrix
+    """
+
+    def __init__(self, name, address, **kwargs):
+        super().__init__(name, address, terminator='\r', **kwargs)
+
+        self.add_function(name='clear_status',
```
jenshnielsen: @WilliamHPNielsen What is the long term plan for functions?

astafan8: to deprecate them.

@StefanD986 , please, don't use functions, and create bound methods instead. Thanks!

qcodes/instrument_drivers/Keysight/keysight_b220x.py
```diff
@@ -0,0 +1,181 @@
+from qcodes import VisaInstrument
+from qcodes.utils.validators import MultiType, Ints, Enum, Lists
+
```
astafan8: > Caution: If the Free connection rule has been specified, ensure multiple input ports are not connected to the same output port. Such configurations can cause damage

Can we implement a check for this in the driver? And smart users can disable the check, while those who "don't know what they are doing" can have this check enabled which might save the live to their measured samples :)

astafan8: > Note: when switching from free to single mode, 1-to-multiple connections will not be changed until the connection is changed!!

Does it make sense to implement such a warning in the driver? Or implement that the change of the 1-to-multiple connections?

I assume it's not an instrument bug, but a feature, right?))

astafan8: Notebook:
> As explained above in Single Route Mode the switch matrix implicitly opens a previous connection when a new connections is made.

Is it a typo? should it be "... the switch matrix implicitly __closes__ a previous connection ..."?

astafan8: Notebook:
> Note: Everytime set_unused_inputs is called, the previous setting is overwritten!

typo: `unused_inputs.set(...)`

astafan8: Notebook:
For the ground port mode:
> Note: Using the port anyway will silently fail, that means it will not raise a python error. Only if you actively query the error queue of the instrument you will see the error.

Is it possible to implement such a check so that users see the error automagically?

astafan8: Notebook:
For the `couple_port_autodetect` example: for the good example, the output says `set()` while i guess it should be `{(3,4), (7,8)}` according to the comment in the code?

qcodes/instrument_drivers/Keysight/keysight_b220x.py
```diff
@@ -0,0 +1,181 @@
...
+                           get_cmd=':COUP? 0',
+                           set_cmd=':COUP 0,{}',
+                           val_mapping={True: 1,
+                                        False: 0},
+                           docstring="Param: True for ON, False for OFF"
+                           )
+
+
+    def _connect(self, channel1, channel2):
+        self.write(":CLOS (@{card:01d}{ch1:02d}{ch2:02d})".format(card=0,
```
astafan8: please make the `card` a private attribute at least, so that "just in case" it can be easily set to whatever.

(i think you told me but i forgot) Also, is it possible for a single instrument to have multiple cards? if so, then i'd at least add a note in the driver's docstring that "multi-card operation" is not supported.

qcodes/instrument_drivers/Keysight/keysight_b220x.py
```diff
@@ -0,0 +1,181 @@
...
+    def _disconnect(self, channel1, channel2):
+        self.write(":OPEN (@{card:01d}{ch1:02d}{ch2:02d})".format(card=0,
+                                                                  ch1=channel1,
+                                                                  ch2=channel2))
+
+    @staticmethod
+    def parse_channel_list(channel_list_str):
+        """Generates a set of (input, output) tuples from a SCPI channel list."""
+        pattern = r'(?P<card>\d{0,1}?)(?P<input>\d{1,2})(?P<output>\d{2})(?=(?:[,\)\r\n]|$))'
+        return {(int(match['input']), int(match['output'])) for match in re.finditer(pattern, channel_list_str)}
```
astafan8: please make all the lines not exceed 80 characters.

qcodes/tests/drivers/test_keysight_b220x.py
```diff
@@ -0,0 +1,216 @@
+import pytest
+from pyvisa.errors import VisaIOError
+from qcodes.instrument_drivers.Keysight.keysight_b220x import KeysightB220X
+
```
astafan8: in case users have an actual instrument connected, shall they be warned that before running the test they should disconnect their samples or smth?

qcodes/tests/drivers/test_keysight_b220x.py
```diff
@@ -0,0 +1,216 @@
...
+@pytest.fixture
+def uut() -> KeysightB220X:
+    try:
+        uut = KeysightB220X('switch_matrix',
+                            address='GPIB::22::INSTR')
```
astafan8: the test will not work if the address is incorrect. Shall we put a "garbage" address here so that the test will not get accidentally executed? and perhaps a nice error message that the user should enter a correct address here in order to run the tests on the real instrument?

qcodes/tests/drivers/test_keysight_b220x.py
```diff
@@ -0,0 +1,216 @@
...
+    uut.couple_port_autodetect()
+    assert 0 == uut.get_status()
+
+
+def test_get_error(uut):
+    uut.get_error()
+    assert 0 == uut.get_status()
+
+
+class Test_parse_channel_list:
```
astafan8: great!

qcodes/tests/drivers/test_keysight_b220x.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def test_idn_command(uut):
+    assert "AGILENT" in uut.IDN()['vendor']
+    assert 0 == uut.get_status()
+
+
+def test_connect(uut):
+    uut.connect(2, 48)
+    assert 0 == uut.get_status()
```
astafan8: in case the real instrument is connected, does it make sense to also assert its state? For example, after connecting these ports to assert the set of connections?

qcodes/instrument_drivers/Keysight/keysight_b220x.py
```diff
@@ -0,0 +1,181 @@
...
+                           )
+
+        self.add_function(name='connect',
+                          call_cmd=self._connect,
+                          args=[Ints(1, 14), Ints(1, 48)],
+                          docstring='connects given pair. Two arguments: input, output')
+
+        self.add_function(name='disconnect',
+                          call_cmd=self._disconnect,
+                          args=[Ints(1, 14), Ints(1, 48)],
```
astafan8: I tempt to recommend defining `Ints(1, 14)` as `self._available_input_ports` (or similar) and `Ints(1, 48)` as `self._available_output_ports` so that these definitions are re-used.

qcodes/instrument_drivers/Keysight/keysight_b220x.py
astafan8: I looked more carefully, and i see that the card number is in many commands. so, it's beneficial to put it an attribute because this will help when one will want to implement a driver for the multi-card case that is based on this driver.

qcodes/instrument_drivers/Keysight/keysight_b220x.py
StefanD986: Not a typo, but I can rephrase that to make it clearer what is happening. What I mean is that the previous connection is disconnected.

StefanD986: In principle yes, however I think none of the drivers automatically poll the status Byte of the instruments. So the behavior is not unique to this instrument. I just felt like this is an interesting case for the user to know.

In general i like the idea of polling instrument status, however since this behavior is common to all SCPI based instruments I’d say automatic status polling should be implemented in in one spot for all scpi instruments (some subclass of VisaInstrument maybe)

astafan8: @QCoDeS/core what's your opinion?

I, frankly, seen status polling in some particular drivers (but i don't remember where) in places where it is directly beneficial.

If you say that it's safer/better/whatever for the user to poll the status explicitly, then fine :)

qcodes/tests/drivers/test_keysight_b220x.py
StefanD986: I would love to do that, however this will only work with the real instrument. When using the pyvisa-sim simulated instrument it is not „smart enough“ to behave correctly. 

Of course I could work around that with some „if simulated:“ etc, but I thought it’s not really worth the effort. After all I mostly want to validate that the command is recognized as „valid“. I then have to trust in the manufacturer to react properly to the command.

astafan8: ok. i see.

qcodes/instrument_drivers/Keysight/keysight_b220x.py
astafan8: ah, not i get it. But it would still be great if you write this more explicitly.

StefanD986: I think I will add a warning that shows when initialising the instrument or when doing a reset.
But I am reluctant to add additional checks for that for the following reasons:
 - *Free* mode is the default behaviour after reset. So it seems to be deemed *safe enough* by Keysight.
 - The only way to add these checks is basically to poll the active connections after each command sent.
   - Users can still use the instrument front panel to override what the active connections and the driver will never know until it polls the connections again.
 - If I add checks for that, users might rely on checks to prevent "stupid mistakes". I'd rather have them pay attention themselves.

StefanD986: Let's say its an undocumented feature of the instrument 😜.

I will add a warning, but I will not implement a check or automatic release of the multiple connections (for similar reasons I wrote above in https://github.com/QCoDeS/Qcodes/pull/1373#discussion_r232925963)

StefanD986: Yes, I will rephrase that.

StefanD986: I will change that. (I renamed the function somewhere and forgot to update the description in the notebook.)

qcodes/instrument_drivers/Keysight/keysight_b220x.py
StefanD986: Good point. I will change that.

qcodes/instrument_drivers/Keysight/keysight_b220x.py
StefanD986: Good point. I will do that.

qcodes/tests/drivers/test_keysight_b220x.py
StefanD986: Done. changed address to `'insert_Keysight_B2200_VISA_resource_name_here'`

qcodes/instrument_drivers/Keysight/keysight_b220x.py
StefanD986: Good to know. I will change them to methods. (the add_function felt a bit cumbersome anyway)

qcodes/instrument_drivers/Keysight/keysight_b220x.py
StefanD986: Done ✅

qcodes/instrument_drivers/Keysight/keysight_b220x.py
StefanD986: Done. ✅ 
 - replaced the card parameter in all SCPI commands with a private variable self._card
 - added comment to driver doctoring that this driver only implements Auto mode.

qcodes/instrument_drivers/Keysight/keysight_b220x.py
StefanD986: Done ✅

qcodes/instrument_drivers/Keysight/keysight_b220x.py
astafan8: > I'd rather have them pay attention themselves.
can't agree more with that in this particular case :)

thanks for explanation

astafan8: ok, perfect!

qcodes/tests/drivers/test_keysight_b220x.py
jenshnielsen: Perhaps the tests should run using the simulated instrument by default. 
I am thinking just do

```
use_simulated=True
inst_addr = None
```
at the top and then connect to the simulated unless `use_simulated` is set to False and a valid address supplied

astafan8: yeah, that's much clearer.

qcodes/instrument_drivers/Keysight/keysight_b220x.py
StefanD986: Changed it to 
`As explained above in Single Route Mode the switch matrix implicitly disconnects the previous path when a new connections is made.`

StefanD986: changed to 
`Note: Everytime unused_inputs(<list of unused inputs>) is called, the previous setting is overwritten!`

docs/examples/driver_examples/Qcodes_example_with_Keysight_B2200_femto_leakage_switch_matrix.ipynb
StefanD986: I have to dig a bit more. I have seen it a couple of times after a reset, but its hard to reproduce.

qcodes/instrument_drivers/Keysight/keysight_b220x.py
StefanD986: I changed all `add_function` to methods.

qcodes/instrument_drivers/Keysight/keysight_b220x.py
StefanD986: Added a warning `When going from *free* to *single* mode existing connections are not released.`

qcodes/tests/drivers/test_keysight_b220x.py
StefanD986: As per @astafan8's suggestion in https://github.com/QCoDeS/Qcodes/pull/1373#discussion_r231457289, I changed the default physical instrument address to something that will require the user who wants to run the test with the physical instrument to actively change this address in the test code. Without that the test will always execute using the simulated instrument by default.

docs/examples/driver_examples/Qcodes_example_with_Keysight_B2200_femto_leakage_switch_matrix.ipynb
StefanD986: Looks like I used the wrong term char (`\r` instead of `\n`). I will change it.

qcodes/instrument_drivers/Keysight/keysight_b220x.py
StefanD986: I now added a decorator around the function calls that automatically clears the status byte *before* the function call, and polls the status byte *after* the function call.

A warning is emitted to the user if the status byte is != 0.

qcodes/instrument_drivers/Keysight/keysight_b220x.py
```diff
@@ -0,0 +1,358 @@
...
+                           get_cmd=':COUP:PORT? {card}'.format(card=self._card),
+                           set_cmd=":COUP:PORT {card},'{{}}'".format(
+                               card=self._card),
+                           set_parser=lambda value: str(value).strip('[]()'),
+                           get_parser=lambda response: [int(x) for x in
+                                                        response.strip(
+                                                            "'").split(',') if
+                                                        x.strip().isdigit()],
+                           vals=Lists(Enum(1, 3, 5, 7, 9, 11, 13))
+                           )
```
astafan8: `couple_ports` and `unused_inputs` have same `get_parser` and `set_parser`. What about making those reused private methods?

qcodes/instrument_drivers/Keysight/keysight_b220x.py
```diff
@@ -0,0 +1,358 @@
...
+    @post_execution_status_poll
+    def _get_connection_rule(self):
+        return self.ask(':CONN:RULE? {card}'.format(card=self._card))
+
+    @staticmethod
+    def parse_channel_list(channel_list: str):
+        """Generate a set of (input, output) tuples from a SCPI channel
+        list string.
+        """
+        pattern = r'(?P<card>\d{0,1}?)(?P<input>\d{1,2})(?P<output>\d{2})(?=(' \
```
astafan8: it's probably not a big deal here, but compiling this regexp outside of the method will increase its performance.

qcodes/tests/drivers/test_keysight_b220x.py
```diff
@@ -0,0 +1,280 @@
+import pytest
+from pyvisa.errors import VisaIOError
+from qcodes.instrument_drivers.Keysight.keysight_b220x import KeysightB220X
+import warnings
+
```
astafan8: this is just great :)

StefanD986:
Didn’t see any smoke. Go ahead.

PR #1376, opened 2018-11-07
Minicircuits USB_SPDT catch the correct error

jenshnielsen:
So that we actually print the error message when this fails. 

In addition fix an infinite (well up to  maximum recursion depth) error that will happen when closing the instrument and some docstring fixes

### DIFF
--- a/qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
+++ b/qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
@@ -6,6 +6,8 @@ from qcodes.instrument_drivers.Minicircuits.Base_SPDT import (
 
 try:
     import clr
+    clr.AddReference('System.IO')
+    from System.IO import FileNotFoundException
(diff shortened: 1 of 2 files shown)


qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
```diff
@@ -26,16 +29,22 @@ class USB_SPDT(SPDT_Base):
     Mini-Circuits SPDT RF switch
 
     Args:
-            name (str): the name of the instrument
-            serial_number (str, optional): the serial number of the device
+            name: the name of the instrument
+            driver_path: path to the dll
+            serial_number: the serial number of the device
                (printed on the sticker on the back side, without s/n)
             kwargs (dict): kwargs to be passed to Instrument class.
```
astafan8:
```suggestion
            kwargs: kwargs to be passed to Instrument class.
```

PR #1379, opened 2018-11-08
[WIP] Extract API calls in Alazar class + some async work

cgranade:
### NOTE: see comments below to understand the future of this PR.

This PR follows discussions with @astafan8, and extracts all C-level calls to the Alazar SDK into its own class to make it easier for that class to manage the use of background threads. To do so, this PR includes a metatype that converts a list of ctypes signatures into sync and async methods that both ensure that all calls to the Alazar SDK are executed in the same ThreadPoolExecutor. Refactoring in this way allows for the `acquire` method to be simplified, as ownership of threads and some memory management can be extracted. For example, work is in progress on this PR to allow for buffer allocation and posting to be done concurrently by using `asyncio.gather` to join after all buffers have been successfully posted. Similarly, extracting and unifying the sync/async APIs in this way allowed for making `find_boards_async()`, which may help speed up init scripts for stations which depend on Alazar initialization.

Still TODO:
- [ ] Docstrings and comments are still pretty sparse.
- [ ] More logic needs extracted from acquire to make it easier to reuse buffers across acquire calls.
- [ ] Testing
- [x] The DLL wrapper is still somewhat coupled to the Alazar usage, but could be generalized to provide thread-safe async wrappers for common C APIs.
- [x] The ctypes return values of metatype-time generated methods are saved as type annotations, but argument types are not due to complications with mypy's semantics for `*args`.
- [ ] Other methods like `get_idn` are still blocking, reducing parallelization of init scripts

### DIFF
--- a/qcodes/tests/drivers/test_MercuryiPS.py
+++ b/qcodes/tests/drivers/test_MercuryiPS.py
@@ -67,21 +67,6 @@ def test_simple_setting(driver):
     driver.GRPX.field_target(0.1)
     assert driver.GRPX.field_target() == 0.1
 
-def test_vector_setting(driver):
-    assert driver.field_target().distance(FieldVector(0, 0, 0)) <= 1e-8
-    driver.field_target(FieldVector(r=0.1, theta=0, phi=0))
-    assert driver.field_target().distance(
-        FieldVector(r=0.1, theta=0, phi=0)
-    ) <= 1e-8
-
-
-def test_vector_ramp_rate(driver):
-    driver.field_ramp_rate(FieldVector(0.1, 0.1, 0.1))
-    assert driver.field_ramp_rate().distance(
-        FieldVector(0.1, 0.1, 0.1)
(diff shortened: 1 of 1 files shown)


astafan8:
I'm going to continue working on this PR. (the field-vector related stuff does not belong to this PR, and will be removed from it)

astafan8:
I removed field-vector-related changes from this PR. Also resolved merge conflict which came due to changes in mercury IPS driver.

What's left in this PR is:
* Alazar API refactoring
* attempt to make ALazar driver async
* attempt to make mercury IPS driver async ( + test )
* async utils

All the async things will be done later in other PRs. The Alazar API refactoring will be continued, but yet in a different PR.

astafan8:
UPDATE: what's left in this PR:
* attempt to make ALazar driver async
* attempt to make mercury IPS driver async ( + test )
* async utils

All of these will be done later in other PRs.

This PR is left open for reference and history reasons.

PR #1380, opened 2018-11-09
SR830: Add increment/decrement sensitivity functions.

spxtr:
I often need to auto range lock-ins mid-sweep. I think it's usually best for instrument drivers to only act as instrument drivers, so I went with as small a change as possible to let me implement autoranging outside of the driver. It would have also been possible to add autoranging behavior into the driver itself, and if y'all prefer that then I'd be happy to update this PR.

The two functions in this change make it possible to do things like the following in my measurement sweeps:

```python
def autorange(srs, max_changes=1):
    def autorange_once():
        r = srs.R.get()
        sens = srs.sensitivity.get()
        if r > 0.9 * sens:
            return srs.increment_sensitivity()
        elif r < 0.1 * sens:
            return srs.decrement_sensitivity()
        return False

    sets = 0
    while autorange_once() and sets < max_changes:
        sets += 1
        time.sleep(srs.time_constant.get())
```

@StefanD986 @astafan8

### DIFF
--- a/qcodes/instrument_drivers/stanford_research/SR830.py
+++ b/qcodes/instrument_drivers/stanford_research/SR830.py
@@ -496,7 +496,8 @@ class SR830(VisaInstrument):
         return self._change_sensitivity(-1)
 
     def _change_sensitivity(self, dn):
-        n = int(self.ask('SENS?'))
+        _ = self.sensitivity.get()
+        n = self.sensitivity.raw_value
         if self.input_config() in ['a', 'a-b']:
             n_to = self._N_TO_VOLT
         else:
@@ -505,7 +506,7 @@ class SR830(VisaInstrument):
         if n + dn > max(n_to.keys()) or n + dn < min(n_to.keys()):
             return False
 
-        self.write(f'SENS {n + dn:d}')
+        self.sensitivity.set(n_to[n + dn])
         return True
 
     def _set_buffer_SR(self, SR):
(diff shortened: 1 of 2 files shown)


jenshnielsen:
@spxtr I think your idea is good and it makes a lot of sense to implement this and then make it possible for the user to implement the autorange. Would you be willing to document the feature in the example notebook here https://github.com/QCoDeS/Qcodes/blob/master/docs/examples/driver_examples/Qcodes%20example%20with%20Stanford%20SR830.ipynb

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -473,6 +473,41 @@ def parse_offset_get(s):
...
+        to pushing the sensitivity down button on the front panel. This has no
+        effect if the sensitivity is already at the minimum.
+
+        Returns:
+            Whether or not the sensitivity was actually changed.
+        """
+        return self._change_sensitivity(-1)
+
+    def _change_sensitivity(self, dn):
+        n = int(self.ask('SENS?'))
```
astafan8: I'd like to suggest reusing the `sensitivity` parameter, so that duplicated bare SCPI commands are not scattered around the driver (unless really needed).

So, in order to do that, what about the following:
```python
self.sensitivity.get()  # or self.sensitivity()
n = self.sensitivity.raw_value
```
and for setting the new sensitivity value below, just use the `n_to`:
```python
self.sensitivity.set(n_to(n + dn))  # or self.sensitivity(n_to(n + dn))
```
What do you think?

spxtr: Done. This does make sense, I didn't know the raw value was accessible in this way.

PR #1381, opened 2018-11-13
Driver/awg5014 upload lomentum sequence

Dominik-Vogel:
Adds the capability to use forged sequences from the lomentum pulse buidling library with the AWG5014.

This driver does not create any dependency but makes a conditional import.
@QCoDeS/core all tests are passing. I would like to merge this pr so folks can start using lomentum.


Dominik-Vogel:
@QCoDeS/core can we merge this driver? All tests are still passing, I tried it out with the new name of the pulse-building lib and the upload still works. This PR only introduces a new function that does not interfere with the rest of qcodes.

PR #1382, opened 2018-11-13
Improvement / dataset metadata

WilliamHPNielsen:
There is currently no way to query for the metadata associated with a particular dataset. One can only look up metadata by tag, which requires some auxiliary knowledge about the metadata.

- Add a `metadata` property to a `DataSet` object to see all associated metadata.

This little improvement is needed for #1352 in order to ensure correct copying of metadata.

@astafan8

### DIFF
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -24,7 +24,9 @@ from qcodes.dataset.sqlite_base import (atomic, atomic_transaction,
                                         VALUE, VALUES, get_data,
                                         get_values,
                                         get_setpoints,
-                                        get_metadata, one,
+                                        get_metadata,
+                                        get_metadata_from_run_id,
+                                        one,
                                         get_experiment_name_from_experiment_id,
                                         get_sample_name_from_experiment_id,
                                         get_guid_from_run_id,
@@ -231,6 +233,7 @@ class DataSet(Sized):
             self._completed = completed(self.conn, self.run_id)
             self._started = self.number_of_results > 0
             self._description = self._get_run_description_from_db()
+            self._metadata = get_metadata_from_run_id(self.conn, run_id)
 
         else:
             # Actually perform all the side effects needed for the creation
@@ -252,6 +255,7 @@ class DataSet(Sized):
             self._started = False
             specs = specs or []
             self._description = RunDescriber(InterDependencies(*specs))
+            self._metadata = get_metadata_from_run_id(self.conn, self.run_id)
 
     @property
     def run_id(self):
@@ -341,6 +345,10 @@ class DataSet(Sized):
     @property
     def description(self) -> RunDescriber:
         return self._description
+
+    @property
+    def metadata(self) -> Dict:
+        return self._metadata
 
     def run_timestamp(self, fmt: str="%Y-%m-%d %H:%M:%S") -> str:
         """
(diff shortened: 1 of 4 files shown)


WilliamHPNielsen:
Comment: you might worry why `snapshot` is missing from `RUNS_TABLE_COLUMNS`. See the test. I think I made the correct choice, but perhaps we should upgrade the schema later.

qcodes/dataset/sqlite_base.py
```diff
@@ -1904,6 +1911,39 @@ def get_metadata(conn: SomeConnection, tag: str, table_name: str):
                             "result_table_name", table_name)
 
 
+def get_metadata_from_run_id(conn: SomeConnection, run_id: int) -> Dict:
+    """
+    Get all metadata associated with the specified run
+    """
+    # TODO: promote snapshot to be present at creation time
```
astafan8: does this TODO mean upgrading the schema to have snapshot column in it?

qcodes/dataset/sqlite_base.py
```diff
@@ -1904,6 +1911,39 @@ def get_metadata(conn: SomeConnection, tag: str, table_name: str):
...
+        if row['name'] not in non_metadata:
+            possible_tags.append(row['name'])
+
+    # and then fetch whatever metadata the run might have
+    for tag in possible_tags:
+        query = f"""
+                SELECT "{tag}"
+                FROM runs
+                WHERE run_id = ?
+                AND "{tag}" IS NOT NULL
```
astafan8: i was trying to find when this extra condition would give problems, but couldn't.

qcodes/dataset/data_set.py
```diff
@@ -342,6 +346,10 @@ def run_timestamp_raw(self) -> float:
     def description(self) -> RunDescriber:
         return self._description
 
+    @property
+    def metadata(self) -> Dict:
```
astafan8: should metadata be added to the persistent traits list?

qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -526,3 +526,23 @@ def test_get_description(some_paramspecs):
     loaded_ds = DataSet(run_id=1)
 
     assert loaded_ds.description == desc
+
```
astafan8: not sure that i understood your comment about snapshot. so you've made it both: non metadata AND not RUNS_TABLE_COLUMNS, right? and both should be fixed with upgrading the schema when snapshot is a necessary column of the runs table, right? if so, i agree. We should also do smth with metadata later so that arbitrary columns do not get created for metadata tags.

qcodes/dataset/sqlite_base.py
astafan8: Will `DataSet.add_metadata({'mytag': None})` produce a new column `mytag` with a `NULL` value in it? if yes, then either such an `add_metadata` call should not be allowed, or `"{tag}" IS NOT NULL` part shall be removed.

qcodes/dataset/data_set.py
WilliamHPNielsen: Yes, but the persistent traits do not exist yet, only on another feature branch.

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: You are absolutely right. Since a single run with metadata `tag` creates the column and thereby implicitly NULL-populates it for all other runs, I think the sensible solution is to prohibit `None` as a valid metadata value. And if we do so, that prohibition should be enforced loud and clear.

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: Yes, exactly. I think it's just a blunder that it is not already there.

PR #1383, opened 2018-11-14
[WIP] DataStorageInterface and Sqlite DSI

Dominik-Vogel:
# Summary

Presents an interface to abstract over the storage media in `DataSet`.

Also includes its implementation for the SQLite storage backend.

### This is to be merged into a feature branch, not to the master. 
Edit: [I think this is now the feature branch, no?]

# List of most important changes being introduced:
* `DataStorageInterface` is defined to abstract storage and retrieval of datasets (runs)
* `SqliteStorageInterface` implements `DataStorageInterface` for direct local sqlite DB storage that is implemented directly in `DataSet` on master branch.
* Remove deprecated methods of `DataSet`: ...
* Do not allow to `add_results` (with 's') if `DataSet` is completed
* Introduce `pristine` and `running` properties of `DataSet` which correspond to its possible state, `started` property is removed (although an attribute remains)
* ...


qcodes/dataset/sqlite_storage_interface.py
```diff
@@ -0,0 +1,30 @@
...
+    DataStorageInterface, VALUES)
+from .sqlite_base import (
+    connect, select_one_where, insert_values, insert_many_values)
+from qcodes.dataset.database import get_DB_location
+
+
+class SqliteStorageInterface(DataStorageInterface):
+    """
+    """
+    def __init__(self, guid: str, ds):
```
jenshnielsen: The signature is not consistent with the ABC, should the ds be passed in the ABC too?
?

WilliamHPNielsen: I thought the storage interface instance should be an attribute of the dataset?

WilliamHPNielsen: This cyclic reference makes me uneasy. Can we not circumvent it?

astafan8: shall we fix the wrapping issue instead?

Dominik-Vogel: From my understanding of abstract base classes there is no need for the constructor of a child class to have the same call signature as the base, on the contrary it is most likely to be different. But I am happy to learn from your experience @jenshnielsen, it would be great if we could have a brief chat about this.

@WilliamHPNielsen, @astafan8, I agree that this is not a good solution here with the dataset being passed to the init. As you mentioned this is a quick and dirty fix for the issue I wrote about on Slack. My intention was to provide an interface that passes the basic tests and serves as a basis for the development of the other back-ends. As we have tight schedule we should first get our prototypes up and running before we fix the wrapping issue.

jenshnielsen: It not strictly disallowed but considered bad style and violates the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle)

astafan8:
Note that this branch now contains both `DataStorageInterface` and its implementation for sqlite.

WilliamHPNielsen:
Closed in favour of #1415

PR #1386, opened 2018-11-16
DataSet.get_data bugs with start/end arguments

### LINKED ISSUE #1384: DataSet.get_data: retrieving parts of dataset - actual start value depends on end
i encountered the following just now when trying to load only part of a dataset using get_data: the start index that's actually returned depends on whether the end index has been supplied or left as `None`.

example screenshot:

image

my expected behavior would be that the first element returned should not depend on the value of `end`.

astafan8:
Fixes #1384.

There are two bugs in the underlying `get_data` function of `sqlite_base.py`:
* `0` are treated as `None`
* sql query has a typo in the `where` clause for `start` for the case where both `start` and `end` are not `None`

This PR fixes both. 

Note that the current behavior suggests a `1`-based indexing (unlike python `0`-based), with both `start` and `end` indexes being _included_ into retuned data. This behavior is captured by the test suite.

Ideally, the very same test should be done for `get_data` from `sqlite_base.py` but did not make it to this PR in order to avoid code duplication.

Note that `get_data` from `sqlite_base.py` was also refactored to reduce the possibility for bugs like the ones mentioned above. In case it made the function less readable, the refactoring can be reverted.

### DIFF
--- a/qcodes/tests/dataset/test_dataset_basic.py
+++ b/qcodes/tests/dataset/test_dataset_basic.py
@@ -614,8 +614,7 @@ class TestGetData:
             (n_vals + 2, None, []),
 
             # test for end only
-            pytest.param(None, 0, [], marks=xfail(
-                reason="Now returns `xdata`, treats 0 as None")),
+            (None, 0, []),
             (None, 2, xdata[:2]),
             (None, -2, []),
             (None, n_vals, xdata),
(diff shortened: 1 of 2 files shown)


astafan8:

qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -1,6 +1,9 @@
 import itertools
 
 import pytest
+from pytest import mark
+
+xfail = mark.xfail
```
jenshnielsen: is this used?

qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -556,3 +558,61 @@ def test_metadata():
     with pytest.raises(ValueError, match=match):
         for tag, value in sorry_metadata.items():
             ds1.add_metadata(tag, value)
+
+
+class TestGetData:
+    x = ParamSpec("x", paramtype='numeric')
+    n_vals = 5
+    xvals = list(range(n_vals))
```
jenshnielsen: The list should be redundant when you only use this to iterate over?

astafan8: yes, you are right. But I am also trying to convey a message that if for whatever reason somebody wants to use `xvals = [345, 456, 22, 11, 4]`, they can (as long as they keep the value of `n_vals` in sync) :)

jenshnielsen: Ok

PR #1387, opened 2018-11-19
Fix unclarity about post_delay/inter_delay of Parameter, remove deprecated get/set delay and get/set step

astafan8:
`Parameter`'s `post_delay` and `inter_delay` properties have confusing descriptions. This PR addresses this.

Also remove deprecated `get_delay`, `set_delay`, `get_step`, `set_step`. Make `delay` kwarg a noop.

Discovered by @damazter

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -462,7 +462,7 @@ class _BaseParameter(Metadatable):
                     t_elapsed = time.perf_counter() - self._t_last_set
                     if t_elapsed < self.inter_delay:
                         # Sleep until time since last set is larger than
-                        # self.post_delay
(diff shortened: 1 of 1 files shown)


astafan8:
@jenshnielsen Do you mean `set_delay`/`get_delay`? Shall i do it in this PR?

Shall I also remove `set_step`/`get_step` ?

jenshnielsen:
Yes set/get delay and make the delay kwarg a noop. Fine to remove the get/set step too I think

jenshnielsen:
@astafan8 I think this looks great. If we are extra paranoid we could make the code raise if you pass delay as a kwarg but I will not insist that is needed

PR #1389, opened 2018-11-19
Fix bug where atomic and ConnectionPlus might not commit to database (+ fix DataSet.unsubscribe) (+ deprecate DataSet modify_results, modify_result, add_parameter_values)

astafan8:
`DataSet.<methods>` were reassigning `DataSet.conn` that initially is `sqlite3.Connection` to `ConnectionPlus`. Due to bugs in `atomic` and `ConnectionPlus`, this made operations on `DataSet` after using those methods __NOT commit__. For example, if you did `DataSet.modify_result`, and after that `DataSet.add_result`, then the last operation would not change the database file at all.

Bugs of `atomic` and `ConnectionPlus`:
- `atomic` wasn't reverting the original value of `atomic_in_progress` upon exit of the context manager
- `ConnectionPlus` had the default initial value of `atomic_in_progress` as `True` for an unknown reason

Main changes in this PR:
- [x] ~~Make `path_to_db` an attribute of `ConnectionPlus`~~ will be done in a separate PR
- [x] connect now always returns `ConnectionPlus`
- [x] `atomic` requires `ConnectionPlus`
- [x] `DataSet` uses only `ConnectionPlus` internally
- [x] All of qcodes uses only `ConnectionPlus`
- [x] Fix serious bug in `atomic` context manager and `ConnectionPlus` which prevents committing when `atomic` is called on an instance of `ConnectionPlus` (not `sqlite3.Connection`)
- [x] Fix serious bug in some `DataSet` methods which did not commit due to the abovementioned bug in `atomic` context manager and `ConnectionPlus`; the methods are `unsubscribe`, `unsubscribe_all`, `modify_result`, `modify_results`, `add_parameter_values`; also, do not redefine `self.conn` in these methods
- [x] Use `atomic` instead of explicit `connection.commit()` in production code
- [x] Cover `ConnectionPlus` and `atomic` with tests
- [x] Add extra tests for `connect` and some other related functions
- [x] Since in many tests we actually need to test that some code has indeed made modifications to the database file, it turns out that we have to do it by using another connection object (read more [here](https://docs.python.org/3.6/library/sqlite3.html#sqlite3.Connection.commit)); hence, some tests have been modified to do so.

Also (sorry, it is now not very easy to extract those in a separate PR):
- [x] `UPGRADE_ACTIONS` and `NEWEST_VERSION` are extracted from `perform_db_upgrade` to be used for correctly parametrizing tests
- [x] `DataSet.unsubscribe` now works (and is tested), it used to not work because database trigger id is different from the id of the `_Subscribe` instance (trigger has 'sub' prefix)
- [x] Deprecate `DataSet.modify_results` (with `s`), and add a minor test that shows that this method does not always work; also its signature and docstring are confusing, I could not make it work for multiple results; let's revisit this feature in future and make it work, if needed.
- [x] Deprecate `DataSet.add_parameter_results` because it does not work as the docstring suggests, see its test code for more information.
- [x] Minor fixes in how tests relying on actual db file fixtures are skipped.
- [x] Ignore .db fixture files with intended db versions that are generated for some tests


astafan8:
@QCoDeS/core ready to review! :)

WilliamHPNielsen:
>  Make everything a ConnectionPlus so that there is no need for connection upgrades.
What do you think?

I think that is indeed the best solution.

astafan8:
@QCoDeS/core I removed "SomeConnection", it's 'ConnectionPlus" now only. Could you have a last quick look?

PR #1392, opened 2018-11-22
Feature/dem 613/update s5i driver

qSaevar:
- Update qutech S5i driver

@astafan8

### DIFF
--- a/qcodes/instrument_drivers/QuTech/S5i.py
+++ b/qcodes/instrument_drivers/QuTech/S5i.py
@@ -14,43 +14,49 @@ class S5i(Instrument):
 
     Args:
         name (str): name of the instrument.
-
         spi_rack (SPI_rack): instance of the SPI_rack class as defined in
             the spirack package. This class manages communication with the
             individual modules.
-
         module (int): module number as set on the hardware.
     """
 
     def __init__(self, name, spi_rack, module, **kwargs):
         super().__init__(name, **kwargs)
 
-        self.s5i = S5i_module(spi_rack, module)
+        self.s5i = S5i_module(spi_rack, module, frequency=41e6, enable_output=1,
+                              output_level=0)
 
-        self.add_parameter('use_external_reference',
-                           label='Use external reference',
-                           get_cmd=self._use_external_reference,
-                           set_cmd=self.s5i.use_external_reference,
-                           vals=Bool())
+        self.add_parameter('enabled',
+                           label='RF_on/off',
+                           set_cmd=self.s5i.enable_output_soft,
+                           vals=Bool(),
+                           docstring='Switched output on/off')
 
         self.add_parameter('frequency_stepsize',
                            label='Frequency stepsize',
                            get_cmd=self._get_stepsize,
                            set_cmd=self.s5i.set_stepsize,
                            units='Hz',
-                           vals=Numbers())
+                           vals=Numbers(),
+                           docstring='Set the optimal frequency stepzise for '
+                                     'a minimal phase noise')
 
         self.add_parameter('frequency',
                            label='Frequency',
                            get_cmd=self._get_rf_frequency,
                            set_cmd=self.s5i.set_frequency,
                            units='Hz',
-                           vals=Numbers())
+                           vals=Numbers(40e6, 4e9),
+                           docstring='Set RF frequency, range 40MHz - 4GHz')
+
+        self.add_parameter('power',
+                           label='Output Power',
+                           set_cmd=self.s5i.set_output_power,
+                           unit='dBm',
+                           vals=Numbers(-14, 20),
+                           docstring='Set power in dBm, range -20 to + 15 dBm')
 
         self.add_function('optimize_for_frequency', call_cmd=self._optimize)
-
-    def _use_external_reference(self):
-        return self.s5i.use_external
 
     def _get_stepsize(self):
         return self.s5i.stepsize
@@ -59,4 +65,5 @@ class S5i(Instrument):
         return self.s5i.rf_frequency
 
     def _optimize(self):
-        self.s5i.set_frequency_optimally(self.s5i.rf_frequency)
+        stepsize = self.s5i.get_optimal_stepsize(self.s5i.rf_frequency)
+        self.s5i.set_stepsize(stepsize)
--- a/qcodes/instrument_drivers/QuTech/S5i.py
+++ b/qcodes/instrument_drivers/QuTech/S5i.py
@@ -47,14 +47,14 @@ class S5i(Instrument):
                            set_cmd=self.s5i.set_frequency,
                            units='Hz',
                            vals=Numbers(40e6, 4e9),
-                           docstring='Set RF frequency, range 40MHz - 4GHz')
+                           docstring='Set RF frequency, range 40MHz to 4GHz')
 
         self.add_parameter('power',
                            label='Output Power',
                            set_cmd=self.s5i.set_output_power,
                            unit='dBm',
                            vals=Numbers(-14, 20),
-                           docstring='Set power in dBm, range -20 to + 15 dBm')
+                           docstring='Set power in dBm, range -14 to +20 dBm')
 
         self.add_function('optimize_for_frequency', call_cmd=self._optimize)
 
--- a/qcodes/instrument_drivers/QuTech/S5i.py
+++ b/qcodes/instrument_drivers/QuTech/S5i.py
@@ -20,17 +20,20 @@ class S5i(Instrument):
         module (int): module number as set on the hardware.
     """
 
-    def __init__(self, name, spi_rack, module, **kwargs):
+    def __init__(self, name, spi_rack, module, frequency=41e6, enable_output=1,
+                 output_level=0, **kwargs):
         super().__init__(name, **kwargs)
 
-        self.s5i = S5i_module(spi_rack, module, frequency=41e6, enable_output=1,
-                              output_level=0)
+        self.s5i = S5i_module(spi_rack, module, frequency=frequency,
+                              enable_output=enable_output,
+                              output_level=output_level)
 
-        self.add_parameter('enabled',
-                           label='RF_on/off',
+        self.add_parameter('output_enabled',
+                           label='RF output enabled',
+                           initial_value=enable_output,
                            set_cmd=self.s5i.enable_output_soft,
                            vals=Bool(),
-                           docstring='Switched output on/off')
+                           docstring='Switches output on/off')
 
         self.add_parameter('frequency_stepsize',
                            label='Frequency stepsize',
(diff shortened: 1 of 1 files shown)


astafan8:
@qSaevar Could you briefly describe in the description what you did and why? (so that I don't ask gazillion of questions inside the code) for example, "removed parameter X because Y", etc. (and next time, don't commit all changes at once - instead, make one atomic change like "removed parameter X because Y", commit it, and only then commit a different change i.e. "added parameter Z because G")

qSaevar:
@astafan8 The underlying driver, `S5i_module`, has been updated and this PR is first and foremost addressing those changes. The `use_external_reference` method has been removed from the underlying driver, so I removed that parameter. Also the `set_frequency_optimally` was removed from `S5i_module` and replaced with `get_optimal_stepsize` so I updated `_optimize` accordingly.

But since I was changing the driver, I also added the `enabled` parameter that can be used to switch the module's output on and off. I added docstrings to the parameters and sharpened the `vals` to only allow what is possible in the underlying driver for `power` and `frequency`.

The updated driver has been tested on hardware.

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -14,43 +14,49 @@ class S5i(Instrument):
...
+                           docstring='Switched output on/off')
 
         self.add_parameter('frequency_stepsize',
                            label='Frequency stepsize',
                            get_cmd=self._get_stepsize,
                            set_cmd=self.s5i.set_stepsize,
                            units='Hz',
-                           vals=Numbers())
+                           vals=Numbers(),
+                           docstring='Set the optimal frequency stepzise for '
```
astafan8:
```suggestion
                           docstring='Set the optimal frequency stepsize for '
```

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -14,43 +14,49 @@ class S5i(Instrument):
...
-        self.s5i = S5i_module(spi_rack, module)
+        self.s5i = S5i_module(spi_rack, module, frequency=41e6, enable_output=1,
+                              output_level=0)
 
-        self.add_parameter('use_external_reference',
-                           label='Use external reference',
-                           get_cmd=self._use_external_reference,
-                           set_cmd=self.s5i.use_external_reference,
-                           vals=Bool())
+        self.add_parameter('enabled',
```
astafan8: Isn't this a better name? because `S5i.enabled(False)` looks like the whole instrument is disabled, while what's happening is `S5i.output_enabled(False)`.
```suggestion
        self.add_parameter('output_enabled',
```

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -14,43 +14,49 @@ class S5i(Instrument):
...
+                           label='RF_on/off',
```
astafan8:
```suggestion
                           label='RF output enabled',
```

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -14,43 +14,49 @@ class S5i(Instrument):
...
+                           set_cmd=self.s5i.enable_output_soft,
+                           vals=Bool(),
+                           docstring='Switched output on/off')
```
astafan8:
```suggestion
                           docstring='Switches output on/off')
```

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -14,43 +14,49 @@ class S5i(Instrument):
...
 
         self.add_parameter('frequency_stepsize',
                            label='Frequency stepsize',
                            get_cmd=self._get_stepsize,
                            set_cmd=self.s5i.set_stepsize,
                            units='Hz',
-                           vals=Numbers())
+                           vals=Numbers(),
```
astafan8: no limits here?

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -14,43 +14,49 @@ class S5i(Instrument):
...
+                                     'a minimal phase noise')
 
         self.add_parameter('frequency',
                            label='Frequency',
                            get_cmd=self._get_rf_frequency,
                            set_cmd=self.s5i.set_frequency,
                            units='Hz',
-                           vals=Numbers())
+                           vals=Numbers(40e6, 4e9),
+                           docstring='Set RF frequency, range 40MHz to 4GHz')
```
astafan8: There is no need to specify the allowed values in the docstring, because qcodes is smart enough to add this information automatically based on the value of `vals`. Try `s5i.frequency.__doc__` on an instance of "s5i" instrument.
```suggestion
                           docstring='Set RF frequency')
```

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -14,43 +14,49 @@ class S5i(Instrument):
...
 
-        self.add_function('optimize_for_frequency', call_cmd=self._optimize)
+        self.add_parameter('power',
+                           label='Output Power',
+                           set_cmd=self.s5i.set_output_power,
+                           unit='dBm',
+                           vals=Numbers(-14, 20),
+                           docstring='Set power in dBm, range -14 to +20 dBm')
```
astafan8: Same here, no need to specify range and unit. it's already there in the docstring.
```suggestion
                           docstring='Set output power')
```

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -14,43 +14,49 @@ class S5i(Instrument):
...
 
-    def _use_external_reference(self):
-        return self.s5i.use_external
+        self.add_function('optimize_for_frequency', call_cmd=self._optimize)
```
astafan8: please don't use functions. We are in the process of deprecating them (refer to the documentation about how to write an instrument driver for more info). Just use a bound method.

So either this:
```suggestion
    def optimize_for_frequency(self):
        """
        < some meaningful docstring here the explains the method :) >
        """
        self._optimize()
```

Or just rename `_optimize`, perhaps?

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -14,43 +14,49 @@ class S5i(Instrument):
...
-
         module (int): module number as set on the hardware.
     """
 
     def __init__(self, name, spi_rack, module, **kwargs):
         super().__init__(name, **kwargs)
 
-        self.s5i = S5i_module(spi_rack, module)
+        self.s5i = S5i_module(spi_rack, module, frequency=41e6, enable_output=1,
+                              output_level=0)
```
astafan8: these kwargs resemble the parameters below. 

* Shouldn't the parameters below _immediately_ reflect these values?
* Shouldn't these values be settable upon instantiation of `S5i` class? perhaps, in some time, the frequency will change, or you will _not_ want to enable output at the initialization.

Please, i really beg, keep things parameterizable/settable/adjustible, so that when one will need to change `41e6` in `frequency=41e6`, he will _not_ need to make a new PR against qcodes.

qcodes/instrument_drivers/QuTech/S5i.py
qSaevar: done

qcodes/instrument_drivers/QuTech/S5i.py
qSaevar: done

qcodes/instrument_drivers/QuTech/S5i.py
qSaevar: done

qcodes/instrument_drivers/QuTech/S5i.py
qSaevar: done

qcodes/instrument_drivers/QuTech/S5i.py
qSaevar: There are, but bit complicated as it needs to be integer division of reference frequency and below 1024.
`reference_frequency / stepsize < 1024`

qSaevar: It's checked in the underlying driver

qcodes/instrument_drivers/QuTech/S5i.py
qSaevar: done

qcodes/instrument_drivers/QuTech/S5i.py
qSaevar: done

qcodes/instrument_drivers/QuTech/S5i.py
qSaevar: done

qSaevar:
I addressed the comments, all but one, as setting limits on the stepsize is dependent on the set frequency.

qSaevar:
And for good measure we would like to test this driver again on hardware before it is merged.

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -14,49 +14,67 @@ class S5i(Instrument):
...
         spi_rack (SPI_rack): instance of the SPI_rack class as defined in
             the spirack package. This class manages communication with the
             individual modules.
-
         module (int): module number as set on the hardware.
     """
 
-    def __init__(self, name, spi_rack, module, **kwargs):
+    def __init__(self, name, spi_rack, module, frequency=41e6, enable_output=1,
+                 output_level=0, **kwargs):
```
astafan8: `enable_output` - why does it have an integer value? is it required by `S5i_module`? 

my suggestion would be to change the type to boolean, so that `enable_output=True` is passed to `__init__`, and then, if `S5i_module` requires that, pass `enable_output=int(enable_output)` to `S5i_module` constructor. As a bonus, this way the initial value for the "enable_output" parameter below will be set correctly.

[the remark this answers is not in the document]
qSaevar: done

qSaevar: done

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -14,49 +14,70 @@ class S5i(Instrument):
...
 
-        self.add_function('optimize_for_frequency', call_cmd=self._optimize)
+    def optimize_for_frequency(self):
+        """
+        This method finds the optimum stepsize for the set frequency.
 
-    def _use_external_reference(self):
-        return self.s5i.use_external
+        The stepsize affects the phase noise of the instrument. The smaller the
+        stepsize, the greater is the pahse noise. So this method sets the
```
eendebakpt:
```suggestion
        stepsize, the greater is the phase noise. So this method sets the
```

qcodes/instrument_drivers/QuTech/S5i.py
```diff
@@ -14,49 +14,70 @@ class S5i(Instrument):
...
+                           vals=Numbers(),
+                           docstring='Set the optimal frequency stepsize for '
+                                     'a minimal phase noise')
 
         self.add_parameter('frequency',
                            label='Frequency',
+                           initial_value=frequency,
                            get_cmd=self._get_rf_frequency,
                            set_cmd=self.s5i.set_frequency,
                            units='Hz',
```
eendebakpt: The `units` was deprecated, we should use `unit` instead. (also for one another parameter) @qSaevar

qSaevar: done

qcodes/instrument_drivers/QuTech/S5i.py
qSaevar: done

quantumkoen:
@astafan8 i have tested the latest changes on this branch against our setup, and can confirm that all is working as expected.

PR #1393, opened 2018-11-22
34465 a bugfix

jdwatson:
Removed set commands in ArrayMeasurement class of Keysight 34465A driver.  Existing class definition prevents the user from making changes to sampling parameters (i.e. it overwrites them).  Setting these type of parameters seems to belong in a measurement script rather than the driver.


jdwatson:
pinging @astafan8

WilliamHPNielsen:
Hi @jdwatson, thanks for bringing this up. While I agree that `ArrayMeasurement` is not general enough to be of use to most people (it was basically written specifically for a particular benchmark), I don't think we should abruptly change the behaviour of things. Some users might rely on `ArrayMeasurement` as it currently is. Instead, why don't we implement a more generic buffer parameter for this instrument?

PR #1395, opened 2018-11-26
Feature / let database show version without upgrading

WilliamHPNielsen:
Needed over at #1352 

- Add a little convenience function to get the DB version of a database and the latest available DB version

@QCoDeS/core

### DIFF
--- a/qcodes/dataset/sqlite_base.py
+++ b/qcodes/dataset/sqlite_base.py
@@ -692,6 +692,25 @@ def perform_db_upgrade_2_to_3(conn: SomeConnection) -> None:
             log.debug(f"Upgrade in transition, run number {run_id}: OK")
 
 
+def get_db_version_and_newest_available_version(path_to_db: str) -> Tuple[int,
+                                                                          int]:
+    """
+    Connect to a DB without performing any upgrades and get the version of
+    that database file along with the newest available version (the one that
+    a normal "connect" will automatically upgrade to)
+
+    Args:
+        path_to_db: the absolute path to the DB file
+
+    Returns:
+        A tuple of (db_version, latest_available_version)
+    """
+    conn = connect(path_to_db, version=0)
+    db_version = get_user_version(conn)
+
+    return (db_version, len(_UPGRADE_ACTIONS))
+
+
 def transaction(conn: SomeConnection,
                 sql: str, *args: Any) -> sqlite3.Cursor:
     """Perform a transaction.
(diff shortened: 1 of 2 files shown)


qcodes/dataset/sqlite_base.py
```diff
@@ -172,6 +180,8 @@ def do_upgrade(conn: SomeConnection) -> None:
         log.info(f'Succesfully performed upgrade {from_version} '
                  f'-> {to_version}')
 
+    _UPGRADE_ACTIONS[to_version] = do_upgrade
```
astafan8: brilliant!

qcodes/dataset/sqlite_base.py
```diff
@@ -354,16 +364,14 @@ def perform_db_upgrade(conn: SomeConnection, version: int=-1) -> None:
           'newest version'
     """
 
-    upgrade_actions = [perform_db_upgrade_0_to_1, perform_db_upgrade_1_to_2,
-                       perform_db_upgrade_2_to_3]
-    newest_version = len(upgrade_actions)
+    newest_version = len(_UPGRADE_ACTIONS)
```
astafan8: Shall this be extracted as well? well, at least it is extracted in #1389 in order to be used for testing.

qcodes/dataset/sqlite_base.py
```diff
@@ -684,6 +692,25 @@ def perform_db_upgrade_2_to_3(conn: SomeConnection) -> None:
             log.debug(f"Upgrade in transition, run number {run_id}: OK")
 
 
+def get_db_version_and_newest_available_version(path_to_db: str) -> Tuple[int,
```
astafan8: Why returning `newest_version` with a function? I'd suggest to put `newest_version` at the module level, no?

WilliamHPNielsen: This is actually mainly to avoid confusion with `get_user_version` which is quite different from this function.

As discussed offline, that is not so easy/pretty to put the newest version at the module level. Either it's a constant that gets evaluated at the bottom of the module (which is weirdish) or it's a function that returns `len(_UPGRADE_ACTIONS)`, which seems like overkill.

astafan8: agree. I'd just add one argument for the "overkill" option - it makes sense to have access to the newest version value without the need to `get_db_version` at the same time, for example, for reusing that value in tests. I'm gonna add the "overkill" option in #1389 :)

WilliamHPNielsen: I have no objections to that.

PR #1396, opened 2018-11-26
Add function to fix wrong descriptions

WilliamHPNielsen:
Background: "in the wild", we have encountered a database with corrupt `run_description`s.

- Add a function to fix wrong descriptions
- Add a test for it
- Generate version 3 database files, as a side effect of this PR

@astafan8

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -28,6 +28,9 @@ install:
 test_script:
   - cd qcodes
   - python tests\dataset\legacy_DB_generation\generate_version_0.py
+  - python tests\dataset\legacy_DB_generation\generate_version_1.py
+  - python tests\dataset\legacy_DB_generation\generate_version_2.py
+  - python tests\dataset\legacy_DB_generation\generate_version_3.py
   - cd tests
   - pytest
   - cd ..\..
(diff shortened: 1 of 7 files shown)


.appveyor.yml
```diff
@@ -28,6 +28,9 @@ install:
 test_script:
   - cd qcodes
   - python tests\dataset\legacy_DB_generation\generate_version_0.py
+  - python tests\dataset\legacy_DB_generation\generate_version_1.py
```
WilliamHPNielsen: Perhaps we don't need to keep the appveyor file up to date. I'd say it's safe to delete this file now.

qcodes/tests/dataset/legacy_DB_generation/generate_version_3.py
```diff
@@ -0,0 +1,139 @@
...
+                                         (params[4], z))
+
+        run_ids.append(datasaver.run_id)
+
+    assert [1, 2, 3] == run_ids, 'Run ids of generated runs are not as ' \
+                                 'expected after generating runs #1-3'
+
+    # Formulate SQL query for adjusting run_description column
+
+    set_run_description_sql = f"""
```
WilliamHPNielsen: Why not use the API ;P

.appveyor.yml
astafan8: in another PR somewhen in the future in the galaxy far, or now?

qcodes/tests/dataset/legacy_DB_generation/generate_version_3.py
astafan8: because it restricts what the `description` can be which is good for api but is not relevant here.

PR #1400, opened 2018-11-28
Add `get_parameter_data` aka `get_columns`

Dominik-Vogel:
This pr add a function to the sqlite_base module and also a method to the DataSet, that allows to retrieve data in a similar way as `get_data` does, but instead of returning a list of rows it provides a much more useful dictionary whose keys are the parameter names and whose entries are the columns of the respective parameter as numpy arrays.

I changed the name to `get_parameter_data` in order to not use data storage representation specific terms such as 'column'. 

This PR allows for an easy implementation of other export protocols.

It furthermore establishes the basis for a fast compiled implementation of the same.


qcodes/dataset/data_set.py
```diff
@@ -708,21 +733,41 @@ def get_data(self,
...
-        data = get_data(self.conn, self.table_name, valid_param_names,
+        valid_param_names = self._validate_parameters(*params)
+        return get_data(self.conn, self.table_name, valid_param_names,
+                        start, end)
+
+    def get_parameter_data(
+            self,
+            *params: Union[str, ParamSpec, _BaseParameter],
+            start: Optional[int] = None,
+            end: Optional[int] = None) -> Dict[str, numpy.ndarray]:
```
jenshnielsen: The method fails rather loudly if no parameters are supplied. (The same is the case for get_data) but never the less perhaps we should test for this

```
ERROR:qcodes.dataset.sqlite_base:Rolling back due to unhandled exception
  File "c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py", line 793, in atomic
    yield conn
  File "c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py", line 756, in atomic_transaction
    c = transaction(conn, sql, *args)
  File "c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py", line 733, in transaction
    c.execute(sql)
sqlite3.OperationalError: near "FROM": syntax error

OperationalError                          Traceback (most recent call last)
c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py in atomic(conn)
    792             conn.cursor().execute('BEGIN')
--> 793         yield conn
    794     except Exception as e:

c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py in atomic_transaction(conn, sql, *args)
    755     with atomic(conn) as conn:
--> 756         c = transaction(conn, sql, *args)
    757     return c

c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py in transaction(conn, sql, *args)
    732     else:
--> 733         c.execute(sql)
    734     return c

OperationalError: near "FROM": syntax error

The above exception was the direct cause of the following exception:

RuntimeError                              Traceback (most recent call last)
<ipython-input-12-4bf7cb4ff8cf> in <module>
----> 1 datasaver.dataset.get_parameter_data()

c:\users\jenielse\source\repos\qcodes\qcodes\dataset\data_set.py in get_parameter_data(self, start, end, *params)
    768         valid_param_names = self._validate_parameters(*params)
    769         return get_parameter_data(self.conn, self.table_name, valid_param_names,
--> 770                         start, end)
    771 
    772     def get_values(self, param_name: str) -> List[List[Any]]:

c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py in get_parameter_data(conn, table_name, columns, start, end)
   1161     # This method is going to form the entry point for a compiled version
   1162 
-> 1163     res = get_data(conn, table_name, columns, start, end)
   1164     res_t = list(map(list, zip(*res)))
   1165     return {name: np.array(column_data)

c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py in get_data(conn, table_name, columns, start, end)
   1132 
   1133     query = _build_data_query(table_name, columns, start, end)
-> 1134     c = atomic_transaction(conn, query)
   1135     res = many_many(c, *columns)
   1136 

c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py in atomic_transaction(conn, sql, *args)
    754     """
    755     with atomic(conn) as conn:
--> 756         c = transaction(conn, sql, *args)
    757     return c
    758 

~\AppData\Local\Continuum\miniconda3\envs\qcodes\lib\contextlib.py in __exit__(self, type, value, traceback)
     97                 value = type()
     98             try:
---> 99                 self.gen.throw(type, value, traceback)
    100             except StopIteration as exc:
    101                 # Suppress StopIteration *unless* it's the same exception that

c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py in atomic(conn)
    795         conn.rollback()
    796         log.exception("Rolling back due to unhandled exception")
--> 797         raise RuntimeError("Rolling back due to unhandled exception") from e
    798     else:
    799         if is_outmost:

RuntimeError: Rolling back due to unhandled exception


```

Dominik-Vogel:
@astafan8 thank you! 
1) It will not be completely substituted, but it will rather form the entry point. Calling it results in looking up the presence of the compiled code. The query will still be build as is and then instead of executing it on python it will be executed from the compiled library. No,`get_data` will not be substituted. This is because getting the data in form of a dictionary of columns seems a lot more natural. 
2) Yes, it does ;-) As far as I can see there is no way in python to implement such a query faster. I have benchmarked some ideas but could not get any improvement.

jenshnielsen:
I did a bunch of manual tests with combinations of array type and regular parameters. They seem to work great. I will transform them into regular tests. @Dominik-Vogel is it ok to push here?

Dominik-Vogel:
yes feel free @jenshnielsen !

qcodes/dataset/data_set.py
```diff
@@ -708,21 +734,45 @@ def get_data(self,
...
-            else:
-                try:
-                    maybeParam = maybeParam.name
-                except Exception as e:
-                    raise ValueError(
-                        "This parameter does not have  a name") from e
-            valid_param_names.append(maybeParam)
-        data = get_data(self.conn, self.table_name, valid_param_names,
+        valid_param_names = self._validate_parameters(*params)
+        if len(valid_param_names) == 0:
```
jenshnielsen: This check does not catch the error in `get_parameter_data`, would it not be better to move this into the `sqlite_base` `get_data` ?

Dominik-Vogel: ok now I see, sorry, yes it is indeed better to have it in `sqlite_base`.

Dominik-Vogel: I will move it and let the function return an empty list plus issue a warning.

jenshnielsen:
Started adding function that gets the data and it's dependencies and automatically expands scalar data to match arrays. 

TODO
- [ ] Tests
- [ ] Support for scalar strings
- [ ] Connect this to datwriter/pandas/xarray
- [ ] A less stupid name for the function

jenshnielsen:
I think this is ready. Working on xarray/pandas on the next branch

qcodes/tests/dataset/helper_functions.py
```diff
@@ -0,0 +1,33 @@
+from typing import Sequence, Tuple, Dict
+
+import numpy as np
+
+
+def verify_data_dict(data: Dict[str, Dict[str, np.ndarray]],
```
astafan8: I might be missing a point here but should this function do more checks? like if names of the `expected_names[param]` are exactly the keys of the inner dicts?

qcodes/tests/dataset/helper_functions.py
```diff
@@ -0,0 +1,33 @@
...
+                     parameter_names: Sequence[str],
+                     expected_names: Dict[str, Sequence[str]],
+                     expected_shapes: Dict[str, Sequence[Tuple[int, ...]]]) -> None:
+    """
+    Simple helper function to verify a dict of data
+
+    Args:
+        data: The data to verify
+        parameter_names: names of the parameters requested
```
astafan8: this is a bit confusing. "requested" for "checking", right? could you rephrase it please?

qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -767,3 +773,130 @@ def ds_with_vals(self, dataset):
...
+                          expected_toplevel_params,
+                          expected_names,
+                          expected_shapes)
+
+
+def parameter_test_helper(ds, toplevel_names, expected_names, expected_shapes):
+    data = ds.get_parameter_data(*toplevel_names)
+    all_data = ds.get_parameter_data()
+
+    assert data.keys() == all_data.keys()
```
astafan8: this is a weird helper function because it only works for the cases where `toplevel_names` for a given `ds` happen to coincide with the names of non-dependencies (which are standalone + dependent parameters). I suggest to make the helper more generic (and perhaps also add tests where `set(data.keys()).is_subset_of(set(all_data.keys()))`?). Otherwise, this implicit assumption should be mentioned in comments.

qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -767,3 +773,130 @@ def ds_with_vals(self, dataset):
     def test_get_data_with_start_and_end_args(self, ds_with_vals,
                                               start, end, expected):
         assert expected == ds_with_vals.get_data(self.x, start=start, end=end)
+
+
```
astafan8: Why do the tests have not have assertions on the actual data being retrieved (and it's type)? Is it because we just trust `get_data`? also no tests cover the start/end feature.

qcodes/tests/dataset/dataset_fixtures.py
```diff
@@ -0,0 +1,129 @@
...
+                                       label=f'param_{n_params}',
+                                       unit='Ohm',
+                                       depends_on=params_indep)]
+    for p in params:
+        dataset.add_parameter(p)
+    dataset.add_results([{p.name: np.random.rand(1)[0] for p in params}
+                         for _ in range(n_rows)])
+    dataset.mark_complete()
+    yield dataset
+
```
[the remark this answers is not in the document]
jenshnielsen: I don't think so. This is done further up the chain. This fixture does not create a new connection

qcodes/tests/dataset/dataset_fixtures.py
```diff
@@ -0,0 +1,129 @@
...
+                                       'numeric',
+                                       label=f'param_{n_params}',
+                                       unit='Ohm',
+                                       depends_on=params_indep[0:1])]
+    for p in params:
+        dataset.add_parameter(p)
+    dataset.add_results([{p.name: np.random.rand(1)[0] for p in params}
+                         for _ in range(n_rows)])
+    dataset.mark_complete()
+    yield dataset
```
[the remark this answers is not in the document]
jenshnielsen: again this is closed further up the chain

qcodes/dataset/sqlite_base.py
```diff
@@ -1182,27 +1207,85 @@ def get_data(conn: ConnectionPlus,
...
+            numeric_elms = [i for i, x in enumerate(types)
+                            if x == "numeric"]
+            text_elms = [i for i, x in enumerate(types)
+                         if x == "text"]
+            for row in res:
+                for element in numeric_elms:
+                    row[element] = np.full_like(row[first_array_element],
+                                                row[element],
+                                                dtype=np.float)
+                    # todo should we handle int/float types here
```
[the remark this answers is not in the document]
jenshnielsen: The problem are that they are not correct! Consider the case where you store a [0.0, 0.1, 0.2, 0.3 ...] in a row. The first element will be converted to an int but the rest of the row will remain as floats! The datatype of all elements in a row should be the same

qcodes/dataset/sqlite_base.py
```diff
@@ -1182,27 +1207,85 @@ def get_data(conn: ConnectionPlus,
...
+                for element in text_elms:
+                    strlen = len(row[element])
+                    row[element] = np.full_like(row[first_array_element],
+                                                row[element],
+                                                dtype=f'U{strlen}')
+
+        # Benchmarking shows that transposing the data with python types is
+        # faster than transposing the data using np.array.transpose
+        # This method is going to form the entry point for a compiled version
```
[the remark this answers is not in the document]
jenshnielsen: This is @Dominik-Vogel original comment. It's very clear to me that this method referees to the method that this is in but lets just remove it for now as it adds little value

qcodes/dataset/sqlite_base.py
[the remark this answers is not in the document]
astafan8: "in a row" you mean in the same table row? to my knowledge this is only possible via "array" paramtype, hence the np.float type will be preserved.

if one stores these numbers in a row, one after another, using "numeric" paramtype then, indeed, our adapters will do what you said, which is incorrect. you are right about this.

Then, let's stay with floats for now? and i think a note in a docstring is still of value.

qcodes/tests/dataset/dataset_fixtures.py
```diff
@@ -0,0 +1,129 @@
...
+        yield datasaver.dataset
+    finally:
+        datasaver.dataset.conn.close()
+
+
+@pytest.fixture
+def standalone_parameters_dataset(dataset):
+    n_params = 3
+    n_rows = 10**3
+    params_indep = [ParamSpec(f'param_{i}',
```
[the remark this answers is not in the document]
jenshnielsen: I am not sure I understand. Part of the issue with this is that there is no way to tell from a parameter alone if its standalone or a dependency of another parameter. This is reflected in the tests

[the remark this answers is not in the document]
astafan8: its ok

jenshnielsen:
@astafan8 I think this addresses all the issues you found

PR #1402, opened 2018-11-28
Various IPS120 driver improvements

spxtr:
First I make it support both GPIB and serial. Next, fix some bugs in the driver. I haven't yet tested this on my own machine, will do so soon.

@jenshnielsen


qcodes/instrument_drivers/oxford/IPS120.py
```diff
@@ -19,32 +19,42 @@
...
+
+
+    def __init__(self, name, address, use_gpib=False, number=2, **kwargs):
         """Initializes the Oxford Instruments IPS 120 Magnet Power Supply.
 
         Args:
             name (string)    : name of the instrument
             address (string) : instrument address
-            number (int)     : ISOBUS instrument number
+            use_gpib (bool)  : whether to use GPIB or serial
```
jenshnielsen: I am not that familiar with the IPS120. Are GPIB and serial the only 2 supported ways of communicating. The Mercury also supports comm over ethernet which is similar to the gpib (so that the serial) is the odd one out. But I am not sure if that applies to this one too?

qcodes/instrument_drivers/oxford/IPS120.py
```diff
@@ -292,11 +310,8 @@ def _set_remote_status(self, mode):
             2: "Local and unlocked",
             3: "Remote and unlocked",
         }
-        if status.__contains__(mode):
-            log.info('Setting remote control status to %s' %
-                status.get(
-                    mode,
-                    "Unknown"))
+        if mode in status:
+            log.info('Setting remote control status to %s' % status[mode])
```
jenshnielsen: I think this should be `self.log` See https://qcodes.github.io/Qcodes/examples/Creating%20Instrument%20Drivers.html#Logging

qcodes/instrument_drivers/oxford/IPS120.py
spxtr: Only these two.

qcodes/instrument_drivers/oxford/IPS120.py
spxtr: Done.

spxtr:
OK, I made the driver use `self.log` instead of log, pulled status dictionaries out, and tidied some random things up. Please let me know how else I can improve it.

I have tested something very similar to this on a real machine, but not this exact change, so please let me test it again before merging.

qcodes/instrument_drivers/oxford/IPS120.py
jenshnielsen: Cool then this is fine

qcodes/instrument_drivers/oxford/IPS120.py
```diff
@@ -894,36 +884,9 @@ def _get_polarity(self):
...
-        status1 = {
-            0: "Desired: Positive, Magnet: Positive, Commanded: Positive",
-            1: "Desired: Positive, Magnet: Positive, Commanded: Negative",
-            2: "Desired: Positive, Magnet: Negative, Commanded: Positive",
-            3: "Desired: Positive, Magnet: Negative, Commanded: Negative",
-            4: "Desired: Negative, Magnet: Positive, Commanded: Positive",
-            5: "Desired: Negative, Magnet: Positive, Commanded: Negative",
-            6: "Desired: Negative, Magnet: Negative, Commanded: Positive",
-            7: "Desired: Negative, Magnet: Negative, Commanded: Negative"
         }
```
jenshnielsen: Looks like these 2 lines should be deleted too

qcodes/instrument_drivers/oxford/IPS120.py
```diff
@@ -19,32 +19,87 @@
...
-
-    Note: Since the ISOBUS allows for several instruments to be managed in parallel, the command
-    which is sent to the device starts with '@n', where n is the ISOBUS instrument number.
+    The IPS 120 can connect through both RS232 serial as well as GPIB. The
+    commands sent in both cases are similar. When using the serial connection,
+    commands are prefaced with '@n' where n is the ISOBUS number.
     """
 
-    def __init__(self, name, address, number=2, **kwargs):
+    _GET_STATUS_MODE = = {
```
jenshnielsen: An extra `=` has crept in here

jenshnielsen:
It looks like there are a few small syntax errors. Other than than I think this is ready to merge once you confirm that it has been tested on a real device

jenshnielsen:
@spxtr Did you ever test this on a real device

spxtr:
We've warmed up the system for maintenance, once we cool it back down I'll test this. Should be in a week or two.

qcodes/instrument_drivers/oxford/IPS120.py
```diff
@@ -144,31 +199,37 @@ def __init__(self, name, address, number=2, **kwargs):
...
         """
-        log.info('Send the following command to the device: %s' %
-            message)
+        self.log.info('Send the following command to the device: %s' % message)
+
+        if self._use_gpib:
+            return self.ask(message)
+
         self.visa_handle.write('@%s%s' % (self._number, message))
         sleep(70e-3)  # wait for the device to be able to respond
```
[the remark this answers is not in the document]
spxtr: Sure.

qcodes/instrument_drivers/oxford/IPS120.py
```diff
@@ -664,35 +690,22 @@ def _get_switch_heater(self):
...
         Set the switch heater Off or On. Note: After issuing a command it is necessary to wait
         several seconds for the switch to respond.
         Args:
             mode (int) :
-            0 : "Off"
-            1 : "On, if PSU = Magnet"
-            2 : "On, No checks" (not available)
+            0 : Off
+            1 : On, if PSU = Magnet
+            2 : On, No checks (not available)
```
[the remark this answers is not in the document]
spxtr: Removed them.

qcodes/instrument_drivers/oxford/IPS120.py
```diff
@@ -658,41 +687,23 @@ def _get_switch_heater(self):
...
-            0: "Off",
-            1: "On, if PSU = Magnet"
-        }
-        if status.__contains__(mode):
-            log.info('Setting switch heater to %s' %
-                status.get(
-                    mode,
-                    "Unknown"))
+            0 : Off
+            1 : On
```
astafan8: do i understand correctly that `switch_heater` parameter has two roles - on get it returns one of the 5 statuses (as mentioned in `_GET_STATUS_SWITCH_HEATER `), but it is possible to set the switch only to `on` or `off`, meaning to `0` or `1` and not to `2`, `5`, or `8` -- is this correct? If so, this is quite dirty, but since users have been using it this way and we know of no complaints, let's keep it like this.

qcodes/instrument_drivers/oxford/IPS120.py
```diff
@@ -707,13 +718,11 @@ def heater_on(self):
         """Switch the heater on, with PSU = Magnet current check"""
         current_in_magnet = self.persistent_current()
         current_in_leads = self.current()
-        if self.switch_heater() == 1:
```
astafan8: i'm confused of how this could even ever work. `switch_heater` parameter, when called `get` on, returns a "string". Before this PR, that string was compared against an integer value of `1` - which is incorrect. Now, thanks to the change in this PR, the "string" coming from `switch_heater()` call gets compared against another string that now comes from `_GET_STATUS_SWITCH_HEATER` (for "index" `1`) - which is correct. Am i mistaken somewhere or just `heater_on` was never used by anyone in this world?

spxtr: You are correct, it never worked.

qcodes/instrument_drivers/oxford/IPS120.py
spxtr: You are correct, it's not pretty.

spxtr:
Yes, I tested this on a real machine. At least the basic functions like manipulating the switch heater, running to fields, and entering/leaving persistent mode worked.

qcodes/instrument_drivers/oxford/IPS120.py
astafan8: awesome! thanks to refactoring, it all works now :)

PR #1404, opened 2018-11-29
Use log instead of print in AWG70000A driver

astafan8:
Substitute the annoying `print` statements in the driver with `log.debug`. Logging can be easily enabled when needed, qcodes has convenient infrastructure as well.


jenshnielsen:
Are we not supposed to use `self.log` following @Dominik-Vogel s changes

astafan8:
@Dominik-Vogel @jenshnielsen now it's done the right way :) thanks for pointing out!

PR #1409, opened 2018-12-04
Driver/keithley s46

sohailc:
Adds a driver for the S46 RF switch


jenshnielsen:
Looks good, Rather than using your own caching i think you should consider using the caching build in to the parameter already via the get_latest function. If you look at the QDac driver https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/QDev/QDac_channels.py you can see that this has a similar function that returns the full status of all the channels. This is called in `_get_status` exactly once when initialized and then updates the value of all the voltages on all the channels. After that the snapshot and other functions that needs the value of all channels can simply ensure that `_get_status` is called exactly once before using all the values

qcodes/instrument_drivers/tektronix/Keithley_s46.py
```diff
@@ -0,0 +1,225 @@
...
+        return [
+            channel for channel in self.channels if
+            channel.channel_number in self.get_closed_channel_numbers()
+        ]
+
+    def open_all_channels(self) -> None:
+        """
+        Please do not write ':OPEN ALL' to the instrument as this will
+        circumvent the lock.
+        """
```
jenshnielsen: Is this slow compared to `:OPEN ALL`? Is that an issue or perhaps it doesn't matter?

qcodes/tests/drivers/test_keithley_s46.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    A version of the driver which logs every ask command. We need this to
+    assert that ':CLOS?' is called once during initialization
+    """
+    def __init__(self, *args, **kwargs):
+        self._ask_log = []
+        super().__init__(*args, **kwargs)
+
+    def ask(self, cmd: str):
+        self._ask_log.append(cmd)
```
jenshnielsen: You could probably do this with the existing logging in the instrument and pytests log capturing functions

qcodes/instrument_drivers/tektronix/Keithley_s46.py
sohailc: Its probably slower, but I do not think it matters.

PR #1413, opened 2018-12-05
val_mapping creator function to solve on/off/true/false issue

### LINKED ISSUE #924: on/off/0/1/True/False
I understand it's been hard to enforce a convention for this and it is different across drivers, but could we at least converge on a recommended one and enforce that as much as possible? 

Possibly this would be at least partly resolved (and also make it backwards compatible) to have a parser so that at least setting 'True' or 'on' or '1' would all map to the expected instrument setting . I understand that what the get function should return is a bit trickier (as I would argue for 1 and 0 but I also realise that could be problematic if an instrument has more settings than on and off; maybe we could make a decision on this and then enforce it though, can't be worse than the mix of things that currently exists?)

nataliejpg:
Fixes #924

I made a function which given the instrument values for 'on' and 'off' maps all the obvious inputs (1, True, 'On', 'on', 'ON') to on and off. The inverse always maps back to booleans which I chose relatively arbitrarily as 'the most sensible' to be the convention. Happy to extend it if more are voted for. 

Also, Implemented in the rhode schwarz SGS100A for now but I imagine other drivers would also benefit from it.
 
TODO:
- [x] Test
- [x] Add a note about it to driver creating notebook doc
- [ ] Use in other drivers?


qcodes/utils/helpers.py
```diff
@@ -666,6 +667,23 @@ def inner(**inner_kwargs):
...
+    return OrderedDict([('On', on_val),
+                       ('ON', on_val),
+                       ('on', on_val),
+                       ('1', on_val),
+                       (True, on_val),
+                       ('Off', off_val),
+                       ('OFF', off_val),
+                       ('off', off_val),
+                       ('0', off_val),
+                       (False, off_val)])
```
astafan8: you state that the inverse gets mapped to booleans. Is it ensured just by the fact that the True/False values are the latter ones in the lists corresponding to on_val and off_val? I'd love to have this ensured in a more strict way.

the minimal thing that can be done to achieve this is to extract as a function the code that creates the inverse val_mapping inside the `Parameter`, say `create_inverse_val_mapping`, and add a test that ensures that `create_inverse_val_mapping(create_on_off_val_mapping(on_val='on_value', off_val='off_value')) == {True: 'on_value', False: 'off_value'}`. Do you think that would suffice, @QCoDeS/core ?

jenshnielsen: Yes this seems reasonable to add a test for this behavior.  
The inverse val mapping does indeed just iterate the key value pairs
so the last pair will always overwrite any previous definition. I think a test is enough. 
My primary concern would be that someone adds a new pair to the dict to handle 'oN' or some such 
thing and does not realize that it should not be added to the end. That is prevented by a test.

astafan8:
I've refactored the code, and added tests.

Questions:
* What other "sensible" on/off inputs we need?
* Do we also need enabled/disabled val mapping? Or shall that case be strongly recommended to be True/False and parameter should always be called `smth_enabled` so that the API is `smth_enabled(True)`?

jenshnielsen:
* I think this is fine
* I vote for leaving that out for now

astafan8:

Question: shall i also go through all drivers quickly and use this new function in cases where it makes sense and does not break instrument-driver's API?

astafan8:
@nataliejpg as the original author of the PR and the main feature requester, could you also have a look? :)

nataliejpg:
looks good to me @astafan8, thanks for the work. I didn't understand your question:

>  Do we also need enabled/disabled val mapping? Or shall that case be strongly recommended to be True/False and parameter should always be called `smth_enabled` so that the API is `smth_enabled(True)`?

astafan8:
@nataliejpg 
> I didn't understand your question:
> > Do we also need enabled/disabled val mapping? Or shall that case be strongly recommended to be True/False and parameter should always be called `smth_enabled` so that the API is `smth_enabled(True)`?

I also don't really remember :( I think it was about parameters which have enabled/disabled meaning which might probably be different from on/off meaning in some cases (definitely not in this PR's rhode schwarz driver), if we need a `create_en_dis_abled_val_mapping` function of some sort.

But now i think we don't, and also this question of mine is a bit "stupid" :)

PR #1418, opened 2018-12-06
Fix/infiniium

sohailc:
This PR fixes the Infiniium oscilloscope driver by Agilent. In particular, the issue that an exception is raised when xorigin and xincrem parameters change to much between preparing curvedata and actual acquisition. What we want to do instead is interpolate between the actual sampling points such that we get the data at the requested sampling points.

@Dominik-Vogel I believe you were the original author of this driver, so I am including you as a reviewer.


sohailc:
@Dominik-Vogel Can you have a look, please?

WilliamHPNielsen:
@sohailc I'm glad we agree about the interpolation. Regarding the time frame for #1263, it may indeed be the best thing to do the workaround that you suggest. Not much will happen this year at least. 

But just for my understanding, since I haven't used the scope, can you perhaps elaborate on how often this exception is thrown? And is there no known way to prevent it? "Hit or miss" does sound very bad.

sohailc:
How often an exception is thrown depends on you settings (e.g. the length of the trace) and how long the time period is between preparing for the acquisition and performing the acquisition. But even under the best of circumstances, for me it failed ~20-30% of the time.

WilliamHPNielsen:
Okay, that's bad indeed.

PR #1421, opened 2018-12-20
Extension/lakeshore read curve file

sohailc:
Added the ability to parse and load curve files.


qcodes/instrument_drivers/Lakeshore/Model_325.py
```diff
@@ -1,11 +1,87 @@
...
+    Item3: value
+
+    Header1   Header2   Header3
+    data11     data12    data13
+    data21     data22    data23
+    """
+
+    def split_data_line(
+            line: str,
+            parsers: Union[type, List]=str
```
astafan8: the `List` case seems to never been used, hence please remove the support for it.

qcodes/instrument_drivers/Lakeshore/Model_325.py
```diff
@@ -1,11 +1,87 @@
...
+    file_data = dict([strip(line.split(":")) for line in metadata_lines])
+    # After meta data we have a data header
+    header_items = strip(split_data_line(next(lines)))
+    # After that we have the curve data
+    data = [
+        split_data_line(line, parsers=float)
+        for line in lines if line.strip() != ""
+    ]
+
+    file_data["data"] = dict(
```
astafan8: if for some reason one of the metadata items is called "data", then it will be overwritten with this operation. I hope it will never happen, however, I'd recommend to make the code more robust by putting all the metadata into its own dictionary within `file_data`.

qcodes/instrument_drivers/Lakeshore/Model_325.py
```diff
@@ -1,11 +1,87 @@
...
+    Read a curve file with extension *.330
+    The format of the file is as follows:
+
+    Item1: value
+    Item2: value
+    Item3: value
+
+    Header1   Header2   Header3
+    data11     data12    data13
+    data21     data22    data23
```
astafan8: please also document the format of the returned dictionary.

qcodes/instrument_drivers/Lakeshore/Model_325.py
```diff
@@ -1,11 +1,87 @@
...
+    return file_data
+
+
+def sanitize_data(file_data: dict) -> None:
+    """
+    Data as found in the curve files are slightly different then
+    the dictionary as expected by the 'upload_curve' method of the
+    driver
+
+    Note: This function modifies the input
```
astafan8: why was this decision made instead of letting this function return a new dictionary object?

qcodes/instrument_drivers/Lakeshore/Model_325.py
```diff
@@ -1,11 +1,87 @@
...
+    """
+    data_dict = file_data["data"]
+    # We do not need the index column
+    del data_dict["No."]
+    # Rename the 'Units' column to the appropriate name
+    # Look up under the 'Data Format' entry to find what units we have
+    data_format = file_data['Data Format']
+    # This is a string in the form '4      (Log Ohms/Kelvin)'
+    data_format_int = int(data_format[0])
```
astafan8: what is the total number of supported units? is it indeed 4 as `valid_sensor_units` suggests? regardless, i'd make this line a bit more robust via `... = int(data_format.split()[0])`, is it fair to do so?

qcodes/instrument_drivers/Lakeshore/Model_325.py
```diff
@@ -469,8 +545,26 @@ def upload_curve(
...
         curve.set_data(data_dict, sensor_unit=sensor_unit)
+
+    def upload_curve_from_file(self, index: int, file_path: str) -> None:
+        """
+        Upload a curve from a curve file. Note that we only support
+        curve files with extension *.330
+        """
+        _, filename = os.path.split(file_path)
+        if not filename.endswith(".330"):
+            raise ValueError("Only curve files with extension *.330 is supported")
```
astafan8:
```suggestion
            raise ValueError("Only curve files with extension *.330 are supported")
```

qcodes/instrument_drivers/Lakeshore/Model_325.py
```diff
@@ -469,8 +545,26 @@ def upload_curve(
...
         curve.serial_number(serial_number)
         curve.format(f"{sensor_unit}/K")
         curve.set_data(data_dict, sensor_unit=sensor_unit)
+
+    def upload_curve_from_file(self, index: int, file_path: str) -> None:
+        """
+        Upload a curve from a curve file. Note that we only support
+        curve files with extension *.330
+        """
+        _, filename = os.path.split(file_path)
```
astafan8: why is this needed? isn't `filename.endswith(".330") == file_path.endswith(".330")`?

qcodes/instrument_drivers/Lakeshore/Model_325.py
sohailc: Ok

qcodes/instrument_drivers/Lakeshore/Model_325.py
sohailc: Ok

qcodes/instrument_drivers/Lakeshore/Model_325.py
sohailc: Good idea :-)

qcodes/instrument_drivers/Lakeshore/Model_325.py
sohailc: Yeah its cleaner to make and modify a copy

PR #1425, opened 2018-12-27
Consolidates requirements files as outlined in #1424

### LINKED ISSUE #1424: Tests fail due to missing websockets requirement
While setting up Qcodes for a colleague we ran into the issue that running the unit tests locally fails with ModuleNotFoundError (`websockets`).

This was probably introduced as part of #1407 . The fix is easy, by adding websockets to the test_requirements.txt file. 

But the real issue I think is that there are several places where dependencies are defined (setup.py, requirements.txt, test_requirements.txt, docs_requirements.txt). I am wondering why the CI didn't catch that? does it execute the unit tests in the same env as where it install Qcodes?

If you encounter a bug use the following template.
If you have a feature request feel free to freestyle.
### Steps to reproduce
1. Clone latest Qcodes
2. Create virtualenv, and `pip install -r requirements.txt -r test_requirements.txt -r docs_requirements.txt`
3. run pytest in qcodes/tests

### Expected behaviour
Tests should all pass

### Actual behaviour
Tests throw:

```Python
qcodes/tests/test_monitor.py:None (qcodes/tests/test_monitor.py)
ImportError while importing test module '/Users/swdroge/PycharmProjects/Qcodes/qcodes/tests/test_monitor.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
test_monitor.py:9: in <module>
    import websockets
E   ModuleNotFoundError: No module named 'websockets'
```

### System
**operating system**: OSX & Windows

**qcodes branch**: master

**qcodes commit**: 023d4189

StefanD986:
Fixes #1424 .

There are several files where we track dependencies of Qcodes, which can roughly broken down into runtime and development dependencies. 

Previous state of the requirements files is:

  - `setup.py`: runtime requirements
  - `requirements.txt`: runtime requirements (but not all)
  - `test_requirements.txt`: development requirements such as test frameworks
  - `docs_requirements.txt`: packages needed to build docs, plus jupyter to execute examples, and support packages for some of the notebooks (pandas)

This PR suggests to consolidate the requirements into 2-3 files to reduce duplication:

  - Let's use `setup.py` for runtime dependencies only
  - let's use `requirements.txt` for additional development dependencies only.
    - As development dependencies are a superset of runtime dependencies, let's include setup.py automatically by adding `.` to `requirements.txt` instead of repeating them here.
    - Since we all write tests for our code 😉 test frameworks are development dependencies and are added to requirements.txt

(For now I did not touch docs_requirements.txt and also not the conda package spec.)

@WilliamHPNielsen


requirements.txt
```diff
@@ -1,13 +1,15 @@
-numpy==1.15.*
-matplotlib==3.0.*
+.
```
jenshnielsen: I think this is a really bad idea. I would never expect that `pip install -r requirements.txt` installs the package it self. That breaks a fundamental contract with the user. More over this is unwanted in cases were I want to install all the dependencies as regular packages but install qcodes as editable. In the best case this results in a redundant install of qcodes.

astafan8:
Closing this because the discussion seems to indicate that we will stick with the current way environment/requirement/setup files are organized. See also discussion in #1556

PR #1429, opened 2019-01-04
Don't enforce equal length blob data in add_results

ThorvaldLarsen:
It is currently not possible to add measurements of arraytypes of different lengths with a single call of `add_results`. 

This PR stops checking lengths of parameters that are registered as array-types (i.e. saved as blobs in the database). The change has been tested at one setup when saving a mix of scalar parameter, arrayparameter, and multiparameter. Also `plot_by_id` has no problems extracting the data.


jenshnielsen:
@ThorvaldLarsen Thanks, there seems to be a problem with the tests. I will take care of that and add another test for your specific usecase

PR #1430, opened 2019-01-04
Don't enforce equal length blob data in add_results

jenshnielsen:
Replaces #1429


jenshnielsen:
@ThorvaldLarsen @QCoDeS/core This should now pass the existing tests and have it's own test and be ready for review

ThorvaldLarsen:
@jenshnielsen do you want me to test this at T3 before merge?

jenshnielsen:
We realized that this needs some more work as it does not enforce that dependent parameters have equal lenght. I think @WilliamHPNielsen is working on that

jenshnielsen:
This is redundant now

PR #1434, opened 2019-01-07
Pytest 4.1 compatability. Don't call fixtures in notebook and use deprecated function

WilliamHPNielsen:
Fixes the failure of tests due a new version of pytest. From version 4.1 onward calling a fixture is no longer a warning, but an error. So let's not do that anymore.

- Don't call fictures in notebook

@jenshnielsen


jenshnielsen:
This still leaves the example notebook depending on pytest. I think we should move the generator out of that file to avoid that

PR #1433, opened 2019-01-04
Driver: Keysight 344xxA - flexible measurements and modularization

astafan8:
Improve old Keysight 344xxA DMM (digital multimeter) driver by adding methods and parameters that allow to perform measurement more flexibly than the existing `data_buffer` `ArrayParameter` implements. Also, add a new version Keysight 344xxA DMM drivers where parameters are logically grouped in submodules whenever possible.

New driver:
* uses submodules for traiggering/sampling/display parts
* does not have `data_buffer` `ArrayParameter`
* tests were ported

Old driver:
* Deprecate `data_buffer` `ArrayParameter`
* Add many useful missing methods and parameters, like `read` and `fetch`

Both drivers:
* Add example notebook for the drivers with an example of measurement routines
* Extract error handling to a common mixin class `KeysightErrorQueueMixin` that is used in 4 Keysight drivers
* Add many more docstrings
* Add `MIN`/`MAX`/`DEF`/`INF` values to parameters that didn't have them
* Fix a number of cases where a particular model has or doesn't have a particular parameter or a value of a parameter

Replaces and Closes https://github.com/QCoDeS/Qcodes/pull/1393.

PR #1435, opened 2019-01-08
Remove deprecated keysight waveform drivers drivers

jenshnielsen:
And update the example notebook to not use them


jenshnielsen:
As noted in #1433

PR #1436, opened 2019-01-09
Move legacy notebooks to a separate folder

jenshnielsen:
And add a note to the docs/readme explaining why @QCoDeS/core


astafan8:
I did download and checked the zipped docs, looks ok for this particular doc-cleanup-step.

PR #1437, opened 2019-01-09
Avoid warnings in CI, and a bit more

astafan8:
Python 3.6 warning-related:
* Remove warning from `test_wrong_address` where deprecated driver has been used
- [x] Remove the `set` warning from the MultiParameter and extended it's test a bit to reflect this (rather than filtering it) (~Suppress warnings originating from poor implementations of MultiParameter and MultiChannelInstrumentParameter~)
* Remove deprecated `.value` usage on HDF5 datasets (`h5py`)
* Avoid hypothesis warnings by using pytest parametrization when appropriate
* Avoid hypothesis timeout-related warnings by limiting max_examples for a test

Python 3.7. warning-related:
- [x] deprecated asyncio stuff in `monitor`
- [x] deprecated import from `collections` in `qcmatplotlib.py`

The "more":
* Move `error_caused_by` to `tests/common.py` module

### DIFF
--- a/qcodes/data/hdf5_format.py
+++ b/qcodes/data/hdf5_format.py
@@ -97,7 +97,7 @@ class HDF5Format(Formatter):
             set_arrays = [s.decode() for s in set_arrays]
             # else:
             #     set_arrays = ()
-            vals = dat_arr.value[:, 0]
+            vals = dat_arr[:, 0]
             if 'shape' in dat_arr.attrs.keys():
                 # extend with NaN if needed
                 esize = np.prod(dat_arr.attrs['shape'])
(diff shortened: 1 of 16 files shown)


astafan8:
@QCoDeS/core presumably ready for review.

Please, tell me if https://github.com/QCoDeS/Qcodes/pull/1437/commits/665899dc8887e55da12a797235cb19aea4c7b0f4 is a correct fix. I'm lacking knowledge, hence confidence.

jenshnielsen:
It is unclear to me if ellipses or the empty tuple is the right way to get the full array from h5py. Both seems to work equivalently in my testing.

I honestly think that we should just remove the set warning from the multiparameter rather than filtering it. Alex has confirmed long time ago that it was just there because the old loop does not support setting it.

jenshnielsen:
The drop in coverage seems to be a bug in codecov. It counts correctly that the old driver is not covered but omits the new one because it was uncovered before this pr.

astafan8:
Ok, i'll stick to `...` then.

Ok, i'll remove the `set` warning.

jenshnielsen:
The remaining warnings are all coming from third party libraries. I have verified that they have been fixed (or at least the relevant lines have changed) in the master branch of Pandas, websockets, jsonschema, and pyyaml so we expect them to disappear in future releases

astafan8:
Thanks, @jenshnielsen , I approve!

Issue #1268, opened 2018-09-13
Docs don't build with Sphinx 1.8.0


jenshnielsen:
```
reading sources... [ 98%] user/index
reading sources... [ 99%] user/intro
reading sources... [100%] user/tutorial

/home/travis/build/QCoDeS/Qcodes/qcodes/utils/deferred_operations.py:45: UserWarning: DeferredOperations has been deprecated and will be removed in the not-too-distant future.
  warnings.warn("DeferredOperations has been deprecated and will be removed "

Warning, treated as error:
Invalid type annotation found on <function ArrayParameter.__init__ at 0x2b9cbee131e0>. Ingored: NameError("name 'Instrument' is not defined",)
make[1]: *** [html] Error 2
make[1]: Leaving directory `/home/travis/build/QCoDeS/Qcodes/docs'
make: *** [html-api] Error 2
```

astafan8:
Closing this as the issue has been resolved in Sphinx and we are using sphinx `2.0.1` at least.

PR #1441, opened 2019-01-14
Build docs with sphinx 1.8.3

jenshnielsen:
The bug causing #1268 has been resolved so lets try building the docs with Sphinx 1.8.3


jenshnielsen:
closes #1268

PR #1442, opened 2019-01-14
Fix logic error in db upgrade

jenshnielsen:
Throw a more explicit exception if the paramtype of a column in a table cannot be found
Previously this would fail with paramtype not defined. This is not a fix for that but will at least give a meaningful error message

Seen when trying to upgrade a db where the run was created but the results table was missing, due to another exception thrown. (I am looking into reproducing that)


jenshnielsen:
The failure should be fixed by #1443

jenshnielsen:
It's a good question what we should do. Most likely this will happen because the run has been created in the runs table but not no matching result table has been created. This is caused by #1444 due to the missing atomic transaction. I can think of 2 solutions. Drop the run completely as it has no value? or hard code a default value

astafan8:
ok. But wouldn't it be more valuable to have a "fixxer" method that'd set paramtype to given value? My idea is that if a database contains an "empty" run, that run should be a valid "empty" run (whatever that means). For example, if the run is there but the "empty" results table is missing, i'd suggest to just add that empty table and that's it, so that the run becomes a valid "empty" one.

I think it makes sense that db upgraders assume correct databases, not "corrupted" ones. This makes them simple (or at least simpler). THen if there are problems with the database like the one you've mentioned, then i'd recommend to fix the db first, and then apply the upgrade.

Would you agree?

jenshnielsen:
@QCoDeS/core Should we merge this one and worry about later if this actually happens. I suspect we removed the cause of this when we prevented running older versions on newer dbs

PR #1443, opened 2019-01-14
Try different way of starting xvfb

jenshnielsen:
It seems like Travis has deployed a new image today that breaks the way we used to start xvfb. This should fix that

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ install:
 
 before_script: # configure a headless display to test plot generation
     - "export DISPLAY=:99.0"
-    - "sh -e /etc/init.d/xvfb start"
+    - Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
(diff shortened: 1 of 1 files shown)


jenshnielsen:
See https://travis-ci.org/QCoDeS/Qcodes/jobs/479404509 for an example of the failure

jenshnielsen:
If you compare that build log to an earlier one you can see that there is a new image from Jan 14th compared to the image from November used previously

.travis.yml
```diff
@@ -31,7 +31,7 @@ install:
 
 before_script: # configure a headless display to test plot generation
     - "export DISPLAY=:99.0"
-    - "sh -e /etc/init.d/xvfb start"
+    - Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
```
astafan8: capital "X"? and it works? :)

Issue #1444, opened 2019-01-14
Adding metadata dict more than one level deep creates invalid run


jenshnielsen:
```
import qcodes as qc
from qcodes import ParamSpec, new_data_set, new_experiment
dataSet = new_data_set("sweep gate")
exp = new_experiment("majorana-qbuit", sample_name="flying brdige")
parameter_a = ParamSpec("a", "numeric")
parameter_b = ParamSpec("b", "numeric", key="value", number=1)
parameter_c = ParamSpec("c", "array")
dataSet = new_data_set("sweep gate", specs=[parameter_a,parameter_b,parameter_c], metadata={'a': {'b': 1}})
```
results in

```
Rolling back due to unhandled exception
Traceback (most recent call last):
  File "c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py", line 819, in atomic
    yield conn
  File "c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py", line 781, in atomic_transaction
    c = transaction(atomic_conn, sql, *args)
  File "c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py", line 756, in transaction
    c.execute(sql, args)
sqlite3.InterfaceError: Error binding parameter 0 - probably unsupported type.

---------------------------------------------------------------------------
InterfaceError                            Traceback (most recent call last)
c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py in atomic(conn)
    818             conn.cursor().execute('BEGIN')
--> 819         yield conn
    820     except Exception as e:

c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py in atomic_transaction(conn, sql, *args)
    780     with atomic(conn) as atomic_conn:
--> 781         c = transaction(atomic_conn, sql, *args)
    782     return c

c:\users\jenielse\source\repos\qcodes\qcodes\dataset\sqlite_base.py in transaction(conn, sql, *args)
    755     if len(args) > 0:
--> 756         c.execute(sql, args)
    757     else:

InterfaceError: Error binding parameter 0 - probably unsupported type.

The above exception was the direct cause of the following exception:

RuntimeError                              Traceback (most recent call last)
<ipython-input-8-6818d1064158> in <module>
----> 1 dataSet = new_data_set("sweep gate", specs=[parameter_a,parameter_b,parameter_c], metadata={'a': {'b': 1}})
[...]

PR #1445, opened 2019-01-14
Ensure that partial run creation is completly rolled back

jenshnielsen:
This should prevent bugs such as the first part of #1444 (invalid metadata dict) invalidating the database. This should really have a test if possible.


astafan8:
Yes, this is a great fix. I also agree with the need for a test.

PR #1449, opened 2019-01-17
fix: Handle 'inf' in _convert_numeric

simonzihlmann:
Reading data from the sqlite database goes through a converter (_convert_numeric) in qcodes/dataset/sqlite_base.py that now can handle 'inf' as the numeric value properly. Previously, it threw an error because it incorrectly assumed that 'inf' would be an integer.

@astafan8


qcodes/dataset/sqlite_base.py
```diff
@@ -255,8 +256,13 @@ def _convert_numeric(value: bytes) -> Union[float, int, str]:
     # outcome is 'nan'
     if np.isnan(numeric):
         return numeric
+    
+    # If that worked, e.g. did not raise an exception, then we check if the
+    # outcome is 'inf'
+    if np.isinf(numeric):
```
astafan8: What about the case of negative infinity? Shall it also be covered in a similar way? If so, please also add a test for it.

qcodes/dataset/sqlite_base.py
```diff
@@ -255,8 +256,13 @@ def _convert_numeric(value: bytes) -> Union[float, int, str]:
     # outcome is 'nan'
     if np.isnan(numeric):
         return numeric
+    
+    # If that worked, e.g. did not raise an exception, then we check if the
```
WilliamHPNielsen: I think this comment is misleading (probably just copy-pasted from above). The "If that worked, e.g. did not raise an exception" part does not really make sense here; there is no reason that the nan-check would raise an exception.

PR #1451, opened 2019-01-21
Use getattr to get ask/write from instruments

### LINKED ISSUE #1450: Creating parameters with get/set_cmd in a virtual instrument causes AttributeError
When creating a virtual instrument (i.e. a wrapper around a device) which has parameters with a `get_cmd` or a `set_cmd` method, the creation will fail as it assumes that all instruments have a `write` or an `ask` method, which is not true for virtual instruments which subclass `InstrumentBase`.

For example, the following code snippet will cause an `AttributeError`:

```py3
from qcodes.instrument.base import InstrumentBase

class VirtualInstrument(InstrumentBase):
    def __init__(self, name):
        super().__init__(self, name)

        add_parameter("Bla", get_cmd=self.get_bla)

    def get_bla(self):
        return 1
```

Though this example is contrived, it came up in https://github.com/QNLSydney/qcodes_measurements/blob/master/qcodes_measurements/device/channel_wrapper.py which labels gates/keeps track of gate connection state.

## Potential Solutions
 - Change the initialization of Parameter to use:
```python3
exec_str_ask = getattr(instrument, "ask", None) if instrument else None
```
```python3
exec_str_write = getattr(instrument, "write", None) if instrument else None
```
 - Subclass from `Instrument` instead, although this potentially causes some issues?
    - A nonsensical value for IDN unless overwritten, which also leads to a nonsensical connect_message.
    - Linting errors on `ask_raw` and `write_raw` which must be overwritten.

spauka:
This allows us to use InstrumentBase as the base class for virtual instruments

Fixes #1450.

- Use getattr to get ask/write from instruments

@jenshnielsen @astafan8

### DIFF
--- a/qcodes/tests/instrument_mocks.py
+++ b/qcodes/tests/instrument_mocks.py
@@ -4,7 +4,7 @@ import logging
 
 import numpy as np
 
-from qcodes.instrument.base import Instrument
+from qcodes.instrument.base import Instrument, InstrumentBase
 from qcodes.utils.validators import Numbers
 from qcodes.instrument.parameter import MultiParameter, Parameter, ArrayParameter
 from qcodes.instrument.channel import InstrumentChannel, ChannelList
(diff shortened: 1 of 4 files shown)


jenshnielsen:
Seems reasonable enough. Could you add a test based on the code in #1450

spauka:
@jenshnielsen @astafan8 added some extra tests on all the changes.

Not sure why the Azure test failed, I don't seem to have access to it...

jenshnielsen:
It seems to be a random networking error during the docs build. I have restarted

qcodes/tests/instrument_mocks.py
```diff
@@ -56,25 +56,31 @@ def _measure_skewed_parabola(self):
...
         '''
         return ((self.x.get()**2 + self.y.get()**2 +
-                self.z.get()**2)*(1 + abs(self.y.get()-self.x.get())) +
-                self.noise.get()*np.random.rand(1))
+                 self.z.get()**2)*(1 + abs(self.y.get()-self.x.get())) +
+                 self.noise.get()*np.random.rand(1))
 
 
-class MockMetaParabola(Instrument):
+class MockMetaParabola(InstrumentBase):
```
astafan8: well done finding this test suite :)

PR #1452, opened 2019-01-23
Only output db_debug in tests when env variable is set

jenshnielsen:
Normally the sqlite debug output is not very useful but included in every failing test. This sets the output for the tests depending on an environmental variable. If you want to include it you can set the env variable before running the tests.


qcodes/tests/dataset/temporary_databases.py
```diff
@@ -20,7 +20,8 @@ def empty_temp_db():
     # create a temp database for testing
     with tempfile.TemporaryDirectory() as tmpdirname:
         qc.config["core"]["db_location"] = os.path.join(tmpdirname, 'temp.db')
-        qc.config["core"]["db_debug"] = True
+        if os.environ.get('QCODES_SQL_DEBUG'):
+            qc.config["core"]["db_debug"] = True
```
astafan8: Perhaps also this in order to ignore whatever the config setting is?
```suggestion
            qc.config["core"]["db_debug"] = True
        else:
            qc.config["core"]["db_debug"] = False
```

WilliamHPNielsen:
> For the CI, shall this variable exist and set to True as default? for eample, for azure, shall it be configured in our "pipeline" so that it can be also manually changed when manually queueing a build?

I'd say "yes" and "yagni".

jenshnielsen:
I would argue that it should be set to False by default also in CI. I cannot think of a single time this has been useful but I can think of lots of times where it actively made it harder to figure out what is going on

WilliamHPNielsen:
I yield and agree make it "no" everywhere by default.

PR #1453, opened 2019-01-23
Fix issues with infered_from and add checks for infered_from and depends_on that are strings

### LINKED ISSUE #1448: Documentation error v0.1.11: load_or_create_experiment
Hello, thank you for this awesome library. I am using it for my master thesis at Universidad Autónoma de Madrid (Spain) and find it really useful. While learning I have found an error in the documentation.

I am using qcodes v0.1.11 (last version I have been able to install with pip using your instructions (http://qcodes.github.io/Qcodes/start/index.html#installing-the-latest-qcodes-release). When I check the documentation of this version I can see there is a function `load_or_create_experiment` (http://qcodes.github.io/Qcodes/api/generated/qcodes.dataset.experiment_container.load_or_create_experiment.html) However, I cannot import it in qcodes.

### Steps to reproduce
```python
sys.version

'3.6.7 |Anaconda, Inc.| (default, Dec 10 2018, 20:35:02) [MSC v.1915 64 bit (AMD64)]'

import qcodes
qcodes.__version__
'0.1.11'

from qcodes.dataset import experiment_container
experiment_container.load_or_create_experiment()

    AttributeError                            Traceback (most recent call last)

    <ipython-input-16-44f48f692e42> in <module>
    ----> 1 experiment_container.load_or_create_experiment()
    

    AttributeError: module 'qcodes.dataset.experiment_container' has no attribute 'load_or_create_experiment'
```

### Expected behaviour

I am still a novice with github and qcodes but I have been able to find that this function was introduced last summer while the last release of qcodes dates from April. Thus, I would expect not to see this function in the documentation of v0.1.11. This might be happening also for other new features, but I haven't explored further.

If I want to use these new features, would you recommend me to clone the github repo and install it from source or are you planning to make a new release soon?

jenshnielsen:
Fixes #1448 

Perform db upgrade to correct:
* infered_from was passed as a string not a list of strings
* infered_from was not handled correctly for parameters that were neither deps nor dependents
In addition:
* Add logic to catch infered_from and depends_on being passed as a string. 
* move db generation scripts to their own repo to disentangle qcodes git hash from the code using it. Otherwise it is very painful to change test fixtures for older db versions 

As noted below I have not found a good way to get mypy to catch this kind of error. 
You cannot use `List[Union[ParamSpec, Str]]` as list is an invariant type https://github.com/python/mypy/issues/3351, the current type checking does not catch is since str in it self is a Sequence
 
@WilliamHPNielsen could you do a review


jenshnielsen:
* Agree on the test.
* I think we need an upgrade. We can perform the fix now because we still have the redundant information 
in the other table but once we do the upgrade to drop that table this needs to be fixed beforehand. I.e in a previous upgrade. If it's in a fix we may not catch all instances before doing the dropping upgrade.

Mypy:

I expected that 
```
S = TypeVar('S')
Listuple = Union[List[S], Tuple[S, ...]]
```
But that does not seem to be that simple.

jenshnielsen:
You cannot meaningfully use a List of a Union unfortunately https://github.com/python/mypy/issues/3351

qcodes/dataset/sqlite_base.py
```diff
@@ -727,7 +713,7 @@ def perform_db_upgrade_2_to_3(conn: ConnectionPlus) -> None:
 def perform_db_upgrade_3_to_4(conn: ConnectionPlus) -> None:
     """
     Perform the upgrade from version 3 to version 4. This really
-    repeats the version 3 upgrade as it originally had a bug in
+    repeats the version 3 upgrade as it originally had two bugs in
```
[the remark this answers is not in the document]
jenshnielsen: Sure, done

qcodes/dataset/sqlite_base.py
```diff
@@ -720,6 +709,77 @@ def perform_db_upgrade_2_to_3(conn: ConnectionPlus) -> None:
...
+@upgrader
+def perform_db_upgrade_3_to_4(conn: ConnectionPlus) -> None:
+    """
+    Perform the upgrade from version 3 to version 4. This really
+    repeats the version 3 upgrade as it originally had two bugs in
+    the inferred annotation. inferred_from was passed incorrectly
+    resulting in the parameter being marked inferred_from for each char
+    in the inferred_from variable and inferred_from was not handled
+    correctly for parameters that were neither dependencies nor dependent on
+    other parameters. Both have since been fixes so rerun the upgrade.
```
WilliamHPNielsen: I guess "fixes" -> "fixed"?

qcodes/tests/dataset/test_database_creation_and_upgrading.py
```diff
@@ -259,16 +260,275 @@ def test_perform_actual_upgrade_2_to_3_some_runs():
...
+    version will be corrected when upgraded to v4.
+    """
+
+    v3fixpath = os.path.join(fixturepath, 'db_files', 'version3')
+
+    dbname_old = os.path.join(v3fixpath, 'some_runs_upgraded_2.db')
+
+    if not os.path.exists(dbname_old):
+        pytest.skip("No db-file fixtures found. You can generate test db-files"
+                    " using the scripts in the legacy_DB_generation folder")
```
WilliamHPNielsen: The location of the scripts to generate fixtures has changed, so this skip message should reflect that change.

WilliamHPNielsen:
Modulo a few comments above, this looks ready.

jenshnielsen:
Sure that sounds good

PR #1455, opened 2019-01-28
add flush option to hdf5 write_metadata

eendebakpt:
Fixes issue with HDF5 files that are not properly closed.

When metadata is written after the main data has been written, the HDF5 file was properly closed. This could lead to currupt HDF5 files. This PR fixes the issue by flushing the HDF5 file by default after writing the metadata.


astafan8:
Could you expand on what this PR is about and in what situation the problem happens? Why wouldn't a file be "not properly closed"? And what issue exactly occurs when "a file is not properly closed"?

Also, this change broke some tests, so, please, fix them. Ideally, the change in the test will help communicating why the new behavior is more beneficial.

eendebakpt:
@astafan8 I updated the PR description. The failing test was invalid, so I removed it.

The test checked whether copying a generated HDF5 file without performing an explicit flush would generate an error. Although copying the file without an explicit flush can result in an error, this is not necessary (since the flush can be triggered by other methods as well).

peendebak:
@astafan8 The PR passes all tests.

PR #1456, opened 2019-01-28
Feature/rigol dg1062/adding duty cycle parameter

vhartong:
- Added Pulse waveform to the list
- Added translation dictionary for waveforms (Device responses are sometimes different 👎 )
- Adding function duty cycle parameter for Square wave and Pulse Wave
- General code cleanup + two bugs


qcodes/instrument_drivers/rigol/DG1062.py
```diff
@@ -160,8 +165,8 @@ def __init__(self, parent: 'DG1062', name: str, channel: int) ->None:
                 ),
                 vals.Enum("INF", "MIN", "MAX", "HighZ")
             ),
-            set_parser=lambda value: "INF" if value == "HighZ" else value,
             get_parser=lambda value: "HighZ"
+            set_parser=lambda value: "INF" if value == "HighZ" else value,
             if float(value) > DG1062Channel.max_impedance else float(value)
         )
```
astafan8: something terribly wrong happened here :) the automatic testing is alarming: https://ms-quantum.visualstudio.com/QubitWorkbench/_build/results?buildId=10600.

I think when moving `set_parser` inserted it in the middle of the lambda expression for `get_parser` which breaks everything :) I guess the following is what you wanted, right?:
```python
<...>
),  # <- comma from the line before
get_parser=lambda value: "HighZ" 
    if float(value) > DG1062Channel.max_impedance else float(value), # <- comma
set_parser=lambda value: "INF" if value == "HighZ" else value  # <- no comma here because of the brace on the next line
)
<...>
```

Also, you sure this final version of the code has been fully tested on a real device? :)

qcodes/instrument_drivers/rigol/DG1062.py
```diff
@@ -102,20 +102,25 @@ def trigger(self) ->None:
...
-        ["HARM", "NOIS", "RAMP", "SIN", "SQU", "TRI", "USER"]
+        ["HARM", "NOIS", "RAMP", "SIN", "SQU", "TRI", "USER", "PULS"]
     }
 
     waveform_params["DC"] = ["freq", "ampl", "offset"]
     waveform_params["ARB"] = ["sample_rate", "ampl", "offset"]
 
+    """
+    Responses from the machine don't always match the name to set the function, hence a translater
+    """
```
astafan8: great catch!

qcodes/instrument_drivers/rigol/DG1062.py
```diff
@@ -182,8 +187,15 @@ def __init__(self, parent: 'DG1062', name: str, channel: int) ->None:
 
         self.add_parameter(
             "state",
-            set_cmd=f"OUTPUT{channel}:STATE {{}}",
             get_cmd=f"OUTPUT{channel}:STATE?",
+            set_cmd=f"OUTPUT{channel}:STATE {{}}",
+        )
+        
+        self.add_parameter(
+            "dcycle",
```
astafan8: could you rename this to `duty_cycle` for readability? good names give much more long-term benefit that saving on a couple of characters (it does depend on a particular case, but here - for sure).

qcodes/instrument_drivers/rigol/DG1062.py
```diff
@@ -182,8 +187,15 @@ def __init__(self, parent: 'DG1062', name: str, channel: int) ->None:
 
         self.add_parameter(
             "state",
-            set_cmd=f"OUTPUT{channel}:STATE {{}}",
             get_cmd=f"OUTPUT{channel}:STATE?",
+            set_cmd=f"OUTPUT{channel}:STATE {{}}",
+        )
+        
+        self.add_parameter(
```
astafan8: Would you mind adding a `label`, `unit`, and `docstring` to this parameter? (this is done by adding new arguments with these names to the `add_parameter` call, see how it's done in other drivers, even for some parameters in this driver)

qcodes/instrument_drivers/rigol/DG1062.py
```diff
@@ -290,7 +304,32 @@ def _set_waveform_params(self, **params_dict: Dict) ->None:
         string += ",".join(
             ["{:7e}".format(params_dict[param]) for param in param_names])
         self.parent.write_raw(string)
+        
+    def _get_duty_cycle(self):
+        """
+        Simple function to extract the duty cycle after checking waveform
+        """
+        self.waveform()
+        wf = self.waveform.get_latest()
```
astafan8: in case it's necessary for `_get_duty_cycle` function to operate on the real actual latest waveform of the instrument, then `wf = self.waveform.get_latest()` is redundant because `wf = self.waveform()` does it already.

in case it's necessary for `_get_duty_cycle` function to avoid additional communication with the instrument, then `self.waveform()` should be removed because it does perform the communication; while `wf = self.waveform.get_latest()` only does the job of returning the "latest value of the parameter that is known to the driver (yes, the state of the instrument might have changed since the value of the parameter has been requested due to, for example, manually clicking buttons of the instrument)".

So, depending on the intention, please, get rid of either of the lines.

qcodes/instrument_drivers/rigol/DG1062.py
```diff
@@ -290,7 +304,32 @@ def _set_waveform_params(self, **params_dict: Dict) ->None:
...
+
+        if wf == 'PULS' or wf == 'SQU':
```
astafan8: the following is more pythonic, readable, future-proof
```suggestion
        if wf in ['PULS', 'SQU']:
```

qcodes/instrument_drivers/rigol/DG1062.py
```diff
@@ -290,7 +304,32 @@ def _set_waveform_params(self, **params_dict: Dict) ->None:
...
+            duty_cycle = self.parent.ask_raw(f":SOUR{self.channel}:FUNC:{wf}:DCYC?")
```
astafan8: It is quite weird that throughout this driver `self.parent.ask_raw` and `self.parent.write_raw` are used instead of just `self.ask` and `self.write` which do the same, but are much cleaner to use from the code/design point of view.

For this driver and this PR, ok, let's keep it consistent. But in general, it's expected that `self.write` and `self.ask` of the `InstrumentChannel` (which is the class of `self` here) class are used, unless there is some really important reason not to do it.

qcodes/instrument_drivers/rigol/DG1062.py
```diff
@@ -290,7 +304,32 @@ def _set_waveform_params(self, **params_dict: Dict) ->None:
         string += ",".join(
             ["{:7e}".format(params_dict[param]) for param in param_names])
         self.parent.write_raw(string)
+        
+    def _get_duty_cycle(self):
+        """
+        Simple function to extract the duty cycle after checking waveform
```
astafan8: This docstring has ~70% of useful information - "extract the duty cycle after checking waveform"  - and ~30% of useless information - "Simple function to". Could you rephrase it to smth like "Read duty cycle value. Also reads the waveform parameter of the instrument."?

Moreover, one might argue that the fact that this method raises when used for waveforms that don't have a meaningful duty cycle, should also be mentioned in the docstring. In this particular case, this note belongs to the docstring of the parameter (see the `docstring` kwargs for `self.add_parameter('duty_cycle', ...)`.

Finally, this method is fairly short, hence it's code perfectly expresses what is going on, hence the docstring is not really needed, in my humble opinion.

To conclude, please:
* depending on your taste, either remove the docstring or make it "100%" useful :)
* add a notion about "allowed waveforms" to the docstring of the parameter in `add_parameter(...)` call.

qcodes/instrument_drivers/rigol/DG1062.py
```diff
@@ -290,7 +304,32 @@ def _set_waveform_params(self, **params_dict: Dict) ->None:
...
+        wf = self.waveform.get_latest()
+
+        if wf == 'PULS' or wf == 'SQU':
+            duty_cycle = self.parent.ask_raw(f":SOUR{self.channel}:FUNC:{wf}:DCYC?")
+        else:
+            raise ValueError(f"Can not read duty cycle for current function: {wf}")
 
+        return duty_cycle
+
+    def _set_duty_cycle(self,duty_cycle):
```
astafan8: The same comments from `_get_duty_cycle` apply to this method.

qcodes/instrument_drivers/rigol/DG1062.py
```diff
@@ -290,7 +304,32 @@ def _set_waveform_params(self, **params_dict: Dict) ->None:
         string += ",".join(
             ["{:7e}".format(params_dict[param]) for param in param_names])
         self.parent.write_raw(string)
+        
+    def _get_duty_cycle(self):
```
astafan8: what's the type of the return value?`float`? if so, please, add a type annotation like this:
```suggestion
    def _get_duty_cycle(self) -> float:
```

qcodes/instrument_drivers/rigol/DG1062.py
vhartong: Yes, this caused a problem. Thanks for picking it up

qcodes/instrument_drivers/rigol/DG1062.py
vhartong: Done.

qcodes/instrument_drivers/rigol/DG1062.py
vhartong: duty cycle does not have a unit, writing up doc_string: done

qcodes/instrument_drivers/rigol/DG1062.py
vhartong: completely agree, looks way better

qcodes/instrument_drivers/rigol/DG1062.py
vhartong: to be honest, i tried to stay consistent within the code. maybe something for the next merge?

astafan8: sure.

PR #1459, opened 2019-01-31
Fixed issue with legacy import

ectof:
replaced array.name for array.array_id in dataset.legacy_import.
Reason: the array.name was often non-unique for distinct parameters, e.g in a 2d sweep of two gates both parameters would have array.name 'v'

tested by Eoin

@jenshnielsen


jenshnielsen:
@ectof I think this looks good. I have updated the automatic tests to reflect the changes

PR #1461, opened 2019-02-04
add option for pretrigger memsize to channel readout

peendebak:
The option to acquire a single value from the M4i digitizer records an array of data and then takes the mean. Part of this data is from the actual moment the value is requested (the pre-trigger part of the buffer). This PR make sthe pre-trigger size configurable and changes the default to 16 (as small as possible).

@jenshnielsen @astafan8


qcodes/instrument_drivers/Spectrum/M4i.py
```diff
@@ -611,7 +611,7 @@ def convert_to_voltage(self, data, input_range):
         return data * input_range / resolution
 
     def initialize_channels(self, channels=None, mV_range=1000, input_path=0,
-                            termination=0, coupling=0, compensation=None, memsize=2**12):
+                            termination=0, coupling=0, compensation=None, memsize=2**12, pretrigger_memsize = None):
```
astafan8: Why not make the default value of this argument `16` instead of `None`? This will remove the need for a couple of lines of code below, and simplify the code because `pretrigger_memsize` will become of type `int` instead of `Optional[int]`.

qcodes/instrument_drivers/Spectrum/M4i.py
```diff
@@ -621,9 +621,13 @@ def initialize_channels(self, channels=None, mV_range=1000, input_path=0,
             mV_range, input_path, termination, coupling, compensation: passed
                 to the set_channel_settings function
             memsize (int): memory size to use for simple channel readout
+            pretrigger_memsize (None or int): pretrigger memory size to use. If None calculate a default value
```
astafan8: The ".. calculate a default value" part is misleading because nothing is being calculated, the default value is just hardcoded. I recommend to remove this notion. It is much more useful to add the fact that "16" is apparently the smallest possible value for pretrigger memory - this will be very helpful for users.

qcodes/instrument_drivers/Spectrum/M4i.py
eendebakpt: @astafan8 I agree. There might be future cases where we want to calculate the actual value, but since that is currently not there let's just use the smallest value possible.

PR #1462, opened 2019-02-05
Document exporting of data from dataset

jenshnielsen:
Planing to also add a more detailed notebook


WilliamHPNielsen:
I like it. It's a really good idea to introduce data export/retrieval upfront in the context manager notebook and to sprinkle examples of using it out through the notebook. 

Comments on the notebooks:

- The measurement context manager notebook still has an experimental feel to it. Many paragraphs are unchanged since I originally wrote them to convince the world that we could do better than the old `Loop`. We should rewrite the whole thing into a more authoritative tutorial tone. Perhaps in a separate PR. The changes introduced in this PR are clear improvements.
- The `Working-With...` notebook is sweet! There is a fair amount of typos, though. I wouldn't mind going through it and fixing them. It's usually faster/easier for another person than the original author to do that.

jenshnielsen:
Sure that sounds good. I agree that we should rewrite the context manager notebook and it should be more of a getting started kind of thing

PR #1464, opened 2019-02-05
Metadatable should accept extra kwargs

StefanD986:
Several subclasses of Metadatable accept extra (keyword)arguments.
Through this it can happen that extra arguments are handed up the
hierarchy to Metadatable. This is why Metadatable should accept extra
kwargs.

- Adds `**kwargs` argument to `Metadatable.__init__`

@jenshnielsen @astafan8

### DIFF
--- a/qcodes/utils/metadata.py
+++ b/qcodes/utils/metadata.py
@@ -22,7 +22,7 @@ ParameterDict = Dict[ParameterKey, T]
 RunId = NewType('RunId', int)
 
 class Metadatable:
-    def __init__(self, metadata=None):
+    def __init__(self, metadata=None, **kwargs):
         self.metadata = {}
         self.load_metadata(metadata or {})
 


StefanD986:
> good, but please fix the test appropriately (see the CI report)

Yes I saw the failing test, but I was unsure whether I *should* fix it, because how the test is written it seems someone wanted to explicitly ensure that there are no extra kwargs. So before I fix it in the test I want to discuss what was the reason for this test.

(also see my question on slack:) 
> Just saw that CI is failing #1464 at test_init() Can someone explain the reason why we have a test that ensures that Metadatable doesn’t get extra arguments?

WilliamHPNielsen:
@StefanD986, I don't think there is any strong reason for that test, it has kind of just always been that way (from the very first commit, afaics). Having said that, I am a little skeptical to the idea of having the superclass init just silently swallow some kwargs to avoid an error. Why do the subclasses have to pass on their kwargs to `MetaDatable`? Perhaps you can provide an example of a situation where that happens?

WilliamHPNielsen:
@StefanD986, I don't think there is any strong reason for that test, it has kind of just always been that way (from the very first commit, afaics). Having said that, I am a little skeptical to the idea of having the superclass init just silently swallow some kwargs to avoid an error. Why do the subclasses have to pass on their kwargs to `MetaDatable`? Perhaps you can provide an example of a situation where that happens?

StefanD986:
@WilliamHPNielsen you're right. I was able to refactor my code in a way that this is not a problem anymore.

PR #1466, opened 2019-02-11
Fix typechecking with mypy=0.670

jenshnielsen:
To avoid failures in the future this locks mypy to version 0.670. If we go down this route someone should remember to regularly update our dependency on mypy


astafan8:
> If we go down this route someone should remember to regularly update our dependency on mypy

knowing how many times we had to discover failing builds because of mypy update on monday mornings, i think it's ok that we bump the version to the latest every 1-2-3 weeks.

PR #1467, opened 2019-02-11
Add simultaneous blocking ramp parameters MercuryIPS

ectof:
Added additional parameters to oxford\MercuryIPS_VISA.py allowing the magnet to be set and ramped with a single call. This is useful when running in a loop.

@WilliamHPNielsen

### DIFF
--- a/qcodes/instrument_drivers/oxford/MercuryiPS_VISA.py
+++ b/qcodes/instrument_drivers/oxford/MercuryiPS_VISA.py
@@ -277,8 +277,22 @@ class MercuryiPS(VisaInstrument):
 
             self.add_parameter(name=f'{coord}_measured',
                                label=f'{coord.upper()} measured field',
-                               unit='T',
+                               unit=unit,
                                get_cmd=partial(self._get_measured, [coord]))
+
+            self.add_parameter(name=f'{coord}_saferamp',
+                               label=f'{coord.upper()} ramp field',
+                               unit=unit,
+                               get_cmd=partial(self._get_component, coord),
+                               set_cmd=lambda x, y=coord: (self._set_target(y, x),
+                                                           self.ramp('safe')))
+
+            self.add_parameter(name=f'{coord}_simulramp',
+                               label=f'{coord.upper()} ramp field',
+                               unit=unit,
+                               get_cmd=partial(self._get_component, coord),
+                               set_cmd=lambda x, y=coord: (self._set_target(y, x),
+                                                           self.ramp('simul')))
 
         # FieldVector-valued parameters #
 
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/oxford/MercuryiPS_VISA.py
```diff
@@ -468,11 +563,12 @@ def ramp(self, mode: str="safe") -> None:
...
                 if slave.field_ramp_rate() == 0:
                     raise ValueError(f'Can not ramp {slave}; ramp rate set to'
-                                     ' zero!')
+                                        ' zero!')
 
         # then the actual ramp
         {'simul': self._ramp_simultaneously,
-         'safe': self._ramp_safely}[mode]()
+            'safe': self._ramp_safely,
+            'simul_block':self._ramp_simultaneously_blocking}[mode]()
```
astafan8: there is some weird excessive indentation in the whole commit, could this be fixed for readability?

qcodes/instrument_drivers/oxford/MercuryiPS_VISA.py
```diff
@@ -396,6 +477,20 @@ def _ramp_simultaneously(self) -> None:
...
+        """
+        Ramp all three fields to their target simultaneously at their given
+        ramp rates. NOTE: there is NO guarantee that this does not take you
+        out of your safe region. Use with care. This function is BLOCKING.
+        """
+        for slave in self.submodules.values():
+            slave.ramp_to_target()
+
+        for slave in np.array(list(self.submodules.values())):
+            # wait for the ramp to finish, we don't car about the order
```
astafan8:
```suggestion
            # wait for the ramp to finish, we don't care about the order
```

qcodes/instrument_drivers/oxford/MercuryiPS_VISA.py
```diff
@@ -396,6 +477,20 @@ def _ramp_simultaneously(self) -> None:
...
             slave.ramp_to_target()
 
+    def _ramp_simultaneously_blocking(self) -> None:
+        """
+        Ramp all three fields to their target simultaneously at their given
+        ramp rates. NOTE: there is NO guarantee that this does not take you
+        out of your safe region. Use with care. This function is BLOCKING.
+        """
+        for slave in self.submodules.values():
+            slave.ramp_to_target()
```
astafan8: these two lines can be safely substituted with `self._ramp_simultaneously()`, right? This reduces code duplication, makes the code more readable.

qcodes/instrument_drivers/oxford/MercuryiPS_VISA.py
```diff
@@ -299,8 +299,89 @@ def __init__(self, name: str, address: str, visalib=None,
...
 
+    def _theta_saferamp_set(self, target):
+        """
+		Set the theta target and ramp to it
+		"""
+        self.theta_target(target)
+        self.ramp(mode='safe')
+        self.theta_measured()
+
+    def _theta_saferamp_get(self):
```
astafan8: what are these `get*` methods needed for if they just refer to the parameters which actually represent the values (like `theta_measured`)? I'd suggest to use set-only parameters for the ramping, and if values need to be read, then appropriate corresponding parameters are used. The reason for this is to remove redundancy if it brings extra maintainability and perhaps confusion.

qcodes/instrument_drivers/oxford/MercuryiPS_VISA.py
```diff
@@ -299,8 +299,89 @@ def __init__(self, name: str, address: str, visalib=None,
...
+						   set_cmd=self._theta_saferamp_set,
+						   get_cmd=self._theta_saferamp_get)
+		
+        self.add_parameter('r_saferamp',
+						   label='B_Radius',
+						   unit='T',
+						   set_cmd=self._r_saferamp_set,
+						   get_cmd=self._r_saferamp_get)
+
+        self.add_parameter('r_simulramp',
```
astafan8: shall there be also `phi_simulramp` and `theta_simulramp`? Also, shall there be corresponding parameters for x/y/z/rho?

qcodes/instrument_drivers/oxford/MercuryiPS_VISA.py
```diff
@@ -299,8 +299,89 @@ def __init__(self, name: str, address: str, visalib=None,
...
+
         self.connect_message()
 
+    def _theta_saferamp_set(self, target):
+        """
+		Set the theta target and ramp to it
+		"""
+        self.theta_target(target)
+        self.ramp(mode='safe')
+        self.theta_measured()
```
astafan8: what is this call needed for? in order to update the value of the parameter, i guess? how is it used by Loop?

ectof:
I have added a couple of commits, that I think address the comments.

For the blocking ramp parameters (X_saferamp etc.), these were primarily useful for the legacy dataset as convenience functions. They are less useful for the new dataset where it's easier to reason inside a loop.

However, I think a universal update call is important and it should be called at the end of the blocking ramps. Otherwise the coordinates get stuck in an inconsistent state, that gets written to metadata etc., updating this shouldn't be left to the user. Perhaps later the code could be refactored in terms of properties, but this seems a reasonable workaround for now

ectof:
Sorry, I just fixed a couple more bugs.
Yes it's been tested on T11, I think everything is working OK

astafan8:
@ectof I made a commit because type checker was showing errors - using private method instead of lambdas. Could you quickly test this change on the real instrument?

ectof:
@astafan8 I just checked on the instrument and it works as expected. Thanks!

PR #1468, opened 2019-02-12
[wip] [discussion] Dev/move deps to interdependencies

WilliamHPNielsen:
This PR may replace #1438 

The idea is to move all information about how parameters are related to each other into a new `InterDependencies` object and thus out of the the `ParamSpec`.

- Make `ParamSpec` only have four attributes: name, datatype, label, unit
- Make `InterDependencies` take three input arguments: dependencies, inferences, and standalones
- Make the type of those arguments do most of the validation (nice, huh?)

All validation takes place upon instantiating an `InterDependencies` object, i.e. if it exists, it is valid.

You can get an impression of how this will work by looking at the `ParamSpecBase` and the `InterDependencies_` objects of this PR. If you like it, we can begin to discuss how to proceed with rolling this out, deprecating old things etc. I have a plan for that, but let's first decide whether we like this design.

Open questions:

- [ ] How should the new `InterDependencies` object serialize itself? Do we need a DB upgrade?
- [ ] Are the rules imposed by `InterDependencies_` for how parameters may relate to each other the correct rules. I _think_ they are, but I'd like your input on that.

@QCoDeS/core 

AB#2861

### DIFF
--- a/qcodes/dataset/dependencies.py
+++ b/qcodes/dataset/dependencies.py
@@ -1,8 +1,8 @@
 from typing import Dict, Any, Tuple, Sequence
 
-from qcodes.dataset.param_spec import ParamSpec
+from qcodes.dataset.param_spec import ParamSpecBase, ParamSpec
 
-ParamSpecTree = Dict[ParamSpec, Tuple[ParamSpec, ...]]
+ParamSpecTree = Dict[ParamSpecBase, Tuple[ParamSpecBase, ...]]
 
 
 class InterDependencies_:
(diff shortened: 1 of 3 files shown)


WilliamHPNielsen:
Just to flesh it out a bit, you would use it like this:

```python

ps1 = ParamSpec(name='dac_ch1_cv', paramtype='numerical', label='Gate Voltage', unit='V')
ps2 =  ParamSpec(name='dac_ch2_cv', paramtype='numerical', label='Plunger Voltage', unit='V')
ps3 = ParamSpec(name='lockin1_g', paramtype='numerical', label='conductance', unit='e^2/h')

idps = InterDependencies(dependencies={ps3: (ps1, ps2)}
```

Dominik-Vogel:
As we discussed over lunch the other day, I really like the idea. This way we can keep the `ParamSpec` and the `Interdependencies` both imutable.
Serialization is a good question.
Relevant for that may be:
How will Paramspecs be used? Shall we ask/get users to always recycle when possible?
Maybe we need a ParamSpecID

WilliamHPNielsen:
AB#2861

WilliamHPNielsen:
Closed in favour of #1477 using the cool new draft PR feature

PR #1469, opened 2019-02-13
Fix build in Azure devops

jenshnielsen:


jenshnielsen:
@QCoDeS/core This should be ready for review once CI passes

PR #1471, opened 2019-02-13
Alazar: refactor DLL API related things

astafan8:
### Main point of the PR

This PR extracts all C-level calls to the Alazar SDK into its own class to make it easier for that class to manage the use of background threads (and other stuff, if needed, like asyncio), and remove all these low-level details out of the driver class to make the latter manageable.

This is done by:
* Make the driver use a dedicated `AlazarAPIAPI` class that is based on `WrappedDll` class and `DllWrapperMeta` metaclass.
  * DllWrapperMeta ensures that there is one instance of the class per given DLL path (kind of singleton).
  * `WrappedDll` uses `DllWrapperMeta` class:
    * creates lock and thread pool executor for the DLL calls
    * assigns ctypes to DLL function arguments and returns (specified in `signatures` attribute)
    * `WrappedDll` allows `AlazarATSAPI` methods (that represent Alazar ATS API functions) to accept qcodes `Parameters`, and makes them aware of `TraceParameters` of the alazar driver.
  * `AlazarATSAPI` class defines `ctypes` signatures for relevant DLL functions, as well as bound methods which represent them; the class also implements convenience methods which make work with the API functions more convenient from the driver class point of view.

Note that the new DLL wrapper class here is somewhat coupled to the Alazar usage, generalizing it to "any DLL" does not bring much value now.

Thanks to @cgranade for the initial bulk of the work in https://github.com/QCoDeS/Qcodes/pull/1379 PR. However, asyncio-related parts were not taken into this PR.

### Other changes in this PR:
* move many convenience methods from alazar ats driver class to alazar api class
* add minor tests for alazar api class
* add minor tests for alazar dma buffer class
* create a module with alazar api constants, populate it with ones that are used in the driver
* attempt to cleanup docstrings in the driver
* refactor get_idn of the alazar driver
* refactor everything related to querying capabilities from alazar driver class into CapabilityHelper class that the driver uses
* Fix get_num_channels alazar driver method for 4 and 8 channel configurations
* Make the Alazar example notebook runnable without any additional hardware
* use instance logger in alazar

### To-Do:
- [x] squash when merge
- [x] (thanks @jenshnielsen) run test acquisitions to see if the driver hasn't been broken :)
- [x] (thanks @jenshnielsen) run test acquisitions to see if the changes don't make it slower :)

### Epilogue:
From Alazar API header files:
```c++
#define MEDIMUM_EXTERNAL_CLOCK	0x00000003UL //TYPO
#define MEDIUM_EXTERNAL_CLOCK	0x00000003UL
```


qcodes/instrument_drivers/AlazarTech/dll_wrapper.py
```diff
@@ -0,0 +1,106 @@
...
+        c_func = getattr(self._dll, full_name)
+        future = self._executor.submit(
+            api_call_task,
+            self._lock, c_func,
+            partial(mark_params_as_updated, *args),
+            *normalize_params(*args)
+        )
+        return future.result()
+
+    def async_call(self: TApi, *args) -> signature.return_type:
```
astafan8: @QCoDeS/core do we follow yagni and remove all traces of async for now (in this PR)?

jenshnielsen: if you mean asyncio then yes I think so. (the driver inherently uses the async alazar c-api and that should not be removed).  If we do not remove asyncio support we will have to sort out the issues pointed out with async in the other pr first.

astafan8: (yes, i meant `asyncio`). ok, will remove `asyncio` for this PR.

astafan8:
@QCoDeS/core I'd like you to have a look at the code to ensure this PR is going the right direction. The work btw is almost done, see the to-do list. If you like the general structure - perfect.

qcodes/instrument_drivers/AlazarTech/ats_api.py
```diff
@@ -0,0 +1,409 @@
...
+import ctypes
+
+from .dll_wrapper import WrappedDll, Signature
+from .constants import BOARD_NAMES, REGISTER_READING_PWD, Capability
+
+
+# Define aliases for ctypes that match Alazar's notation.
+U8 = ctypes.c_uint8
+U32 = ctypes.c_uint32
+HANDLE = U32
```
Dominik-Vogel: Ok this is unrelated to this PR, but nevertheless worth bringing up: Should the `HANDLE` type not be U64, because it represents a `void*` on a 64bit plattform?

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -33,185 +32,38 @@ class AlazarTech_ATS(Instrument):
...
-        34: 'ATS9130',
-        35: 'ATS9352',
-        36: 'ATS9453',
-    }
-
     @classmethod
-    def find_boards(cls, dll_path: str=None) -> List[dict]:
-        """
-        Find Alazar boards connected
+    def find_boards(cls, dll_path: str = None) -> List[dict]:
```
Dominik-Vogel: should we not preserve the doc string here in an updated version?

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -1005,7 +716,8 @@ def get_num_channels(byte_rep: int) -> int:
         """
         one_channels = tuple(2**i for i in range(16))
         two_channels = (3, 5, 6, 9, 10, 12)
-        four_channels = (255,)
+        four_channels = (15,)
+        eight_channels = (255,)
         sixteen_channels = (65535,)
```
Dominik-Vogel: Could these lines also be accommodated in the constants.py? (maybe including the elif chain...)

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -1014,45 +726,18 @@ def get_num_channels(byte_rep: int) -> int:
...
             return 4
+        elif byte_rep in eight_channels:
+            return 8
         elif byte_rep in sixteen_channels:
             return 16
         else:
             raise RuntimeError('Invalid channel configuration supplied')
 
-
     def _read_register(self, offset: int) -> int:
```
Dominik-Vogel: shall we not preserve the doc strings?

qcodes/instrument_drivers/AlazarTech/ats_api.py
astafan8: this is an interesting question.

`AlazarApi.h` does say that `HANDLE` is `void *`, but in the `ats_api.py` from Alazar SDK `U32` is used for handles, for example:
```python
ats.AlazarGetBoardBySystemID.restype = U32  # <-- THIS IS NOT `void *`
ats.AlazarGetBoardBySystemID.argtypes = [U32, U32]
self.systemId = systemId
self.boardId = boardId
self.handle = ats.AlazarGetBoardBySystemID(systemId, boardId)
if self.handle == 0:
    raise Exception("Board %d.%d not found" % (systemId, boardId))
```

Frankly, i don't know what to do. I can try to substitute `U32` with `ctypes.c_void_p` type, and change the type annotation of handle from `int` to this everywhere.

But i think, i will actually just email Alazar and ask what the correct type should be.

qcodes/instrument_drivers/AlazarTech/ATS.py
astafan8: They are moved to the api class from the driver class, so they are preserved. The thing is that register manipulation are private methods, hence direct driver users should not use them, and since the actual code of the methods is not in the api class, i decided to keep the docstrings there, and not copy them over to save the driver class from the details basically.

Is it ok? Or shall i write some docstrings here, or copy-paste docstrings (don't want that actually)?

qcodes/instrument_drivers/AlazarTech/ATS.py
astafan8: yes, i think it's a good idea. i'll see how to refactor this.

qcodes/instrument_drivers/AlazarTech/ATS.py
Dominik-Vogel: That is fine I think  :-)

qcodes/instrument_drivers/AlazarTech/ATS.py
astafan8: sure!

qcodes/instrument_drivers/AlazarTech/ats_api.py
```diff
@@ -0,0 +1,409 @@
...
+    ## CLASS MEMBERS ##
+
+    # Only allow a single instance per DLL path.
+    __instances: Dict[str, TApi] = {}
+
+    def __new__(cls: Type[TApi], dll_path: str) -> TApi:
+        if dll_path in cls.__instances:
+            logger.debug(
+                f"Found existing ATS API instance for DLL path {dll_path}.")
+            return cls.__instances[dll_path]
```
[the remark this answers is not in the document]
astafan8: Yes, so the point is to have a singleton of the alazar api, basically. The reason for this is to ensure that the same lock guards dll library calls from all the "clients" of the api class.

> even though your original instance has been destroyed
I think you've spotted a bug - when an instance is destroyed, it does not get removed from the `__instances` dictionary. I'll fix it (and write a test for this part actually).

qcodes/instrument_drivers/AlazarTech/ats_api.py
astafan8: changed to `void *` after emailing with AlazarTech

qcodes/instrument_drivers/AlazarTech/ats_api.py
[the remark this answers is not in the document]
astafan8: Done. The whole "singleton" infrastructure is revisited and implemented properly now. And tested.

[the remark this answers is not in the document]
astafan8: Question though @QCoDeS/core. As far as `LoadLibrary` from `ctypes` is concerned, both `..\AlazarApi` and `..\AlazarApi.dll` would result in the same loaded library with even the same `_handle` value. But `DllWrapperMeta` will differentiate between them and create another instance of alazar api. Shall it be fixed with checking for explicit existence of the `dll_path`? See the note in this test https://github.com/QCoDeS/Qcodes/pull/1471/files#diff-098d9377f1f97ea1e441123b925e60d6R118

qcodes/instrument_drivers/AlazarTech/ATS.py
astafan8: Done.

astafan8:
@QCoDeS/core Could you have another look at the PR? I finally approached the "actual testing" stage. And sorry that this PR takes so much time. Please have a look at the updated description as well.

qcodes/instrument_drivers/AlazarTech/ats_api.py
```diff
@@ -0,0 +1,502 @@
...
+from qcodes.instrument.parameter import _BaseParameter as Parameter
+
+# Define aliases for ctypes that match Alazar's notation.
+U8 = ctypes.c_uint8
+U32 = ctypes.c_uint32
+HANDLE = ctypes.c_void_p
+
+POINTER_U8 = Any
+POINTER_U32 = Any
+
```
jenshnielsen: Could we give these slightly different names so that it's clear that. U8 and U32 are ctypes types while (POINTER_U8 and POINTER_U32) are just typing aliases with no implementation. I.e they are only used for typing, in the code you use POINTER(U8)

qcodes/instrument_drivers/AlazarTech/ats_api.py
```diff
@@ -0,0 +1,502 @@
+"""
+This module provides a class that encapsulates Alazar ATS API, 
+:class:`AlazarATSAPI`. The class is used to expose Alazar API functions
+of its C library in a python-friendly way.
+"""
+
+
+from typing import Dict, Tuple, Union, Any
```
jenshnielsen: Verified that all signatures in this file matches the latest sdk manual

qcodes/instrument_drivers/AlazarTech/utils.py
```diff
@@ -1,12 +1,24 @@
...
 class TraceParameter(Parameter):
     """
     A parameter that keeps track of if its value has been synced to
-    the Instrument by setting a flag when changed.
+    the ``Instrument``. To achieve that, this parameter sets
+    the ``_parameters_synced`` attribute of the ``Instrument`` to ``False``
+    when the value of the parameter is changed.
+
+    This parameter is useful for the Alazar card driver. Syncing parameters to
+    an Alazar card is not instance, hence it makes sense to first set the
```
jenshnielsen: Syncing parameters to an Alazar card is slow?

qcodes/instrument_drivers/AlazarTech/ats_api.py
jenshnielsen: I assume this is needed since you cant use POINTER(U8) in the type signature

astafan8: exactly, i wasted quite some time trying to figure out the way to do it properly, and failed so far. googling didn't help either, seems that people never do this.

Shall I rename them to `POINTER_c_uint32`/`POINTER_c_uint8` so that they look more like `ctypes` types?

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -33,185 +34,47 @@ class AlazarTech_ATS(Instrument):
...
         for system_id in range(1, system_count + 1):
-            board_count = dll.AlazarBoardsInSystemBySystemID(system_id)
+            board_count = api.boards_in_system_by_system_id(system_id)
             for board_id in range(1, board_count + 1):
-                boards.append(cls.get_board_info(dll, system_id, board_id))
+                boards.append(cls.get_board_info(api, system_id, board_id))
         return boards
 
-        # TODO(nataliejpg) this needs fixing..., dll can't be a string
+    # TODO(nataliejpg) this needs fixing..., dll can't be a string
```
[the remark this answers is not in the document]
astafan8: indeed.

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -1,29 +1,30 @@
...
 from qcodes.instrument.parameter import Parameter
+from .ats_api import AlazarATSAPI
 from .utils import TraceParameter
+from .helpers import CapabilityHelper
+from .constants import NUMBER_OF_CHANNELS_FROM_BYTE_REPR
 
-logger = logging.getLogger(__name__)
 
-# TODO(damazter) (C) logging
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
astafan8: definitely!

Btw, would you allow to not pass ATS's logger instance to the underlying "api" class? being lazy here)

qcodes/instrument_drivers/AlazarTech/ats_api.py
astafan8: oh, thanks a lot!

qcodes/instrument_drivers/AlazarTech/utils.py
astafan8: yes, i'll fix the mistyping.

astafan8:
@jenshnielsen If you have time, i'd appreciate if you test this PR on your setup. I am also going to test it with my 9360 card. In particular, it would be great if you have some tests at hand which can show that the overall performance did not degrade.

qcodes/instrument_drivers/AlazarTech/ATS.py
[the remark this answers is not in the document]
jenshnielsen: Sure that sounds fine at least for now

astafan8:
After CI is done, i'll squash-merge it (rebasing ~2000 commits was not successful at all).

PR #1478, opened 2019-02-15
Reorganise API docs

jenshnielsen:
This is by no means complete but is a start towards a more sensible organization of the api docs


qcodes/instrument/parameter.py
```diff
@@ -548,11 +580,15 @@ def get_ramp_values(self, value: Union[float, int, Sized],
...
+        Validate the value supplied.
 
         Args:
-            value (any): value to validate
+            value: value to validate
 
+        Raises:
+            TypeError: If the value is of the wrong type.
+            ValueError: If the value is outside the bounds specified by the
+            validator.
```
astafan8: interesting that this line does not need extra indentation

qcodes/instrument/parameter.py
```diff
@@ -564,6 +600,16 @@ def validate(self, value: ParamDataType) -> None:
 
     @property
     def step(self) -> Optional[Number]:
+        """
+        Stepsize that this Parameter uses steps during set operations.
```
astafan8:
```suggestion
        Stepsize that this Parameter uses during set operation.
```

qcodes/instrument/parameter.py
```diff
@@ -669,13 +715,19 @@ def inter_delay(self, inter_delay: Number) -> None:
...
+        """
         return "_".join(self.name_parts)
 
     def set_validator(self, vals):
         """
-            Deprecated Set a validator `vals` for this parameter.
-                Args:
-                    vals (Validator):  validator to set
+        Deprecated Set a validator `vals` for this parameter.
+        Replace the `vals` method directly instead.
```
astafan8:
```suggestion
        
        Deprecated - reassign the `vals` attribute directly instead.
```

docs/api/index.rst
```diff
@@ -12,6 +12,7 @@ QCoDeS API
    config
    dataset
    instrument
+   parameters
    logger
    math
    monitor
```
astafan8: shall the order here make sense? i see it's not alphabetical, but i can't see what the current logic is. i'd suggest to go from config through parameters and instruments to measurement and dataset, with monitor, logger and other stuff inbetween where relevant.

docs/api/dataset.rst
```diff
@@ -0,0 +1,101 @@
...
+   :members:
+
+
+plotting
+--------
+
+.. automodule:: qcodes.dataset.plotting
+   :members:
+
+sqlite_base
```
astafan8: do we really want to generate docs for everything? i thought that for example sqlite_base is something internal that noone should use outside qcodes internals, am i wrong?

qcodes/instrument/parameter.py
```diff
@@ -669,13 +715,19 @@ def inter_delay(self, inter_delay: Number) -> None:
...
+        submodule that the parameter may be bound to.
+        """
         return "_".join(self.name_parts)
 
     def set_validator(self, vals):
         """
-            Deprecated Set a validator `vals` for this parameter.
-                Args:
-                    vals (Validator):  validator to set
+        Deprecated Set a validator `vals` for this parameter.
```
astafan8: some punctuation for readability
```suggestion
        (Deprecated) Set a validator `vals` for this parameter.
```

docs/api/index.rst
jenshnielsen: That is why this is WIP

docs/api/dataset.rst
jenshnielsen: This is still undecided and why this is marked wip

qcodes/instrument/parameter.py
jenshnielsen: It probably does

jenshnielsen:
I think this is ready now. 

* There is a `.rst` file per python file (module) that we document
* These are  kept in a folder structure that matches the code
* We only document the relevant modules. (Apart from utils etc where we document everything) 
* The individual python files are documented with a `automodule` clause. This means that everything is 
included for a module. We could get better results by going more fine grained and chose the specific classes to document but I think this is the best trade off between maintainability and granularity)

docs/api/dataset/index.rst
```diff
@@ -0,0 +1,30 @@
...
+
+.. autosummary::
+
+    qcodes.dataset
+    qcodes.dataset.measurements
+    qcodes.dataset.plotting
+    qcodes.dataset.data_set
+    qcodes.dataset.database_extract_runs
+    qcodes.dataset.legacy_import
+    qcodes.dataset.param_spec
```
astafan8: as per some discussions last year, i still don't think that ParamSpec should be exposed to the users. at least for now.

docs/api/instrument/index.rst
```diff
@@ -0,0 +1,25 @@
+.. _instrument_api :
+
+qcodes.instrument (Instruments)
+===============================
+
+.. autosummary::
+
+    qcodes.instrument
+    qcodes.instrument.visa
+    qcodes.instrument.ip
```
astafan8: "creating instrument drivers" says
> - `IPInstrument` - a deprecated driver just for ethernet connections. Do not use this; use `VisaInstrument` instead.

Shall we not include it in the docs then?

docs/api/utils/index.rst
```diff
@@ -0,0 +1,36 @@
...
+   :hidden:
+
+   command
+   deprecate
+   helpers
+   magic
+   metadata
+   plotting
+   slack
+   threading
```
astafan8: so what are the steps for the developer once he adds a new python module that deserves being included in the docs? perhaps, those can be added to "contributing" or wherever?

jenshnielsen: Yes that would be great once that is rewritten.

astafan8: can a rudimentary description be a part of this PR?

astafan8:
Is it possible and is it preferred to go from
```
qcodes.dataset.database_extract_runs.extract_runs_into_db(source_db_path: str, target_db_path: str, *run_ids, upgrade_source_db: bool = False, upgrade_target_db: bool = False) → None
```
to 
```
extract_runs_into_db(source_db_path: str, target_db_path: str, *run_ids, upgrade_source_db: bool = False, upgrade_target_db: bool = False) → None
```
when generating docs?

jenshnielsen:
> 
> 
> Is it possible and is it preferred to go from
> 
> ```
> qcodes.dataset.database_extract_runs.extract_runs_into_db(source_db_path: str, target_db_path: str, *run_ids, upgrade_source_db: bool = False, upgrade_target_db: bool = False) → None
> ```
> 
> to
> 
> ```
> extract_runs_into_db(source_db_path: str, target_db_path: str, *run_ids, upgrade_source_db: bool = False, upgrade_target_db: bool = False) → None
> ```
> 
> when generating docs?

I would mostly be against that. Explicit is better than implicit and all that

docs/api/utils/index.rst
jenshnielsen: Yes, but the problem is not the description. The problem is that it would land in the middle of another document where the rest no longer applies

astafan8: i'd be ok with that because without the description i'd have to bother you when i make "my next PR" :) Full-blown proper fixing of that document can be done later.

jenshnielsen: Ok I guess it should go here https://qcodes.github.io/Qcodes/community/contributing.html#development ?

astafan8: good! thanks!

astafan8:
I'm just going to throw them here since i've found them, but it's up to the author to decide what to do with them.

Have problems with formatting:
* class qcodes.utils.validators.Validator
* qcodes.utils.plotting.Number has a docstring which does not correspond to it
* qcodes.utils.plotting.auto_range_iqr - interesting "return type" entry, i don't know how it should be actually, maybe it's ok and just confusing for me
* qcodes.utils.helpers.NumpyJSONEncoder - list of supported conversions in "default" method is badly formatted. Also, the docstring includes the base class docstring which i think is irrelevant here.
* station.remove_component has an extra bullet point for "raises" section
* field_vector, set_component method: extra bullet point in the "parameters" section
* qcodes.logger.log_analysis - some of the "returns" sections are not formatted correctly (they look like quotes instead)
* qcodes.logger.instrument_logger.InstrumentLoggerAdapter docstring can be improved to have code/examples/plain text distinguished
* qcodes.instrument.group_parameter.GroupParameter - it's a bit bad than the doc is polluted with the methods of the parent class
* qcodes.instrument.parameter module has a nice module-level doc but cool stull like ParameterWithSetpoints is not mentioned there while ArrayParameter is
* qcodes.instrument.channel.AutoLoadableChannelList - the bullet-point list in the docstring is not renderred as such

jenshnielsen:
@astafan8 I was going to suggest that we should split up all the modules and each do a pr per module to bring the docs up to spec for that module. 

* That would be fix the issues that you have seen
* Add missing module level docstrings
* Add other missing docstrings

I really just wanted this to be about a sane overall layout

astafan8:
> we should split up all the modules and each do a pr per module to bring the docs up to spec for that module

Perfect! Let's do that soon.

jenshnielsen:
I will do the contributing guild updates in a new pr now

PR #1480, opened 2019-02-20
Dev/b2200 switch matrix multiple channel connect

StefanD986:
New feature for B2200 switch matrix driver:

- Adds methods to (dis)connect multiple input->output paths with one call (`connect_paths()` and `disconnect_paths()`)
- This is (much) faster and nicer than calling `connect()` multiple times in a loop.

Includes tests and updated example notebook.

@astafan8


qcodes/instrument_drivers/Keysight/keysight_b220x.py
```diff
@@ -4,6 +4,7 @@
 
 from qcodes import VisaInstrument
 from qcodes.utils.validators import MultiType, Ints, Enum, Lists
+from typing import List, Tuple
 
 
 def post_execution_status_poll(func):
```
astafan8: as the mypy output of the CI suggests, you need to add type hints to this decorator. you can read about this online, but perhaps smth simple like `def post_execution_status_poll(func: Callable) -> Callable:` could work.

PR #1482, opened 2019-02-20
Feature: give DataSet an explicit start method

WilliamHPNielsen:
This PR is sort of a pave-the-way PR that will make #1477 easier to implement. Having said that, this PR also fixes some annoying issues (see below)

The main idea here is that a `DataSet` can be in one of three states: pristine, running (started but not yet completed), and completed. When the `DataSet` is pristine, no results can be added to it. Its parameters may still be modified, i.e. parameters can be added and removed. The `DataSet.mark_started()` call then writes the parameters to the DB. Now results may be added and the parameters can no longer change. 

- Introduce `DataSet.mark_started()` and make it obligatory to call before adding results
- Delay writing the parameters onto the DB until the `DataSet` is marked as started
- Remove `DataSet` methods that have been deprecated for at least 3 months

This PR fixes the current unwanted behaviour:
- Parameters may be added at any time (also for completed runs) and immediately modify the `results_table`
- It is an error to call `add_result` on a completed run, but *not* to call `add_results` on a completed dataset

Open questions [EDIT: The PR does as the question suggests now]:
- Should we also put writing the `run_timestamp` into the `_perform_start_actions`? I'd say yes, since that time is closer to being the "true" start time than when the `DataSet` object was created. It will also make it unambiguous whether a `DataSet` has been started when loading it from disk; simply see whether `run_timestamp` is NULL. 

@QCoDeS/core


qcodes/tests/dataset/test_dataset_loading.py
```diff
@@ -123,11 +138,14 @@ def test_completed_timestamp_for_not_completed_dataset():
...
+    assert ds.pristine is True
+    assert ds.started is False
+    assert ds.running is False
+    assert ds.completed is False
 
-    assert None is ds.completed_timestamp_raw
+    assert ds.completed_timestamp_raw is None
 
-    assert None is ds.completed_timestamp()
+    assert ds.completed_timestamp() is None
```
astafan8: you'd be surprised but the reason to write `assert None is ...` is that for some reason the expected order is `assert expected is actual`, not `assert actual is expected`, as per how pytest creates this insight into the compared values.

qcodes/dataset/data_set.py
```diff
@@ -540,6 +546,14 @@ def add_snapshot(self, snapshot: str, overwrite: bool=False) -> None:
             log.warning('This dataset already has a snapshot. Use overwrite'
                         '=True to overwrite that')
 
+    @property
+    def pristine(self) -> bool:
+        return not(self._started or self._completed)
+
+    @property
+    def running(self) -> bool:
```
astafan8: shall here be docstrings for these 3 properties?

WilliamHPNielsen: Ah, docstrings, yes, I've heard of those :)

WilliamHPNielsen:
@QCoDeS/core I'd say this is done now. Give it your worst!

jenshnielsen:
Looks good, As we discussed offline I think it would be good with tests that covers the 4 errors that are raised with respect to pristine. In the code cove they look uncovered at the moment

WilliamHPNielsen:
Okay, now the diff coverage is 100%.

qcodes/dataset/data_set.py
```diff
@@ -353,7 +369,7 @@ def sample_name(self) -> str:
         return get_sample_name_from_experiment_id(self.conn, self.exp_id)
 
     @property
-    def run_timestamp_raw(self) -> float:
+    def run_timestamp_raw(self) -> Optional[float]:
```
astafan8: Too late but ---- the same kind of change has to also be made for `def run_timestamp` method - it has to handle the None case nicely (perhaps also return None instead of formattted time string). I wonder why it wasn't caught by tests :(((

WilliamHPNielsen: Aiiii, good catch!

The `run_timestamp` will work for `None`; that just gives CURRENT TIME. Silent data corruption, omg everyone freak out!

I'll make a quickfix PR.

PR #1483, opened 2019-02-21
Delete old notebook

WilliamHPNielsen:
The `example-creation` notebook is written in great haste and does not really contribute with anything not covered in the `The Experiment Container` notebook. So in the interest of not confusion users and having less old crap to maintain, I propose that we delete it.

@QCoDeS/core


jenshnielsen:
This looks fine but there is a link to this notebook in https://github.com/QCoDeS/Qcodes/blob/master/docs/dataset/introduction.rst that needs to be replaced

WilliamHPNielsen:
Whoops, that is a good point!

PR #1484, opened 2019-02-21
Add complex column for "scalar" complex data

jenshnielsen:
Complex array data types are already handled correctly. 

This is one of the possible solutions. Store complex numbers in a blob for each row similar to how arrays are stored. 

Possible alternatives.

* A complex row type only for complex values
* An interface that maps 1 complex number to 2 columns automatically

@QCoDeS/core What do you think should we go this way or try an alternative solution?


qcodes/dataset/sqlite_base.py
```diff
@@ -246,7 +246,12 @@ def _convert_numeric(value: bytes) -> Union[float, int, str]:
         # if the reason was the conversion to float, and, if so, we are sure
         # that we need to return a string
         if "could not convert string to float" in str(e):
```
jenshnielsen: This logic is somewhat convoluted and the biggest downside of this solution

jenshnielsen:
@WilliamHPNielsen and I had an offline discussion and agreed that it's better with a specific complex column datatype

qcodes/dataset/measurements.py
```diff
@@ -179,6 +181,8 @@ def add_result(self, *res_tuple: res_type) -> None:
                 value = np.atleast_1d(value)
                 array_size = len(value.ravel())
                 if param_spec.type != 'array' and array_size > 1:
+                    # TODO this does not check that the dtype matches the sqlite
+                    # column type
```
jenshnielsen: This is not a bug in this pr but does never the less need to be addressed

qcodes/dataset/sqlite_base.py
```diff
@@ -271,6 +276,13 @@ def _adapt_float(fl: float) -> Union[float, str]:
     return float(fl)
 
 
+def _adapt_complex(value: Union[complex, np.complex]) -> sqlite3.Binary:
```
astafan8: Is this type annotation complete?

also, shall we make a `Complex` type as `Union[complex, np.complex, np.complex64, ...]`?

jenshnielsen:
> 
> 
> Is the idea that 'complex' is sort of the same as 'numeric' (as in not 'array') but just represents complex numbers? if so, are there any complications for 'array' of complex numbers that we need to care about (i can think of only updating docs everywhere saying "for storing complex arrays (not scalars) just use 'array' type")?

Yes that is a separate task to add tests and docs for complex array parameters (which already seems to work)

qcodes/dataset/sqlite_base.py
jenshnielsen: Yes

qcodes/utils/validators.py
```diff
@@ -301,6 +301,28 @@ def validate(self, value: numbertypes, context: str = '') -> None:
...
+    validtypes = (complex, np.complex, np.complex64, np.complex128)
+
+    def __init__(self) -> None:
+
+        self._valid_values = ((1+1j), )
+
+    def validate(self, value: numbertypes, context: str = '') -> None:
+        if not isinstance(value, self.validtypes):
+            raise TypeError(
+                '{} is not an int or float; {}'.format(repr(value), context))
```
```suggestion
                '{} is not complex; {}'.format(repr(value), context))
```

qcodes/utils/validators.py
```diff
@@ -301,6 +301,28 @@ def validate(self, value: numbertypes, context: str = '') -> None:
         super().validate(castvalue, context=context)
 
 
+class ComplexNum(Validator):
```
jenshnielsen: The validator needs a test

qcodes/dataset/measurements.py
```diff
@@ -190,17 +194,23 @@ def add_result(self, *res_tuple: res_type) -> None:
...
                                      f"because its type class is "
-                                     f"'text', not 'numeric'.")
+                                     f"{param_spec.type!r}, not 'numeric'.")
             elif isinstance(value, str):
                 if param_spec.type != 'text':
                     raise ValueError(f"It is not possible to save a string "
                                      f"value for parameter {paramstr!r} "
                                      f"because its type class is "
                                      f"{param_spec.type!r}, not 'text'.")
+            elif isinstance(value, (complex, np.complex, np.complex128, np.complex64)):
```
astafan8: shall we define `complex_types = (complex, np.complex, np.complex128, np.complex64)` in, say, `qcodes.math.<some module>` and re-use them everywhere in the code? (i actually wonder why numpy does not have such a definition in the library for the convenience of the humanity...)

qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -1545,6 +1574,66 @@ def test_datasaver_2d_multi_parameters_array(channel_array_instrument):
...
+
+    with meas.run() as datasaver:
+        setparam.set(0)
+        with pytest.raises(ValueError, match=expected_msg):
+            datasaver.add_result((setparam, setparam()),
+                                 (param, param()))
+
+
+@pytest.mark.usefixtures("experiment")
+def test_save_complex_num_as_complex_raises(complex_num_instrument):
```
astafan8:
```suggestion
def test_save_numeric_as_complex_raises(complex_num_instrument):
```

qcodes/dataset/measurements.py
```diff
@@ -190,17 +194,23 @@ def add_result(self, *res_tuple: res_type) -> None:
...
-                if param_spec.type == 'text':
+                if param_spec.type != 'numeric':
                     raise ValueError(f"It is not possible to save a numeric "
                                      f"value for parameter {paramstr!r} "
                                      f"because its type class is "
-                                     f"'text', not 'numeric'.")
+                                     f"{param_spec.type!r}, not 'numeric'.")
             elif isinstance(value, str):
                 if param_spec.type != 'text':
                     raise ValueError(f"It is not possible to save a string "
```
astafan8: just wondering, do we have parametrization on existing param_types for tests for these exceptions? if so, are we sure that the tests are executed also for the new 'complex' paramtype ?

jenshnielsen: I think we should completely rewrite this logic

jenshnielsen:
@WilliamHPNielsen It would be great to look at this together to make sure that I understand the refactoring that you did correctly

qcodes/dataset/measurements.py
jenshnielsen: Moved this to utils and prepared for the other complex pr

astafan8: thanks!

WilliamHPNielsen:
It looks good and correct, but perhaps we could add tests where we save a complex parameter alongside an array and where we save a list of complex numbers (although I hate that behaviour, it should be supported for now). Just to be sure that complex number do indeed work as scalars and strings in terms of the unraveling/raveling logic.

qcodes/dataset/measurements.py
```diff
@@ -411,6 +414,8 @@ def reshaper(val, paramtype):
                 return float(val)
             elif paramtype == 'text':
                 return str(val)
+            elif paramtype == 'complex':
+                return complex(val)
```
jenshnielsen: Even if this is perhaps a bit pathological it should have a test

astafan8: ❤️

qcodes/utils/validators.py
```diff
@@ -301,6 +303,28 @@ def validate(self, value: numbertypes, context: str = '') -> None:
...
```
jenshnielsen: Should this be renames ComplexNumbers for consistency

astafan8: yes :)

qcodes/dataset/measurements.py
```diff
@@ -47,9 +47,11 @@ class ParameterTypeError(Exception):
 
 def is_number(thing: Any) -> bool:
     """
-    Test if an object can be converted to a number UNLESS it is a string
+    Test if an object can be converted to a float UNLESS it is a string or
+    complex.
```
jenshnielsen: This is actually unused by now. After #1477. I will deprecate in another pr

jenshnielsen:
@QCoDeS/core I think this is now functionally complete. I don't like how many places this hooks into the api but I believe it is correct and has sufficient tests

qcodes/dataset/sqlite_base.py
```diff
@@ -1350,10 +1368,13 @@ def get_parameter_data(conn: ConnectionPlus,
 
         # if we have array type parameters expand all other parameters
         # to arrays
-        if 'array' in types and ('numeric' in types or 'text' in types):
+        if 'array' in types and ('numeric' in types or 'text' in types
+                                 or 'complex in types'):
```
astafan8: Looks like typo. Is it caught by any tests? what kind of tests would catch this?
```suggestion
                                 or 'complex' in types):
```

qcodes/utils/validators.py
```diff
@@ -301,6 +303,28 @@ def validate(self, value: numbertypes, context: str = '') -> None:
...
+    def __init__(self) -> None:
+
+        self._valid_values = ((1+1j), )
+
+    def validate(self, value: numbertypes, context: str = '') -> None:
+        if not isinstance(value, self.validtypes):
+            raise TypeError(
+                '{} is not complex; {}'.format(repr(value), context))
+
+    is_numeric = False  # there is on meaningful way to sweep a complex number
```
```suggestion
    is_numeric = False  # there is no meaningful way to sweep a complex number
```

qcodes/dataset/sqlite_base.py
```diff
@@ -24,13 +24,15 @@
...
+from qcodes.utils.types import complex_types
+
 
 log = logging.getLogger(__name__)
 
 # represent the type of  data we can/want map to sqlite column
 VALUE = Union[str, Number, List, ndarray, bool]
 VALUES = List[VALUE]
-
+complex_type_union = Union[complex, np.complex, np.complex64, np.complex128]
```
astafan8: shall this be moved to `qcodes/utils/types.py`?

qcodes/dataset/sqlite_base.py
jenshnielsen: As I understand it we cannot write a test that would catch this. The non empty string always evaluates to True. However removing the the `and ...` part of the if statement will only result in a slight performance degradation so there is no way of testing for it

astafan8: exactly, i got similar conclusions as well ((

anyway, this is yet another example that confirms that "code review" is important and necessary for ensuring code quality blablablabla.

PR #1485, opened 2019-02-21
Docs/better parameter docs

jenshnielsen:
Pulled the improvements to the api docs out of #1478 To make that pr more focused. We can certainly improve these docs further but I think this is already a significant update


qcodes/instrument/group_parameter.py
```diff
@@ -1,3 +1,10 @@
+"""
+This module implements a :class:`Group` intended to hold multiple
+parameters that are to be get and set by the same command. The parameters
```
WilliamHPNielsen: I guess we use "gotten" instead of "get" here? I know that it's a controversial word, but I think that it is what we normally use.

qcodes/instrument/parameter.py
```diff
@@ -568,6 +604,16 @@ def validate(self, value: ParamDataType) -> None:
 
     @property
     def step(self) -> Optional[Number]:
+        """
+        Stepsize that this Parameter uses during set operation.
+        If step is a positive number, this is the maximum value change
```
WilliamHPNielsen: What if step is a negative number? Do we raise?

qcodes/instrument/parameter.py
```diff
@@ -673,13 +719,20 @@ def inter_delay(self, inter_delay: Number) -> None:
 
     @property
     def full_name(self) -> str:
+        """
+        Name of the parameter including the name of the instrument and
+        submodule that the parameter may be bound to.
```
WilliamHPNielsen: Let's go all in and describe how the extra names are included. Maybe like this:

```
"""
Name of the parameter including the name of the instrument and
submodule that the parameter may be bound to. The names are separated
by underscores, like this: 'instrument_submodule_parameter'.
"""
```

qcodes/instrument/parameter.py
jenshnielsen: This is documented below in the nice docstring of the setter, which unfortunately is never used in the rendered docs 😛

qcodes/instrument/group_parameter.py
astafan8: "gotten" and "setten" ? :)

I'd actually agree with that especially because it allows to distinguish between "get" method and "get" word.

WilliamHPNielsen: Setten? -Get outta here!

PR #1487, opened 2019-02-22
Add linux to azure builds

jenshnielsen:


qcodes/tests/test_plot_utils.py
```diff
@@ -3,7 +3,9 @@
 """
 
 from pytest import fixture
```
jenshnielsen: Not needed should be reverted

jenshnielsen:
This seems to work but is way more ugly than I hoped 😞

jenshnielsen:
@astafan8 All the steps that are the same are reused. It's just that many of the steps are only almost identical 😞

astafan8:
@jenshnielsen shall we get this in?

i don't have anything against this as long as we are sure that the CI for the two platforms are executed in parallel (not sequentially).

jenshnielsen:
They run parallelly but I think we should clean this up a bit more before merging.

jenshnielsen:
Given the current issues with conda activation. I think this is as clean as its going to get until something is fixed externally. I added a comment to explain the current state

PR #1489, opened 2019-02-25
Add support for complex values to Array validator

jenshnielsen:
In the process discovered that the original typecheking for real values was actually not working so fixed that and added a test for it.


qcodes/tests/test_validators.py
```diff
@@ -580,6 +580,79 @@ def test_type(self):
...
+        a.validate(np.arange(10, dtype=np.complex128))
+        with self.assertRaises(TypeError):
+            a.validate(np.arange(10, dtype=np.complex64))
+
+    def test_real_subtypes(self):
+        """
+        Test that validating a concrete real type into an array that
+        only support other concrete types raises as expected
+        """
+        types = [np.int8, np.int16, np.int32, np.int64,
```
jenshnielsen: We should probably extract all these types to some common place in utils

jenshnielsen:
@QCoDeS/core I think this is ready for review

qcodes/utils/validators.py
```diff
@@ -488,32 +488,72 @@ def __repr__(self) -> str:
...
+        if any(np.issubsctype(type(min_value), real_type) for real_type in
+               self.real_types):
+            self._min_value = min_value
+        else:
+            raise TypeError('min_value must be a (real) number and the '
+                            'validator must only allow real (non complex) '
+                            'data types')
+
+        if any(np.issubsctype(type(max_value), real_type) for real_type in
+               self.real_types):
```
astafan8: why is it `real_types` and not `valid_types`? if it's array of integers, should the min/max be allowed to be float (i'd say, no)? if it's array of floats, should min/max be allowed to be integers (i'd say, yes, but that's not very clean)?

qcodes/utils/validators.py
```diff
@@ -488,32 +488,72 @@ def __repr__(self) -> str:
...
+            raise TypeError("min and max_value is not supported for complex "
+                            "types.")
+
+        if any(np.issubsctype(type(min_value), real_type) for real_type in
+               self.real_types):
+            self._min_value = min_value
+        else:
+            raise TypeError('min_value must be a (real) number and the '
+                            'validator must only allow real (non complex) '
+                            'data types')
```
astafan8: The second part of the exception message `and the validator must only allow real (non complex) data types` is not checked in this particulat `if`, but is checked above. Isn't it redundant here?

Also, i guess it might help user if the error message contains smth like `the type of min_value is {type(min_value)}`, m?)

qcodes/utils/validators.py
```diff
@@ -488,32 +488,72 @@ def __repr__(self) -> str:
...
+
+    Min and max validation is not supported for complex numbers.
 
     Args:
         min_value:  Min value allowed, default inf.
         max_value: Max value allowed, default inf.
         shape: The shape of the array, tuple of either ints or Callables taking
             no arguments that return the size along that dim as an int.
+        valid_types: Sequence of types that the validator should support. Should
+            be a subset of the supported types or None. If None all real
```
astafan8: Punctuation facilitates reading)
```suggestion
            be a subset of the supported types, or None. If None, all real
```

qcodes/utils/validators.py
jenshnielsen: Because that would be an api change

jenshnielsen: Because min and max cannot be complex. Complex numbers do not have partial order

astafan8: are you saying that previously it was allowed to specify the min/max with value of some type regardless of what `valid_types` are? wow.

qcodes/utils/validators.py
jenshnielsen: It is checked here. Thats why we check against real types.

There are two different cases that needs to be checked. 

* Trying to set min and max (to real numbers) for anything that accepts complex. That is done above.
* Trying to set the min value to a complex number. That is done here.

qcodes/utils/validators.py
jenshnielsen: But sure we can extend it to also check that min and max is of that same type

qcodes/tests/test_validators.py
```diff
@@ -1,13 +1,27 @@
...
+numpy_concrete_ints = (np.int8, np.int16, np.int32, np.int64,
+                       np.uint8, np.uint16, np.uint32, np.uint64)
+numpy_non_concrete_ints = (np.int, np.int_, np.uint, np.integer)
+
+numpy_concrete_floats = (np.float16, np.float32, np.float64)
+numpy_non_concrete_floats = (np.float, np.float_,
+                             np.floating)
+
+numpy_concrete_complex = (np.complex64, np.complex128)
+numpy_non_concrete_complex = (np.complex, np.complex_, np.complexfloating)
```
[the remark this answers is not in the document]
jenshnielsen: I would move them when we actually use them somewhere else

qcodes/utils/validators.py
astafan8: sure. but this (valid_types==complex + min/max are specified) is caught above, right?

My question is why the type of min/max value is being checked against all real_types instead of valid_types of the validator (the `self`)?

qcodes/utils/validators.py
```diff
@@ -488,32 +488,72 @@ def __repr__(self) -> str:
...
         shape: The shape of the array, tuple of either ints or Callables taking
             no arguments that return the size along that dim as an int.
+        valid_types: Sequence of types that the validator should support. Should
+            be a subset of the supported types, or None. If None, all real
+            datatypes will validate.
     """
 
-    validtypes = (int, float, np.integer, np.floating)
+    real_types = (np.integer, np.floating)
+    supported_types = real_types + (np.complexfloating,)
```
astafan8: police question - should these two be private or constant? as opposed to "just attributes"?

qcodes/utils/validators.py
astafan8: i get this: `* Trying to set the min value to a complex number. That is done here.`. but then why is `and the validator must only allow real (non complex) data types` mentioned in the error message?

qcodes/utils/validators.py
jenshnielsen: I don't think so. See comment below

astafan8: > to also check that min and max is of that same type

i think you can catch two hairs at the same time by saying "min and max have to be of that same type as the `valid_types`", would you agree?

qcodes/utils/validators.py
```diff
@@ -488,32 +488,98 @@ def __repr__(self) -> str:
...
+            self.valid_types)
+
+        max_value_is_valid_type = any(
+            np.issubsctype(type(max_value), valid_type) for valid_type in
+            self.valid_types)
+
+        if min_value_is_valid_type or min_value is None:
+            self._min_value = min_value
+        else:
+            raise TypeError(f'min_value must an instance of valid_types. '
```
```suggestion
            raise TypeError(f'min_value must be an instance of valid_types. '
```

qcodes/utils/validators.py
```diff
@@ -488,32 +488,98 @@ def __repr__(self) -> str:
...
+                            f'It is {min_value} of '
+                            f'type {type(min_value)}')
+
+        if max_value_is_valid_type or max_value is None:
             self._max_value = max_value
         else:
-            raise TypeError('max_value must be a number bigger than min_value')
+            raise TypeError(f'max_value must an instance of valid_types. '
```
```suggestion
            raise TypeError(f'max_value must be an instance of valid_types. '
```

qcodes/tests/test_validators.py
```diff
@@ -580,6 +596,119 @@ def test_type(self):
...
+        a = Arrays(valid_types=(np.complex128,))
+
+        a.validate(np.arange(10, dtype=np.complex128))
+        with pytest.raises(TypeError, match=r"is not any of "
+                                            r"\(<class 'numpy.complex128'>,\)"
+                                            r" it is complex64"):
+            a.validate(np.arange(10, dtype=np.complex64))
+
+    def test_min_max_real_ints_raises(self):
+        with pytest.raises(TypeError, match="min_value must an instance "
```
astafan8:
```suggestion
        with pytest.raises(TypeError, match="min_value must be an instance "
```

qcodes/tests/test_validators.py
```diff
@@ -580,6 +596,119 @@ def test_type(self):
...
+                                            "of valid_types."):
+            Arrays(valid_types=(np.integer,), min_value=1.0)
+        with pytest.raises(TypeError, match="max_value must an instance "
```
astafan8:
```suggestion
        with pytest.raises(TypeError, match="max_value must be an instance "
```

jenshnielsen:
@QCoDeS/core I think this is ready to land but needs a new review since I dismissed @astafan8 s original review. I don't think there are any important changes since then

astafan8:
Then I'll let the other core devs do this review.

jenshnielsen:
@astafan8 Just to be clear I did not dismiss the review because there was anything wrong with it :) Just because I did not want this to land before #1477 landed

astafan8:
@jenshnielsen sure, i didn't mean that :) just let's get some other pair of eyes to have a look at the change :)

qcodes/utils/validators.py
```diff
@@ -512,32 +512,98 @@ def __repr__(self) -> str:
...
-        min_value:  Min value allowed, default inf.
-        max_value: Max value allowed, default inf.
+        min_value:  Min value allowed, default None for which min value
+            check is not performed
+        max_value: Max value allowed, default None for which max value
+            check is not performed
         shape: The shape of the array, tuple of either ints or Callables taking
             no arguments that return the size along that dim as an int.
+        valid_types: Sequence of types that the validator should support. Should
+            be a subset of the supported types, or None. If None, all real
```
WilliamHPNielsen: It may be helpful to write here what the supported types are?

WilliamHPNielsen: Nevermind, it's written 11 lines above, I'm just blind :)

WilliamHPNielsen:
I think this looks good. Looking forward to see the first instrument use this.

PR #1491, opened 2019-02-26
Dont upgrade conda

jenshnielsen:
At the moment it seems like conda 4.6 is very unstable

Hopefully fixes ci builds on Azure

On the other hand there has definitely been issues where an old version of conda has not mixed well with new packages


jenshnielsen:
We went with #1476 instead

PR #1494, opened 2019-02-27
Inferred paramtype from Parameter and validator

jenshnielsen:
This is a suggestion to how this could be done. The tests will need some work at the very least


qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -1590,6 +1590,31 @@ def test_datasaver_2d_multi_parameters_array(channel_array_instrument):
...
+    # overwrite the default with sensible alternatives
+    assert Measurement._infer_paramtype(chan.dummy_array_parameter,
+                                        'numeric') == 'numeric'
+    assert Measurement._infer_paramtype(chan.dummy_parameter_with_setpoints,
+                                        'numeric') == 'numeric'
+    assert Measurement._infer_paramtype(chan.dummy_multi_parameter,
+                                        'array') == 'array'
+    assert Measurement._infer_paramtype(chan.dummy_2d_multi_parameter,
+                                        'array') == 'array'
+
```
jenshnielsen: We are missing a test for string type validators

jenshnielsen:
@QCoDeS/core  Could anyone have a look at this?

qcodes/dataset/measurements.py
```diff
@@ -625,22 +626,29 @@ def register_parameter(
             basis: The parameters that this parameter is inferred from. If
                 this parameter is not inferred from any other parameters,
                 this should be left blank.
-            paramtype: type of the parameter, i.e. the SQL storage class
+            paramtype: type of the parameter, i.e. the SQL storage class,
+                If None the paramtype will be inferred from the paramter type
```
WilliamHPNielsen: "paramter" -> "parameter"

qcodes/dataset/measurements.py
```diff
@@ -625,22 +626,29 @@ def register_parameter(
...
+                and the validator.
```
WilliamHPNielsen: I would propose to write "and the validator of the supplied parameter" to be fool-proof.

qcodes/dataset/measurements.py
```diff
@@ -625,22 +626,29 @@ def register_parameter(
             basis: The parameters that this parameter is inferred from. If
                 this parameter is not inferred from any other parameters,
                 this should be left blank.
-            paramtype: type of the parameter, i.e. the SQL storage class
+            paramtype: type of the parameter, i.e. the SQL storage class,
+                If None the paramtype will be inferred from the parameter type
+                and the validator of the supplied parameter" to be fool-proof.
```
astafan8:
```suggestion
                and the validator of the supplied parameter.
```

qcodes/dataset/measurements.py
```diff
@@ -669,7 +677,35 @@ def register_parameter(
...
+            paramtype = 'array'
+        elif isinstance(parameter, ArrayParameter) and paramtype is None:
+            paramtype = 'array'
+        elif isinstance(parameter.vals, vals.Strings) and paramtype is None:
+            paramtype = 'text'
+        # TODO complex case here once support is added
+        # should we try to figure out if parts of a multiparamter are
+        # arrays or something else?
+        # default to numeric as before
+        elif paramtype is None:
```
astafan8: I suggest to move the `if paramtype is None` to the caller - then the code of the inference method is purely about inferring.

qcodes/dataset/measurements.py
```diff
@@ -669,7 +677,35 @@ def register_parameter(
...
+        if isinstance(parameter.vals, vals.Arrays) and paramtype is None:
+            paramtype = 'array'
+        elif isinstance(parameter, ArrayParameter) and paramtype is None:
+            paramtype = 'array'
+        elif isinstance(parameter.vals, vals.Strings) and paramtype is None:
+            paramtype = 'text'
+        # TODO complex case here once support is added
+        # should we try to figure out if parts of a multiparamter are
+        # arrays or something else?
+        # default to numeric as before
```
astafan8:
```suggestion
        # default to numeric
```

qcodes/dataset/measurements.py
```diff
@@ -669,7 +677,35 @@ def register_parameter(
...
+        """
+        if isinstance(parameter.vals, vals.Arrays) and paramtype is None:
+            paramtype = 'array'
+        elif isinstance(parameter, ArrayParameter) and paramtype is None:
+            paramtype = 'array'
+        elif isinstance(parameter.vals, vals.Strings) and paramtype is None:
+            paramtype = 'text'
+        # TODO complex case here once support is added
+        # should we try to figure out if parts of a multiparamter are
+        # arrays or something else?
```
astafan8: nah... not for the current multiparameter. perhaps for the "new" one later in the future.

jenshnielsen: As we talked about this is a prerequisite for dropping the unraveling of array like parameters

jenshnielsen:
@astafan8 I tend to think that we are moving towards making the storage class an implementation detail that the user hopefully does not need to care about and hence we do not need to document it. but I could be wrong

astafan8:
@jenshnielsen I completely support the following:
> ... we are moving towards making the storage class an implementation detail that the user hopefully does not need to care about

So, to me this only means that the "user guide" surely does not need to mention it. However, for more advanced users (and for use as well) having this design/behavior documented would be helpful. If for some reason (say, performance optimizations) a user asks "I learned about these paramtypes, and it seems that 'array' would be faster for my case, but i see that it was saved as 'numeric'. why is that and how can i change it?", I'd like to be able to point him to some docs from which he will deduct that for his case he might want to explicitly pass a paramtype, or just set a proper validator of the parameter, or else.

Does this make sense? :)

WilliamHPNielsen:
Let's get this in, no?

PR #1496, opened 2019-03-01
Azure pipeline: independent steps

astafan8:
* Run mypy, pytest, db generation, docs build independent of each others success or failure (say, still run pytest if mypy failed)
* publish coverage results after pytest results, not after docs build
* make more steps succeededOrFailed instead of always
* add more display names
* finally ignore .vscode files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,6 +85,7 @@ tmp/
 docs/examples/data/*
 
 .idea/
+.vscode/
 
 # Mac files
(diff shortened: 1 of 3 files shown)


jenshnielsen:
👍 I think you can change the condition on publish test results to succeededOrFailed too (from always) If the pipeline is canceled there will not be anything to publish anyway

PR #1497, opened 2019-03-01
Cast array params to np array before inserting

jenshnielsen:
Otherwise this will not work correctly when inserting as arrays

### DIFF
--- a/qcodes/dataset/measurements.py
+++ b/qcodes/dataset/measurements.py
@@ -159,7 +159,7 @@ class DataSaver:
                                                   res,
                                                   found_parameters)
 
-        for partial_result in res:
+        for i, partial_result in enumerate(res):
             parameter = partial_result[0]
             paramstr = str(parameter)
             value = partial_result[1]
@@ -174,9 +174,10 @@ class DataSaver:
                 inserting_as_arrays = True
                 inserting_this_as_array = True
             if any(isinstance(value, typ) for typ in array_like_types):
-
                 value = cast(np.ndarray, partial_result[1])
                 value = np.atleast_1d(value)
+                # we want to insert the casted version
+                res[i] = (parameter, value)
                 array_size = len(value.ravel())
                 if param_spec.type != 'array' and array_size > 1:
                     inserting_unrolled_array = True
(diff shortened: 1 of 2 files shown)


qcodes/dataset/measurements.py
```diff
@@ -174,9 +174,10 @@ def add_result(self, *res_tuple: res_type) -> None:
                 inserting_as_arrays = True
                 inserting_this_as_array = True
             if any(isinstance(value, typ) for typ in array_like_types):
-
                 value = cast(np.ndarray, partial_result[1])
                 value = np.atleast_1d(value)
+                # we want to insert the casted version
```
astafan8: it's not only casted but also converted, right? for example, if `partial_result[1]` is a tuple, the tuple gets also converted to numpy array by `np.atleast_1d`? or does `cast(np.ndarray, ...` do that already?

jenshnielsen: Converting from one type to another is a cast. However the function `cast` is indeed badly named and does not perform any real cast but purely changes the nominal type. The real `cast` is indeed performed by `atleast1d` as you say

astafan8: thanks, good to know. i got pythonized too much))

qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -800,158 +800,151 @@ def test_datasaver_foul_input():
...
 
 @settings(max_examples=5, deadline=None)
 @given(N=hst.integers(min_value=5, max_value=500),
-       M=hst.integers(min_value=4, max_value=250))
+       M=hst.integers(min_value=4, max_value=250),
+       seed=hst.integers(min_value=0, max_value=np.iinfo(np.uint32).max))
 @pytest.mark.usefixtures("experiment")
-def test_datasaver_array_parameters(SpectrumAnalyzer, DAC, N, M):
-    spectrum = SpectrumAnalyzer.spectrum
+@pytest.mark.parametrize("param_type", ['array', 'tuple', 'list'])
```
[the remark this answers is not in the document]
astafan8: thanks!

jenshnielsen:
I believe that coverage drops because https://github.com/QCoDeS/Qcodes/blob/master/qcodes/dataset/measurements.py#L264 is no longer hit (the conversion is now taken care of by atleast_1d)

PR #1498, opened 2019-03-01
Use perf_counter, not monotonic for DB flushing

astafan8:
Let's see it works. perf_counter has more resolution than monotonic, which is better for smaller write_periods. Presumably.


WilliamHPNielsen:
Looks good. Any reason for this PR to be marked as draft? It looks pretty done to me.

PR #1500, opened 2019-03-04
Feature/dem 804/implement missing uhfli commands

qSaevar:
- Add support for enabling, and amplitude, for all devNNN/sigouts/ in ZIUHFL driver
- Change name of parameters `signal_outputN_enable` and `signal_outputN_amplitude` to `signalN_outputM_enable1` and `signalN_outputM_amplitude` in ZIUHFLI driver.

@WilliamHPNielsen

### DIFF
--- a/qcodes/instrument_drivers/ZI/ZIUHFLI.py
+++ b/qcodes/instrument_drivers/ZI/ZIUHFLI.py
@@ -929,14 +929,6 @@ class ZIUHFLI(Instrument):
                                 val_mapping={'ON': 1, 'OFF': 0},
                                 vals=vals.Enum('ON', 'OFF') )
 
-            self.add_parameter('signal_output{}_amplitude'.format(sigout),
-                                label='Signal output amplitude',
-                                set_cmd=partial(self._sigout_setter,
-                                                sigout-1, 1, outputamps[sigout]),
-                                get_cmd=partial(self._sigout_getter,
-                                               sigout-1, 1, outputamps[sigout]),
-                                unit='V')
-
             self.add_parameter('signal_output{}_ampdef'.format(sigout),
                                 get_cmd=None, set_cmd=None,
                                 initial_value='Vpk',
@@ -970,16 +962,27 @@ class ZIUHFLI(Instrument):
                                 val_mapping={'ON': 1, 'OFF': 0},
                                 vals=vals.Enum('ON', 'OFF') )
 
-            self.add_parameter('signal_output{}_enable'.format(sigout),
-                                label="Enable signal output's amplitude.",
-                                set_cmd=partial(self._sigout_setter,
-                                                sigout-1, 0,
-                                                outputampenable[sigout]),
-                                get_cmd=partial(self._sigout_getter,
-                                                sigout-1, 0,
-                                                outputampenable[sigout]),
-                                val_mapping={'ON': 1, 'OFF': 0},
-                                vals=vals.Enum('ON', 'OFF') )
+            for sigout in range(1, 9):
+                self.add_parameter('signal_output{}_enable'.format(sigout),
+                                   label="Enable signal output's amplitude.",
+                                   set_cmd=partial(self._sigout_setter,
+                                                   sigout - 1, 0,
+                                                   outputampenable[sigout]),
+                                   get_cmd=partial(self._sigout_getter,
+                                                   sigout - 1, 0,
+                                                   outputampenable[sigout]),
+                                   val_mapping={'ON': 1, 'OFF': 0},
+                                   vals=vals.Enum('ON', 'OFF'))
+
+                self.add_parameter('signal_output{}_amplitude'.format(sigout),
+                                   label='Signal output amplitude',
+                                   set_cmd=partial(self._sigout_setter,
+                                                   sigout - 1, 1,
+                                                   outputamps[sigout]),
+                                   get_cmd=partial(self._sigout_getter,
+                                                   sigout - 1, 1,
+                                                   outputamps[sigout]),
+                                   unit='V')
 
         auxoutputchannels = ChannelList(self, "AUXOutputChannels", AUXOutputChannel,
                                snapshotable=False)
--- a/qcodes/instrument_drivers/ZI/ZIUHFLI.py
+++ b/qcodes/instrument_drivers/ZI/ZIUHFLI.py
@@ -5,7 +5,7 @@ import numpy as np
 from functools import partial
 from math import sqrt
 
-from typing import Callable, List, Union, cast
+from typing import Callable, List, Union, cast, Optional
 
 from qcodes.utils.helpers import create_on_off_val_mapping
 
@@ -931,13 +931,6 @@ class ZIUHFLI(Instrument):
                                 val_mapping={'ON': 1, 'OFF': 0},
                                 vals=vals.Enum('ON', 'OFF') )
 
-            self.add_parameter('signal_output{}_ampdef'.format(sigout),
-                                get_cmd=None, set_cmd=None,
-                                initial_value='Vpk',
-                                label="Signal output amplitude's definition",
-                                unit='V',
-                                vals=vals.Enum('Vpk','Vrms', 'dBm'))
-
             self.add_parameter('signal_output{}_range'.format(sigout),
                                 label='Signal output range',
                                 set_cmd=partial(self._sigout_setter,
@@ -945,6 +938,13 @@ class ZIUHFLI(Instrument):
                                 get_cmd=partial(self._sigout_getter,
                                                 sigout-1, 1, 'range'),
                                 vals=vals.Enum(0.075, 0.15, 0.75, 1.5))
+
+            self.add_parameter('signal_output{}_ampdef'.format(sigout),
+                               get_cmd=None, set_cmd=None,
+                               initial_value='Vpk',
+                               label="Signal output amplitude's definition",
+                               unit='V',
+                               vals=vals.Enum('Vpk', 'Vrms', 'dBm'))
 
             self.add_parameter('signal_output{}_offset'.format(sigout),
                                 label='Signal output offset',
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -969,16 +961,29 @@ def __init__(self, name: str, device_ID: str, **kwargs) -> None:
...
+                self.add_parameter(
+                    'signal{}_output{}_enable'.format(sigout, output),
+                    label="Enable signal output's amplitude.",
+                    set_cmd=partial(self._sigout_setter,
+                                    sigout - 1, 0,
+                                    'enables/{}'.format(output - 1)),
+                    get_cmd=partial(self._sigout_getter,
+                                    sigout - 1, 0,
+                                    'enables/{}'.format(output - 1)),
+                    val_mapping={'ON': 1, 'OFF': 0},
```
astafan8: for all the on/off parameters, you might want to use [`create_on_off_val_mapping` function from](https://github.com/QCoDeS/Qcodes/blob/master/qcodes/utils/helpers.py#L670), [read more about it here](http://qcodes.github.io/Qcodes/examples/Creating%20Instrument%20Drivers.html?highlight=drivers#On/Off-parameters).

jenshnielsen:
Will this not only work if you have the Multi-frequency upgrade? See #978 where i started implementing this too

qcodes/instrument_drivers/ZI/ZIUHFLI.py
qSaevar: I will have a look.

qSaevar:
@jenshnielsen that could be. I'll see if we have a device without the upgrade for further testing/developing.

jenshnielsen:
@qSaevar  We have one without the upgrade so I can test that side but I do not have one with the upgrade. I never finished the other pr because of I couldn't test it probably. 
But if we can split the testing between us that would be great. Feel free to steal the code from that pr to detect if the feature is enabled

qcodes/instrument_drivers/ZI/ZIUHFLI.py
qSaevar: ✔️

qSaevar:
@jenshnielsen I added a check, as in your PR, to see if MF is enabled. I can test this on a device later today.

qSaevar:
@jenshnielsen I merged the code from PR 978 into this one and tested it on a device with MF enabled and it works. I do not have a device with out MF so I can not test that,

jenshnielsen:
@qSaevar Thanks, I will test it on our device without mf. Hopefully tomorrow.

qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -1758,41 +1797,51 @@ def update_range():
...
+                         demodulators this acts on
         """
 
         querystr = '/{}/sigouts/{}/{}'.format(self.device, number, setting)
+        if output_mode is not None:
+            querystr += '/{}'.format(output_mode)
         if mode == 0:
             value = self.daq.getInt(querystr)
-        if mode == 1:
+        elif mode == 1:
```
astafan8: so the mode, is it of type "int" or "bool"? and i guess the docstring has to contain the info that mode 0 is int and mode 1 is double.

qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -929,21 +931,6 @@ def __init__(self, name: str, device_ID: str, **kwargs) -> None:
                                 val_mapping={'ON': 1, 'OFF': 0},
```
astafan8: since `create_on_off_val_mapping` is introduced to this driver, shall it be applied to all the other parameters? :)

qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -952,6 +939,13 @@ def __init__(self, name: str, device_ID: str, **kwargs) -> None:
                                                 sigout-1, 1, 'range'),
                                 vals=vals.Enum(0.075, 0.15, 0.75, 1.5))
 
+            self.add_parameter('signal_output{}_ampdef'.format(sigout),
+                               get_cmd=None, set_cmd=None,
+                               initial_value='Vpk',
+                               label="Signal output amplitude's definition",
+                               unit='V',
```
astafan8: this parameter does not have a reasonable unit.
```suggestion
                               unit='',
```

qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -970,16 +964,45 @@ def __init__(self, name: str, device_ID: str, **kwargs) -> None:
...
-                                vals=vals.Enum('ON', 'OFF') )
+            if 'MF' in self.props['options']:
+                for modeout in range(1, 9):
+                    self.add_parameter('signal_output{}_amplitude{}'.format(sigout, modeout),
+                                       label='Signal output amplitude',
+                                       set_cmd=partial(self._sigout_setter,
+                                                       sigout - 1, 1, 'amplitudes', output_mode=modeout - 1),
+                                       get_cmd=partial(self._sigout_getter,
+                                                       sigout - 1, 1, 'amplitudes', output_mode=modeout - 1),
+                                       unit='V')
```
astafan8: I'd add a `docstring=...` to this parameter explaining that "the actual unit and representation is defined by `signal_output{}_ampdef` parameter". (same for the non_MF parameters)

qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -970,16 +964,45 @@ def __init__(self, name: str, device_ID: str, **kwargs) -> None:
...
+                                       set_cmd=partial(self._sigout_setter,
+                                                       sigout - 1, 0,
+                                                       'enables', output_mode=modeout - 1),
+                                       get_cmd=partial(self._sigout_getter,
+                                                       sigout - 1, 0,
+                                                       'enables', output_mode=modeout - 1),
+                                       val_mapping=create_on_off_val_mapping())
+            else:
+                self.add_parameter('signal_output{}_enable'.format(sigout),
+                                   label="Enable signal output's amplitude.",
```
astafan8: what is meant with "enabling an amplitude"? is it just enabling that output so that the signal goes out of the device, or is it something else?

qcodes/instrument_drivers/ZI/ZIUHFLI.py
astafan8: probably not in this PR, this requires changes in the logic because the logic is wired to 'ON'/'OFF' values and should be rewired to use True/False.... next time)

qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -952,6 +939,13 @@ def __init__(self, name: str, device_ID: str, **kwargs) -> None:
                                                 sigout-1, 1, 'range'),
                                 vals=vals.Enum(0.075, 0.15, 0.75, 1.5))
 
+            self.add_parameter('signal_output{}_ampdef'.format(sigout),
+                               get_cmd=None, set_cmd=None,
```
astafan8: Perhaps not very much related to this PR, but still. i'm not sure i understand how this parameter works. I see that it's used only in the private "_setter" function, but not used in the private "_getter" function. So, my questions is how does the instrument know that it has to accept or return values in Vpk/Vrms/dBm? Have i missed something? Or is this parameter just pure convenience for the users of the driver?

qcodes/instrument_drivers/ZI/ZIUHFLI.py
qSaevar: I changed the typehint to int

qcodes/instrument_drivers/ZI/ZIUHFLI.py
qSaevar: ✔️

qcodes/instrument_drivers/ZI/ZIUHFLI.py
qSaevar: Yes, it´s just enable signal output, the actual output. I fixed the label and added docstring

qcodes/instrument_drivers/ZI/ZIUHFLI.py
qSaevar: I moved the parameter around and got the blame. I did not add it or change it. But that said, I don't really understand the value of this parameter. The device has its amplitude in V, i.e it sets the peak amplitude of the signal output and scales the signal accordingly. I only see it in the setter, so if you set it to dBm, and then set the amplitude with `signal_output{}_amplitude{}` then I guess you will not get the same value back when retrieving the amplitude. But maybe someone else has a better understanding of this parameter.

astafan8: >  and got the blame

no blame :) I was seeking clarification for the sake of better understanding the driver and the change in this PR, nothing more)

Alright, then let's keep everything as is. I guess the existence of this parameter does not affect, say, your work, right?

qSaevar: No we do not use the parameter. When you view code on github, there is a blame button that you can click to see who did what. That´s what I ment with getting the blame. But I moved that parameter back, and some of the blame with it :)

qSaevar:
@astafan8 I addressed the comments you made.

qcodes/instrument_drivers/ZI/ZIUHFLI.py
astafan8: ah, ok, i see))

qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -1648,65 +1689,72 @@ def _get_demod_sample(self, number: int, demod_param: str) -> float:
         datadict['phi'] = np.angle(datadict['x'] + 1j * datadict['y'], deg=True)
         return datadict[demod_param]
 
-    def _sigout_setter(self, number, mode, setting, value):
+    def _sigout_setter(self, number: int,
+                       mode: bool,
```
astafan8: sorry for being so picky, but this confusion with `mode` argument is not only in `_sigout_getter` but also here, and i guess somehwere else in the driver. would you mind fixing that as the last thing for this PR?

qSaevar: Sure, this I'll hunt it down. This was a copy/paste from PR-978 and I guess I did not review much.

qSaevar: done,

PR #1501, opened 2019-03-04
Use numpy 1.16 by default

jenshnielsen:
Since its now available in both Conda and Pip form


jenshnielsen:
This has uncovered a corner case bug in numpy reported here https://github.com/numpy/numpy/issues/13089

WilliamHPNielsen:
I'm slightly confused... The CI passed using `numpy 16.2.1`, but it was our CI that uncovered the corner case bug? Or did you discover it locally?

Is it safe to merge, or should we wait for this https://github.com/numpy/numpy/pull/13092 to get in first?

jenshnielsen:
@WilliamHPNielsen It is a hypothesis test that generates the issue. So it is not deterministic. We may see random failures if we merge this but they are unlikely to happen in real life.

jenshnielsen:
We can also restrict the inputs to be more realistic

WilliamHPNielsen:
I see, thanks for explaining. I guess it's best to wait, then, depending a bit on how much work the restricting of inputs is. If it's a broad range of strategies in many tests (I guess we use linspace a lot), then it's not worth the bother, I think.

astafan8:
https://github.com/numpy/numpy/pull/13092 was merged. Let's see what CI says now. If there are no problems, shall we merge? @QCoDeS/core

jenshnielsen:
Yes, this was backported to numpy 1.16.3 which is released and now available in conda too

jenshnielsen:
@QCoDeS/core I think this can land now if someone approves it

PR #1503, opened 2019-03-04
Andor DU401 driver

SvenBo90:
feat: New driver

Driver for Andor DU401 and example notebook

no breaking changes


qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+    def get_camera_serial_number(self):
+        c_serial_number = ctypes.c_int()
+        code = self.dll.GetCameraSerialNumber(ctypes.byref(c_serial_number))
+        self.error_check(code, 'GetCameraSerialNumber')
+        return c_serial_number.value
+
+    def get_hardware_version(self):
+        c_pcb = ctypes.c_int()
+        c_decode = ctypes.c_int()
+        c_dummy1 = ctypes.c_int()
```
astafan8: the dummy arguments, do they not have any meaning? or smth like "reserved" by the API? Anyway, please add a comment for this in the code (or rename them to something more clear if possible).

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+    def start_acquisition(self):
+        code = self.dll.StartAcquisition()
+        self.error_check(code, 'StartAcquisition')
+
+    def wait_for_acquisition(self):
+        code = self.dll.WaitForAcquisition()
+        self.error_check(code, 'WaitForAcquisition')
+
+
+class Spectrum(Parameter):
```
astafan8: I personally do not think there is a need for making this parameter a separate class. The `get_raw` method can be implemented in the driver class.

Moreover, it seems that this parameter is a good candidate for `ParameterWithSetpoints` (with `Array` validator and callable `shape`), [it is mentioned here](http://qcodes.github.io/Qcodes/examples/Creating%20Instrument%20Drivers.html?highlight=parameterwithsetpoints#Custom-Parameter-classes). I encourage you to read more about it [here](http://qcodes.github.io/Qcodes/examples/writing_drivers/Simple-Example-of-ParameterWithSetpoints.html), [here](http://qcodes.github.io/Qcodes/examples/writing_drivers/A-ParameterWithSetpoints-Example-with-Dual-Setpoints.html), [here](http://qcodes.github.io/Qcodes/examples/driver_examples/Qcodes-example-with-Signal-Hound-USB-SA124B-ParameterWithSetpoints.html), and see if you like this approach. I think that it fits perfectly here.

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+
+class Andor_DU401(Instrument):
+    """
+    Instrument driver for the Andor DU401 BU2 CCD.
+
+    Args:
+        name (str): Instrument name.
+        dll_path (str): Path to the atmcd64.dll file. If not set, a default path is used.
+        camera_id (int): ID for the desired CCD.
+        setup (bool): Flag for the setup of the CCD. If true, some default settings will be sent to the CCD.
```
astafan8: how "default" are these settings? In principle, the driver should not impose anything that is specific to only a subset of measurements that is supports. If it does, then in becomes much more difficult to implement other measurements (or measurement modes) without breaking user's code (who started relying on some defaults).

I'd strongly recommend to remove this "setup" feature and the corresponding flag, and let the users decide what the defaults are.

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+
+        self.add_parameter('acquisition_mode',
+                           set_cmd=self.atmcd64d.set_acqiusition_mode,
+                           val_mapping={
+                               'single scan': 1,
+                               'accumulate': 2
+                           },
+                           label='acquisition mode')
+
+        self.add_parameter('cooler',
```
astafan8: what about `cooled_enabled`? :)

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+        def exposure_time_parser(ans):
+            return float(ans[0])
+        self.add_parameter('exposure_time',
+                           get_cmd=self.atmcd64d.get_acquisition_timings,
+                           set_cmd=self.atmcd64d.set_exposure_time,
+                           get_parser=exposure_time_parser,
+                           unit='s',
+                           label='exposure time')
+
+        self.add_parameter('filter_mode',
```
astafan8: if there are only two filter modes (on and off), then isn't `filter_enabled` a better name? :)

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+        x_pixels (int): Number of pixels on the x axis.
+        y_pixels (int): Number of pixels on the y axis.
+
+    """
+
+    # TODO (SvenBo90): implement further acquisition modes
+    # TODO (SvenBo90): implement further read modes
+    # TODO (SvenBo90): implement further trigger modes
+    # TODO (SvenBo90): add and delete parameters dynamically when switching acquisition mode or read mode
+    # TODO (SvenBo90): handle shutter closing and opening timings
```
astafan8: these are worth mentioning in the docstring of the driver a format similar to "Note that the driver does not support this and that". Perhaps, also in the example notebook.

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+
+        self.add_parameter('filter_mode',
+                           get_cmd=self.atmcd64d.get_filter_mode,
+                           set_cmd=self.atmcd64d.set_filter_mode,
+                           val_mapping=create_on_off_val_mapping(on_val=2, off_val=0),
+                           label='filter mode')
+
+        self.add_parameter('number_accumulations',
+                           set_cmd=self.atmcd64d.set_number_accumulations,
+                           label='number accumulations')
```
astafan8: Would this be more correct?
```suggestion
                           label='Number of accumulations')
```

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+
+        self.add_parameter('read_mode',
+                           set_cmd=self.atmcd64d.set_read_mode,
+                           val_mapping={'full vertical binning': 0})
+
+        min_temperature, max_temperature = self.atmcd64d.get_temperature_range()
+        self.add_parameter('set_temperature',
```
astafan8: why are `set_temperature` and `temperature` different parameters?

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+                           shape=(1, self.x_pixels),
+                           label='spectrum')
+
+        self.add_parameter('temperature',
+                           get_cmd=self.atmcd64d.get_temperature,
+                           unit=u"\u00b0"+'C',
+                           label='temperature')
+
+        self.add_parameter('trigger_mode',
+                           set_cmd=self.atmcd64d.set_trigger_mode,
```
astafan8: isn't it possible to read form the instrument the current trigger mode?

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+
+        # initialization
+        self.atmcd64d.initialize(' ')
+        self.atmcd64d.set_current_camera(self.atmcd64d.get_camera_handle(camera_id))
+
+        # get camera information
+        self.serial_number = self.atmcd64d.get_camera_serial_number()
+        self.head_model = self.atmcd64d.get_head_model()
+        self.firmware_version = self.atmcd64d.get_hardware_version()[4]
+        self.firmware_build = self.atmcd64d.get_hardware_version()[5]
```
astafan8: what is the reason for separating the firmware version from the firmware build? Ideally, they should be part of a single "firmware version" value ))

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+
+    # set methods
+    def set_cooler(self, cooler_on):
+        if cooler_on == 1:
+            self.atmcd64d.cooler_on()
+        elif cooler_on == 0:
+            self.atmcd64d.cooler_off()
+
+    def set_shutter_mode(self, shutter_mode):
+        self.atmcd64d.set_shutter(1, shutter_mode, 30, 30)
```
astafan8: what do these magic numbers mean in this method call? Shouldn't they become constants (or perhaps private class attributes) so that they at least can be changed if necessary?

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+        if setup:
+            self.cooler.set(True)
+            self.set_temperature.set(-60)
+            self.read_mode.set('full vertical binning')
+            self.acquisition_mode.set('single scan')
+            self.trigger_mode.set('internal')
+            self.shutter_mode.set('fully auto')
+
+        # print connect message
+        self.connect_message(idn_param='IDN')
```
astafan8: This should not be needed.
```suggestion
        self.connect_message()
```

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+                           set_cmd=self.atmcd64d.set_filter_mode,
+                           val_mapping=create_on_off_val_mapping(on_val=2, off_val=0),
+                           label='filter mode')
+
+        self.add_parameter('number_accumulations',
+                           set_cmd=self.atmcd64d.set_number_accumulations,
+                           label='number accumulations')
+
+        self.add_parameter('read_mode',
+                           set_cmd=self.atmcd64d.set_read_mode,
```
astafan8: isn't it possible to read form the instrument the current value of this parameter?

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+                           unit='s',
+                           label='exposure time')
+
+        self.add_parameter('filter_mode',
+                           get_cmd=self.atmcd64d.get_filter_mode,
+                           set_cmd=self.atmcd64d.set_filter_mode,
+                           val_mapping=create_on_off_val_mapping(on_val=2, off_val=0),
+                           label='filter mode')
+
+        self.add_parameter('number_accumulations',
```
astafan8: isn't it possible to read form the instrument the current value of this parameter?

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+        def accumulation_cycle_time_parser(ans):
+            return float(ans[1])
+        self.add_parameter('accumulation_cycle_time',
+                           get_cmd=self.atmcd64d.get_acquisition_timings,
+                           set_cmd=self.atmcd64d.set_accumulation_cycle_time,
+                           get_parser=accumulation_cycle_time_parser,
+                           unit='s',
+                           label='accumulation cycle time')
+
+        self.add_parameter('acquisition_mode',
```
astafan8: isn't it possible to read form the instrument the current value of this parameter?

qcodes/instrument_drivers/Andor/DU401.py
```diff
@@ -0,0 +1,424 @@
...
+
+        min_temperature, max_temperature = self.atmcd64d.get_temperature_range()
+        self.add_parameter('set_temperature',
+                           set_cmd=self.atmcd64d.set_temperature,
+                           vals=Ints(min_value=min_temperature,
+                                     max_value=max_temperature),
+                           unit=u"\u00b0"+'C',
+                           label='set temperature')
+
+        self.add_parameter('shutter_mode',
```
astafan8: isn't it possible to read form the instrument the current value of this parameter?

astafan8:
@SvenBo90 @lankes-fzj 

CLosing this in favor of https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/32

PR #1504, opened 2019-03-04
Shamrock SR750 driver

SvenBo90:
feat: New driver

Driver for Shamrock SR750 and example notebook

no breaking changes


qcodes/instrument_drivers/Shamrock/SR750.py
```diff
@@ -0,0 +1,305 @@
...
+        code = self.dll.ShamrockClose()
+        self.error_check(code, 'ShamrockClose')
+
+    def get_calibration(self, device, number_pixels):
+        c_device = ctypes.c_int(device)
+        c_calibration = (ctypes.c_float * number_pixels)()
+        c_number_pixels = ctypes.c_int(number_pixels)
+        code = self.dll.ShamrockGetCalibration(c_device, c_calibration, c_number_pixels)
+        self.error_check(code, 'ShamrockGetCalibration')
+        return [max(0, v) for v in c_calibration]
```
astafan8:
```suggestion
        return [max(0.0, v) for v in c_calibration]
```
This alteration will make sure that the returned list contains only 'float's, and no 'int's. In python, `max(0, -2.5)` returns `0` of type `int`. `0.0` is of type `float`.

qcodes/instrument_drivers/Shamrock/SR750.py
```diff
@@ -0,0 +1,305 @@
...
+                           set_cmd=self.set_grating,
+                           get_parser=int,
+                           vals=Ints(min_value=1,
+                                     max_value=self.number_gratings),
+                           label='grating')
+
+        self.add_parameter('groove_density',
+                           get_cmd=self.get_groove_density,
+                           get_parser=int,
+                           unit='l/mm',
```
astafan8: what is `l` here? :) shouldn't it be `g` for `groove`?

qcodes/instrument_drivers/Shamrock/SR750.py
```diff
@@ -0,0 +1,305 @@
...
+                           get_cmd=self.get_wavelength,
+                           set_cmd=self.set_wavelength,
+                           get_parser=float,
+                           vals=Numbers(min_value=min_wavelength,
+                                        max_value=max_wavelength),
+                           unit='nm',
+                           label='wavelength')
+
+        # print connect message
+        self.connect_message(idn_param='IDN')
```
astafan8: This should be enough:
```suggestion
        self.connect_message()
```

qcodes/instrument_drivers/Shamrock/SR750.py
```diff
@@ -0,0 +1,305 @@
...
+
+        # send CCD info to Shamrock
+        self.ShamrockCIF.set_number_pixels(self.device_id, ccd_number_pixels)
+        self.ShamrockCIF.set_pixel_width(self.device_id, ccd_pixel_width)
+
+        # add the instrument parameters
+        self.add_parameter('blaze',
+                           get_cmd=self.get_blaze,
+                           get_parser=int,
+                           label='blaze')
```
astafan8: Would you mind capitalizing the labels?

qcodes/instrument_drivers/Shamrock/SR750.py
```diff
@@ -0,0 +1,305 @@
...
+                           vals=Numbers(min_value=min_wavelength,
+                                        max_value=max_wavelength),
+                           unit='nm',
+                           label='wavelength')
+
+        # print connect message
+        self.connect_message(idn_param='IDN')
+
+    # get methods
+    def get_blaze(self):
```
astafan8: since the users are expected to interact with parameters, and not with such methods directly, it makes sense to make get/set methods that are used in parameters, private, but prefixing their names with `_`. For example, `get_blaze` -> `_get_blaze`. This way, when users will be interacting with the driver and rely on autocomplete for available parameters, they won't see `get_blaze` next to the `blaze` parameter.

qcodes/instrument_drivers/Shamrock/SR750.py
astafan8: i noticed that in the notebook the connecting took ~70s (`Connected to: Shamrock None (serial:SR-2975, firmware:None) in 71.63s`) - that's quite long. Is it ok for this instrument?

qcodes/instrument_drivers/Shamrock/SR750.py
```diff
@@ -0,0 +1,305 @@
...
+        self.ShamrockCIF.set_number_pixels(self.device_id, ccd_number_pixels)
+        self.ShamrockCIF.set_pixel_width(self.device_id, ccd_pixel_width)
+
+        # add the instrument parameters
+        self.add_parameter('blaze',
+                           get_cmd=self.get_blaze,
+                           get_parser=int,
+                           label='blaze')
+
+        self.add_parameter('calibration',
```
astafan8: this parameter is not settable, also, when gotten, it returns an array. In this situation, the question is - does "it" have to be a parameter? Why not having it just in the form of `get_calibration` method? Also, a related question, does the calibration array have to be stored in a snapshot (it's quite long, but does it provide value when included in the snapshot)?

qcodes/instrument_drivers/Shamrock/SR750.py
```diff
@@ -0,0 +1,305 @@
...
+    def get_idn(self):
+        return {'vendor': 'Shamrock', 'serial': self.serial_number}
+
+    def get_slit(self):
+        return self.ShamrockCIF.get_slit(self.device_id)
+
+    def get_wavelength(self):
+        return self.ShamrockCIF.get_wavelength(self.device_id)
+
+    # set methods
```
astafan8: for these kinds of separators of code, it makes them more readable when they are surrounded by new lines, and not stick to the definition of the next method as it is in this case. So, i'd recommend to add a blank line right here for readablity (and for other "separators" like this).

qcodes/instrument_drivers/Shamrock/SR750.py
```diff
@@ -0,0 +1,320 @@
...
+        serial_number: Serial number of the spectrometer.
+        number_gratings: Number of gratings on the spectrometer.
+    """
+
+    def __init__(self, name: str, dll_path: Optional[str] = None, device_id: Optional[int] = 0, ccd_number_pixels: Optional[int] = 1024, ccd_pixel_width: Optional[int] = 26, **kwargs):
+
+        super().__init__(name, **kwargs)
+
+        # link to dll
+        self.ShamrockCIF: ShamrockCIF = ShamrockCIF(dll_path=dll_path)
```
astafan8: For the future, it is very beneficial for testing the driver without the need for the actual DLL file to be present, to have an Optional argument for the "dll-wrapper" instance, so that if nothing is passed "ShamrockCIF" gets created as here, but if smth is passed then that is used instead of creating "ShamrockCIF" instance.

qcodes/instrument_drivers/Shamrock/SR750.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+class ShamrockCIF:
+    """
+    Wrapper class for the ShamrockCIF.dll library.
+
+    The class has been tested for a Shamrock SR750.
+
+    Args:
+        dll_path (str): Path to the ShamrockCIF.dll file. If not set, a default path is used.
```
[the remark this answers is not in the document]
astafan8: thank you for using the type annotations!

qcodes/instrument_drivers/Shamrock/SR750.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def error_check(self, code, function_name=''):
+        if code in self._success_codes.keys():
+            if self.verbose:
+                print("ShamrockCIF: [%s]: %s" % (function_name, self._success_codes[code]))
+        elif code in self._error_codes.keys():
+            print("ShamrockCIF: [%s]: %s" % (function_name, self._error_codes[code]))
+            raise Exception(self._error_codes[code])
+        else:
+            print("ShamrockCIF: [%s]: Unknown code: %s" % (function_name, code))
```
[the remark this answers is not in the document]
astafan8: thanks for using the logger!

qcodes/instrument_drivers/Shamrock/SR750.py
```diff
@@ -0,0 +1,305 @@
...
+        self.dll = ctypes.windll.LoadLibrary(dll_path or self._dll_path)
+        os.chdir(current_path)
+
+    def error_check(self, code, function_name=''):
+        if code in self._success_codes.keys():
+            if self.verbose:
+                print("ShamrockCIF: [%s]: %s" % (function_name, self._success_codes[code]))
+        elif code in self._error_codes.keys():
+            print("ShamrockCIF: [%s]: %s" % (function_name, self._error_codes[code]))
+            raise Exception(self._error_codes[code])
```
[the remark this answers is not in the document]
astafan8: saw the change, great, thanks!

qcodes/instrument_drivers/Shamrock/SR750.py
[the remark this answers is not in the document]
astafan8: ok, i see, left out of scope.

qcodes/instrument_drivers/Shamrock/SR750.py
astafan8: great!

qcodes/instrument_drivers/Shamrock/SR750.py
astafan8: thanks for this change.

qcodes/instrument_drivers/Shamrock/SR750.py
astafan8: ok, now it's 8s, good :)

qcodes/instrument_drivers/Shamrock/SR750.py
astafan8: @SvenBo90 what is your take on this?

qcodes/instrument_drivers/Shamrock/SR750.py
```diff
@@ -0,0 +1,305 @@
...
+
+    """
+
+    # default dll path
+    _dll_path = 'C:\\Program Files\\Andor SDK\\Shamrock64\\ShamrockCIF.dll'
+
+    # success and error codes
+    _success_codes = {20202: 'SHAMROCK_SUCCESS'}
+    _error_codes = {
+        20201: 'SHAMROCK_COMMUNICATION_ERROR', 20266: 'SHAMROCK_P1INVALID', 20267: 'SHAMROCK_P2INVALID',
```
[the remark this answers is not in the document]
astafan8: great! it's much more readable now!

qcodes/instrument_drivers/Shamrock/SR750.py
astafan8: ok, i guess, "l" is what it should be.

lankes-fzj:
@astafan8, I took over the work on @SvenBo90 's drivers. For simplicity, this pull request will be continued in #1888.

jenshnielsen:
https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/28

PR #1506, opened 2019-03-04
Aviosys IP Power 9258S Driver

SvenBo90:
feat: New driver

Driver for Aviosys IP Power 9258S and example notebook

no breaking changes


qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
+from qcodes import Instrument, InstrumentChannel, ChannelList
+from qcodes.utils.validators import Enum
+from qcodes.utils.helpers import create_on_off_val_mapping
+import urllib
```
astafan8: @QCoDeS/core what do you think about the choice of `urllib` over [`requests`](http://docs.python-requests.org/en/master/)? Is it good to stick to the built-in in this case?

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
...
+        # add parameters
+        self.add_parameter('power',
+                           get_cmd=self.get_power,
+                           set_cmd=self.set_power,
+                           get_parser=int,
+                           val_mapping=create_on_off_val_mapping(on_val=1, off_val=0),
+                           label='power {}'.format(self.channel))
+
+    # get methods
+    def get_power(self):
```
astafan8: since the users are expected to interact with parameters, and not with such methods directly, it makes sense to make these get/set methods (that are used in parameters) private, by prefixing their names with `_`. Would you mind doing this for "the get/set methods that are used in parameters" in this PR?

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
...
+        request = urllib.request.Request(self.parent.address+'/set.cmd?cmd=getpower')
+        response = urllib.request.urlopen(request)
+        request_read = response.read()
+        return request_read.decode("utf-8")[4+int(self.channel_id)*6]
```
astafan8: the code of this method mixes two responsibilities - one for the url requests, the other one for extracting the response into the value of power. Could you make the distinction clear by, for example, splitting this method into two?

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # set methods
+    def set_power(self, power):
+        request = urllib.request.Request(self.parent.address+'/set.cmd?cmd=setpower+p6%s=%s' % (self.channel_id, power))
```
astafan8: there is already quite some code duplication regarding `urllib.request.Request(...` etc. This can be drastically improved by defining helper functions in the instrument class which just take the important part of the URL as an argument.

One direction that can be taken is to override `ask_raw` and `write_raw` methods of the instrument driver class, so that they take care of the `urllib` business, and prepending `self.parent.address` to the URL, and other common stuff. Once that is done, the code of `set_power` on this channel will look much simpler, smth along the following lines:
```python
def set_power(self, power):
    self.write(cmd='setpower', args=f'+p6{self.channel_id}={power}')
```
which is much sorter and more readable. The `urllib` business would be then located in one place, and `set_power` code is much simpler. Together, this reduces possibilities for typos and bugs.

Moreover, if later there will be need for other parameters (don't know if this instrument has more than just "power enabled" :) ), it will be very easy to do because the `urllib` business and even the format of the URL string is taken care of by the `ask_raw`/`write_raw` methods of the instrument driver.

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
...
+        self.channel = channel
+        self.channel_id = self._channel_ids[channel]
+
+        # add parameters
+        self.add_parameter('power',
+                           get_cmd=self.get_power,
+                           set_cmd=self.set_power,
+                           get_parser=int,
+                           val_mapping=create_on_off_val_mapping(on_val=1, off_val=0),
+                           label='power {}'.format(self.channel))
```
astafan8: would you mind capitalizing parameter labels?

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
...
+
+        super().__init__(parent, name)
+
+        # validate the channel id
+        self._channel_values.validate(channel)
+        self.channel = channel
+        self.channel_id = self._channel_ids[channel]
+
+        # add parameters
+        self.add_parameter('power',
```
astafan8: would you agree that `enabled` or `power_enabled` is a better name instead of `power` because it is more specific and aligns well with the boolean True/False values that the parameter has? Also, in the example notebook, you already invite the users to use True/False for setting `power` parameter as opposed to setting it to 'on' or 'off', which makes me think that `enabled` or `power_enabled` is really the name of the parameter you were looking for.

If so, could you rename it to `power_enabled`/`enabled`?

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # set methods
+    def set_power(self, power):
+        request = urllib.request.Request(self.parent.address+'/set.cmd?cmd=setpower+p6%s=%s' % (self.channel_id, power))
+        urllib.request.urlopen(request)
+
+
+class Aviosys_IP_Power_9258S(Instrument):
+    """
+    Instrument driver for the Aviosys IP Power 9258 remote socket controller.
```
astafan8:
```suggestion
    Instrument driver for the Aviosys IP Power 9258 remote socket controller.

```

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
...
+    Args:
+        name (str): Instrument name.
+        address (str): http address.
+        login_name (str): http login name.
+        login_password (str) http login password.
+    Attributes:
```
astafan8:
```suggestion

    Attributes:
```

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
astafan8: `int` in `int(self.channel_id)` is redundant because `self.channel_id` is already an integer according to `_channel_ids` dictionary, right?

astafan8: out of curiosity, what does a typical `request_read` in this case look like? what are the first 4 characters for and what are the 6 characters per channel for?

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
...
+    _channel_values = Enum('A', 'B', 'C', 'D')
+    _channel_ids = {'A': 1, 'B': 2, 'C': 3, 'D': 4}
+
+    def __init__(self, parent, name, channel):
+
+        super().__init__(parent, name)
+
+        # validate the channel id
+        self._channel_values.validate(channel)
+        self.channel = channel
```
astafan8: for improving the clarity, could you rename the `channel` attribute to `_id_name` and `channel_id` attribute to `_id_number` (also making them private)? because there are: channel objects, channel name of the device ('A'), channel name as a submodule of the driver class ('A'), name of the channel object ('ChanA' as the driver code below suggests), etc. It's important to not confuse those, and names is a good way of achieving that.

The confusion is already here within this PR. In the driver code below, the term `channel_id` was used locally to refer to `_id_name` while here, in the code of the channel, `channel_id` is actually `_id_number`. It's not a runtime problem, but it's not good for readability (hence debugging, extending, etc.).

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
...
+        # set up http connection
+        password_manager = urllib.request.HTTPPasswordMgrWithDefaultRealm()
+        password_manager.add_password(None, self.address, login_name, login_password)
+        handler = urllib.request.HTTPBasicAuthHandler(password_manager)
+        opener = urllib.request.build_opener(handler)
+        urllib.request.install_opener(opener)
+
+        # add channels
+        channels = ChannelList(self, "PowerChannels", PowerChannel, snapshotable=False)
+        for channel_id in ('A', 'B', 'C', 'D'):
```
astafan8: What about the following?
```suggestion
        for channel_id in PowerChannel._channel_ids.keys():
```

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # set up http connection
+        password_manager = urllib.request.HTTPPasswordMgrWithDefaultRealm()
+        password_manager.add_password(None, self.address, login_name, login_password)
+        handler = urllib.request.HTTPBasicAuthHandler(password_manager)
+        opener = urllib.request.build_opener(handler)
+        urllib.request.install_opener(opener)
+
+        # add channels
+        channels = ChannelList(self, "PowerChannels", PowerChannel, snapshotable=False)
```
astafan8: why is `snapshotable` `False`? Isn't it useful to know what power channels were enabled/disabled for a particular measurement?

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class PowerChannel(InstrumentChannel):
+    """
+    Channel class for a socket on the Aviosys IP Power 9258S.
+    Args:
+        parent (Instrument): Parent instrument.
+        name (str): Channel name.
+        channel (str): Alphabetic channel id.
+    Attributes:
```
astafan8:
```suggestion

    Attributes:
```

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
+from qcodes import Instrument, InstrumentChannel, ChannelList
+from qcodes.utils.validators import Enum
+from qcodes.utils.helpers import create_on_off_val_mapping
+import urllib
+
+
+class PowerChannel(InstrumentChannel):
+    """
+    Channel class for a socket on the Aviosys IP Power 9258S.
+    Args:
```
astafan8:
```suggestion

    Args:
```

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
...
+    # set methods
+    def set_power(self, power):
+        request = urllib.request.Request(self.parent.address+'/set.cmd?cmd=setpower+p6%s=%s' % (self.channel_id, power))
+        urllib.request.urlopen(request)
+
+
+class Aviosys_IP_Power_9258S(Instrument):
+    """
+    Instrument driver for the Aviosys IP Power 9258 remote socket controller.
+    Args:
```
astafan8:
```suggestion

    Args:
```

qcodes/instrument_drivers/Aviosys/IP_Power_9258S.py
```diff
@@ -0,0 +1,92 @@
...
+        parent (Instrument): Parent instrument.
+        name (str): Channel name.
+        channel (str): Alphabetic channel id.
+    Attributes:
+        channel (str): Alphabetic channel id.
+        channel_id (int): Numeric channel id.
+    """
+
+    _channel_values = Enum('A', 'B', 'C', 'D')
```
astafan8: This removes code duplication and makes life easy in case there are more channels added later (the latter is not really applicable here, but still).
```suggestion
    _channel_values = Enum(*_channel_ids.keys())
```

jenshnielsen:
replaced by https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/30

PR #1509, opened 2019-03-05
Add boxcar average option to M4i driver

peendebak:

- Add option for boxcar averaging
- Update class docstring
- Cleanup code by using the `__transfer_buffer_numpy` method

@WilliamHPNielsen @astafan8


qcodes/instrument_drivers/Spectrum/M4i.py
```diff
@@ -957,22 +1012,7 @@ def blockavg_hardware_trigger_acquisition(self, mV_range, nr_averages=10,
...
-        # data acquisition
-        self._def_transfer64bit(
-            pyspcm.SPCM_BUF_DATA, pyspcm.SPCM_DIR_CARDTOPC, 0, data_pointer, 0, sizeof32bit * memsize * numch)
-        self.general_command(pyspcm.M2CMD_DATA_STARTDMA |
-                             pyspcm.M2CMD_DATA_WAITDMA)
-
-        # convert buffer to numpy array
-        data = ct.cast(data_pointer, ct.POINTER(buffer_size))
-        output = np.frombuffer(data.contents, dtype=ct.c_int32) / nr_averages
-        self._debug = output
```
astafan8: the same code should then also be removed from (line 866](https://github.com/QCoDeS/Qcodes/pull/1509/files#diff-28e84092864b9f07030db603c5e1f52eR844)

qcodes/instrument_drivers/Spectrum/M4i.py
```diff
@@ -777,22 +786,41 @@ def start_acquisition(self, mV_range, memsize, posttrigger_size=None, verbose=0)
...
+        """
         # setup software buffer
-        buffer_size = ct.c_int16 * memsize * numch
+        if bytes_per_sample == 2:
+            buffer_size = ct.c_int16 * memsize * numch
+        elif bytes_per_sample == 4:
+            buffer_size = ct.c_int32 * memsize * numch
+        else:
+            raise ValueError('bytes_per_sample should be 2 or 4')
         data_buffer = (buffer_size)()
```
astafan8: `buffer_size` is very bad name for the variable. It's not the size, it's a ctype object for a data_buffer of a correct size.

qcodes/instrument_drivers/Spectrum/M4i.py
```diff
@@ -777,22 +786,41 @@ def start_acquisition(self, mV_range, memsize, posttrigger_size=None, verbose=0)
 
         return {'memsize': memsize, 'numch': numch, 'mV_range': mV_range}
 
-    def _transfer_buffer_numpy(self, memsize, numch):
-        """ Transfer buffer to numpy array """
+    def _transfer_buffer_numpy(self, memsize, numch, bytes_per_sample=2):
```
astafan8: do i understand correctly that this method not only transfers the data from the buffer to the PC, but also triggers the acquisition of the data (because of the `self.general_command(...)` call)? If so, then the name of this method is confusing. I'd suggest to call it `_acquire_and_transfer` or similar. The fact that a numpy array is returned can be documented in the docstrings, and (even better) in the type annotation of the method.

qcodes/instrument_drivers/Spectrum/M4i.py
```diff
@@ -777,22 +786,41 @@ def start_acquisition(self, mV_range, memsize, posttrigger_size=None, verbose=0)
...
 
         # convert buffer to numpy array
         data = ct.cast(data_pointer, ct.POINTER(buffer_size))
-        output = np.frombuffer(data.contents, dtype=ct.c_int16)
+        if bytes_per_sample == 2:
+            output = np.frombuffer(data.contents, dtype=ct.c_int16)
+        elif bytes_per_sample == 4:
+            output = np.frombuffer(data.contents, dtype=ct.c_int32)
+        else:
+            raise ValueError('bytes_per_sample should be 2 or 4')
```
astafan8: what is this exception repeated for? exactly same exception is going to be raised at the very beginning of the method is bytes_per_sample is not 2 or 4. Basically, this else clause is unreachable.

qcodes/instrument_drivers/Spectrum/M4i.py
```diff
@@ -777,22 +786,41 @@ def start_acquisition(self, mV_range, memsize, posttrigger_size=None, verbose=0)
...
+            bytes_per_sample (int): specifies the datatype. 2 for int16, 4 for int32
+        Returns:
+            array: transfered data
+
+        """
         # setup software buffer
-        buffer_size = ct.c_int16 * memsize * numch
+        if bytes_per_sample == 2:
+            buffer_size = ct.c_int16 * memsize * numch
+        elif bytes_per_sample == 4:
```
astafan8: i'd highly recommend to have the if-clause for bytes_per_sample only once in the code. Basically, bytes_per_sample define what the ctype type is going to be used. Once the ctype is determined, it can be used for buffer_size and for dtype in np.from_buffer without the need for repeating the if-clause. This make the code shorter, more readable, and reduces the possibility for bugs and typos.

qcodes/instrument_drivers/Spectrum/M4i.py
```diff
@@ -729,20 +750,8 @@ def multiple_trigger_acquisition(self, mV_range, memsize, seg_size, posttrigger_
...
-        self._def_transfer64bit(
-            pyspcm.SPCM_BUF_DATA, pyspcm.SPCM_DIR_CARDTOPC, 0, data_pointer, 0, 2 * memsize * numch)
-        self.general_command(pyspcm.M2CMD_DATA_STARTDMA |
-                             pyspcm.M2CMD_DATA_WAITDMA)
-
-        # convert buffer to numpy array
-        data = ct.cast(data_pointer, ct.POINTER(buffer_size))
-        output = np.frombuffer(data.contents, dtype=ct.c_int16)
+        # convert transfer data to numpy array
+        output = self._transfer_buffer_numpy(memsize, numch, bytes_per_sample=2)
```
astafan8: should bytes_per_sample also be an argument of the multiple_trigger_acquisition method? because now there is no way to run multiple_trigger_acquisition for bytes_per_sample==4.

qcodes/instrument_drivers/Spectrum/M4i.py
```diff
@@ -864,6 +892,34 @@ def gated_trigger_acquisition(self, mV_range, memsize, pretrigger_size, posttrig
...
+        self.segment_size(memsize)
+        self.posttrigger_memory_size(posttrigger_size)
+        self.data_memory_size(memsize * self.box_averages())
+        numch = self._num_channels()
+
+        self.trigger_or_mask(pyspcm.SPC_TMASK_SOFTWARE)
+        self.general_command(pyspcm.M2CMD_CARD_START |
+                             pyspcm.M2CMD_CARD_ENABLETRIGGER | pyspcm.M2CMD_CARD_WAITREADY)
+
+        output = self._transfer_buffer_numpy(memsize, numch, bytes_per_sample=4)
```
astafan8: should bytes_per_sample also be an argument of the single_software_trigger_acquisition_boxcar method? because now there is no way to run single_software_trigger_acquisition_boxcar for bytes_per_sample==2.

qcodes/instrument_drivers/Spectrum/M4i.py
peendebak: No, the acquisition starts with other commands (e.g. M2CMD_CARD_START). It is true the methods blocks untill the acquisition is complete. I added a type annotation for this method (but will not annotate the entire driver)

qcodes/instrument_drivers/Spectrum/M4i.py
peendebak: The boxcar acquisition only uses the int32 datatype. (this is required since the boxcar method adds values, so the int16 would overflow)

peendebak:
@astafan8 Addressed the comments. I am not sure about my solution to fix a mypy error (line 804). Do you have a better alternative?

astafan8:
I'd suggest the following:

```python
sample_ctype: Type
if bytes_per_sample == 2:
    sample_ctype = ct.c_int16
elif bytes_per_sample == 4:
    sample_ctype = ct.c_int32
else:
```

or a stricter version

```python
sample_ctype: Union[Type[ct.c_int16], Type[ct.c_int32]]
```

Could you try them and see which one works?

qcodes/instrument_drivers/Spectrum/M4i.py
peendebak: The `multiple_trigger_acquisition` does not generated values with 4 bytes precision.

qcodes/instrument_drivers/Spectrum/M4i.py
```diff
@@ -777,22 +788,39 @@ def start_acquisition(self, mV_range, memsize, posttrigger_size=None, verbose=0)
...
-            pyspcm.SPCM_BUF_DATA, pyspcm.SPCM_DIR_CARDTOPC, 0, data_pointer, 0, 2 * memsize * numch)
+            pyspcm.SPCM_BUF_DATA, pyspcm.SPCM_DIR_CARDTOPC, 0, data_pointer, 0, bytes_per_sample * memsize * numch)
         self.general_command(pyspcm.M2CMD_DATA_STARTDMA |
                              pyspcm.M2CMD_DATA_WAITDMA)
 
         # convert buffer to numpy array
-        data = ct.cast(data_pointer, ct.POINTER(buffer_size))
-        output = np.frombuffer(data.contents, dtype=ct.c_int16)
+        data = ct.cast(data_pointer, ct.POINTER(ctype_buffer_type))
+        output = np.frombuffer(data.contents, dtype=sample_ctype)
```
astafan8: isn't it possible to use `data_buffer` here instead of the casting and `data.contents`?

peendebak: @astafan8 Thanks for the suggestion. I updated and tested the code.

PR #1510, opened 2019-03-06
Feature/dem 805/implement sampling rate float parameter

qSaevar:
- Set and Get the sampling rate of the ZIUHFLI using float. 

@WilliamHPNielsen @jenshnielsen @astafan8

### DIFF
--- a/qcodes/instrument_drivers/ZI/ZIUHFLI.py
+++ b/qcodes/instrument_drivers/ZI/ZIUHFLI.py
@@ -1335,8 +1335,9 @@ class ZIUHFLI(Instrument):
                            set_cmd=partial(self._set_samplingrate_as_float,
                                            self._samplingrate_codes),
                            unit='Hz',
-                           get_cmd=partial(self._get_samplingrate_as_float),
-                           docstring=""" Set the scope sampling rate with float.
+                           get_cmd=self._get_samplingrate_as_float,
+                           docstring=""" A numeric representation of the scope's
+                             samplingrate parameter.
                              If a sampling rate is set that does not match one 
                              of the allowed sampling rates, then round to the 
                              nearest allowed sampling rate. Returns the actual
--- a/qcodes/instrument_drivers/ZI/ZIUHFLI.py
+++ b/qcodes/instrument_drivers/ZI/ZIUHFLI.py
@@ -1,3 +1,4 @@
+import math
 import time
 import logging
 import numpy as np
@@ -1332,17 +1333,15 @@ class ZIUHFLI(Instrument):
 
         self.add_parameter('scope_samplingrate_float',
                            label="Scope's sampling rate as float",
-                           set_cmd=partial(self._set_samplingrate_as_float,
-                                           self._samplingrate_codes),
+                           set_cmd=self._set_samplingrate_as_float,
                            unit='Hz',
                            get_cmd=self._get_samplingrate_as_float,
+                           vals=vals.Enum(
+                               [ZIUHFLI._convert_to_float(freq) for freq in
+                                self._samplingrate_codes.keys()]),
                            docstring=""" A numeric representation of the scope's
-                             samplingrate parameter.
-                             If a sampling rate is set that does not match one 
-                             of the allowed sampling rates, then round to the 
-                             nearest allowed sampling rate. Returns the actual
-                             sampling rate as reported by the parameter 
-                             scope_samplingrate."""
+                             samplingrate parameter. Sets and gets the sampling 
+                             rate by using the scope_samplingrate parameter."""
                            )
 
         self.add_parameter('scope_length',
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -1330,6 +1330,20 @@ def __init__(self, name: str, device_ID: str, **kwargs) -> None:
                             vals=vals.Enum(*list(self._samplingrate_codes.keys()))
                             )
 
+        self.add_parameter('scope_samplingrate_float',
+                           label="Scope's sampling rate as float",
+                           set_cmd=partial(self._set_samplingrate_as_float,
+                                           self._samplingrate_codes),
+                           unit='Hz',
+                           get_cmd=partial(self._get_samplingrate_as_float),
```
astafan8: no need for partial here

qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -2168,6 +2182,23 @@ def getduration():
...
+        converter = {'hz': 'e0', 'khz': 'e3', 'mhz': 'e6', 'ghz': 'e9',
+                     'thz': 'e12'}
+        value, suffix = frequency.split(' ')
+        return float(''.join([value, converter[suffix.lower()]]))
+
+    def _set_samplingrate_as_float(self, samplingrate_codes, frequency):
+        nearest_frequency = min(samplingrate_codes.keys(), key=lambda k: abs(
+            frequency - ZIUHFLI._convert_to_float(k)))
+
+        self.scope_samplingrate(nearest_frequency)
```
astafan8: smth like `self.scope_samplingrate_float._save_val(ZIUHFLI._convert_to_float(nearest_frequency))` is missing here, right, @QCoDeS/core ?

qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -1330,6 +1330,20 @@ def __init__(self, name: str, device_ID: str, **kwargs) -> None:
                             vals=vals.Enum(*list(self._samplingrate_codes.keys()))
                             )
 
+        self.add_parameter('scope_samplingrate_float',
+                           label="Scope's sampling rate as float",
+                           set_cmd=partial(self._set_samplingrate_as_float,
+                                           self._samplingrate_codes),
+                           unit='Hz',
+                           get_cmd=partial(self._get_samplingrate_as_float),
+                           docstring=""" Set the scope sampling rate with float.
```
astafan8: i personally think the word "float" is abused in this PR. smth like "numeric" or "number" seems to be better. but i leave this to the author.

qcodes/instrument_drivers/ZI/ZIUHFLI.py
qSaevar: ✔️

qcodes/instrument_drivers/ZI/ZIUHFLI.py
qSaevar: right?

qcodes/instrument_drivers/ZI/ZIUHFLI.py
qSaevar: ✔️

qSaevar:
@astafan8 I addressed the your comments except the one with `self.scope_samplingrate_float._save_val(ZIUHFLI._convert_to_float(nearest_frequency))` as it was left with a question mark. Should I do it?

qcodes/instrument_drivers/ZI/ZIUHFLI.py
astafan8: @QCoDeS/core ? :)

WilliamHPNielsen: See review below.

qSaevar:
@WilliamHPNielsen I very well see your point and I'm not going to try and convince you. This behavior of setting the nearest point is how it works in the m4i driver, so there was a request for it in the uhfli driver as well.
But having the option of excepting and returning the sampling rate as float would still be very handy to have in many cases. So what if I make the "nearest" frequency a static method that finds the nearest allowed frequency, given some floating point number, that can in turn be used to set the sampling frequency? Something like:
```
sampling_rate = ZIUHFLI.round_to_nearest_sampling_frequency(desired_sampling_rate)
uhfli.scope_samplingrate_float(sampling_rate)
```
?

WilliamHPNielsen:
@qSaevar I think what you propose with
```python
sampling_rate = ZIUHFLI.round_to_nearest_sampling_frequency(desired_sampling_rate)
uhfli.scope_samplingrate_float(sampling_rate)
```
is a nice and clean solution. 

Then the `scope_samplingrate_float` parameter can have an `Enum` validator only accepting the correct values, and the `round_to_nearest_sampling_frequency` method should then take care not to introduce floating-point errors, but that should not be too hard to implement.

qSaevar:
@WilliamHPNielsen I updated the PR so now the `scope_samplingrate_float` parameter only accepts sampling rates that are allowed by the device. There is also a helping method that returns the closest sampling rate from some desired value and it is no longer linear as before.

WilliamHPNielsen:
Looks terrific!

PR #1512, opened 2019-03-07
Fix/do not retrive nulls

WilliamHPNielsen:
The `get_parameter_data` method of the DataSet currently retrieves everything in each result column of the SQLite result table, including `NULL`s. This PR suggests to change that behaviour to only fetch the non-NULL values. The filtering is conditioned on the top-level parameter's values.

- introduce a new sqlite function to retrieve N+1 columns of data, filtered conditioned on one of the columns being non-`NULL`
- make `get_parameter_data` use that
- add two tests for scenarios where `NULL`s are present in the result table

@QCoDeS/core 

EDIT: This is needed before I can proceed with #1477


qcodes/dataset/sqlite_base.py
```diff
@@ -1332,7 +1334,16 @@ def get_parameter_data(conn: ConnectionPlus,
...
-        res = get_data(conn, table_name, param_names, start=start, end=end)
+        res = get_parameter_tree_values(conn,
+                                        table_name,
+                                        output_param,
+                                        *param_names[1:])
+
+        start = start or 1
+        end = end or len(res)
+
+        res = res[start-1:end]
```
jenshnielsen: I am not too happy with this. It seems pointless to have these start and end arguments if they don't actually limit what you get from the database

jenshnielsen: Perhaps we should just deprecate start/end here

qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -1011,12 +1062,29 @@ def test_get_parameter_data_independent_parameters(standalone_parameters_dataset
...
+        toplevel_names: names of the toplevel parameters of the dataset
+        expected_names: names of the parameters expected to be loaded for a
+            given parameter as a sequence indexed by the parameter.
+        expected_shapes: expected shapes of the parameters loaded. The shapes
+            should be stored as a tuple per parameter in a sequence containing
+            all the loaded parameters for a given requested parameter.
+        expected_values: expected content of the data arrays stored in a
+            sequenceexpected_names:
+
+    """
```
jenshnielsen: +1 I should have written this earlier

qcodes/dataset/sqlite_base.py
astafan8: I agree.

since it's a relatively new method, we can just remove these arguments :)

alternatively, the start/end part of `get_data` can be factored out and re-used in the query in `get_parameter_tree_values`.

astafan8: The only thing, it is not trivial when to apply start/end - to the whole table or to the de-NULL-ed table?)

qcodes/dataset/sqlite_base.py
```diff
@@ -1332,7 +1334,16 @@ def get_parameter_data(conn: ConnectionPlus,
         param_names = [param.name for param in paramspecs]
         types = [param.type for param in paramspecs]
```
astafan8: it's not related to this PR, but [the behavior from this line](https://github.com/QCoDeS/Qcodes/pull/1512/files#diff-c7bb6e3db6832c32d8ae8596d8991d78R1328) does not seem to be documented in the docstring.

This is the missing part that i've found in the DataSet method's docstring:
```
If no paramerers are supplied the values will
be returned for all parameters that are not them self dependencies.
```

qcodes/dataset/sqlite_base.py
jenshnielsen: @astafan8 I think that's why. But as I see it the start and end only have value because they can limit how much data that you load from the db. That does not happen if you load everything and then limit

qcodes/dataset/sqlite_base.py
```diff
@@ -1363,6 +1374,7 @@ def get_parameter_data(conn: ConnectionPlus,
         output[output_param] = {name: np.array(column_data)
                          for name, column_data
                          in zip(param_names, res_t)}
```
astafan8: another comment unrelated to this PR: `res_t` can be left as a generator (without the outermost `list(...)`) since it's consumed by `zip`. probably, performance will be better.

qcodes/tests/dataset/dataset_fixtures.py
```diff
@@ -35,6 +35,27 @@ def scalar_dataset(dataset):
     yield dataset
 
 
+@pytest.fixture
+def scalar_dataset_with_nulls(dataset):
+    """
+    A very simple dataset. A scalar is varied, and two parameters are measured
```
astafan8: i'd argue that this is not "simple" :)

This is simple:
```
1 | 25 | 46
```

This is less simple:
```
1 | 25   | NULL
1 | NULL | 42
```

qcodes/tests/dataset/dataset_fixtures.py
```diff
@@ -35,6 +35,27 @@ def scalar_dataset(dataset):
...
+
+    for p in [sp, val1, val2]:
+        dataset.add_parameter(p)
+
+    dataset.mark_started()
+
+    dataset.add_results([{sp.name: 0, val1.name: 1},
+                         {sp.name: 0, val2.name: 2}])
+    dataset.mark_completed()
+    yield dataset
```
astafan8: i forgot, do we need the following dance here? :
```python
try:
    yield dataset
finally:
    dataset.conn.close()
```

qcodes/dataset/sqlite_base.py
astafan8: exactly, extend `_build_data_query` and re-use it - what do you think?

qcodes/dataset/sqlite_base.py
```diff
@@ -1332,7 +1334,16 @@ def get_parameter_data(conn: ConnectionPlus,
...
         types = [param.type for param in paramspecs]
 
-        res = get_data(conn, table_name, param_names, start=start, end=end)
+        res = get_parameter_tree_values(conn,
+                                        table_name,
+                                        output_param,
+                                        *param_names[1:])
+
+        start = start or 1
+        end = end or len(res)
```
astafan8: if i remember correctly, `0 or len(res)` returns `len(res)`, hence just simple `or` is not enough for `Optional[int]`. same for `start`.
```suggestion
        end = end if end is not None else len(res)
```

qcodes/tests/dataset/dataset_fixtures.py
WilliamHPNielsen: Yes, it's needed. While yielding, a test may fail and raise.

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: It's not the first time you point this out to me. Perhaps I'll eventually learn 🙂

qcodes/dataset/sqlite_base.py
WilliamHPNielsen: So I went and benchmarked this quickly, and the results are not conclusive. One method was not faster in all cases (somewhat surprisingly). But it *feels* faster to use a generator, so I'm gonna go with that. It was also fastest in most cases.

WilliamHPNielsen:
@QCoDeS/core I think the latest push addresses all concerns. Do you agree?

qcodes/dataset/sqlite_base.py
```diff
@@ -1390,6 +1401,70 @@ def get_values(conn: ConnectionPlus,
...
+            returned. None is equivalent to "all the rest". If start > end,
+            nothing is returned.
+
+    Returns:
+        A list of list. The outer list index is row number, the inner list
+        index is parameter value (first toplevel_param, then other_param_names)
+    """
+
+    offset = (start - 1) if start is not None else 0
+    limit = (end - start + 1) if end is not None else -1
```
jenshnielsen: I think mypy is complaining that this line does not check if start is None

WilliamHPNielsen: `mypy` has a point. Fixed this.

PR #1513, opened 2019-03-07
use pytest azure pipelines

jenshnielsen:
this replaces coverage and test upload


jenshnielsen:
I am not convinced this is an improvement

PR #1516, opened 2019-03-08
Update Offline Plotting Tutorial with kwargs examples

WilliamHPNielsen:
A user in Copenhagen asked me how to:

- Set the colormap for a single call to `plot_by_id`
- Set the coloraxis to a log scale in `plot_by_id`
- Set the range for the colorscale in `plot_by_id`

This is all easily done by passing the appropriate `kwargs` to `plot_by_id`.

- Document how to do the three things mentioned above.

@QCoDeS/core


jenshnielsen:
Looks good.

A few optional suggestions

* Should the logscale be done on data that is always > 0 
* Should @Dominik-Vogel s autoscaling be referenced in the vmin/vmax example

jenshnielsen:
@WilliamHPNielsen I think you might need to increase the timeout for that notebook https://nbsphinx.readthedocs.io/en/0.4.2/timeout.html

PR #1517, opened 2019-03-08
Update logger docs

jenshnielsen:
Start working on improving the api docs for the logging modue


jenshnielsen:
@Dominik-Vogel Could you have a look at this since you wrote most of the logging module

qcodes/logger/instrument_logger.py
```diff
@@ -1,3 +1,9 @@
+"""
+This module defines a :class:`logging.LoggerAdapter` and
+:class:`logging.Filter` used to enable capture of output from specific
```
Dominik-Vogel:
```suggestion
:class:`logging.Filter` used to enable the capturing of output from specific
```

PR #1518, opened 2019-03-11
Build docs with nitpick mode

jenshnielsen:
Nitpick mode will check all internal links. This required a lot of changes since all types should be linked

nitpick_ignore defines all the things that we currently ignore. 

Eventually we want to move the type args from the docstring into the signature using pep484 typing/mypy but I have not attempted to do that here. But I have tried to make the types more consistent by removing non existing types such as string (which should have been str) 

I have opted for making the type annotations in the docstrings closer to the mypy compatible types by allowing Optional etc in the docstrings. This requires adding these to ignore. Eventually once we convert everything to mypy style this can probably go. 

There are a few strange types (especially in the drivers) that do not resolve. For the moment I have just added workarounds for these to be ignored. Eventually we should fix this.

There are a lot of stupid workarounds for the IPython Magic that imports IPythons docs which are not compatible with the typing standards. We could also just exclude this from the docs?


azure-pipelines.yml
```diff
@@ -68,7 +68,7 @@ jobs:
           CALL conda activate qcodes
           cd docs
           REM Turn warnings into errors
-          set SPHINXOPTS=-W -v
```
jenshnielsen: remember to revert this

jenshnielsen: Done

qcodes/config/config.py
```diff
@@ -188,9 +188,9 @@ def validate(self, json_config=None, schema=None, extra_schema_path=None):
         validators are used.
 
         Args:
-            json_config (Optiona[string]) : json file to validate
-            schema (Optiona[dict]): schema dictionary
-            extra_schema_path (Optiona[string]): schema path that contains
+            json_config (Optional[str]) : json file to validate
+            schema (Optional[dict]): schema dictionary
+            extra_schema_path (Optional[str]): schema path that contains
                     extra validators to be added to schema dictionary
```
astafan8:
```suggestion
                extra validators to be added to schema dictionary
```

qcodes/instrument/base.py
```diff
@@ -108,8 +108,8 @@ def add_function(self, name: str, **kwargs) -> None:
 
         Args:
             name (str): how the Function will be stored within
-            ``instrument.Functions`` and also how you  address it using the
-            shortcut methods: ``instrument.call(func_name, *args)`` etc.
+                ``instrument.Functions`` and also how you  address it using the
+                shortcut methods: ``instrument.call(func_name, *args)`` etc.
```
astafan8:
```suggestion
```

qcodes/utils/plotting.py
```diff
@@ -186,7 +186,7 @@ def apply_auto_color_scale(colorbar: matplotlib.pyplot.colorbar,
     :meth:`apply_color_scale_limits`.
     If you want to adjust the color scale based on the configuration file
     `qcodesrc.json`, use :meth:`auto_color_scale_from_config`, which is used
-    In :meth:`plot_by_id`.
+    In :meth:`qcodes.dataset.plotting.plot_by_id`.
```
astafan8: shouldn't this be a "function" `:func`?

jenshnielsen: That seems to be true

PR #1521, opened 2019-03-13
increase deadline on this test as it was failing randomly

jenshnielsen:
Failed randomly in https://travis-ci.org/QCoDeS/Qcodes/jobs/505673065 as part of #1517


WilliamHPNielsen:
Yeah, that also happened in #1512

PR #1522, opened 2019-03-13
[DEM-831] Compiler warnings in ZIHDAWG8 should be logged or raised.

qSaevar:
- Compiler warnings in ZIHDAWG8 are caught and logged
- Compiler warnings can be elevated to errors

@astafan8

### DIFF
--- a/qcodes/instrument_drivers/ZI/ZIHDAWG8.py
+++ b/qcodes/instrument_drivers/ZI/ZIHDAWG8.py
@@ -22,8 +22,6 @@ WARNING_TRUNCATING = r"^Warning \(line: [0-9]+\): waveform \S+ cut " \
                      r" or 4 samples for dual channel waveforms\)$"
 WARNING_ANY = r"^Warning \(line: [0-9]+\):.*$"
 
-log = logging.getLogger(__name__)
-
 
 class CompilerError(Exception):
     """ Errors that occur during compilation of sequence programs."""
@@ -64,7 +62,7 @@ class ZIHDAWG8(Instrument):
 
         Args:
             *warnings: One or more regular expression that matches a compiler
-            warning.
+                warning.
 
         """
         for warning in warnings:
@@ -236,7 +234,7 @@ class ZIHDAWG8(Instrument):
             for warning_as_error in self._warnings_as_errors:
                 if re.search(warning_as_error, warning) is not None:
                     raise CompilerError('Warning treated as an error.', warning)
-            log.warning(warning)
+            self.log.warning(warning)
 
     def upload_waveform(self, awg_number: int, waveform: list,
                         index: int) -> None:
--- a/qcodes/instrument_drivers/ZI/ZIHDAWG8.py
+++ b/qcodes/instrument_drivers/ZI/ZIHDAWG8.py
@@ -1,6 +1,8 @@
 import csv
 import json
+import logging
 import os
+import re
 import textwrap
 import time
 from functools import partial
@@ -10,6 +12,21 @@ import zhinst.utils
 
 from qcodes import Instrument
 from qcodes.utils import validators as validators
+
+WARNING_CLIPPING = r"^Warning \(line: [0-9]+\): [a-zA-Z0-9_]+ has a higher " \
+                   r"amplitude than 1.0, waveform amplitude will be limited " \
+                   r"to 1.0$"
+WARNING_TRUNCATING = r"^Warning \(line: [0-9]+\): waveform \S+ cut " \
+                     r"down to playable length from [0-9]+ to [0-9]+ samples " \
+                     r"\(should be a multiple of 8 samples for single channel" \
+                     r" or 4 samples for dual channel waveforms\)$"
+WARNING_ANY = r"^Warning \(line: [0-9]+\):.*$"
+
+log = logging.getLogger(__name__)
+
+
+class CompilerError(Exception):
+    """ Errors that occur during compilation of sequence programs."""
 
 
 class ZIHDAWG8(Instrument):
@@ -39,6 +56,19 @@ class ZIHDAWG8(Instrument):
         self.awg_module.execute()
         node_tree = self.download_device_node_tree()
         self.create_parameters_from_node_tree(node_tree)
+        self._warnings_as_errors = []
+
+    def elevate_compiler_warning_to_error(self, *warnings: str) -> None:
+        """ Add a compiler warning that should be treated as an error during
+            sequence program compilation.
+
+        Args:
+            *warnings: One or more regular expression that matches a compiler
+            warning.
+
+        """
+        for warning in warnings:
+            self._warnings_as_errors.append(warning)
 
     def snapshot_base(self, update: bool = False,
                       params_to_skip_update: Sequence[str] = None) -> Dict:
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/ZI/ZIHDAWG8.py
```diff
@@ -176,21 +206,38 @@ def upload_sequence_program(self, awg_number: int,
...
         return self.awg_module.getInt('awgModule/compiler/status')
 
+    def _handle_compiler_warnings(self, status_string: str) -> None:
+        warnings = [warning for warning in status_string.split('\n') if
+                    re.search(WARNING_ANY, warning) is not None]
+        for warning in warnings:
+            for warning_as_error in self._warnings_as_errors:
+                if re.search(warning_as_error, warning) is not None:
+                    raise CompilerError('Warning treated as an error.', warning)
+            log.warning(warning)
```
jenshnielsen: I suggest using the log attached to the instrument see https://qcodes.github.io/Qcodes/examples/Creating%20Instrument%20Drivers.html#Logging for more info

qSaevar: ✔️

qSaevar:
I addressed the comment and tested the branch on hardware. Is there anything else I should do?

qcodes/instrument_drivers/ZI/ZIHDAWG8.py
```diff
@@ -39,6 +53,19 @@ def __init__(self, name: str, device_id: str, **kwargs) -> None:
         self.awg_module.execute()
         node_tree = self.download_device_node_tree()
         self.create_parameters_from_node_tree(node_tree)
+        self._warnings_as_errors: List[str] = []
+
+    def elevate_compiler_warning_to_error(self, *warnings: str) -> None:
```
astafan8: since there is this "add" functionality, i'd expect also a function for "removing" an already added warning so that it's treated again as a warning, right?)

to make it even easier, instead of creating the "add" and "remove" functions, i'd just make the `_warnings_as_errors` list "public" and write smth in the docs or the example notebook on how to use it.

qSaevar:
> filtering with regexp is absolutely not convenient in this case, and requires some effort to test/debug. i wish these warnings had "id" by which one could decide to throw an exception with it or not.

Filtering using regular expressions is needed as the status string has more info then just warnings. It is the output from the compiler and is only accessible as a single string.  What type of "id" are you thinking about? 

> if there are "errors", the only the first one is raised as an exception, and others are not included. I'd suggest that a more convenient way is to collect all "errors" and raise them as a single exception.

That I will do.

qcodes/instrument_drivers/ZI/ZIHDAWG8.py
```diff
@@ -177,21 +204,37 @@ def upload_sequence_program(self, awg_number: int,
...
+            CompilerError: If error occurs during compilation of the sequence
+            program, or if a warning is elevated to an error.
         """
         self.awg_module.set('awgModule/index', awg_number)
         self.awg_module.set('awgModule/compiler/sourcestring', sequence_program)
         while len(self.awg_module.get('awgModule/compiler/sourcestring')
-                      ['compiler']['sourcestring'][0]) > 0:
-            time.sleep(0.1)
+                  ['compiler']['sourcestring'][0]) > 0:
+            time.sleep(0.01)
```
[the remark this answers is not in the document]
qSaevar: ✔️

qcodes/instrument_drivers/ZI/ZIHDAWG8.py
qSaevar: ✔️

qSaevar:
@astafan8 I addressed the comments except the one with the error "ids".

qcodes/instrument_drivers/ZI/ZIHDAWG8.py
```diff
@@ -177,21 +199,38 @@ def upload_sequence_program(self, awg_number: int,
...
+            CompilerError: If error occurs during compilation of the sequence
+            program, or if a warning is elevated to an error.
         """
         self.awg_module.set('awgModule/index', awg_number)
         self.awg_module.set('awgModule/compiler/sourcestring', sequence_program)
         while len(self.awg_module.get('awgModule/compiler/sourcestring')
-                      ['compiler']['sourcestring'][0]) > 0:
-            time.sleep(0.1)
+                  ['compiler']['sourcestring'][0]) > 0:
+            pass
```
astafan8: isn't it more efficient (for some reason, i don't know) so still sleep here because the instrument is busy with compilation and if we are polling it constantly and very frequently, it slows down the compilation progress?

qcodes/instrument_drivers/ZI/ZIHDAWG8.py
```diff
@@ -177,21 +199,38 @@ def upload_sequence_program(self, awg_number: int,
...
         return self.awg_module.getInt('awgModule/compiler/status')
 
+    def _handle_compiler_warnings(self, status_string: str) -> None:
+        warnings = [warning for warning in status_string.split('\n') if
+                    re.search(WARNING_ANY, warning) is not None]
+        for warning in warnings:
+            for warning_as_error in self.warnings_as_errors:
+                if re.search(warning_as_error, warning) is not None:
+                    raise CompilerError('Warning treated as an error.',
+                                        *warnings)
```
astafan8: this still does not address my comment about "raising single exception for all the warnings that were treated as errors". This is because the exception is being raised from inside the for-loops. Could you make sure that all items from `warnings` that should be treated as errors (and only they) are passed to the exception?

qcodes/instrument_drivers/ZI/ZIHDAWG8.py
```diff
@@ -19,6 +33,13 @@ class ZIHDAWG8(Instrument):
     (tested using LabOne 18.05.54618 and firmware 53866).
     Furthermore, the Data Server and Web Server must be running and a connection
     between the two must be made.
+
+    Compiler warnings, when uploading and compiling a sequence program, can
+    be treated as errors. This is desirable if the user of the driver does
+    not want clipping or truncating of waveform to happen silently by the
+    compiler. Warnings are constants on the module level and can be added to the
+    drivers attribute warnings_as_errors. If warning are added, they will raise
```
astafan8: Perfect! This is just the syntax for sphinx - a popular tool for generating documentation. Also, this way 
warnings_as_errors is not glued into the other text and "stands out" :)
```suggestion
    drivers attribute :attr:`warnings_as_errors`. If warning are added, they will raise
```

qcodes/instrument_drivers/ZI/ZIHDAWG8.py
qSaevar: Yes, that could be. When I thought what is proper sleeping time, I could only come up with None. But I'll reintroduce it as a private property of, say, 10ms.

qcodes/instrument_drivers/ZI/ZIHDAWG8.py
qSaevar: sphinx is not happy about this:
`sphinx.errors.SphinxWarning: /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/qcodes/instrument_drivers/ZI/ZIHDAWG8.py:docstring of qcodes.instrument_drivers.ZI.ZIHDAWG8.ZIHDAWG8:7:py:attr reference target not found: warnings_as_error`

qcodes/instrument_drivers/ZI/ZIHDAWG8.py
```diff
@@ -177,21 +200,41 @@ def upload_sequence_program(self, awg_number: int,
...
+                    re.search(WARNING_ANY, warning) is not None]
+        errors = []
+        for warning in warnings:
+            for warning_as_error in self.warnings_as_errors:
+                if re.search(warning_as_error, warning) is not None:
+                    errors.append(warning)
+        if len(errors) > 0:
+            raise CompilerError('Warning treated as an error.', *errors)
+
+        [self.log.warning(warning) for warning in warnings]
```
astafan8: sorry, and now it would be great as a user to get the "warning" warnings before the "error" warnings so that you get the "full compiler response", not just the "error" part in case there are "error" warnings. Fair wish? :)

qcodes/instrument_drivers/ZI/ZIHDAWG8.py
astafan8: weird. i guess your work around is fine. (i'm still learning myself how sphinx works and how one should write correct references to attributes and such)

qcodes/instrument_drivers/ZI/ZIHDAWG8.py
astafan8: awesome :)

qcodes/instrument_drivers/ZI/ZIHDAWG8.py
astafan8: and just make this a nice for loop so that `mypy` does not complain :)

qSaevar: fair

qSaevar: it's gone

qcodes/instrument_drivers/ZI/ZIHDAWG8.py
qSaevar: It did not work

astafan8: sad.

then wrapping it at least into double quote thingies like below should be nice:
```suggestion
    drivers attribute ``warnings_as_errors``. If warning are added, they will raise
```

PR #1523, opened 2019-03-19
Alazar add get/set parameter

jenshnielsen:
This is useful as it among other things can be used to inspect the status of the buffers added to the card.

### DIFF
--- a/qcodes/instrument_drivers/AlazarTech/ats_api.py
+++ b/qcodes/instrument_drivers/AlazarTech/ats_api.py
@@ -21,11 +21,13 @@ from qcodes.instrument.parameter import _BaseParameter as Parameter
 U8 = ctypes.c_uint8
 U16 = ctypes.c_uint16
 U32 = ctypes.c_uint32
+C_LONG = ctypes.c_long
 HANDLE = ctypes.c_void_p
 
 POINTER_c_uint8 = Any
 POINTER_c_uint16 = Any
 POINTER_c_uint32 = Any
+POINTER_c_long = Any
 
 
 class AlazarATSAPI(WrappedDll):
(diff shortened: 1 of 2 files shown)


qcodes/instrument_drivers/AlazarTech/constants.py
```diff
@@ -519,3 +519,83 @@ class RecordAverageOption(IntEnum):
...
+    # The number of bits per sample
+    SETGET_ASYNC_BUFFSIZE_BYTES = 0x10000039
+    # The size of API-allocated DMA buffers in bytes
+    SETGET_ASYNC_BUFFCOUNT = 0x10000040
+    # The number of API-allocated DMA buffers
+    GET_ASYNC_BUFFERS_PENDING = 0x10000050
+    # DMA buffers currently posted to the board
+    GET_ASYNC_BUFFERS_PENDING_FULL = 0x10000051
+    # DMA buffers waiting to be processed by the application
+    GET_ASYNC_BUFFERS_PENDING_EMPTY = 0x10000052
```
astafan8: this is really cool. I wish we knew about it before)

qcodes/instrument_drivers/AlazarTech/constants.py
```diff
@@ -519,3 +519,83 @@ class RecordAverageOption(IntEnum):
 LED_OFF = 0
 LED_ON = 1
 
+
+class Parameters(IntEnum):
```
astafan8: with all the other enum classes defined before this PR, i was trying to follow the convention that the name of the class represents what one of the elements is, hence it's single (not plural). So, i'd suggest to rename this class to `Parameter` (not a good name in the qcodes realm but whatever), and the other new enums to "not plural".

qcodes/instrument_drivers/AlazarTech/constants.py
```diff
@@ -519,3 +519,83 @@ class RecordAverageOption(IntEnum):
...
+    PACK_MODE = 0x10000072
+    # Get/Set the pack mode as a member of PACK_MODES enum
+    SET_SINGLE_CHANNEL_MODE = 0x10000043
+    # Reserve all the on-board memory to the channel passed as
+    # argument. Single-port only.
+    API_FLAGS = 0x10000090
+    # State of the API logging as a member of
+    # API_TRACE_STATES enum
+
+class ECC_MODES(IntEnum):
```
astafan8: another convention that i tried to follow was to have pythonic names for the enum classes - `ECCMode` instead of `ECC_MODES`. Shall we continue with this convention?

qcodes/instrument_drivers/AlazarTech/ats_api.py
```diff
@@ -21,17 +21,19 @@
 U8 = ctypes.c_uint8
 U16 = ctypes.c_uint16
 U32 = ctypes.c_uint32
+C_LONG = ctypes.c_long
```
astafan8: nice! `AlazarBeforeAsyncRead` also uses c_long so shall it now use this alias as well?

jenshnielsen:
@astafan8 I think this addresses all your comments

jenshnielsen:
@astafan8 How you you feel about calling it `AlazarParameter` `Parameter` breaks the docs build. We could fix that but I think it's too generic anyway

astafan8:
@jenshnielsen I see, yes, `AlazarParameter` should be quite good actually.

PR #1524, opened 2019-03-20
Alazar Driver: Add api instance as initialization parameter

Dominik-Vogel:
With this small change I will be able to create a mock of the alazar driver that does not require an alazar card in order to instantiate and also potentially benchmark the alazar and alazar controllers, by simply providing an instance of  an api object that wraps all the dll calls without actually calling into the dll.


qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -1,940 +1,940 @@
...
+        return {
+            'system_id': system_id,
+            'board_id': board_id,
+            'board_kind': board_model,
+            'max_samples': max_s,
+            'bits_per_sample': bps
+        }
+
+    def __init__(self, name: str, system_id: int=1, board_id: int=1,
+                 dll_path: str=None, api: Optional[AlazarATSAPI] = None, **kwargs) -> None:
```
astafan8:
```suggestion
                 dll_path: Optional[str] = None, api: Optional[AlazarATSAPI] = None, **kwargs) -> None:
```

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -1,940 +1,940 @@
...
+
+        return {
+            'system_id': system_id,
+            'board_id': board_id,
+            'board_kind': board_model,
+            'max_samples': max_s,
+            'bits_per_sample': bps
+        }
+
+    def __init__(self, name: str, system_id: int=1, board_id: int=1,
```
astafan8:
```suggestion
    def __init__(self, name: str, system_id: int = 1, board_id: int = 1,
```

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -401,6 +403,7 @@ def acquire(self, mode=None, samples_per_record=None,
             enable_record_headers:
             alloc_buffers:
             fifo_only_streaming:
+
```
astafan8:
```suggestion
```

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -113,10 +113,12 @@ def get_board_info(cls, api: AlazarATSAPI, system_id: int,
             'bits_per_sample': bps
         }
 
-    def __init__(self, name: str, system_id: int=1, board_id: int=1,
-                 dll_path: str=None, **kwargs) -> None:
+    def __init__(
+            self, name: str, system_id: int = 1, board_id: int = 1,
+            dll_path: Optional[str] = None,
+            api: Optional[AlazarATSAPI] = None, **kwargs) -> None:
```
jenshnielsen: The api arg should be documented above.

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -39,6 +39,9 @@ class AlazarTech_ATS(Instrument):
         system_id: target system id for this board
         board_id: target board id within the system for this board
         dll_path: path to the ATS driver dll library file
+        api: AlazarATSAPI interface, defaults to the dll api. This argument
+            enables to provide another api, e.g. for a simulated driver for
```
jenshnielsen: I would write. `This argument makes it possible to provide another api` ... or `This argument enables the user to provide another api`

PR #1526, opened 2019-03-21
Lakeshore Model 336: make runnable, add minimal test, and fixes along the way

astafan8:
This PR is a result of a wish of using Lakeshore Model 336 driver in Delft, which turned into fixing some bugs in the drivers.

Most notable:
* Add test for Model 336 that is very similar to test for Model 372 (quick copy-paste mostly)
* Fix BaseOutput's input_channel parameter creation. It has to be a group parameter, but other kwargs like val_mapping and/or vals and others can be now specified in the subclasses to allow aproprivate variations between models. in Lakeshore Model 336, Model 372, and their base class.
* Map input_channel values and sensor channel names. Add map between BaseOutput's input_channel values and sensor channel object names. Model 336 uses letters for sensor channel names, hence the values of the BaseOutput's input_channel parameter naturally become strings. Model 372 uses numbers (it has 16 channels, while 366 has 4) for sensor channels, hence the values of the BaseOutput's input_channel parameter naturally become integers. To bridge this difference, and make wait_until_set_point_reached work for any Lakeshore model driver, a new mapping with a long ugly name input_channel_parameter_values_to_channel_name_on_instrument is introduced.
* Fix set_range_from_temperature bug for upper range - If the given temperature is from the highest range, the logic in the function would throw an exception because the index that bisect returns
is +1 relative to what it should be. Now it's fixed, and a test is added. For Lakeshore drivers, all models

@QCoDeS/core This PR "as is" does not act on the fact that the driver and test infrastructure for it is quite a bit convoluted and difficult to wrap ones head around. Hence, this PR only makes fixes, so that Model 336 driver works. Would you insist on reworking the structure of the driver and the test infrastructure in this PR or shall we leave it for "later"?

### DIFF
--- a/qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
+++ b/qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
@@ -268,9 +268,12 @@ class BaseOutput(InstrumentChannel):
                                    self.wait_equilibration_time.get_latest())
 
         active_channel_id = self.input_channel()
-        active_channel_number_in_list = active_channel_id - 1
-        active_channel = \
-            self.root_instrument.channels[active_channel_number_in_list]
+        active_channel = getattr(
+            self.root_instrument,
+            (self.root_instrument
+                .input_channel_parameter_values_to_channel_name_on_instrument[active_channel_id]
+            )
+        )
 
         if active_channel.units() != 'kelvin':
             raise ValueError(f"Waiting until the setpoint is reached requires "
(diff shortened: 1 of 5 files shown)


qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
```diff
@@ -266,15 +273,18 @@ def wait_until_set_point_reached(self,
                                    self.wait_equilibration_time.get_latest())
 
         active_channel_id = self.input_channel()
-        active_channel_number_in_list = active_channel_id - 1
-        active_channel = \
-            self.root_instrument.channels[active_channel_number_in_list]
+        active_channel = getattr(
+            self.root_instrument,
+            (self.root_instrument
```
jenshnielsen: For readability I would prefer making this its own variable outside the getattr. channel_name or some such thing

astafan8: Done.

Issue #1525, opened 2019-03-20
AMI430 driver


damazter:
I want to change the physical coils of the magnet, and I would like to update the safety limits accordingly.

I use the code:
```
ami_yellow.current_ramp_limit(0.1762) # A/s
ami_yellow.ramp_rate(0.1762 * 0.01490) # T/s <<<-- exception is raised here
ami_yellow.coil_constant(0.01490) # T/A
ami_yellow.current_limit(67.11) # A
```

(unfortunately traceback is lost, but the exception originated from `_set_ramp_rate` method, it's the ValueError that that method may raise (see line 369 of AMI430.py).)

It seems that the reason for this exception is that field_ramp_limit was not updated when I set current_ramp_limit above. Hence, the driver thinks that my ramp rate is out of bounds even though it is not (because i set the current_ramp_limit before).

Then, i tried the following:
```
ami_yellow.current_ramp_limit(0.1762) # A/s
ami_yellow.field_ramp_limit(0.1762 * 0.01490) # T/s <<<-- NOTE that I ALSO set field_ramp_limit parameter this time
ami_yellow.ramp_rate(0.1762*0.01490) # T/s
ami_yellow.coil_constant(0.01490) # T/A
ami_yellow.current_limit(67.11) # A
```

And this time there was no exception raised.

However, setting field_ramp_limit seems to overwrite my current_ramp_limit (because coil constant has not updated yet, i.e. it is not "0.01490" yet but some other previous value), so current_ramp_limit apparently gets set to a value which is out of my safety limits.

My proposed solution is a method of setting all safety limits and constants in a single function, and sending those to the instrument in a single call, preferebly in __init__. 
Moreover, I would even propose to make the coil_constant, current_ramp_limit (and thereby field_ramp_limit) and current_limit to unchangable after __init__ (so that when coils are changed, the user has to reinstantiate the magent driver).

Along the way found the following bug:
* scale of field_ramp_limit (which is dependent on the coil constant) is fixed in __init__, and this will bring problems if users change coil constanst.

PR #1527, opened 2019-03-21
AMI430: fix coil_constant-related bugs of current- and field- parameters

astafan8:
Fixes:
* `has_current_rating` argument/attribute should only matter for `field_rating` parameter, not for `field_ramp_limit` parameter.
* Update `scale` of `field_limit` parameter when a `coil_constant` is set
* Add tests for field* and current* parameters so that they are proportional via coil_constant
* Add tests reducing `ramp_rate` if new `current_ramp_limit` or `field_ramp_limit` is lower than `ramp_rate`
* Apply PEP8
* Fix AMI430 simulation yaml file

This PR might end up resolving https://github.com/QCoDeS/Qcodes/issues/1525.


astafan8:
@damazter Could you see if this PR fixes the coil-constant issue https://github.com/QCoDeS/Qcodes/issues/1525 ?

jenshnielsen:
Did we ever figure out why this difference exists. Afaik it had something todo with the driver/docs version on AMIs webpage being out of date and the new versions only being on bitbucket? It would be good to at least update the comments about this.

astafan8:
@jenshnielsen with the "difference" you mean the `has_current_rating` attribute? and with "update comment" you mean to update the following one?:
https://github.com/QCoDeS/Qcodes/blob/4c4f8f5edd9cd94fabf3ceb07cb4df556d347ece/qcodes/instrument_drivers/american_magnetics/AMI430.py#L177-180

WilliamHPNielsen:
That comment still reflects my knowledge, unfortunately. But I have not compared firmware versions of different models, so there is still the possibility that we'll be able to figure out roughly in what firmware version that feature was added/removed.

astafan8:
Thanks to @damazter @damazter2 rigorous testing of this change, we now confirm that the coil-constant related bugs are solved. By request of Damaz, two tests were added: one that runs the coil-constant/field/current-limit tests with arbitrary values and a number of times, and one new test for the fact that ramp_rate is reduced when ramp_limit is made lower than the current (now) ramp_rate.

After CI shows green, I am going to merge this. @QCoDeS/core any objections?

qcodes/tests/drivers/test_ami430.py
```diff
@@ -440,6 +555,230 @@ def test_blocking_ramp_parameter(current_driver, caplog):
...
+    Note that the 'factor's are randomized "manually" because of the
+    possibility to hit the limits of the parameters.
+    """
+    with warnings.catch_warnings():
+        # this is to avoid AMI430Warning about "maximum ramp rate", which
+        # may show up but is not relevant to this test
+        warnings.simplefilter('ignore', category=AMI430Warning)
+
+        test_current_and_field_params_interlink_at_init(ami430)
+
```
jenshnielsen: Would it make sense to make this a bit more DRY by looping over 3 tuples or pytest parametrize ?

astafan8: sure. i can probably hard-code some values and use permutations from itertools or smth.

but only if you insist :)

astafan8: @jenshnielsen ? :)

jenshnielsen: I would prefer but I don't insist. That's why its a question :)

astafan8: ok :)

PR #1529, opened 2019-04-01
build docs with sphinx 2.0

jenshnielsen:
Sphinx 2.0 is out. So lets see if our docs build with it.


jenshnielsen:
@QCoDeS/core Any objection to this. I think the changes needed are correct and harmless

PR #1530, opened 2019-04-02
Check line endings in CI

jenshnielsen:
When reviewing https://github.com/QCoDeS/Qcodes/pull/1524/ I realized that it will probably be useful make the CI enforce consistent line endings all over the project.


astafan8:
Hopefully a related question: why do we not use `.gitattributes` that checks out and checks in with `\n` on all systems?

jenshnielsen:
I don't have any experience with it but that seems like a reasonable thing to try out

astafan8:
I used it in the previous projects. It really removes this line-end hassle for the given repository, regardless of what the global git settings are.

Just to be clear, i think running a linter is also useful. The only thing i don't know is what the overlap with mypy and codacy there is.

jenshnielsen:
I don't think there is any overlap with mypy. Codacy uses pylint under the hood. But it does not expose this setting which is off by default

WilliamHPNielsen:
Seems like a good idea. Shouldn't we just get this in?

PR #1532, opened 2019-04-04
Make error_caused_by more robust

jenshnielsen:
* Handles cases where line information is missing (such as worked around in #1477 
* Only match against the root cause of the exception. This seems like what was always intended
* Add tests of the above
* Remove one incorrect use of error_caused_by

### DIFF
--- a/qcodes/tests/common.py
+++ b/qcodes/tests/common.py
@@ -105,7 +105,16 @@ def error_caused_by(excinfo: 'ExceptionInfo', cause: str) -> bool:
         cause: the error message or a substring of it
     """
     chain = excinfo.getrepr().chain
-    cause_found = False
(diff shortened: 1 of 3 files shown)


jenshnielsen:
We could make this look through the whole chain if we think that is better? 
I would argue that we should always omit the last element even in that case? 

@QCoDeS/core What do you think

jenshnielsen:
@astafan8 yes. In the last example it would IMHO be ok for it to match the `KeyError` and the `RuntimeError` but not the `ValueError`

astafan8:
@jenshnielsen thanks for explaining - i agree with you. `ValueError` is checked for at `pytest.raises(..)` stage, so no need match it again.

PR #1535, opened 2019-04-09
Limit maximum buffer size as documented for AlazarBeforeAsyncRead

jenshnielsen:
As documented here http://www.alazartech.com/Support/Download%20Files/ATS-SDK-Guide-7.2.3.pdf#section*.110 

We do not use the headers so that is not taken into account. My testing on the 9360 seems to indicate that it actually supports upto 96 MB but to stay on the safe side I have gone with the docs. We do not support and never will the old pci cards with lower limits so we do not need to take that into account


qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -445,6 +445,36 @@ def acquire(self, mode=None, samples_per_record=None,
...
+        bytes_per_buffer = (bytes_per_record *
+                            records_per_buffer * number_of_channels)
+
+        sample_type = ctypes.c_uint16 if bytes_per_sample > 1 else ctypes.c_uint8
+
+        max_buffer_size = 64*1024*1024
+        # 64 MB see docs of AlazarBeforeAsyncRead
+        # http://www.alazartech.com/Support/Download%20Files/ATS-SDK-Guide-7.2.3.pdf#section*.110
+
+        requested_buffer_size = (bits_per_sample * samples_per_record *
```
astafan8: why is `bits_per_sample` with `//8` used instead of `bytes_per_sample` from above?

astafan8: I find it hard to understand why `number_of_channels` of channels is not involved in this formula. The respective formula in SDK pdf also does not contain `number_of_channels`. Why is that? Are on-board-memory-buffers considered per channel, and host-memory-buffers not?

astafan8: and an extra question that is related to the above, what are the meanings of `requested_buffer_size` and `bytes_per_buffer` variables? (The names are not helping me (( ). My initial expectation is that `requested_buffer_size == bytes_per_buffer` and hence is not needed. What am i missing?

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -445,6 +445,36 @@ def acquire(self, mode=None, samples_per_record=None,
...
+        channels_binrep = self.channel_selection.raw_value
+        number_of_channels = self.get_num_channels(channels_binrep)
+
+        # bytes per buffer
+        bytes_per_buffer = (bytes_per_record *
+                            records_per_buffer * number_of_channels)
+
+        sample_type = ctypes.c_uint16 if bytes_per_sample > 1 else ctypes.c_uint8
+
+        max_buffer_size = 64*1024*1024
```
astafan8: since this is really a constant, shall it be located outside of this method?

qcodes/instrument_drivers/AlazarTech/ATS.py
astafan8: when i read a C++ example of DualPort NPT acquisition for ATS9360, i see the formula that `bytesPerBuffer = bytesPerRecord * recordsPerBuffer * channelCount` - channel number is involved :)

jenshnielsen: This buffer size is only a fraction of the total buffer size (1/64) in this case. I don't think that the limitation comes from filling the internal memory but from some other implementation detail that they don't reveal to use

jenshnielsen: Because there are 2 different considerations. The absolute size of the buffer (in this case there are 1.5 bytes per sample) 
and the padded to whole bytes per sample size. As far as I can see from the docs the internal buffer utilization needs to be calculated from the fractional value but we transfer whole bytes. I will clarify the variable names.

jenshnielsen: A variation of the above

qcodes/instrument_drivers/AlazarTech/ATS.py
jenshnielsen: Sure

qcodes/instrument_drivers/AlazarTech/ATS.py
jenshnielsen: The calculation with unites is 12 bits/sample / 8 bits/byte * x samples/record *y  records/ buffer which should give (12/8) * x * y bytes/buffer

astafan8: ok, thanks! So i was confusing on-board-memory buffers with host-memory buffers.

So the limit is for the on-board-memory buffer, right? and for some reason the limit does not involve number of channels, right?

And thanks for detailed comments around this formula in the newer commits!

astafan8: > This buffer size is only a fraction of the total buffer size (1/64) in this case.

Sorry, i'm getting more confused. What did you mean with "this buffer size" and "total buffer size" and where does "1/64" comes from? and how does this relate to taking `number_of_channels` into account or not?

> ... from some other implementation detail ...

i agree with you, i have the same feeling.

astafan8: still don't get why "internal buffer" (the on-board-memory buffer) size limit does not take number of channels into account :(

jenshnielsen: Sorry should have said total onboard memory. I mean given that we don't know the reason for this limitation (other than it's for sure not filling all the memory cf the card which would for sure depend on number of channels) all questions about this are speculations. This limitation might just be that the whatever fpga logic that captures data from a single channel only has a memory address space large enough to address 64 MB and there is a matching address space for the next channel. We really don't know at this stage.

astafan8: ok, thanks for explaining! I hope AlazarTech come back with useful answers :)

jenshnielsen:
@QCoDeS/core We don't really have a useful answer from AlazarTech. They confirm that this limit is here (at least on some cards) but also recommend to make the buffer smaller than this in general < 8 MB. 
I think we should put this in as is since it at least makes it much less confusing for users.

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -445,6 +445,40 @@ def acquire(self, mode=None, samples_per_record=None,
...
+            ctypes.c_uint16 if whole_bytes_per_sample > 1 else ctypes.c_uint8)
+
+        internal_buffer_size_requested = (bits_per_sample * samples_per_record *
+                                          records_per_buffer) // 8
+
+        if internal_buffer_size_requested > max_buffer_size:
+            raise RuntimeError(f"Requested a buffer of size: "
+                               f"{internal_buffer_size_requested / 1024 ** 2}"
+                               f" MB. The maximum supported size is "
+                               f"{max_buffer_size / 1024 ** 2} MB.")
```
astafan8: How about also adding the recommended here as well?
```suggestion
                               f"{max_buffer_size / 1024 ** 2} MB "
                               f"(recommended is <8MB).")
```

PR #1536, opened 2019-04-09
feat: Add driver for the Mightex SSE-1304-U CCD spectrometer

maxgoetsch:
feature: Driver for the Mightex SSE-1304-U CCD spectrometer.

- Add driver for the device.

@jenshnielsen


WilliamHPNielsen:
Hi @maxgoetsch, welcome to QCoDeS! Thanks for contributing to the driver base. The PR looks good, but before we can merge it, the CI should pass and the code should adhere to our house style. I'll drop you some comments on that below.

qcodes/instrument_drivers/mightex/SSE1304U.py
```diff
@@ -0,0 +1,175 @@
...
+import numpy as np
+import numpy.ctypeslib as npct
+
+import qcodes as qc
+import qcodes.utils.validators as vals
+
+from .dllwrapper import MightexDLLWrapper, FrameRecord  
+
+class WavelengthAxis(qc.Parameter):
+    ''' Holds the wavelengths corresponding to CCD pixels. '''
```
WilliamHPNielsen: Please use triple **double quotes** for the docstrings.

qcodes/instrument_drivers/mightex/dllwrapper.py
```diff
@@ -0,0 +1,54 @@
+import ctypes as ct
+import numpy.ctypeslib as npct
+
+''' This module communicates with the SSE SDK dll provided by Mightex. '''
```
WilliamHPNielsen: The module level docstring should be at the very top of the module. Thank you for including one.

WilliamHPNielsen:
There are some issues with `mypy`. It basically boils down to the implicit assumption of the `WavelengthAxis` and `IntensitySpectrum` that they be added to the correct type of instrument, which `mypy` cannot implicitly make. I would suggest making that assumption explicit, something along the lines of
```python
class WavelengthAxis(qc.Parameter):
    """ Holds the wavelengths corresponding to CCD pixels """

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        if not isinstance(self.root_instrument, MightexSSE_1304_U):
            Raise ValueError('WaveLengthAxis added to the wrong instrument '
                             'type. This parameter is only compatible with '
                             'MightexSSE_1304_U.')
        self.spectrometer: MightexSSE_1304_U = self.root_instrument

    def get_raw(self) -> np.ndarray:
        return self.spectrometer.calib(np.arange(0, 3648))
```

maxgoetsch:
Hi @WilliamHPNielsen, thanks for reviewing! Is there something wrong with the docstrings? It seems like the FrameRecord struct causes the errors, but I don't know why.

WilliamHPNielsen:
@maxgoetsch The docstrings should be okay now, but please have a look at trailing whitespace and line endings.

Apart from the formal requirement of CI passing, it'd be good if all parameters could be in SI units. I know it seems like an inconvenience (like, who in the universe ever recorded a wavelength in meters?), but we enforce it to ensure uniformity across the drivers.

Another thing to consider is whether the `__init__` method of a driver should have the side effect of changing the settings of an instrument. Usually we say that the `__init__` should only establish a connection and leave the instrument as it was. Is there a good reason to do that differently here?

qcodes/instrument_drivers/mightex/SSE1304U.py
```diff
@@ -0,0 +1,191 @@
...
+        """ Sets the number of frames that should be averaged
+        for each measurement. """
+        self._avg = avg if avg > 0 else 1
+
+    def rec_darkframe(self) -> np.ndarray:
+        self.drkdata = self.get_single_frame()
+        return self.drkdata
+
+    def set_darkframe(self, drk: np.ndarray):
+        self.drkdata = drk
```
WilliamHPNielsen: I guess a newline is needed at the end of the file.

qcodes/instrument_drivers/mightex/dllwrapper.py
```diff
@@ -0,0 +1,54 @@
...
+        self.set_exposuretime.argtypes = [ct.c_int, ct.c_int]
+
+        self.set_avg_framenum = self._dll.MTSSE_SetDeviceAverageFrameNum
+        self.set_avg_framenum.argtypes = [ct.c_int, ct.c_int]
+
+        self.set_activestatus = self._dll.MTSSE_SetDeviceActiveStatus
+        self.set_activestatus.argtypes = [ct.c_int, ct.c_int]
+
+        self.set_auto_dark = self._dll.MTSSE_SetDeviceSpectrometerAutoDarkStatus
+        self.set_auto_dark.argtypes = [ct.c_int, ct.c_int, ct.c_int]
```
WilliamHPNielsen: Also here, please add a newline.

WilliamHPNielsen:
@maxgoetsch I am still awaiting your response regarding the `__init__` method changing the settings on the instrument. What's your take on this?

maxgoetsch:
Hi @WilliamHPNielsen, I'm sorry to leave you waiting. You're right, it's not necessary to change settings in the unit, i will change that asap.

WilliamHPNielsen:
@maxgoetsch, no problem, I did not mean to make you hurry up, I just wanted to make sure that you were not waiting for *me* to make the next step. We don't want deadlocks in the review ;)

Looking forward to seeing the changes.

maxgoetsch:
@WilliamHPNielsen I just recognized that I included this change already in a previous commit. The only thing that is still changed in the __init__ is resetting the dark data since it is stored only software side in the class. Do you know a way to get around this?

WilliamHPNielsen:
@maxgoetsch, if a setting does not change anything on the instrument itself, I think it's fair enough to reset it every time you reconnect to the instrument. It is not an anti-pattern that a new driver instance has no knowledge of earlier instances; it is indeed expected. So that's fine. But if you find yourself in a situation where you constantly need to reconnect, e.g. because of crappy instrument firmware that randomly drops connections, we could think about persisting that setting to disk. But I'd say that's probably overkill.

One more thing to consider: since 3648 seems to be a magic number, it might make sense to refactor it into a class variable or instance attribute. And to put in a check in `set_darkframe` that the input has the right shape.

maxgoetsch:
@WilliamHPNielsen I added the CCD size as a class parameter. I think trying to reshape the array in `set_darkframe` should be sufficient since it raises an exception if the shape does not match.

qcodes/instrument_drivers/mightex/SSE1304U.py
```diff
@@ -0,0 +1,196 @@
...
+
+        return npct.as_array(buffer.contents.CCDData,
+                             shape=(MightexSSE_1304_U.PIXELS,)).copy()
+
+    @property
+    def calib(self):
+        """ Returns a function to calibrate pixels to wavelength. """
+        # These values are taken from the default calibration file
+        # shipped with the device.
+        return np.poly1d([
```
sohailc: What happens when we use this driver on another instance of this instrument? You probably meant this to be temporary, working code instead of something you wish to commit definitely to the code base (right?) but I would still fix this asap. Hardcoding values like this is a very big no-no. 

Add a method like `load_calibration_data` which takes a file descriptor (not name) as input and stores the calibration parameters as a private class attribute (e.g. `self._calibration_parameters`). The advantage of a file descriptor instead of a name is that it will make you function more easily testable in CI. An example of how this would work is shown here: 

test: https://github.com/QCoDeS/Qcodes/blob/master/qcodes/tests/drivers/test_lakeshore_file_parser.py
driver: https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/Lakeshore/Model_325.py

maxgoetsch: @sohailc Do you think it is a bad idea to load the calibration by just supplying the filename? If yes, I'll change the code.

astafan8:
@maxgoetsch 

The driver in this pull request will not be supported by QCoDeS developers, and, hence, is a community driver. Please close this pull-request and open a new one against [qcodes_contrib_drivers](https://github.com/QCoDeS/Qcodes_contrib_drivers) repository. QCoDeS developers will close this pull-request unconditionally after 2 weeks.

[Read about qcodes_contrib_drivers repository and python package here.](https://github.com/QCoDeS/Qcodes_contrib_drivers#qcodes-contrib-drivers)

astafan8:
@maxgoetsch

I'm closing this PR. Feel free to open this against [Qcodes_contrib_drivers](https://github.com/QCoDeS/Qcodes_contrib_drivers).

PR #1539, opened 2019-04-15
Add validator: EnumSCPIArgs

sohailc:
The SCPI langauge is case in-sensitive. Furthermore, not all characters in a argument are necessary. For example, the command "SOUR:CURRent 1.0" and "SOUR:CURR 1.0" and "SOUR:current 1.0" are all valid and will produce identical results on e.g. the Keithley 2450 source meter.

In manuals, this is specified by reserving upper case letters for parts which are mandatory and lower case letters for optional parts.

This PR adds a validator which takes this into account. For example: 

```python
self.add_parameter(
            "function",
            set_cmd=":SENS:FUNC \"{}\"",
            get_cmd=":SENS:FUNC?",
            vals=vals.EnumSCPIArgs("CURRent", "VOLTage"),
            get_parser=self._function_get_parser
)
```
Now we can do keithley.function("current") or keithley.function("CURR") which will have the same effect. 

@WilliamHPNielsen can you please have a look


WilliamHPNielsen:
This is an interesting idea, but I think I don't quite get it. Some questions for my understanding:

This is for the case when the user directly writes SCPI commands (or parts of them) to a parameter, right? Why is it preferable to allow the parameter value to be many different things? Doesn't this make the snapshot harder to parse? At a quick glance, it looks like a leaky abstraction, i.e. that the SCPI layer leaks up into the driver that was supposed to abstract that layer away. I'm probably missing something.

Next, for the actual implementation, I tried to read up on what the SCPI standard actually says. I always have troubles finding my way through that mess, but I *think* that page 6-1 of this document http://www.ivifoundation.org/docs/scpi-99.pdf pointing to page 60 in this document http://webuser.unicas.it/misure/MAQ_OLD%20(VO)/Dispense/DISP_7STANDARD%20IEEE%20488_2%201992.pdf tells us that, as you also point out, SCPI is indeed case insensitive. Then `"currenT"` should be valid too, right?

astafan8:
> At a quick glance, it looks like a leaky abstraction, i.e. that the SCPI layer leaks up into the driver that was supposed to abstract that layer away.

I absolutely completely agree with this. (Didn't notice it at a first glance)

I'd suggest to solve this use case with a `val_mapping` from human-readable/human-usable values to corresponding SCPI commands (or even first to smth else, and that smth else to SCPI commands).

PR #1533, opened 2019-04-05
added a driver for the Keithley 2450

sohailc:
A driver for the Keithley 2450 has been added.

PR #1541, opened 2019-04-16
Keithley 6500 parameters had invalid SCPI commands

qSaevar:
In Keithley_6500 driver fix parameters:

- trigger_source
- trigger_delay
- display_enable

@astafan8


jenshnielsen:
It looks like the dmm supports compatibility mode with old dmms. I suspect the original driver was written to support this

qSaevar:
@jenshnielsen It is stated in the class docstring that this is based on the keithley 2000 (where the SCPI commands are valid) and that this is a bet-version. Don't know if there has been a 6500 in the past where the three parameters that I changed worked. My guess is not.

qcodes/instrument_drivers/tektronix/Keithley_6500.py
```diff
@@ -109,10 +109,15 @@ def __init__(self, name, address, reset_device=False, **kwargs):
...
-                           get_cmd='DISP:ENAB?',
-                           set_cmd='DISP:ENAB {}', set_parser=int, vals=Bool())
+        self.add_parameter('display_backlight',
+                           docstring='Control the brightness of the display '
+                                     'backligt. OFF turns the display off and'
+                                     'BLACkout also turns off indicators and '
+                                     'key lights on the device.',
+                           get_cmd='DISP:LIGH:STAT?',
+                           set_cmd='DISP:LIGH:STAT {}',
+                           vals=Enum('ON100', 'ON75', 'ON50', 'ON25', 'OFF',
```
astafan8: i'd recommend using val_mapping here and give user-friendly names to each value.

qcodes/instrument_drivers/tektronix/Keithley_6500.py
```diff
@@ -121,17 +126,32 @@ def __init__(self, name, address, reset_device=False, **kwargs):
...
+                               docstring='Set and read trigger delay for '
+                                         'timer %i.' % trigger,
+                               get_parser=float,
+                               get_cmd='TRIG:TIM%i:DEL?' % trigger,
+                               set_cmd='TRIG:TIM%i:DEL {}' % trigger,
+                               unit='s', vals=Numbers(min_value=0,
+                                                      max_value=999999.999))
+
+            self.add_parameter('trigger%i_source' % trigger,
+                               docstring='Set, and read, the event that sets'
```
astafan8: this docstring is confusing: "set and read the event"?

jenshnielsen:
@qSaevar Yes my point is that the 6500 can be put into a 2000 compatibility mode. I think that could be what this driver is targeting. See  #1533 for something similar example for the the 2450

qSaevar:
@jenshnielsen Ok, I see you point. So, solutions: Flag in the constructor for compatibility mode set by the user? Second driver for the full feature support? Some auto detection for the mode the device is in?

qcodes/instrument_drivers/tektronix/Keithley_6500.py
qSaevar: It is confusing, fixed.

qcodes/instrument_drivers/tektronix/Keithley_6500.py
qSaevar: Done

jenshnielsen:
@qSaevar Yes I think we should test which mode it's in and error appropriately if it's in the wrong mode. I support using the default mode and not any compatibility mode

qSaevar:
@jenshnielsen ok, so I'll add something like:
if language not SCPI:
  raise MeaningfulError
At the end of the constructor.

qSaevar:
I did add a check for the correct command set.

qSaevar:
@astafan8 I addressed your comments, @jenshnielsen concerns and updated the branch. Is there anything else that needs to be done?

qcodes/instrument_drivers/tektronix/Keithley_6500.py
```diff
@@ -149,6 +177,12 @@ def __init__(self, name, address, reset_device=False, **kwargs):
         self.write('FORM:DATA ASCII')
         self.connect_message()
 
+        command_set = self.ask_raw('*LANG?')
```
astafan8: i think this code should go above the `self.write` call. perhaps even to the top of the __init__ (right after the super __init__ call).

Also, no need for using ask_raw - just use ask :)

qSaevar: ✔️

qSaevar: done

PR #1544, opened 2019-04-24
Driver / Keysight : Add 33510B to model list

occoder:
and make the dict style more diff friendly


WilliamHPNielsen:
Hi @occoder, and welcome to QCoDeS! Thank you for improving the Keysight driver.

Just to be completely sure, have you tested that the driver works correctly with the 33510B model and that the only difference between that and other models is the max frequency?

occoder:
Hi  @WilliamHPNielsen 
So far, I just tested a few commands shown in  Qcodes/docs/examples/driver_examples/Qcodes example with Keysight 33500B.ipynb on the real 33510B hardware. They all worked sound.
I may not be able to cover all the corner cases at present.
As a start point, this change seems to be sufficient.
Cheers.

WilliamHPNielsen:
I guess we can do no better than to say "if the example notebook works, everything works".

PR #1546, opened 2019-04-25
Number of oscillators should be 8 if MF enabled

qSaevar:

- If MF is enabled in ZI UHFLI there are 8 oscillators that can be controlled.
- Sampling rate as float should return exact value, not rounded off value. 

@WilliamHPNielsen

### DIFF
--- a/qcodes/instrument_drivers/ZI/ZIUHFLI.py
+++ b/qcodes/instrument_drivers/ZI/ZIUHFLI.py
@@ -1379,9 +1379,8 @@ class ZIUHFLI(Instrument):
                            set_cmd=self._set_samplingrate_as_float,
                            unit='Hz',
                            get_cmd=self._get_samplingrate_as_float,
-                           vals=vals.Enum(
-                               *[ZIUHFLI._convert_to_float(freq) for freq in
-                                self._samplingrate_codes.keys()]),
+                           vals=vals.Enum(*[1.8e9 / 2 ** v for v in
+                                            self._samplingrate_codes.values()]),
                            docstring=""" A numeric representation of the scope's
                              samplingrate parameter. Sets and gets the sampling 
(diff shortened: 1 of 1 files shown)


WilliamHPNielsen:
Let me see if I get this...

I'll just trust you on the number of oscillators. Regarding the frequency, the point is that the graphical interface specifies the available sampling rates as 
```
['1.80 GHz', '900 MHz', '450 MHz', '225 MHz', '113 MHz', '56.2 MHz',
'28.1 MHz', '14.0 MHz', '7.03 MHz', '3.50 MHz', '1.75 MHz', '880 kHz', '440 kHz', '220
kHz', '110 kHz', '54.9 kHz', '27.5 kHz']
```
but you and I both know that these numbers are not exact, merely rounded off to 3 significant digits. In fact, the _true_ frequencies are `1.80 GHz` divided by `2**n`.

Okay, I get it. I agree.

WilliamHPNielsen:
The only weakness is: how well do we know the original frequency? 

@qSaevar ?

qSaevar:
@WilliamHPNielsen the original frequency? Do you mean the base 1.8GHz or the internal ref frequency? In the manual it is stated that the short term stability of the internal clock is 0.00005 ppm with typical initial accuracy of +/- 0.5 ppm and max of +/- 1 ppm. But then there is the possibility of using an external reference clock...

qSaevar:
@WilliamHPNielsen I fixed the setter for sampling rate and I have tested this on hardware. Should I do anything else? Or what is the status of this PR.

WilliamHPNielsen:
@qSaevar I meant the base frequency of 1.8 GHz. If that could potentially be 1.804 GHz, things would be bad. But you answered that question now.

We'll merge once CI is happy.

qSaevar:
@WilliamHPNielsen That's great, thanks

PR #1547, opened 2019-04-25
Remove dataset add_parameter

WilliamHPNielsen:
With the merge of #1477, the method `add_parameter` of the `DataSet` became obsolete. This PR removes the method from the `DataSet` API by breaking it. Removing that method will in turn make development of the interdependencies stuff easier, since the method `extend_with_paramspec` will no longer have to exist.

If you wish, you can view this PR as the first "clean up the fallout of #1477" PR. There are more to come, including schema upgrades etc.

~Still pending: upgrade all tests to use `set_interdependencies` instead of `add_parameter`.~

I am open to input on how to best break/remove/deprecate `DataSet.add_parameter`, but I stress the point that its existence is a bit of a roadblock and that I therefore find it justified to simply kill it without going through a deprecation cycle. Looking ahead, we will need to break a lot of stuff and I think we should allow ourselves to move quickly.

@QCoDeS/core this is now ready for review.


WilliamHPNielsen:
@QCoDeS/core, bump, bump. All the commits here are contained in #1555, but it might be good to get this in first. We can also do both in one go. What do you think?

astafan8:
Let's do these PRs one-by-one. I'll have a look at this tomorrow.

qcodes/dataset/data_set.py
```diff
@@ -497,27 +497,11 @@ def toggle_debug(self):
...
-        if self.parameters:
-            old_params = self.parameters.split(',')
-        else:
-            old_params = []
-
-        if spec.name in old_params:
-            raise ValueError(f'Duplicate parameter name: {spec.name}')
-
-        self._interdeps = self._interdeps._extend_with_paramspec(spec)
+        raise NotImplementedError('This method has been removed. '
```
astafan8: please add a test for this :)

jenshnielsen:
I am not super keen on removing code that is public api without a deprecation cycle but I don't have a good solution. Should we check that this is not called from any code that we know of before doing this?

astafan8:
Code that i know (and i checked) of does not use DataSet.add_parameter.

WilliamHPNielsen:
No occurrences in main qcodes (obviously) nor in the current master of `qdev-wrappers`.

PR #1554, opened 2019-05-02
add parameter for Keysight E8267D IQ modulator 

eendebakpt:
@astafan8 @WilliamHPNielsen

### DIFF
--- a/qcodes/instrument_drivers/Keysight/Keysight_E8267D.py
+++ b/qcodes/instrument_drivers/Keysight/Keysight_E8267D.py
@@ -61,7 +61,7 @@ class Keysight_E8267D(VisaInstrument):
                            set_cmd='OUTP {}',
                            get_parser=parse_on_off,
                            vals=on_off_validator)
-        self.add_parameter('status',
+        self.add_parameter('IQmodulator',
                            get_cmd='DM:STATe?',
                            set_cmd='DM:STATe {}',
                            get_parser=parse_on_off,
--- a/qcodes/instrument_drivers/Keysight/Keysight_E8267D.py
+++ b/qcodes/instrument_drivers/Keysight/Keysight_E8267D.py
@@ -65,8 +65,15 @@ class Keysight_E8267D(VisaInstrument):
                            get_cmd='DM:STATe?',
                            set_cmd='DM:STATe {}',
                            get_parser=parse_on_off,
-                           on_off_validator,
+                           vals=on_off_validator,
                            docstring='Enables or disables the internal I/Q modulator. Source can be external or internal.')
+
+        for source in [1, 2]:
+            self.add_parameter(f'IQsource{source}',
+                               get_cmd=f'DM:SOUR{source}?',
+                               set_cmd=f'DM:SOUR{source} {{}}',
+                               vals=vals.Enum('OFF', 'EXT', 'EXT600', 'INT'),
+                               docstring=f'Source {source} for I/Q modulation.')
 
         self.connect_message()
 
--- a/qcodes/instrument_drivers/Keysight/Keysight_E8267D.py
+++ b/qcodes/instrument_drivers/Keysight/Keysight_E8267D.py
@@ -1,5 +1,6 @@
-from numpy import rad2deg, deg2rad
+import numpy as np
 from qcodes import VisaInstrument, validators as vals
+from qcodes.utils.validators import Numbers
 
 
 def parse_on_off(stat):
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/Keysight/Keysight_E8267D.py
```diff
@@ -27,14 +27,19 @@ class Keysight_E8267D(VisaInstrument):
     def __init__(self, name, address, step_attenuator=False, **kwargs):
         super().__init__(name, address, **kwargs)
 
+        # Only listed most common spellings idealy want a
+        # .upper val for Enum or string
+        on_off_validator = vals.Enum('on', 'On', 'ON',
```
astafan8: this is a great case for using [`create_on_off_val_mapping`](http://qcodes.github.io/Qcodes/examples/Creating%20Instrument%20Drivers.html?highlight=create_on_off#On/Off-parameters) function. please use it :)

qcodes/instrument_drivers/Keysight/Keysight_E8267D.py
```diff
@@ -55,10 +60,20 @@ def __init__(self, name, address, step_attenuator=False, **kwargs):
                            get_cmd=':OUTP?',
                            set_cmd='OUTP {}',
                            get_parser=parse_on_off,
-                           # Only listed most common spellings idealy want a
-                           # .upper val for Enum or string
-                           vals=vals.Enum('on', 'On', 'ON',
-                                          'off', 'Off', 'OFF'))
+                           vals=on_off_validator)
+        self.add_parameter('IQmodulator',
```
astafan8: please also follow the naming recommendation [mentioned here](http://qcodes.github.io/Qcodes/examples/Creating%20Instrument%20Drivers.html?highlight=create_on_off#On/Off-parameters), so that this parameter should be called smth like `IQ_modulator_enabled`.

qcodes/instrument_drivers/Keysight/Keysight_E8267D.py
```diff
@@ -55,10 +60,20 @@ def __init__(self, name, address, step_attenuator=False, **kwargs):
...
+                           set_cmd='DM:STATe {}',
+                           get_parser=parse_on_off,
+                           vals=on_off_validator,
+                           docstring='Enables or disables the internal I/Q modulator. Source can be external or internal.')
+
+        for source in [1, 2]:
+            self.add_parameter(f'IQsource{source}',
+                               get_cmd=f'DM:SOUR{source}?',
+                               set_cmd=f'DM:SOUR{source} {{}}',
+                               vals=vals.Enum('OFF', 'EXT', 'EXT600', 'INT'),
```
astafan8: does it make sense to give these values human-friendly names, and use val_mapping to map them to these SCPI values? i'd really recommend this.

qcodes/instrument_drivers/Keysight/Keysight_E8267D.py
peendebak: @astafan8 That would indeed be a good case, but it would make the driver backwards incompatible (with the `create_on_off_val_mapping` the parameter will return a 0, instead of 'Off').

qcodes/instrument_drivers/Keysight/Keysight_E8267D.py
peendebak: From the manual:

*EXTernal*  This choice selects an external 50 ohm source as the I/Q input to I/Q modulator.
*INTernal*    This  choice  is  for  backward  compatibility  with  ESG  E44xxB  models  and  performs  the  same  function  as  the  BBG1  selection.
*BBG1*    This  choice  selects  the  baseband  generator  as  the  source  for  the  I/Q  modulator.
*EXT600*    This  choice  selects  a  600  ohm  impedance  for  the  I  and  Q  input  connectors  and  routes  the  applied  signals  to  the  I/Q  modulator.
*OFF*    This  choice  disables  the  I/Q  inpu
```
What would good names be? 

@astafan8

File: qcodes/instrument_drivers/Keysight/Keysight_E8267D.py (hunk 2 above)
Comment (peendebak) [reply qcodes/instrument_drivers/Keysight/Keysight_E8267D.py]: Done

File: qcodes/instrument_drivers/Keysight/Keysight_E8267D.py (hunk 1 above)
Comment (astafan8) [reply qcodes/instrument_drivers/Keysight/Keysight_E8267D.py]: damn, indeed. ok, let's not introduce backwards incompatibility.

File: qcodes/instrument_drivers/Keysight/Keysight_E8267D.py (hunk 3 above)
Comment (astafan8) [reply qcodes/instrument_drivers/Keysight/Keysight_E8267D.py]: well, i would do smth like the following:
```
{
"external_50_ohm": "EXT",
"internal": "INT",
"baseband_generator": "BBG1",
"external_600_ohm": "EXT600",
"off": "OFF"
}
```
and also add a `docstring=....` kwarg to this `add_parameter` call with the text that explain all these values so that the user does not need to look into the manual.

But if you think it's overkill, or unnecessary humanization of the values, feel free to use whatever you think is reasonable.

File: qcodes/instrument_drivers/Keysight/Keysight_E8267D.py [hunk 4]
```diff
@@ -27,21 +28,39 @@ class Keysight_E8267D(VisaInstrument):
...
+                           get_cmd='FREQ:OFFS?',
+                           set_cmd='FREQ:OFFS {}',
+                           get_parser=float,
+                           vals=Numbers(min_value=-200e9,
+                                        max_value=200e9))
+        self.add_parameter('frequency_mode',
+                           label='Frequency mode',
+                           set_cmd='FREQ:MODE {}',
+                           get_cmd='FREQ:MODE?',
+                           vals=vals.Enum('FIX', 'CW', 'SWE', 'LIST'))
```
astafan8: same comment here as for the "IQsource#" parameters - would human-readable values be better here? say (just guessing) "fixed" for "FIX", "continuous_wave" for "CW", etc? and a nice `docstring=...`?

peendebak: I copied this one from the agilent E8267 driver. But it might make sense to change the names now.

peendebak:
@astafan8 I added `create_on_off_val_mapping`. The argument names I kept the same, so they keep matching the command names on the actual instrument.

qcodes/instrument_drivers/Keysight/Keysight_E8267D.py
astafan8: ok, it's fine. And thanks for the docstring.

PR #1556, opened 2019-05-03
the requirements should be listed in the setup file and the requireme…

sohailc:
As a library, QCoDeS should have its dependencies included in the 'install_requires' argument in the setup.py file, not in the requirements.txt file. This is explained nicely here: https://caremad.io/posts/2013/07/setup-vs-requirement/ 

Motivation for this change: 
When listing QCoDeS as a dependency in the requirements.txt file of and external project which relies on QCoDeS, I noticed that not all the dependencies of QCoDeS are installed automatically, when installing my own project. Specifically, matplotlib, pyqtgraph and PyQt seemed to be missing. This is because these dependencies are not listed in the 'install_requires' list given as argument to the setup function. Evidently, the setup.py file and the requirements.txt file grew out of sync. 

The proper solution to this is to list all requirements in the setup file and having the requirements.txt reference the setup.py file, as explained in the URL above.


WilliamHPNielsen:
I think the discussion here is relevant: #1425

jenshnielsen:
Hi @sohailc  The intention of separating things into to requirements files is to separate the concrete mimiumum dependencies along the lines that article suggests. I.e qcodes can run with any version of numpy  >= 1.10 so installing qcodes will not force you to upgrade. But to create a reproducible environment we list a specific version. This is lost with your change. The same logic applies to h5py 
We do neither need nor want to enforce that all users of qcodes internally or externally have exactly the same versions installed. 


If you look in the extra section of the setup.py file you will see that matplotlib and and pyqtgraph are already listed. This is done like that because they were only needed for certain specific features. 
The version of matplotlib listed here is obviously out of date so we should fix that
and I think at this stage we can just make matplotlib a general dependency. 

```
extras = {
    'MatPlot': ('matplotlib', '2.2.3'),
    'QtPlot': ('pyqtgraph', '0.10.0'),
    'coverage tests': ('coverage', '4.0'),
    'Slack': ('slacker', '0.9.42')
}
```

QCoDeS does not depend on PyQtGraph. Never has and never will. 
It is an optional dependency that is only used for one optional part of the code. For the same
reason pandas is not required by default. 

We cannot list pyqt5 (in addition to the fact that its not a dependency since matplotlib is functional with multiple backends and pyqtgraph is not a dependency) in setup.py due to a bug in the naming of the package using conda.

Consider the following example. 

```
conda crate -n qttest
conda install pyqt
# now if you list pyqt5 in setup.py this is equivalent to pip installing it.
pip install pyqt5
# observe how two differently named packages are installed on top of each other
$ conda list
# packages in environment at C:\Users\jenielse\AppData\Local\Continuum\miniconda3\envs\qttest:
#
# Name                    Version                   Build  Channel
ca-certificates           2019.1.23                     0
certifi                   2019.3.9                 py37_0
icu                       58.2                 ha66f8fd_1
jpeg                      9b                   hb83a4c4_2
libpng                    1.6.37               h2a8f88b_0
openssl                   1.1.1b               he774522_1
pip                       19.1                     py37_0
pyqt                      5.9.2            py37h6538335_2
pyqt5                     5.12.1                   pypi_0    pypi
pyqt5-sip                 4.19.15                  pypi_0    pypi
python                    3.7.3                h8c8aaf0_1
qt                        5.9.7            vc14h73c81de_0
setuptools                41.0.1                   py37_0
sip                       4.19.8           py37h6538335_0
sqlite                    3.28.0               he774522_0
vc                        14.1                 h0510ff6_4
vs2015_runtime            14.15.26706          h3a45250_4
wheel                     0.33.1                   py37_0
wincertstore              0.2                      py37_0
zlib                      1.2.11               h62dcd97_3
(qttest)
```
Observe the inconsistent environment with two differently named packages installed on top of each other. In the best case this will give you whatever package was installed last. In the worst case it will result in crashes when importing pyqt.  


By the way putting the `-e .` in the requirements.txt file seems to me to be an extremely bad idea. What happens if you want to deploy qcodes as a non editable install using requirements.txt. You can't without changing the file or reinstalling QCoDeS afterwards. You could put `.` in the file but then you have the opposite problem.

sohailc:
Thanks for your detailed answer @jenshnielsen. I guess I lack understanding here. 

But let's say I have a library that requires QCoDeS, so I make a requirements.txt file and list QCoDeS in there. Running `pip install -e .` results in QCoDeS being installed in my environment, but importing QCoDeS from the python prompt generates a warning that matplotlib and pyqtgraph are not present. I had hoped that all dependencies of QCoDeS would have been installed, including matplotlib, so I would not get this warning every time I import my library. 

I guess the correct way to do this is to list matplotlib separately in my own requirements file. 

Thanks again for the detailed explaination.

astafan8:
@QCoDeS/core @sohailc shall we close this? or what is the status? my humble opinion is that we should keep running with our current way of sorting out package dependencies.

sohailc:
Yeah I will close this

PR #1557, opened 2019-05-06
Deprecate is_number

jenshnielsen:
No longer used after #1477 and should probably never have been public in the first place


qcodes/dataset/measurements.py
```diff
@@ -45,6 +46,7 @@ class ParameterTypeError(Exception):
     pass
 
 
+@deprecate("This function is no longer used and should have been private")
```
astafan8:
```suggestion
@deprecate("This function is no longer used and will be removed soon.")
```

PR #1558, opened 2019-05-06
AMI430 driver: fix using numeric field_limit

astafan8:
The original user problem was that the driver would raise `ValueError` when initialized with a floating point value. This prevented the user to set the field limit to smth like 2.1T, and the workaround was to use an unsafe limit of 3T. 

This is the code:
https://github.com/QCoDeS/Qcodes/blob/1923c8d5c60580c902c93ec5e22b2cfd2bede401/qcodes/instrument_drivers/american_magnetics/AMI430.py#L493-L497

The problem is that `repr(field_limit).isnumeric()` is used to check whether the `field_limit` is "numeric" - which returns `True` for integer values but returns `False` for floating point values. Also, it's pretty bizarre that a string method is used to check for "numeric-ness", and is used wrongly.

This PR implements a fix for it, see the list below.

List of changes:
* Use `isinstance` for type checking instead of `repr(field_limit).isnumeric()`
* Make sure numeric field_limit is converted to `float` internally
* Annotate field_limit argument type in AMI430 driver
* Clarify error message for the field limit type upon AMI430 initialization
* Test using int and float field_limit in AMI430 driver
* Make separate fixture for magnet axes instances in AMI430 magnet driver tests

Thanks to @vhartong for informing about this issue.

### DIFF
--- a/qcodes/instrument_drivers/american_magnetics/AMI430.py
+++ b/qcodes/instrument_drivers/american_magnetics/AMI430.py
@@ -672,8 +672,7 @@ class AMI430_3D(Instrument):
         )
 
     def _verify_safe_setpoint(self, setpoint_values):
-
-        if repr(self._field_limit).isnumeric():
+        if isinstance(self._field_limit, numbers.Real):
             return np.linalg.norm(setpoint_values) < self._field_limit
 
         answer = any([limit_function(*setpoint_values) for
--- a/qcodes/instrument_drivers/american_magnetics/AMI430.py
+++ b/qcodes/instrument_drivers/american_magnetics/AMI430.py
@@ -14,6 +14,9 @@ from qcodes.utils.validators import Bool, Numbers, Ints, Anything
 
 log = logging.getLogger(__name__)
 
+CartesianFieldLimitFunction = \
+    Callable[[numbers.Real, numbers.Real, numbers.Real], bool]
+
 
 class AMI430Exception(Exception):
     pass
@@ -476,7 +479,8 @@ class AMI430(IPInstrument):
 class AMI430_3D(Instrument):
     def __init__(self, name,
                  instrument_x, instrument_y, instrument_z,
-                 field_limit: Union[numbers.Real, Iterable[Callable]],
+                 field_limit: Union[numbers.Real,
+                                    Iterable[CartesianFieldLimitFunction]],
                  **kwargs):
         super().__init__(name, **kwargs)
 
@@ -494,7 +498,7 @@ class AMI430_3D(Instrument):
         self._instrument_y = instrument_y
         self._instrument_z = instrument_z
 
-        self._field_limit: Union[float, Iterable[Callable]]
+        self._field_limit: Union[float, Iterable[CartesianFieldLimitFunction]]
         if isinstance(field_limit, collections.abc.Iterable):
             self._field_limit = field_limit
         elif isinstance(field_limit, numbers.Real):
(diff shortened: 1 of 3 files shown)


qcodes/tests/drivers/test_ami430.py
```diff
@@ -424,3 +435,19 @@ def test_blocking_ramp_parameter(current_driver, caplog):
 
         assert len([mssg for mssg in messages if 'blocking' in mssg]) == 0
 
+
+@pytest.mark.parametrize('field_limit', [2, 2.2], ids=['integer', 'float'])
```
jenshnielsen: Should we have tests with numpy integer and float data types just in case

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -472,8 +474,10 @@ def _update_units(self, ramp_rate_units=None, field_units=None):
 
 
 class AMI430_3D(Instrument):
-    def __init__(self, name, instrument_x, instrument_y,
-                 instrument_z, field_limit, **kwargs):
+    def __init__(self, name,
+                 instrument_x, instrument_y, instrument_z,
+                 field_limit: Union[numbers.Real, Iterable[Callable]],
```
jenshnielsen: Should this be a `Callable[[numerictype]], bool]` to make it even more strict

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -490,11 +494,15 @@ def __init__(self, name, instrument_x, instrument_y,
         self._instrument_y = instrument_y
         self._instrument_z = instrument_z
 
-        if repr(field_limit).isnumeric() or isinstance(field_limit, collections.abc.Iterable):
+        self._field_limit: Union[float, Iterable[Callable]]
+        if isinstance(field_limit, collections.abc.Iterable):
             self._field_limit = field_limit
+        elif isinstance(field_limit, numbers.Real):
```
jenshnielsen: Does this evaluate to True for numpy float/ints

qcodes/tests/drivers/test_ami430.py
astafan8: sure! shall i parametrize over all concrete and non-concrete types from qcodes/utils/types.py?

jenshnielsen: Perhaps that is overkill but sounds good

qcodes/instrument_drivers/american_magnetics/AMI430.py
astafan8: yes, for all the "real" types mentioned in qcodes.utils.types this evaluates to True (tested, not just claiming).

qcodes/tests/drivers/test_ami430.py
astafan8: yes, it is an overkill... but i still did it, so let me push it as well...

qcodes/instrument_drivers/american_magnetics/AMI430.py
astafan8: sure. shall i use numbers.Real again for `numerictype`?

jenshnielsen: Yes if that works. Otherwise we may need to define a union in types.py

astafan8: done.

astafan8:
@jenshnielsen is this good? if so, after testing on real magnet (hopefully today) i will merge.

astafan8:
Tested with Victor just now. It works. So i merge :)

PR #1560, opened 2019-05-13
Extend QCoDeS-Station to be configurable through yaml file

Dominik-Vogel:
# Configurable Station
This PR integrates the `StationConfigurator` from qdev-dk/qdev-wrappers#68 into the QCoDeS station object.

(see https://github.com/QCoDeS/Qcodes/pull/1256 for complete discussion history)

The StationConfigurator has been a separate object in qdev_wrappers because the Station lives in qcodes core, but it conceptually belongs into the Station.
The Station is meant to be a representation of the physical setup. And this description is given by the station config yaml file.

## What the `Station` does
So far the station has methods for snapshotting:
 - snapshot_base
 - add_component
and default measurements:
 - set_measurement
 - measure

## Additions
and in this pr methods for serialization from a yaml are added:
 - load_config_file
 - load_instrument

I suggest that we deprecate the default measurement functionality in another PR, because as far as I know it is not used and it is only used by the legacy loop. Removing these functions would further decouple dependencies.

## Migrating from the `StationConfigurator`

Previous bootstrapping looked like this:
- create `Station` (no arguments)
- create `StationConfigurator`, pass station
(- call station specific init method/script to specify `sample_name` see point 2 in future improvements.)

With this an experiment is bootstrapped by simply:
-create station (potentially with sample name as argument)

Moving from the `StationConfigurator` to the new `Station` object will require two simple changes:
 - change `<station_configurator_object>.load_instrument(...)` to `<station_obect>.load_instrument(...)` and remove instantiation of `<station_configurator_object>` 
 - change the configuration in ~\qcodesrc.json by replacing the section name `station_configurator` by `station`

# Future improvements

To go in small steps this PR addresses the minimal integration and documentation of the station configurator in the station. In later PRs documentation and possible refactorings of the internal structure might follow. Lets think about what we are committing to with this PR.

The next steps are:
  1. add support for virtual instruments:
    - add `source_instrument` as an attribute of the `add_parameter` section to allow to instantiation of `DelegateParameter`s with a source from another instrument.
    - enable to automatically load those instruments if they are defined in the same yaml file.
    - add a suitable base class for a purely virtual instrument.
  2. Accommodate the control over the sample name in the Station object. (this would also satisfy @ThorvaldLarsen s request.)


astafan8:
@jenshnielsen @WilliamHPNielsen please, review this. I did my review in the previous PR, we need fresh eyes :)

Dominik-Vogel:
Thanks Jens for all the good comments, I'll work on them.

qcodes/station.py
```diff
@@ -217,3 +258,224 @@ def __getitem__(self, key):
...
+                        'Could not load default config for Station: \n'
+                        f'File {get_config_default_file()} not found. \n'
+                        'You can change the default config file in '
+                        '`qcodesrc.json`.')
+                return
+
+        with open(path, 'r') as f:
+            self.load_config(f)
+
+    def load_config(self, config: Union[str, IO[AnyStr]]) -> None:
```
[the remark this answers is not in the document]
Dominik-Vogel: Maybe it would make most sense to make this a private method.

qcodes/station.py
```diff
@@ -217,3 +258,245 @@ def __getitem__(self, key):
...
+        self.remove_component(instrument.name)
+        # del will remove weakref and close the instrument
+        instrument.close()
+        del instrument
+
+    def load_instrument(self, identifier: str,
+                        revive_instance: bool = False,
+                        **kwargs) -> Instrument:
+        """
+        Creates an :class:`~qcodes.Instrument` instance as described by the
```
jenshnielsen: I think you should either use the full path to the actual Class or just simply  `.Instrument` which should resolve to the correct one

Dominik-Vogel: I was also unsure about it, so I tested it locally and it worked. As CI is failing I will update it to the full path.

WilliamHPNielsen:
I have no objections.

PR #1561, opened 2019-05-13
Remove warning when taking snapshot of ZIHDAWG8

qSaevar:
- update parameter in ZIHDAWG.snapshot_base should work as advertised.
- feature_code should be ignored when getting an update snapshot.

@WilliamHPNielsen

### DIFF
--- a/qcodes/instrument_drivers/ZI/ZIHDAWG8.py
+++ b/qcodes/instrument_drivers/ZI/ZIHDAWG8.py
@@ -64,7 +64,7 @@ class ZIHDAWG8(Instrument):
         self._compiler_sleep_time = 0.01
 
     def snapshot_base(self, update: bool = True,
-                      params_to_skip_update: Sequence[str] = 'features_code') -> Dict:
+                      params_to_skip_update: Sequence[str] = ('features_code',)) -> Dict:
         """ Override the base method to ignore 'feature_code' by default."""
         return super(ZIHDAWG8, self).snapshot_base(update=update,
                                                    params_to_skip_update=params_to_skip_update)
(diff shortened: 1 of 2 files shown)


qcodes/instrument_drivers/ZI/ZIHDAWG8.py
```diff
@@ -63,11 +63,12 @@ def __init__(self, name: str, device_id: str, **kwargs) -> None:
         self.warnings_as_errors: List[str] = []
         self._compiler_sleep_time = 0.01
 
-    def snapshot_base(self, update: bool = False,
+    def snapshot_base(self, update: bool = True,
```
jenshnielsen: This silently drops ```params_to_skip_update```, perhaps we should either append `features_code` to whatever is passed or simple make the default value ```('features_code',)```

qSaevar: I like appending `features_code` to whatever is passed, rather than having it as a default value.

qcodes/instrument_drivers/ZI/ZIHDAWG8.py
```diff
@@ -63,11 +63,15 @@ def __init__(self, name: str, device_id: str, **kwargs) -> None:
         self.warnings_as_errors: List[str] = []
         self._compiler_sleep_time = 0.01
 
-    def snapshot_base(self, update: bool = False,
-                      params_to_skip_update: Sequence[str] = None) -> Dict:
-        """ Override the base method to ensure all values are up-to-date"""
-        return super(ZIHDAWG8, self).snapshot_base(update=True,
-                                                   params_to_skip_update=None)
+    def snapshot_base(self, update: bool = True,
+                      params_to_skip_update: Sequence[str] = 'features_code') -> Dict:
```
jenshnielsen: As this is a sequence of strings it should be `('features_code',)`

qSaevar: That's right

jenshnielsen:
Will merge once ci passes

PR #1562, opened 2019-05-13
DB upgrade: version 5 - snapshot column ALWAYS present in runs table

astafan8:
Make `snapshot` column (of type "TEXT") always present in `runs` table. This way it becomes part of the stable schema, not a "metadata" column anymore.

Relies on version 4 database generator from here https://github.com/QCoDeS/qcodes_generate_test_db/pull/3.

Also couldn't resist fixing some of the style stuff.

ToDo:
- [x] remove checking out branch of `qcodes_generate_test_db`
- [x] update the hash in `qcodes_generate_test_db` to really the latest hash before merging this PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,7 @@ script:
       cd ..
       git clone https://github.com/QCoDeS/qcodes_generate_test_db.git
       cd qcodes_generate_test_db
+      git checkout bumping-to-version-5
       python generate_version_0.py
       python generate_version_1.py
       python generate_version_2.py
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -45,11 +45,13 @@ script:
       cd ..
       git clone https://github.com/QCoDeS/qcodes_generate_test_db.git
       cd qcodes_generate_test_db
+      git checkout bumping-to-version-5
```
jenshnielsen: I assume that this should go and the branch merged before this goes in?

astafan8: absolutely (see todo above)

PR #1565, opened 2019-05-15
Don't upload htmlcov in Azure pipelines

jenshnielsen:
Pipelines warns that its unused anyway


jenshnielsen:
Travis uses a different codecover tool so I think this is fine

PR #1567, opened 2019-05-15
Add pip to conda env

jenshnielsen:
Conda warns that pip is not installed but used
fix this by adding pip as a dependency


jenshnielsen:
The warning it prints is 

```Warning: you have pip-installed dependencies in your environment file, but you do not list pip itself as one of your conda dependencies.  Conda may not use the correct pip to install your packages, and they may end up in the wrong place.  Please add an explicit pip dependency.  I'm adding one for you, but still nagging you.
```

I interpret that as it will do the "right" thing but that may change in the future and is unsupported

PR #1568, opened 2019-05-16
Reorg Parameter notebooks and add basic complex numbers notebook

WilliamHPNielsen:
As a warm-up exercise for extending `plot_by_id` to handle complex input, I made a small notebook about complex-valued parameters. We might as well get that in. And we might as well move the notebooks a bit about, now that we're at it.

The notebook doesn't use the dataset, and should as such live at the "top level".

- Add a little notebook about complex-valued parameters. In practice, that means the `ComplexNumbers` and the `Arrays` validators

@QCoDeS/core


docs/examples/Complex Numbers.ipynb
```diff
@@ -0,0 +1,318 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Yeah, those are complex numbers\n"
+     ]
+    }
+   ],
+   "source": [
+    "amps_val_strict.validate(np.array([1.0 + 0j, 2.0, 3.0, 4.0, 5.0]))\n",
```
jenshnielsen: should we make it clearer whats going on by doing something along the lines of

```
amp_array = np.array([1.0 + 0j, 2.0, 3.0, 4.0, 5.0])
print(amp_array.dtype)
amps_val_strict.validate(amp_array)
```

docs/examples/Complex Numbers.ipynb
```diff
@@ -0,0 +1,318 @@
...
+    "The proper validator should accept complex numbers and should reject anything of the wrong shape. Note that we get to decide whether we want to accept \"non-strictly complex\" data."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "amps_val_strict = Arrays(shape=(5,), valid_types=(np.complex128, np.complex64))\n",
```
jenshnielsen: Should we use `np.complex` in the same way that we use the none size specific int and float types

docs/examples/Complex Numbers.ipynb
WilliamHPNielsen: Yes!

docs/examples/Complex Numbers.ipynb
WilliamHPNielsen: Certainly.

jenshnielsen:
Looks good. Since we have learned that sphinx does not always check links to notebooks can you grep and make sure that there are no now broken links to these notebooks?

jenshnielsen:
Perhaps this notebook https://qcodes.github.io/Qcodes/examples/writing_drivers/Simple-Example-of-ParameterWithSetpoints.html should be moved there too

jenshnielsen:
@WilliamHPNielsen Should we get this in? Would be nice to ,move that other notebook

docs/user/tutorial.rst
```diff
@@ -8,7 +8,7 @@ Tutorial
 Writing a Driver
 ----------------
 
-First look at what parameters are and how to create them: `qcodes parameter <https://github.com/QCoDeS/Qcodes/blob/master/docs/examples/Parameters.ipynb>`__ .
```
jenshnielsen: We really need to update the user guide and tutorial asap

astafan8: yep .... :(

jenshnielsen:
I have moved that notebook and updated all links in the documentation that points to any of the moved notebooks

PR #1570, opened 2019-05-17
Reorganization of sqlite_base.py

astafan8:
The reason of this refactoring is to allow database upgraders to have snippets of code from qcodes that are only relevant to those database versions. For example, when we remove layouts and dependencies tables we will still preserve the code that works with them "as is" in a module for the relevant database upgrade, and will change the sqlite API that qcodes uses as we need (this is the reason for Codacy to report problems on this PR).

The new structure is somewhat explained in the top-level docstrings of the new `sqlite` modules and sub-modules. I also tried to make good commits, so it might be helpful to just follow the commit history to see what happened.

The code has only been moving, and import statements have been adjusted. So, the functionality and bodies of functions are intact.

Note that `dataset/sqlite_base.py` and `dataset/database.py` modules are now NOT used inside qcodes, but are left with imported entities to facilitate migration of other dev branches of qcodes, and, __unfortunately__, some third-party tools. All the new development shall be done in the new `sqlite` package.

Note imports of two notebooks were updated due to this change.

not gonna do now:
- !!! apply some privatization scheme so that sqlite stuff does not get used all-over-the-place (to-be-discussed offline)
- move sqlite3 adapters and converters into their own module(s)
- go through all of the `dataset` module and move all sqlite-query code (for example, in experiment container module) to `sqlite.*`

@QCoDeS/core


astafan8:
@QCoDeS/core I would like to try to be annoying and propose to take care of `apply some privatization scheme so that sqlite stuff does not get used all-over-the-place` in a different PR. So, with your review, please tell if you agree with the structure or not (suggestions/questions are welcome).

qcodes/dataset/sqlite/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+         /     |     V       V
+        |      |  .query_helpers
+        |      |   |       |
+        |      V   V       |
+        |  .db_upgrades    |
+        |      /           V
+        |     /        .queries
+        v    v
+      .database
+
```
jenshnielsen: Cool, now we just need to remember to keep this up to date

qcodes/dataset/sqlite_base.py
```diff
@@ -1,2755 +1,47 @@
...
-                        "run_counter", "format_string"]
-
-    for mcond in match_conditions:
-        if mcond not in valid_conditions:
-            raise ValueError(f"{mcond} is not a valid match condition.")
-
-    end_time = match_conditions.get('end_time', None)
-    time_eq = "=" if end_time is not None else "IS"
-
-    sample_name = match_conditi
```
jenshnielsen: As above should this raise a warning on import?

qcodes/dataset/database.py
```diff
@@ -1,57 +1,9 @@
...
-    row = cursor.fetchall()[0]
-
-    return row[2]
+"""
+Code of this module has been moved to `.sqlite.database`. This module now
+only re-imports the functions which it used to contain, for backwards
+compatibility. Do not import functions from this module because it will be
+removed soon.
+"""
+
```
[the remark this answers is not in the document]
astafan8: yes, good point. i'll add a warning

qcodes/dataset/sqlite_base.py
astafan8: yes, good point. i'll add a warning

qcodes/dataset/sqlite/__init__.py
astafan8: yes ;) that's why the text above the diagram a bit vague :)

PR #1572, opened 2019-05-20
Dont read from layouts and dependencies tables in QCoDeS

astafan8:
This PR removes usage of functions which __read__ information from `layouts` and `dependencies` tables. Reading information from `layouts` and `dependencies` tables is substituted with reading `RunDescriber` object (stored in `run_description` column of `runs` table). This PR does NOT remove _writing_ to `layouts` and `dependencies` tables, this will be done in another PR.

The big news in this PR is that `get_data_by_id` and thus `plot_by_id` is now free from using the `layouts` and `dependencies` tables.

This PR is just one of the PRs on the way towards removing the `layouts` and `dependencies` tables completely from QCoDeS database.

ToDo:
- [x] whenever necessary remove usage of private attributes to access RunDescriber and/or InterDependencies
- [x] `flatten` data arrays `get_data_by_id` or not? the tests of `get_data_by_id` say "yes". but perhaps when plot_by_id is improved (that's in a different PR) then it can use get_parameter_data directly, and then `get_data_by_id` can be deprecated; but within the scope of this PR we just make sure that the tests work and plotting notebooks still produce correct plots (since plot_by_id uses get_data_by_id that is drastically adjusted in this PR)


WilliamHPNielsen:
I'd say this PR is ready for review. Note that it is not completely true that we don't read from `layouts` and `dependencies`; there is a single place (in the current API) where that happens: in the `create_run` function from `sqlite.queries`.

@QCoDeS/core

astafan8:
> there is a single place (in the current API) where we still read from `layouts` and `dependencies`: in the `create_run` function from `sqlite.queries`

do you mean the `_get_layout_id` inside `_add_parameters_to_layout_and_deps` inside `_insert_run`? if so, that's fine, but this "reading" is now done only with "private" methods of `sqlite.queries` and it will go away together with the tables (because everything will be tied to `run_description` column)

PR #1573, opened 2019-05-20
Driver for Keysight B1500 Parameter Analyzer

StefanD986:
Adds a driver for the Keysight B1500 Parameter Analyzer

As further described in the example notebook, this implements:
 - a high level interface that implements a small subset of commonly used functionality
   - the SMU high level driver works
   - the CMU high level driver still needs some work and practice testing
 - a low-level helper interface to generate messages that can be sent to the instrument
   - The implementation is complete, so that full functionality of the instrument can be used via low-level commands
   - The typehints for the low-level function are complete which is a first help for the user
   - The docstring for each command is not complete (halfway during writing the documentation I decided that users should refer to the manual, due to the complexity of the instrument)
 - Example notebook that outlines the usage of both, the low-level as well as the high-level interface usage.
 - Access to more functions in the High Level Interface
   - NPLC Setting
   - ADC type configuration

The functionality of the High-level driver is limited, however complete enough for what we currently use in the lab.

The API docs for the modules of this driver will NOT be generated because I have no clue how to avoid sphinx warning about duplicated documentation of class `KeysightB1500` due to the convenience imports in `keysightb1500/__init__.py` module. I fought for this for too long, so, just ignoring this.

Additionally had to:
- improve a bit `makefile` and `make.bat` file for building docs
- fix typing and docstring of `is_sequence_of`
- fix `create_on_off_val_mapping` for the case of default arguments (`True`/`False`)

Here is a ToDo List of items that we might need in the near future __BUT NOT IN THIS PR__:
 - Sampling Measurement Mode to measure time traces
 - use `InstrumentChannel` for each channel rather than for each module? (modules may in principle have multiple channels, but it's not the case for the modules implemented so far)
- deal with the complex data output formats that the instrument supports (there are quite a few TODOs in the code for that), but it's not critical for now.


astafan8:
@QCoDeS/core just for your information, I will take care of this PR. I'll keep you posted on this.

StefanD986:
👏🏼 thanks for taking this over!

One small note one the `AUX` device class. This was poorly documented by me: 

AUX was just meant as a placeholder (because we don’t actually have a B1530A card to test it), which is why this wasn’t included in the enum. 

It should probably be something that makes more sense in this context like `AWG` if you actually want to use it.

astafan8:
Once I resolve the problem with the docs, I will merge. Enough of adding features and things within the scope of this PR.

StefanD986:
Haha no worries, I love to see the progress ;)

Op vr 7 jun. 2019 om 17:01 schreef Mikhail Astafev <notifications@github.com

qcodes/instrument_drivers/Keysight/keysightb1500/message_builder.py
```diff
@@ -0,0 +1,3886 @@
...
+        To make usage easier also take a look at the classes defined in
+        keysightb1500.constants which defines a lot of the integer constants
+        that the commands expect as arguments.
+        """
+        self._msg = CommandList()
+
+    @property
+    def message(self) -> str:
+        joined = str(self._msg)
+        if len(joined) > 250:
```
jenshnielsen: It seems like this should be an error

jenshnielsen:
Furthermore the tests should be moved to a subfolder of the tests folder

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500.py
```diff
@@ -0,0 +1,178 @@
...
+
+            self.add_module(name=module.short_name, module=module)
+
+    @staticmethod
+    def from_model_name(model: str, slot_nr: int, parent: 'KeysightB1500',
+                        name: Optional[str] = None) -> 'B1500Module':
+        """
+        Creates the correct instance type for instrument by model name.
+
+        :param model: Model name such as 'B1517A'
```
jenshnielsen: These type of parameter docs all needs to be rewritten to google style which qcodes uses everywhere https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

jenshnielsen:
I think the docs build failure is due to the Class being documented both at the Original location and one level up due to the `__all__` definition in the `__init__.py` file. This is ambiguous since sphinx does not know which version to link against

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500.py
astafan8: done

qcodes/instrument_drivers/Keysight/keysightb1500/message_builder.py
astafan8: done

astafan8:
@jenshnielsen could you have a final look (since you already had a look before)? and then i'll merge.

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        self.write("*RST")
+
+    def get_status(self) -> int:
+        return int(self.ask("*STB?"))
+
+    # TODO: Data Output parser: At least for Format FMT1,0 and maybe for a
+    # second (binary) format. 8 byte binary format would be nice because it
+    # comes with time stamp
+    # FMT1,0: ASCII (12 digits data with header) <CR/LF^EOI>
```
jenshnielsen: Is it ok to merge with this todo? It's not clear to me as someone not familiar with the instrument what limitations this gives.

astafan8: yes, that is OK. the improvements on the data format will come in the future PRs. for the use cases in the labs, this is enough so far.

I also don't like to leave TODOs in the code, but it's ok for now, because there are expected more PRs in the coming month or two which might clean most of those up.

StefanD986: I'll comment on this, since I wrote this `TODO`: 
The B1500 provides many (I think >16) output formats, both ASCII and binary. The output format only tells how information (values, status messages, etc) is encoded in each item in the response. Additionally, what items are included depends on the measurement mode, and with which instruments modules the mode was configured.

My original plan was to create a parser for at least two of the formats that brings the response into some structured form. But due to the variety of result formats this is not trivial, which is why I went for the quick solution of having a separate parser for the mode that we currently use (which at the moment I think is just the Highspeed Spot IV measurement and the Highspeed spot Capacitance measurement.

If we ever want to make use of the other measurement modes I think it is still a good idea to pick this TODO up.

astafan8: thanks, Stefan, for clarification!

PR #1574, opened 2019-05-22
Driver DynaCool: Add magnet setting with wait

ThorvaldLarsen:
Adding function to set magnetic field and wait until magnet is a setpoint.
This PR still needs to be tested.

@Dominik-Vogel


qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
```diff
@@ -90,9 +90,10 @@ def __init__(self, name: str,
                            get_cmd='TEMP?')
 
         self.add_parameter('field',
-                           label='Field strength',
+                           label='Field',
                            unit='A/m',
-                           get_cmd=self._present_field_getter)
+                           get_cmd=self._present_field_getter
```
Dominik-Vogel:
```suggestion
                           get_cmd=self._present_field_getter,
```

ThorvaldLarsen:
Maybe one can look at difference between present field and expected field rather than the magnet state for to determine if the sweep is done? This was actually my initial plan but magnet state seemed like at slightly easier way to implement the same thing.

Dominik-Vogel:
That seems like a good plan. Maybe we can combine and wait until the current field has changed and then wait until we are no longer in the ramping state. I fully agree, using the state seems like a cleaner way of doing things.

WilliamHPNielsen:
I propose to close this PR in favour of #1586 . @ThorvaldLarsen and @Dominik-Vogel, what do you think?

PR #1576, opened 2019-05-22
Keithley6500 trigger_timer validator to narrow

qSaevar:
- Keithley6500 trigger_timer validator to should allow 0

@astafan8

### DIFF
--- a/qcodes/instrument_drivers/tektronix/Keithley_6500.py
+++ b/qcodes/instrument_drivers/tektronix/Keithley_6500.py
@@ -167,6 +167,7 @@ class Keithley_6500(VisaInstrument):
                                             'bus': 'COMM',
                                             'external': 'EXT'})
 
+        # Control interval between scans, default is 0.
         self.add_parameter('trigger_timer',
                            get_parser=float,
                            get_cmd='ROUT:SCAN:INT?',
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/tektronix/Keithley_6500.py
```diff
@@ -171,7 +171,7 @@ def __init__(self, name, address, reset_device=False, **kwargs):
                            get_parser=float,
                            get_cmd='ROUT:SCAN:INT?',
                            set_cmd='ROUT:SCAN:INT {}',
-                           unit='s', vals=Numbers(min_value=0.001, max_value=999999.999))
```
astafan8: and what does the value of 0 mean in this case? trigger immediately? if so, i guess this is to be set using trigger_source parameter, no?

qSaevar: trigger immediately I think. The problem here is that the default value from the device is 0, so if you read the value from the device and try to reapply it, it fails.

astafan8: oh.... ok, could you add that clarification as a comment next to that parameter definition? and then we'll merge.

qSaevar: done

qcodes/instrument_drivers/tektronix/Keithley_6500.py
```diff
@@ -167,11 +167,12 @@ def __init__(self, name, address, reset_device=False, **kwargs):
                                             'bus': 'COMM',
                                             'external': 'EXT'})
 
+        # Control interval between scans, default is 0.
```
astafan8: I meant to add this part :) :
```suggestion
        # Control interval between scans; the default value from the instrument is 0,
        # hence 0 is included in the validator's range of this parameter.
```

PR #1578, opened 2019-05-23
Deprecation of Qdac in favor of QDac_channels

GateBuilder:
Added deprecation call for QDac in favor of QDac_channels to prevent confusion.

@Dominik-Vogel


qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -30,7 +31,8 @@ class QDac(VisaInstrument):
 
     # set nonzero value (seconds) to accept older status when reading settings
     max_status_age = 1
-
+    # this driver will be deprecated in favor of QDac_channels.py
+    @deprecate
```
astafan8: allow me to suggest to have a dedicated `deprecate` decorator for `__init__` - the warning message could then say that "some CLASS is deprecated, and other class should be used" instead of say "function SMTH is deprecated...", see here:
https://github.com/QCoDeS/Qcodes/blob/0f9176495d3bd02e6a05a221ef0b9635810c0f02/qcodes/utils/deprecate.py#L10-L10

astafan8: please use `reason` and `alternative` arguments of this decorator to provide our users with reason and alternative. after that, the comment `this driver will be deprecated in favor of QDac_channels.py` will be redundant. (Also note that it is extremely unlikely that the users will read that comment.)

qcodes/instrument_drivers/QDev/QDac.py
```diff
@@ -30,8 +31,9 @@ class QDac(VisaInstrument):
 
     # set nonzero value (seconds) to accept older status when reading settings
     max_status_age = 1
-
+    
     def __init__(self, name, address, num_chans=48, update_currents=True):
+        warnings.warn("The QDac Class is deprecated; use QDacChannel instead.", DeprecationWarning, 2)
```
WilliamHPNielsen: I don't mean to nitpick, but the `QDac` class from `QDac.py` is being deprecated in favour of the `QDac` class from `QDac_channels.py`. The module `QDac_channels.py` also contains a class called `QDacChannel`, but that is not an instrument driver; the driver is called `QDac`.

jenshnielsen:
@GateBuilder Note that it looks like there are differences between when DeprecationWarnings are shown in python 3.6 and 3.7 Compare https://docs.python.org/3.6/library/warnings.html#warning-categories to https://docs.python.org/3.7/library/warnings.html#warning-categories

I suggest that we extend the decorator to also support classes. This could simply be done by taking an optional type argument. 

@astafan8 I suggest that we create a qcodes deprecation warning that inherit from UserWarning or some other warnings categorie that is not DeprecationWarning. Similar to how it is done here https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/cbook/deprecation.py in matplotlib

astafan8:
@GateBuilder I'm not sure I understand why the reality is such that we should remember to "specify the stacklevel" to some value when issuing a warning.

@jenshnielsen I like your suggestion to create `QcodesDeprecationWarning` from `UserWarning` as done in matplotlib. This seems to me a cleaner solution than taking care of the stacklevel.

I am also for extending the `deprecate` decorator (we could in principle just copy-paste the code from matplotlib, adjust it to our needs, and add minor tests for it). But i'd leave this up to the author because i don't want this PR to explode. Ideally, if to be done, the extension of the `deprecate` decorator should be a PR of its own :)

qcodes/__init__.py
```diff
@@ -73,7 +72,7 @@
 from qcodes.instrument.sweep_values import SweepFixedValues, SweepValues
 
 from qcodes.utils import validators
-from qcodes.utils.zmq_helpers import Publisher
```
jenshnielsen: These changes should not be here.

Dominik-Vogel: I was just commenting on that. Give me a minute :-)

jenshnielsen: Sorry no rush :)

jenshnielsen:
I think this looks good. I suggest we merge this via a squash merge as the history has become somewhat confusing on this branch.

Dominik-Vogel:
Sorry for the commit mess, I forgot that I did not have a clean master... I added reverting commits and will squash merge


@GateBuilder

jenshnielsen:
I always try to remember merging and rebaseing against origin/master and not the local master to prevent this

Dominik-Vogel:
I pulled from upstream, but I think the problem was that I created the branch via `git checkout -b` while I was on my dirty local master.

PR #1579, opened 2019-05-23
Driver for Tektronix Oscilloscopes: MSO/DPO5000/B, DPO7000/C, DPO70000/B/C/D/DX, DSA70000/B/C/D, and MSO70000/C/DX 

sohailc:
This PR adds a driver for the following oscilloscopes: Tektronix MSO/DPO5000/B, DPO7000/C, DPO70000/B/C/D/DX, DSA70000/B/C/D, and MSO70000/C/DX Series Digital Oscilloscopes.


qcodes/instrument_drivers/tektronix/DPO7200xx.py
```diff
@@ -0,0 +1,605 @@
...
+        super().__init__(parent, name)
+
+        self.add_parameter(
+            "raw_data_offset",
+            get_cmd="WFMOutPRE:YOFF?",
+            get_parser=float,
+            docstring="""
+            Raw acquisition values range from min to max. 
+            For instance, for unsigned binary values of one 
+            byte, min=0 and max=255. The data offset specifies
```
astafan8: does this mean that setting the min/max via a validator is left out of scope of this PR?

qcodes/instrument_drivers/tektronix/DPO7200xx.py
```diff
@@ -0,0 +1,605 @@
...
+            vals=Arrays(shape=(self.get_trace_length,)),
+            unit=waveform_format.x_unit()
+        )
+
+        self.add_parameter(
+            "trace",
+            label="Voltage",
+            get_cmd=self._get_trace_data,
+            vals=Arrays(shape=(self.get_trace_length,)),
+            unit=waveform_format.y_unit(),
```
astafan8: perhaps a stupid question - does the unit have to be dynamic? so that if it changes, the unit of this parameter gets updated?

qcodes/instrument_drivers/tektronix/DPO7200xx.py
```diff
@@ -0,0 +1,605 @@
...
+
+class TektronixDPOMeasurement(InstrumentChannel):
+    """
+    The measurement submodule
+    """
+    # It was found by trial and error that adjusting
+    # the measurement type and source takes some time
+    # to reflect properly on the value of the
+    # measurement. Wait a minimum of ...
+    minimum_adjustment_time = 0.1
```
astafan8: should this be private ? :) and explained in the example notebook as well?

qcodes/instrument_drivers/tektronix/DPO7200xx.py
```diff
@@ -0,0 +1,605 @@
...
+    def __init__(
+            self,
+            parent: Instrument,
+            name: str,
+            measurement_number: int
+    ) -> None:
+
+        super().__init__(parent, name)
+        self._measurement_number = measurement_number
+        self._adjustment_time = time.time()
```
astafan8: shall perf_counter be used for precision?

qcodes/instrument_drivers/tektronix/DPO7200xx.py
```diff
@@ -0,0 +1,605 @@
...
+
+
+class TektronixDPO7000xx(VisaInstrument):
+    """
+    QCoDeS driver for the MSO/DPO5000/B, DPO7000/C,
+    DPO70000/B/C/D/DX/SX, DSA70000/B/C/D, and
+    MSO70000/C/DX Series Digital Oscilloscopes
+    """
+    channel_count = 4
+    measurement_count = 8
```
astafan8: may i suggest to rename these to `number_of_*`?

sohailc:
@astafan8 Thanks so much for your comments. This was really helpful and I am adjusting my PR according to your comments. My latest commit is still work in progress. Let me send you a message when I think we should have a look at my PR again. 

Thanks again!

qcodes/instrument_drivers/tektronix/DPO7200xx.py
sohailc: No, its not out of the scope of this PR. The value of raw data offset is implicitly defined by several parameters, including the data encoding used (see the `encoding` parameter on TektronixDPOData), which in turn is determined by the data format (e.g. signed integer, floating points) and bytes per sample. See also the `TektronixDPOWaveformFormat` module. 

There is no way to directly set the value of `raw_data_offset` though.

qcodes/instrument_drivers/tektronix/DPO7200xx.py
sohailc: not a stupid question at all! I modified my original PR so this is not relevant anymore, but the units on the channels will always be the same and there is no way for the user to change this. So this can be queried once.

qcodes/instrument_drivers/tektronix/DPO7200xx.py
sohailc: I made this private now

qcodes/instrument_drivers/tektronix/DPO7200xx.py
sohailc: Thanks for letting me know about perf_counter! I did not know about this method. I have used it now in my new PR

qcodes/instrument_drivers/tektronix/DPO7200xx.py
sohailc: Ok

sohailc:
@astafan8 Can we have another look at this? I did my best to apply the modifications you suggested. 

Thanks, 

Sohail

qcodes/instrument_drivers/tektronix/DPO7200xx.py
astafan8: ok

qcodes/instrument_drivers/tektronix/DPO7200xx.py
```diff
@@ -0,0 +1,704 @@
...
+
+        super().__init__(parent, name)
+
+        self.add_parameter(
+            "mode",
+            get_cmd="HORizontal:MODE?",
+            set_cmd="HORizontal:MODE {}",
+            vals=Enum("auto", "constant", "manual"),
+            get_parser=str.lower,
+            docstring="""
```
astafan8: do these long docstrings render nicely in jupyter, meaning with correct spacing and indentation? if not, then i'd suggest to wrap them in a call to [`textwrap.dedent`](https://docs.python.org/3.6/library/textwrap.html#textwrap.dedent) (i think it's done that way in some other driver already).

sohailc:
@astafan8, *please do not merge this yet*. There are some missing features I forgot to add.

astafan8:
@sohailc ok, fine.

But why does that feature have to be added in __THIS__ PR? Can't we merge this first, and then the submodule is added in another PR?

sohailc:
I thought it handier to add all needed features in this PR. Anyway, I have added the submodule now so if it passes the tests and you approve of the code, can you please merge it in.

PR #1582, opened 2019-05-27
Add set_dac_unit method to D5a instrument driver

qSaevar:
- Add `set_dac_unit` method to D5a instrument for chaning the units and gain between volts and millivolts.

@astafan8


qcodes/instrument_drivers/QuTech/D5a.py
```diff
@@ -98,6 +103,11 @@ def __init__(self, name, spi_rack, module, inter_delay=0.1, dac_step=10e-3,
                                vals=Enum(*self._span_set_map.keys()),
                                docstring='Change the output span of the DAC. This command also updates the validator.')
 
+    def _set_dac_unit(self, unit: str) -> None:
```
astafan8: and for consistency `self._mV` should also be changed.

qSaevar: I removed `self._mV` as it is not used.

qSaevar:
@astafan8 as there is no real need for `.get()`-ing the `set_dac_unit` parameter I converted it to a method as you suggested.

qSaevar:
No, I don't mind. Thanks

PR #1584, opened 2019-05-29
Fix / snapshot_base params_to_skip_update

WilliamHPNielsen:
We found out that there's currently a bug in `snapshot_base`: if the snapshot encounters a parameter that should be skipped (as given by the `params_to_skip_update` argument), it will skip the update of all subsequent parameters.

- Add a test for this behaviour
- Fix the skipping logic
- Add `Optional` to the type annotations for `params_to_skip_update`

@QCoDeS/core


qcodes/tests/test_snapshot.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    params = ['v1', 'v2', 'v3', 'v4']
+
+
+    inst = SnapShotTestInstrument('inst',
+                                  params=params,
+                                  params_to_skip=[])
+
+    request.addfinalizer(inst.close)
```
astafan8: i'd move these two lines inside the loop instead of
```
inst._params_to_skip = param_to_skip
inst.reset_counter()
```
in order to reduce unnecessary state management.

qcodes/tests/test_snapshot.py
```diff
@@ -0,0 +1,39 @@
...
+Test module for snapshots of instruments and parameters
+"""
+
+from qcodes.tests.instrument_mocks import SnapShotTestInstrument
+
+
+def test_snapshot_skip_params_update(request):
+    """
+    Test that params_to_skip_update works as expected, in particular that only
+    the parameters given by that variable are skipped
```
astafan8: i find it a bit confusing that `params_to_skip_update` argument is not explicitly used in this test. i expected to see many `inst.snapshot(...)` with different combinations of `update` and `params_to_skip_update` values. can the test be improved so that the tested cases are more explicit and not hidden inside `SnapShotTestInstrument`?

qcodes/utils/metadata.py
```diff
@@ -55,8 +58,8 @@ def snapshot(self, update=False):
 
         return snap
 
-    def snapshot_base(self, update: bool=False,
-                      params_to_skip_update: Sequence[str]=None):
+    def snapshot_base(self, update: bool = False,
```
astafan8: it's a bit weird that already Metadatable knows about parameters (due to `params_to_skip_update`) while the parameters appear only in the instrument.... but that's not to be dealt with in this PR, of course, just nagging.

qcodes/instrument/channel.py
```diff
@@ -383,7 +383,9 @@ def lock(self) -> None:
         self._channels = tuple(self._channels)
         self._locked = True
 
-    def snapshot_base(self, update: bool=False, params_to_skip_update: Optional[Sequence[str]]=None):
+    def snapshot_base(self, update: bool = True,
+                      params_to_skip_update: Optional[Sequence[str]] = None
```
astafan8: this might be irrelevant for this PR, but i don't get why is `params_to_skip_update` NOT passed to the `chan.snapshot(update=update)` calls inside this method? it just seems to me that this feature is not well thought of and perhaps not well implemented. again, this is just nagging.

PR #1586, opened 2019-06-03
Driver / dynacool add blocking and non-blocking ramp

WilliamHPNielsen:
This PR fixes the same problem as #1574 does, but it makes the DynaCool driver more similar to the MercuryiPS driver.

- Add a `field_target` parameter
- Add a separate `ramp` method to ramp the magnetic field, either in a blocking or non-blocking way
- Fix some unit confusion; we want to work in T

It would be awesome if @ThorvaldLarsen and @tomasstankevic could have a look at this PR.

@QCoDeS/core


qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
```diff
@@ -194,11 +212,89 @@ def get_idn(self) -> Dict[str, Optional[str]]:
...
+        done in two steps, since users have reported that the magnet state does
+        not immediately change to 'ramping' when asked to ramp.
+        """
+
+        target_in_T = target_in_oe*1e-4
+
+        start_field = self.field_measured()
+        ramp_range = np.abs(start_field - target_in_T)
+
+        # TODO (WilliamHPNielsen): do we need some non-zero tolerance here?
```
astafan8: i'd say, yes) some numpy.allclose -like thingy

qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
```diff
@@ -194,11 +212,89 @@ def get_idn(self) -> Dict[str, Optional[str]]:
...
+        if ramp_range == 0:
+            return
+
+        self._field_setter(param='field_setpoint', value=target_in_oe)
+
+        # step 1: wait for the magnet to actually start ramping
+        # NB: depending on the `field_approach`, we may reach the target
+        # several times before the ramp is over (oscillations around target)
+        while np.abs(self.field_measured() - start_field) < ramp_range:
+            sleep(0.1)
```
astafan8: please make it a private attribute of the class to allow adjustments.

qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
```diff
@@ -194,11 +212,89 @@ def get_idn(self) -> Dict[str, Optional[str]]:
...
+        start_field = self.field_measured()
+        ramp_range = np.abs(start_field - target_in_T)
+
+        # TODO (WilliamHPNielsen): do we need some non-zero tolerance here?
+        if ramp_range == 0:
+            return
+
+        self._field_setter(param='field_setpoint', value=target_in_oe)
+
+        # step 1: wait for the magnet to actually start ramping
```
astafan8: shouldn't the lines above this one be outside of this method? because the seem to apply to the "non-blocking ramp" as well.

qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
```diff
@@ -194,11 +212,89 @@ def get_idn(self) -> Dict[str, Optional[str]]:
...
+        # NB: depending on the `field_approach`, we may reach the target
+        # several times before the ramp is over (oscillations around target)
+        while np.abs(self.field_measured() - start_field) < ramp_range:
+            sleep(0.1)
+
+        # step 2: wait for the magnet to report that is has reached the
+        # setpoint
+
+        while self.magnet_state() == 'ramping':
+            sleep(0.1)
```
astafan8: please make it a private attribute of the class to allow adjustments.

qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
```diff
@@ -91,30 +93,43 @@ def __init__(self, name: str,
...
+                           label='Field target',
+                           unit='T',
+                           get_cmd=None,
+                           set_cmd=None,
+                           vals=vals.Numbers(-14, 14))
+
         self.add_parameter('field_setpoint',
                            label='Field setpoint',
-                           unit='A/m',
+                           unit='T',
```
astafan8: that's backwards incompatible, but i guess that's the point anyway, right?

astafan8: actually since the units are backwards INcompatible, perhaps, we can safely get rid of the "deprecated" parameters now? or not?

qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
WilliamHPNielsen: `np.allclose` is surely better than `==`. But we still need some input from scientists of what the noise is.

qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
WilliamHPNielsen: That's a good idea.

qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
WilliamHPNielsen: This is actually a bugfix, see the original commit: https://github.com/QCoDeS/Qcodes/commit/c9a8bf99d18abc12bb63dab1548973ee990739d7

We've been working in Teslas all along, we just forgot to update the units. So I would not consider this to be backwards incompatible.

WilliamHPNielsen: I'd say we keep them for a little while.

qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
WilliamHPNielsen: Sure.

qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
WilliamHPNielsen: Sure.

WilliamHPNielsen:
@astafan8, this should address all your concerns. Let me know if I missed something.

qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
astafan8: ah, i see, great ))

astafan8:
@WilliamHPNielsen yes, thank you :)

ThorvaldLarsen:
@WilliamHPNielsen as I understand it this does not solve my problem.
I want to do:
`do1d(dyna.blocking_ramp_to_target,0,1,101,0,some_measurement_param)`
With this PR there is no parameter that will set the target field _and_ perform blocking ramp to target. I realize this is not an optimal way to perform a magnet sweep but this should still be possible.
This is the same as the `f'{coord}_ramp'` parameters in the mercury driver which will set the field and perform a blocking ramp.

qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
tomasstankevic: Why do we need to check if it's reached the target value? Is it not enough to read the state from the magnet that it's holding? 
There is certainly some tolerance, but I can't say how much, maybe we can come up with what tolerance is needed? @ThorvaldLarsen  For example when using approach mode "No o-shoot" the field never reaches exactly zero, it has some 0.16 Oe remnant magnetization, however the magnet state says "holding".

WilliamHPNielsen:
@ThorvaldLarsen, fair enough. How about now?

ThorvaldLarsen:
Yes, I think that should do it. Thanks!

WilliamHPNielsen:
@ThorvaldLarsen and @tomasstankevic, we should not merge this before (at least) one of you has had the chance to verify that the new parameters work on an actual instrument. At your earliest convenience, please do that (or get someone else to do it) and ping back the outcome here. Thanks!

qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
```diff
@@ -91,30 +98,50 @@ def __init__(self, name: str,
...
+                           label='Field',
+                           unit='T',
                            get_cmd=self._present_field_getter)
 
+        self.add_parameter('field_target',
+                           label='Field target',
+                           unit='T',
+                           get_cmd=None,
+                           set_cmd=None,
+                           vals=vals.Numbers(-14, 14))
```
jenshnielsen: I suggest adding docstrings to these 4 parameters clairyfing their difference

astafan8:
@ThorvaldLarsen and @tomasstankevic, did you happen to test the new parameters?

WilliamHPNielsen:
Merging (after CI) as this driver version has been in use at MQML for weeks now with no problems.

PR #1587, opened 2019-06-03
Fix/weird travis test fail

WilliamHPNielsen:
For some combinations of `pytest` version `4.6.1` and [we don't quite know what] version `?`, tests that use `assert all(...) is True` or `assert not all(...)` fail.

Since we can't really figure out which package versions to avoid, let's avoid using that syntax instead.

- Reword two tests

@astafan8


WilliamHPNielsen:
Just for completeness, I think it's this issue that we saw: https://github.com/pytest-dev/pytest/issues/5372

So the problem would also go away in `pytest` version `4.6.2`, it seems.

astafan8:
thanks for the completeness!

PR #1591, opened 2019-06-06
Bump sphinx to 2.1.0

astafan8:
actually, there is hope that this will help with this error https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=1996 , as i've seen a possibly related [bug has been fixed](https://www.sphinx-doc.org/en/master/changes.html#release-2-1-0-released-jun-02-2019) `#6379: py domain: Module index (py-modindex.html) has duplicate titles`.

### DIFF
--- a/docs_requirements.txt
+++ b/docs_requirements.txt
@@ -1,4 +1,4 @@
-sphinx==2.0.1
+sphinx==2.1.0
 sphinx_rtd_theme
 sphinx-jsonschema
 nbsphinx


jenshnielsen:
It looks like the auto documented api docs are not valid rst. I suggest that we postpone this until 2.1.1 is released. This could be due to https://github.com/sphinx-doc/sphinx/issues/6452

PR #1592, opened 2019-06-07
Plotting / support complex numbers in plot_by_id

WilliamHPNielsen:
Support the plotting of complex numbers in `plot_by_id`. Still WIP.

- Add a preparser that splits complex-valued parameters into two real-valued ones

Still pending:

- [x] Actually use the preparser in `plot_by_id`
- [x] Update the documentation to show how to plot complex numbers
- [x] Change `phase_and_mag` into `mag_and_phase`

@QCoDeS/core


qcodes/dataset/plotting.py
```diff
@@ -269,6 +272,110 @@ def plot_by_id(run_id: int,
...
+        conversion: the conversion method, either "real_and_imag" or
+            "phase_and_mag"
+        degress: whether to return the phase in degrees. The default is to
+            return the phase in radians
+    """
+
+    if conversion not in ['real_and_imag', 'phase_and_mag']:
+        raise ValueError(f'Invalid conversion given. Received {conversion}, '
+                         'but can only accept "real_and_imag" or '
+                         '"phase_and_mag".')
```
astafan8: awesome!

`mag_and_phase`? :)

WilliamHPNielsen: Hmm, that's probably true. `(r, theta)` sounds natural.

WilliamHPNielsen:
@ThorvaldLarsen, you were one of the first people to request this feature. I'd be good to know if you think that this PR does the job, or whether we overlooked something.

qcodes/dataset/plotting.py
```diff
@@ -269,6 +288,110 @@ def plot_by_id(run_id: int,
...
+    # we build a new NamedData object from the given `alldata` input.
+    # Note that the length of `newdata` will be larger than that of `alldata`
+    # in the case of complex top-level parameters. This is the reason why we
+    # use two variables below,
+
+    for group in alldata:
+        new_group = []
+        new_groups: NamedData = [[], []]
+        for index, parameter in enumerate(group):
+            data: np.ndarray = parameter['data']
```
astafan8: `data` seems to be unused, right?
```suggestion
```

qcodes/dataset/plotting.py
```diff
@@ -269,6 +288,110 @@ def plot_by_id(run_id: int,
...
+    new_parameters: List[Dict[str, Union[str, np.ndarray]]] = []
+
+    for n in range(2):
+        d: Dict[str, Union[str, np.ndarray]]
+        d = {'name': new_names[n], 'label': new_labels[n],
+             'unit': new_units[n], 'data': new_data[n]}
+        new_parameters.append(d)
+
+    return tuple(new_parameters)  # type: ignore
+
```
astafan8: may i suggest a subjectively simpler implementation of this logic?:
```
new_parameters = (
    {'name': name, 'label': label,
     'unit': unit, 'data': data}
    for name, label, unit, data in zip(
        new_names, new_labels, new_units, new_data)
)

return tuple(new_parameters)
```

qcodes/dataset/plotting.py
```diff
@@ -269,6 +288,110 @@ def plot_by_id(run_id: int,
...
+    if conversion not in ['real_and_imag', 'mag_and_phase']:
+        raise ValueError(f'Invalid conversion given. Received {conversion}, '
+                         'but can only accept "real_and_imag" or '
+                         '"mag_and_phase".')
+
+    newdata = []
+
+    # we build a new NamedData object from the given `alldata` input.
+    # Note that the length of `newdata` will be larger than that of `alldata`
+    # in the case of complex top-level parameters. This is the reason why we
```
astafan8: Could you add a clarification phrase after `... in the case of complex top-level parameters.` along the lines of
```
Note that the length of `newdata` will be larger than that of `alldata` 
in the case of complex top-level parameters due to the fact that in this 
function a SINGLE complex-valued top-level parameter gets split into 
TWO real-valued top-level parameters.
```
?

This allows to understand the need for both `new_group` and `new_groups` before reaching the end of the function.

qcodes/tests/dataset/test_plotting.py
```diff
@@ -148,3 +149,97 @@ def test_appropriate_kwargs():
     with _appropriate_kwargs('2D_point', False, **{}) as ap_kwargs:
         assert len(ap_kwargs) == 1
         assert ap_kwargs['cmap'] == qc.config.plotting.default_color_map
+
+
+def test__complex_to_real_preparser():
+
+    # FIRST CASE: one real setpoint, one complex value
```
astafan8: Shall the `test__complex_to_real_preparser` be split into at least two independent ones (ideally into 4 parts actually)? I'd also suggest for this test to go into its own file.

ThorvaldLarsen:
@WilliamHPNielsen this looks great to me! Thanks for getting it done.

qcodes/dataset/plotting.py
WilliamHPNielsen: I think we use it to check whether the data values are complex (next line).

qcodes/dataset/plotting.py
WilliamHPNielsen: I was hoping that all the extra type annotations would somehow be helpful. I guess they're not?

qcodes/dataset/plotting.py
WilliamHPNielsen: More explanation is always better.

qcodes/dataset/plotting.py
astafan8: they are but the `for n in range(2):` part kind of kills it, at least for me. Hence, the suggestion.

qcodes/dataset/plotting.py
astafan8: sono idioto :)

qcodes/tests/dataset/test_plotting.py
WilliamHPNielsen: I agree with splitting it into two tests. I don't think it should go into a separate file unless the preparser also leaves the plotting module (into `plotting_helpers.py` or smth)

qcodes/dataset/plotting.py
```diff
@@ -269,6 +288,120 @@ def plot_by_id(run_id: int,
...
+            new_names, new_labels, new_units, new_data))
+
+    # The reason we ignore the type in the return is that I cannot figure
+    # out how to get mypy to correctly infer the type of iterated values
+    # (the name, label, unit, and data above)
+
+    return new_parameters  # type: ignore
+
+
+def _make_named_data_dict(
```
astafan8: unused?

jenshnielsen:
The notebook should use 

load_or_create_experiment and explicitly pass the experiment to the measurement
plot_dataset()

qcodes/dataset/plotting.py
astafan8: removed.

astafan8:
> The notebook should use load_or_create_experiment and explicitly pass the experiment to the measurement

Done. plot_dataset() change I would defer to another PR. so that we change plot_by_id for plot_dataset in the whole qcodes at once.

PR #1597, opened 2019-06-13
ZI UHF-LI: fix `scope_mode` and `scope_average_weight` parameters

qSaevar:
- Ignore  `sweeper_sweeptime` in `snapshot(update=True)` if `_sweeper_signals` is an empty list
- Getter for `scope_mode` and `scope_average_weight` should work

When taking a snapshot of the ZIUHFLI driver with update=True the following warnings are generated:
```
[ziuhfli(ZIUHFLI)] Snapshot: Could not update parameter: sweeper_sweeptime
[ziuhfli(ZIUHFLI)] Snapshot: Could not update parameter: scope_mode
[ziuhfli(ZIUHFLI)] Snapshot: Could not update parameter: scope_average_weight
``` 
This should no longer be the case.

@WilliamHPNielsen

### DIFF
--- a/qcodes/instrument_drivers/ZI/ZIUHFLI.py
+++ b/qcodes/instrument_drivers/ZI/ZIUHFLI.py
@@ -5,7 +5,7 @@ import numpy as np
 from functools import partial
 from math import sqrt
 
-from typing import Callable, List, Union, cast, Optional
+from typing import Callable, List, Union, cast, Optional, Sequence, Dict
 
 from qcodes.utils.helpers import create_on_off_val_mapping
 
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -2265,7 +2278,7 @@ def getduration():
             returndict =  self.scope.get(querystr)
             # The dict may have different 'depths' depending on the parameter.
             # The depth is encoded in the setting string (number of '/')
-            keys = setting.split('/')[1:]
```
WilliamHPNielsen: Is this change the change responsible for making the getters for `scope_mode` and `scope_average_weight` work?

qSaevar: yes

qSaevar: as the `setting` for `scope_mode` is just `mode` and for `scope_average_weight` is `averager/weight` then omitting the first part before a `/` of the setting leaves an empty list for the `mode` and a key error for the `weight`

WilliamHPNielsen: I see. And the `_scope_getter` is only used for the `scope_mode`, `scope_average_weight`, and `scope_duration` parameters, the latter of which is handled explicitly by the `_scope_getter` function.

The changes look correct. I just wonder why the `[1:]` was put in to begin with. Perhaps the setting nesting level changed in some firmware upgrade? If you can somehow check that, that would be fantastic. But since these changes make your version work, let's get them in.

qSaevar: I went over the commit history of the original PR, and it seems that in commit `1066df6b01d6016da8ff7f34ccdfe201aea0fcfd` when the special case for `duration` was added this bug was introduced. Before that commit, `rawvalue = returndict` was outside and after the while loop `while keys != []` and then it should have worked.

WilliamHPNielsen: Wow, good job! Thank you.

PR #1598, opened 2019-06-13
Close instruments in tests.

jenshnielsen:
This avoids a bunch of warnings from pytest. 

Furthermore fix a test that depended on the state of the simulated instrument when being run.
This is caused by pyvisa-sim remembering state even when closing instruments

### DIFF
--- a/qcodes/tests/drivers/test_keithley_s46.py
+++ b/qcodes/tests/drivers/test_keithley_s46.py
@@ -33,8 +33,10 @@ def s46_six():
     """
     driver = S46('s46_six', address='GPIB::2::INSTR', visalib=visalib)
 
-    yield driver
-    driver.close()
+    try:
+        yield driver
+    finally:
+        driver.close()
 
 
 @pytest.fixture(scope='module')
@@ -44,8 +46,10 @@ def s46_four():
     """
     driver = S46('s46_four', address='GPIB::3::INSTR', visalib=visalib)
 
-    yield driver
-    driver.close()
+    try:
+        yield driver
+    finally:
+        driver.close()
 
 
 def test_runtime_error_on_bad_init():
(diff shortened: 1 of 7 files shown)


WilliamHPNielsen:
This looks good, I like it.

We should probably consider whether we want to modify `pyvisa-sim` to be less stateful. Or perhaps we should just have a test capturing the fact that state persists through the `inst.close` method?

WilliamHPNielsen:
This looks good, I like it.

We should probably consider whether we want to modify `pyvisa-sim` to be less stateful. Or perhaps we should just have a test capturing the fact that state persists through the `inst.close` method?

qcodes/tests/test_visa.py
```diff
@@ -134,28 +134,32 @@ def open_resource(self, address):
...
         self.assertEqual(address_opened[0], 'ASRL3')
+        inst.close()
 
         # this one doesn't
-        MockBackendVisaInstrument('name4', address='ASRL4', visalib='@py')
+        inst = MockBackendVisaInstrument('name4', address='ASRL4', visalib='@py')
         self.assertEqual(rm_mock.call_count, 4)
         self.assertEqual(rm_mock.call_args, (('@py',),))
         self.assertEqual(address_opened[0], 'ASRL4')
+        inst.close()
```
astafan8: this test just calls for being split into multiple.

qcodes/tests/drivers/test_keithley_s46.py
```diff
@@ -66,8 +70,9 @@ def test_query_close_once_at_init(caplog):
     Test that, during initialisation, we query the closed channels only once
     """
     with caplog.at_level(logging.DEBUG):
-        S46('s46_test_query_once', address='GPIB::2::INSTR', visalib=visalib)
+        inst = S46('s46_test_query_once', address='GPIB::2::INSTR', visalib=visalib)
         assert caplog.text.count(":CLOS?") == 1
+        inst.close()
```
astafan8: these calls to `close` (throughout the PR) should ideally should be taken care of by `request.addfinalizer(..)`. If we have more unforeseen issues in the future, i wouldn't want the absence of `request.addfinalizer(..)` be part of that issue.

jenshnielsen: This test does not use a fixture. The tests that does used try/yield finally which will be equivalently good. The fix of this test and others that do not use fixtures is out of scope for this pr. To be solved probably this will need to resolve the general issue with exceptions raised in instrument construction leaving instruments in an incomplete state

astafan8: ok, ok) for later)

PR #1599, opened 2019-06-13
BugFix: _update_units in AMI430 

ThorvaldLarsen:
Dave found bug with the AMI driver when trying to change the units using `field_units` parameter.

So, in this PR:
* This is fixed by using inverted_val_mapping instead of val_mapping in the logic of the parameter set function.
* Also added tests for this.
* Fix AMI430 to update coil constant parameter when field units changed. The call to method `self._update_coil_constant()` has been changed to call on parameter `self.coil_constant()` so that the possible new value of the coil constant from the instrument is being stored in the parameter (can be retrieved using `self.coil_constant.get_latest()` afterwards).


astafan8:
@ThorvaldLarsen Great that this was spotted, and thanks for the PR!

However, i'm not sure I understand what steps need to be taken in order to come to this bug. Would you be comfortable adding these steps in a form of a test in `qcodes/tests/drivers/test_ami430.py` file? ([this test case](https://github.com/QCoDeS/Qcodes/blob/master/qcodes/tests/drivers/test_ami430.py#L435) might serve as a template).

ThorvaldLarsen:
@astafan8 the bug appears if you try to use the parameter. If you set the units the code will fail in 
 two lines below in `ramp_rate_units = AMI430._SHORT_UNITS[ramp_rate_units]` as `ramp_rate_units` is not a field in the dictionary with the `val_mapping` done incorrectly.
Clearly, nobody has ever used this parameter before.

astafan8:
aa, got it))

> Clearly, nobody has ever used this parameter before.

so we are going to fix this by adding some tests...

astafan8:
@ThorvaldLarsen I added some tests, and fixed some minor bug on the way as well. Will you have time to re-test the PR again, just in case? once you confirm, i'll merge.

ThorvaldLarsen:
Great! Thanks for PR through fast.
Will try to get it tested tomorrow.

astafan8:
@ThorvaldLarsen did you happen to test this?

astafan8:
@ThorvaldLarsen has this been tested in the end on a real magnet?

PR #1605, opened 2019-06-20
[WIP] Persist run id to a new column

jenshnielsen:
Will ultimately make it easier to reference runs that have been moved from one db to another. 

In the long run we should focus on improving the experiment interface and make it possible to look up data by result counter. 

Stuff to do:
- [x] Provide api to load from captured_run_id and disambiguate as needed
- [x] test the upgrader
- [ ] captured_run_id is kind of a bad name. Any better suggestions?


qcodes/dataset/data_set.py
```diff
@@ -1111,3 +1180,28 @@ def new_data_set(name, exp_id: Optional[int] = None,
...
+
+    Args:
+        guids: Sequence of one or more guids
+        conn: A ConnectionPlus object with a connection to the database.
+
+    Returns: ASCII art table of information about the supplied guids.
+    """
+    from tabulate import tabulate
+    headers = ["captured_run_id", "experiment_name", "sample_name",
+               "sample_id", "location", "work_station"]
```
WilliamHPNielsen: Should `captured_counter` not be one of the headers?

qcodes/tests/dataset/test_database_extract_runs.py
```diff
@@ -411,17 +419,190 @@ def test_load_by_X_functions(two_empty_temp_db_connections,
...
+                              name='exp2',
+                              sample_name='no_sample')
+
+    source_2_datasets = [DataSet(conn=source_conn_2,
+                                 exp_id=source_2_exp.exp_id) for i in range(10)]
+
+    source_all_datasets = source_1_datasets + source_2_datasets
+
+    shuffled_datasets = source_all_datasets.copy()
+    random.shuffle(shuffled_datasets)
```
WilliamHPNielsen: I have a weak opinion that we should avoid randomness inside tests. I can take on a (probably cumbersome) rewrite of this logic with itertools and/or hypothesis.

jenshnielsen: Sure if you like. I could not be bothered in this case

qcodes/dataset/sqlite/db_upgrades/__init__.py
```diff
@@ -254,8 +254,8 @@ def perform_db_upgrade_6_to_7(conn: ConnectionPlus) -> None:
     """
     Perform the upgrade from version 6 to version 7
 
-    Add a captured_runid and captured_counter column to the runs table and
-    assign the value from the run_id and counter to these columns.
+    Add a captured_run_id and captured_counter column to the runs table and
+    assign the value from the run_id and result_counter to these columns.
```
jenshnielsen: This should indeed really be called `result_counter` but its called counter internally :( not sure what is best

qcodes/dataset/data_set.py
jenshnielsen: Yes that is probably a good idea

jenshnielsen:
replaced by #1637

PR #1611, opened 2019-06-21
Add option in get_parameter_data() to only return values of the top level parameter

### LINKED ISSUE #1601: [feature request] get_parameter_data option to not return setpoints
In the case of a `Multiparameter` with many internal setpoints that stay constant while an external setpoint is swept, the output of `Dataset.get_parameter_data` can get unwieldy. A useful option would be to only return setpoints once, then something similar to `Dataset.get_data`.

```
qc.new_experiment(name='get_data_issue', sample_name='none')

class MockMultiParam(MultiParameter):
    def __init__(self):
        super().__init__("multiparam", names=('A','B'), shapes=((1e6,),)*2, setpoints = ((range(1e6),),)*2, setpoint_names = (('t',),)*2 )
        self.i = 0
        
    def get_raw(self):
        val = (np.arange(1e6) + self.i*10*1e6,)*2
        self.i +=1
        return val

mock_param = MockMultiParam()
t2_param = Parameter('t2', set_cmd=None, get_cmd=None)

meas = Measurement()
meas.register_parameter(t2_param)
meas.register_parameter(mock_param, paramtype='array', setpoints=(external_param,))

with meas.run() as datasaver:  
    for t2 in range(100):
        datasaver.add_result((t2_param, t2), (mock_param, mock_param()))
        run_id = datasaver.run_id

print(ds.get_data('B'))
print(ds.get_parameter_data('B'))
```

thibaudruelle:
Fixes #1601.

- add `return_setpoints` option (defaulting to `True`) to `get_parameter_data()` in both `dataset.Dataset` and `dataset.sqlite.queries`

@astafan8

### DIFF
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -761,7 +761,8 @@ class DataSet(Sized):
             self,
             *params: Union[str, ParamSpec, _BaseParameter],
             start: Optional[int] = None,
-            end: Optional[int] = None) -> Dict[str, Dict[str, numpy.ndarray]]:
+            end: Optional[int] = None,
+            include_setpoints = True) -> Dict[str, Dict[str, numpy.ndarray]]:
         """
         Returns the values stored in the DataSet for the specified parameters
         and their dependencies. If no paramerers are supplied the values will
@@ -789,6 +790,8 @@ class DataSet(Sized):
                 if None
             end: end value of selection range (by results count); ignored if
                 None
+            include_setpoints: if False, only return the values for the
+                top level parameters instead of the entire trees
 
         Returns:
             Dictionary from requested parameters to Dict of parameter names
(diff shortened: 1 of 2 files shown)


jenshnielsen:
@thibaudruelle Thanks, This looks good. We should perhaps add a test

astafan8:
Indeed, we'd need a test.

My worry is that for some dataset the data returned with `include_setpoints=True` may be different from `include_setpoints=False`. But it's a worry, i can't immediately come up with such a dataset.

thibaudruelle:
What kind of test would you write ? A list of expected / actual return value comparisons as for `get_data` ?

astafan8:
> What kind of test would you write ?

yes, feel free to get inspired by our tests, for example, [this one](https://github.com/QCoDeS/Qcodes/blob/987dcf340df4caf57f8eeb01569c0af94d03f68b/qcodes/tests/dataset/test_measurement_context_manager.py#L1787)

The key is to test the behavior with this flag, for various "parameter trees".

PR #1612, opened 2019-06-21
Explicit experiment in notebooks

jenshnielsen:
Update a bunch of notebooks to use an explicit experiment. Along the way make it possible to pass an experiment when loading legacy data. 

There are a lot of other improvements that could be made to these notebooks. These are explicitly out of scope for this pr. 
The doND notebook has not been updated. This needs an active experiment to make any sense


docs/examples/DataSet/Load old data.ipynb
```diff
@@ -17,8 +17,8 @@
     "import qcodes as qc\n",
     "from qcodes.dataset.legacy_import import import_dat_file\n",
     "\n",
-    "from qcodes.dataset.database import initialise_database\n",
-    "from qcodes.dataset.experiment_container import new_experiment\n",
+    "from qcodes import initialise_database\n",
+    "from qcodes.dataset.experiment_container import load_or_create_experiment\n",
```
astafan8:
```suggestion
    "from qcodes import load_or_create_experiment\n",
```

docs/examples/DataSet/Pedestrian example of subscribing to a DataSet.ipynb
```diff
@@ -210,112 +200,45 @@
...
+      "    self.run()\n",
+      "  File \"c:\\users\\jenielse\\source\\repos\\qcodes\\qcodes\\dataset\\data_set.py\", line 143, in run\n",
+      "    self._loop()\n",
+      "  File \"c:\\users\\jenielse\\source\\repos\\qcodes\\qcodes\\dataset\\data_set.py\", line 168, in _loop\n",
+      "    self._call_callback_on_queue_data()\n",
+      "  File \"c:\\users\\jenielse\\source\\repos\\qcodes\\qcodes\\dataset\\data_set.py\", line 157, in _call_callback_on_queue_data\n",
+      "    self.callback(result_list, self._data_set_len, self.state)\n",
+      "  File \"<ipython-input-9-348631bac194>\", line 15, in ASCII_plotter_5bit\n",
+      "    plotline[yvalue] = 'O'\n",
+      "IndexError: list assignment index out of range\n",
```
astafan8: exception?

qcodes/dataset/legacy_import.py
```diff
@@ -1,20 +1,27 @@
...
     """
     Register parameters for all DataArrays in a given QCoDeS legacy dataset
 
     This tries to infer the name, label and unit along with any setpoints
     for the given array.
+
+    Args:
+        dataset: Legacy dataset to register parameters from.
+        exp: experiment that the legacy dataset should be bound to. If
+            None
```
astafan8: if None? some text seems to be missing

qcodes/dataset/legacy_import.py
```diff
@@ -65,14 +72,21 @@ def store_array_to_database_alt(meas, array):
...
+def import_dat_file(location: str,
+                    exp: Optional[Experiment] = None) -> List[int]:
     """
-    This imports a QCoDeS legacy DataSet
+    This imports a QCoDeS legacy DataSet into the database.
+
+    Args:
+        location: Path to file containing legacy dataset
+        exp: Specify the experiment to store data to.
+            If None the default one is used.
```
astafan8: perhaps not "default" but "latest" or just refer to the docs of "Measurement" object?

docs/examples/DataSet/Load old data.ipynb
jenshnielsen: I guess we should stick to that but we really should get rid of that importing random mixture of everything from the toplevel module.

qcodes/dataset/legacy_import.py
jenshnielsen: The default may change, but sure we can link to the measurement object.

docs/examples/DataSet/Pedestrian example of subscribing to a DataSet.ipynb
jenshnielsen: This is indeed strange I did not change this logic

qcodes/dataset/legacy_import.py
jenshnielsen: The docs of measurement are equally bad :( I will link that and fix it there

jenshnielsen:
@astafan8 All suggestions should have been addressed. I can no longer reproduce the subscription issue, but that is in any case not specific to this branch

PR #1614, opened 2019-06-24
Add plot_dataset method 

jenshnielsen:
And make plot by id uses this under the hood. There seems to be no good reason why you 
need to supply the `run_id` to plot if you already have a loaded dataset. So refactor the code a bit to make
it possible to directly plot a dataset.


qcodes/dataset/plotting.py
```diff
@@ -90,7 +90,33 @@ def plot_by_id(run_id: int,
...
+    return plot_dataset(dataset,
+                        axes,
+                        colorbars,
+                        rescale_axes,
+                        auto_color_scale,
+                        cutoff_percentile,
+                        **kwargs)
+
+
+def plot_dataset(dataset: DataSet,
```
astafan8: shall this also be a method on the DataSet for convenience?

qcodes/dataset/plotting.py
```diff
@@ -90,7 +90,33 @@ def plot_by_id(run_id: int,
                                                  Number]] = None,
                **kwargs) -> AxesTupleList:
     """
-    Construct all plots for a given run
+    Construct all plots for a given `run_id`. All other arguments are forwarded
+    to :meth:`.plot_dataset` See this for more details.
```
astafan8:
```suggestion
    to :meth:`.plot_dataset`, see this for more details.
```

qcodes/dataset/plotting.py
jenshnielsen: As agreed offline we will leave this is as now

PR #1615, opened 2019-06-24
Feature: allow users to link datasets to each other

WilliamHPNielsen:
Add an interface (in the broadest possible sense of that word) for users to annotate that the measurement they are about to perform has N parents and also allow some text-based annotations for that relationship (e.g. "fit", "analysis").

It is only possible to link to datasets in the same DB file as the new measurement will be written to.


emartinez:
@WilliamHPNielsen I like this :)

qcodes/dataset/linked_datasets/links.py
```diff
@@ -0,0 +1,85 @@
+"""
+This module defines the Edge dataclass as well as two functions to read and
```
astafan8:
```suggestion
This module defines the Link dataclass as well as two functions to read and
```

qcodes/dataset/linked_datasets/links.py
```diff
@@ -0,0 +1,85 @@
...
+write an Edge object to/from string, respectively
```
astafan8:
```suggestion
write an Link object to/from string, respectively
```

qcodes/dataset/linked_datasets/links.py
```diff
@@ -0,0 +1,85 @@
...
+        self.validate_node(self.head, "head")
+        self.validate_node(self.tail, "tail")
+
+
+def link_to_str(link: Link) -> str:
+    """
+    Convert a Link to a string
+    """
+    data_attrs = ['head', 'tail', 'edge_type', 'description']
+    ldict = {da: getattr(link, da) for da in data_attrs}
```
astafan8: this can be done using `asdict` function from `dataclasses` module.

qcodes/dataset/linked_datasets/links.py
```diff
@@ -0,0 +1,85 @@
...
+    link = Link(**ldict)
+    return link
+
+
+def links_to_str(links: List[Link]) -> str:
+    """
+    Convert a list of links to string. Note that this is the output that gets
+    stored in the DB file
+    """
+    output = json.dumps([link_to_str(link) for link in links])
```
astafan8: why json-ifying a list of json strings instead of json-ifying a list of Link objects? suggestion: `json.dumps(links)`

qcodes/dataset/measurements.py
```diff
@@ -720,6 +731,27 @@ def _paramspecbase_from_strings(
...
+        Args:
+            parent: the parent dataset
+            link_type: a name for the type of parent-child link
+            description: a free-text description of the relationship
+        """
+        # we save the information in a way that is very compatible with the
+        # Link object we will eventually make out of this information. We
+        # cannot create a Link object just yet, because the DataSet of this
+        # Measurement has not been given a GUID yet
+        parent_dict = {'tail': parent.guid, 'edge_type': link_type,
```
astafan8: shall this then also become a dataclass of its own? say `DataSetReference` or `ParentReference` or else

qcodes/dataset/sqlite/queries.py
```diff
@@ -1161,6 +1162,21 @@ def _update_run_description(conn: ConnectionPlus, run_id: int,
...
+    Update (i.e. overwrite) the parent_datasets field for the given run_id
+    """
+
+    sql = """
+          UPDATE runs
+          SET parent_datasets = ?
+          WHERE run_id = ?
+          """
+    with atomic(conn) as conn:
+        conn.cursor().execute(sql, (links_str, run_id))
```
astafan8: what is a difference between datasets that have parent_datasets equal to `NULL` and `"[]"`? if there's no difference, then shall this column be populated with `"[]"` and hence `if maybe_link_str is None: link_str = json.dumps([])` in `get_parent_dataset_links` is not needed?

qcodes/dataset/linked_datasets/links.py
WilliamHPNielsen: It shall be done so.

qcodes/dataset/linked_datasets/links.py
WilliamHPNielsen: `Link` objects are not `json` serializable.

qcodes/dataset/measurements.py
WilliamHPNielsen: I'd say that's overkill for now. We probably want to change from `dataclass` to some protobuf-based solution once we start evolving the `Link`s. Let's wait until we know more before we implement too much.

qcodes/dataset/sqlite/queries.py
WilliamHPNielsen: I think you are right, there should not be two different values (`NULL`/`None` and `"[]"`/`[]`) that represent the same thing. I'll make the upgrader populate the column accordingly and apply the simplification you propose.

qcodes/dataset/measurements.py
astafan8: agree

WilliamHPNielsen:
Closing this in favour of #1639

PR #1618, opened 2019-06-25
Don't get db location from config if conn supplied

jenshnielsen:
The conn could point to any file on disk so this will in general not be correct.
Fix this by looking up the db path from the connection supplied and using that rather than relying on the config. Only fall back to config if no path or conn is supplied. 

Also fix a small issue where the path separators could be inconsistent if the incorrect platform dependent path sepeartor was used in the config file. e.g. on windows `~/experiments.db` would previously expand to `c:\\users\\username/experiments.db` 

Another issue discovered as part of #1605 

This should preferably have a test.

### DIFF
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -28,7 +28,7 @@ from qcodes.dataset.sqlite.queries import add_parameter, create_run, \
 from qcodes.dataset.sqlite.query_helpers import select_one_where, length, \
     insert_many_values, insert_values, VALUE, one
 from qcodes.dataset.sqlite.database import get_DB_location, connect, \
-    init_conn
+    conn_from_dbpath_or_conn
 from qcodes.instrument.parameter import _BaseParameter
 from qcodes.dataset.descriptions.rundescriber import RunDescriber
 from qcodes.dataset.descriptions.dependencies import (InterDependencies_,
@@ -237,7 +237,7 @@ class DataSet(Sized):
                              "been passed together with non-None values. "
                              "This is not allowed.")
 
-        self.conn, self._path_to_db = init_conn(conn, path_to_db)
+        self.conn = conn_from_dbpath_or_conn(conn, path_to_db)
 
         self.conn = make_connection_plus_from(conn) if conn is not None else \
             connect(self.path_to_db)
@@ -294,7 +294,7 @@ class DataSet(Sized):
 
     @property
     def path_to_db(self):
-        return self._path_to_db
+        return self.conn.path_to_dbfile
 
     @property
     def name(self):
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -24,11 +24,11 @@ from qcodes.dataset.sqlite.queries import add_parameter, create_run, \
     get_sample_name_from_experiment_id, get_guid_from_run_id, \
     get_runid_from_guid, get_run_timestamp_from_run_id, get_run_description,\
     get_completed_timestamp_from_run_id, update_run_description, run_exists,\
-    remove_trigger, set_run_timestamp, get_db_location_from_conn
+    remove_trigger, set_run_timestamp
 from qcodes.dataset.sqlite.query_helpers import select_one_where, length, \
     insert_many_values, insert_values, VALUE, one
 from qcodes.dataset.sqlite.database import get_DB_location, connect, \
-    get_DB_debug
+    get_DB_debug, path_to_dbfile
 from qcodes.instrument.parameter import _BaseParameter
 from qcodes.dataset.descriptions.rundescriber import RunDescriber
 from qcodes.dataset.descriptions.dependencies import (InterDependencies_,
@@ -293,7 +293,7 @@ class DataSet(Sized):
             raise ValueError('Received BOTH conn and path_to_db. Please '
                              'provide only one or the other.')
         if conn is not None:
-            path_to_db = get_db_location_from_conn(conn)
+            path_to_db = path_to_dbfile(conn)
         else:
             path_to_db = path_to_db or get_DB_location()
         conn = conn or connect(path_to_db, get_DB_debug())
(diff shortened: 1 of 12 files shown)


qcodes/dataset/data_set.py
```diff
@@ -284,6 +287,18 @@ def __init__(self, path_to_db: str = None,
                 self._interdeps = InterDependencies_()
             self._metadata = get_metadata_from_run_id(self.conn, self.run_id)
 
+    @staticmethod
```
jenshnielsen: @astafan8 This static method is duplicated in the experiment class. I would like to extract it into a helper function but I am not sure where that belongs in the new structure.

qcodes/dataset/sqlite/queries.py
```diff
@@ -1590,3 +1590,16 @@ def remove_trigger(conn: ConnectionPlus, trigger_id: str) -> None:
         name: id of the trigger
     """
     transaction(conn, f"DROP TRIGGER IF EXISTS {trigger_id};")
+
+
+def get_db_location_from_conn(conn: ConnectionPlus) -> str:
```
jenshnielsen: This is not needed. We already have path_to_dbfile

astafan8: does `path_to_dbfile` use the same query?

either way, it would be really nice to remove redundancy.

qcodes/dataset/sqlite/database.py
```diff
@@ -190,7 +190,7 @@ def get_db_version_and_newest_available_version(path_to_db: str) -> Tuple[int,
 
 
 def get_DB_location() -> str:
-    return expanduser(qcodes.config["core"]["db_location"])
+    return normpath(expanduser(qcodes.config["core"]["db_location"]))
```
astafan8: is `get_DB_location` used everywhere in qcodes? in other words, are there other places where "normpath(expanduser" would be added?

qcodes/dataset/data_set.py
astafan8: it seems to belong to `qcodes.dataset.sqlite.database` as an convenient thing on top of `connect`/`get_DB_location`/etc. Is this reasonable?

qcodes/dataset/data_set.py
```diff
@@ -284,6 +287,18 @@ def __init__(self, path_to_db: str = None,
...
+    def _init_conn(conn, path_to_db):
+        if path_to_db is not None and conn is not None:
+            raise ValueError('Received BOTH conn and path_to_db. Please '
+                             'provide only one or the other.')
+        if conn is not None:
+            path_to_db = path_to_dbfile(conn)
+        else:
+            path_to_db = path_to_db or get_DB_location()
+        conn = conn or connect(path_to_db, get_DB_debug())
+        return conn, path_to_db
```
astafan8: wow, i never realized how complicatd this logic is. can we use this opportunity to make the code more readable? the `path_to_db or ...` parts distract me a lot.

qcodes/dataset/sqlite/database.py
jenshnielsen: I think so. The exanduser thing is a bit of a pain point. So we have put it here

qcodes/dataset/data_set.py
jenshnielsen: Sure I was guessing you were the clever guy who implemented this 😛

jenshnielsen:
@astafan8 This now makes the path to db a property of the connection+. Do you want to have another look?

qcodes/dataset/sqlite/connection.py
```diff
@@ -151,3 +155,14 @@ def atomic_transaction(conn: ConnectionPlus,
...
+
+def path_to_dbfile(conn: Union[ConnectionPlus, sqlite3.Connection]) -> str:
+    """
+    Return the path of the database file that the conn object is connected to
+    """
+    cursor = conn.cursor()
+    cursor.execute("PRAGMA database_list")
+    row = cursor.fetchall()[0]
+
+    return row[2]
```
astafan8: i liked the version of this function from an earlier commit that had more comments and links to sqlite docs or smth :)

qcodes/dataset/sqlite/database.py
```diff
@@ -227,12 +227,35 @@ def initialise_or_create_database_at(db_file_with_abs_path: str) -> None:
...
+    that take either a `ConnectionPlus` or the path to a db file.
+    If neither is given this will fall back to the default db location.
+    It is an error to supply both.
+
+    Args:
+        conn: A ConnectionPlus object pointing to a sqlite database
+        path_to_db: The path to a db file.
+
+    Returns:
+        Tuple of `ConnectionPlus` and path to db file.
```
astafan8:
```suggestion
        A `ConnectionPlus` object
```

qcodes/tests/dataset/temporary_databases.py
```diff
@@ -28,6 +28,20 @@ def empty_temp_db():
         yield
 
 
+@pytest.fixture(scope='function')
+def empty_temp_db_connection():
+    """
+    Yield the paths of an empty file.
```
astafan8:
```suggestion
    Yield connection to an empty temporary DB file.
```

qcodes/dataset/sqlite/connection.py
```diff
@@ -27,8 +27,10 @@ class ConnectionPlus(wrapt.ObjectProxy):
         atomic_in_progress: a bool describing whether the connection is
             currently in the middle of an atomic block of transactions, thus
             allowing to nest `atomic` context managers
+        path_to_dbfile: Path to the database file of the connection.
     """
     atomic_in_progress: bool = False
+    path_to_dbfile = ''
```
astafan8: could you add some assertions for this to [ConnectionPlus tests as well](https://github.com/QCoDeS/Qcodes/blob/9bf5110c006cb5b01e5368405fe952fe25a613aa/qcodes/tests/dataset/test_sqlite_connection.py#L42) ?

qcodes/dataset/sqlite/connection.py
jenshnielsen: Yes that was from before I realized that we already had this function :) I can put the link here too

jenshnielsen: This is the one that we always had. Just moved to make the dep tree work out

PR #1619, opened 2019-06-25
Two fixes for IPS120 driver.

spxtr:
The dictionary fix is obvious. It's just fallout from #1402.

The leave_persistent_mode fix is less obvious.

Before this change, if we set the field to 0 and then set the
magnet in persistent mode, it will do everything properly, however
the "persistent mode" light will not go on and the magnet will end in
the "zero field, heater off" state. Then, when you try to leave
persistent mode, it will do nothing, since there is no persistent
field. This is obnoxious, since we would like to do sweeps that
involve operations like this:

1. leave persistent mode
2. set a new magnetic field
3. enter persistent mode
4. do a sweep

These will fail if you include 0 in your sweep. It will fail in an
annoying way: all future fields will just be 0.

This change fixes this problem. If the state is "heater off, field
at zero" then it just turns the heater on.

@jenshnielsen @astafan8


astafan8:
@spxtr I don't have experience with this instrument, so I am gonna just trust that you did the right choice. Can you confirm that this exact change has been tested on a real instrument?

spxtr:
Yes, tested on a real instrument.

PR #1622, opened 2019-07-02
Add a method on the channel parameter to check if it is closed.

sohailc:
Currently, to check if a channel on the Keithley S46 is closed we have to write the following: 

```python
s46.A1.get() == "close"
```

This can be error prone: typos can cause this code to fail (e.g. 'closed' instead of 'close', 'Close' with a capital C, accidental spaces, etc). 

Furthermore, a bug of this nature can go unnoticed for a long time and might be hard to track down, as the code above will not generate an exception when 'close' is spelled wrong. 

The following code is much more robust

```python
s46.A1.is_closed()
```

An understandable error message will be thrown when the above code is run with a typo in the method name.

### DIFF
--- a/qcodes/instrument_drivers/tektronix/Keithley_s46.py
+++ b/qcodes/instrument_drivers/tektronix/Keithley_s46.py
@@ -106,6 +106,14 @@ class S46Parameter(Parameter):
                                "that is not attached to an instrument.")
         self._instrument.write(f":{new_state} (@{self._channel_number})")
 
+    def is_closed(self) -> bool:
+        """
+        A public method to query the channel state without relying on string comparison.
+
+        Returns: True if channels is closed, False otherwise.
+        """
+        return self.get() == "close"
+
     @property
     def channel_number(self) -> int:
         return self._channel_number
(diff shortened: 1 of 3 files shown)


qcodes/instrument_drivers/tektronix/Keithley_s46.py
```diff
@@ -106,6 +106,14 @@ def _set(self, new_state) -> None:
                                "that is not attached to an instrument.")
         self._instrument.write(f":{new_state} (@{self._channel_number})")
 
+    def is_closed(self) -> bool:
+        """
+        A public method to query the channel state without relying on string comparison.
```
astafan8: This part of the docstring seems irrelevant, the description of the return values seems clear enough.
```suggestion
```

sohailc: Ok, I will change this

astafan8:
> carefully determine the order of tests to get around pyvisa-sim limit…  …
…ations: `test_is_closed` should be run before opening/closing channels on the mock instrument.

please do not depend on the order the tests are executed. this is so annoying to debug if there are problems (and in fact we recently had such an experience). just prepare the insturment in the correct state in the test (or in a separate fixture), and then test the method.

sohailc:
@astafan8 can you have another look at this, please?

PR #1624, opened 2019-07-08
Updater function for QCoDeS

GateBuilder:
This PR adds the functionality of updating QCoDeS installation with a simple function call.

General capabilities, besides updating QCoDeS, of this PR are:

- User can choose to back up his/her (any) environment at will. 
- User can use the function to update Conda and Anaconda Python Environment, explicitly.
- User can upgrade pip installed and outdated packages.

@jenshnielsen

### DIFF
--- a/qcodes/utils/qcodes_updater.py
+++ b/qcodes/utils/qcodes_updater.py
@@ -1,0 +1,145 @@
+import os
+import subprocess
+import sys
+import shutil
+import datetime
+import time
+import json
+import warnings
+
+import qcodes
+
+
+def update_qcodes_installation(back_up=True, env_name='qcodes', env_backup_name='qcodes_backup',
+                               conda_update=True, env_update=True, pip_upgrade=False, qcodes_update=True):
+
+    """
+    Update current QCoDeS installation.
+
+    Args:
+        back_up (bool): Back up current installation before update
+
+        env_name (str): Name of the environment to be updated
+
+        env_backup_name (str): Name of the back up environment to be created
+
+        conda_update (bool): Update 'conda' manager of Anaconda Python installation
+
+        env_update (bool): Update QCoDeS python environment
+
+        pip_upgrade (bool): When True, upgrade 'pip' and all the packages in the environment
+        listed as 'outdated' by pip. This argument should be used with care.
+
+        qcodes_update (bool): Updates QCoDeS module
+
+    Raises:
+        ImportError: If QCoDeS module cannot be imported after the update
+    """
+
+    # Will back up the QCoDeS Anaconda Python environment.
+
+    _source = os.sep.join(qcodes.__file__.split(os.sep)[:-2])
+
+    if not back_up == False:
+
+        _name = str(datetime.date.today()) # Current date.
+        env_backup_name = env_backup_name + '_' + _name
+
+        print("Existing {} environment will be backed up as {}...\n".format(env_name, env_backup_name))
+
+        subprocess.run('conda create --name {} --clone {}'.format(env_backup_name, env_name), shell=True)
+
+        # Copy QCoDeS root to qcodes_backup
+
+        _destination = _source + '_backup' + '_' + _name
+
+        shutil.copytree(_source, _destination, symlinks=True, ignore=None)
+
+        # Re-install QCoDeS from the back up root
+
+        subprocess.run('activate {} && pip uninstall qcodes && pip install -e {}'
+                       .format(env_backup_name, _destination), shell=True)
+
+    # Will update the Conda environment.
+
+    if not conda_update == False:
+
+        print("Updating Conda...\n")
+        subprocess.run('conda update -n base conda -c defaults', shell=True)
+
+    # Update QCoDeS Environment
+
+    # Note that Conda will not upgrade a package, even if there is a newer
+    # version, in the case that the package satisfies the requirement specified
+    # by QCoDeS.
+
+    if not env_update == False:
+
+        print("Updating QCoDeS environment...\n")
+        subprocess.run('cd {} && conda env update'.format(_source), shell=True)
+
+    # Update pip
+
+    if pip_upgrade == True:
+
+        print("Upgrading pip...\n")
+        subprocess.run('pip install --user --upgrade pip', shell=True)
+
+        print("Querying outdated packages...\n")
+        out_pkgs = subprocess.run('pip list --outdated --format=json', shell=False,
+                                  check=False, stdout=subprocess.PIPE).stdout
+        _out_pkgs = out_pkgs.decode('utf8').replace("'", '"')
+        _out_pkgs_dict = json.loads(_out_pkgs)
+
+        print("Upgrading packages...")
+
+        for i in range(len(_out_pkgs_dict)):
+            pkg = _out_pkgs_dict[i]["name"]
+            if not (pkg == 'qcodes' and pkg == 'qdev_wrappers'):
+                subprocess.run('pip install -U --upgrade-strategy=only-if-needed ' +
+                               '{}'.format(pkg), shell=True)
+            else:
+                pass
+
+    # Update QCoDeS
+
+    if not qcodes_update == False:
+
+        print('Making a pull request from QCoDeS repositories...\n')
+
+        subprocess.run('cd {} && git stash && git checkout master && git pull && git stash pop'.format(_source), shell=True)
+
+        print("Updating QCoDeS...\n")
+
+        subprocess.run('pip install -e {}'.format(_source), shell=True)
+
+    # Test the installation
+
+        try:
+
+            import qcodes as qc
+            print("QCoDeS version {} is succesfully installed.".format(qc.__version__))
+            sys.exit()
+
+        except ImportError:
+
+            warnings.warn("An unknown issue occured during update.\nThe changes shall be rolled back.", UserWarning, 2)
+
+            subprocess.run('conda deactivate && conda remove --name qcodes --all', shell=True)
+
+            print("Cloning QCoDeS from back up...\n")
+
+            subprocess.run('conda create --name {} --clone {}'.format('qcodes', env_backup_name), shell=True)
+            subprocess.run('conda remove --name {} --all'.format(env_backup_name), shell=True)
+
+            # Roll back to the roots
+
+            shutil.rmtree(_source)
+            shutil.copytree(_destination, _source, symlinks=True, ignore=None)
+
+            subprocess.run('activate qcodes && pip uninstall qcodes && pip install -e {}'
+                           .format(_source), shell=True)
+
+            shutil.rmtree(_destination)
+
+            sys.exit()


qcodes/utils/qcodes_updater.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def update_qcodes_installation(back_up=True, env_name='qcodes', env_backup_name='qcodes_backup',
+                               conda_update=True, env_update=True, pip_upgrade=False, qcodes_update=True):
+
+    """
+    Update current QCoDeS installation.
+
+    Args:
+        back_up (bool): Back up current installation before update
```
jenshnielsen: Please remove these blank lines between each line

qcodes/utils/qcodes_updater.py
```diff
@@ -0,0 +1,142 @@
...
+    # by QCoDeS.
+
+    if not env_update == False:
+
+        print("Updating QCoDeS environment...\n")
+        subprocess.run('cd {} && conda env update'.format(_source), shell=True)
+
+    # Update pip
+
+    if pip_upgrade == True:
```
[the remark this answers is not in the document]
GateBuilder: I am currently implementing a "version checker" in such cases where after updating pip if there is a confliction in versions that are needed and with the ones that are currently installed, it will correct them. But until it is done, ı agree, I will remove pip option.

jenshnielsen:
I would not document this now but rely on users locally testing it until we are happy that it works

qcodes/utils/qcodes_updater.py
```diff
@@ -0,0 +1,101 @@
...
+    # by QCoDeS.
+
+    if env_update:
+        print("Updating QCoDeS environment...\n")
+        subprocess.run('conda env update', shell=True, cwd=_source)
+
+    # Update QCoDeS
+
+    if qcodes_update:
+        print('Making a pull request from QCoDeS repositories...\n')
```
jenshnielsen: This should read something like `Updating from QCoDeS master`

qcodes/utils/update/qcodes_updater.py
```diff
@@ -0,0 +1,110 @@
...
+        env_name: Name of the environment to be updated
+        env_backup_name: Name of the back up environment to be created
+    """
+    _source = os.sep.join(qcodes.__file__.split(os.sep)[:-2])
+
+    # Make sure that git working tree is clean
+    repo = git.Repo(_source)
+    if repo.is_dirty():
+        print("QCoDeS working tree is not clean. Please commit or stash your changes, and try again.")
+        sys.exit()
```
jenshnielsen: Since this is a function and not a script I think it should just return

qcodes/utils/update/qcodes_updater.py
```diff
@@ -0,0 +1,110 @@
...
+        subprocess.run(f'conda create --name {env_name} --clone {env_backup_name}', shell=True)
+        subprocess.run(f'conda remove --name {env_backup_name} --all', shell=True)
+
+        # Roll back to the roots
+        shutil.rmtree(_source)
+        shutil.copytree(_destination, _source, symlinks=True, ignore=None)
+        subprocess.run(f'activate qcodes && pip uninstall qcodes && pip install -e {_source}', shell=True)
+        shutil.rmtree(_destination)
+
+        sys.exit()
```
jenshnielsen: as above

jenshnielsen:
I take back what I wrote about docs. I think it should be documented here https://qcodes.github.io/Qcodes/start/index.html#installation

qcodes/utils/update/qcodes_updater.py
GateBuilder: Yes, in case the back up function called directly, a simple return would suffice. However, when this one called from one of the update functions (they are designed to do so), and If the tree is dirty, it will return None, and an unnecessary error will raise as update functions expects 4 strings as return values. I think code should terminate if there is no possibility for an update?

qcodes/utils/update/qcodes_updater.py
GateBuilder: I agree, this is unnecessary and the line will be removed.

qcodes/utils/update/qcodes_updater.py
WilliamHPNielsen: I think that reasoning is a bit upside-down. It's the responsibility of whoever calls the function to correctly handle the different outcomes of that call, not the responsibility of the function itself to try to anticipate where and how it is going to get called.

In this case, I'd suggest that the update function raises if the backup fails. Or, alternatively, that the backup function raises if the tree is not clean. The update function can then catch that error and re-raise it with some context.

docs/start/index.rst
```diff
@@ -92,6 +92,36 @@ If you have installed with git, pull the QCoDeS repository using your
...
+    from qcodes.utils.update.qcodes_updater import update_qcodes_installation
+    update_qcodes_installation()
+
+which will first back up your current installation, then will update your
+Anaconda Python environment and finally will do an update of QCoDeS through
+official git repositories.
+
+You can choose to update Anaconda Python environment explicitly
+
+.. code:: bash
```
jenshnielsen: why not `.. code:: python` ?

qcodes/utils/update/qcodes_updater.py
```diff
@@ -0,0 +1,127 @@
...
+    # QCoDeS source path
+    _source = os.sep.join(qcodes.__file__.split(os.sep)[:-2])
+    # Make sure that git working tree is clean
+    repo = git.Repo(_source)
+    if (inspect.stack()[1].function == 'update_qcodes_installation' and repo.is_dirty()):
+        raise git.exc.GitError
+    elif (inspect.stack()[1].function == 'conda_env_update' and repo.is_dirty()):
+        raise git.exc.GitError
+    elif repo.is_dirty():
+        print("QCoDeS working tree is not clean. Please commit or stash your changes, and try again.")
```
jenshnielsen: I suggest that we just replace the 3 sections above with raising the same exception. Then you move the alternative handling of the exception into whatever calls the function. We can define our own exception to make it explicit that we only catch that exception.

qcodes/utils/update/qcodes_updater.py
```diff
@@ -0,0 +1,127 @@
...
+        subprocess.run(f'conda create --name {env_backup_name} --clone {env_name}', shell=True)
+
+        # Copy QCoDeS root to qcodes_backup
+        _destination = _source + '_backup' + '_' + _time
+        shutil.copytree(_source, _destination, symlinks=True, ignore=None)
+
+        # Re-install QCoDeS from the back up root
+        subprocess.run(f'activate {env_backup_name} && pip uninstall qcodes && pip install -e {_destination}', shell=True)
+
+        return _source, _destination, env_name, env_backup_name
```
jenshnielsen: In general there is no need to use private `_` variables inside a function. They are private to the function scope anyway

qcodes/utils/update/qcodes_updater.py
```diff
@@ -0,0 +1,127 @@
...
+        raise git.exc.GitError
+    elif repo.is_dirty():
+        print("QCoDeS working tree is not clean. Please commit or stash your changes, and try again.")
+    else:
+        _time = str(datetime.date.today()) # Current date.
+        env_backup_name = env_backup_name + '_' + _time
+
+        print(f"Existing {env_name} environment will be backed up as {env_backup_name}...\n")
+        subprocess.run(f'conda create --name {env_backup_name} --clone {env_name}', shell=True)
+
```
jenshnielsen: Did you check that thi

qcodes/utils/update/qcodes_updater.py
```diff
@@ -0,0 +1,127 @@
...
+    elif (inspect.stack()[1].function == 'conda_env_update' and repo.is_dirty()):
+        raise git.exc.GitError
+    elif repo.is_dirty():
+        print("QCoDeS working tree is not clean. Please commit or stash your changes, and try again.")
+    else:
+        _time = str(datetime.date.today()) # Current date.
+        env_backup_name = env_backup_name + '_' + _time
+
+        print(f"Existing {env_name} environment will be backed up as {env_backup_name}...\n")
+        subprocess.run(f'conda create --name {env_backup_name} --clone {env_name}', shell=True)
```
jenshnielsen: Did you check that this correctly copies pip installed packages

qcodes/utils/update/qcodes_updater.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # Copy QCoDeS root to qcodes_backup
+        _destination = _source + '_backup' + '_' + _time
+        shutil.copytree(_source, _destination, symlinks=True, ignore=None)
+
+        # Re-install QCoDeS from the back up root
+        subprocess.run(f'activate {env_backup_name} && pip uninstall qcodes && pip install -e {_destination}', shell=True)
```
jenshnielsen: How will this work if qcodes is not installed as an editable install (with `-e`)

qcodes/utils/update/qcodes_updater.py
```diff
@@ -0,0 +1,127 @@
...
+    try:
+        if not inspect.stack()[1].function == 'update_qcodes_installation':
+            # Back up first
+            _source, _destination, env_name, env_backup_name = qcodes_backup()
+            # Then, update
+            execute_update(_source)
+        else:
+            # QCoDeS source path
+            _source = os.sep.join(qcodes.__file__.split(os.sep)[:-2])
+            execute_update(_source)
```
jenshnielsen: Again I really want to discourage functions where the functionality depends on who calls them.

qcodes/utils/update/qcodes_updater.py
```diff
@@ -0,0 +1,127 @@
...
+        env_name: Name of the environment to be updated
+        env_backup_name: Name of the back up environment to be created
+
+    Raises:
+        git.exc.GitError: If working tree is dirty
+    """
+    # QCoDeS source path
+    _source = os.sep.join(qcodes.__file__.split(os.sep)[:-2])
+    # Make sure that git working tree is clean
+    repo = git.Repo(_source)
```
jenshnielsen: We probably need code here to check if this is even a repo. E.g qcodes could be installed directly from pip

qcodes/utils/update/qcodes_updater.py
GateBuilder: Yes, Conda clone literally creates a clone of the environment with all the installed packages and preserve the versions.

qcodes/utils/update/qcodes_updater.py
GateBuilder: Yes, it does not handle that situation and this question is related to the last one. I will resolve them together.

WilliamHPNielsen:
I like where this PR is going. I still have some reservations before merging, though. 

Since all subprocess calls can fail, I think it might be good to have some error handling in place. I tried out `conda_env_update` just now, which failed during the first call to subprocess. It would probably make sense for the function to abort after the intial fais, but currently, the rest of the updating continues. What do you think?

I think it would be good to revisit the documentation for the installation instructions. It is a tad confusing that updating the environment is documented twice in different ways.

I have left two more comments inline.

qcodes/utils/update/qcodes_updater.py
```diff
@@ -0,0 +1,174 @@
...
+
+def address_keeper() -> Tuple[str, str]:
+    # QCoDeS source path
+    src = os.sep.join(qcodes.__file__.split(os.sep)[:-2])
+    time_stamp = str(datetime.date.today()) # Current date.
+    dst = src + '_backup' + '_' + time_stamp
+    return src, dst
+
+
+def qcodes_is_editable() -> bool:
```
WilliamHPNielsen: Now that we have the same function in `installation_info`, we should probably use that one here to avoid code duplication.

qcodes/utils/update/qcodes_updater.py
```diff
@@ -0,0 +1,174 @@
...
+import subprocess
+import urllib.request
+import warnings
+import qcodes
+
+class DirtyTreeException(Exception):
+    pass
+
+
+def address_keeper() -> Tuple[str, str]:
```
WilliamHPNielsen: As a public function, this function should have a docstring.

qcodes/utils/update/qcodes_updater.py
```diff
@@ -0,0 +1,174 @@
...
+    if qcodes_is_editable():
+        source, destination = address_keeper()
+        # Make sure that git working tree is clean
+        repo = git.Repo(source)
+        if repo.is_dirty():
+            raise DirtyTreeException("QCoDeS working tree is not clean. " +
+                                     "Please commit or stash your changes " +
+                                     "and try again.")
+        else:
+            print(f'Existing {env_name} environment will be backed up as{env_backup_name}...\n')
```
WilliamHPNielsen: Missing space after `backed up as`

PR #1626, opened 2019-07-11
Tektronix AWG 5014 : Fix range/validators of Channel offset, low pass filter, Marker high level and low level  

lakhotiaharshit:
Instrument: AWG5014

Fixed ranges 
1)  Channel offset from: -0.1 - 0.1 to -2.25 to 2.25
2) Channel low pass filter: removed validator 9.9e37
3) Marker high level:  -2.7 - 2.7 to -0.9 - 2.7 
4) Marker low level: -2.7 - 2.7 to -1.0 -2.6
All the values were checked using a physical instrument.


lakhotiaharshit:
@WilliamHPNielsen As we discussed. I validated the range of the parameters using a physical instrument. You were indeed right as I found 3 more incorrect validators.

PR #1628, opened 2019-07-12
Tektronix AWG 70000 5028 resolution docstrings

### LINKED ISSUE #1134: AWG set channel resolution: docstring should reflect allowed values from _chan_resolutions
All three Tektronix AWG devices have their drivers implemented based one AWG70000A driver. AWG5208 is quite different from the other two AWGs as it has 4 markers per channel instead of 2, which is reflected on this line: [_chan_resolutions = {'5208': [12, 13, 14, 15, 16],...](https://github.com/QCoDeS/Qcodes/blob/fe9a68f45cd0a1976e21e1375806ae381113cdbb/qcodes/instrument_drivers/tektronix/AWG70000A.py#L61).

The issue is: the fact about 4 markers instead of 2 is not reflected in the docstring of the `set_resolution` method [see here](https://github.com/QCoDeS/Qcodes/blob/fe9a68f45cd0a1976e21e1375806ae381113cdbb/qcodes/instrument_drivers/tektronix/AWG70000A.py#L289):
```
             docstring=("""
                                      8 bit resolution allows for two
                                      markers, 9 bit resolution
                                      allows for one, and 10 bit
                                      does NOT allow for markers"""))
```

ToDo:
- [ ] make dosctring content dependent on the number of channel resolutions
- [ ] add a test for the dosctring text 😺 

alternatively, perhaps, you'd propose just to generalize docstring in order not to include this dynamic information that is dependent on the number of channels?

WilliamHPNielsen:
Fixes #1134 

- Add a test for the docstring
- Make the docstring model dependent

@astafan8

### DIFF
--- a/qcodes/.coveragerc
+++ b/qcodes/.coveragerc
@@ -9,7 +9,6 @@ omit =
     instrument_drivers/QuTech/*
     instrument_drivers/cryogenic/*
     instrument_drivers/rigol/*
-    instrument_drivers/tektronix/*
     instrument_drivers/Lakeshore/*
     instrument_drivers/Spectrum/*
     instrument_drivers/agilent/*
--- a/qcodes/instrument_drivers/tektronix/AWG70000A.py
+++ b/qcodes/instrument_drivers/tektronix/AWG70000A.py
@@ -62,6 +62,20 @@ _chan_resolutions = {'5208': [12, 13, 14, 15, 16],
                      '70001A': [8, 9, 10],
                      '70002A': [8, 9, 10]}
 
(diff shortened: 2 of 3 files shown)


jenshnielsen:
The coverage report reveals that this is not included in the coverage. Could you change .coverage to not exclude this driver (and possibly other Tektronix drivers that have tests)?

WilliamHPNielsen:
Okay, let's see how much the coverage drops.

jenshnielsen:
I think we can land this with the coverage enabled for Tektronix

PR #1629, opened 2019-07-12
Feature : an elapsed time parameter

WilliamHPNielsen:
There has been recurring requests to get a default QCoDeS parameter to measure elapsed wall clock time in an experiment. This PR introduces a simple `ElapsedTimeParameter` to conveniently measure - you guessed it - elapsed wall clock time.

- Add a parameter to do the job
- Add an example notebook

@QCoDeS/core


WilliamHPNielsen:
I realise that some users might like to have a parameter to measure more absolute time (time stamps). I think that is beyond the scope of this PR.

qcodes/utils/time.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def __init__(self, name: str, label: str = 'Elapsed time', **kwargs):
+
+        super().__init__(name=name,
+                         label=label,
+                         unit='s',
+                         set_cmd=False,
+                         **kwargs)
+
+        self.t0: float = perf_counter()
```
astafan8: should this be private? :)
```suggestion
        self._t0: float = perf_counter()
```

qcodes/utils/time.py
```diff
@@ -0,0 +1,38 @@
...
+    Parameter to measure elapsed time. Measures wall clock time since the
+    last reset of the instance's clock. The clock is reset upon creation of the
+    instance. The constructor passes kwargs along to the Parameter constructor.
+
+    Args:
+        name: the local name of the parameter. Should be a valid
+            identifier, ie no spaces or special characters. If this parameter
+            is part of an Instrument or Station, this is how it will be
+            referenced from that parent, ie ``instrument.name`` or
+            ``instrument.parameters[name]``
```
astafan8: is it really needed to explain this in the docstring of this parameter? this applies to all parameters, right?

qcodes/utils/time.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+    def __init__(self, name: str, label: str = 'Elapsed time', **kwargs):
```
astafan8: it's be nice to block people from passing `get_cmd` but ok)

docs/examples/DataSet/Measuring X as a function of time.ipynb
```diff
@@ -0,0 +1,179 @@
...
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "\n",
+    "import numpy as np\n",
+    "\n",
+    "from qcodes.utils.time import ElapsedTimeParameter\n",
+    "from qcodes.instrument.parameter import Parameter\n",
+    "from qcodes.dataset.measurements import Measurement\n",
+    "from qcodes.dataset.sqlite.database import initialise_or_create_database_at\n",
```
astafan8: i'm not yet comfortable with exposing where this function comes from. i prefer that for now it comes just from `qcodes.`, otherwise people will start looking into what else they can use from `.sqlite.*` while they are not allowed because it's all our internal stuff that can change, etc.

docs/examples/DataSet/Measuring X as a function of time.ipynb
```diff
@@ -0,0 +1,179 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Measuring X as a function of time\n",
+    "\n",
+    "Sometimes we'd like to measure something as a function of elapsed wall clock time. This notebook provides a simple example of doing that using the `ElapsedTimeParameter`."
```
astafan8: I'd recommend to add here the reason behind `ElapsedTimeParameter` - the fact that qcodes provides it, hence all qcodes users in principle can use the same parameter which makes it easier to relate measurements which use this parameter (sorry, on friday evening my language skills are gone, but i hope you got what i mean :) )

docs/examples/DataSet/Measuring X as a function of time.ipynb
```diff
@@ -0,0 +1,179 @@
...
+      ]
+     },
+     "execution_count": 2,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "initialise_or_create_database_at(os.path.join(os.getcwd(), 'x_as_a_function_of_time.db'))\n",
+    "new_experiment('tutorial', 'no_sample')"
```
jenshnielsen: I changed most (all) of the other example notebooks around to use load_or_create_experiment and explicitly pass that to the measurement. I think this is the best solution until we have an explicit experiment switcher

qcodes/utils/time.py
WilliamHPNielsen: Correct. It has been shortened now.

qcodes/utils/time.py
WilliamHPNielsen: Done. Along with `unit` and `set_cmd`, which have hardcoded values for this particular parameter.

docs/examples/DataSet/Measuring X as a function of time.ipynb
WilliamHPNielsen: I have moved one import to resolve this. We should make a separate PR that attacks the mess of the qcodes namespace.

docs/examples/DataSet/Measuring X as a function of time.ipynb
WilliamHPNielsen: I have added a little bit. I'm not too sure what you are after... -do we need to document why standardization is helpful?

docs/examples/DataSet/Measuring X as a function of time.ipynb
WilliamHPNielsen: Absolutely.

qcodes/utils/time.py
WilliamHPNielsen: Okay.

WilliamHPNielsen:
After discussions with Jens, I have moved the code from `utils` to a new module next to the parameters. I agree that this makes more sense.

docs/examples/DataSet/Measuring X as a function of time.ipynb
astafan8: yes, in a subtle way, like "this parameter provides a single way of measuring time which is awesome because standardization is awesome in this case" :)

jenshnielsen:
I think we should include this in the api docs in docs/api/parameter/* Just let me know if you need a hand setting that up correctly

WilliamHPNielsen:
@jenshnielsen was this what you had in mind?

PR #1630, opened 2019-07-16
Docs examples

GateBuilder:
This PR proposes editorial changes and corrections to the following example notebooks:

- Configuring_QCoDeS.ipynb
- Station.ipynb

to increase readability and clarity while decreasing incohenrecies.

@WilliamHPNielsen


docs/examples/Station.ipynb
```diff
@@ -40,22 +40,22 @@
...
+    "The concept of a station, in essence, is a programmatical representation of such a bucket. Instruments, parameters, and other \"components\" of an experiment can be added to a station. The user gets a station instance that can be referred to in order to access those \"components\".\n",
     "\n",
-    "Moreover, stations are very helpful when capturing the state of the experimental setup, known as snapshot. Refer to the respective section about station snapshot below.\n",
+    "A station can be configured from a text file which simplifies the initialization of the instruments. In particular, in this tutorial, we shall provide an example configuration of a station via a YAML file.\n",
     "\n",
-    "Last but not least, station can be configured in a text file which simplifies initialization of the instruments. Read more of this below."
+    "A special use case of a station in an experiment would be the capturing the state of an experimental setup, known as a snapshot. We shall devote a subsection for the concept of a snapshot as a particular method of the station. "
```
astafan8: the "as a particular method of the station" is confusing. what does it suppose to mean?
```suggestion
    "A special use case of a station in an experiment would be the capturing of the state of an experimental setup, known as a snapshot. We shall devote a subsection for the concept of a snapshot. "
```

docs/examples/Configuring_QCoDeS.ipynb
```diff
@@ -17,16 +17,17 @@
...
+    "The config file controls various options to QCoDeS such as the default path and name of the database in which your data is stored and logging level of the debug output. QCoDeS is shipped with a default configuration. As we shall see, you can overwrite these default values in several different ways to customize the configuration. In particular, you may want to change the path of your database which is initialised by the `qcodes.dataset` with the default name `experiments.db`. By default the path of your database is `~/experiments.db` (here, `~` stands for the path of the user's home directory). In the following example, I have changed the default path of my database, flagged with the key `db_location`, \n",
```
astafan8: "is initialised by the `qcodes.dataset`"? what does this mean?)
```suggestion
    "The config file controls various options to QCoDeS such as the default path and name of the database in which your data is stored and logging level of the debug output. QCoDeS is shipped with a default configuration. As we shall see, you can overwrite these default values in several different ways to customize the configuration. In particular, you may want to change the path of your database which by default is `~/experiments.db` (here, `~` stands for the path of the user's home directory). In the following example, I have changed the default path of my database, flagged with the key `db_location`, \n",
```

astafan8: to my taste, the "the default path of my database, flagged with the key `db_location`" part is confusing, i think it can be written in a more clear way.

also "my" and "I" - i'm looking at `'c:\\\\Users\\\\jenielse\\\\mymainfolder\\\\experiments.db'` below, and does this mean that this notebook is written with Jens as a narrator in mind? :)

docs/examples/Configuring_QCoDeS.ipynb
```diff
@@ -186,8 +187,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This will add a base_location with value /dev/random to the current config, and validate it's value to be of type string, will also set the description and what one would want to have as default.\n",
-    "The new entry is saved in the 'user' part of the config."
+    "This will add a base_location with value /dev/random to the current configuration, and validate it's value to be of type string, will also set the description and what one would want to have as default.\n",
```
astafan8: since we're improving the documentation, i guess the formatting of text can be improved on the way, right?
```suggestion
    "This will add a `base_location` with value `/dev/random` to the current configuration, and validate it's value to be of type string, will also set the description and what one would want to have as default.\n",
```

docs/examples/Configuring_QCoDeS.ipynb
```diff
@@ -359,8 +360,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "one can change the core values at runtime, but there is no guarantee that they are valid.\n",
-    "Changing core values will also shadow the config from being updated if the defaults are updated upsteam (i.e. in the repo). So use it with a care."
+    "One can change the core values at runtime, but there is no guarantee that they are valid.\n",
+    "Changing core values will also shadow the configuration from being updated if the defaults are updated in the upsteam (i.e. in the repository). So use it with a care."
```
astafan8: this sentences could be made more clear. for example, smth like "Since user config shadows the default config that comes with QCoDeS, apply care when changing values under `core` section. That section is meant for settings that are determined by QCoDeS core developers. .....bla .... bla ...bla"

docs/examples/Station.ipynb
GateBuilder: I wanted to point out that snapshot is in fact a more general concept, while, here we deal with it as a method of the station object. But we can drop that line as you suggested, no problem.

docs/examples/Configuring_QCoDeS.ipynb
GateBuilder: Before it wan not clear in the text that which key in the dictionary actually corresponds to the location of the dbase.  I can rewrite it. 

I think Jens first wrote this example, so I decided to leave those path names as it is. I believe that It will make no difference.

astafan8: > Before it was not clear in the text that which key in the dictionary actually corresponds to the location of the dbase

ah. ok, but the thing is that the schema of the config file is also added to the docs, and the schema explains all the deafault keys, so perhaps it's more efficient to refer to it?

GateBuilder: In this case, I believe, this should stay as in the cell immediately after, the core dictionary is provided. I will just change "flagged with" with "represented by". A general reference to the schema is not necessary, IMHO.

astafan8: ok, btw this is where the config is completely descibed (what i referred to as "schema") http://qcodes.github.io/Qcodes/user/configuration.html#default-config

PR #1633, opened 2019-07-18
use versioneer to manage version numbers

jenshnielsen:
This means that if you install from a non release version you will get version numbers such as 0.4.0+30.g160bfbc45 where 0.4.0 is the last release before your current version. 30 are the number of commits since the release. g indicates that we use git and 160bfbc45 is the short commit hash.

This also means that there is no longer any need to manually update the version number in version.py 

versioneer.py and _version.py are auto generated and it probably does not make sense to review these files

Replaces #1631


jenshnielsen:
Should probably exclude this code from codecov

qcodes/__init__.py
```diff
@@ -115,4 +115,8 @@ def test(**kwargs):
     return retcode
 
 
-test.__test__ = False # type: ignore # Don't try to run this method as a test
+test.__test__ = False  # type: ignore # Don't try to run this method as a test
+
+from ._version import get_versions
```
astafan8: why was it needed for duplicate code? isn't `from . import version \n __version__ = version.__version__` enough?

versioneer.py
```diff
@@ -0,0 +1,1822 @@
+
+# Version: 0.18
```
astafan8: will we ever need to upgrade this?))

astafan8:
also, could you update the "how to release wiki" that "manually bumping the version is not needed" after this is merged?)

versioneer.py
jenshnielsen: Yes its documented how to upgrade here https://github.com/warner/python-versioneer#updating-versioneer 

I will add a link

jenshnielsen:
@astafan8 yes the wiki is on my todo list after this is merged.

qcodes/__init__.py
jenshnielsen: Sure

versioneer.py
jenshnielsen: It's already documented in the doc string below. I think this is ok

astafan8: yes, it is perfect! thanks for pointing me there

PR #1636, opened 2019-07-22
Base Sphinx makefiles on the modern template

jenshnielsen:
This significantly reduces the amount of boilerplate code in the makefiles and should hopefully make the logic a bit easier to follow


docs/Makefile
```diff
@@ -1,267 +1,48 @@
...
-SPHINXBUILD   = sphinx-build
-PAPER         =
+# You can set these variables from the command line, and also
+# from the environment for the first two.
+SPHINXOPTS    ?=
+SPHINXBUILD   ?= sphinx-build
+SOURCEDIR     = .
 BUILDDIR      = _build
-# this is to to autmagic gh-pages
-GH_PAGES_SOURCES = docs qcodes CONTRIBUTING.rst LICENSE
```
jenshnielsen: This should not have been deleted

jenshnielsen:
@astafan8 I am not sure what you mean. Functionally there should be no big change. It is mainly that instead of having all these exilic targets on the line of `target foo: call sphinx build with arg foo` there is just a default target

astafan8:
@jenshnielsen ah, got it now, thanks!

PR #1638, opened 2019-07-23
update m4i docstrings

peendebak:
The m4i acquisition methods do not all set properties for the triggering in the same way. Fixing this would be backwards incompatible. This PR at least updates some of the docstrings.

@astafan8


astafan8:
@peendebak please fix the codacy things, and then i'll merge.

PR #1639, opened 2019-07-23
Feature : allow users to link runs

WilliamHPNielsen:
This PR supersedes #1615 

I need to 

- [x] Take care of merge conflict resolving fallout :P
- [ ] ~Make the version 7 -> 8 upgrader populate the `parent_datasets` column with the default value `"[]"`~ No reason for this, since we've agreed to treat that column as optional
- [x] Update the "generate-fixture-dbs" repo with some version 7 DBs
- [x] Use those DBs in CI for this branch

and then we should be ready to merge.

@QCoDeS/core


qcodes/dataset/data_set.py
```diff
@@ -497,6 +515,29 @@ def add_parameter(self, spec: ParamSpec):
                                   'Please use DataSet.set_interdependencies '
                                   'instead.')
 
+    def set_parent_dataset_links(self, links: List[Link]) -> None:
```
jenshnielsen: Should this be a property setter?

WilliamHPNielsen: Yes, that's probably more natural. The same holds true for `set_interdependencies`, but we can change that later.

qcodes/dataset/linked_datasets/links.py
```diff
@@ -0,0 +1,84 @@
...
+    @staticmethod
+    def validate_node(node_guid: str, node: str) -> None:
+        """
+        Validate that the guid given is a valid guid.
+
+        Args:
+            node_guid: the guid
+            node: either "head" or "tail"
+        """
+        if _guid_pattern.match(node_guid):
```
jenshnielsen: Should we not refactor this out into function in the guids.py (is_valid_guid or some such thing)

qcodes/dataset/measurements.py
```diff
@@ -721,6 +732,27 @@ def _paramspecbase_from_strings(
 
         return tuple(depends_on), tuple(inf_from)
 
+    def register_parent(
+            self: T, parent: DataSet, link_type: str,
```
jenshnielsen: So currently `link_type` can be anything?

qcodes/dataset/sqlite/queries.py
```diff
@@ -1512,6 +1531,37 @@ def get_run_description(conn: ConnectionPlus, run_id: int) -> str:
...
+    """
+
+    # The RuntimeError comes from the atomic contextmanager as it rolls back
+    # the transaction. We assume that the rollback is caused by the column
+    # `parent_datasets` not being present. We find it fair to interpret this
+    # situation as "no parent datasets were registered"
+    #
+    # We cannot in general trust that NULLs will not appear in the column,
+    # so we must explicitly handle them.
+
```
jenshnielsen: I would prefer to use pragma table info to check for the existence. How is a null value in the db handled here?

qcodes/dataset/measurements.py
WilliamHPNielsen: Yes. It's completely the responsibility of the users to ensure that they always write `"analysis"` and never `"analsyis"`, or at least that they can handle that ambiguity. Do you think we can do better than this?

qcodes/dataset/sqlite/queries.py
WilliamHPNielsen: Sure, we can use the pragma. It's a good idea to not just catch-all on `RuntimeError`.

A `NULL` value retrieved as a `None`. The case when `maybe_str_link is None` is handled a few lines down.

qcodes/dataset/measurements.py
jenshnielsen: I think its a good idea to normalize but on the other hand i have no idea of the relevant types so lets leave it for now

qcodes/dataset/sqlite/queries.py
jenshnielsen: Yes either A more specific exception catching or the pragma is fine with me

qcodes/dataset/linked_datasets/links.py
WilliamHPNielsen: Yes, let's do that. Note that the verification is only of the format. Nonsensical values can still slip through.

jenshnielsen: Yes but that can be improved in a future PR

PR #1640, opened 2019-07-24
Use mypy 0.720

jenshnielsen:
This is likely to trigger a few errors so mark it wip for now


qcodes/dataset/plotting.py
```diff
@@ -182,30 +182,32 @@ def plot_dataset(dataset: DataSet,
     nplots = len(alldata)
 
     if isinstance(axes, matplotlib.axes.Axes):
-        axes = [axes]
+        axeslist = [axes]
```
WilliamHPNielsen: I like the way `mypy` thinks :) Readability +1

PR #1644, opened 2019-07-25
Build travis against all branches

jenshnielsen:
I do honestly do not see the benefit of this limitation. And it prevents prs targeting feature branches to be build correctly.


WilliamHPNielsen:
I think it's a remnant from ancient pre-fork days, where all PRs were made from branches on the main fork. However the history, I agree with your conclusion.

PR #1646, opened 2019-07-29
Extract and write data from dbase as .dat file format

GateBuilder:
This PR proposes a method, "write_data_to_gnuplot_file()", in QCoDeS dataset which extracts the data and write it in an external file with the desired format (preferably as .dat).

@jenshnielsen


qcodes/dataset/data_set.py
```diff
@@ -936,6 +942,43 @@ def get_data_as_pandas_dataframe(self,
             dfs[name] = df
         return dfs
 
+    def write_data_to_gnuplot_file(self, single_file: bool = False,
```
jenshnielsen: imho this type should also be Optional[bool]

qcodes/dataset/data_set.py
```diff
@@ -936,6 +942,43 @@ def get_data_as_pandas_dataframe(self,
...
+                                   npath: Optional[str] = None) -> None:
+        """
+        A function to export data as GNU plot format.
```
jenshnielsen: I guess gnuplot is one word and not capitalized. Note that it does not actually have anything to do with the GNU project https://en.wikipedia.org/wiki/Gnuplot

qcodes/dataset/data_set.py
```diff
@@ -936,6 +942,43 @@ def get_data_as_pandas_dataframe(self,
...
+        for parametername, df in dfdict.items():
+            if not single_file:
+                df.to_csv(path_or_buf=f'{parametername}.dat', header=False, sep='\t')
+            else:
+                dfs_to_save.append(df)
+                df_length = len(dfs_to_save[0])
+                if any(len(df) != df_length for df in dfs_to_save):
+                    raise DataLengthException("You cannot concatenate data " +
+                                              "with different length to a " +
+                                              "single file.")
```
jenshnielsen: The section of code below should not be in the loop as far as I can see. Currently you will save the same file multiple times

qcodes/dataset/data_set.py
jenshnielsen: Never mind

jenshnielsen:
Thanks this looks a lot better now. I wonder if we should make it possible to write a header matching what the old dataset did. But that can also be done in a new pr.

WilliamHPNielsen:
FYI: tests are being written, coming soon.

astafan8:
@GateBuilder I've found a similar [Delft-home-made function for exporting data into a text file](https://github.com/kouwenhovenlab/pytopo/blob/master/pytopo/qctools/dataset_export/spyview.py). I think it can be a good inspiration with regards to features and arguments that users would want to see. Feel free to pm me for more info.

astafan8:
May I also suggest that this method is actually a function in a different python module? And, if really requested by users, DataSet class just happens to have a method with the same name that automatically redirects to that function? This was the code of DataSet class is kept "not-that-big" (and the function in principle can be implemented as a composition of smaller functions that can be unit-tested separately, etc etc.)

qcodes/dataset/data_set.py
```diff
@@ -936,6 +943,46 @@ def get_data_as_pandas_dataframe(self,
             dfs[name] = df
         return dfs
 
+    def write_data_to_text_file(self, path: str,
+                                   single_file: bool = False,
+                                   single_file_name: Optional[str] = None) -> None:
+        """
+        An auxiliary function to export data to a text file.
```
WilliamHPNielsen: It would be terrific with a (short) explanation of the format here. Especially the `single_file` case, that one gets
```
x1 y1
x2 y2
... ...
xN yN
x1 z1
x2 z2
... ...
xN zN
```
which is not obvious from the file name.

jenshnielsen: Is that the format that we want. For that format there seems to be no limitations with lenght of the columns. 

I expected the format to be

```
x1 y1 z1
x2 y2 z2
... ...   ...
xN yN zN
```

GateBuilder: Yes this is the format it saves. There was a miscommunication between me and William. So no problem, I will add this info to docstring.

PR #1648, opened 2019-07-29
Dev / log install info

WilliamHPNielsen:
It can be very useful to have some basic information about the installation of qcodes in the logs.

This PR seeks to enrichen the log output with more useful info, in particular installation info. I am, however, not really sure about the logic when looking up the environment. It goes something like
 - First check if the installation is from pip or from source
 - If from pip:
   -  record the version number and call it a day
 - If from source
   - check if editable
   - compare `pip freeze` to `requirements.txt` (?)

The idea is to do all this as `qcodes.logger.start_logging` is called.

Or what do you think, @QCoDeS/core ? I'd like to make as few `subprocess` calls to `pip` as possible.

### DIFF
--- a/qcodes/utils/installation_info.py
+++ b/qcodes/utils/installation_info.py
@@ -1,0 +1,44 @@
+"""
+This module contains helper functions that provide information about how
+QCoDeS is installed and about what other packages are installed along with
+QCoDeS
+"""
+from typing import Dict, List
+import subprocess
+
+
+def _pip_list_parser(pip_list_raw: List[str]) -> Dict[str, Dict[str, str]]:
+    """
+    Parse the raw output of pip list into a dict with package name keys and
+    a dict {'version': ..., 'location': ...} values
+
+    Args:
+        pip_list_raw: a list of strings, one line per list element
+    """
+    out: Dict[str, Dict[str, str]] = {}
+
+    # The first two lines are the header and header-content separator
+    for line in pip_list_raw[2:]:
+        fields = line.split()
+        pkg, ver = fields[0], fields[1]
+        if len(fields) == 3:
+            loc = fields[2]
+        else:
+            loc = ''
+
+        out.update({pkg: {'version': ver, 'location': loc}})
+
+    return out
+
+
+def is_qcodes_installed_editably() -> bool:
+    """
+    Return a boolean with the answer to the question 'is QCoDeS installed in
+    editable mode?'
+    """
+    pipproc = subprocess.run(['pip', 'list', '-e'], stdout=subprocess.PIPE)
+    lines = pipproc.stdout.decode('utf-8').split('\r\n')[:-1]
+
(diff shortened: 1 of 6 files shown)


qcodes/utils/installation_info.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Args:
+        pip_list_raw: a list of strings, one line per list element
+    """
+    out: Dict[str, Dict[str, str]] = {}
+
+    # The first two lines are the header and header-content separator
+    for line in pip_list_raw[2:]:
+        fields = line.split()
+        pkg, ver = fields[0], fields[1]
```
jenshnielsen: This version number cannot be trusted. It reflects the version  at install time for an editable install not necessarily the version at run time. The version number in qcodes.__version__ and qcodes.version.__version__ can however be truested to be updated

jenshnielsen:
I guess the lowest hanging fruit would be to just log `qcodes.version.__version__` this seems to always be up to date. 

Pip does not have a public api but perhaps something like https://pypi.org/project/pip-api/ might be better than rolling out your own subprocess module

jenshnielsen:
I guess logging all installed versions of dependencies would be useful but that is regardless of the install type?

WilliamHPNielsen:
Okay, I think this is a first good stab at the problem.

My only concern is that the `subprocess` call to `pip` is a bit slow. But then again, it happens once.

qcodes/utils/installation_info.py
```diff
@@ -0,0 +1,119 @@
...
+def _pip_list_parser(pip_list_raw: List[str]) -> Dict[str, Dict[str, str]]:
+    """
+    Parse the raw output of pip list into a dict with package name keys and
+    a dict {'version': ..., 'location': ...} values
+
+    Args:
+        pip_list_raw: a list of strings, one line per list element
+    """
+    out: Dict[str, Dict[str, str]] = {}
+
```
jenshnielsen: Pips output format changed some years back. I suspect this may fail for older versions of pip . Do we care.

jenshnielsen: I just discovered that there is also `pip list -e --format json` that could make parsing simpler?

jenshnielsen: Specifically the default changed here https://pip-python3.readthedocs.io/en/stable/reference/pip_list.html#cmdoption-format from legacy to columns at some point. 

Specifying the format upfront is probably a good idea

jenshnielsen: The changes are documented here https://pip.pypa.io/en/stable/news/ 

It seems like columns was added in 9.0 (2016), the default changed in 10.0 (2018) and the legacy format removed in 18.0 (2018).

WilliamHPNielsen:
@jenshnielsen I think the latest commit addresses all your concerns.

PR #1650, opened 2019-07-31
Azure Use newest non deprecated version of publish task

jenshnielsen:
Apparently there is now a newer version of the PublishTestResults task in Azure pipelines and the old one triggers a deprecation warning. So lets use the new one. 

https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=azure-devops&tabs=yaml

https://github.com/MicrosoftDocs/vsts-docs/issues/2181


jenshnielsen:
Also take the chance to remove the html reports. We no longer use them so in the coverage task so we might as well save the time used to generate them.

WilliamHPNielsen:
This looks fine.

PR #1651, opened 2019-07-31
Feature: Instruments snapshot exclude parameters

nyxus:
- Added way to exclude parameters from the instruments snapshot
- Added another test case to the param_to_skip_update of the instruments snapshot test

At the DiCarlo lab, we use a couple of devices which have many (instruments) parameters (E.g., QWG, HDAWG8, UHFQC). Most of these parameters are not useful for us to log (or backup) via the snapshot feature. There are so many useless parameters that we have problems with backing up all the information. With this feature, the user or device manufacture can specify which parameters not to include.

This code is tested on our devices, and I added tests to the snapshot tests.   

@WilliamHPNielsen, @AdriaanRol

### DIFF
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -156,7 +156,8 @@ class InstrumentBase(Metadatable, DelegateAttributes):
         self.submodules[name] = submodule
 
     def snapshot_base(self, update: bool=False,
-                      params_to_skip_update: Optional[Sequence[str]] = None
+                      params_to_skip_update: Optional[Sequence[str]] = None,
+                      params_to_exclude: Sequence[str] = None
                       ) -> Dict:
         """
         State of the instrument as a JSON-compatible dict (everything that
@@ -172,6 +173,7 @@ class InstrumentBase(Metadatable, DelegateAttributes):
                 different way (as in the qdac). If you want to skip the
                 update of certain parameters in all snapshots, use the
                 `snapshot_get`  attribute of those parameters instead.
+            params_to_exclude: Sequence[str] = None
 
         Returns:
             dict: base snapshot
@@ -179,6 +181,9 @@ class InstrumentBase(Metadatable, DelegateAttributes):
 
         if params_to_skip_update is None:
             params_to_skip_update = []
+
+        if params_to_exclude is None:
+            params_to_exclude = []
 
         snap = {
             "functions": {name: func.snapshot(update=update)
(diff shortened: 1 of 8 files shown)


jenshnielsen:
Thanks. Is there a particular reason why the snapshot_value and snapshot_get options on the individual parameters does not enable you to do what you want? I.e the two bools documented here https://qcodes.github.io/Qcodes/api/parameters/parameter.html#qcodes.instrument.parameter.Parameter

nyxus:
@jenshnielsen, both the `snapshot_get` and `snapshot_value` still leave the parameter in the output of the snapshot. They change the behaviour of the parameter when the snapshot is called. 
For example, the QWG (inhouse developed AWG) has parameters to assign waves to codewords on a channel (`codeword_1_ch1_waveform('sine')`). The QWG has 4 channels and per-channel 1024 codewords. In total, this adds 4096 parameters (or entries) to the snapshot. The same problems occur for the HDAWG8 which has 8 channels and even more codewords than the QWG. We do not need these parameters in the snapshot because the information is logged (also) on a different level. Therefore the `params_to_exclude` which will completely leave out these kinds of parameters from the snapshot.

nyxus:
I'm looking at the errors of the CI, and I do not know why they fail. 
Error 1:
```
qcodes/instrument_drivers/ZI/ZIUHFLI.py:1655:5: error: Signature of "snapshot_base" incompatible with supertype "InstrumentBase"
qcodes/instrument_drivers/ZI/ZIHDAWG8.py:66:5: error: Signature of "snapshot_base" incompatible with supertype "InstrumentBase"
```
I added a new optional parameter, why does it fail? 

Error 2:
The Travis errors are also a bit weird. But looking at other builds the also fail with the same errors. So I guess this has nothing to do with this PR?

jenshnielsen:
@nyxus thanks for the explanation that makes sense. One could also consider extending the parameter to fully have this feature. I am honestly not sure what is the cleanest.

 As a historic context. The parameter_to_skip update functionality is mainly there so that we can handle instruments that have a shortcut functionality allowing the update to be done as a one shot operation for all parameters. 

The 2 drivers overwrites the snapshot_base method. I think they just need to be updated too so that the signature of the method is kept in sync.

nyxus:
After some internal discussion in DCL we decided to also implement this feature as suggested by @jenshnielsen. In this way clear that a property of a parameter instead of a separate list. It also continues the idea of the behaviour of the snapshot to include in the parameter itself (see `snapshot_get` and `snapshot_value`). See PR #1653 for the new implementation.

WilliamHPNielsen:
Closing this in favour of #1653

PR #1653, opened 2019-08-02
 Feature: Added instrument paramater snapshot_exclude 

nyxus:
- Added a parameter to the instrument parameter to exclude it from the snapshot. 

Note that this is very similar to #1651, the goal is the same but it has a different implementation. To see the reason why we need this please look at #1651 which has a detailed problem description. This implementation was suggested by @jenshnielsen, which uses the instrument parameter to store the `snapshot_exclude` instead of providing a list to the snapshot function. 

@jenshnielsen, @AdriaanRol

### DIFF
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -202,7 +202,7 @@ class InstrumentBase(Metadatable, DelegateAttributes):
             except:
                 # really log this twice. Once verbose for the UI and once
                 # at lower level with more info for file based loggers
-                logging.info("Snapshot: Could not update parameter: {}".format(name))
+                self.log.warning(f"Snapshot: Could not update parameter: {name}")
                 self.log.info(f"Details for Snapshot:", exc_info=True)
                 snap['parameters'][name] = param.snapshot(update=False)
 
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -160,7 +160,8 @@ class InstrumentBase(Metadatable, DelegateAttributes):
                       ) -> Dict:
         """
         State of the instrument as a JSON-compatible dict (everything that
-        the custom JSON encoder class :class:'qcodes.utils.helpers.NumpyJSONEncoder'
+        the custom JSON encoder class
+        :class:'qcodes.utils.helpers.NumpyJSONEncoder'
         supports).
 
         Args:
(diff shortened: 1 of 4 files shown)


nyxus:
Note that I still need to add tests.

nyxus:
I've added tests and also tested the new implementation on a device on my desk. Can someone review the code? @jenshnielsen, @WilliamHPNielsen?

qcodes/instrument/base.py
```diff
@@ -190,19 +190,22 @@ def snapshot_base(self, update: bool=False,
...
+                snap['parameters'][name] = param.snapshot(update=update_par)
             except:
                 # really log this twice. Once verbose for the UI and once
                 # at lower level with more info for file based loggers
-                self.log.warning(f"Snapshot: Could not update parameter: "
-                                 f"{name}")
-                self.log.info(f"Details for Snapshot:",
-                              exc_info=True)
-
+                logging.info("Snapshot: Could not update parameter: {}".format(name))
```
jenshnielsen: Please change this back to `self.log.warning`

qcodes/tests/test_snapshot.py
```diff
@@ -37,6 +37,44 @@ def test_snapshot_skip_params_update(request, params, params_to_skip):
...
+    for param_exl in params_to_exclude:
+        inst.parameters[param_exl].snapshot_exclude = True
+
+    snap = inst.snapshot()
+    snap_params = [x for x in snap["parameters"]]
+    params = [x for x in params if x not in params_to_exclude]
+
+    assert all(elem not in snap_params for elem in params_to_exclude), f"Parameter(s) is not excluded from the " \
+        f"snapshot. expected: {params}, actual: {snap_params}"
+
```
jenshnielsen: Would it make sense to extend this to also test that all of the parameters in `params` are indeed in the snapshot?

qcodes/instrument/base.py
nyxus: Done

qcodes/tests/test_snapshot.py
nyxus: I was thinking the same thing. And the code is so simple nor hurts it, so added to the test.

nyxus:
I've adapted the code after review and fixed the line length issues.

qcodes/instrument/parameter.py
```diff
@@ -156,6 +156,12 @@ class _BaseParameter(Metadatable):
         snapshot_value (Optional[bool]): False prevents parameter value to be
             stored in the snapshot. Useful if the value is large.
 
+        snapshot_exclude (Optional[bool]): True prevents parameter to be
+            included in the snapshot. Useful if there are many of the same
+            parameter which are clogging up the snapshot.
+
```
jenshnielsen: Please remove this blank line. I think this will render incorrectly in the html of the docs otherwise

jenshnielsen:
Looks good. I will leave this for a bit to givee others a chance to merge

qcodes/instrument/parameter.py
nyxus: Done!

PR #1654, opened 2019-08-04
Feature/m4i/exact sample rate

peendebak:
The M4i `sample_rate` parameter returns the sample rate, but rounded to an integer number of Hz. This PR adds a method to return the exact sample rate.

The original parameter is unchanged and naming is similar to the old parameter.

@astafan8 @jenshnielsen


peendebak:
@jenshnielsen I executed
```
prev_rate=0
for rate in np.linspace(.005e6, 250e6, 300):
    m4i.sample_rate(rate)
    if prev_rate ==m4i.sample_rate():
        continue
    prev_rate = m4i.sample_rate()
    print(f'sample_rate {m4i.sample_rate()} exact_sample_rate {m4i.exact_sample_rate()}' )
```
to check. The `exact_sample_rate` gives the correct values.

jenshnielsen:
@peendebak Cool thanks

PR #1656, opened 2019-08-05
Add functionalities to dpo measurement

sohailc:
We add the following functionalities to the measurement submodule of the driver:

1) Ability to gather measurement statistics: mean, max, min and standard deviation 
2) Automatically turn a measurement on when calling a measurement method (Set `state` to True) 
3) Ability to globally set the statistics mode, time constant and buffer resetting

@astafan8 Can you have a look, please?

### DIFF
--- a/qcodes/instrument_drivers/tektronix/DPO7200xx.py
+++ b/qcodes/instrument_drivers/tektronix/DPO7200xx.py
@@ -6,7 +6,7 @@ MSO70000/C/DX Series Digital Oscilloscopes
 import textwrap
 import time
 from functools import partial
-from typing import Union, Callable
+from typing import Union, Callable, cast
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


qcodes/instrument_drivers/tektronix/DPO7200xx.py
```diff
@@ -777,17 +839,37 @@ def _set_source(self, source_number: int, value: str) -> None:
...
+                "OFF", "ALL", "VALUEMean", "MINMax",
+                "MEANSTDdev"
+            )
+        )
+
+        self.add_parameter(
+            "time_constant",
+            get_cmd="MEASUrement:STATIstics:WEIghting?",
+            set_cmd="MEASUrement:STATIstics:WEIghting {}",
+            get_parser=int
```
jenshnielsen: It's not obvious to me how this works. Would it make sense to include a docstring for this parameter?

sohailc:
@jenshnielsen, I think I have addressed your comment. Can you merge this PR if there are no further issues?

PR #1658, opened 2019-08-06
Subscriber: return a dict from parameter name to result rather than a tuple

jenshnielsen:
This makes the results much easier to work with resolves an issue where the order of the values in the tuple may be random. 

TODO
- [ ] Change documentation to reflect the new subscriber api
- [ ] This is an api break, is that ok? Alternatively we could make the subscriber optionally 
      be able to return either tuples or dicts. The default would then be lists but with a warning that it      changes eventually


astafan8:
> This is an api break, is that ok? Alternatively we could make the subscriber optionally
be able to return either tuples or dicts. The default would then be lists but with a warning that it changes eventually

ideally, i'd vote for the "return either tuples or dicts" solution.

But in this particular case, i'm in for not following the discipline (which one should never do) - here breaking the api will force some user's code to break, so that they are forced to make it better (in my opinion) by using parameter names instead of indices.

jenshnielsen:
@astafan8  I would also be happy with the both options for now, warn and then switch default and eventually make array an error forcing users to adapt but giving them some time to do it

PR #1659, opened 2019-08-06
Integrate applications insight into logging

WilliamHPNielsen:
Telemetry can be very useful.

- Integrate application insights into qcodes 
- Make it explicitly opt-in for the user to send any information anywhere

@QCoDeS/core


setup.py
```diff
@@ -30,6 +30,7 @@ def readme():
     'pandas',
     'tabulate',
     'tqdm'
```
jenshnielsen: Missing `,`

qcodes/logger/logger.py
```diff
@@ -145,6 +151,17 @@ def get_log_file_name() -> str:
...
 
+
+def flush_telemetry_traces() -> None:
+    """
+    Flush the traces of the telemetry logger. If telemetry is not enabled, this
+    function does nothing.
+    """
+    if qc.config.telemetry.enabled:
+        global telemetry_handler
+        telemetry_handler.flush()
```
jenshnielsen: Does it make sense to hook this function up to atexit?

WilliamHPNielsen: Good question. The `logging.shutdown()` already flushes, so it might be superfluous. I'll try to dig into that.

qcodes/tests/__init__.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+from qcodes.config import Config
+
+
+@pytest.fixture(scope="session", autouse=True)
+def disable_telemetry():
+    """
```
jenshnielsen: Do we really need to write to home? Can we not just modify the `qcodes.config` instance in memory. Alternatively could we save to current dir or use the env variable to load a custom file

WilliamHPNielsen:
@QCoDeS/core a question: I included the benchmark notebook for you to look at, but should we include it in the docs? It could be good for users to know at least something about the impact of enabling telemetry.

qcodes/tests/__init__.py
WilliamHPNielsen: That dreaded `qc.config`... Let me look at it. I agree that producing files on disk is not a nice side effect of running the tests. And of course, a file in the current dir could even break the effect of this fixture.

jenshnielsen:
You could also put the notebook in the reports repo

qcodes/tests/__init__.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+import qcodes as qc
+
+
+@pytest.fixture(scope="session", autouse=True)
+def disable_telemetry():
```
jenshnielsen: In antisipation that we are going to also want to disable other things (subscribers at least) should we call this method something like `setup_config_for_testing`?

WilliamHPNielsen:
Re the notebook: I'll put it in the reports repo in a separate PR.

qcodes/tests/__init__.py
WilliamHPNielsen: Or is it cleaner to use one fixture per disabled feature?

WilliamHPNielsen: In any case, this has now been moved into `conftest.py` (Thanks, @Dominik-Vogel!)

qcodes/tests/conftest.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+import qcodes as qc
+
+
+@pytest.fixture(scope="session", autouse=True)
+def disable_telemetry():
+    """
+    We do not want the tests to send up telemetric information. This fixture
+    can be imported to disable the telemetry. Each `test_*.py` file that
```
jenshnielsen: Last part of the docstring is out of date?

qcodes/logger/logger.py
```diff
@@ -18,9 +18,15 @@
...
 from typing import Optional, Union, Sequence
 
+from applicationinsights.logging.LoggingHandler import LoggingHandler
+
 import qcodes as qc
 import qcodes.utils.installation_info as ii
 
+# We need to declare the type of this global variable up here. See
+# https://github.com/python/mypy/issues/5732 for reference
+telemetry_handler: LoggingHandler
```
jenshnielsen: Is it worth considering wrapping this in "if typechecking" and move the imports down to make appinsights an optional dependency?

qcodes/tests/conftest.py
WilliamHPNielsen: Good point!

PR #1661, opened 2019-08-09
Disable subscribers during tests

jenshnielsen:
They are unwanted here and may interfere with tests
Inspired by the solution in #1659


Dominik-Vogel:
I don't quite understand. This value is set to `[]` by `qcodes/config/qcodesrc.json` unless explicitly set in the test. What is the use case of this fixture?

Dominik-Vogel:
Thanks for the offline explanation @jenshnielsen . We can merge that for now. I suggest we add an auto fixture based on https://github.com/QCoDeS/Qcodes/blob/a777a65e98a3086c156dbe640ce5a12cc0140867/qcodes/tests/test_config.py#L142 to make the state of the config independent of the local settings.

PR #1666, opened 2019-08-13
Set journal mode of the db on creation.

jenshnielsen:
This should make it simpler for users to test different journal modes. This does not change the default journal mode.


jenshnielsen:
@astafan8 That should address your comments

PR #1668, opened 2019-08-13
Add logging line to measurement

WilliamHPNielsen:
This is a very natural thing to log; that a measurement started.

@QCoDeS/core


WilliamHPNielsen:
@astafan8 It's not completely clear where the completion logging should happen, but probably in the dataset as the dataset is marked complete. Exceptions are already caught by the logging.

PR #1670, opened 2019-08-15
Give all upgraders a progress bar

WilliamHPNielsen:
All the DB upgrades should have a progress bar to make the output of the upgrading less confusing.

- Add a progress bar to all updaters

This PR doesn't feel particularly DRY, but I could only think of complicated ways of adding the progress bar to the `upgrader` decorator.

@QCoDeS/core

### DIFF
--- a/qcodes/dataset/sqlite/db_upgrades/__init__.py
+++ b/qcodes/dataset/sqlite/db_upgrades/__init__.py
@@ -17,6 +17,7 @@ from functools import wraps
 from typing import Dict, Callable
 
 import numpy as np
+from tqdm import tqdm
 
 from qcodes.dataset.guids import generate_guid
 from qcodes.dataset.sqlite.connection import ConnectionPlus, \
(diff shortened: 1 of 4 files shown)


jenshnielsen:
Could I suggest that we switch to using `from tqdm.auto import tqdm` 

That gives a much nicer output in the notebook for instance.

Compare 

newupgrade

to 

oldupgrade

jenshnielsen:
Forget that. It seems like that does not work in Spyder :(

jenshnielsen:
I wish we could at least avoid it printing to std err but use regular std out

jenshnielsen:
I think we can do that with `tqdm(range(1, no_of_runs+1), file=sys.stdout)` 

Happy to push that change if you are ok with it

WilliamHPNielsen:
@jenshnielsen, I like your suggestions. Too bad with spyder. But going for stdout is already an improvement. Push away :)

jenshnielsen:
@WilliamHPNielsen pushed the change

PR #1678, opened 2019-08-19
Speedup import

jenshnielsen:
Speed up import by avoiding to import a number of 3th party packages before they are needed

This opens the discussion of how the toplevel qcodes namespace should look like. Currently it does a number of imports that are more or less well related and some of these are relatively slow.


Dominik-Vogel:
I just pulled your branch for my testing, and I am super happy how much faster it makes the testing cycle! Great job!

jenshnielsen:
@QCoDeS/core this is ready for review

Issue #1645, opened 2019-07-29
Bad black magic import of config


WilliamHPNielsen:
As discovered by the dynamic duo @jenshnielsen and `mypy` in #1640, the import of the config module and the later overriding assignemnt in the main `__init__.py` file is a bit creepy.

https://github.com/QCoDeS/Qcodes/blob/1d08d81adbb173e5f8bd486ee2a2fc3454641b1e/qcodes/__init__.py#L7-L14

Like, what is `qcodes.config` now? What should it be?

I propose that we resolve this by renaming (with an appropriate deprecation cycle) the `qcodes.config` module to `qcodes.configuration`. I think that will have the minimal impact on user workflows.

@QCoDeS/core what do you think?

jenshnielsen:
I agree, This will however need careful testing to ensure that the currently (by blackmagic working) logic does not break

astafan8:
i agree with the issue, and the proposed solution.

Akshita07:
This issue was solved by #1682

PR #1682, opened 2019-08-20
Move config to configuration

jenshnielsen:
This resolves the ambiguity noted in #1645 by moving the config module to configuration such that

* qcodes.config  is the current active config
* qcodes.configuration is the configuration module.

- [x] Update any notebooks and documentation. 
- [x] I have chosen not to leave around the old alias as it was super confusing and not clear how that would resolve compared to the current config. 

@QCoDeS/core This is now ready for final review


WilliamHPNielsen:
@jenshnielsen, to answer your question
> Should we leave around qcodes.config as an alias for the time being and remove it later?

Wouldn't it work to leave it around with a deprecation warning like we've done for some of the `dataset` modules, e.g. `qcodes.dataset.sqlite_base.py`? If that's possible, I'd vote for us taking the same approach here.

jenshnielsen:
@QCoDeS/core please have a look at the current state. This now incorporates @astafan8 s work on types from #1824

Before merging I need to:

* write a deprecation message for the changelog 
* double check if the docs needs updates

docs/user/configuration.rst
```diff
@@ -25,49 +25,52 @@ Default config
...
 The following  table explains the default configuration of qcodes.
 
-.. jsonschema:: ../../qcodes/config/qcodesrc_schema.json
+.. jsonschema:: ../../qcodes/configuration/qcodesrc_schema.json
 
 
 Using the config
 ----------------
-QCoDeS exposes the default values in the module namespace as `defaults`
+QCoDeS exposes the default values in the module namespace as `config`
```
astafan8:
```suggestion
QCoDeS exposes the configuration values in the `qcodes` module namespace as `config`:
```

qcodes/__init__.py
```diff
@@ -4,13 +4,13 @@
...
 
-from qcodes.config import Config
+import qcodes.configuration as qcconfig
 from qcodes.logger import start_all_logging
 from qcodes.logger.logger import conditionally_start_all_logging
 from qcodes.utils.helpers import add_to_spyder_UMR_excludelist
 from .version import __version__
 
-config: Config = Config()
+config: qcconfig.Config = qcconfig.Config()
```
astafan8: is the type annotation still needed here? :(
```suggestion
config = qcconfig.Config()
```

qcodes/tests/test_config.py
```diff
@@ -152,39 +153,39 @@ def default_config(user_config: Optional[str] = None):
...
     """
-    home_file_name = qcodes.Config.home_file_name
-    schema_home_file_name = qcodes.Config.schema_home_file_name
-    env_file_name = qcodes.Config.env_file_name
-    schema_env_file_name = qcodes.Config.schema_env_file_name
-    cwd_file_name = qcodes.Config.cwd_file_name
-    schema_cwd_file_name = qcodes.Config.schema_cwd_file_name
-
-    qcodes.Config.home_file_name = ''
+    home_file_name = qcodes.configuration.Config.home_file_name
```
astafan8: is `qcodes.configuration.` prefix needed if `Config` class has alrady been imported into this namespace?

qcodes/__init__.py
jenshnielsen: yes

astafan8: ok, I lack some mypy understanding

jenshnielsen: I honestly don't know why but mypy fails without it

PR #1683, opened 2019-08-20
Notebook Dataset Performance: Correct typos and add figure

lakhotiaharshit:
Example notebook data performance contains many typing errors/ grammatical mistakes etc. Importantly, output figures are not displayed on the webpage. This PR fixes these issues.  

- Display output figures.
- Correct typing errors.
- Fix the deprecation warning related to "The module `qcodes.dataset.database` is deprecated".
- Have short (no repetition) and long(100 repetition)  experiment the same 'sizes' of data so that comparison is easier.  

@WilliamHPNielsen 
@jenshnielsen 
@Dominik-Vogel


jenshnielsen:
The figures in the html output at https://qcodes.github.io/Qcodes/examples/DataSet/Dataset%20Performance.html are generated in CI. I suspect that the space in the filename may also be a problem that causes missing figures.

jenshnielsen:
Could you also rewrite the last few cells to `use get_parameter_data`?

jenshnielsen:
I suggest that we delete everything from 
```
This is probably more useful as a numpy array. Here we use squeze to get rid of any singleton dimensions.
```
and down. We do not want to promote get_data_by_id anymore

lakhotiaharshit:
> The figures in the html output at https://qcodes.github.io/Qcodes/examples/DataSet/Dataset%20Performance.html are generated in CI. I suspect that the space in the filename may also be a problem that causes missing figures.

@jenshnielsen Changing the name to have no spaces did not help. But making figures static (by removing %matplotlib) did help in getting the figures in the HTML.  I suspect we have to add some more extensions to Sphinx.
Some suggest the extensions in the discussion below.  
https://stackoverflow.com/questions/30397384/embedding-matplotlib-plots-in-sphinx-how-to-show-documentation.

jenshnielsen:
Ok thats a shame. We are using a different extension to build the notebooks to rst https://github.com/spatialaudio/nbsphinx I am not sure these things are compatible. 

I think its fine to change the notebook from `%matplotlib notebook` to `%matplotlib inline` which will give static figures.

PR #1684, opened 2019-08-21
Keithley 2600 Driver: Implement time trace and autorange

GateBuilder:
This PR proposes the following additions to the Keithley 2600 series SourceMeter Driver:

- Add time trace functionality
- Add current and voltage autorange functionality

Subsumes and closes #1617.
Subsumes and closes #1623.

Progress:

- [x] Implement time trace
- [x] Implement autorange
- [x] Make sure that there is a clear communication between manual and autorange parameters
- [x] Add docstrings
- [x] Add tests
- [x] Add examples

@jenshnielsen


qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -79,6 +81,102 @@ def get_raw(self) -> np.ndarray:
...
+    current and voltage.
+    """
+
+    def __init__(self, name: str, instrument: Instrument, **kwargs) -> None:
+
+        super().__init__(name=name,
+                         instrument=instrument,
+                         **kwargs)
+
+    def _prepareTimeTrace(self, mode: str) -> None:
```
WilliamHPNielsen: Is there still a reason for this function to take a `mode` arg?

As far as I can see, this function is now a (weak) validation function; it either does nothing or gives a warning. That is not a bad thing to have, but the name and docstring should reflect the usage.

GateBuilder: Hi, I changed the name to _validateTimeTrace. It needs mode argument as indeed it checks the mode and raises a ValueError. These are written in the docstring as well.

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -182,7 +182,7 @@ def get_raw(self) -> np.ndarray:
 
 class TimeAxis(Parameter):
     """
-    A simple :class:`Parameter` that holds the time interval
```
jenshnielsen: Why this change. I think the correct way to do this is 
```
`:class:`.Parameter`
```

See the docs here for how this is resolved http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#python-roles

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -74,11 +76,132 @@ def get_raw(self) -> np.ndarray:
...
+        Args:
+            mode: User defined mode for the timetrace. It can be either
+            current, "i", or voltage "v".
+
+        Raises:
+            ValueError: If the timetrace mode is not set to either curent or
+            voltage.
+        """
+
+        if mode not in ['i', 'v']:
```
WilliamHPNielsen: Since the `timetrace_mode` parameter already has a validator, there is no reason to re-validate here (or anywhere else) as long as that parameter's value is used as `mode`.

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -74,11 +76,125 @@ def get_raw(self) -> np.ndarray:
...
+            self.unit='V'
+            self.label='Voltage'
+
+    def _time_trace(self, npts: int, dt: float, mode: str) -> np.ndarray:
+        """
+        The function that prepares a Lua script for timetrace data acquisition.
+
+        Args:
+            npts: Number of points.
+            dt: Infinitesimal time resolution.
```
WilliamHPNielsen: There is nothing infinitesimal about the time resolution. But it may be good to clarify the units (seconds).

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -74,11 +76,125 @@ def get_raw(self) -> np.ndarray:
...
+
+        Raises:
+            RuntimeError: If no instrument attached to Parameter.
+        """
+
+        if self.instrument is None:
+            raise RuntimeError("No instrument attached to Parameter.")
+
+        channel = self.instrument.channel
+        self.shape = (npts,)
```
WilliamHPNielsen: There is no reason for `TimeTrace` to have a `shape` attribute.

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -74,11 +76,125 @@ def get_raw(self) -> np.ndarray:
...
+            current, "i", or voltage "v".
+        """
+        if mode == 'i':
+            self.unit='A'
+            self.label='Current'
+        if mode == 'v':
+            self.unit='V'
+            self.label='Voltage'
+
+    def _time_trace(self, npts: int, dt: float, mode: str) -> np.ndarray:
```
WilliamHPNielsen: Why does this method take arguments? All of `npts`, `dt`, and `mode` are attributes of `self`, aren't they?

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
WilliamHPNielsen: But the best thing is probably to just get rid of the input parameters (see comment above).

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -74,11 +76,125 @@ def get_raw(self) -> np.ndarray:
...
+        self._check_time_trace()
+        npts = self.instrument.npts()
+        dt = self.instrument.dt()
+        data = self._time_trace(npts, dt, mode)
+        return data
+
+
+class TimeAxis(Parameter):
+    """
+    A simple :class:`.Parameter` that holds the time interval
```
WilliamHPNielsen: Perhaps it's just me, but I find the wording confusing. The `TimeAxis` doesn't hold the time interval, it holds all the times (relative to the measurement start) at which the points of the `TimeTrace` were acquired.

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -74,11 +76,125 @@ def get_raw(self) -> np.ndarray:
...
+        """
+
+        if self.instrument is None:
+            raise RuntimeError("No instrument attached to Parameter.")
+
+        channel = self.instrument.channel
+        self.shape = (npts,)
+
+        script = ['{}.measure.count={}'.format(channel, self.shape[0]),
+        'oldint={}.measure.interval'.format(channel),
```
WilliamHPNielsen: Can we indent the entire Lua script for readability?

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -224,6 +379,35 @@ def __init__(self, parent: Instrument, name: str, channel: str) -> None:
         self.add_parameter('fastsweep',
                            parameter_class=LuaSweepParameter)
 
+        self.add_parameter('npts',
```
WilliamHPNielsen: Since the `npts` parameter only affects the number of points of the `TimeTrace`, maybe it should have a name signifying that, e.g. `time_trace_npts`? The same goes for `dt` below.

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -224,6 +379,35 @@ def __init__(self, parent: Instrument, name: str, channel: str) -> None:
...
+
+        self.add_parameter('timetrace',
+                           vals=vals.Arrays(shape=(self.npts,)),
+                           setpoints=(self.time_axis,),
+                           parameter_class=TimeTrace)
+
+        self.add_parameter('timetrace_mode',
+                           get_cmd = None,
+                           set_cmd = self.timetrace._set_mode,
+                           vals=vals.Enum('i', 'v'))
```
WilliamHPNielsen: There are different schools of thought on this topic, but I prefer elaborate, easy-to-read names over quickly typed names. Should we change `"i"` and `"v"` to `"current"` and `"voltage"`?

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -311,7 +495,20 @@ def _fast_sweep(self, start: float, stop: float, steps: int,
...
+        This is the function that sends the Lua script to be executed and
+        returns the corresponding data from the buffer.
+
+        Args:
+            _script: The Lua script to be executed.
+            steps: Number of points.
+        """
+        nplc = self.nplc()
+
+        self.write(self._parent._scriptwrapper(program=_script, debug=True))
```
WilliamHPNielsen: I think `root_instrument` is preferred over `_parent`.

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -338,6 +535,17 @@ def _fast_sweep(self, start: float, stop: float, steps: int,
 
         return outdata
 
+    def _check_autorange_v(self, val: int) -> None:
```
WilliamHPNielsen: This is surely a detail, but I find the name of this function confusing; it does not check anything. Consistently with other drivers, I think we should call it `_set_sourcerange_i`.

WilliamHPNielsen: Why isn't there a similar function for the measurerange?

astafan8: > This is surely a detail

in my book it is a very important detail :) i support this comment.

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -146,18 +140,20 @@ def _time_trace(self, npts: int, dt: float, mode: str) -> np.ndarray:
...
-        'printbuffer(1, {}, {}.nvbuffer1.readings)'.format(self.shape[0], channel)]
+        npts = self.instrument.npts()
+        dt = self.instrument.dt()
+        mode = self.instrument.timetrace_mode()
+
+        script = ['{}.measure.count={}'.format(channel, npts),
+                  'oldint={}.measure.interval'.format(channel),
+                  '{}.measure.interval={}'.format(channel, dt),
+                  '{}.nvbuffer1.clear()'.format(channel),
+                  '{}.measure.{}({}.nvbuffer1)'.format(channel, mode, channel),
```
WilliamHPNielsen: Note that if we do go ahead and change the `mode` values to be `"current" ` and `"voltage"`, then this line needs to change.

qcodes/tests/drivers/test_keithley_26xx.py
```diff
@@ -41,8 +43,8 @@ def test_smu_channels_and_their_parameters(driver):
         assert 'current' == smu.mode()
         smu.mode('voltage')
 
-        assert 'off' == smu.output()
-        smu.output('on')
+        assert 0 == smu.output()
+        smu.output(1)
```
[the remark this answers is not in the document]
WilliamHPNielsen: @astafan8 is right, but we usually adhere to the `assert actual_value == expected_value` way of asserting in tests. So that would be `assert smu.output() == 1`.

This is a long standing war between Mikhail and me 😄

WilliamHPNielsen:
While we're at it, it might be good to remove the introduction from the notebook saying that this is the notebook for the NEW driver. It's been two years now.

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -311,10 +490,23 @@ def _fast_sweep(self, start: float, stop: float, steps: int,
...
+            steps: Number of points.
+        """
+        nplc = self.nplc()
+
+        self.write(self.root_instrument._scriptwrapper(program=_script, debug=True))
         # we must wait for the script to execute
-        oldtimeout = self._parent.visa_handle.timeout
-        self._parent.visa_handle.timeout = 2*1000*steps*nplc/50 + 5000
+        oldtimeout = self.root_instrument.visa_handle.timeout
+        self.root_instrument.visa_handle.timeout = 2*1000*steps*nplc/50 + 5000
```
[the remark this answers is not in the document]
WilliamHPNielsen: Also note that the magic number `50` here is the line frequency. We should never hard-code that value.

qcodes/tests/drivers/test_keithley_26xx.py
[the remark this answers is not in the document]
astafan8: yes yes :) in this case i followed the format of the already existing `assert`ions

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -74,11 +76,120 @@ def get_raw(self) -> np.ndarray:
...
+
+    def _check_time_trace(self) -> None:
+        """
+        A helper function that compares the integration time with measurement
+        interval for accurate results.
+
+        Raises:
+            RuntimeError: If no instrument attached to Parameter.
+        """
+        if self.instrument is None:
```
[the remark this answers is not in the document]
GateBuilder: Yes you are right, driver works perfectly fine without them. However, CI did not like the code without them in my early commits, so I had to add these checks.  Thus, they should stay.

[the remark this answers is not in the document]
astafan8: hm... would it work if this check goes to the `__init__` and `mypy` is made happy in that one place? instead of duplicating this check in all the methods of the class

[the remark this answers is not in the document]
GateBuilder: Yes, I concur with your criticism of repeating code lines, however, moving it to ``__init__`` unfortunately won't work. I will keep this conversation open until you are satisfied, however, driver is ready to be merged.

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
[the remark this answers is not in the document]
GateBuilder: Alright, well I do not see the uselfulness of a context manager in this code, I should deal with this within a new PR, as this is out of scope. Recall that this code is not only for the time_trace but rather a legacy one and needed for ``_fast_sweep`` as well. So your naming scheme should be revised.

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -198,6 +339,15 @@ def __init__(self, parent: Instrument, name: str, channel: str) -> None:
                                      'current this will have no effect, set '
                                      '`sourcerange_i` instead',
                            vals=vals.Enum(*iranges[self.model]))
+
+        self.add_parameter('measure_autorange_i',
+                           label='current autorange',
+                           get_parser=float,
```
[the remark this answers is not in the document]
GateBuilder: Ok, ``get_parser`` is needed for the snapshot to be able to updated, correctly.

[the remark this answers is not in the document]
astafan8: > needed for the snapshot to be able to updated

really? snapshot supports boolean values, what was the issue exactly?

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -509,7 +506,7 @@ def _execute_lua(self, _script: list, steps: int) -> np.ndarray:
         """
         nplc = self.nplc()
         linefreq = self.linefreq()
-        _time_trace_extra_visa_timeout = self._time_trace_extra_visa_timeout()
+        _time_trace_extra_visa_timeout = self._extra_visa_timeout
         estimated_measurement_duration = 2*1000*steps*nplc/linefreq
         new_visa_timeot = (estimated_measurement_duration
```
WilliamHPNielsen: Note that there's a small typo here, `new_visa_timeot` -> `new_visa_timeout`.

GateBuilder: ahh..thanks.

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -74,11 +76,114 @@ def get_raw(self) -> np.ndarray:
...
+        data = self._time_trace()
+        return data
+
+
+class TimeAxis(Parameter):
+    """
+    A simple :class:`.Parameter` that holds all the times (relative to the
+    measurement start) at which the points of the time trace were acquired.
+    """
+    def __init__(self, name: str, instrument: Instrument, **kwargs) -> None:
```
astafan8: this init does nothing, hence can be removed.

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -172,24 +297,42 @@ def __init__(self, parent: Instrument, name: str, channel: str) -> None:
...
                                      'This affects the range and the '
                                      'precision of the source.',
                            vals=vals.Enum(*iranges[self.model]))
 
+        self.add_parameter('source_autorange_i_enabled',
+                            label='current source autorange',
+                            get_cmd=f'{channel}.source.autorangei',
+                            get_parser=float,
+                            set_cmd=f'{channel}.source.autorangei={{}}',
+                            docstring='Set autorange on/off for source voltage.',
```
astafan8: typo
```suggestion
                            docstring='Set autorange on/off for source current.',
```

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -311,33 +493,65 @@ def _fast_sweep(self, start: float, stop: float, steps: int,
                   'printbuffer(1, {}, {}.nvbuffer1.readings)'.format(steps,
                                                                      channel)]
 
-        self.write(self._parent._scriptwrapper(program=script, debug=True))
-        # we must wait for the script to execute
-        oldtimeout = self._parent.visa_handle.timeout
-        self._parent.visa_handle.timeout = 2*1000*steps*nplc/50 + 5000
+        return self._execute_lua(script, steps)
+
+    def _execute_lua(self, _script: list, steps: int) -> np.ndarray:
```
astafan8: let's make it more specific:
```suggestion
    def _execute_lua(self, _script: List[str], steps: int) -> np.ndarray:
```

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -311,33 +493,65 @@ def _fast_sweep(self, start: float, stop: float, steps: int,
...
+        returns the corresponding data from the buffer.
+
+        Args:
+            _script: The Lua script to be executed.
+            steps: Number of points.
+        """
+        nplc = self.nplc()
+        linefreq = self.linefreq()
+        _time_trace_extra_visa_timeout = self._extra_visa_timeout
+        estimated_measurement_duration = 2*1000*steps*nplc/linefreq
```
astafan8: analogous to `_extra_visa_timeout`, shall the `2` become an attribute? for example, `_measurement_duration_factor`.

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -311,33 +493,65 @@ def _fast_sweep(self, start: float, stop: float, steps: int,
...
         data = data[2:-1]
-
         outdata = np.array(list(struct.iter_unpack('<f', data)))
         outdata = np.reshape(outdata, len(outdata))
+        return outdata
 
-        self._parent.visa_handle.timeout = oldtimeout
+    def _set_sourcerange_v(self, val: float) -> None:
+        channel = self.channel
+        self.source_autorange_v_enabled.set(0)
```
astafan8:
```suggestion
        self.source_autorange_v_enabled(False)
```

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -311,33 +493,65 @@ def _fast_sweep(self, start: float, stop: float, steps: int,
...
+        self.write(f'{channel}.source.rangev={val}')
 
-        return outdata
+    def _set_measurerange_v(self, val: float) -> None:
+        channel = self.channel
+        self.measure_autorange_v_enabled.set(0)
```
astafan8:
```suggestion
        self.measure_autorange_v_enabled(False)
```

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -311,33 +493,65 @@ def _fast_sweep(self, start: float, stop: float, steps: int,
...
+        self.write(f'{channel}.measure.rangev={val}')
+
+    def _set_sourcerange_i(self, val: float) -> None:
+        channel = self.channel
+        self.source_autorange_i_enabled.set(0)
```
astafan8:
```suggestion
        self.source_autorange_i_enabled(False)
```

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -311,33 +493,65 @@ def _fast_sweep(self, start: float, stop: float, steps: int,
...
+        self.write(f'{channel}.source.rangei={val}')
 
+    def _set_measurerange_i(self, val: float) -> None:
+        channel = self.channel
+        self.measure_autorange_i_enabled.set(0)
```
astafan8:
```suggestion
        self.measure_autorange_i_enabled(False)
```

qcodes/tests/drivers/test_keithley_26xx.py
```diff
@@ -40,27 +42,85 @@ def test_smu_channels_and_their_parameters(driver):
 
         assert 'current' == smu.mode()
         smu.mode('voltage')
+        assert smu.mode() == 'voltage'
 
-        assert 'off' == smu.output()
-        smu.output('on')
+        assert 0 == smu.output()
```
astafan8: For all the parameters that use create_on_off_val_mapping, could you change the test code so that it asserts for `True`/`False`, for example, like this:
```suggestion
        assert True is smu.output()
```
This is needed so that when somebody reads the test it is clear that these parameters are intended for `True`/`False` values (and not for "0"/"0.0"/etc, although those also work because it's python and things like `True == 0.0` are true)

astafan8:
This PR requires https://github.com/QCoDeS/Qcodes/pull/1732 bug fix to be merged first.

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -224,6 +372,36 @@ def __init__(self, parent: Instrument, name: str, channel: str) -> None:
...
+                           label='Time resolution',
+                           unit='s',
+                           get_cmd=None,
+                           set_cmd=None)
+
+        self.add_parameter(name='time_axis',
+                           label='Time',
+                           unit='s',
+                           vals=vals.Arrays(shape=(self.timetrace_npts,)),
+                           parameter_class=TimeAxis)
```
WilliamHPNielsen: The `time_axis` parameter should not be included in the snapshot as it doesn't give any new information not covered by other parameters and also is potentially huge. I think passing `snapshot_value=False` to `add_parameter` should do the trick.

WilliamHPNielsen:
I am happy now. There's still a fair amount of work left to be done on this driver, but that is all beyond the scope of this PR which introduces two cool new features. 

I left one comment that we should address about the snapshot of the `time_axis`. Then it's merge time!

PR #1686, opened 2019-08-27
Improve requirements 

jenshnielsen:
* Add env markers for dataclasses in requirements.txt
* Matplotlib is a non optional dependency by now so add it to setup.py
* Add websockets to requirements.txt to align requirements.txt with setup.py

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,6 +12,7 @@ broadbean>=0.9.1
 wrapt
 pandas
 tqdm
-dataclasses  # this won't be needed when we move to python 3.7
+dataclasses;python_version<'3.7'  # this won't be needed when we move to python 3.7
+websockets==8.0.2
 tabulate
 applicationinsights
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -12,6 +12,7 @@ broadbean>=0.9.1
 wrapt
 pandas
 tqdm
-dataclasses  # this won't be needed when we move to python 3.7
+dataclasses;python_version<'3.7'  # this won't be needed when we move to python 3.7
+websockets==8.0.2
```
astafan8: does this need to be reflected in [`environment.yaml`](https://github.com/QCoDeS/Qcodes/blob/99f543dcd671b372a563bd6e8df938e18c3f02fb/environment.yml#L36) ?

jenshnielsen: Yes I think we should pin it there too

PR #1689, opened 2019-08-27
[WIP] KeysightB1500: Add ADC type, mode and averaging

lakhotiaharshit:
B1500 sampling is taking a long time. Hence user would like to have the parameter to set the averaging. 
Here I add a new parameter to do the same.

### DIFF
--- a/docs/examples/driver_examples/Qcodes example with Keysight B1500 Parameter Analyzer.ipynb
+++ b/docs/examples/driver_examples/Qcodes example with Keysight B1500 Parameter Analyzer.ipynb
@@ -515,7 +515,23 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Set SMU voltage"
+    "Set the operation mode and the setup parameter of the A/D converter (ADC) for each ADC type. By default the ADC is set tio 'high speed' type, mode is set set to 'manual' and the number of averaging is set to 1. "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "b1500.smu1.set_adc_type_mode_and_coeff()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Set the voltage"
    ]
   },
   {
@@ -1086,7 +1102,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.3"
+   "version": "3.6.9"
   },
   "nbsphinx": {
    "execute": "never"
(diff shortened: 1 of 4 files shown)


docs/examples/driver_examples/Qcodes example with Keysight B1500 Parameter Analyzer.ipynb
```diff
@@ -515,7 +515,23 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Set SMU voltage"
+    "Set the operation mode and the setup parameter of the A/D converter (ADC) for each ADC type. By default the ADC is set tio 'high speed' type, mode is set set to 'manual' and the number of averaging is set to 1. "
```
astafan8:
```suggestion
    "Set the operation mode and the setup parameter of the A/D converter (ADC) for each ADC type. By default the ADC is set to 'high speed' type, mode is set to 'manual' and the number of averaging is set to 1. "
```

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_sampling_measurement.py
```diff
@@ -1,7 +1,6 @@
 import warnings
```
astafan8: since there are only formatting changes, please revert the diff of this file

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1517A.py
```diff
@@ -284,3 +284,17 @@ def use_high_resolution_adc(self) -> None:
                    .aad(chnum=self.channels[0],
                         adc_type=AAD.Type.HIGH_RESOLUTION)
                    .message)
+
+    def set_adc_type_mode_and_coeff(self, *,
```
astafan8: this is a duplicate of [private method `_setup_integration_time` from `B1500` class ("mainframe")](https://github.com/QCoDeS/Qcodes/blob/99f543dcd671b372a563bd6e8df938e18c3f02fb/qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500.py#L138-L145). Hence:
1. it's only necessary to make it public (either by wrapping it in a new function or by just renaming it from "private" to "public")
2. since this function does not set anything that is directly(!) specific to SMU, it should not live in the SMU module driver. as per design of the actual instrument, one assigns particular ADC to a SMU, which establishes the linkage between settings of that ADC and that SMU. it does seem convenient to have this method in SMU module class, but it'm not comfortable with complicating the perception of an already complicated instrument.

lakhotiaharshit:
Closing this pull request because it rewrites the features already present in the driver. Instead of this, a new PR will be opened which will document the already available feature in this driver for the purpose of sampling measurement.

PR #1691, opened 2019-08-28
Fix docstrings in qcodes.instrument: visa, channel, base

GateBuilder:
This PR proposes some fixes in the docstrings of 

- qcodes.instrument.visa
- qcodes.instrument.channel
- qcodes.instrument.base

@WilliamHPNielsen

### DIFF
--- a/qcodes/instrument/channel.py
+++ b/qcodes/instrument/channel.py
@@ -16,19 +16,19 @@ class InstrumentChannel(InstrumentBase):
     Base class for a channel in an instrument
 
     Args:
-        parent (Instrument): the instrument to which this channel should be
-          attached
+        parent (Instrument): The instrument to which this channel should be
+          attached.
 
-        name (str): the name of this channel
+        name (str): The name of this channel.
 
     Attributes:
-        name (str): the name of this channel
+        name (str): The name of this channel.
 
         parameters (Dict[Parameter]): All the parameters supported by this
-          channel. Usually populated via ``add_parameter``
+          channel. Usually populated via ``add_parameter``.
 
         functions (Dict[Function]): All the functions supported by this
-          channel. Usually populated via ``add_function``
+          channel. Usually populated via ``add_function``.
     """
 
     def __init__(self,
@@ -89,7 +89,7 @@ class MultiChannelInstrumentParameter(MultiParameter):
     """
     Parameter to get or set multiple channels simultaneously.
 
-    Will normally be created by a ChannelList and not directly by anything
+    Will normally be created by a :class:`ChannelList` and not directly by anything
     else.
 
     Args:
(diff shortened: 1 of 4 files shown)


qcodes/instrument_drivers/Gentec/Gentec_Maestro.py
```diff
@@ -52,16 +52,17 @@ def zero_offset_get_parser(ans):
         # print connect message
         self.connect_message()
 
-    # get methods
-    def get_idn(self):
-        return {'vendor': 'Gentec', 'model': self.model, 'firmware': self.firmware_version}
+        # get methods
```
jenshnielsen: These changes do not look like they belong here ?

GateBuilder: The build for the docs were failing because of this. So I did the changes here. There is a problem then again. I will check it from the master branch.

WilliamHPNielsen: Sorry for just jumping in, but the added indentation is wrong. The methods `get_idn`, `clear_zero_offset`, `set_zero_offset`, and `_query_version` should be where they were, i.e. as bound methods of the class, not as methods inside the scope of `__init__`.

GateBuilder: Yes I corrected it now.

qcodes/instrument/channel.py
```diff
@@ -141,31 +142,31 @@ class ChannelList(Metadatable):
...
-            the object to be returned by the ChanneList's __getattr__ method.
-            Should be a subclass of MultiChannelInstrumentParameter.
+            the object to be returned by the ``__getattr__`` method of :class:`ChannelList`.
+            Should be a subclass of :class:`MultiChannelInstrumentParameter`.
 
     Raises:
-        ValueError: If chan_type is not a subclass of InstrumentChannel
-        ValueError: If multichan_paramclass if not a subclass of
-            MultiChannelInstrumentParameter (note that a class is a subclass
+        ValueError: If :class:`chan_type` is not a subclass of :class:`InstrumentChannel`
```
astafan8:
```suggestion
        ValueError: If ``chan_type`` is not a subclass of :class:`InstrumentChannel`
```

qcodes/instrument_drivers/Gentec/Gentec_Maestro.py
```diff
@@ -3,7 +3,7 @@
 
 
 class Gentec_Maestro(VisaInstrument):
-    """
+    r"""
```
astafan8: what is `r` needed for?

GateBuilder: In its most general form it is related to this issue:
 https://github.com/sphinx-doc/sphinx/issues/4479

In this case, I successfully traced back the error to the point where I removed an "r" in visa.py as I thought it was unnecessary. I set it back (see commit 7c24ccf) and that fixed it while I let this one also stay as I did not want to make another unnecessary commit and it does not do any harm.

jenshnielsen: So to  add a bit more technical detail. The r marks the string as raw this means that rather interpreting `\n` as a new line it will be the literal string `\n` This is needed for the docstring of visa since it contains the terminator docs. I don't think it would be needed here but indeed as long as you don't want to write any none escaped sequences it does no harm

PR #1692, opened 2019-08-29
Fix docstrings of qcodes.instrument: parameter, specialized_parameter, group_parameter

lakhotiaharshit:
Fixes documentation of user-facing api docs for parameters.

Files fixed:
qcodes.instrument.parameter
qcodes.instrument.group_parameter
qcodes.instrument.specialized_parameters


qcodes/instrument/parameter.py
```diff
@@ -1686,18 +1682,18 @@ def combine(*parameters, name, label=None, unit=None, units=None,
     Combine parameters into one sweepable parameter
 
     Args:
-        *parameters (qcodes.instrument.parameter.Parameter): the parameters to
-            combine
-        name (str): the name of the paramter
```
jenshnielsen: In the future please only remove types from the docstrings when they have been included in the function signatures

qcodes/instrument/parameter.py
```diff
@@ -1654,7 +1650,7 @@ def __init__(self, parameter, max_val_age=None):
 
     def get(self):
         """Return latest value if time since get was less than
-        `self.max_val_age`, otherwise perform `get()` and return result
+        `self.max_val_age`, otherwise perform ``get()`` and return result
```
astafan8: i guess the reference to the class attribute should be done properly, right? (but i don't remember how to do it)
```suggestion
        :attr:`max_val_age`, otherwise perform ``get()`` and return result
```

qcodes/instrument/group_parameter.py
```diff
@@ -79,9 +77,9 @@ class Group:
     via the same command. The command has to be a string, for example,
     a VISA command.
 
-    The `Group`'s methods are used within `GroupParameter` in order to
-    properly implement setting and getting of a single parameter in the
-    situation where one command sets or gets more than one parameter.
+    The :meth:`Group`'s methods are used within :class:`GroupParameter` in
```
astafan8: this is about "methods of Group class" (not about a method called "Group").
```suggestion
    The :class:`Group`'s methods are used within :class:`GroupParameter` in
```

lakhotiaharshit:
> the main reason to request changes is Jens' comment - please either "not remove the type annotations from the docstrings" or "if you prefer the type annotations from the docstrings to be removed, please add them to the function/method signatures".

I am summarizing our weekly (friday) discussion on this point for the sake of completeness and future reference: "In this PR we will remove the type annotations if they are already in the function signature. If they are not in the signature we will keep the documentation as it is and not try to add the types in main code because that needs lot more effort which is beyond the purview of this PR"

qcodes/instrument/specialized_parameters.py
```diff
@@ -1,7 +1,7 @@
 """
-Module for specialized parameters. The qcodes.instrument.parameters module
-provides generic parameters for different generic cases. This module provides
-useful/convenient specializations of such generic parameters.
+Module for specialized parameters. The :class:`qcodes.instrument.parameters`
```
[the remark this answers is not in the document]
lakhotiaharshit: Done

qcodes/instrument/parameter.py
lakhotiaharshit: I also could not find the proper way to refer to the instance attribute.

using :attr: gives error: 
qcodes.instrument.parameter.GetLatest.get:1: WARNING: py:attr reference target not found: max_val_age

At the moment I have kept as it is. Just removed the self from the attribute.

qcodes/instrument/group_parameter.py
lakhotiaharshit: Corrected this.

jenshnielsen:
As far as I can see this still drops the types from some docstrings where the types have not yet been added to the function signature. Do we want to revert this?

astafan8:
> As far as I can see this still drops the types from some docstrings where the types have not yet been added to the function signature. Do we want to revert this?

yes. either revert, or add types to function signature (but only if it's not a big effort)

lakhotiaharshit:
> As far as I can see this still drops the types from some docstrings where the types have not yet been added to the function signature. Do we want to revert this?

Thanks for noticing. I had missed them. I am working on it now.

PR #1694, opened 2019-09-04
Alazar: refactor `AcquisitionInterface` out of `AcquisitionController`

Dominik-Vogel:
The Alazar `AcquistionController` is an interface for the alazar ats instrument, that takes care of handling the buffer data. In the current implementation this interface is a qcodes instrument.

In my opinion this is an unnecessary coupling between pure number crunching code and qcodes instruments. I therefore suggest to refactor an `AcquisitionInterface` out of the `AcquisitionController`, which only contains the interface methods.
By letting the `AcquistionController` inherit from `AcquistionInterface` and `Instrument` the original functionality should stay untouched.

Additionally I removed the `NotImplementedError`s from some of the methods, which don't need to be implemented.

### DIFF
--- a/qcodes/instrument_drivers/AlazarTech/ATS.py
+++ b/qcodes/instrument_drivers/AlazarTech/ATS.py
@@ -866,7 +866,7 @@ class Buffer:
                 'Memory should have been released before buffer was deleted.')
 
 
-class AcquisitionController(Instrument):
+class AcquisitionInterface:
     """
     This class represents all choices that the end-user has to make regarding
     the data-acquisition. this class should be subclassed to program these
@@ -889,22 +889,9 @@ class AcquisitionController(Instrument):
         _alazar: a reference to the alazar instrument driver
     """
 
-    def __init__(self, name, alazar_name, **kwargs):
-        """
-        Args:
-            alazar_name: The name of the alazar instrument on the server
-        """
-        super().__init__(name, **kwargs)
-        self._alazar = self.find_instrument(alazar_name,
-                                            instrument_class=AlazarTech_ATS)
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
 
-    def _get_alazar(self):
-        """
-        returns a reference to the alazar instrument. A call to self._alazar is
-        quicker, so use that if in need for speed
-        :return: reference to the Alazar instrument
-        """
-        return self._alazar
 
     def pre_start_capture(self):
         """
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -889,38 +889,23 @@ class AcquisitionController(Instrument):
...
 
-    def __init__(self, name, alazar_name, **kwargs):
-        """
-        Args:
-            alazar_name: The name of the alazar instrument on the server
-        """
-        super().__init__(name, **kwargs)
-        self._alazar = self.find_instrument(alazar_name,
-                                            instrument_class=AlazarTech_ATS)
+    def __init__(self, *args, **kwargs):
```
astafan8: this is not needed, right?

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -866,7 +866,7 @@ def __del__(self):
                 'Memory should have been released before buffer was deleted.')
 
 
-class AcquisitionController(Instrument):
+class AcquisitionInterface:
     """
     This class represents all choices that the end-user has to make regarding
     the data-acquisition. this class should be subclassed to program these
```
astafan8: shall `acquisitioncontroller` be substrituded with `acquisitioninterface` in this docstring? (see below this line)

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -889,38 +889,23 @@ class AcquisitionController(Instrument):
         _alazar: a reference to the alazar instrument driver
```
astafan8: should be removed as the interface doesn't have this attribute

jenshnielsen: One should not document private attributes anyway

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -962,3 +947,44 @@ def buffer_done_callback(self, buffers_completed):
...
+    def __init__(self, name, alazar_name, **kwargs):
+        """
+        Args:
+            alazar_name: The name of the alazar instrument on the server
+        """
+        super().__init__(name, **kwargs)
+        self._alazar = self.find_instrument(alazar_name,
+                                            instrument_class=AlazarTech_ATS)
+
+    def _get_alazar(self):
```
jenshnielsen: We should really deprecate this. But that is for another time

Dominik-Vogel: fully agree

qcodes/instrument_drivers/AlazarTech/ATS.py
Dominik-Vogel: Oh yes, I forgot the doc string, thanks!

qcodes/instrument_drivers/AlazarTech/ATS.py
Dominik-Vogel: I think it might be, when using the interface as a mixin.

PR #1695, opened 2019-09-04
dev / add logging to measurement runner __exit__

WilliamHPNielsen:
This PR does what @astafan8 suggested in https://github.com/QCoDeS/Qcodes/pull/1668 😄 

- Add a logging line to the `__exit__` method of the `Measurement` `Runner`
- Also log exceptions, in case there were any

I opted for just stuffing in the traceback, as I think human-readability is probably to be favoured over some structured view of the error chain. We can change or extend this later as needed.

@QCoDeS/core


qcodes/dataset/measurements.py
```diff
@@ -639,10 +641,21 @@ def __exit__(self, exception_type, exception_value, traceback) -> None:
...
+        if exception_type:
+            # if an exception happened during the measurement,
+            # log the exception
+            stream = io.StringIO()
+            tb_module.print_exception(exception_type,
+                                      exception_value,
+                                      traceback,
+                                      file=stream)
+            log.warning('An exception occured in measurement with guid: '
+                        f'{self.ds.guid}; Traceback: {stream.getvalue()}')
```
astafan8: newlines for readability? :)
```suggestion
                        f'{self.ds.guid};\nTraceback:\n{stream.getvalue()}')
```

PR #1699, opened 2019-09-06
Yokogawa GS200: Exclude/Include snapshot depending on mode

lakhotiaharshit:
Yokogawa GS200 can be a voltage source or current source. When in one mode the parameters of another mode can be excluded from the snapshot. 

Things to do: 
- [x] Test with a real instrument.


qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -352,6 +357,21 @@ def _ramp_source(self, ramp_to: float, step: float, delay: float) -> None:
         self.output_level.step = saved_step
         self.output_level.inter_delay = saved_inter_delay
 
+    def _exclude_snapshot(self, mode: str) -> bool:
```
astafan8: it seems to me that this method is not really needed if it's only used at init of the instrument. would you agree?

lakhotiaharshit: @astafan8 I have removed the _exclude_snapshot method and excluded current and current_range from the snapshot at init.

jenshnielsen:
I would not be against adding a comment above the changes to the snapshot explaining why

lakhotiaharshit:
> I would not be against adding a comment above the changes to the snapshot explaining why

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -469,9 +478,17 @@ def _set_source_mode(self, mode: str) -> None:
...
+            self.voltage_range.snapshot_exclude = False
+            self.voltage.snapshot_exclude = False
+            self.current_range.snapshot_exclude = True
+            self.current.snapshot_exclude = True
         else:
             self.range = self.current_range
             self.output_level = self.current
+            self.voltage_range.snapshot_exclude = True
+            self.voltage.snapshot_exclude = True
+            self.current_range.snapshot_exclude = False
```
Dominik-Vogel: @lakhotiaharshit I have a suspicion that setting the snapshot to True, will require getting the parameter, or at least filling the parameter cache with a known value.

Dominik-Vogel: Could you check up on this?

astafan8: @Dominik-Vogel I don't think that `require` is the correct word here.

> setting the snapshot to True, will require getting the parameter, or at least filling the parameter cache with a known value

Did you mean to say that when the parameter gets "switched" from `.snapshot_exclude = False` to `.snapshot_exclude = True` (because the instrument mode is switched between voltage/current modes), it's probably useful that the value of that parameter gets updated immediately such that value is "valuable" for use, for example, when `snapshot(..)` is requested?

i also think that this driver generally needs improvement for proper representation of the fact that it can operate in these two modes. for example, why it is the case that when parameter is in current mode the user still has the voltage parameter available on the instrument and it's get/set don't throw errors saying "you're not in a voltage mode hence you can use the voltage parameter". So i think that this particular PR **quick-fixed** the snapshotting of the instrument, and that's it.

Dominik-Vogel: Mikhail, I think we do agree in general. Let me though elaborate a little on why I used the stronger term `require`: As far as I understand we do have the policy, that the station snapshot contains the current state of the station. Or maybe even: the station snapshot contains the current state of the station at any time. For this reason we do read all the parameter values on adding an instrument to the station. Now there are exceptions, marked with `snapshot_exclude` to cover the case of parameters that are not definable at the given time. 
Therefore I think that when changing the validity of parameters should update them in the snapshot automatically.
(I just commented on this PR because this fix showed the problem more clearly, I have no critique for the PR itself, but just wanted to get Harshits opinion, who probably has thought about it a lot more that I have)

PR #1700, opened 2019-09-09
Fix docstrings: math.field_vector, monitor.monitor and .station

GateBuilder:
This PR proposes fixes for the docstrings in 

- [x] qcodes.math.field_vector
- [x] qcodes.monitor.monitor
- [x] qcodes.station

@astafan8


qcodes/math/field_vector.py
```diff
@@ -1,11 +1,3 @@
-"""
```
WilliamHPNielsen: Note that we strongly encourage the use of module docstrings. I agree that this particular docstring should be moved to the class, but it's worth thinking about putting something else in here. It can be as brief as the one here: https://github.com/QCoDeS/Qcodes/blob/master/qcodes/utils/types.py

qcodes/math/field_vector.py
```diff
@@ -1,3 +1,8 @@
+"""
+A convenient class to keep track of vectors representing physical fields.
```
WilliamHPNielsen: Sorry to be such a nitpicker, but the module docstring should document the module, which is not a class. So smth like "A helper module containing a class to keep track of vectors in different coordinate systems" would be more on point.

qcodes/math/field_vector.py
```diff
@@ -7,11 +12,11 @@
 
 class FieldVector(object):
     """
-    A convenient class to keep track of vectors representing physical fields. The
-    idea is that a vector instance stores a representation in Cartesian, spherical
-    and cylindrical coordinates. Giving either (x, y, z) values or (rho, phi, z)
-    values or (r, theta, phi) values at instantiation, we will calculate the other
-    representation immediately.
+    The idea is that a vector instance stores a representation in Cartesian,
```
WilliamHPNielsen: I think it's still good to have "A class representing X" in the docstring. But that may be a matter of individual preference.

qcodes/station.py
```diff
@@ -112,22 +111,23 @@ def snapshot_base(self, update: bool = True,
...
-        State of the station as a JSON-compatible dict (everything that
-        the custom JSON encoder class :class:'qcodes.utils.helpers.NumpyJSONEncoder'
+        State of the station as a JSON-compatible dictionary (everything that
+        the custom JSON encoder class :class:`qcodes.utils.helpers.NumpyJSONEncoder`
         supports).
 
-        Note: in the station contains an instrument that has already been
-        closed, not only will it not be snapshotted, it will also be removed
+        Note: If the station contains an instrument that has already been
+        closed, not only it will not be snapshotted, it will also be removed
```
WilliamHPNielsen: I think the "will it" was correct.

qcodes/math/field_vector.py
```diff
@@ -191,12 +197,12 @@ def set_component(self, **new_values):
...
             # (x, y, z) values are recomputed, as is the rho coordinate.
             # Internally we arrange this by setting x, y, z and rho to None
-            # and calling self._compute_unknowns()
+            # and calling self._compute_unknowns().
             >>> f.set_component(r=10)
 
         Args:
             new_values (dict): keys representing parameter names and values the
-                values to be set
+                               values to be set.
```
WilliamHPNielsen: A question that probably reveals my ignorance about sphinx: why is this extra indentation needed?

jenshnielsen: As far as I know it's not https://www.sphinx-doc.org/en/1.5/ext/example_google.html

WilliamHPNielsen: I see. Should we have a standard for indentation? We should probably follow `Sections are created with a section header and a colon followed by a block of indented text.` from that link.

WilliamHPNielsen: Or rather:
```
 The format for a parameter is::

        name (type): description
            The description may span multiple lines. Following
            lines should be indented. The "(type)" is optional.

            Multiple paragraphs are supported in parameter
            descriptions.
```

WilliamHPNielsen: I guess what I'm trying to say is that
```
            new_values (dict): keys representing parameter names and values the
                values to be set
```
is to be preferred over
```
            new_values (dict): keys representing parameter names and values the
                               values to be set.
 ```

Do you agree? Do we care?

jenshnielsen: In this case I believe the default is and should be indent by 4 spaces

astafan8: > In this case I believe the default is and should be indent by 4 spaces

+1

PR #1701, opened 2019-09-09
Block keyboard interrupts in critical parts of the code.

jenshnielsen:
Mainly when interacting with visa or writing to the database

Missing

- [X] Is there a good way to test this automatically. Not really. The failed attempt is in https://github.com/jenshnielsen/Qcodes/tree/block_interrupt_with_tests
- [x] Performance implications. Performance implications are acceptable see below
- [x] Document measurement interrupt
- [x] Document how to protect your own code e.g. driver


qcodes/dataset/measurements.py
```diff
@@ -146,30 +147,30 @@ def add_result(self, *res_tuple: res_type) -> None:
         # of all parameters. This also allows users to call
         # add_result with the arguments in any particular order, i.e. NOT
         # enforcing that setpoints come before dependent variables.
+        with DelayedKeyboardInterrupt():
```
jenshnielsen: This may not be needed as long as the flush method is protected

jenshnielsen:
There is also a small chance that a keyboard interrupt could happen while setting up the table for a run. That should probably be protected too

jenshnielsen:
Or perhaps the atomic context manger should just block keyboard interrupts

astafan8:
> Is there a good way to test this automatically

allow me to suggest something (don't know if it's good or not):
* for unit test i can propose the following draft:
```python
exception_to_raise = KeyboardInterrupt('from inside')
exception_caught = False
a = []
try:
        a.append('before but inside')
        raise exception_to_raise
        a.append('after but inside')
    a.append('after but outside')
except KeyboardInterrupt as e:
    exception_caught = True
    assert e == exception_to_raise  # or similar
finally:
    assert exception_caught
    assert a == ['before but inside', 'after but inside']  # ... so that 'after but outside' is not in
```
* for more interesting tests, i'd suggest a combination of the "unit test idea" with mocking something inside `with DelayedKeyboardInterrupt():` so that that something throws `KeyboardInterrupt`.

astafan8:
> Performance implications. This seems to be a bit slower than without. Can it be rewritten to not create as many contexts

is there any info on how much is "a bit"? if it's in the range of ~3% for visa communication - i wouldn't worry much, the benefit is more important. for the data saving and other sqlite queries - that perhaps needs decisions on a case by case basis: i believe that `add_results` should be inside one such context manager for the greater good, and for other parts perhaps less so.

astafan8:
my only concern about this idea is the fact that `KeyboardInterrupt` itself is overridden - i'd expect that `KeyboardInterrupt` will remain as an "emergency stop", and there is some other `Ctrl+Q` (or else) that performs the "safe interrupt". My thinking is: what if, say, during a for-some-reason long visa communication the user decided to interrupt everything because of an emergency - i doubt that in this situation he would want to wait until the visa command returns etc. If `KeyboardInterrupt` itself is overridden, the user has no means of interrupting the execution until that context manager is exited (well, other than shutting down python kernel in some hacky way, as opposed to just `Ctrl+C`). Alternatively, we could also consider implementing "single `Ctrl+C`" as "safe interrupt" and "more than one `Ctrl+C`" as "standard keyboard interrupt", but this sounds more complicated to implement than ~`Ctrl+Q`.

or are these worries of mine solved in some way that i failed to see just yet?

jenshnielsen:
It might be worth considering taking inspiration from how things work in Jupyter notebook with regards to double ctrl-c to interrupt. The relevant code is notebook/notebookapp.py:_confirm_exit

jenshnielsen:
@astafan8  I added an option to trigger immediately by doing a second interrupt. 
You can test this with a stupid example like the following


```
from qcodes.utils.delaykeyboardinterrupt import DelayedKeyboardInterrupt
    import time
    for i in range(20):
        time.sleep(0.1)
        print(i)
    print("done looping")
print("And after")
```

This seems to work well in both console jupyter notebook and spyder with the one exception that you cannot trigger a second keyboard interrupt in spyder by pressing the button (its grayed out) but you can trigger both the first and second one with a ctrl-c

jenshnielsen:
More or less failed attempts of writing tests for this are on https://github.com/jenshnielsen/Qcodes/tree/block_interrupt_with_tests

jenshnielsen:
Doing some simple benchmarking 

```
%%timeit
for i in range(1000):
    pass
17.8 µs ± 925 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

```

```
%%timeit
for i in range(1000):
        pass
23.3 ms ± 2.26 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

```
%%timeit
for i in range(1000):
            pass
28.7 ms ± 3.99 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

The cost of the context manager is about 2.5 us. I think that is acceptable given that its meant to wrap around database commits and network communication that is much slower. The extra cost of nesting is not significant

lakhotiaharshit:
> from qcodes.utils.delaykeyboardinterrupt import DelayedKeyboardInterrupt
> with DelayedKeyboardInterrupt():
>     import time
>     for i in range(20):
>         time.sleep(0.1)
>         print(i)
>     print("done looping")
> print("And after")

@Mikha

> @astafan8  I added an option to trigger immediately by doing a second interrupt.
> You can test this with a stupid example like the following
> from qcodes.utils.delaykeyboardinterrupt import DelayedKeyboardInterrupt
> with DelayedKeyboardInterrupt():
>     import time
>     for i in range(20):
>         time.sleep(0.1)
>         print(i)
>     print("done looping")
> print("And after")
> 
> This seems to work well in both console jupyter notebook and spyder with the one exception that you cannot trigger a second keyboard interrupt in spyder by pressing the button (its grayed out) but you can trigger both the first and second one with a ctrl-c

@astafan8, @jenshnielsen I tried manually testing the keyboard interrupts and everything works fine. I tried two cases: 1) Using the above code. Both single and double interrupt works in jupyter notebook on a windows machine. 2) I tried to interrupt while data was being added to the database. This was the case where the user would lock their database. Here also both the interrupts work as expected. Single interrupt never left the database locked.

jenshnielsen:
TODO

- [ ] Document measurement interrupt
- [ ] Document how to protect your own code e.g. driver

jenshnielsen:
@lakhotiaharshit  Thanks for testing :)

jenshnielsen:
@QCoDeS/core I think this is ready for review now

docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb
```diff
@@ -513,6 +513,61 @@
...
+   "source": [
+    "## Handling interruption of measurements\n",
+    "\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "A QCoDeS driver should be prepared for interruptions of the measurement triggered by a KeyBoardInterrupt from the enduser. \n",
```
astafan8:
```suggestion
    "A QCoDeS driver should be prepared for interruptions of the measurement triggered by a KeyboardInterrupt from the enduser. \n",
```

qcodes/utils/delaykeyboardinterrupt.py
```diff
@@ -0,0 +1,41 @@
...
+    KeyboardInterrupt immediately.
+
+    Inspired by https://stackoverflow.com/questions/842557/how-to-prevent-a-block-of-code-from-being-interrupted-by-keyboardinterrupt-in-py
+    """
+    signal_received = None
+    old_handler = None
+
+    def __enter__(self) -> None:
+        if signal.getsignal(signal.SIGINT) is signal.default_int_handler:
+            self.old_handler = signal.signal(signal.SIGINT, self.handler)
```
astafan8: which `self.handler` is meant here? perhaps `self.old_handler`?
```suggestion
            self.old_handler = signal.signal(signal.SIGINT, self.old_handler)
```

PR #1706, opened 2019-09-10
Fix docstrings qcodes.utils: comand, deprecate, helpers, magic and metadata

GateBuilder:
This PR proposes fixes for the docstrings of:

- qcodes.utils.command
- qcodes.utils.deprecate
- qcodes.utils.helpers
- qcodes.utils.magic
- qcodes.utils.metadata

@astafan8

### DIFF
--- a/qcodes/utils/helpers.py
+++ b/qcodes/utils/helpers.py
@@ -29,29 +29,29 @@ log = logging.getLogger(__name__)
 class NumpyJSONEncoder(json.JSONEncoder):
     """
     This JSON encoder adds support for serializing types that the built-in
-    `json` module does not support out-of-the-box. See the docstring of the
-    `default` method for the description of all conversions.
+    ``json`` module does not support out-of-the-box. See the docstring of the
+    ``default`` method for the description of all conversions.
     """
 
     def default(self, obj):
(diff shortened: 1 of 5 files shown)


qcodes/utils/helpers.py
```diff
@@ -88,7 +89,7 @@ def default(self, obj):
 
 
 def tprint(string, dt=1, tag='default'):
-    """ Print progress of a loop every dt seconds """
+    """ Print progress of a loop every dt seconds. """
```
astafan8:
```suggestion
    """Print progress of a loop every ``dt`` seconds."""
```

qcodes/utils/helpers.py
```diff
@@ -354,8 +359,8 @@ def __exit__(self, type, value, tb):
 
 def make_unique(s, existing):
     """
-    make string s unique, able to be added to a sequence `existing` of
-    existing names without duplication, by appending _<int> to it if needed
+    Make string `s` unique, able to be added to a sequence `existing` of
```
astafan8:
```suggestion
    Make string ``s`` unique, able to be added to a sequence ``existing`` of
```

PR #1707, opened 2019-09-10
Alazar Simulator

Dominik-Vogel:
In this PR a simulator for the Alazar cards gets added. It is far from complete but enables simple testing.

An example can be found in the test.

### DIFF
--- a/qcodes/instrument/mockers/simulated_ats_api.py
+++ b/qcodes/instrument/mockers/simulated_ats_api.py
@@ -1,0 +1,112 @@
+"""
+This module provides an api that can be passed instead for the
+:class:`qcodes.instrument_drivers.AlazarTech.ats_api.AlazarATSAPI`
+, which does not require a real alazar card present in the system.
+This mock does currently represent a ATS9360 and does not provide
+any functionality whatsoever.
+"""
+
+
+from typing import Tuple, Callable, Optional
+import numpy as np
+import ctypes
+
+from qcodes.instrument_drivers.AlazarTech.dll_wrapper import (
+    _mark_params_as_updated, ReturnCode)
+from qcodes.instrument_drivers.AlazarTech.constants import (
+    Capability)
+from qcodes.instrument_drivers.AlazarTech.ats_api import AlazarATSAPI
+
+
+class SimulatedAlazarATSAPI(AlazarATSAPI):
+
+    registers = {
+        8: 70254688,
+        58: np.uint32(1 << 26)  # Trigger hold off
+    }
+
+    def __init__(
+            self,
+            dll_path: str = None,  # Need this for meta super class
+            buffer_generator: Optional[Callable[[np.ndarray], None]] = None,
+            dtype: type = np.uint16
+    ):
+        # don't call `super().__init__` here to avoid dependence on the
+        # alazar driver, when loading the dll
+        self._buffer_generator = (
+            buffer_generator or self._default_buffer_generator)
+        self.dtype = dtype
+        self.buffers = {}
+
+    def _sync_dll_call(self, c_name: str, *args):
+        _mark_params_as_updated(*args)
+
+    def get_board_by_system_id(self, system_id: int, board_id: int) -> int:
+        # This is the handle....
+        return 12335
+
+    def busy(self, handle: int) -> int:
+        return 0
+
+    # OTHER API-RELATED METHODS
+
+    def get_board_model(self, handle: int) -> str:
+        return 'ATS9360'
+
+    def get_channel_info_(self, handle: int) -> Tuple[int, int]:
+        return 4294967294, 12
+
+    def get_cpld_version_(self, handle: int) -> str:
+        return "25.16"
+
+    def get_driver_version_(self) -> str:
+        return '6.5.1'
+
+    def get_sdk_version_(self) -> str:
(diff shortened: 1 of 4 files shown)


qcodes/instrument/mockers/simulated_ats_api.py
```diff
@@ -0,0 +1,112 @@
...
+            buffer: ctypes.c_void_p,
+            timeout_in_ms: int
+    ) -> ReturnCode:
+        b = self.buffers.get(buffer.value, None)
+        assert b is not None
+        self._buffer_generator(b)
+        return self._sync_dll_call(
+            'AlazarWaitAsyncBufferComplete', handle, buffer, timeout_in_ms)
+
+    @staticmethod
```
astafan8: why is it a static method and not just a function?

qcodes/instrument/mockers/simulated_ats_api.py
```diff
@@ -0,0 +1,112 @@
...
+
+    registers = {
+        8: 70254688,
+        58: np.uint32(1 << 26)  # Trigger hold off
+    }
+
+    def __init__(
+            self,
+            dll_path: str = None,  # Need this for meta super class
+            buffer_generator: Optional[Callable[[np.ndarray], None]] = None,
```
astafan8: yes! this is the king :)

qcodes/tests/mockers/test_simulated_ats_api.py
```diff
@@ -0,0 +1,63 @@
...
+
+@pytest.fixture(scope='function')
+def simulated_alazar():
+    driver = ATS9360(
+        'Alazar',
+        api=SimulatedAlazarATSAPI(dll_path='simulated'))
+    with driver.syncing():
+        driver.sample_rate(1_000_000_000)
+        driver.decimation(1)
+        driver.trigger_delay(0)
```
astafan8: is setting decimation and trigger_delay really needed? when i was using a similar simulation class a couple of days ago, only the sample_rate was needed to be set, and that's it.

qcodes/tests/mockers/test_simulated_ats_api.py
```diff
@@ -0,0 +1,63 @@
...
+    yield TestAcquisitionController()
+
+
+def test_simulated_alazar(simulated_alazar, alazar_ctrl):
+    alazar = simulated_alazar
+    alazar.aux_io_mode('AUX_IN_TRIGGER_ENABLE')
+    alazar.aux_io_param('TRIG_SLOPE_POSITIVE')
+    alazar.sync_settings_to_card()
+
+    alazar.acquire(
```
astafan8: can we add an assertion of what the call to acquire returns? this would really show that the simulator works and also would test that the default buffer generator works.

qcodes/tests/mockers/test_simulated_ats_api.py
```diff
@@ -0,0 +1,63 @@
...
+        def post_acquire(self):
+            return self.buffers
+
+    yield TestAcquisitionController()
+
+
+def test_simulated_alazar(simulated_alazar, alazar_ctrl):
+    alazar = simulated_alazar
+    alazar.aux_io_mode('AUX_IN_TRIGGER_ENABLE')
+    alazar.aux_io_param('TRIG_SLOPE_POSITIVE')
```
astafan8: is setting these two parameters needed?

qcodes/tests/mockers/test_simulated_ats_api.py
Dominik-Vogel: Good point, I will remove it.

qcodes/tests/mockers/test_simulated_ats_api.py
Dominik-Vogel: nope, all that came from some code I used with a real alazar. I removed it now.

qcodes/instrument/mockers/simulated_ats_api.py
Dominik-Vogel: This is probably a matter of taste. 
My thought was: it only ever gets used from within the class, and to improve readability, I like to define things on the most restrictive scope possible. Following that idea it should be a local function inside the constructor.
What are your thoughts? Why should this live on a module level?

qcodes/tests/mockers/test_simulated_ats_api.py
Dominik-Vogel: Yes that is a good suggestion!

qcodes/instrument/mockers/simulated_ats_api.py
astafan8: > define things on the most restrictive scope possible

that's perfect, i agree with that. i don't like local function because it's impossible to test them by themselves.

my general thoughts were that the buffer generator is an interface kind of thing that the simulator uses (or expects from outside) rather than something that it "is capable of providing". Hence providing a default from outside of the class sounds logical as opposed to have the default defined inside the class.

ok. since it's anyway 'private' to this class/module, let's keep it as it. but big thanks for explaining the thoughts!

qcodes/tests/mockers/test_simulated_ats_api.py
astafan8: awesome, good catch!

PR #1710, opened 2019-09-11
Remove lower-case keysight folder

### LINKED ISSUE #1709: Filename collision due to case-sensitivity in Keysight folder
When pulling the qcodes repository on Windows, there is a filename collision between the uppercase and lowercase Keysight folders.

The error message is as follows:
```bash
$ git clone https://github.com/QCoDeS/Qcodes.git
Cloning into 'Qcodes'...
remote: Enumerating objects: 1522, done.
remote: Counting objects: 100% (1522/1522), done.
remote: Compressing objects: 100% (655/655), done.

Receiving objects: 100% (112398/112398), 242.65 MiB | 2.58 MiB/s, done.
Resolving deltas: 100% (87395/87395), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'qcodes/instrument_drivers/Keysight/__init__.py'
  'qcodes/instrument_drivers/keysight/__init__.py'
```

I propose we remove the lowercase keysight folder as it has now been deprecated for over 2 years.

### System
Windows/OSX and other OS's with case insensitive file systems.

spauka:
Closes #1709.

- Removes the lower case keysight folder, which has been deprecated for over 2 years. Fixes a warning when pulling on Windows/OSX.

@jenshnielsen


jenshnielsen:
@spauka The reason we have not removed the folder is that doing so gives a conflict for any user doing a git pull on a case insensitive file system. This was never intended as a deprecation cycle

jenshnielsen:
If we can find a clean way to remove it. Perhaps because new versions of git has fixed that issue I am all for it

jenshnielsen:
That being said at least we can remove the `__init__.py` file from the lower case folder. That should fix the warning i guess?

jenshnielsen:
More specifically the issue is that when I switch to your branch I see

```
(base) PS C:\Users\jenielse\source\repos\Qcodes> git status                                                                                                                                                                                                                                                                                                               On branch fix/case-sensitive-folders                                                                                                                                                                                                                                                                                                                                      Your branch is up to date with 'spauka/fix/case-sensitive-folders'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        deleted:    qcodes/instrument_drivers/Keysight/__init__.py
```
E.g the `__init__.py` file in the upper case folder has been deleted and I need to unstage that change manually. This will be the case for all users with a local clone on a case insensitive OS as far as I can see

spauka:
Oh haha, wow that's gross... I guess I got hints of this problem when I had to use a linux box to even make the patch 😛

I'll have to check tomorrow that this doesn't screw things up on Windows but renaming the folder before deleting has been suggested to work.

jenshnielsen:
@spauka Yes gross indeed. If you can find a way to make it work that would be much appreciated. As you can see we eventually gave up :(

spauka:
OK, I all my ideas failed 😢, so I'm going to close this for now. I'll reopen if I can find a solution.

jenshnielsen:
:( Thanks for trying

PR #1712, opened 2019-09-11
CI: Link commands in all scripts by their exit code, use mostly 'bash'

astafan8:
There have been cases where within a `bash` or `script` step one command would fail with non-0 exit code but the command after that would continue to execute. This is not desired because, for example, we don't want to run `mypy` if the `"qcodes"` environment wasn't activated.

In this PR, we use `&&` to glue commands together so that non-0 exit code prevents continuation of execution of a command sequence. Moreover, most of the `script` steps got converted to `bash` steps.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -45,53 +45,53 @@ jobs:
         displayName: "Add conda to PATH on Linux"
         condition: eq( variables['Agent.OS'], 'Linux' )
       - script: |
-          CALL conda update -n base conda -y
+          CALL conda update -n base conda -y &&
           CALL conda init --all
         displayName: "Conda setup on Windows"
         condition: eq( variables['Agent.OS'], 'Windows_NT' )
       - bash: |
-          sudo conda update -n base conda -y
+          sudo conda update -n base conda -y &&
           sudo conda init bash
         displayName: "Conda setup on Linux"
         condition: eq( variables['Agent.OS'], 'Linux' )
       - bash: |
-          conda --version
-          conda env create --file environment.yml
-          source activate qcodes
-          pip install -r test_requirements.txt
-          pip install -r docs_requirements.txt
+          conda --version &&
+          conda env create --file environment.yml &&
+          source activate qcodes &&
+          pip install -r test_requirements.txt &&
+          pip install -r docs_requirements.txt &&
           pip install -e .
         displayName: "Install environment, qcodes"
       - bash: |
-          source activate qcodes
+          source activate qcodes &&
           mypy qcodes
         displayName: "mypy"
       - bash: |
-          source activate qcodes
-          cd ..
-          git clone https://github.com/QCoDeS/qcodes_generate_test_db.git
-          cd qcodes_generate_test_db
-          python generate_version_0.py
-          python generate_version_1.py
-          python generate_version_2.py
-          python generate_version_3.py
-          python generate_version_4a.py
-          python generate_version_4.py
-          python generate_version_5.py
-          python generate_version_6.py
+          source activate qcodes &&
+          cd .. &&
+          git clone https://github.com/QCoDeS/qcodes_generate_test_db.git &&
+          cd qcodes_generate_test_db &&
+          python generate_version_0.py &&
+          python generate_version_1.py &&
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -44,54 +44,54 @@ jobs:
...
+          python generate_version_6.py &&
           python generate_version_7.py
         displayName: "Generate db fixtures"
         condition: succeededOrFailed()
-      - script: |
-          CALL C:\Miniconda\condabin\conda_hook.bat
-          CALL conda activate qcodes
-          cd qcodes
+      - bash: |
+          C:\Miniconda\condabin\conda_hook.bat &&
```
jenshnielsen: Running a bat script from bash cannot work

jenshnielsen: For bash we are still using the old fashioned way of activating the env with source. 
Running the bash script that enables `conda init` in a platform independent way seems more complicated than it should be as the location is platform dependent and you need sudo on linux but sudo is not available in git bash on windows

astafan8: > Running a bat script from bash cannot work
yep, learned it the hard way. also that conda_hook.bat is not meant for bash... so, lack-of-knowledge everywhere :)

as for `conda init` - such a shame :( but ok. hope the `source activate` will be allowed for some time by conda devs.

azure-pipelines.yml
```diff
@@ -107,20 +107,20 @@ jobs:
         inputs:
           codeCoverageTool: Cobertura
           summaryFileLocation: '$(System.DefaultWorkingDirectory)/qcodes/coverage.xml'
-      - script: |
-          CALL C:\Miniconda\condabin\conda_hook.bat
-          CALL conda activate qcodes
-          cd docs
-          REM Turn warnings into errors
-          set SPHINXOPTS=-W -v
```
jenshnielsen: I think we should leave this block as a cmd or powerscript. We need to test the make.bat file and that cannot run in bash

astafan8: what if i do `./make.bat`? according to google search it will execute the bat file in cmd

jenshnielsen: I assume you mean bash and not cmd? Is set valid in bash I would assume not

astafan8: ah, sorry, i completely misunderstood your comment, i thought you were referring to `make.bat` call.

indeed, instead of `set` it should be `export`.

i hope bash will tell me that i'm stupid.. :)

jenshnielsen: I am referring to the make.bat call. .bat files are windows batch scripts they are intended to run in cmd(aka batch) and not in bash. You are right however that in git bash on window you can run a .bat script if you do `./make.bat` even if you are in bash 

in cmd `./make.bat` is a syntax error

astafan8: ok. decided to revert back to script for simplicity

astafan8:
@jenshnielsen could you have a look again ? after that we can press "update branch", ensure that everything ends with green, and squash-merge it.

azure-pipelines.yml
```diff
@@ -108,19 +108,20 @@ jobs:
...
       - script: |
-          CALL C:\Miniconda\condabin\conda_hook.bat
-          CALL conda activate qcodes
-          cd docs
-          REM Turn warnings into errors
-          set SPHINXOPTS=-W -v
+          CALL C:\Miniconda\condabin\conda_hook.bat && ^
+          CALL conda activate qcodes && ^
+          cd docs && ^
+          REM Turn warnings into errors && ^
```
jenshnielsen: I have never seen this construct before. Do you have some docs for this?

astafan8: which construct do you mean exactly? the `&&` does not let the next command execute if the preceeding command exits with non-0 exit code. the `^` is a line continuation in `cmd.exe`/`batch` (somehow it is not needed for `bash` case).

jenshnielsen: I get it . It's actually a bit more subtle than than the `^` escapes (Ignores) the next char after it. That is a new line in this case at the end of a line it does indeed have this effect.

astafan8:
@jenshnielsen is this good to go?

PR #1713, opened 2019-09-11
Add `Station.close_all_registered_instruments`

Dominik-Vogel:
Add a method to the Station that closes all instruments that have been registered as a component to the Station.
Open questions:
 - should Instruments also be removed from the station?

### DIFF
--- a/qcodes/station.py
+++ b/qcodes/station.py
@@ -275,9 +275,10 @@ class Station(Metadatable, DelegateAttributes):
         The instruments will stay registered as a component to the
         `Station`.
         """
-        for c in self.components.values():
-            if isinstance(c, Instrument):
-                c.close()
+        for k, v in self.components.items():
+            if isinstance(v, Instrument):
+                v.close()
+                self.remove_component(k)
 
     def load_config_file(self, filename: Optional[str] = None):
         """
--- a/qcodes/station.py
+++ b/qcodes/station.py
@@ -266,7 +266,7 @@ class Station(Metadatable, DelegateAttributes):
     def close_all_registered_instruments(self) -> None:
         """
         Closes all instruments that are registered to this `Station`
-        object by calling the :meth:`Instrument.close`-method on
+        object by calling the :meth:`.base.Instrument.close`-method on
         each one.
         The instruments will stay registered as a component to the
         `Station`.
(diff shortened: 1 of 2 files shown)


astafan8:
> should Instruments also be removed from the station?

this depends on the use case. could you expand on what motivated such a method in the first place?

qcodes/station.py
```diff
@@ -265,6 +265,19 @@ def __getitem__(self, key):
...
 
+    def close_all_registered_instruments(self) -> None:
+        """
+        Closes all instruments that are registered to this `Station`
+        object by calling the :meth:`.base.Instrument.close`-method on
+        each one.
+        The instruments will stay registered as a component to the
+        `Station`.
+        """
+        for k, v in self.components.items():
```
jenshnielsen: This ends up changing the dict while iteration over it. I think you need to do a `for k, v in list(self.components.items())` to clone the dict before iterating over it

qcodes/station.py
```diff
@@ -265,6 +265,19 @@ def __getitem__(self, key):
 
     delegate_attr_dicts = ['components']
 
+    def close_all_registered_instruments(self) -> None:
+        """
+        Closes all instruments that are registered to this `Station`
+        object by calling the :meth:`.base.Instrument.close`-method on
+        each one.
+        The instruments will stay registered as a component to the
+        `Station`.
```
jenshnielsen: Is this still true?

qcodes/station.py
```diff
@@ -275,10 +275,9 @@ def close_all_registered_instruments(self) -> None:
         The instruments will stay registered as a component to the
         `Station`.
         """
-        for k, v in self.components.items():
-            if isinstance(v, Instrument):
-                inst = v
```
spxtr: Out of curiosity, why do `inst = v` and then `inst.close()` rather than just `v.close()`?

astafan8: this is a style thing. "v" is just a value from a dictionary of components, hence it can be an instrument but also something else. after the isinstance part, the "v" is clarified to be an "inst" instrument, and this is expressed via defining a new variable because closing an instrument is a valid operation, while closing "a component" may not be valid because not all station components are expected to support the "close" method.

PR #1714, opened 2019-09-11
Station: fail hard when adding an `Instrument` with a name that is already registered

Dominik-Vogel:
Currently, when adding an instrument to the `Station` with a name that has already been taken, the instrument will be renamed by `make_unique`.

This behavior is, imho, not acceptable, because the instrument name is the best inter process identifier for instruments that we have. Therefore the name may not change silently from the specified name.

This PR removes `make_unique` and raises an error if a name is already taken.

Warning: this might break existing scripts!

### DIFF
--- a/qcodes/utils/helpers.py
+++ b/qcodes/utils/helpers.py
@@ -352,22 +352,6 @@ class LogCapture():
             self.logger.addHandler(handler)
 
 
-def make_unique(s, existing):
-    """
-    make string s unique, able to be added to a sequence `existing` of
-    existing names without duplication, by appending _<int> to it if needed
-    """
-    n = 1
-    s_out = s
-    existing = set(existing)
-
-    while s_out in existing:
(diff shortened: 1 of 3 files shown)


jenshnielsen:
In general I agree. We could consider deprecating make_unique for now and remote it later

qcodes/utils/helpers.py
```diff
@@ -352,6 +352,9 @@ def __exit__(self, type, value, tb):
             self.logger.addHandler(handler)
 
 
+@deprecate(
+    reason=("Instrument names are cross process identifiers that should be "
```
astafan8: to me the deprecation reason is specific for the use of the function, but in principle this function could've been used not only in station. but that's not important at all :)

Dominik-Vogel: This is very true, agreed!

PR #1716, opened 2019-09-12
Update Infiniium.py: add option to save waveform to usb

pjramsey37:
Adding a basic function save_data for saving displayed waveforms as a data list, and an accompanying parameter data_format for setting the format to save it in.

Fixes #issuenumber.

- I missed a comma after the vals line in the added data_format parameter. This needs adding in.
- 
- 

@mention one core developer


qcodes/instrument_drivers/Keysight/Infiniium.py
```diff
@@ -527,6 +527,12 @@ def __init__(self, name, address, timeout=20, **kwargs):
                             unit='s',
                             get_parser=float
                             )
+        
+        self.add_parameter('data_format',
+                           set_cmd='SAV:WAV:FORM {}',
+                           vals=Enum('CSV', 'ASC', 'BIN'),
```
astafan8: i'd strongly recommend to use a `val_mapping` here in order to hide the VISA values, for example:
```python
val_mapping={'csv': 'CSV', 'asc': 'ASC', 'binary': 'BIN'}
```
(would be nice to have something meaningful for the "ASC" case...)

pjramsey37: I'm new to github, so don't know how to edit my submission. For the record though, "asc" is short for "ASCIIXY", but I don't have any real understanding of what an ASCIIXY file is.

pjramsey37: I how now added val_mapping, but of course SCPI doesn't care about case, and accepts both the full name of the file type and the short-hand, so to my mind it would be preferable to have the validator accept "ASC", "ASCIIXY", "BIN", "BINARY", "CSV" and all case variations (i.e. make the validator completely case-insensitive), which would remove the need for any value mapping.

jenshnielsen:
@pjramsey37 Did you close this on purpose?

pjramsey37:
@jenshnielsen Yes, but I guess maybe I shouldn't have done? I have no clue what a pull request is - I created it because I couldn't work out how to edit my submission and that was the only way I could see to alert someone else to the mistake I'd left in the code, and I figured now it was fixed I have no further need of it.

WilliamHPNielsen:
I think we should reopen this PR, then, shouldn't we?

astafan8:
> how to edit my submission

@pjramsey37 when looking at the diff of the file, there's a menu on the right where you can choose to "edit file". You could also work locally using your own fork of qcodes instead of using github for that (if this didn't make any sense to you, i suggest to read some tutorials about `github`, and perhaps `git`).

-----

back to the initial purpose of the PR:

> Adding a basic function save_data for saving displayed waveforms as a data list, and an accompanying parameter data_format for setting the format to save it in.

@pjramsey37 is the PR still relevant? if yes, then let's make sure that all the tests pass and there are no typos or mistakes code-wise, and after that let's merge it, so that other users can benefit from your contribution :)

astafan8:
@pjramsey37 is this change still needed? we haven't heard from you for a while... If it still is, let's merge, otherwise I will close.

pjramsey37:
@astafan8 Sorry. When my job ended I just gave up on sorting this since I couldn't see how to fix the merging issues. I'm hoping I've now fixed the issues which prevented merging (which mostly related to line length).

PR #1718, opened 2019-09-13
AMI430 test: fix flakiness from asserting timestamps of Parameters

astafan8:
The `Parameter` uses `datetime.now()` to define a timestamp of when it's value was set. `datetime.now()` may not always provide good resolution, so adding a `time.sleep` to the test code is a quick solution for the flakiness.

### DIFF
--- a/qcodes/tests/drivers/test_ami430.py
+++ b/qcodes/tests/drivers/test_ami430.py
@@ -1,6 +1,7 @@
 import io
 import numpy as np
 import re
+import time
 import pytest
 from hypothesis import given, settings
 from hypothesis.strategies import floats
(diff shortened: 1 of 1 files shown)


jenshnielsen:
I agree. In principle 0.1 s should be more than enough On windows I get:

```
In [6]: time.get_clock_info('time')                                                                                                                                                                               Out[6]: namespace(adjustable=True, implementation='GetSystemTimeAsFileTime()', monotonic=False, resolution=0.015625)
```

astafan8:
@jenshnielsen shall i used then `time.get_clock_info('time').resolution` instead of a number? :)

jenshnielsen:
Yes as I understand the docs here that is the time resolution used https://docs.python.org/3/library/datetime.html#datetime.datetime.now To be save i would do `2*time.get_clock_info('time').resolution` or some such thing?

astafan8:
@jenshnielsen `2*time.get_clock_info('time').resolution` - magnifico! thanks!

PR #1721, opened 2019-09-16
Use importlib metadata

jenshnielsen:
This will be part of the standard lib in 3.8 and seems to be the recommend way. In addition  
use `requirements-parser` to do the parsing of the version requirements rather than writing our own parser

### DIFF
--- a/qcodes/utils/installation_info.py
+++ b/qcodes/utils/installation_info.py
@@ -63,6 +63,7 @@ def get_qcodes_requirements_versions() -> Dict[str, str]:
     """
     Return a dictionary of the currently installed versions of the packages
     that QCoDeS requires. The dict maps package name to version string.
+    If an (optional) dependency is not installed the name maps to "None".
     """
 
     req_names = get_qcodes_requirements()
@@ -73,6 +74,6 @@ def get_qcodes_requirements_versions() -> Dict[str, str]:
         try:
             req_versions[req] = version(req)
         except PackageNotFoundError:
-            pass
+            req_versions[req] = "None"
 
     return req_versions
--- a/qcodes/utils/installation_info.py
+++ b/qcodes/utils/installation_info.py
@@ -5,24 +5,16 @@ QCoDeS
 """
 from typing import Dict, List, Optional
 import subprocess
-import pkg_resources
-import importlib
 import json
 import logging
+import requirements
 
+try:
+    from importlib.metadata import distribution, version, PackageNotFoundError
+except ImportError:
+    # 3.7 and earlier
+    from importlib_metadata import distribution, version, PackageNotFoundError
 import qcodes
-
-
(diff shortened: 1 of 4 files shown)


jenshnielsen:
Ideally this should have #1722 merged first to test the python 3.8 implementation

qcodes/utils/installation_info.py
```diff
@@ -77,28 +67,12 @@ def get_qcodes_requirements_versions() -> Dict[str, str]:
...
-    for req_module in req_modules:
-        mod = importlib.import_module(req_module)
-        if req_module in _PACKAGE_NAMES:
-            req_pkg = _PACKAGE_NAMES[req_module]
-        else:
-            req_pkg = req_module
-        req_versions.update({req_pkg: mod.__version__})  # type: ignore
+    for req in req_names:
+        try:
+            req_versions[req] = version(req)
```
jenshnielsen: Unlike the previous implementation this captures extra requirements too. These may not be installed so here we skip import errors. Should we be more clever on that?

jenshnielsen:
This means that we uses the metadata recorded for the package rather than the `.__version__` attribute. This has both positive and negative results. It will not capture editable installs using versioneer or something similar (We still use the `__version__` field for qcodes so I don't think this is an issue. On the other hand this means that we actually capture version of all dependencies and not just the ones that report a `__version__` field

requirements.txt
```diff
@@ -16,3 +16,5 @@ dataclasses;python_version<'3.7'  # this won't be needed when we move to python
 websockets==8.0.2
 tabulate
 applicationinsights
+requirements-parser
```
astafan8: should these also be in `environment.yml` for conda?

qcodes/utils/installation_info.py
astafan8: > more clever

i can only think of logging.

requirements.txt
jenshnielsen: Yes good point

qcodes/utils/installation_info.py
jenshnielsen: Ok since the points of this is logging the info I have opted to log this with a version of "None"

jenshnielsen:
@WilliamHPNielsen As the original author I will wait for your comments before doing anything more with this

qcodes/utils/installation_info.py
```diff
@@ -73,32 +63,17 @@ def get_qcodes_requirements_versions() -> Dict[str, str]:
     """
     Return a dictionary of the currently installed versions of the packages
     that QCoDeS requires. The dict maps package name to version string.
+    If an (optional) dependency is not installed the name maps to "None".
```
WilliamHPNielsen: May I suggest "Not installed" instead of "None"?

PR #1723, opened 2019-09-17
Clean up warnings triggered by python 3.8

jenshnielsen:
This cleans up new warnings triggered by python 3.8

* Don't use `is` to compare against string literals
* Don't use deprecated and unused encoding kwarg for json.loads. This has been unused since 3.1
* Stop using deprecated `@coroutine` decorator
* Remove pre python 3.5 workarounds in coroutine tests


jenshnielsen:
There are still a number of warnings. They are as far as I can see triggered from dependencies and will hopefully be resolved upstream soon.

PR #1725, opened 2019-09-18
User-facing api docs: utils: plotting, slack, threading, validator, zmq_helpers

lakhotiaharshit:
Corrects user-facing API docs in utils. 
- plotting
- slack
- threading
- validator
- zmq_helpers

@mention one core developer
@astafan8


jenshnielsen:
I fixed a bunch of other small issues: 

* Before Args: there should always be a blank line.
* Args: is a special section -> So you should not have stuff like ` Inherited Args (see validators.Ints)`
* Context: should be in an Args section

A few more things that could be fixed 

- [ ] The docstring for the baseclass here https://qcodes.github.io/Qcodes/api/utils/validators.html#qcodes.utils.validators.Validator does not render nicely
- [ ] Numbers https://qcodes.github.io/Qcodes/api/utils/validators.html#qcodes.utils.validators.Numbers is missing a intro part of the docstring

qcodes/utils/validators.py
```diff
@@ -189,6 +192,8 @@ def __repr__(self) -> str:
 
 class Numbers(Validator):
     """
+    Requires a number  of type int, float, numpy.integer or numpy.floating.
+
     Args:
         min_value: Minimal value allowed, default -inf.
         max_value: Maximal value allowed, default inf.
```
jenshnielsen: Could you fix the todo below too?

lakhotiaharshit: OK

jenshnielsen: Thanks :)

jenshnielsen:
Could you also add a short docstring at the top of the module for validators.py 

This will be rendered here file:///C:/Users/jenielse/source/repos/Qcodes/docs/_build/html/api/utils/index.html on the right side of the table

qcodes/utils/validators.py
lakhotiaharshit: Fixed the docs to included `Raises` in all the methods.

jenshnielsen:
Thanks this looks great

PR #1731, opened 2019-09-23
Make InstrumentBase name and short_name read-only properties

peendebak:
Make the `InstrumentBase.name` and `InstrumentBase.short_name` read-only because those are not supposed to be changed after instrument instantiation.

@astafan8


jenshnielsen:
This looks good. I think this should have a breaking change note in the changelog. @astafan8 Would you mind starting the changelog for 0.7 and document this as a breaking change ?

astafan8:
@peendebak could you invite me as collaborator to your fork so that i can push commits to this branch?

astafan8:
@peendebak could you invite me as collaborator to your fork so that i can push commits to this branch?

PR #1733, opened 2019-09-25
Add function for operation complete status check to VisaInstrument

GateBuilder:
This PR proposes a function in ``VisaInstrument`` class that checks the Operation Complete Query (OPC) bit and ensures that the operation is complete, i.e., the bit is set to ``0``. The suggested function repeatedly returns the content of the Event Status Register (ESR) and clears it.

Subsumes and closes [#1168](https://github.com/QCoDeS/Qcodes/pull/1168).

@jenshnielsen 
@astafan8

### DIFF
--- a/qcodes/instrument/visa.py
+++ b/qcodes/instrument/visa.py
@@ -2,6 +2,7 @@
 from typing import Sequence, Optional, Dict
 import warnings
 import logging
+import time
 
 import visa
 import pyvisa.constants as vi_const
@@ -258,3 +259,21 @@ class VisaInstrument(Instrument):
         snap['timeout'] = self.timeout.get()
 
         return snap
+
+    def wait_until_complete(self, check_after: float, sleep_time: float) -> None:
+        """
+        A utility function that asynchronously queries the Event Status Register (ESR)
+        and checks if the ``0`` bit (the Operation Complete Query (OPC) bit) is set, i.e.,
+        the operation is complete. This function repeatedly returns the content of the ESR
+        and clears it.
+
+        Args:
+            check_after: Time to wait after which polling the OPC status to ESR starts.
+            sleep_time: Time to sleep at each iteration in the polling loop.
+        """
+        log.debug("Clearing ESR")
+        self.ask_raw("*ESR?")
+        self.write_raw("*OPC")
+        time.sleep(check_after)
+        while int(self.ask_raw("*ESR?").strip())%2==0:
+            time.sleep(sleep_time)


qcodes/instrument/visa.py
```diff
@@ -258,3 +259,21 @@ def snapshot_base(self, update: bool = True,
         snap['timeout'] = self.timeout.get()
 
         return snap
+
+    def wait_until_complete(self, check_after: float, sleep_time: float) -> None:
```
astafan8: shall `sleep_time` have a default value?

qcodes/instrument/visa.py
```diff
@@ -258,3 +259,21 @@ def snapshot_base(self, update: bool = True,
...
+        """
+        A utility function that asynchronously queries the Event Status Register (ESR)
+        and checks if the ``0`` bit (the Operation Complete Query (OPC) bit) is set, i.e.,
+        the operation is complete. This function repeatedly returns the content of the ESR
+        and clears it.
+
+        Args:
+            check_after: Time to wait after which polling the OPC status to ESR starts.
```
astafan8:
```suggestion
            check_after: Time in seconds to wait after which polling the OPC status to ESR starts.
```

qcodes/instrument/visa.py
```diff
@@ -258,3 +259,21 @@ def snapshot_base(self, update: bool = True,
...
+            sleep_time: Time to sleep at each iteration in the polling loop.
```
astafan8:
```suggestion
            sleep_time: Time in seconds to sleep at each iteration in the polling loop.
```

qcodes/instrument/visa.py
```diff
@@ -258,3 +259,21 @@ def snapshot_base(self, update: bool = True,
...
+        """
+        log.debug("Clearing ESR")
+        self.ask_raw("*ESR?")
```
astafan8: shouldn't this be `*CLS` as described in https://cdn.rohde-schwarz.com/pws/dl_downloads/dl_application/application_notes/1gp79/1GP79_1E_SCPI_Programming_Guide_SigGens.pdf ? i don't know, just trying to sanity-check.

qcodes/instrument/visa.py
```diff
@@ -258,3 +259,21 @@ def snapshot_base(self, update: bool = True,
...
+        A utility function that asynchronously queries the Event Status Register (ESR)
+        and checks if the ``0`` bit (the Operation Complete Query (OPC) bit) is set, i.e.,
+        the operation is complete. This function repeatedly returns the content of the ESR
+        and clears it.
+
+        Args:
+            check_after: Time to wait after which polling the OPC status to ESR starts.
+            sleep_time: Time to sleep at each iteration in the polling loop.
+        """
+        log.debug("Clearing ESR")
```
astafan8: we should probably use logger instance that belongs to the instrument instance, right?
```suggestion
        self.log.debug("Clearing ESR")
```

qcodes/instrument/visa.py
GateBuilder: Yeah, it seems this completely depends on experiment and/or the instrument. I did not want to set a default for this reason. But can be done, if you think that would be more proper?

astafan8: i trust your judgement on this. if really needed, it can be added later.

qcodes/instrument/visa.py
GateBuilder: I actually read it. If you read section 4.2.2.1, this PR just Pythonic implementation of it. There, CLS is used for initial clean up, can be implemented here as well, I am not sure.

astafan8: aha, i see. indeed, we should not need "initial clean up" here.

but why do we need this initial `*ESR?` call then? the document does not mention it. it just says that one should call `*ESR?` in order to poll only.

GateBuilder: Ok, I believe  ``*ESR?`` and ``*CLS`` is inter-changeable here, as both clears the ESR. The advantage of ``*ESR?`` is it also returns the content if there is any.

astafan8: ok, good then.

jenshnielsen:
I wonder what happens if this is used after a short command that has already finished by the time this is executed.  In my understanding that would mean that the opc bit has already been set and is cleared by `*ESR?` and the instrument will then probably deadlock? 

It seems to me that the correct way would be.

```
self.ask_raw("*ESR?")
self.write("startpotentiallyslowcmd")
while int(self.ask_raw("*ESR?").strip())%2==0:
     time.sleep(sleep_time)
```

This cannot be done easily since it's across multiple function calls in qcodes but perhaps the call to ```self.ask_raw("*ESR?")``` should simply not be here?

jenshnielsen:
@astafan8 This still needs testing on real instruments so i am bumping this to 0.10

jenshnielsen:
It seems like if this should be done it should be done using the pyvisa event handling https://pyvisa.readthedocs.io/en/latest/introduction/event_handling.html

umithardal:
Hi guys! Thanks for finishing this. I am having problem to sign in to my account (Gatebuilder). Let me know if I need to do anything!

PR #1738, opened 2019-09-27
Use Mypy 0.730

jenshnielsen:
This allows us to restrict ignore comments to only specific error codes. 

This has 2 benefits imho:

* We don't ignore any other errors than we should 
* This self documents why we need the ignore statements. 

For earlier versions of mypy this is simply ignored.


jenshnielsen:
For reference the error codes are given here https://mypy.readthedocs.io/en/latest/error_codes.html

PR #1739, opened 2019-09-27
Add mypy types to visa backend

jenshnielsen:
Noticed from #1738 that this important code is not typed

### DIFF
--- a/qcodes/instrument/visa.py
+++ b/qcodes/instrument/visa.py
@@ -1,5 +1,5 @@
 """Visa instrument driver based on pyvisa."""
-from typing import Sequence, Optional, Dict
+from typing import Sequence, Optional, Dict, Union
 import warnings
 import logging
 
@@ -81,7 +81,7 @@ class VisaInstrument(Instrument):
             self.visalib = visalib
 
         self.visabackend = None
-
+        self.visa_handle: visa.ResourceManager
         try:
             self.set_address(address)
         except Exception as e:
@@ -95,7 +95,7 @@ class VisaInstrument(Instrument):
         self.set_terminator(terminator)
         self.timeout.set(timeout)
 
-    def set_address(self, address):
+    def set_address(self, address: str):
         """
         Set the address for this instrument.
 
@@ -125,7 +125,7 @@ class VisaInstrument(Instrument):
         self.visa_handle = resource_manager.open_resource(address)
         self._address = address
 
-    def device_clear(self):
+    def device_clear(self) -> None:
         """Clear the buffers of the device"""
 
         # Serial instruments have a separate flush method to clear
@@ -159,9 +159,9 @@ class VisaInstrument(Instrument):
         self._terminator = terminator
 
         if self.visabackend == 'sim':
-                self.visa_handle.write_termination = terminator
+            self.visa_handle.write_termination = terminator
 
-    def _set_visa_timeout(self, timeout):
+    def _set_visa_timeout(self, timeout: Optional[Union[float, int]]):
 
         if timeout is None:
             self.visa_handle.timeout = None
@@ -169,7 +169,7 @@ class VisaInstrument(Instrument):
             # pyvisa uses milliseconds but we use seconds
             self.visa_handle.timeout = timeout * 1000.0
 
-    def _get_visa_timeout(self):
+    def _get_visa_timeout(self) -> None:
 
         timeout_ms = self.visa_handle.timeout
         if timeout_ms is None:
@@ -178,7 +178,7 @@ class VisaInstrument(Instrument):
             # pyvisa uses milliseconds but we use seconds
             return timeout_ms / 1000
 
-    def close(self):
+    def close(self) -> None:
         """Disconnect and irreversibly tear down the instrument."""
         if getattr(self, 'visa_handle', None):
             self.visa_handle.close()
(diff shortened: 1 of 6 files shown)


qcodes/instrument/visa.py
```diff
@@ -24,37 +24,36 @@ class VisaInstrument(Instrument):
...
-
+        name: What this instrument is called locally.
+        address: The visa resource name to use to connect.
+        timeout: seconds to allow for responses. Default 5.
         terminator: Read termination character(s) to look for. Default ``''``.
-
         device_clear: Perform a device clear. Default True.
-
+        visalib: Visa backend to use when connecting to this instrument.
+            This should be in the form of a string '@<backend>'.
```
jenshnielsen: Why is this like that you may ask. I think we should remove this logic and just make it a string in a different pr

qcodes/instrument/visa.py
```diff
@@ -159,17 +156,17 @@ def set_terminator(self, terminator: str):
...
+    def _set_visa_timeout(self, timeout: Optional[Union[float, int]]):
 
         if timeout is None:
             self.visa_handle.timeout = None
         else:
             # pyvisa uses milliseconds but we use seconds
             self.visa_handle.timeout = timeout * 1000.0
 
-    def _get_visa_timeout(self):
+    def _get_visa_timeout(self) -> None:
```
astafan8:
```suggestion
    def _get_visa_timeout(self) -> Optional[float]:
```

qcodes/instrument_drivers/rigol/DG1062.py
```diff
@@ -344,7 +345,7 @@ class DG1062(VisaInstrument):
     waveforms = DG1062Channel.waveforms
 
     def __init__(self, name: str, address: str,
-                 **kwargs: Dict) ->None:
+                 **kwargs) -> None:
```
astafan8: no need for it as it is implicit
```suggestion
                 **kwargs):
```

qcodes/instrument/visa.py
```diff
@@ -24,37 +24,36 @@ class VisaInstrument(Instrument):
...
-
         device_clear: Perform a device clear. Default True.
-
+        visalib: Visa backend to use when connecting to this instrument.
+            This should be in the form of a string '@<backend>'.
+            By default the NI backend is used, but '@py' will use the
+            ``pyvisa-py`` backend. Note that QCoDeS does not install (or even require)
+            ANY backends, it is up to the user to do that. see eg:
+            http://pyvisa.readthedocs.org/en/stable/names.html
         metadata (Optional[Dict]): additional static metadata to add to this
```
jenshnielsen: This is not a actually an explicit kwarg here and it is unused in the base class (`Instrument`). We should either remove it or actually use it

qcodes/instrument/visa.py
astafan8: do you mean the `@` logic? i believe that it comes from pyvisa

jenshnielsen: Yes but that does not mean that it's not pointless :) As far as I can see the only thing `self.visalib` is used for is to split off the `@` to form `self.visabackend` in `set_adress`

qcodes/instrument/visa.py
Dominik-Vogel: That is interesting. I think it should have gone into `Metadatable` but somebody forgot to put it into the `super().__init__` call in `InstrumentBase`.

qcodes/instrument/visa.py
```diff
@@ -24,37 +24,36 @@ class VisaInstrument(Instrument):
...
+    See help for :class:`qcodes.Instrument` for additional information on writing
     instrument subclasses.
 
     Attributes:
         visa_handle (pyvisa.resources.Resource): The communication channel.
     """
 
-    def __init__(self, name, address=None, timeout=5,
-                 terminator='', device_clear=True, visalib=None, **kwargs):
+    def __init__(self, name: str, address: str, timeout: Union[int, float] = 5,
```
jenshnielsen: I do not see the point of the default value of address being None. This however does break a test of simulated instruments

qcodes/instrument/visa.py
jenshnielsen: This will be resolved in #1740

qcodes/instrument/visa.py
astafan8: that's good, i agree.

qcodes/instrument/visa.py
astafan8: ok, understood. yes, i tend to agree. and then perhaps we can finally remove the logic about warnings and `@` being in `address` and `visalib` arguments.

qcodes/utils/validators.py
```diff
@@ -331,7 +331,7 @@ class Enum(Validator):
     eg. Enum(val1, val2, val3)
     """
 
-    def __init__(self, *values: Hashable) -> None:
+    def __init__(self, *values: Union[Hashable, None]) -> None:
```
```suggestion
    def __init__(self, *values: Optional[Hashable]) -> None:
```

qcodes/tests/test_visa.py
```diff
@@ -91,7 +92,7 @@ class TestVisaInstrument(TestCase):
     ]
 
     def test_ask_write_local(self):
-        mv = MockVisa('Joe')
+        mv = MockVisa('Joe', 'none')
```
astafan8: for readability?
```suggestion
        mv = MockVisa('Joe', 'none_address')
```

PR #1740, opened 2019-09-27
Fix instrument metadata

jenshnielsen:
Actually forward metadata to the Metadatable superclass. This is work in progress since its still lacking tests


jenshnielsen:
I am not sure the changes to remove ```**kwargs`` is the best way to preserve consistency. We could also simply ignore them (with a warning)

PR #1741, opened 2019-09-30
Add name parameter to Measurement class

### LINKED ISSUE #1736: Add name parameter to Measurement class
The `Measurement` class has the attribute `name`, which is currently hardcoded to an empty string ([here](https://github.com/QCoDeS/Qcodes/blob/master/qcodes/dataset/measurements.py#L686)). Because of this, doing `with meas.run() as datasaver: ...` will always give the dataset the name `"results"` ([here](https://github.com/QCoDeS/Qcodes/blob/master/qcodes/dataset/measurements.py#L582)). Would it be possible to add a `name` argument to the Measurement constructor? So that the `__repr__` of a dataset could be slightly more informative.

Or does this mess up something with the database? As in https://github.com/QCoDeS/Qcodes/issues/898

mgunyho:
Fixes #1736.

I added the name as the first kwarg to be consistent with other classes, and also a deprecation warning if the first parameter is instead an Experiment.


mgunyho:
I made the changes you requested. Sorry about not adding a test previously, I didn't look in the subfolders. I hope the test is not too simple.

qcodes/dataset/measurements.py
```diff
@@ -668,36 +667,27 @@ class Measurement:
...
+            produced by the measurement. If not given, a default value of
+            'results' is used for the dataset.
     """
 
-    def __init__(self, name: Optional[str] = '',
-                 exp: Optional[Experiment] = None,
-                 station: Optional[qc.Station] = None) -> None:
+    def __init__(self, exp: Optional[Experiment] = None,
+                 station: Optional[qc.Station] = None,
+                 name: Optional[str] = '') -> None:
```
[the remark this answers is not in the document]
mgunyho: True, the default value is the empty string, not None.

qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -416,6 +416,23 @@ def test_measurement_name(experiment, DAC, DMM):
         assert datasaver.dataset.table_name == expected_name
 
 
+def test_measurement_name_param(experiment, DAC, DMM):
+    fmt = experiment.format_string
+    exp_id = experiment.exp_id
+
+    name = 'yolo'
+
+    meas = Measurement(name=name)
```
astafan8:
```suggestion
    meas = Measurement(name=name, exp=experiment)
```

qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -416,6 +416,23 @@ def test_measurement_name(experiment, DAC, DMM):
...
+
+    meas.register_parameter(DAC.ch1)
+    meas.register_parameter(DMM.v1, setpoints=[DAC.ch1])
+
+    with meas.run() as datasaver:
+        run_id = datasaver.run_id
+        expected_name = fmt.format(name, exp_id, run_id)
+        assert datasaver.dataset.table_name == expected_name
```
astafan8: also assert name of the dataset
```suggestion
        assert datasaver.dataset.table_name == expected_name
        assert datasaver.dataset.name == name
```

qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -416,6 +416,23 @@ def test_measurement_name(experiment, DAC, DMM):
         assert datasaver.dataset.table_name == expected_name
 
 
+def test_measurement_name_param(experiment, DAC, DMM):
```
astafan8: could you also add a test where name argument is not passes and assert that the name of the dataset becomes "results"?

PR #1742, opened 2019-10-01
Fix/group parameter initial value

### LINKED ISSUE #1730: GroupParameter initial_value cannot be set
Here's a MWE:
```py
from qcodes.instrument.group_parameter import GroupParameter, Group
from qcodes import Instrument

class MyInstrument(Instrument):
    def __init__(self, name, *args, **kwargs):
        super().__init__(name, *args, **kwargs)

        self.add_parameter(name="foo",
                           initial_value=42,
                           parameter_class=GroupParameter
                          )

        self.group = Group([self.foo])

instr = MyInstrument("test")
```

### Expected behaviour
The instrument should have the GroupParameter with the given initial value.

### Actual behaviour
Raises `RuntimeError("Trying to set Group value but no  group defined")`.

### Proposed fix
The `GroupParameter` should defer setting the initial value until it has been added to the group. One way of doing it would be to add something like the following to `GroupParameter.__init__`, before the `super().__init__` call:
```py
if "initial_value" in kwargs:
    self._initial_value = kwargs["initial_value"]
    kwargs["initial_value"] = None
```
and then adding a `GroupParameter.add_to_group` method where the value is actually set, and calling that instead of just setting `parameter.group = self` in `Group.__init__`. I'm not 100% sure if this is the right way to do it.

### System
**qcodes branch**: master

**qcodes commit**: c7eef82d9ab68afb3546fb3c736f2d5b2ff02a14

mgunyho:
Fixes #1730 .

- Add initial_value kwarg to GroupParameter and appropriate logic for setting initial values for all parameters in a group, after the group has been created.

@astafan8

### DIFF
--- a/qcodes/instrument/group_parameter.py
+++ b/qcodes/instrument/group_parameter.py
@@ -30,8 +30,11 @@ class GroupParameter(Parameter):
 
     Args:
         name: Name of the parameter.
-        instrument: Instrument that this parameter belongs to; this
-              instrument is used by the group to call its get and set commands.
+        instrument: Instrument that this parameter belongs to; this instrument
+            is used by the group to call its get and set commands.
+        initial_value: Initial value of the parameter. Note that either none or
+            all of the parameters in a :class:`.Group` should have an initial
+            value.
 
         **kwargs: All kwargs used by the :class:`.Parameter` class, except
              ``set_cmd`` and ``get_cmd``.
@@ -40,6 +43,7 @@ class GroupParameter(Parameter):
     def __init__(self,
                  name: str,
                  instrument: Optional['Instrument'] = None,
+                 initial_value: Union[float, int, str, None] = None,
                  **kwargs
                  ) -> None:
 
@@ -48,6 +52,7 @@ class GroupParameter(Parameter):
                              "'get_cmd' kwarg")
 
         self.group: Union[Group, None] = None
+        self._initial_value = initial_value
         super().__init__(name, instrument=instrument, **kwargs)
 
         self.set = self._wrap_set(self.set_raw)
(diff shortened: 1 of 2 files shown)


qcodes/instrument/group_parameter.py
```diff
@@ -176,6 +181,26 @@ def __init__(self,
         else:
             self.get_parser = self._separator_parser(separator)
 
+        have_initial_values = [p._initial_value is not None
+                               for p in parameters]
+        if any(have_initial_values):
+            if not all(have_initial_values):
+                have_initial_values = [p.name for p in parameters
```
astafan8: could you please not re-use an existing variable? i suggest `params_with_initial_values` for example.

qcodes/tests/test_group_parameter.py
```diff
@@ -88,3 +98,19 @@ def test_raises_on_get_set_without_group():
...
+    initial_a = 42
+    initial_b = 43
+    dummy = Dummy("dummy", initial_a=initial_a, initial_b=initial_b)
+
+    assert dummy.a() == initial_a
+    assert dummy.b() == initial_b
+
+def test_raise_on_not_all_initial_values():
+
+    expected_err_msg = (r'Either none or all of the parameters in a group '
```
astafan8: thanks for a clear error message!

qcodes/tests/test_group_parameter.py
```diff
@@ -66,7 +77,6 @@ def test_sanity():
     assert dummy.a() == 3
     assert dummy.b() == 10
 
-
 def test_raise_on_get_set_cmd():
```
astafan8: please, don't remove the space - per pep8 there should be two empty lines between functions.

qcodes/tests/test_group_parameter.py
```diff
@@ -88,3 +98,19 @@ def test_raises_on_get_set_without_group():
     with pytest.raises(RuntimeError) as e:
         param.set(1)
     assert str(e.value) == "('Trying to set Group value but no group defined', 'setting b to 1')"
+
+def test_initial_values():
```
astafan8: please, add another empty line between functions (pep8)

qcodes/tests/test_group_parameter.py
```diff
@@ -88,3 +98,19 @@ def test_raises_on_get_set_without_group():
...
+    initial_a = 42
+    initial_b = 43
+    dummy = Dummy("dummy", initial_a=initial_a, initial_b=initial_b)
+
+    assert dummy.a() == initial_a
+    assert dummy.b() == initial_b
+
+def test_raise_on_not_all_initial_values():
```
astafan8: please, add another empty line between functions (pep8)

qcodes/tests/test_group_parameter.py
```diff
@@ -88,3 +98,19 @@ def test_raises_on_get_set_without_group():
...
+
```
astafan8: and this empty line can be removed :)

mgunyho:
Thanks for the comments! I implemented the fixes you suggested. I don't know why I deleted that one space and didn't adhere to pep8 in general.

PR #1743, opened 2019-10-02
Pep8 and type cleanup in instruments

jenshnielsen:
Was looking over this code and noticed a few issues. 

There are still more things that could be improved but this is imho a clean improvement

### DIFF
--- a/qcodes/instrument/channel.py
+++ b/qcodes/instrument/channel.py
@@ -745,7 +745,7 @@ class AutoLoadableInstrumentChannel(InstrumentChannel):
             raise RuntimeError(
                 "Object does not exist (anymore) on the instrument")
 
-    def write(self, cmd: str) -> Any:
+    def write(self, cmd: str):
         """
         Write to the instrument only if the channel is present on the instrument
         """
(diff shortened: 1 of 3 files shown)


qcodes/instrument/channel.py
```diff
@@ -787,30 +789,31 @@ class AutoLoadableChannelList(ChannelList):
...
+        ValueError: If :class:`chan_type` is not a subclass of
+            :class:`InstrumentChannel`
         ValueError: If ``multichan_paramclass`` is not a subclass of
-            :class:`MultiChannelInstrumentParameter` (note that a class is a subclass
-            of itself).
-
+            :class:`MultiChannelInstrumentParameter` (note that a class is a
+            subclass of itself).
     """
     def __init__(
```
jenshnielsen: I am not a big fan of this indention style but left it for now

qcodes/instrument/channel.py
```diff
@@ -732,33 +734,33 @@ def remove(self)->None:
...
         raise NotImplementedError("Please subclass")
 
-    def _assert_existence(self)->None:
+    def _assert_existence(self) -> None:
         if not self._exists_on_instrument:
             raise RuntimeError(
                 "Object does not exist (anymore) on the instrument")
 
-    def write(self, cmd: str)->Any:
+    def write(self, cmd: str):
```
WilliamHPNielsen: Why not ` -> None` for this one? Is `mypy` unhappy about the `super()`?

jenshnielsen: Because @astafan8 is going to tell me that its redundant :)

qcodes/instrument/channel.py
astafan8: what would you recommend for the case of many arguments each having long type annotations?

jenshnielsen: One per line is fine but this ( on separate lines thing is not great. These lines are actually already too long even with this style

PR #1744, opened 2019-10-02
Bugfix/parameter set to

WilliamHPNielsen:
There is a small bug in `parameter.set_to` that this PR fixes. 

The bug:

Have an instrument driver. 
Have a parameter, `p`, of that driver with a validator that does not accept `None`. 
Connect to instrument (now all parameters have `_latest["value"] == None`). 
Use `with p.set_to(some_valid_value):`. 
Get `ValueError` from the setter of `p` that `None` is invalid.

This PR fixes the bug by querying `p` in case `_latest["value"] == None`.

@QCoDeS/core

### DIFF
--- a/qcodes/tests/test_parameter.py
+++ b/qcodes/tests/test_parameter.py
@@ -1192,11 +1192,26 @@ class TestSetContextManager(TestCase):
 
     def setUp(self):
         self.instrument = DummyInstrument('dummy_holder')
-        self.instrument.add_parameter(
-            "a",
-            set_cmd=None,
-            get_cmd=None
-        )
+
+        self._vp_value = "foo"
+
+        self.instrument.add_parameter("a",
+                                      set_cmd=None,
+                                      get_cmd=None)
+
+        # This parameter mocks an actual instrument parameter; when first
+        # connecting to the instrument, it has the _latest["value"] None.
+        # We must call get() on it to get a valid value
+        self.instrument.add_parameter("validated_param",
+                                      set_cmd=self._vp_setter,
+                                      get_cmd=self._vp_getter,
+                                      vals=vals.Enum("foo", "bar"))
+
+    def _vp_getter(self):
+        return self._vp_value
+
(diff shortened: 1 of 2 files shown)


qcodes/tests/test_parameter.py
```diff
@@ -1192,11 +1192,26 @@ class TestSetContextManager(TestCase):
 
     def setUp(self):
         self.instrument = DummyInstrument('dummy_holder')
-        self.instrument.add_parameter(
-            "a",
-            set_cmd=None,
-            get_cmd=None
-        )
+
+        self._vp_value = "foo"
```
astafan8: could you move this line close to the "validated_param" parameter so that it's easier to notice? it took me a while to understand why _vp_getter and _vp_setter are needed :)

qcodes/tests/test_parameter.py
```diff
@@ -1214,6 +1229,9 @@ def test_context(self):
             assert self.instrument.a.get() == 3
         assert self.instrument.a.get() == 2
 
+    def test_validated_param(self):
+        with self.instrument.validated_param.set_to("bar"):
+            assert self.instrument.validated_param.get() == "bar"
```
astafan8: for strictness :)
```suggestion
            assert self.instrument.validated_param.get() == "bar"
        assert self.instrument.validated_param.get_latest() == "foo"
```

qcodes/tests/test_parameter.py
```diff
@@ -1214,6 +1229,9 @@ def test_context(self):
             assert self.instrument.a.get() == 3
         assert self.instrument.a.get() == 2
 
+    def test_validated_param(self):
```
astafan8: adding precondition explicitly :)
```suggestion
    def test_validated_param(self):
        assert self.instrument.validated_param.get_latest() is None
```

WilliamHPNielsen:
This was meant as a quick bugfix. I have some more features coming for the `set_to` functionality. I'll make a larger PR.

PR #1747, opened 2019-10-03
Disallow missing types for instrument base and visa

jenshnielsen:
Since these are core parts of qcodes.


jenshnielsen:
@astafan8 Absolutely  but one step at a time :)

PR #1748, opened 2019-10-03
explicitly install attrs 

jenshnielsen:
to workaround  what seems to be a bug in pip (19.2.3)

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -10,3 +10,4 @@ codecov
 asv
 gitpython
 pylint
+attrs>=19.2.0 # needed for hypothesis >=4.38.1 but not correctly honored by pip 19.2.3


jenshnielsen:
This is a known issue. https://github.com/pypa/pip/issues/988  Since pytest comes before hypothesis in the requirements file and they both depend on attrs (and pytest supports an older version) only the constraints of pytest are imposed

NeolithEra:
Hi, maybe you can try a dependency diagnosis website http://www.watchman-pypi.com/dependency_analysis, which can help diagnoze dependency conflict problems and automatically visualize your dependency tree and those of your downstream projects.

Best,
Neolith

jenshnielsen:
@NeolithEra Thanks but this is IMHO not a conflict as both pytest and hypothesis happily supports `attrs>=19.2.0` this is just the result of pypi not doing full dependency resolving and missing one of the requirements

NeolithEra:
@jenshnielsen Thanks for your answer. It can visualize the full dependency tree of any project, which helps us diagnoze the dependency problems of Python world.

pradyunsg:
@jenshnielsen That user has been spamming multiple issues and PRs with that comment.

---

Is this requirements file self contained? Would it be possible for you to create a minimal requirements.txt that demonstrates this failure and file an issue over on https://github.com/pradyunsg/zazo? That's where I'm accumulating reports of pip's resolver not making the right choices, so that we have a bunch of tests to benchmark against, when we come around to fixing that issue.

PR #1749, opened 2019-10-04
Parameter set to improvements

WilliamHPNielsen:
This PR fixes a bug in `parameter.set_to` and reduces the amount of instrument communication (calls to `parameter.set`) that happens in the context manager.

The bugfix: if a parameter has been initialized, but not yet gotten, it will have a `_latest["value"]` of `None`. When exiting the context manager, the value that the parameter is restored to  should not be `None`, but instead the result of `parameter.get`.

The improvement: if `parameter` has value `X`, then `with parameter.set_to(X)` is now a NOOP.

The improvement allows us to write parameter getters that hinge on other parameters without introducing unnecessary communication. An example (simplified) for a DMM that can be in voltage or current mode:
```python
def _voltage_getter(self):
    with self.mode.set_to("voltage"):
        value = self.ask("READ?")
    return value

def _current_getter(self):
    with self.mode.set_to("current"):
        value = self.ask("READ?")
    return value
```
Imagine getting the voltage or current in a tight loop. In that case, we do not want the communication to be
```
MODE:VOLTAGE
READ?
MODE:VOLTAGE
```
for each get call.

@QCoDeS/core

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -115,15 +115,16 @@ class _SetParamContext:
     >>> assert abs(dac.voltage() - v) <= tolerance
 
     """
-    def __init__(self, parameter):
+    def __init__(self, parameter: "Parameter", value: Any):
         self._parameter = parameter
+        self._value = value
         self._original_value = self._parameter._latest["value"]
 
         if self._original_value is None:
-            self._original_value = self._parameter.get()
+            self._original_value = self._parameter.get()  # type: ignore
 
     def __enter__(self):
-        pass
+        self._parameter.set(self._value)
 
     def __exit__(self, typ, value, traceback):
         self._parameter.set(self._original_value)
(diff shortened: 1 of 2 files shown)


qcodes/instrument/parameter.py
```diff
@@ -115,15 +115,23 @@ class _SetParamContext:
...
-    def __init__(self, parameter):
+    def __init__(self, parameter: "Parameter", value: Any):
         self._parameter = parameter
+        self._value = value
         self._original_value = self._parameter._latest["value"]
 
+        self._value_is_changing = self._value != self._original_value
+
+        if self._original_value is None and self._value_is_changing:
+            self._original_value = self._parameter.get()  # type: ignore
```
jenshnielsen: Following the merge of #1738 please don't use bare ignores. use `--show-error-codes` to get the error code needed and do `ignore[errorcode]`

WilliamHPNielsen: Gotcha!

PR #1751, opened 2019-10-07
Ensure that get_latest will call get on a parameter

jenshnielsen:
If the parameter has never been captured it seems like `get_latest` should also call `get`

This is draft due to the missing tests

### DIFF
--- a/qcodes/tests/test_parameter.py
+++ b/qcodes/tests/test_parameter.py
@@ -32,8 +32,18 @@ class GettableParam(Parameter):
 
     def get_raw(self):
         self._get_count += 1
-        self._save_val(42)
         return 42
+
+class BetterGettableParam(Parameter):
+    """ Parameter that keeps track of number of get operations,
+        But can actually store values"""
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._get_count = 0
+
+    def get_raw(self):
+        self._get_count += 1
+        return self._latest['raw_value']
 
 
 class DeprecatedParam(Parameter):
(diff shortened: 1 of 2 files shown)


jenshnielsen:
Todo

- [ ] Test that get is called / not called when ts is not set / set
- [ ] Test that error is raised when max_val_age is not None and get is not set
- [ ] Test that get_latest raises when get is false and set / initial value is not called
- [ ] Test that get_latest does not raise when get is false but set / initial value has been called

jenshnielsen:
I uncovered a bug where max_val_age is never used in Parameter. This is fixed too but need a test here

qcodes/tests/test_parameter.py
```diff
@@ -223,6 +225,125 @@ def test_get_latest(self):
         self.assertEqual(local_parameter.get_latest(), 2)
         self.assertGreater(local_parameter.get_latest.get_timestamp(), after_set)
 
+    def test_get_latest_unknown(self):
+        """
+        Test that get latest on a parameter that has not been acquired will
+        trigger a get
```
jenshnielsen: We could make these use parameter sub classes that counts the calls to get

jenshnielsen:
@QCoDeS/core this is ready for review

qcodes/tests/test_parameter.py
```diff
@@ -223,6 +235,124 @@ def test_get_latest(self):
...
+                                   'ts': None}
+        assert local_parameter.get_latest.get_timestamp() is None
+        before_get = datetime.now()
+        assert local_parameter._get_count == 0
+        assert local_parameter.get_latest() == value
+        assert local_parameter._get_count == 1
+        # calling get_latest above will call get since TS is None
+        # and the TS will therefore no longer be None
+        assert local_parameter.get_latest.get_timestamp() is not None
+        assert local_parameter.get_latest.get_timestamp() >= before_get
```
astafan8: Shall we make this stricter by using the sleep for time resolution?

qcodes/tests/test_parameter.py
```diff
@@ -223,6 +235,124 @@ def test_get_latest(self):
...
+        no get cmd raises correctly.
+        """
+        value = 1
+        with self.assertRaises(SyntaxError):
+            _ = Parameter('test_param', set_cmd=None,
+                          get_cmd=False,
+                          max_val_age=1, initial_value=value)
+
+        # _BaseParameter does not have this check since get_cmd could be added
+        # in a subclass. Here we create a subclass that does not and
```
astafan8: Does not what?

qcodes/tests/test_parameter.py
jenshnielsen: I would prefer not to. We have been bitten by time resolution introduced failures before hand. This is enough to test what we want

PR #1752, opened 2019-10-07
Deprecation: Introduce `QCoDeSDeprecationWarning`, extract deprecation function as standalone, apply decorator to all deprecated methods

Dominik-Vogel:
This change simply refactor the code for raising a deprecation warning from the deprecation decorator and exposes it as a public method, so that deprecation warnings can be issued in other cases that for function and methods.
I did this in the context of doing some slight changes to the station configuration yaml file.


qcodes/utils/deprecate.py
```diff
@@ -3,6 +3,21 @@
...
+    alternative: Optional[str] = None
+) -> None:
+    msg = f'The {what} is deprecated'
+    if reason is not None:
+        msg += f', because {reason}'
+    msg += '.'
+    if alternative is not None:
+        msg += f' Use \"{alternative}\" as an alternative.'
+
+    warnings.warn(msg, DeprecationWarning, stacklevel=2)
```
jenshnielsen: I as surprised that using a DeprecationWarning given that these are hidden by default. I suggest making a qcodes deprecation warning that inherits from runtime warning

jenshnielsen: Does it make sense to make stacklevel an optional arg with a default value of 2

Dominik-Vogel: Hehe, I was thinking the same thing, but I guess I was too lazy and thought we could add it once we need it. But you are right, since I am touching this code anyways, I will add it.

Dominik-Vogel: And good point about the `RuntimeWarning`!

jenshnielsen:
Technically the change in warning type should probably be documented. Do you mind starting the changelog and add that there?

Dominik-Vogel:
you are absolutely right. I will start the changelog.

PR #1753, opened 2019-10-07
Station configuration: remove type keyword

Dominik-Vogel:
This PR changes the syntax of the station config files slightly, in order to be more consistent and to allow easier validation in the future.
The change disallows the `type` keyword and will use the full python style path to the instrument in the `driver` keyword. The old syntax will still be available but show a deprecation warning.
This means that
```
instruments:
  mock:
    driver: qcodes.tests.instrument_mocks
    type: DummyChannelInstrument
```
Will show a deprecation warning and will have to be rewritten as
```
instruments:
  mock:
    driver: qcodes.tests.instrument_mocks.DummyChannelInstrument
```
The syntax of the tests has been changed to this new style and a test for the old style has been added.


Dominik-Vogel:
awesome @astafan8 ! That is absolutely right! I will get to it.

jenshnielsen:
@GateBuilder and @lakhotiaharshit You are probably likely to see a bunch or deprecation varnings related to this as 0.7 rolls out

PR #1755, opened 2019-10-09
Set pytest directory to speed up test collection

Dominik-Vogel:
This seems to speed up my local test collection.


Dominik-Vogel:
@jenshnielsen pinging because I added to restrict the file glob pattern.

setup.cfg
```diff
@@ -1,3 +1,8 @@
+[tool:pytest]
+testpaths = "qcodes/tests"
+python_files = test_*.py
+python_functions = test_*
```
astafan8: the python functions and python files entries - these are pytest defaults, right? do we really need to explicitly define them here?

Dominik-Vogel: I wasn't 100% sure, how this was handled, but I read up on it: you are absolutely right. These are the defaults, which get overwritten individually when set. (only difference is python_function = test*, but we really don't care). So I will remove them again.

PR #1756, opened 2019-10-09
Station configuration deprecate parameter limits as string

Dominik-Vogel:
Currently limits of  a parameter can be specified by a comma separated string in the station config file:
```
instruments:
  mock:
    driver: qcodes.tests.instrument_mocks.DummyInstrument
    init:
      gates: {"ch1"}
    parameters:
      ch1:
        limits: -10, 10
```
This PR deprecates that standard in favor of a more consistent usage of the grammar, by using an array:
```
instruments:
  mock:
    driver: qcodes.tests.instrument_mocks.DummyInstrument
    init:
      gates: {"ch1"}
    parameters:
      ch1:
        limits: [-10, 10]
```


qcodes/station.py
```diff
@@ -485,7 +485,14 @@ def setup_parameter_from_dict(instr: Instrument, name: str,
...
                 elif attr == 'limits':
-                    lower, upper = [float(x) for x in val.split(',')]
+                    if isinstance(val, str):
+                        issue_deprecation_warning(
+                            ('use of a comma separated string for the limits'
+                             'keyword'),
+                            alternative='an array like "[lower_lim, upper_lim]"')
+                        lower, upper = [float(x) for x in val.split(',')]
+                    else:
+                        lower, upper = val
```
astafan8: I was about to ask "shall we add some validation here so that if a user incorrectly supplied the limits (for example, only one limit is given), then the error message is more clear than 'could not unpack a list of 1 element into two variables'" - but then i realized that a schema is coming, hence all the validation will be done via the schema. Is my understanding correct?

qcodes/tests/test_station.py
```diff
@@ -549,15 +549,15 @@ def test_setup_delegate_parameters():
...
+        limits: [-10, 10]
     add_parameters:
       gate_a:
         source: ch1
         unit: mV
         label: main gate
         scale: 2
         offset: 1
-        limits: -6, 6
+        limits: [-6, 6]
```
astafan8: shall we have these as floats for the sake of "covering more cases"?
```suggestion
        limits: [-6.0, 6.0]
```

qcodes/tests/test_station.py
```diff
@@ -646,3 +646,19 @@ def test_deprecated_type_keyword():
...
+    driver: qcodes.tests.instrument_mocks.DummyInstrument
+    init:
+      gates: {"ch1"}
+    parameters:
+      ch1:
+        limits: -10, 10
+    """)
+    with warnings.catch_warnings(record=True) as w:
+        st.load_instrument('mock')
+        assert issubclass(w[-1].category, QCoDeSDeprecationWarning)
```
astafan8: newline and indentation :)
```suggestion
    assert issubclass(w[-1].category, QCoDeSDeprecationWarning)

```

qcodes/station.py
Dominik-Vogel: Exactly :-)

qcodes/tests/test_station.py
Dominik-Vogel: Good point. Let me also change the check then.

Dominik-Vogel:
@GateBuilder @lakhotiaharshit you will get similar deprecation warnings from this one, too.

PR #1757, opened 2019-10-09
_BaseParameter cached value: add set_cache method, make _save_val method useless, convert raw_value attribute to property

### LINKED ISSUE #1222: Feature request: set_latest
Parameters roughly speaking work like this:

- To set the parameter value via instrument communication, use `set`
- To get the parameter value via instrument communication, use `get`
- To get the parameter value without instrument communication, use `get_latest`
- To set the parameter value without instrument communication, use... `_save_val`?

Shouldn't that be `set_latest`? Furthermore, `_save_val` currently does not quite work as expected in cases where a `Parameter` has no `get_cmd`, but does have a `val_mapping`. An example:

```python
p1 = Parameter('p1', val_mapping={'one': 1, 'two': 2}, set_cmd=lambda x: x)
p1._save_val('one')
p1.get()
```
Raises a TypeError, because the `get` calls a `lambda: self.raw_value`, but the `raw_value` is `None`, since `_save_val` does not update the `raw_value` if there is a `val_mapping`. Link to code:
https://github.com/QCoDeS/Qcodes/blob/b68979ef084a19373976daaef1d6a19ba5644ae9/qcodes/instrument/parameter.py#L341-L355

Ideally, I would like
```python
p1 = Parameter('p1', val_mapping={'one': 1, 'two': 2}, set_cmd=lambda x: x)
p1.set_latest('one')
```
to set `p1._latest` to `{'value': 'one', 'ts': datetime(blah blah), 'raw_value': 1}`. Similarly for cases with scaling, offset, and parsers.

@QCoDeS/core 

PS: This came up in #1218, so although slightly contrived, in **can** happen in the real world.

astafan8:
## Description

This is needed for setting the cached value of a parameter **without calling the meat of it's `set` method** (for example, without communicating to the instrument, in case the parameter is part of an instrument).

## Use cases that this covers for sure:
- when initializing an instance of an instrument driver, this can be used to set the cached value of a parameter to a known default of an instrument for cases where instrument does not support querying that value, and only supports setting it to smth (good example, Keysight B1500 and its modules).
- resolves #1222 , for DynaCool driver

## ToDo/inventory:
- [x] avoid code duplication of the `wrap_set` method in `set_cached` by refactoring `-`_from_value_to_raw_value` into its own method and use both places (also refactor `_from_raw_value_to_value` from `wrap_get`, just for fun)
- [x] testing, testing, testing
- [x] revise the name of the method - make it `set_cache` from `set_latest`
- [x] add `_update_cache_from` method that is the only one that updates the `_latest` dictionary, and use it in `wrap_get`
- [x] ~~make necessary changes to `DelegateParameter`~~ - a bug discovered where `delegate.get_latest()` does not forward the call to `source.get_latest()` which is a bug, issue https://github.com/QCoDeS/Qcodes/issues/1782 created - fixing will happen in a separate PR.
- [x] ~~make necessary changes to `ParameterWithSetpoints` to support it~~ - the're none, so nothing to do
- [x] add docstrings
- [x] ~~mention somewhere in the documentation~~ **won't do** in this PR since `get_latest` is also not mentioned anywhere
- [x] link to github issues that this PR resolves
- [x] don't use `_save_val` in qcodes infra, use `set_cache` or `_update_cache_with` instead; `_save_val` deprecation will follow in other PRs
- [x] convert `raw_value` attribute to a property: on get it returns `self._latest['raw_value']` and on set it only sets `self._latest['raw_value']`; deprecation, forbidding setting `raw_value` via the property, etc will follow in other PRs
- [x] don't use `raw_value` property in `parameter.py`

## Related tasks for other PRs:
- `_save_val`
  - remove using `_save_val` in all known drivers, try using pre-this-PR API for this, for example `set`; if not possible, then use `set_cache` and bump the dependent qcodes version
    - (https://github.com/QCoDeS/Qcodes/pull/1790) (https://github.com/QCoDeS/Qcodes/pull/1789) `qcodes` drivers
    - (https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/25) `qcodes_contrib_drivers` drivers
  - deprecate `_save_val`, and define qcodes version when it will be removed
  - remove `_save_val`
- `raw_value`
  - remove getting `raw_value` via `raw_value` attribute (well, now property) from all known instrument drivers - try to use existing API, for example, `get_latest`
    - (https://github.com/QCoDeS/Qcodes/pull/1788) `qcodes` drivers
    - ~~`qcodes_contrib_drivers`~~ - there seems to be none
  - remove setting `raw_value` via `raw_value` attribute from all known instrument drivers - try using pre-this-PR API for this, for example `set`; if not possible, then use `set_cache` and bump the dependent qcodes version
    - (https://github.com/QCoDeS/Qcodes/pull/1789) `qcodes` drivers
    - ~~`qcodes_contrib_drivers`~~ - there seems to be none
  - deprecate setting `raw_value` via attribute, and define qcodes version when it will be removed
  - deprecate getting `raw_value` via attribute, and define qcodes version when it will be removed
  - remove `raw_value` altogether
- (https://github.com/QCoDeS/Qcodes/pull/1786) use `get_latest()` for `_SetParamContext._original_value`, not `_latest[`value`]`
- (https://github.com/QCoDeS/Qcodes/pull/1787) for consistency with `get` and `get_timestamp`, add `get_raw_value` method to `GetLatest` class thing that returns `self._latest['raw_value']`
- (https://github.com/QCoDeS/Qcodes/pull/1791) add `get_cache`/`get_cache_raw`/`get_cache_timestamp` which aliases `get_latest`, and refactor the meat from `get_latest` to be used in those new methods, also make sure that `parameter.get_latest` gets the doctoring of `GetLatest.get` method
- add tests for roundtipping of values and raw values on the level of `_from_*` methods for hypothesis-generated values, parsers, and all the other stuff
- fix formatting of `test_parameter.py`, also use  ONLY `pytest` there


astafan8:
@QCoDeS/core Let's start reviewing of this PR. Please have a look at the tests in order to find out if the captured behavior is the one that we want/like. Please also let know if ideologically/conceptually `DelegateParameter` and ParameterWithSetpoints` have to implement something special with respect to `set_cached` method (or if it's good as-is).

qcodes/instrument/parameter.py
```diff
@@ -421,6 +422,65 @@ def snapshot_base(self, update: bool = True,
...
+            raw_value = self.set_parser(raw_value)
+
+        return raw_value
+
+    def set_cached(self, value: ParamDataType) -> None:
+        """
+        Set the cached value of the parameter without invoking the
+        ``set_cmd`` of the parameter (if it has one). For example, in case of
+        an instrument parameter, calling :meth:`set_cached` as opposed to
+        calling :meth:`set` will only change the internally-stored value of
```
jenshnielsen: Docs build fail since you cannot put in a reference to set that is dynamically generated

qcodes/instrument/parameter.py
```diff
@@ -421,6 +422,65 @@ def snapshot_base(self, update: bool = True,
...
+                # Use single offset for all values
+                raw_value += self.offset
+
+        # parser last
+        if self.set_parser is not None:
+            raw_value = self.set_parser(raw_value)
+
+        return raw_value
+
+    def set_cached(self, value: ParamDataType) -> None:
```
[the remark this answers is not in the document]
astafan8: will be renamed to `set_cache` as agreed in an offline discussion

qcodes/tests/test_parameter.py
```diff
@@ -513,6 +538,69 @@ def test_scale_and_offset_raw_value_iterable(self, values, offsets, scales):
         if isinstance(scales, Iterable) and not isinstance(offsets, Iterable):
             event('Scale is array but not offset')
 
+    @settings(max_examples=300)
+    @given(
```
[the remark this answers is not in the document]
astafan8: sure. as we agreed, let's keep this task optional for this PR, but i'll likely do it.

astafan8:
@QCoDeS/core ready for review again - see the description for the scope and future todo (i'll either make PRs for them shortly after this merge or add cards to our board).

qcodes/instrument/parameter.py
```diff
@@ -316,6 +322,38 @@ def __init__(self, name: str,
...
 
+    @property
+    def raw_value(self) -> ParamRawDataType:
+        """
+        Represents the cached raw value of the parameter.
+
+        Setting the ``raw_value`` is not recommended as it may lead to
+        inconsistence state of the parameter.
+
+        Note that this property will be deprecated soon.
```
jenshnielsen: You can have a look at step for how to get the getter and setter into the rendered docstring

qcodes/instrument/parameter.py
```diff
@@ -316,6 +322,38 @@ def __init__(self, name: str,
...
+        """
+        return self._get_cache_raw()
+
+    @raw_value.setter
+    def raw_value(self, new_raw_value: ParamRawDataType) -> None:
+        self._set_cache_raw(new_raw_value)
+
+    def _set_cache_raw(self, raw_value: ParamRawDataType) -> None:
```
jenshnielsen: Do we really need these 2 private methods ?

qcodes/instrument/parameter.py
astafan8: will do! thanks for noting!

qcodes/instrument/parameter.py
astafan8: Basically, having these methods allows to limit the number of the references to `._latest`. 

`_get_cache_raw` is used internally in Parameter/_BaseParameter, and will be superceeded later in another PR by a public `get_cache_raw`.

`_set_cache_raw` is used only in the `raw_value` setter, and serves only the purpose of limiting references to `._latest`.

Would you agree with this approach?

jenshnielsen: I don't feel strongly about it but private setters feels pointless to me. IMHO setters and getters are needed to avoid leaking the (private) storage implementation of the date but since the method is private too there is no worry about leaking. 

They may also be used to do some convert or cast but that is not the case here.

qcodes/instrument/parameter.py
```diff
@@ -994,7 +1085,7 @@ def __init__(self, name: str,
 
         if not hasattr(self, 'set') and set_cmd is not False:
             if set_cmd is None:
-                self.set_raw: Callable = partial(self._save_val, validate=False)
+                self.set_raw: Callable = lambda x: x
```
jenshnielsen: This is ok since set wrapper already calls save val right?

jenshnielsen: And we do have tests for that I assume

astafan8: indeed. `_save_val` (now `_update_cache`) is part of 'set' and 'get' thanks to `wrap_get` and `wrap_set` anyway. Many tests implicitly test this for sure, but perhaps also explicitly.

let me add a test that goes inside the `._latest` and asserts stuff there just to be sure.

qcodes/instrument/parameter.py
astafan8: I see. Let me leave it for now. at the moment i see the value of explicitly pointing that the new "cache stuff" is layered between "raw_value" and the "_latest". Can always be removed in later PRs during "further refactoring"

qcodes/instrument/parameter.py
astafan8: done

PR #1759, opened 2019-10-11
Add validation to station config files

Dominik-Vogel:
With this PR every time a station config file gets loaded it gets validated against a json schema. All `ValidationError` s are caught and reraised as warnings. Should the schema work for the config files in use, this mechanism can be deactivate so that we raise real errors.

The schema is not 100% water tight, due to limitations of the mechanism. It covers all the basic cases and disallows wrong attributes. It restricts parameter and instrument names to valid python names (alphanumeric with hyphen or underscore, for parameters potentially joined with dots), but it does not restrict the names to names that 'exist' e.g. you can source parameters that have a potentially valid name, but are not part of any channel (`this_channel_does_not_exist.neither_does_this_parameter` would be a valid source).

It accepts the deprecated `type` keyword and the deprecated `string` value for `limits`.

Added the feature of setting up vscode integration:
vscode_integration


qcodes/station.py
```diff
@@ -355,11 +369,19 @@ def update_load_instrument_methods():
...
+                                f'lazy loading method {method_name} could '
                                 'be created in the Station.')
-        self._config = YAML().load(config)
+        yaml = YAML().load(config)
+        with open(get_schema_file_path()) as f:
+            schema = json.load(f)
+        try:
+            jsonschema.validate(yaml, schema)
+        except jsonschema.exceptions.ValidationError as e:
+            warnings.warn(e.message, ValidationWarning)
```
jenshnielsen: Do we log this?

Dominik-Vogel: 👍 All warnings get logged.
https://github.com/QCoDeS/Qcodes/blob/72890d0348cf4a141e8e879b950fdd4f4bdb6440/qcodes/logger/logger.py#L214

qcodes/dist/tests/station/example.station.yaml
```diff
@@ -0,0 +1,109 @@
...
+  decadac:
+    driver: qdev-wrappers.customised_instruments.Decadac_ext
+    type: Decadac_ext
+    address: ASRL4::INSTR
+    init:
+      terminator: \n
+    add_parameters:
+      Bx: {source: ch01.fine_volt, label: Bx, unit: T, scale: 28, limits: [-1,1], monitor: true}
+  # this is a mock instrument from the qcode mocks
+  # note the different path for the driver
```
astafan8: which path? is this comment referring to the `driver`+`type` combination?

qcodes/tests/test_station.py
```diff
@@ -659,6 +668,24 @@ def test_deprecated_limits_keyword_as_string():
...
+instruments:
+  mock:
+    driver: qcodes.tests.instrument_mocks.DummyInstrument
+invalid_keyword:
+  more_errors: 42
+    """)
+
+
+def test_config_validation_comprehensive_config():
+    Station(config_file=os.path.join(
```
astafan8: interesting that there are no traces of QCoDeSDeprecationWarning in the stdout of running pytest coming from the fact that the yaml file contains driver+type entries and parameter limits as a string. i expect those to show up because other warnings to pop up, [for example this](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=4929&view=logs&j=d84cbf5c-d59c-5b28-7c9f-55b9660cb0fb&t=97624319-fa93-5383-4c5b-1ee5f22c3215&l=145). @jenshnielsen , is it all again due to the funky default behaviour of warnings in python?

to clarify, my expectation was that this call would be wrapped in smth like "ignore warnings".

qcodes/dist/schemas/station.schema.json
```diff
@@ -0,0 +1,176 @@
...
+                ""
+            ]
+        },
+        "parameter_base": {
+            "$comment": "A common base for the properties shared by 'add_parameters' and 'parameters can first be implemented with the new json-schema draft 2019-09, for which there is no implementation at this point. The new feature is called 'unevaluatedProperties'."
+        }
+    },
+    "properties": {
+        "instruments": {
+            "description": "Instruments defined in this file. Valid names are all alphanumerical names including '-' and '_'.",
```
astafan8: hyphens cannot be part of a python identifier. or am i misunderstanding something?

qcodes/dist/schemas/station.schema.json
```diff
@@ -0,0 +1,176 @@
...
+    "description": "This schema serves the purpose of validating yaml files that describe a QCoDeS station. See more here: https://qcodes.github.io/Qcodes/examples/Station.html#Configuring-the-Station-by-using-a-YAML-configuration-file",
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "definitions": {
+        "instruments": {
+            "enum": [
+                ""
+            ]
+        },
+        "parameter_base": {
+            "$comment": "A common base for the properties shared by 'add_parameters' and 'parameters can first be implemented with the new json-schema draft 2019-09, for which there is no implementation at this point. The new feature is called 'unevaluatedProperties'."
```
astafan8: this comment is confusing. what can exactly "can first be implemented with the new json schema"?

Dominik-Vogel:
Thank you Mikhail for taking the time for a thorough review. I mistakenly staged the wrong example file. In the original I had stripped all the comments and fixed the usage of the driver keyword. It is indeed strange that the warnings are not shown. But I suspect that this is because pytest does not show the stdout when tests succeed, right?

qcodes/tests/test_station.py
Dominik-Vogel: Pytest does not show all warnings by default. Here we are raising a `QCoDeSDeprecationWarning` which is a `RuntimeWarning`. It does not get shown by default. We should consider making these warnings errors by adding the `-W` parameter or changing the pytest.ini 
```
[pytest]
filterwarnings =
    error::QCoDeSDeprecationWarning
```
(see here https://docs.pytest.org/en/latest/warnings.html)
Then we have to add
```
@pytest.mark.filterwarnings("default::QCoDeSDeprecationWarning")
```
to the tests that expect warnings.

qcodes/dist/schemas/station.schema.json
Dominik-Vogel: Thank you for putting it so gallantly. To my shame I did not know that hyphens where not allowed in python identifiers.

qcodes/dist/schemas/station.schema.json
Dominik-Vogel: During my short affair with json-schemas I learned about how their design is a beautiful craftsmanship, just as writing beautiful code. And also here the DRY principle holds true. Currently the schema repeats the same elements for `parameters` and `add_parameters`. This can be refactored with a feature that comes with the new standard that has been released last month, but has not been implemented by the libraries.

astafan8: thanks for the clarification.

Dominik-Vogel:
@astafan8 @jenshnielsen besides some bug fixes I added:
- documentation in station.ipynb
- generation of a 'user' scheme that includes all the instruments of provided modules for autocompletion. The schema lives under ~/.qcodes/schemas
- script function to register 'user' schema with visual studio code

qcodes/utils/installation.py
```diff
@@ -0,0 +1,40 @@
...
+    You can alternatively access this
+    setting via File->Preferences->Settings and search for `yaml.schemas`.
+
+    To enable autocompletinon of QCoDeS instrument from additional packages run
+    `qcodes.station.update_config_schema`.
+
+    For more information consult `qcodes/docs/examples/Station.ipynb`.
+    """
+    if not os.path.exists(SCHEMA_PATH):
+        update_config_schema()
```
astafan8: i have a feeling that this "side effect" should not be part of this function. The only thing i'd expect it to do is just to raise an exception if `SCHEMA_PATH` doesn't exist. I think that implicitly updating the schema here is probably user-friendly but it does upgrade the schema "silently" may not be what user want to happen. But i'm not sure, just raising a feeling.

qcodes/dist/tests/station/example.station.yaml
```diff
@@ -0,0 +1,59 @@
+instruments:
```
astafan8: [the station notebook also contains an example with comments and stuff](https://nbviewer.jupyter.org/github/QCoDeS/Qcodes/blob/f24eb33b6a078cc8d8d408168dc809fbe7170128/docs/examples/Station.ipynb#Example-of-YAML-Station-configuration) - to avoid duplication, would it make sense to move that example into this file, and render this file in the notebook?

qcodes/tests/test_station.py
```diff
@@ -230,6 +239,11 @@ def test_station_after_instrument_is_closed():
                                        'station'):
         station.remove_component('bob')
 
+
+def test_update_config_schema():
+    update_config_schema()
```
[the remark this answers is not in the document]
Dominik-Vogel: That is an excellent idea!

qcodes/dist/tests/station/example.station.yaml
```diff
@@ -0,0 +1,59 @@
...
+      ch01.v:
+        alias: mygate
+      ch03.v: {alias: Q1lplg1, label: my label}
+      ch04.v: {monitor: true}
+  dmm1:
+    type: qcodes.instrument_drivers.agilent.Agilent_34400A.Agilent_34400A
+    enable_forced_reconnect: true
+    address: GPIB::1::65535::INSTR
+    init:
+      visalib: 'Agilent_34400A.yaml@sim'
```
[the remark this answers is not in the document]
Dominik-Vogel: I think this dates back from a time when I was writing the lakeshore driver sim and it was not part of qcodes yet... Well spotted, I fixed it to using the partial path.

qcodes/utils/installation.py
```diff
@@ -0,0 +1,40 @@
...
+    config_path = os.path.expandvars(
+        os.path.join('%APPDATA%', 'Code', 'User', 'settings.json'))
+    with open(config_path, 'r+') as f:
+        data = json.load(f)
+    data.setdefault(
+        'yaml.schemas', {}
+    )[r'file:\\' + os.path.splitdrive(SCHEMA_PATH)[1]] = '*.station.yaml'
```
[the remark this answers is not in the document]
Dominik-Vogel: It might not be. Feel free to test if you have a mac handy.

qcodes/tests/test_station.py
```diff
@@ -659,6 +673,24 @@ def test_deprecated_limits_keyword_as_string():
       ch1:
         limits: -10, 10
     """)
-    with warnings.catch_warnings(record=True) as w:
+    with warnings.catch_warnings(record=True) as ws:
         st.load_instrument('mock')
-    assert issubclass(w[-1].category, QCoDeSDeprecationWarning)
+    assert any(issubclass(w.category, QCoDeSDeprecationWarning)
```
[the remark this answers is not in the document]
Dominik-Vogel: You caught me being lazy ;-). I'll fix it.

qcodes/utils/installation.py
Dominik-Vogel: I have the same feeling as you have here. But let me try to explain my thoughts and my attempt of being pragmatic:
For me it would be most preferable if that function would live somewhere outside of qcodes, because it is just a very specific helper script. Yet, I would want it to be distributed easily to all stations, and the community can also benefit from it. I think putting it into the `utils` folder is a good compromise (even though many other modules here do not qualify as utils...)
It does not really upgrade any schema, it only creates one if none has been there. I think the whole thing is bare convenience and we could thus leave it as it is, but we should extend the docstring to capture that side effect.
But as said above I don't have a strong preference for that solution.

qcodes/dist/tests/station/example.station.yaml
Dominik-Vogel: I have not found a way how to do this nicely. And as divergence is not of an big issue here, I'd suggest: let's leave it for now.

docs/examples/Station.ipynb
```diff
@@ -527,11 +528,11 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 12,
    "metadata": {},
    "outputs": [],
    "source": [
-    "from qcodes.utils.installation import register_station_schema_with_vscode\n",
+    "# from qcodes.utils.installation import register_station_schema_with_vscode\n",
```
astafan8: commented out?

qcodes/utils/installation.py
astafan8: agree. let's leave as is.

docs/examples/Station.ipynb
Dominik-Vogel: Noooo...  ;-( That was a leftover from playing with different options to inline a yaml file.

qcodes/station.py
```diff
@@ -538,3 +572,62 @@ def update_monitor():
...
+
+
+def update_config_schema(
+    additional_instrument_modules: Optional[List[ModuleType]] = None
+) -> None:
+    """Update the json schema file 'station.schema.json'.
+
+    Args:
+        additional_instrument_modules: python modules that contain
+            :class:`qcodes.instrument.InstrumentBase` definitions
```
astafan8:
```suggestion
            :class:`qcodes.instrument.base.InstrumentBase` definitions
```

qcodes/station.py
```diff
@@ -355,11 +378,22 @@ def update_load_instrument_methods():
...
                 else:
-                    log.warning(f'Invalid identifier: ' +
-                                f'for the instrument {instrument_name} no ' +
-                                f'lazy loading method {method_name} could ' +
+                    log.warning(f'Invalid identifier: '
+                                f'for the instrument {instrument_name} no '
+                                f'lazy loading method {method_name} could '
                                 'be created in the Station.')
-        self._config = YAML().load(config)
+        yaml = YAML().load(config)
```
jenshnielsen: Could you add a comment here explaining why this uses the template for future reference

qcodes/station.py
```diff
@@ -538,3 +575,62 @@ def update_monitor():
...
+                if (issubclass(instr[1], InstrumentBase) and
+                    instr[1].__module__.startswith(module.__name__))
+            ]
+            res.update(new_members)
+        return tuple(res)
+
+    def update_schema_file(
+        template_path: str,
+        output_path: str,
+        instrument_names: Tuple[str]
```
jenshnielsen: Tuple[str, ...]

PR #1760, opened 2019-10-11
Station config: rename `Driver` to `Type`

Dominik-Vogel:
In PR #1753 the type keyword has been removed in favor of the driver keyword. This PR changes this fusion of the two keywords to only using the `type` keyword. The reason for that is that `type` is more general and will be less confusing when allowing to use the config file to create channels with the same syntax as instruments.


jenshnielsen:
@Dominik-Vogel Something strange happened to the whitespace/newlines in the notebook. Otherwise this looks good. Could you write a note in the changelog summarizing the changes you made to the station format in this and the 2 other prs

Dominik-Vogel:
@jenshnielsen the whitespaces must have been the new vscode version. I can try to find some time tonight to load it in jupyter and resave it as well as fixing the backlog.

jenshnielsen:
@Dominik-Vogel I recreated the notebook and am happy to add the changelog.

Dominik-Vogel:
Thank you guys for fixing this!

PR #1763, opened 2019-10-15
AimTTi PL601-P Power Supply Driver

GateBuilder:
This PR introduces the QCoDeS driver for the AimTTi PL601-P programmable power supply.

@WilliamHPNielsen


jenshnielsen:
The docs build fails because it tries to execute the notebook. You will need to mark the notebook as not executed during docs build as noted here https://qcodes.github.io/Qcodes/examples/writing_drivers/Creating-Instrument-Drivers.html#Documentation

qcodes/instrument_drivers/AimTTi/AimTTi_PL601P_channels.py
```diff
@@ -0,0 +1,307 @@
...
+class NotKnownModel(Exception):
+    pass
+
+class AimTTiChannel(InstrumentChannel):
+    """
+    This is the class that holds the output channels of AimTTi power
+    supply.
+    """
+    def __init__(self, parent: Instrument, name: str,
+                 channel: str, **kwargs) -> None:
```
jenshnielsen: channel should probably be an int ?

qcodes/instrument_drivers/AimTTi/AimTTi_PL601P_channels.py
```diff
@@ -0,0 +1,307 @@
...
+
+        channels = ChannelList(self, "Channels", AimTTiChannel,
+                               snapshotable=False)
+
+        _model = self.get_idn()['model']
+
+        _numOutputChannels = {'PL068-P': 1, 'PL155-P': 1, 'PL303-P': 1,
+                             'PL601-P': 1, 'PL303QMD-P': 2, 'PL303QMT': 3}
+
+        if not _model in _numOutputChannels.keys():
```
jenshnielsen: I think you also need to check that explicitly that _model is not none here to make mypy happy since the return type is optional[str]

qcodes/instrument_drivers/AimTTi/AimTTi_PL601P_channels.py
```diff
@@ -0,0 +1,311 @@
...
+            raise NotKnownModel("Unknown model, connection cannot be "
+                                "established.")
+
+        if _model is None:
+            raise RuntimeError("Model name cannot be empty "
+                               "and should be string.")
```
WilliamHPNielsen: I don't think this error message is very helpful to the user (they didn't provide the model name). Under which circumstances would this error be encountered? Perhaps a `_model` of `None` should also raise `NotKnownModel`.

qcodes/instrument_drivers/AimTTi/AimTTi_PL601P_channels.py
```diff
@@ -0,0 +1,311 @@
...
+        A bound method that decreases the current output of the corresponding
+        channel by an amount the step size set by the user via ``curr_step_size``
+        parameter.
+        """
+        channel_id = self.channel
+        self.write(f'DECI{channel_id}')
+         # Clear the cache.
+        _ = self._get_current_value()
+
+    def save_setup(self, slote: int) -> None:
```
WilliamHPNielsen: Should `slote` be `slot`?

qcodes/tests/drivers/test_AimTTİ_PL601P.py
```diff
@@ -0,0 +1,30 @@
...
+    driver.close()
+
+
+@pytest.fixture(scope='function')
+def channels(driver):
+    ch_names = {'ch1'}
+    assert ch_names == set(list(driver.submodules.keys()))
+
+    yield tuple(getattr(driver, ch_name)
+                for ch_name in ch_names)
```
WilliamHPNielsen: Is this fixture ever used?

qcodes/instrument_drivers/AimTTi/AimTTi_PL601P_channels.py
```diff
@@ -0,0 +1,311 @@
...
+        if not slote in self.set_up_store_slots:
+            raise RuntimeError("Slote number should be an integer between"
+                               "0 adn 9.")
+
+        channel_id = self.channel
+        self.write(f'RCL{channel_id} {slote}')
+        # Update snapshot after load.
+        _ = self.snapshot(update=True)
+
+    def set_damping(self, val: int=0) -> None:
```
WilliamHPNielsen: Could `damping` not be a parameter? It looks to always have a well-defined state, and it seems relevant for the user to be able to query the state.

Okay, just read the notebook: 
> Finally, the current meter averaging can be turned on and off remotely. As there is no remote query for the status, user should observe the "Meter Average" signal light. 

That sucks. But a parameter with `get_cmd=None` would still be preferable, I think.

WilliamHPNielsen: Even if this should remain a function, I don't think there should be a default value for `val`. How is it obvious that `set_damping()` switches the damping off and not on? "Explicit is better than implicit" 🙂

qcodes/instrument_drivers/AimTTi/AimTTi_PL601P_channels.py
```diff
@@ -0,0 +1,311 @@
...
+        Returns '1' if the interface lock is owned by the requesting instance,
+        '0' if there is no active lock and '-1' if the lock is unavailable.
+        """
+        is_lockedSTR = self.ask_raw('IFLOCK?')
+        is_locked = is_lockedSTR.strip()
+        return int(is_locked)
+
+    def lock_interface(self) -> int:
+        """
+        Requests instruement interface lock. Returns '1' if successful and
```
WilliamHPNielsen: `instruement` -> `instrument`

qcodes/instrument_drivers/AimTTi/AimTTi_PL601P_channels.py
```diff
@@ -0,0 +1,311 @@
...
+        '-1' if the lock is unavailable.
+        """
+        lockSTR = self.ask_raw('IFLOCK')
+        lock = lockSTR.strip()
+        return int(lock)
+
+    def unlock_interface(self) -> int:
+        """
+        Requests the release of instruement interface lock. Returns '0'
```
WilliamHPNielsen: `instruement` -> `instrument`

docs/examples/driver_examples/QCoDeS example with Aim TTi PL601-P.ipynb
```diff
@@ -0,0 +1,903 @@
...
+   ],
+   "source": [
+    "tti.ch1.curr_range()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Here, we note that the output must be switched off before changing the current range. This is automatically handled by the QCoDeS driver. However, driver does not automatically switch on the output after the new current range is set. This should be handled by the user."
```
WilliamHPNielsen: I would argue that the driver should handle the output (see my comment below)

qcodes/instrument_drivers/AimTTi/AimTTi_PL601P_channels.py
```diff
@@ -0,0 +1,311 @@
...
+    def decrement_volt_by_step_size(self) -> None:
+        """
+        A bound method that decreases the voltage output of the corresponding
+        channel by an amount the step size set by the user via ``volt_step_size``
+        parameter.
+        """
+        channel_id = self.channel
+        self.write(f'DECV{channel_id}')
+         # Clear the cache.
+        _ = self._get_voltage_value()
```
WilliamHPNielsen: Does this update the `volt` parameter? If I do 
```python
aim.volt(4)
aim.decrement_by_step_size()
print(aim.volt.get_latest())
```
what do I then get? I think the `decrement_volt_by_step_size` should instead call `self.volt.get()`.

qcodes/tests/drivers/test_AimTTİ_PL601P.py
```diff
@@ -0,0 +1,30 @@
...
+
+@pytest.fixture(scope='function')
+def driver():
+    driver = AimTTi('AimTTi', address='GPIB::1::INSTR', visalib=visalib)
+
+    yield driver
+    driver.close()
+
+
+@pytest.fixture(scope='function')
```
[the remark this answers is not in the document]
GateBuilder: As there is a duplicate, I am resolving this one.

qcodes/instrument_drivers/AimTTi/AimTTi_PL601P_channels.py
```diff
@@ -0,0 +1,311 @@
...
+        return float(_c_step_size_split[1])
+
+    def _set_current_range(self, val: int) -> None:
+        """
+        This is the private function that ensures that the output is switched
+        off before changing the current range, as pointed out by the instrument
+        manual.
+        """
+        channel_id = self.channel
+        self.output(False)
```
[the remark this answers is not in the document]
GateBuilder: I thought about this while writing the driver, i.e., automatically turn on the output after changing the range. Indded, this context manager handles this elegantly. However, while it is safe from the software and as well as aim tti hardware sides, such an automation eliminates the optional time interval for the user to double check its set up before any connection with other hardware for which the driver has no control over.  I recommend that this should stay as it is.

qcodes/instrument_drivers/AimTTi/AimTTi_PL601P_channels.py
GateBuilder: This is a pretty good catch! 💯  Fixed. :)

docs/examples/driver_examples/QCoDeS example with Aim TTi PL601-P.ipynb
GateBuilder: See my comment above.

qcodes/instrument_drivers/AimTTi/AimTTi_PL601P_channels.py
GateBuilder: I think you are right that there should not be a default value, I removed it. However, it should stay as a function due to in fact that how the instrument works. For this and, for example, increment and decrement by step size properties, I have first created parameters which failed. They only work if they are methods of the class to which we pass the corresponding values as arguments.

qcodes/tests/drivers/test_AimTTİ_PL601P.py
GateBuilder: Currently not. I will put a small test to use the fixture for the time being, as I want to keep it for future. I will figure out pytest-sims issue and complete the tests eventually.

GateBuilder: Ok I removed it, did not make any sense to have it there.

WilliamHPNielsen:
@GateBuilder would you mind revealing in some more detail what happens in the following case:
 if I set the step size to, say, 0.5 V, what then happens if I set `aim.volt(aim.volt() + 0.1)` ?

WilliamHPNielsen:
@GateBuilder sweet! Thanks for explaining.

PR #1777, opened 2019-10-21
Fix example notebooks: Context Manager, Dataset Performance and Extract Runs

GateBuilder:
This PR proposes some fixes to the example notebooks:

- The Context Manager
- Dataset Performance
- Extracting runs from one DB file to another

@Dominik-Vogel


docs/examples/DataSet/Dataset Context Manager.ipynb
```diff
@@ -131,10 +140,10 @@
     "\n",
```
lakhotiaharshit: Should we link the example notebook for the `The Experiment Container` here? http://qcodes.github.io/Qcodes/examples/DataSet/The-Experiment-Container.html

docs/examples/DataSet/Dataset Context Manager.ipynb
```diff
@@ -131,10 +140,10 @@
...
     "Furthermore, datasets are associated to an experiment. By default the run\n",
     "is appended to the latest existing experiments. If no experiment has been created,\n",
-    "we must create one. We do that by calling `load_or_create_experiment`\n",
+    "we must create one. We do that by calling `load_or_create_experiment` method.\n",
```
astafan8:
```suggestion
    "we must create one. We do that by calling `load_or_create_experiment` function.\n",
```

docs/examples/DataSet/Dataset Context Manager.ipynb
```diff
@@ -131,10 +140,10 @@
...
     "\n",
-    "Here we explicitly pass the loaded or created experiment to the `Measurement` to ensure that we are always\n",
-    "useing the `dataset_context_manager` experiment throughout this tutorial."
+    "Here we explicitly pass the loaded or created experiment to the `Measurement` object to ensure that we are always\n",
+    "using the `dataset_context_manager` throughout the experiment introduced within this tutorial."
```
astafan8: here, the point is in referencing the specific experiment object.
```suggestion
    "using the `dataset_context_manager` `Experiment` created within this tutorial."
```

docs/examples/DataSet/Dataset Context Manager.ipynb
```diff
@@ -1003,14 +1012,14 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The dataset implements a number of methods for accessing the data of a given dataset. Here we will concentrate on the two most user friendly methods."
+    "QCoDeS ``dataset`` implements a number of methods for accessing the data of a given dataset. Here we will concentrate on the two most user friendly methods."
```
astafan8: should here be just the instance or the class?
```suggestion
    "QCoDeS ``DataSet`` implements a number of methods for accessing the data of a given dataset. Here we will concentrate on the two most user friendly methods."
```

docs/examples/DataSet/Extracting-runs-from-one-DB-file-to-another.ipynb
```diff
@@ -6,7 +6,7 @@
    "source": [
     "# Extracting runs from one DB file to another\n",
     "\n",
-    "This notebook shows how to use the `extract_runs_into_db` function to extract runs from DB file (the source DB) into another DB file (the target DB). If the target DB does not exist, it will be created. The runs are **NOT** removed from the original DB file; they are copied over.\n"
+    "This notebook shows how to use the `extract_runs_into_db` method to extract runs from a database (DB) file (the source DB) into another DB file (the target DB). If the target DB does not exist, it will be created. The runs are **NOT** removed from the original DB file; they are copied over.\n"
```
astafan8: as per python definitions, it's a function, not a method :)
```suggestion
    "This notebook shows how to use the `extract_runs_into_db` function to extract runs from a database (DB) file (the source DB) into another DB file (the target DB). If the target DB does not exist, it will be created. The runs are **NOT** removed from the original DB file; they are copied over.\n"
```

GateBuilder: :)

docs/examples/DataSet/Dataset Context Manager.ipynb
```diff
@@ -7,7 +7,9 @@
     "# The Context Manager aka the Measurement Object\n",
     "\n",
     "This notebook shows some ways of performing different measurements using \n",
-    "QCoDeS parameters and the new DataSet accessed via a context manager.\n"
+    "QCoDeS parameters and the new DataSet accessed via a context manager. Here, it is assumed that the reader has some degree of familiarity with fundamental objects and methods of QCoDeS.\n",
```
Dominik-Vogel: When adding some requirements to the reader here, I would be more specific, just along the lines as Harshit is saying: name the concepts required to understand and link to the documents where they can be learned.

docs/examples/DataSet/Dataset Context Manager.ipynb
```diff
@@ -131,10 +140,10 @@
     "\n",
     "Furthermore, datasets are associated to an experiment. By default the run\n",
     "is appended to the latest existing experiments. If no experiment has been created,\n",
-    "we must create one. We do that by calling `load_or_create_experiment`\n",
+    "we must create one. We do that by calling `load_or_create_experiment` function.\n",
```
Dominik-Vogel: I preferred the previous version. It is irrelevant whether `load_or_create_experiment` is a function or any other kind of `Callable`.

docs/examples/DataSet/Dataset Context Manager.ipynb
```diff
@@ -1098,7 +1107,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The data can also be exported as one or more [Pandas](https://pandas.pydata.org/) DataFrames. The DataFrames are returns as a dict from measured parameters to DataFrames."
+    "The data can also be exported as one or more [Pandas](https://pandas.pydata.org/) DataFrames. The DataFrames returns as a dictionary from measured parameters to DataFrames."
```
Dominik-Vogel: I think the original intended to be:
The DataFrames are returned as a....

docs/examples/DataSet/Dataset Context Manager.ipynb
```diff
@@ -1221,15 +1230,15 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "To load existing datasets QCoDeS provides several functions. The most useful and generic is `load_by_run_spec`. \n",
+    "To load existing datasets QCoDeS provides several functions. The most useful and generic method is called `load_by_run_spec`. \n",
```
Dominik-Vogel: `load_by_run_spec`  is not a method but a function.

docs/examples/DataSet/Dataset Context Manager.ipynb
```diff
@@ -6801,7 +6810,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Unlike the data measured on a grid above all the measured data points have an unique combination of the two dependent parameters. When exporting to XArray NaN's will therefore replace all the missing combinations of `dac_ch1` and `dac_ch2` and the data is unlikely to be useful in this format. "
+    "Unlike the data measured on a grid, all the measured data points have an unique combination of the two dependent parameters. When exporting to XArray NaN's will therefore replace all the missing combinations of `dac_ch1` and `dac_ch2` and the data is unlikely to be useful in this format. "
```
Dominik-Vogel: Maybe:
Unlike the data measured above, which lies on a grid, here, all the ...
(not sure if this was the intention)

docs/examples/DataSet/Dataset_Performance.ipynb
```diff
@@ -329,7 +329,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "As usual you can load the data using load_by_id but you will notice that the different storage methods\n",
+    "As usual you can load the data using ``load_by_id`` method but you will notice that the different storage methods\n",
```
Dominik-Vogel: Missing article for `load_by_id`. Method->function.
(Maybe we need rewrite to make `load_by_spec` the default, @jenshnielsen what do you think?)

docs/examples/DataSet/Dataset Context Manager.ipynb
GateBuilder: Well, I would prefer not doing that as every required concept will surely have some additional dependencies and prerequisities, as well. Having a general statement or none (as in the previous version) is sufficient, IMHO. Harshit's above point is OK because the concept is explicitly referred and thus should be linked :)

docs/examples/DataSet/Dataset Context Manager.ipynb
GateBuilder: From software development point of view, yes, I agree with you. But this is a document? We are not supposed to follow some coherency in the English? I can not, personally, treat documents as code. It should be readable.

astafan8: @Dominik-Vogel i think that in this particular case the difference is subtle. However, for other cases I think it's good to mention whether a "symbol" is a function or some method of some class because this helps the beginner reader to find where the symbol is.

Dominik-Vogel: I am fine with this. (We can add a 'the')

docs/examples/DataSet/Dataset Context Manager.ipynb
Dominik-Vogel: I am fine with this, you have the better overview, Umit.

docs/examples/DataSet/Dataset_Performance.ipynb
```diff
@@ -79,7 +79,7 @@
    "metadata": {},
```
lakhotiaharshit: >from qcodes.dataset.experiment_container import (Experiment,,
                                                     load_last_experiment,                                                
                                                        new_experiment)

I think these imports are not used anywhere in the notebook.

GateBuilder: As we mention about the container, I prefer these imports to stay.

docs/examples/DataSet/Extracting-runs-from-one-DB-file-to-another.ipynb
```diff
@@ -408,7 +408,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "And the original run_id is preserved as `captured_run_id` "
+    "And the original ``run_id`` is preserved as `captured_run_id` "
```
astafan8: single quotes is enough. right?
```suggestion
    "And the original `run_id` is preserved as `captured_run_id` "
```

docs/examples/DataSet/Dataset Context Manager.ipynb
```diff
@@ -127,14 +136,14 @@
     "\n",
     "If this is the first time you create a dataset, the underlying database file has\n",
     "most likely not been created. The following cell creates the database file. Please\n",
-    "refer to documentation on `The Experiment Container` for details.\n",
+    "refer to documentation on [`The Experiment Container`](http://qcodes.github.io/Qcodes/examples/DataSet/The-Experiment-Container.html) for details.\n",
```
astafan8: please use link to the notebook within the repo
```suggestion
    "refer to documentation on [`The Experiment Container`](The-Experiment-Container.ipynb) for details.\n",
```

PR #1779, opened 2019-10-22
Remove outdated documents: Tutorial

GateBuilder:
This PR proposes to remove

- http://qcodes.github.io/Qcodes/user/tutorial.html

due to 

1. The content is outdated.
2. The content that is relevant to legacy code does already exists in:

- http://qcodes.github.io/Qcodes/examples/legacy/Tutorial.html
- http://qcodes.github.io/Qcodes/examples/legacy/Combined%20Parameters.html

I have checked if there are any pointers to this file to ensure there are no broken links. I could not locate anything. 

@jenshnielsen


jenshnielsen:
Thanks. I think you need to remove the tutorial from the index.rst file in the same folder

WilliamHPNielsen:
There is a link in this file: https://github.com/QCoDeS/Qcodes/blob/master/docs/user/index.rst

jenshnielsen:
The driver link in the intro.rst file also refers to this document. I suggest building docs locally when making changes before pushing. This will significantly reduce the feedback loop

jenshnielsen:
It's probably because you need to do a make clean before building when you have removed a file. Otherwise it will be in the cache

PR #1781, opened 2019-10-23
Fix broken link to notebook in parameter module

lakhotiaharshit:
Fixes a broken link to the notebook in the parameter module

@astafan8


qcodes/instrument/parameter.py
```diff
@@ -27,7 +27,7 @@
 - :class:`.ParameterWithSetpoints` is intended for array-values parameters.
     This Parameter class is intended for anything where a call to the instrument
     returns an array of values.
-    `This notebook <../examples/writing_drivers/Simple-Example-of-ParameterWithSetpoints.ipynb>`_.
+    `This notebook <../../examples/writing_drivers/A-ParameterWithSetpoints-Example-with-Dual-Setpoints.ipynb>`_.
```
astafan8: shouldn't it be this https://github.com/QCoDeS/Qcodes/blob/master/docs/examples/Parameters/Simple-Example-of-ParameterWithSetpoints.ipynb ?

astafan8: or just both :)

lakhotiaharshit: Maybe both. I was also confused about what would be a better choice.

astafan8: the SimpleExample is definitely the preferred choice. But let's have both.

astafan8:
CI is complaining :)

PR #1784, opened 2019-10-23
Deprecation of moved community instrument drivers

GateBuilder:
This PR contains the deprecation warnings concerning moved community contributed QCoDeS drivers as well as relevant changes in core readme file and QCoDeS documentation.

@WilliamHPNielsen


qcodes/utils/deprecate.py
```diff
@@ -47,3 +47,9 @@ def decorated_func(*args, **kwargs):
             return func(*args, **kwargs)
         return decorated_func
     return actual_decorator
+
+
+deprecate_moved_to_qcd = partial(deprecate, reason="This driver has been moved"
+                                                   " to Qcodes_contrib_drivers"
+                                                   "and will be removed "
+                                                   "from QCoDes eventually")
```
astafan8: newline
```suggestion
                                                   "from QCoDeS eventually")

```

Dominik-Vogel:
In a short discussion Jens and me came to the conclusion that we should best implement deprecation of a class by extending the existing decorator to distinguish between classes and functions/methods, and in case of  a class to add deprecation to every single method.
The reason for that are cases like: Static and class methods, as well as subclasses that may not call the init method.
An alternative is a metaclass, but this created problems with inheritance in the case of the alazar adaptor (not sure if it can be circumvented though).

Dominik-Vogel:
@GateBuilder  do you want to implement the deprecation mechanism from #1799 here? Let me know if you need any help.

Dominik-Vogel:
#1805 should fix at least one of the issues.

PR #1785, opened 2019-10-23
Fix setting parameter from station config

Dominik-Vogel:
This PR fixes a bug reported by @ThorvaldLarsen for setting-up a parameter of a `ChannelList` through the station config file.
Consider the following example:
```yaml
instruments:
  mock:
    type: qcodes.tests.instrument_mocks.DummyChannelInstrument
    parameters:
      channels.temperature:
          initial_value: 10
```
Here `channels` is of type `ChannelList`.


qcodes/station.py
```diff
@@ -52,6 +53,9 @@ def get_config_use_monitor() -> Optional[str]:
     return qcodes.config["station"]["use_monitor"]
 
 
+ChannelOrInstrumentBase = Union[InstrumentBase, ChannelList]
```
astafan8: shouldn't this also be a channel? (not only channel lists) some drivers have "submodules" with parameters implemented via subclassing InstrumentChannel.
```suggestion
ChannelOrInstrumentBase = Union[InstrumentBase, ChannelList, InstrumentChannel]
```

qcodes/station.py
```diff
@@ -456,31 +460,47 @@ def load_instrument(self, identifier: str,
...
-        # local function to refactor common code from defining new parameter
-        # and setting existing one
-        def resolve_parameter_identifier(instrument: InstrumentBase,
-                                         identifier: str) -> Parameter:
+        def resolve_instrument_identifier(
+            instrument: ChannelOrInstrumentBase,
+            identifier: str
+        ) -> ChannelOrInstrumentBase:
+            """
+            Get the instrument or channel_list described by a nested string.
```
astafan8: shouldn't this be both channel_list and channel?

qcodes/station.py
Dominik-Vogel: It is :-)
```python
class InstrumentChannel(InstrumentBase):
```

astafan8: oh that's super good! stupid me :)

Dominik-Vogel:
This was a premature conversion from draft (I could swear I had the tests passing). I'll notify you once its good to go.

Dominik-Vogel:
Ok, now it's good to go, @astafan8 @jenshnielsen

PR #1791, opened 2019-10-23
_BaseParameter: add get_cache* API and use it in GetLatest and in snapshot_base

astafan8:
### Description (to be reworked)
This is to promote "cache" API as opposed to old `raw_value`, `_save_val`, and other hacks around getting `_latest[...]`. The `_latest` dictionary should only be used by `_BaseParameter` methods which allows it to be refactored in the future as needed. `get_cache` mirrors `get_latest` behavior.

Requires https://github.com/QCoDeS/Qcodes/pull/1757 and https://github.com/QCoDeS/Qcodes/pull/1787 to be merged.

### ToDo:
- [x] modify snapshot_base to use get_cache* API
- [x] ensure that `_latest` dictionary is only referred to in `get_cache*` methods and `_update_latest` method (tests are covered by a separate item)
- [x] modify tests to use get_cache* API everywhere and add trivial tests for get_latest stuff (to just mark that get_latest API still exists)
- [x] use get_cache* API everywhere in qcodes instead of `.get_latest.get_raw_value` and `.get_latest.get_timestamp`
- [ ] refactor `snapshot_base` to use `get_latest`, also add `get_if_cache_is_invalid` to `_get_latest` function
- [ ] introduce proper `Cache` class (in this PR?)
- [ ] (?) use get_cache* API everywhere in qcodes drivers
- [ ] (?) use get_cache* API everywhere in qcodes contrib drivers and other drivers


jenshnielsen:
My vote is for get_cached to behave as get_latest does now. I don't see any use cases where the dump behavior is useful but many places where it can lead to errors.

Dominik-Vogel:
@astafan8 should there not also be a `set_cached_raw`?

astafan8:
> should there not also be a set_cached_raw?

i decided not to add it because the interaction of users with the parameter is intended to happen on the "value" level because a parameter is intended to hide the "raw" stuff (which usually is interaction with an instrument). If you see a use case for this that can't be sorted out via set_cache, let's definitely consider.

astafan8:
> My vote is for get_cached to behave as get_latest does now. I don't see any use cases where the dump behavior is useful but many places where it can lead to errors.

@jenshnielsen then what should be the way to access the cached value of the parameter in [the refactored _get_latest function](https://github.com/QCoDeS/Qcodes/pull/1791/files#diff-7ab35e1b5092c8dd21f4bb9883f61a76R155)?

also, happy to hear about "many places where it can lead to errors" because i'd expect if the difference between get_cache and get_latest is documented properly, there won't be problems.

jenshnielsen:
> @jenshnielsen then what should be the way to access the cached value of the parameter in the 
> refactored _get_latest function?
You don't implement a public api just for the sake of internal use. 

Sure we can handle this by documentation. First step would be not to call this `get_cached` but something that better signals its problematic behavior

Dominik-Vogel:
@astafan8 on 'set_cache_raw' : fair enough :-)

Dominik-Vogel:
We should also implement setter and getter methods for max val age, so that it can be forwarded to the DelegateParameter.

PR #1793, opened 2019-10-28
test on travis using 3.8 release not dev branch

jenshnielsen:
Now that 3.8 is out we should no longer need to use the dev branch


jenshnielsen:
Fixed a few deprecation warnings too. There are 2 more that could be fixed but those are on instruments that will be moved to qcodes_contrib_drivers

PR #1799, opened 2019-10-31
Enable the use of the deprecation decorator with classes

Dominik-Vogel:
This PR allows to use the deprecation decorator for classes.
When a class is decorated all its methods will be deprecated.
This allows deprecation on run time and not import time (which we want)
It is important to not simply decorate the in init method for cases such as
- subclass never calls `super().__init__`
- static- and class- methods

The deprecation still fails for some cases:
- nested classes
- use of class variables
- properties
- setters

I suggest we document these limitations and continue with more important tasks.

This PR also adds the handy context managers `assert_deprecated("message")` and `assert_not_deprecated` for writing tests for deprecation warnings.

The code of this PR can be used in #1784 .


qcodes/utils/deprecate.py
```diff
@@ -46,7 +49,7 @@ def deprecate(
         alternative: The alternative function or class to put in use instead of
             the deprecated one.
     """
-    def actual_decorator(func: Callable) -> Callable:
+    def decorate_callable(func: Callable):
```
astafan8: shall we do smth like mentioned here https://mypy.readthedocs.io/en/latest/generics.html#declaring-decorators ? or is it too much for this PR?

```python
FuncType = Callable[..., Any]
F = TypeVar('F', bound=FuncType)

# A decorator that preserves the signature.
def my_decorator(func: F) -> F:
```

qcodes/utils/deprecate.py
```diff
@@ -55,4 +58,46 @@ def decorated_func(*args, **kwargs):
...
+        yield
+    assert len(ws) == 0
+
+
+@contextmanager
+def assert_deprecated(message: str):
+    with _catch_deprecation_warnings() as ws:
+        yield
+    assert len(ws) == 1
+    assert (ws[0].message.args[0] == message)
```
astafan8:
```suggestion
    assert ws[0].message.args[0] == message
```

qcodes/utils/deprecate.py
```diff
@@ -55,4 +58,46 @@ def decorated_func(*args, **kwargs):
...
+            for m_name in dir(obj):
+                m = getattr(obj, m_name)
+                if isinstance(m, (types.FunctionType, types.MethodType)):
+                    setattr(obj, m_name, decorate_callable(m))
+            return obj
+
     return actual_decorator
+
+
+deprecate_moved_to_qcd = partial(
```
astafan8: belongs to the other PR, right?

qcodes/tests/test_deprecate.py
```diff
@@ -27,8 +26,47 @@ def _add_one(x):
...
+
+def test_deprecated_class():
+
+    @deprecate(reason='this is a test')
+    class C:
+        a = 'pristine'
+
+        def method(self):
+            self.a = 'last called by method'
+
```
astafan8: for the sake of sanity-checking, shall classes with explicit `__init__`, `@property` also be tested?

qcodes/tests/test_deprecate.py
```diff
@@ -27,8 +26,47 @@ def _add_one(x):
...
+        warnings.warn('Some other warning')
+    assert len(ws) == 2
+    with warnings.catch_warnings(record=True) as ws:
+        issue_deprecation_warning('something')
+        warnings.warn('Some other warning')
+    assert len(ws) == 2
+
+
+def test_deprecated_class():
+
```
astafan8: > The deprecation still fails for some less common cases:
> 
> - nested classes
> - use of class variables

Shall this also be captured in tests?

qcodes/utils/deprecate.py
Dominik-Vogel: This is nice! yes, I think we should!

qcodes/tests/test_deprecate.py
Dominik-Vogel: good idea!

qcodes/utils/deprecate.py
Dominik-Vogel: yes, If you insist, I'll remove it ;-)

qcodes/tests/test_deprecate.py
Dominik-Vogel: I think this would be overkill.

astafan8: ok :(

qcodes/utils/deprecate.py
Dominik-Vogel: Does not work as expected :-( leaving this for now.

qcodes/tests/test_deprecate.py
Dominik-Vogel: So property does not work. I suggest we document it and move on to more pressing issues.

astafan8: agree! (please keep the failing test ;) )

PR #1800, opened 2019-11-01
Feature: measurement progress printer

mgunyho:
This contains a small class to monitor the progress of a measurement by adding a subscriber, like so:
```py
from qcodes.utils.measurement_subscribers import TextProgress

meas = qcodes.Measurement()

meas.register_custom_parameter("x")
meas.register_custom_parameter("y", setpoints=("y",))

xs = np.linspace(-1, 1, 2000)

meas.add_subscriber(TextProgress(total_measurements=xs.size), state=None)

with meas.run() as datasaver:
    for x in xs:
        meas.add_result(("x", x), ("y", np.random.randn())) 
        time.sleep(0.01)
```
This prints something like
```
105 / 2000 (5.2%) - elapsed: 0:00:01 - ETA: 0:00:19 (Fri Oct 11 12:34:56)
```
It's nothing fancy, but I think something like this should be built-in.

Some things to consider:
- Is it a good idea to have a full module for this? Or should it just go to `helpers`?
- This should be mentioned in some of the example notebooks. I think it could be added to the `Dataset context manager` notebook as a third example, to the `Example measurements with real instruments`, and maybe to the `15 Minutes to Qcodes` notebook. Do you have other suggestions?
- The start and end times are a bit fudgy, since the subscriber function has no real way of knowing if it is being called on the first or last insertion (maybe the subscriber API could be extended)
- With an `ArrayParameter`/`MultiParameter`, the array sizes need to be taken into account in `total_measurements`, this should be probably mentioned somewhere

@astafan8

### DIFF
--- a/qcodes/tests/dataset/test_measurement_context_manager.py
+++ b/qcodes/tests/dataset/test_measurement_context_manager.py
@@ -760,7 +760,11 @@ def test_example_subscriber_text_progress(capsys, experiment, DAC, DMM, N):
                     percent = f'{(i + 1) / N * 100 :.1f} %'
                     ptn = rf'{i + 1} / {N} \({percent}\) - elapsed: 0:00:0\d'
                     if i > 0:
-                        ptn += r' - remaining: 0:00:0\d'
+                        ptn += r' - ETA: 0:00:0\d'
+                        if variant == 2:
+                            ptn += r' \(\w+ \w+ \d\d \d\d:\d\d:\d\d\)'
+                        elif variant == 3:
+                            ptn += r' \(\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d\)'
 
                 cap = capsys.readouterr()
 
--- a/qcodes/tests/dataset/test_measurement_context_manager.py
+++ b/qcodes/tests/dataset/test_measurement_context_manager.py
@@ -24,6 +24,7 @@ from qcodes.dataset.legacy_import import import_dat_file
 from qcodes.dataset.data_set import load_by_id
 from qcodes.instrument.parameter import expand_setpoints_helper
 from qcodes.utils.validators import Arrays, ComplexNumbers, Numbers
+from qcodes.utils.measurement_subscribers import TextProgress
 # pylint: disable=unused-import
 from qcodes.tests.dataset.temporary_databases import (empty_temp_db,
                                                       experiment)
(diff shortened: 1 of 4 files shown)


lakhotiaharshit:
Indeed this is a very useful functionality. But just from your example, I feel a similar result can be achieved by using [tqdm](https://tqdm.github.io/). Am I missing something important here?

mgunyho:
I tried tqdm at some point. My first issue was that it didn't handle nested loops properly, although the `tqdm_notebook` widget which I hadn't seen earlier seems to work well with that. The second problem I had was that for long measurements, I'd like to have an estimate of the finish time as a date, so I don't have to do so much mental arithmetic to figure out what is the current time plus 14 hours. This doesn't seem to be easy to do with the tqdm formatting options, at least at a quick glance.

qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -713,6 +714,68 @@ def sub_get_y_vals(results, length, state):
...
+                    ptn = rf'{i + 1} - elapsed: 0:00:0\d'
+                else:
+                    percent = f'{(i + 1) / N * 100 :.1f} %'
+                    ptn = rf'{i + 1} / {N} \({percent}\) - elapsed: 0:00:0\d'
+                    if i > 0:
+                        ptn += r' - ETA: 0:00:0\d'
+                        if variant == 2:
+                            ptn += r' \(\w+ \w+ \d\d \d\d:\d\d:\d\d\)'
+                        elif variant == 3:
+                            ptn += r' \(\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d\)'
```
astafan8: this is an impressive amount of effort, to assert stdout with regexps. However, i think it's more useful to inject something into `TextProgress` which would just collect all the "messages" that it wants to print to the screen in some structured format (for example, a dictionary), and then the test would assert the contents of that dictionary. Such an injection in the dirty way can be done using magic mock (i'd recommend to rewrite the `__call__` method to create such a structured "message" and then pass it to some `_print_progress_message` method, which can be magic-mocked to also store the given "messages" somewhere where they can be asserted by the test method). I also do think that despite the use of a mock or similar things, the test readability will increase (as opposed to trying to make sense of the regexps).

qcodes/utils/measurement_subscribers.py
```diff
@@ -0,0 +1,93 @@
...
+``Measurement`` instance with ``add_subscriber``.
+"""
+
+import time
+import datetime
+from typing import (Optional, List, Tuple, Any, Union, MutableSequence,
+                    MutableMapping)
+
+
+class TextProgress:
```
astafan8: by looking at this code i strongly recommend to dig deeper into `tqdm` and instead of reinventing it - extend it when needed (which i don't think that extending is needed per se). this is how to extend tqdm - https://github.com/tqdm/tqdm#description-and-additional-stats , this is about nested progress bars - https://github.com/tqdm/tqdm#nested-progress-bars (their support has reasonable constraints), and jupyter integration is awesome - https://github.com/tqdm/tqdm#ipythonjupyter-integration .

and perhaps this little post will help finding a good way to display progress of the measurement - https://github.com/tqdm/tqdm/wiki/How-to-make-a-great-Progress-Bar

As a bonus, qcodes already has tqdm as a dependency :)

mgunyho:
Thank you for the insightful comments! You are right, I should look into tqdm more carefully instead of reinventing the wheel. I suppose that the takeaway here is that tqdm should be mentioned in the documentation somewhere. I will close this PR now, and maybe I'll open a new one for updating the docs.

> I agree with the "Dataset context manager notebook as a third example"

Do you suggest that tqdm should be adapted to work with the subscriber API? I tried this also, but had problems with the output formatting, although that was without the notebook integration. Or should the example show using tqdm just like
```py
with meas.run() as datasaver:
    for x in tqdm_notebook(xs):
        datsaver.add_result( ... )
```
in which case it's a bit odd that it would be part of a tutorial on using subscribers.

astafan8:
> I will close this PR now, and maybe I'll open a new one for updating the docs.

ok!

> Do you suggest that tqdm should be adapted to work with the subscriber API?

yes, in case you still intend to implement smth like `ProgressBar` subscriber that uses tqdm behind the scenes. If not, then there's no need to mention tqdm in the subscriuber notebook, because as you point out here:

```python
with meas.run() as datasaver:
    for x in tqdm_notebook(xs):
        datsaver.add_result( ... )
```

tqdm can be used just "as is". However, it is probably "nice to hint users towards using tqdm" in the "Dataset context manager notebook" in the way such as:

>If you want to see the progress of your measurement, you can use this very nice `tqdm` package. Btw, qcodes uses it internally already so it is already installed with qcodes. Hence, just read about it and tailor it to your specific needs, for example:
> ```python
> with meas.run() as datasaver:
>     for x in tqdm_notebook(xs):
>         datsaver.add_result((.., x), ... )
> ```

mgunyho:
Just for reference: I looked into tqdm a bit more, to get the ETA to display like I wanted. Apparently, with the Jupyter integration, this doesn't work properly right now (following the instructions [here](https://github.com/tqdm/tqdm/issues/494) lead me to [this issue](https://github.com/tqdm/tqdm/issues/644)). I got it working by hackishly parsing the progress bar string, as suggested in the comments [here](https://github.com/tqdm/tqdm/issues/562). This negates the "low overhead" of tqdm, but maybe that's not such a big issue when doing measurements that anyway take 10-1000 milliseconds (it could probably also be alleviated by using the Qcodes subscriber API).
```py
from datetime import datetime, timedelta
from tqdm.notebook import tqdm

xs = np.linspace(0, 1, 1001)

def parse_progressbar_eta(bar) -> int:
    eta = str(pbar).split("<")[-1].split(",")[0].split(":")[::-1]
    eta = sum(60**i * int(e) for i, e in enumerate(eta))
    return eta

pbar = tqdm(total=xs.size)

for x in xs:

    time.sleep(0.01)

    pbar.update()
    
    eta = parse_progressbar_eta(pbar)
    
    finish_time = datetime.strftime(datetime.now() + timedelta(seconds=eta),
                                    format="%a %b %d %H:%M:%S"
                                   )
    pbar.set_postfix_str(finish_time, refresh=False)
    
pbar.close()
```

PR #1801, opened 2019-11-01
Rename test file to ascii

jenshnielsen:
This fixes an issue with building the docs locally for me


jenshnielsen:
The driver only landed ~14 days ago so this may have always been there. I do not know why the issue is not triggered elsewhere. It may also be specific to my setup

PR #1804, opened 2019-11-04
Configure Codacy from file to reproduce same error messages locally

Dominik-Vogel:
Currently the only python complaints raised by codacy stem from pylint. The standard config of pylint is massively tweaked via a filter set in the codacy ui. In this PR I went through some painful regexps to convert the html of the config page to a useful `.pylintrc` file.

This file can be used instead of the UI by codacy, but it can also be used by running pylint locally.


Dominik-Vogel:
please Mikhail, remove some disable settings. I started with the original settings and already removed some, of which I thought they were purely beneficial to remove.
(TBH, the work was not so big, thanks to regexp)

PR #1806, opened 2019-11-06
Fix deprecation of static methods

Dominik-Vogel:
Turns out that my extensive tests were not extensive enough. Now, laugh and behold v3 of the vicious deprecation decorator, featuring no less then 200 LOC of test.

This time it fixes the issue that static methods could no longer be called uninhibitedly, as in the second part of this test:
```python
    def test_static_method_uninhibited(self):
        with _ignore_warnings():
            assert C.static_method(1) == 2
        with _ignore_warnings():
            c = C('pristine')
            assert c.static_method(1) == 2
``` 

Therefore I found dark magic to refrain from decorating static methods.
Lesson learned: it would have been better to just decorate the `__init__` method....

BTW. This PR does fix all tests of the driver deprecation PR.

### DIFF
--- a/qcodes/utils/deprecate.py
+++ b/qcodes/utils/deprecate.py
@@ -69,7 +69,13 @@ def deprecate(
             for m_name in dir(obj):
                 m = getattr(obj, m_name)
                 if isinstance(m, (types.FunctionType, types.MethodType)):
-                     # pylint: disable=no-value-for-parameter
+                    # skip static methods, since they are not wrapped corectly
+                    # by wrapt.
+                    # if anyone reading this knows how the following line
+                    # works please let me know.
+                    if isinstance(obj.__dict__.get(m_name, None), staticmethod):
+                        continue
+                    # pylint: disable=no-value-for-parameter
                     setattr(obj, m_name, decorate_callable(m))
                     # pylint: enable=no-value-for-parameter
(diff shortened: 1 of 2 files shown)


qcodes/utils/deprecate.py
```diff
@@ -69,7 +69,13 @@ def actual_decorator(obj: Any) -> Any:
             for m_name in dir(obj):
                 m = getattr(obj, m_name)
                 if isinstance(m, (types.FunctionType, types.MethodType)):
-                     # pylint: disable=no-value-for-parameter
+                    # skip static methods, since they are not wrapped corectly
```
astafan8: > When deprecating a classmethod, a staticmethod, or a property, the    ``@deprecated`` decorator should go *under* the ``@classmethod``, etc.    decorator (i.e., `deprecated` should directly decorate the underlying    callable).

taken from here https://matplotlib.org/3.1.1/_modules/matplotlib/cbook/deprecation.html

Dominik-Vogel: Yes, that is true, but `wrapt` should be addressing exactly that. And when decorating a class, there is no way (or at least not any obvious one) to programmatically replace the already defined functions, including their decorator with a version with a changed order of decorators.

PR #1807, opened 2019-11-06
Add types to dataset descriptions

jenshnielsen:
Makes dataset.description fully typed

### DIFF
--- a/qcodes/dataset/descriptions/dependencies.py
+++ b/qcodes/dataset/descriptions/dependencies.py
@@ -18,7 +18,7 @@ class DependencyError(Exception):
     def __init__(self,
                  param_name: str,
                  missing_params: Set[str],
-                 *args):
+                 *args: Any):
         super().__init__(*args)
         self._param_name = param_name
         self._missing_params = missing_params
@@ -32,12 +32,12 @@ class InferenceError(Exception):
     def __init__(self,
                  param_name: str,
                  missing_params: Set[str],
-                 *args):
+                 *args: Any):
         super().__init__(*args)
         self._param_name = param_name
         self._missing_params = missing_params
 
-    def __str__(self):
+    def __str__(self) -> str:
         return (f'{self._param_name} has the following inferences that are '
                 f'missing: {self._missing_params}')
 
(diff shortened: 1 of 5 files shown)


qcodes/dataset/descriptions/dependencies.py
```diff
@@ -488,7 +488,7 @@ def __repr__(self) -> str:
                f"standalones={self.standalones})")
         return rep
 
-    def __eq__(self, other):
+    def __eq__(self, other: Any) -> bool:
 
         def sorter(inp: Any) -> List[Any]:
```
astafan8: shouldn't this one actually `ParamSpecBase`, not `Any`?

qcodes/dataset/descriptions/param_spec.py
```diff
@@ -108,11 +108,11 @@ class ParamSpec(ParamSpecBase):
...
     def __init__(self, name: str,
                  paramtype: str,
-                 label: Optional[str]=None,
-                 unit: Optional[str]=None,
-                 inferred_from: Sequence[Union['ParamSpec', str]]=None,
-                 depends_on: Sequence[Union['ParamSpec', str]]=None,
-                 **metadata) -> None:
+                 label: Optional[str] = None,
+                 unit: Optional[str] = None,
+                 inferred_from: Sequence[Union['ParamSpec', str]] = None,
```
astafan8: Optional[Sequence... ?

qcodes/dataset/descriptions/dependencies.py
jenshnielsen: Nope you are allowed to compare this with anything (but it returns False naturally)

qcodes/dataset/descriptions/param_spec.py
jenshnielsen: Yes both these types are apparently wrong as they are

qcodes/dataset/descriptions/dependencies.py
astafan8: i meant the `Any` in the `sorter` method

jenshnielsen: OK :) I think this is correct

PR #1808, opened 2019-11-06
Customize  the log messages to telemetry when a measurement starts and ends

lakhotiaharshit:
As per our Friday discussion we wanted to log qchar related measurements too. The easiest way to do that is if we have some signatures in the logging info itself so that qchar related measurment can be seperated. With some quick discussion with Mikhail, we thought that since telemetry is the feature of qcodes, it is justified to provide this feature to users through qcodes itself. This PR allows making customizes logging message on top of the standard message that qcodes log. 

To do:
  - [x] Add tests once everyone agrees. 
 - [x] Prettify the test. 

@mention one core developer
@astafan8 @WilliamHPNielsen @jenshnielsen


qcodes/dataset/measurements.py
```diff
@@ -657,7 +661,8 @@ def __exit__(self, exception_type, exception_value, traceback) -> None:
             # and finally mark the dataset as closed, thus
             # finishing the measurement
             self.ds.mark_completed()
-            log.info(f'Finished measurement with guid: {self.ds.guid}')
+            log.info(f'Finished measurement with guid: {self.ds.guid}. '
```
Dominik-Vogel: Do we need to repeat the log info here? It already has the same info on the `starting measurement` part with the matching guid.

lakhotiaharshit: For the telemerty plot shown below, we need both the start and the end time of the measurement. That is why we have the repitition. 

image

Dominik-Vogel: Sorry for the misunderstanding. Reading my comment again, shows myself that I really have not done a good job in communicating ;-). I was only referring to the extra log info, that you implement here.

lakhotiaharshit: Dominik, that is correct. Once I have the GUID from start measurement this should not be needed.

lakhotiaharshit: The offline discussion happened between Dominik, Mikhail and me, and we agreed that It is ok to keep the extra_log_info in the finished measurement for the added convenience in handling data in appinsights and powerBI

qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -217,6 +218,34 @@ def get_raw(self):
...
 
 
+@pytest.fixture
+def meas_with_registered_param(DAC, DMM):
+    meas = Measurement()
+    meas.register_parameter(DAC.ch1)
+    meas.register_parameter(DMM.v1, setpoints=[DAC.ch1])
+    yield meas
+
+@pytest.mark.usefixtures("experiment")
```
astafan8: just decorate the meas_with_registered_param fixture itself and you won't need to repeat it for all the tests :)

PR #1811, opened 2019-11-07
Add types to sqlite queries

jenshnielsen:
This uncovers an actual bug in in database extract runs that was not correctly unpacking the run_ids. 
This bug happened to not matter since numpy.unique flattens nested arrays such that `[[1,2,3,3]] -> np.ndarray([1,2,3])` and was therefore not caught by tests

### DIFF
--- a/qcodes/dataset/database_extract_runs.py
+++ b/qcodes/dataset/database_extract_runs.py
@@ -20,8 +20,8 @@ from qcodes.dataset.linked_datasets.links import links_to_str
 
 def extract_runs_into_db(source_db_path: str,
                          target_db_path: str, *run_ids: int,
-                         upgrade_source_db: bool=False,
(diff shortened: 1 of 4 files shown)


qcodes/dataset/database_extract_runs.py
```diff
@@ -52,11 +54,10 @@ def extract_runs_into_db(source_db_path: str,
                  'upgrade_target_db=True to auto-upgrade the target DB file.')
             return
 
-
     source_conn = connect(source_db_path)
 
     # Validate that all runs are in the source database
-    do_runs_exist = is_run_id_in_database(source_conn, run_ids)
+    do_runs_exist = is_run_id_in_database(source_conn, *run_ids)
```
astafan8: good catch!

PR #1814, opened 2019-11-07
Fix type annotation of `InstrumentChannel.__init__` 

Dominik-Vogel:
This should  not break anything as it is more general.


jenshnielsen:
This uncovered an incorrect type in the znb driver. Which is now fixed. I also took the chance to fix a few other pep8 and type things in the znb driver

PR #1816, opened 2019-11-08
Unique name for qcodes log file for every process id 

lakhotiaharshit:
One possible solution to the logging_error happening because rollover to a new log file being blocked by another process is to uniquely name the logging file for each process id. In this PR apart from adding the process id, I have also added a date to the file to ensure a) file names are as unique as possible b) they are properly arranged and easier to find. 

Though this might clutter ~/.qcodes folder with many log files per day.  

@astafan8 @jenshnielsen @Dominik-Vogel 

To do:
- [x] Fix notebooks to explain the changes.
            - http://qcodes.github.io/Qcodes/examples/15_minutes_to_QCoDeS.html?highlight=log
- [x] Add tests. Although I have tested this manually still it is better to have tests to check if this works as intended and fixes the logging error


jenshnielsen:
Cool, On thing to consider is how we handle the roll over. AFAIK the rollover will append dates to the filename too. Do we end up with two dates in the filename (do we care) is there a better way? Should we just remove the rollover now?

qcodes/logger/logger.py
```diff
@@ -128,14 +129,23 @@ def get_level_code(level: Union[str, int]) -> int:
                            f'{type(level)} to logging level code. Need '
                            'string or int.')
 
+def generate_log_file_name():
+    """
+    Generates the name of the log file based on process id, date, time and
+    PYTHON_LOG_NAME
+    """
+    pid = str(os.getpid())
+    dt_str = datetime.now().strftime("%y%m%d-")
```
astafan8:
```suggestion
    dt_str = datetime.now().strftime("%y%m%d")
```

qcodes/logger/logger.py
```diff
@@ -128,14 +129,23 @@ def get_level_code(level: Union[str, int]) -> int:
...
+    python_log_name = dt_str + pid + PYTHON_LOG_NAME
```
astafan8:
```suggestion
    python_log_name = '-'.join([dt_str, pid, PYTHON_LOG_NAME])
```

lakhotiaharshit:
> I think it's ok to keep the roll-over in case a notebook/kernel is long-running, but then the name of the file should not need to include the date, perhaps some magic hash of the date would be ok ))

I also think roll-over may be needed for long-running notebook. I can add date dependent uuid. But it will make the names of the log file very messy. I have not seen users opening the log files very often so we can live with the messy names. However, the only way to find the relevant log files will by sorting them according to dates.

astafan8:
> However, the only way to find the relevant log files will by sorting them according to dates.

@lakhotiaharshit and by also looking at the output of `start_all_logging` function, right? at the "usually top" of the notebook

lakhotiaharshit:
> However, the only way to find the relevant log files will by sorting them according to dates.
> 
> @lakhotiaharshit and by also looking at the output of start_all_logging function, right? at the "usually top" of the notebook

No, It only gives the location of the command_history.log. It will be good to also add the filename of qcodes.log

astafan8:
> No, It only gives the location of the command_history.log. It will be good to also add the filename of qcodes.log

:) is this smth for this PR as well?

lakhotiaharshit:
> No, It only gives the location of the command_history.log. It will be good to also add the filename of qcodes.log
> 
> :) is this smth for this PR as well?

Yes, this is taken care of in this PR itself. Now the output of start_all_logging() also contains the location of the qcodes log file as shown below:

> start_all_logging()
Logging hadn't been started.
Activating auto-logging. Current session state plus future input saved.
Filename       : C:\Users\a-halakh\.qcodes\logs\command_history.log
Mode           : append
Output logging : True
Raw input log  : False
Timestamping   : True
State          : active
Qcodes Logfile : C:\Users\a-halakh\.qcodes\logs\6337b0ed417631c2579927a115985bcb-11712-qcodes.log

qcodes/tests/test_logger.py
```diff
@@ -95,7 +95,8 @@ def AMI430_3D():
 
 def test_get_log_file_name():
     fp = logger.logger.get_log_file_name().split(os.sep)
-    assert fp[-1] == logger.logger.PYTHON_LOG_NAME
+    assert str(os.getpid()) in fp[-1]
+    assert logger.logger.PYTHON_LOG_NAME in fp[-1]
```
astafan8:
```suggestion
    assert fp[-1].endswith(logger.logger.PYTHON_LOG_NAME)
```

Dominik-Vogel:
My impression is that the resulting names are not very practical to use. We will have some long uuids that don't convey any information to the human eye and some dates at the end. So we end up with a directory with a lot of files in random order.
Therefore I would like to suggest an alternative:
[start date][filename from which `start_all_logging` has been called][additional date from rollover]

PR #1817, opened 2019-11-08
Type all of qcodes.instrument.* except .mockers.ami430

astafan8:
@jenshnielsen feel free to push to this branch/PR if necessary.


qcodes/instrument/ip.py
```diff
@@ -1,47 +1,51 @@
...
     instrument subclasses.
     """
 
-    def __init__(self, name, address=None, port=None, timeout=5,
-                 terminator='\n', persistent=True, write_confirmation=True,
-                 **kwargs):
+    def __init__(self, name: str,
+                 address: Optional[str] = None,
+                 port: Optional[int] = None,
+                 timeout: Union[float, int] = 5,
```
jenshnielsen: Should we have an alias for that here too

qcodes/instrument/parameter.py
```diff
@@ -669,8 +669,7 @@ def set_wrapper(value: ParamDataType, **kwargs: Any) -> None:
         return set_wrapper
 
     def get_ramp_values(self, value: Union[Number, Sized],
-                        step: Number = None) -> List[Union[Number,
-                                                           Sized]]:
+                        step: Number = None) -> Sequence[Union[Number, Sized]]:
```
jenshnielsen: Why is this change needed. It seems like all the returned values are lists as far as I can see

jenshnielsen: BTW the type of permissive_range seems to be wrong

qcodes/instrument/sweep_values.py
```diff
@@ -127,13 +131,21 @@ class SweepFixedValues(SweepValues):
     That allows things like adaptive sampling, where you don't know ahead of
     time what the values will be or even how many there are.
     """
-    def __init__(self, parameter, keys=None, start=None, stop=None,
-                 step=None, num=None):
+    def __init__(self, parameter: '_BaseParameter',
+                 keys: Optional[Any] = None,
+                 start: Optional[Union[int, float]] = None,
+                 stop: Optional[Union[int, float]] = None,
+                 step: Optional[Union[int, float]] = None,
```
jenshnielsen: Alias?

qcodes/instrument/ip.py
Dominik-Vogel: I think this is not necessary if I understand this correctly:
https://www.python.org/dev/peps/pep-0484/#the-numeric-tower
so it should simply be float.
There are some more ocurences and this would imply we can also remove the home defined `Number`  type.

jenshnielsen:
It looks like @Dominik-Vogel is right

```
def my_test_func(a: float):
    return a + 1


if __name__ == "__main__":
    my_test_func(1)
    my_test_func(1.0)
    my_test_func(1.0 + 1.0j)
```

Type checks correctly for int and float but not for complex

jenshnielsen:
And for a bit more fun
```
import numpy as np

def my_test_func(a: float):
    return a + 1


if __name__ == "__main__":
    my_test_func(1)
    my_test_func(1.0)
    my_test_func(1.0 + 1.0j)
    my_test_func(np.int32(1))
    my_test_func(np.float64(1))
```
with https://github.com/numpy/numpy-stubs installed as of master today.

```
(qcodes) PS C:\Users\jenielse\Desktop> mypy .\testnumtypes.py
testnumtypes.py:10: error: Argument 1 to "my_test_func" has incompatible type "complex"; expected "float"
testnumtypes.py:11: error: Argument 1 to "my_test_func" has incompatible type "int32"; expected "float"
testnumtypes.py:12: error: Argument 1 to "my_test_func" has incompatible type "float64"; expected "float"
```

qcodes/instrument/parameter.py
astafan8: i don't really remember anymore, but that had to do with typing sweep_values and Sequence being a convariant(?) type as opposed to List. Let me double check this.

> BTW the type of permissive_range seems to be wrong

what should it be? (i'll also double check)

jenshnielsen: For the record. 

List does not work since its invariant in arguments so that you cannot return a List[float] from 
from a function that has List[Union[float, othertype]] 

Sequence is covariant since it's immutable so that is safe. The other alternative would be to explicitly give all the returned variables the full type. 
 

permissive_range is already fixed below

jenshnielsen:
This looks good. Only remaining point is to change all new Union[float,int] to float

PR #1822, opened 2019-11-12
Fix most of the issues revealed by using numpy stubs in type checking.

jenshnielsen:
This fixes the bulk of the issues discovered using numpy stubs with mypy. 

I will fix the last few that are more challenging and enable checking with stubs in a subsequent pr.

### DIFF
--- a/qcodes/dataset/data_export.py
+++ b/qcodes/dataset/data_export.py
@@ -83,17 +83,17 @@ def get_data_by_id(run_id: int) -> List:
         dep_data_dict_index = None
 
         for param_name, data in data_dict.items():
-            data_dict = {}
+            my_data_dict: Dict[str, Union[str, np.ndarray]] = {}
 
-            data_dict['name'] = param_name
+            my_data_dict['name'] = param_name
 
-            data_dict['data'] = data.flatten()
+            my_data_dict['data'] = data.flatten()
 
             ps = ds.paramspecs[param_name]
-            data_dict['unit'] = ps.unit
-            data_dict['label'] = ps.label
+            my_data_dict['unit'] = ps.unit
+            my_data_dict['label'] = ps.label
 
-            data_dicts_list.append(data_dict)
+            data_dicts_list.append(my_data_dict)
 
             if param_name == dep_name:
                 dep_data_dict_index = len(data_dicts_list) - 1
--- a/qcodes/dataset/data_export.py
+++ b/qcodes/dataset/data_export.py
@@ -1,4 +1,4 @@
-from typing import List, Any, Sequence, Tuple, Dict, Union
+from typing import List, Any, Sequence, Tuple, Dict, Union, cast
 import logging
 
 import numpy as np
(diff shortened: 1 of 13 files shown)


Dominik-Vogel:
@jenshnielsen could you make a short informal write up of how to use the numpy stubs? It looks like we should all know it and that it was not easy to understand.

jenshnielsen:
@Dominik-Vogel There is not much to know actually. The base classes for numpy floats and intergers are `np.floating``` and `np.interger` Other than that we are using the type for arrays `np.ndarray` without any subtyping for element types.

qcodes/dataset/data_export.py
```diff
@@ -83,17 +84,17 @@ def get_data_by_id(run_id: int) -> List:
         dep_data_dict_index = None
 
         for param_name, data in data_dict.items():
-            data_dict = {}
+            my_data_dict: Dict[str, Union[str, np.ndarray]] = {}
```
jenshnielsen: If we would allow a dependency on typing extensions this could be made a Typed dict which would avoid a lot of the casts needed when taking data out of this dict.

jenshnielsen:
@QCoDeS/core This should be ready to land. Can someone have a look thanks?

qcodes/dataset/data_export.py
```diff
@@ -83,17 +85,17 @@ def get_data_by_id(run_id: int) -> List:
...
 
-            data_dict['name'] = param_name
+            my_data_dict['name'] = param_name
```
astafan8: was the renaming really needed? :)

qcodes/dataset/data_export.py
astafan8: `allow a dependency on typing extensions` - shall we "wait for" python 3.8 for that?

qcodes/dataset/measurements.py
```diff
@@ -37,14 +38,13 @@
 log = logging.getLogger(__name__)
 
 array_like_types = (tuple, list, np.ndarray)
-scalar_res_types = Union[str, int, float, np.dtype]
+scalar_res_types = Union[str, int, float, np.integer, np.floating]
```
astafan8: what about `np.complex` and `complex`?

qcodes/dataset/plotting.py
```diff
@@ -367,7 +367,7 @@ def _complex_to_real_preparser(alldata: NamedData,
         new_group = []
         new_groups: NamedData = [[], []]
         for index, parameter in enumerate(group):
-            data: np.ndarray = parameter['data']
+            data: np.ndarray = cast(np.ndarray, parameter['data'])
```
astafan8: can't the type be determined thanks to the cast?
```suggestion
            data = cast(np.ndarray, parameter['data'])
```

qcodes/dataset/data_export.py
jenshnielsen: Yes, have a look at the code please. The code was redefining data_dict while iterating over data_dict.

astafan8: yes, didn't notice it, sorry. that was terrible))

qcodes/dataset/measurements.py
jenshnielsen: What about them :) They were not here so fixing that does not really belong in this pr but sure we can put them in

qcodes/dataset/data_export.py
jenshnielsen: Sure I fail to see why that matters but I don't really feel strongly about it

qcodes/dataset/measurements.py
astafan8: > that does not really belong in this pr

agree

> so fixing

this is what i meant - whether the reason for them not being there is smth good or its a bug.

jenshnielsen: It's a bug we do support them

PR #1824, opened 2019-11-12
Type annotate qcodes.config.* and fix docstrings

astafan8:


jenshnielsen:
I think we should finish #1682 first to fix all the typing issues around the config instance and the config module

jenshnielsen:
This has been cherry-picked into #1682

PR #1825, opened 2019-11-12
Type annotate qcodes.math.*, make FieldVector's attributes always computed

astafan8:
todo until completion:
- [ ] rework the logic regarding `_compute_unknowns` method so that a `FieldVector` instance ALWAYS has ALL of its coordinates calculated (currently it is possible to end up with a `FieldVector` that has some of its attributes equal to `None`)
- [ ] remove `T` `TypeVar` completely becuase `mypy` complained when it was used to annotate dunder methods :( or use `T` such that `mypy` understands the typing of dunders

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -21,3 +21,6 @@ disallow_untyped_defs = False
 
 [mypy-qcodes.dataset.*]
 disallow_untyped_defs = True
+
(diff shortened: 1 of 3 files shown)


astafan8:
closing because adjusting FieldVector is a bit too much work with not too much benefit

PR #1828, opened 2019-11-14
_BaseParameter: add _Cache class, use it in GetLatest and in snapshot_base (and IMPROVE IT), remove set_cache method, remove _latest

astafan8:
### Main thing

This PR takes another approach to #1791 - introducing `_Cache` class which:
- incorporates the `_latest` dictionary with `value`, `raw_value`, and `timestamp` (also `max_val_age`), and 
- `get` method that behaves as `get_latest` and 
- `set` method that behaves as `set_cache` that was introduced in #1757.

### Some details of this PR

- `GetLatest` and `get_latest` remain as is for backwards compatibility, but under the hood they reuse `_Cache` and its methods "as-is". In the near future we will be able to deprecate `GetLatest` altogether 
and move our users to use `.cache` only.
- Note that `_Cache.get` as opposed to `get_latest.get` has an additional `get_if_invalid` flag that is useful to set to `False` when you don't want to call `get` of the parameter if the cached value is invalid (for example, because the parameter has never been captured).
- All docstrings in `parameter.py` were rewritten to reflect the existence of the `_Cache`
- Drivers which used `set_cache` were changed to use `cache.set`.
- Closes  #1791 because it subsumes it.
- refactoring `test_parameter.py` into a package is out of scope for this PR

### Also, `snapshot_base` improvement (SHOULD WE DO IT NOW?)
- Moreover, `_BaseParameter.snapshot_base` has been refactored to use `cache` and rely on `cache.get`. Note that `_BaseParameter._snapshot_get` is passed to `cache.get` via this new `get_if_invalid` flag.
- `_BaseParameter.snapshot_base` got it's own set of parametrized set of tests so that its current behavior is captured (to the best of my abilities)
- Note that due to the new behavior of `_BaseParameter.snapshot_base`, some parameters had to receive `snapshot_get=False` kwarg, also some weird bug in tests for `Loop` was discovered and fixed

### ToDo:
- [ ] rewrite `test_parameter.py` to test `cache.*` instead of `get_latest.*` but also add minor tests for get_latest to ensure that it does the same as cache

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -489,25 +489,6 @@ class _BaseParameter(Metadatable):
 
         return raw_value
 
-    def set_cache(self, value: ParamDataType) -> None:
-        """
-        Set the cached value of the parameter without invoking the
-        ``set_cmd`` of the parameter (if it has one). For example, in case of
-        an instrument parameter, calling :meth:`set_cache` as opposed to
-        calling ``set`` will only change the internally-stored value of
-        the parameter (that is available when calling ``get_latest``),
-        and will pass that value to the instrument.
-
-        Note that this method also respects all the validation, parsing,
-        offsetting, etc that the ``set`` method respects. However,
-        if the parameter has :attr:`step` defined, unlike the ``set`` method,
-        this method does not perform setting the parameter step-by-step.
-
-        Args:
-            value: new value for the parameter
-        """
-        self.cache.set(value)
-
     def _save_val(self, value: ParamDataType, validate: bool = False) -> None:
         """
         Use ``cache.set`` instead of this method. It will be deprecated soon.
@@ -1740,6 +1721,22 @@ class _Cache:
         return self._max_val_age
 
     def set(self, value: ParamDataType) -> None:
+        """
+        Set the cached value of the parameter without invoking the
+        ``set_cmd`` of the parameter (if it has one). For example, in case of
+        an instrument parameter, calling :meth:`cache.set` as opposed to
+        calling ``set`` will only change the internally-stored value of
+        the parameter (that is available when calling ``cache.get`` or
+        ``get_latest``), and will pass that value to the instrument.
+
+        Note that this method also respects all the validation, parsing,
+        offsetting, etc that the parameter's ``set`` method respects. However,
+        if the parameter has :attr:`step` defined, unlike the ``set`` method,
+        this method does not perform setting the parameter step-by-step.
+
+        Args:
+            value: new value for the parameter
+        """
         self._parameter.validate(value)
         raw_value = self._parameter._from_value_to_raw_value(value)
         self._update_with(value=value, raw_value=raw_value)
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -285,12 +285,9 @@ class _BaseParameter(Metadatable):
         # record of latest value and when it was set or measured
         # what exactly this means is different for different subclasses
         # but they all use the same attributes so snapshot is consistent.
-        self._latest: Dict[str, Optional[Union[ParamDataType,
-                                               ParamRawDataType,
-                                               datetime]]]
-        self._latest = {'value': None, 'ts': None, 'raw_value': None}
+        self.cache = _Cache(self, max_val_age=max_val_age)
         self.get_latest: GetLatest
-        self.get_latest = GetLatest(self, max_val_age=max_val_age)
+        self.get_latest = GetLatest(self)
 
         self.get: Callable[..., ParamDataType]
         if hasattr(self, 'get_raw') and not getattr(self.get_raw, '__qcodes_is_abstract_method__', False):
@@ -339,27 +336,11 @@ class _BaseParameter(Metadatable):
         :setter: Setting the ``raw_value`` is not recommended as it may lead to
             inconsistent state of the parameter.
         """
-        return self._get_cache_raw()
+        return self.cache._raw_value
 
     @raw_value.setter
     def raw_value(self, new_raw_value: ParamRawDataType) -> None:
-        self._set_cache_raw(new_raw_value)
-
-    def _set_cache_raw(self, raw_value: ParamRawDataType) -> None:
-        """
-        Sets the cached raw value of the parameter.
-
-        This is a private method for internal QCoDeS use only.
-        """
-        self._latest["raw_value"] = raw_value
-
-    def _get_cache_raw(self) -> ParamRawDataType:
-        """
-        Returns the cached raw value of the parameter.
-
-        This is a private method for internal QCoDeS use only.
-        """
-        return self._latest["raw_value"]
+        self.cache._raw_value = new_raw_value
 
     @abstractmethod
     def get_raw(self) -> ParamRawDataType:
@@ -438,7 +419,11 @@ class _BaseParameter(Metadatable):
                 and self._snapshot_value and update:
             self.get()
 
-        state: Dict[str, Any] = copy(self._latest)
+        state: Dict[str, Any] = {
+            'value': self.cache._value,
+            'raw_value': self.cache._raw_value,
+            'ts': self.cache.timestamp
+        }
         state['__class__'] = full_class(self)
         state['full_name'] = str(self)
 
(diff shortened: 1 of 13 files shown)


qcodes/instrument/parameter.py
```diff
@@ -1745,6 +1695,97 @@ def setpoint_full_names(self) -> Optional[Sequence[Sequence[str]]]:
...
+        return self._max_val_age
+
+    def set(self, value: ParamDataType) -> None:
+        """
+        Set the cached value of the parameter without invoking the
+        ``set_cmd`` of the parameter (if it has one). For example, in case of
+        an instrument parameter, calling :meth:`cache.set` as opposed to
+        calling ``set`` will only change the internally-stored value of
+        the parameter (that is available when calling ``cache.get`` or
+        ``get_latest``), and will pass that value to the instrument.
```
jenshnielsen: and will not pass that value to the instrument?

qcodes/instrument/parameter.py
```diff
@@ -1763,73 +1804,36 @@ class GetLatest(DelegateAttributes):
...
+    def __init__(self, parameter: _BaseParameter):
         self.parameter = parameter
-        self.max_val_age = max_val_age
 
     delegate_attr_objects = ['parameter']
     omit_delegate_attrs = ['set']
 
     def get(self) -> ParamDataType:
-        """Return latest value if time since get was less than
+        """
```
jenshnielsen: At some point should add deprecation warning to this  I guess

qcodes/instrument/parameter.py
```diff
@@ -434,17 +415,14 @@ def snapshot_base(self, update: bool = True,
...
 
-        state: Dict[str, Any] = copy(self._latest)
-        state['__class__'] = full_class(self)
-        state['full_name'] = str(self)
-
-        if not self._snapshot_value:
-            state.pop('value')
-            state.pop('raw_value', None)
+        if self._snapshot_value:
+            get_if_invalid = self._snapshot_get and update
```
jenshnielsen: * If update is true this should explicitly call parameter.get as update means don't trust the cache IMHO
* Do we need to keep the has get check?


So should it be?
```python
has_get = hasattr(self, 'get')
if self._snapshot_value: 
    if update:
         state['value'] = self.get()
    else:
        get_if_invalid = self._snapshot_get and has_get:
        state['value'] = self.cache.get(get_if_invalid=get_if_invalid)
```

qcodes/instrument/parameter.py
```diff
@@ -650,7 +600,7 @@ def set_wrapper(value: ParamDataType, **kwargs: Any) -> None:
 
                     set_function(raw_val_step, **kwargs)
 
-                    self._update_cache_with(value=val_step,
+                    self.cache._update_with(value=val_step,
```
Dominik-Vogel: Maybe we could also include this into the `cache.set` and invoke `set_function` with `cache.raw_value`.

Dominik-Vogel:
I would like to clarify on one of the use cases we should have in mind. I think this use case is exactly the case where things got confusing (and confused) with the raw value:
When having a virtual parameter (I mean one with `get_cmd=None`) then we might want to store its value based on some external computation. as its get simply returns the cache that is straight forward. But we might want to be able to scale that parameter, so we need to set the raw value of the cache.

qcodes/instrument/parameter.py
astafan8: indeed. it's a typo.

qcodes/instrument/parameter.py
astafan8: indeed, that's the plan. but in another PR in the future :)

qcodes/instrument/parameter.py
```diff
@@ -609,7 +559,7 @@ def get_wrapper(*args: Any, **kwargs: Any) -> ParamDataType:
                 if self._validate_on_get:
                     self.validate(value)
 
-                self._update_cache_with(value=value, raw_value=raw_value)
+                self.cache._update_with(value=value, raw_value=raw_value)
```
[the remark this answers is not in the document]
astafan8: (discussed offline) adding such a kwarg is not nice because it interconnects various methods which makes code less clean. `_save_val` had such a kwarg, and hence was too powerful and also not really at the same time. So, offline it was agreed to not do this.

qcodes/instrument/parameter.py
```diff
@@ -1745,6 +1695,97 @@ def setpoint_full_names(self) -> Optional[Sequence[Sequence[str]]]:
             return self.setpoint_names
 
 
+class _Cache:
+    def __init__(self,
+                 parameter: '_BaseParameter',
+                 max_val_age: Optional[float] = None):
+        self._parameter = parameter
+        self._value: ParamDataType = None
+        self._raw_value: ParamRawDataType = None
```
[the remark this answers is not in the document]
astafan8: as we found out offline, it actually should. This will make `_BaseParameter` use only PUBLIC things of the `_Cache`. So, a read-only `raw_value` property has been added now to `_Cache`

qcodes/instrument/parameter.py
```diff
@@ -339,27 +336,11 @@ def raw_value(self) -> ParamRawDataType:
         :setter: Setting the ``raw_value`` is not recommended as it may lead to
             inconsistent state of the parameter.
         """
-        return self._get_cache_raw()
+        return self.cache._raw_value
 
     @raw_value.setter
     def raw_value(self, new_raw_value: ParamRawDataType) -> None:
```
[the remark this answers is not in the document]
astafan8: indeed, the idea is to deprecate-and-remove `_BaseParameter.raw_value` altogether in the future. since last release `raw_value` has been converted from an attribute to a property so that a more potentially harmful setting of the `raw_value` can be deprecated/removed first, and then later removed altogether thanks to introduction of `.cache` with all it's "greatness".

The docstring of the `raw_value` property already notes that it will be deprecated soon and that setting it is dangerous.

qcodes/instrument/parameter.py
astafan8: similar to the `validate` kwarg suggestion, as agreed offline, let's not do this such that the `set_wrapper` method is kept clean/better-structured/more-explicit.

astafan8:
> I would like to clarify on one of the use cases we should have in mind. I think this use case is exactly the case where things got confusing (and confused) with the raw value: When having a virtual parameter (I mean one with get_cmd=None) then we might want to store its value based on some external computation. as its get simply returns the cache that is straight forward. But we might want to be able to scale that parameter, so we need to set the raw value of the cache.

As discussed offline, this case should probably better be tackled via use of DelegateParameter or via wrapping of that external computation in a Parameter or else. The reason for this is in the core concept of a Parameter as smth that has get/set which returns or takes a "value", and "raw_value" being an extra level of complexity that is necessary to accommodate one of the main uses of the Parameter as smth that wraps an instrument-API-specific "raw_value" into a "value" that makes more sense to work with.

astafan8:
> Personally I am inclined to impose the restriction that the Parameter is not supposed to call into any private attributes or mehtods of the the Cache and vice versa.

absolutely agree. As discussed offline, we'll stick for now to only allowing Cache to use private stuff of Parameter but not vice versa.

> This could be achieved by adding the set_raw to the cache, which I feel is okay to also expose to the user.

as agreed offline, let's keep that for later when it's actually needed.

qcodes/instrument/parameter.py
astafan8: @jenshnielsen I thought about this suggestion and I do agree with it!

With one minor modification - [as per docstring of "snapshot_get" flag](https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument/parameter.py#L168-L171), it prevents calling get even `update=True`. Also, i think it's good that if parameter is not gettable even with `update=True` its get doesnt get called to avoid useless exception.

Hence i propose the following (also took the liberty to try to name variables in a meaningful way):
```python
if self._snapshot_value:
    has_get = hasattr(self, 'get')
    allowed_to_call_get_when_snapshotting = self._snapshot_get
    can_call_get_when_snapshotting = allowed_to_call_get_when_snapshotting and has_get
    if can_call_get_when_snapshotting and update:
        state['value'] = self.get()
    else:
        state['value'] = self.cache.get(get_if_invalid=can_call_get_when_snapshotting)
    state['raw_value'] = self.cache.raw_value
```
what do you think?

jenshnielsen: Yes that looks right

astafan8: great! i'm adding this to the code and covering it with tests...

astafan8:
@jenshnielsen and @Dominik-Vogel , after i was done implementing the "new" behavior of `_BaseParameter.snapshot_base` i had to fix a couple of failing tests, see the following corresponding commits:
- https://github.com/QCoDeS/Qcodes/pull/1828/commits/9773d695bfb0610d289ccb9b7b8438cd9eae7261
- https://github.com/QCoDeS/Qcodes/pull/1828/commits/a8dcf2905451fc912332036c9780a7db23f39291
- https://github.com/QCoDeS/Qcodes/pull/1828/commits/6ca47cff3c746bc6259b6d65bbfff3710b5f1801

Could you have a look at this and confirm that "it's OK", and "that's how it all should've already been", and with the change to `snapshot_base` behavior we are not "breaking" some drivers or some user code?

Dominik-Vogel:
To me the changes to the tests look very legitimate. But I did not really understand the test for the loop and I think we should not care too much about it since it will be deprecated at some point.

astafan8:
In order not to block release of qcodes 0.8.0, this PR gets split into two:
* https://github.com/QCoDeS/Qcodes/pull/1832 for introducing `_Cache` class
* https://github.com/QCoDeS/Qcodes/pull/1833 for `snapshot_base` improvement

This PR gets closed.

PR #1830, opened 2019-11-20
Unify default values of arguments for all snapshot_base and snapshot methods in qcodes repo

astafan8:
### Description

To promote "the same" behavior unless needed explicitly. (i think that it's good, but if there are use cases against that i'm happy to hear about those here)

Note that right before merging the commits in this PR will be rebased on the lastest master for a cleaner history :)

### Places in question:
- [ ] https://github.com/QCoDeS/Qcodes/blob/295ad9043ba52d61432b4e73ab1a39714572bb0c/qcodes/instrument_drivers/ZI/ZIHDAWG8.py#L79-L81
  the `ZIHDAWG8` driver - does anyone remember why is this override needed? how is this instrument special to all the other instruments?

### DIFF
--- a/qcodes/instrument/channel.py
+++ b/qcodes/instrument/channel.py
@@ -391,7 +391,7 @@ class ChannelList(Metadatable):
         self._channels = tuple(self._channels)
         self._locked = True
 
-    def snapshot_base(self, update: bool = True,
+    def snapshot_base(self, update: bool = False,
                       params_to_skip_update: Optional[Sequence[str]] = None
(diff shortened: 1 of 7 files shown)


astafan8:
@jenshnielsen @Dominik-Vogel what do you think about this PR? also about the question in the description?

jenshnielsen:
I need a bit of time to reevaluate the chaining before I can make my mind of if this

astafan8:
closing in favor of #4808

PR #1831, opened 2019-11-20
Avoid getting Keysight B1517A voltage/current parameters upon snapshot

astafan8:
Because they may not be always gettable due to the nature of those parameters and the instrument. Dicovered when working on #1828


astafan8:
> Perhaps there is a way to toggle them gettable on/off when the instrument changes state. Similar to what @lakhotiaharshit did in #1699

hmm, interesting idea. then we'd need to add a propery for `_snapshot_get`or promote it to a puclic attribute :)

jenshnielsen:
I think we should do that anyway. I have a few other usecases in mind

PR #1832, opened 2019-11-20
Add _Cache to _BaseParameter, use it in GetLatest, remove set_cache method, remove _latest

astafan8:
### Main thing

This PR takes another approach to #1791 - introducing `_Cache` class which:
- incorporates the `_latest` dictionary with `value`, `raw_value`, and `timestamp` (also `max_val_age`), and 
- `get` method that behaves as `get_latest` and 
- `set` method that behaves as `set_cache` that was introduced in #1757.

This PR subsumes significant part of https://github.com/QCoDeS/Qcodes/pull/1828 but not the improvements to `snapshot_base`.

### Some details of this PR

- `GetLatest` and `get_latest` remain as is for backwards compatibility, but under the hood they reuse `_Cache` and its methods "as-is". In the near future we will be able to deprecate `GetLatest` altogether 
and move our users to use `.cache` only.
- Note that `_Cache.get` as opposed to `get_latest.get` has an additional `get_if_invalid` flag that is useful to set to `False` when you don't want to call `get` of the parameter if the cached value is invalid (for example, because the parameter has never been captured).
- All docstrings in `parameter.py` were rewritten to reflect the existence of the `_Cache`
- Drivers which used `set_cache` were changed to use `cache.set`.
- Closes #1791 because it subsumes it.
- refactoring `test_parameter.py` into a package is out of scope for this PR

### To be done in another PR:
- rewrite `test_parameter.py` to test `cache.*` instead of `get_latest.*` but also add minor tests for `get_latest` to ensure that it does the same as `cache`

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -1891,8 +1891,8 @@ class GetLatest(DelegateAttributes):
     ``max_val_age`` for a parameter that does not have a get function.
 
     The functionality of this class is subsumed and improved in
-    :class:`_Cache` that is accessible via ``.cache`` attribute of the
-    class:`.Parameter`. Use of ``parameter.cache`` is recommended over use of
+    :class:`._Cache` that is accessible via ``.cache`` attribute of the
+    :class:`.Parameter`. Use of ``parameter.cache`` is recommended over use of
     ``parameter.get_latest``.
 
     Examples:
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -295,7 +295,7 @@ class _BaseParameter(Metadatable):
         self.get: Callable[..., ParamDataType]
         implements_get_raw = (
             hasattr(self, 'get_raw')
-            and getattr(self.get_raw, '__qcodes_is_abstract_method__', False)
+            and not getattr(self.get_raw, '__qcodes_is_abstract_method__', False)
         )
         if implements_get_raw:
             self.get = self._wrap_get(self.get_raw)
@@ -310,7 +310,7 @@ class _BaseParameter(Metadatable):
         self.set: Callable[..., None]
         implements_set_raw = (
             hasattr(self, 'set_raw')
-            and getattr(self.set_raw, '__qcodes_is_abstract_method__', False)
+            and not getattr(self.set_raw, '__qcodes_is_abstract_method__', False)
         )
         if implements_set_raw:
             self.set = self._wrap_set(self.set_raw)
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -489,25 +489,6 @@ class _BaseParameter(Metadatable):
 
         return raw_value
 
-    def set_cache(self, value: ParamDataType) -> None:
-        """
-        Set the cached value of the parameter without invoking the
-        ``set_cmd`` of the parameter (if it has one). For example, in case of
-        an instrument parameter, calling :meth:`set_cache` as opposed to
-        calling ``set`` will only change the internally-stored value of
-        the parameter (that is available when calling ``get_latest``),
-        and will pass that value to the instrument.
-
-        Note that this method also respects all the validation, parsing,
-        offsetting, etc that the ``set`` method respects. However,
-        if the parameter has :attr:`step` defined, unlike the ``set`` method,
-        this method does not perform setting the parameter step-by-step.
-
-        Args:
-            value: new value for the parameter
-        """
-        self.cache.set(value)
-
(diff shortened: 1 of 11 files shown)


qcodes/instrument/parameter.py
```diff
@@ -282,29 +282,33 @@ def __init__(self, name: str,
...
+        # It is left for now for backwards compatibility reasons and shall
+        # be deprecated and removed in the future versions.
         self.get_latest: GetLatest
-        self.get_latest = GetLatest(self, max_val_age=max_val_age)
+        self.get_latest = GetLatest(self)
 
         self.get: Callable[..., ParamDataType]
-        if hasattr(self, 'get_raw') and not getattr(self.get_raw, '__qcodes_is_abstract_method__', False):
+        get_raw_is_abstract = getattr(self.get_raw,
+                                      '__qcodes_is_abstract_method__', False)
```
jenshnielsen: Is this refactor safe? This will try to look up attributes on get_raw before checking that there is a get_raw method

astafan8: oh, you're right. good catch, i'll "revert" this.

astafan8: done.

qcodes/instrument/parameter.py
```diff
@@ -1745,15 +1730,174 @@ def setpoint_full_names(self) -> Optional[Sequence[Sequence[str]]]:
...
+                if no_get:
+                    raise RuntimeError(f"Value of parameter "
+                                       f"{(self._parameter.full_name)} "
+                                       f"is unknown and the Parameter does "
+                                       f"not have a get command. "
+                                       f"Please set the value before "
+                                       f"attempting to get it.")
+                return self._parameter.get()
+            else:
+                return self._value
```
Dominik-Vogel: Is there any chance that _value is not `None` here?

qcodes/instrument/parameter.py
```diff
@@ -1745,15 +1730,174 @@ def setpoint_full_names(self) -> Optional[Sequence[Sequence[str]]]:
...
+            return self._value
+        else:
+            if no_get:
+                # TODO: this check should really be at the time of setting
+                #  max_val_age unfortunately this happens in init before
+                #  get wrapping is performed.
+                raise RuntimeError("`max_val_age` is not supported for a "
+                                   "parameter without get command.")
+
+            oldest_ok_val = (datetime.now()
```
Dominik-Vogel: I know that you did not invent this, but I feel that the name of this variable is not good: `..._val` suggests that  it is a value. Maybe `last_accepted_ts` would be better.

qcodes/instrument/parameter.py
```diff
@@ -282,29 +282,37 @@ def __init__(self, name: str,
...
+        # ``_Cache`` stores "latest" value (and raw value) and timestamp
+        # when it was set or measured
+        self.cache = _Cache(self, max_val_age=max_val_age)
+        # ``GetLatest`` is left from previous versions where it would
+        # implement a subset of features which ``_Cache`` has.
+        # It is left for now for backwards compatibility reasons and shall
+        # be deprecated and removed in the future versions.
         self.get_latest: GetLatest
-        self.get_latest = GetLatest(self, max_val_age=max_val_age)
+        self.get_latest = GetLatest(self)
```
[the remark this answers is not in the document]
astafan8: I prefer this so that the type checking related to `.get_latest` still happens.

qcodes/instrument/parameter.py
```diff
@@ -282,29 +282,37 @@ def __init__(self, name: str,
...
 
         self.get: Callable[..., ParamDataType]
-        if hasattr(self, 'get_raw') and not getattr(self.get_raw, '__qcodes_is_abstract_method__', False):
+        has_non_abstract_get_raw = (
+            hasattr(self, 'get_raw')
+            and getattr(self.get_raw, '__qcodes_is_abstract_method__', False)
+        )
+        if has_non_abstract_get_raw:
```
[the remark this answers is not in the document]
astafan8: yes, will change

qcodes/instrument/parameter.py
astafan8: yes. especially when "None" is not a valid value for a parameter.

i'm not sure i understand the question - what did you mean to wonder?

qcodes/instrument/parameter.py
astafan8: sure, will change :)

qcodes/instrument/parameter.py
astafan8: ah, i get it. I think this will likely be rewised in the snapshot_base-improvement-related PR later. also extra tests for _Cache in the not-so-far-future would help.

PR #1833, opened 2019-11-20
Snapshot_base introduce new update=None, and use cache in parameter's snapshot_base

astafan8:
### Main thing

`_BaseParameter.snapshot_base` has been refactored to use `cache` and rely on `cache.get`.
It has gaied a third value for update such that `update=True` means unconditionally update the cache. `update=None` means update if known to be out of date and `update=False` meas do not ever try to update the underlying value. It is the intention that `update=None` should become the default in many places but that will happen in a diffrent pr.

### Other notes
- Note that `_BaseParameter._snapshot_get` is passed to `cache.get` via this new `get_if_invalid` flag.
- `_BaseParameter.snapshot_base` got it's own set of parametrized set of tests so that its current behavior is captured (to the best of my abilities)

### ToDo
- [x] **IMPORTANT** add tests for `get_if_invalid` on `_Cache.get(...)` (also for the whole `_Cache`) **EDIT** done in a different pr
- [x] remove `_value` from `DelegateCache` because it should not be needed ever, thanks to this PR **EDIT** done in a different pr
- [x] due to fixes to tests that have to be done... is the suggested behavior really correct? **EDIT** not needed any more since the change to update=False was reverted

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -433,7 +433,8 @@ class _BaseParameter(Metadatable):
 
         if self._snapshot_value:
             has_get = self.gettable
-            allowed_to_call_get_when_snapshotting = self._snapshot_get and update is not False
+            allowed_to_call_get_when_snapshotting = (self._snapshot_get
+                                                     and update is not False)
             can_call_get_when_snapshotting = (
                     allowed_to_call_get_when_snapshotting and has_get)
 
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -427,21 +427,24 @@ class _BaseParameter(Metadatable):
                 f"Parameter ({self.name}) is used in the snapshot while it "
                 f"should be excluded from the snapshot")
 
-        if self.gettable and self._snapshot_get \
-                and self._snapshot_value and update:
-            self.get()
+        state: Dict[str, Any] = {'__class__': full_class(self),
+                                 'full_name': str(self)}
 
-        state: Dict[str, Any] = {
-            'value': self.cache._value,
-            'raw_value': self.cache.raw_value,
-            'ts': self.cache.timestamp
-        }
-        state['__class__'] = full_class(self)
-        state['full_name'] = str(self)
+        if self._snapshot_value:
+            has_get = self.gettable
+            allowed_to_call_get_when_snapshotting = self._snapshot_get
+            can_call_get_when_snapshotting = (
+                    allowed_to_call_get_when_snapshotting and has_get)
 
-        if not self._snapshot_value:
-            state.pop('value')
-            state.pop('raw_value', None)
+            if can_call_get_when_snapshotting and update:
+                state['value'] = self.get()
+            else:
+                state['value'] = self.cache.get(
+                    get_if_invalid=can_call_get_when_snapshotting)
+
+            state['raw_value'] = self.cache.raw_value
+
+        state['ts'] = self.cache.timestamp
 
         if isinstance(state['ts'], datetime):
             dttime: datetime = state['ts']
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -416,7 +416,7 @@ class _BaseParameter(Metadatable):
             update: If True, update the state by calling ``parameter.get()``
                 unless ``snapshot_get`` of the parameter is ``False``.
                 If ``update`` is ``None``, use the current value from the
-                ``cache`` unless the cache is invalid. If False never call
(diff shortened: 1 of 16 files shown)


jenshnielsen:
@astafan8 I ended up changing the API of update for snapshot_base slightly. Now it takes an optional bool. Such that True and False works as before but if the value is None it will only update the cache if it is known to be invalid (the new behaviour) This is primarily so that we can be sure that https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument/base.py#L217 never triggers an update. I suggest that we then (in a subsequent pr) change the default values of update from False to None in Station

qcodes/tests/test_parameter.py
```diff
@@ -1769,12 +1789,307 @@ def test_parameter_with_overwritten_get_raises():
...
+                return get_func(*args, **kwargs)
+
+            wrapped_func.call_count = lambda: call_count
+
+            return wrapped_func
+
+        p.get = wrap_in_call_counter(p.get)
+        assert p.get.call_count() == 0  # pre-condition
+    else:
+        assert not hasattr(p, 'get')  # pre-condition
```
jenshnielsen, line 1852: p.gettable

qcodes/tests/test_parameter.py
```diff
@@ -1856,3 +2176,200 @@ def test_get_from_cache_does_not_trigger_real_get_if_get_if_invalid_false():
...
+    local_parameter = Parameter('test_param', get_cmd=False)
+
+    if get_if_invalid is True:
+        with pytest.raises(RuntimeError, match="Value of parameter test_param"):
+            local_parameter.cache.get(get_if_invalid=get_if_invalid)
+    elif get_if_invalid == NOT_PASSED:
+        with pytest.raises(RuntimeError, match="Value of parameter test_param"):
+            local_parameter.cache.get()
+    else:
+        assert local_parameter.cache.get(get_if_invalid=get_if_invalid) is None
```
jenshnielsen, line 2377: @astafan8 Maybe we should moveall the new tests to 2 new files. test_parameter_snapshot and test_parameter_cache ? this one is getting a bit extreem

astafan8: @jenshnielsen yes, sure! perhaps it's even time to turn "test_parameter.py" into "parameter" test module

jenshnielsen: Thinking a bit more about this and the conflicts this had in the tests with one of the other prs I will do this in another pr soon

jenshnielsen:
@astafan8 I think this is ready for a final look before merging. Note that the cache tests that used ot be here have already landed via one of the other prs

qcodes/instrument/parameter.py
```diff
@@ -415,8 +415,9 @@ def snapshot_base(self, update: bool = True,
         Args:
             update: If True, update the state by calling ``parameter.get()``
                 unless ``snapshot_get`` of the parameter is ``False``.
-                If ``update`` is ``False``, just use the current value from the
-                ``cache``.
+                If ``update`` is ``None``, use the current value from the
+                ``cache`` unless the cache is invalid. If False never call
```
astafan8, line 419:
```suggestion
                ``cache`` unless the cache is invalid. If ``False``, never call
```

qcodes/loops.py
```diff
@@ -285,16 +285,17 @@ def then(self, *actions, overwrite=False):
...
                       params_to_skip_update: Optional[Sequence[str]] = None):
         """
         State of the loop as a JSON-compatible dict (everything that
         the custom JSON encoder class :class:'qcodes.utils.helpers.NumpyJSONEncoder'
         supports).
 
         Args:
-            update (bool): If True, update the state by querying the underlying
-                sweep_values and actions. If False, just use the latest values
+            update : If True, update the state by querying the underlying
```
astafan8, line 296:
```suggestion
            update: If True, update the state by querying the underlying
```

qcodes/tests/test_parameter.py
```diff
@@ -2052,18 +2073,332 @@ def __init__(self, *args, **kwargs):
...
+            def wrapped_func(*args, **kwargs):
+                nonlocal call_count
+                call_count += 1
+                return get_func(*args, **kwargs)
+
+            wrapped_func.call_count = lambda: call_count
+
+            return wrapped_func
+
+        p.get = wrap_in_call_counter(p.get)
```
astafan8, line 2161: ooops, overwriting `get`...

jenshnielsen: Yes but does overwrite it with a wrapped version of it self so this is safe as is

PR #1835, opened 2019-11-21
Add Keysight 34410A driver

spauka:
Add support for the Keysight 34410A DMM into qcodes, and add support for pulling other types of parameter (current and resistance) into the driver.

- Keysight 34410A support
- Support for current, resistance measurements
- Add some missing mypy annotations

@WilliamHPNielsen

### DIFF
--- a/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
+++ b/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
@@ -149,6 +149,8 @@ class Sample(InstrumentChannel):
 
         if self.parent.is_34465A_34470A:
             _max_sample_count = 1e9
+        elif self.parent.model == "34410A":
+            _max_sample_count = 50_000
         else:
             _max_sample_count = 1e6
 
--- a/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
+++ b/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
@@ -2,8 +2,7 @@ import textwrap
 import numpy as np
 from contextlib import ExitStack
 from functools import partial
-from typing import Sequence, Optional, Dict, Union, Callable, Any, List, \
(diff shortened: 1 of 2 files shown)


qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -710,15 +745,29 @@ def abort_measurement(self) -> None:
...
-        # TODO: do we need to set any other parameters here?
+        Args:
+            sense_function: The parameter to measure. Valid values are those
+                accepted by the sense_function parameter.
 
-        with self.sense_function.set_to('DC Voltage'):
+        Returns:
+            The float value of the parameter.
+        """
+        with self.sense_function.set_to(sense_function):
```
jenshnielsen: In the long run I think we are planing a more fundamental fix for this issue but I think this is a nice improvement as is. @WilliamHPNielsen  @GateBuilder What do you think. I guess this will slightly slowdown the measurements ?

GateBuilder: I believe this adds some communication overhead, but as you pointed out, I expect that to be a very slight one. Overall, this PR is a nice improvement.

astafan8: @jenshnielsen @GateBuilder which "this issue" are you referring to? Do you mean setting sense_function each time "single-point" kind of measurement is done with these parameters?

IMHO, this is OK as a convenience, and a convenience MAY come with extra overhead. In case the user wants to avoid the overhead, he can always resort to using more verbose/explicit API (`init_measurement`+`fetch`).

spauka: Thanks for checking :) FWIW this is based on the previous code for the `volt` parameter, so even with the overhead, it won't be a regression in relation to the previous versions.

PR #1839, opened 2019-11-24
Fix return type of _BaseParameter.__call__

astafan8:
Question - is this fix correct? :)

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -406,6 +406,7 @@ class _BaseParameter(Metadatable):
         else:
             if hasattr(self, 'set'):
                 self.set(*args, **kwargs)
+                return None
             else:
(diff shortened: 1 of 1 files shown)


jenshnielsen:
@astafan8 I believe this is correct. I fixed the issues that came up as a result

qcodes/instrument/parameter.py
```diff
@@ -2187,8 +2188,8 @@ def snapshot_base(self, update: bool = False,
         meta_data['label'] = param.label  # type: ignore[attr-defined]
         meta_data['full_name'] = param.full_name  # type: ignore[attr-defined]
         meta_data['aggregator'] = repr(getattr(self, 'f', None))
-        for param in self.parameters:
```
jenshnielsen: This is the usual reusing a variable name issue for something else. Param is used above for the combinedparamter

PR #1840, opened 2019-11-25
Add initial_cache_value kwarg to Parameter

astafan8:
Similar to `initial_value` kwarg, but `initial_cache_value` will call `self.cache.set`.

ToDo:
- [x] mutually exclude use of `initial_value` and `initial_cache_value`
- [x] add tests: "`initial_cache_value` works", "mutual exclusion with `initial_value`", "`initial_cache_value` works for `DelegateParameter`"
- [x] add docstring explaining `initial_cache_value`
- [x] add docstring explaining `initial_value` (i was surprised to find out that it's not there)

### DIFF
--- a/qcodes/tests/test_delegate_parameter.py
+++ b/qcodes/tests/test_delegate_parameter.py
@@ -281,7 +281,8 @@ def test_raw_value_scaling(make_observable_parameter):
 def test_setting_initial_value_delegate_parameter():
     value = 10
     p = Parameter('testparam', set_cmd=None, get_cmd=None)
-    d = DelegateParameter('test_delegate_parameter', p, initial_value=value)
+    d = DelegateParameter('test_delegate_parameter', p,
+                          initial_value=value)
     assert p.cache._value == value
     assert d.cache._value == value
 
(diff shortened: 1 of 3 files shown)


jenshnielsen:
@astafan8 I think this is ready now. There was indeed a bug in DelegateParameter which is now fixed and has a test.

qcodes/tests/test_delegate_parameter.py
```diff
@@ -277,3 +276,21 @@ def test_raw_value_scaling(make_observable_parameter):
...
+    assert d.cache._value == value
+
+
+def test_setting_initial_cache_delegate_parameter():
+    value = 10
+    p = Parameter('testparam', set_cmd=None, get_cmd=None)
+    d = DelegateParameter('test_delegate_parameter', p,
+                          initial_cache_value=value)
+    assert p.cache._value == value
+    assert d.cache._value == value
```
jenshnielsen, line 296: Replace with a call to get(get_if_invalid=False) once https://github.com/QCoDeS/Qcodes/pull/2008 lands

PR #1843, opened 2019-11-26
Remove no longer needed apt deps from travis

jenshnielsen:
used to build h5py which now has wheels for python 3.8


jenshnielsen:
Looks like there are currently only windows wheels for 3.8 so this will have to wait a bit

astafan8 (approved):

PR #1844, opened 2019-11-26
Fix QDac for qcodes v0.8.0: inconsistent values for vrange

Dominik-Vogel:
This addresses an issues that came to light with the improvement of the cache in qcodes v0.8.0. The 'vrange' parameter was used inconsistently. (So was 'irange')

A lot of refactoring was necessary in order to find the bug, but it is a good time to do the refactoring since the new driver version needs to be tested on a real device anyway.

Ready for review, but not merging, since the hardware tests are missing.


qcodes/instrument_drivers/QDev/QDac_channels.py
```diff
@@ -335,21 +337,30 @@ def _set_vrange(self, chan, switchint):
...
+        self.write('vol {} {}'.format(chan, switchint))
 
-        old = tdict[self.channels[chan-1].vrange.get_latest()]
+        channel = self.channels[chan-1]
+        if channel.vrange.cache() != switchint:
+            cache = channel.v.cache
+            if switchint == 1:
+                cache(cache() * 0.1)
+            elif switchint == 0:
+                cache(cache() / 0.1)
```
astafan8: cache's `__call__` provides convenience only for `get()` call, not for `set(..)`
```suggestion
                cache.set(cache() / 0.1)
```

qcodes/instrument_drivers/QDev/QDac_channels.py
```diff
@@ -414,60 +420,63 @@ def _get_status(self, readcurrents=False):
...
-            chanstr, v, _, vrange, _, irange = line.split('\t')
-            chan = int(chanstr)
+            chan, i_range, v_range, v = parse_line(line)
 
-            irange_trans = {'hi cur': 1, 'lo cur': 0}
+            channel = self.channels[chan - 1]
+            channel.vrange.cache(v_range)
+            self._update_v_validator(channel, v_range)
+            channel.irange.cache(i_range)
+            channel.v.cache(v * 0.1 if v_range else v)
```
astafan8: isn't is possible to call `self._set_vrange(chan, v_range)` here instead of `self._update_v_validator(channel, v_range)` and `channel.v.cache(v * 0.1 if v_range else v)`? i'm especially worried about duplicating the `v * 0.1` part of the logic.

qcodes/instrument_drivers/QDev/QDac_channels.py
```diff
@@ -414,60 +420,63 @@ def _get_status(self, readcurrents=False):
         0-based, ie chan1 is out[0]
```
astafan8: the docstring should be changed to NOT say that this method returns anything.

qcodes/instrument_drivers/QDev/QDac_channels.py
```diff
@@ -367,34 +378,29 @@ def _current_parser(self, s):
...
 
         if 'range' in param:
             value = returnmap[param][value]
 
-        return value
 
-    def _get_status(self, readcurrents=False):
+    def _update_cache(self, readcurrents=False):
         r"""
         Function to query the instrument and get the status of all channels.
```
astafan8:
```suggestion
        Function to query the instrument and get the status of all channels,
        e.g. voltage (``v``), voltage range (``vrange``), and current range (``irange``)
        parameters of all the channels.
```

qcodes/instrument_drivers/QDev/QDac_channels.py
```diff
@@ -367,34 +378,29 @@ def _current_parser(self, s):
         """
         return 1e-6*self._num_verbose(s)
 
+    @deprecate.deprecate(reason='inconsitent values for irange',
```
astafan8:
```suggestion
    @deprecate(reason='inconsitent values for irange',
```

qcodes/instrument_drivers/QDev/QDac_channels.py
Dominik-Vogel: Thank you for catching this. This will save me some time tomorrow testing with the qdac!

qcodes/instrument_drivers/QDev/QDac_channels.py
Dominik-Vogel: You are speaking from my heart.  I really wanted to refactor this and already started doing it. But I had to undo the refactoring. Now I have been thinking about it for a while and defined the right vocabulary, writing a github comment, almost in love-letter format, to you. After that I realized a good way of refactoring and implemented it ;-).

astafan8: great :)

Dominik-Vogel:
This code has now been tested including the actual voltage output:
```python
ch = qdac.ch41

def assert_equal(v):
    assert np.isclose(ch.v.cache(), v, rtol=1e-2)
    assert np.isclose(ch.v(), v, rtol=1e-2)
    assert np.isclose(ch.v(), dmm.volt(), rtol=1e-2)

ch.vrange(0)
ch.v(1)
assert_equal(1)
ch.vrange(1)
assert_equal(0.1)
ch.vrange(0)
assert_equal(1)
ch.vrange(0)
ch.v(0.5)
assert_equal(0.5)
ch.vrange(1)
assert_equal(0.05)
```

PR #1845, opened 2019-11-26
Keysight 344xx: DIG option is always enabled with 3.0 firmware, also add MEM option, use *OPT? for retrieving enabled options

vhartong:
For two types of multimeters (`34465A` and `34470A`) from this driver the digitizer option (aka `DIG`) has been enabled with firmware version `3.0`.

This PR proposes a request that the digitizer is enabled when it is detecting the firmware version to be higher or equal to version `3.0`. `3.0` is the correct firmware version to check against for all the 4 models because the previous version is `2.17`, [as stated in this release notes document](https://www.keysight.com/upload/cmc_upload/All/Firmware_3.0_Release_Notes.pdf).

Also add support for the `'MEM'` option that changes the number of allowed pretrigger samples.

More documentation on the instruments can be found here: 
https://www.keysight.com/en/pdx-2891457-pn-34465A/digital-multimeter-6-digit-truevolt-dmm
See page 26 of user manual: 
afbeelding

Note that the firmware check will not mistakenly add DIG-enabled driver features to models `34460A` and `34461A` because the output of the `_license()` call for them will not contain `'DIG'` ever. Nevertheless, for cleanliness, the `has_DIG` flag can only be `True` if the model flag is also `True`.

Also, we have learned from keysight that `*OPT?` which returns options is a better way to check for options as that will always work (as opposed to the license command). So it is now implemented and used in DIG and MEM flags.


qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -438,7 +439,7 @@ def __init__(self, name: str, address: str, silent: bool = False,
         ####################################
         # Instrument specifications
 
-        self.has_DIG = 'DIG' in self._licenses()
+        self.has_DIG = 'DIG' in self._licenses() or LooseVersion('A.03') <= LooseVersion(idn['firmware'])
```
jenshnielsen: Should this also check that the instrument is the correct model? 

So along the lines of 

```
self.has_DIG = 'DIG' in self._licenses() or (LooseVersion('A.03') <= LooseVersion(idn['firmware']) and model in modelswithdig
```
```

Comment (vhartong) [reply qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py]: That could be an option.. At the moment, the self.has_DIG only seems to affect the dicts for certain models so with this implementation it would seem redundant to also specify the model, small example from the code:
```
        PLCs = {'34410A': [0.006, 0.02, 0.06, 0.2, 1, 2, 10, 100],
                '34460A': [0.02, 0.2, 1, 10, 100],
                '34461A': [0.02, 0.2, 1, 10, 100],
                '34465A': [0.02, 0.06, 0.2, 1, 10, 100],
                '34470A': [0.02, 0.06, 0.2, 1, 10, 100]
                }
        if self.has_DIG:
            PLCs['34465A'] = [0.001, 0.002, 0.006] + PLCs['34465A']
            PLCs['34470A'] = [0.001, 0.002, 0.006] + PLCs['34470A']
```

Comment (jenshnielsen) [reply qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py]: Nope it also affects certain options such as the pre trigger. Most places it checks has_DIG and model number but not every where.

Comment (jenshnielsen) [reply qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py]: * https://github.com/QCoDeS/Qcodes/blob/7ca399a05c3c559160b3a4f2a532319f7551d07f/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py#L87
* https://github.com/QCoDeS/Qcodes/blob/7ca399a05c3c559160b3a4f2a532319f7551d07f/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py#L123
* https://github.com/QCoDeS/Qcodes/blob/7ca399a05c3c559160b3a4f2a532319f7551d07f/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py#L175
* https://github.com/QCoDeS/Qcodes/blob/7ca399a05c3c559160b3a4f2a532319f7551d07f/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py#L363

Comment (jenshnielsen) [reply qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py]: I think that line 175 and 363 will only work for the 34465 and 34470 too. These may now evaluate to true for other models.

Comment (vhartong) [reply qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py]: Yes, this would seem to be the case according to this table: 
afbeelding

Comment (jenshnielsen) [reply qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py]: So I would suggest that we change that line to

```python
# 34465A and 34470A supports additional options with the DIG extension
# this is an optional licenced extension with older versions of the firmware
# but is included in the baseline config from version A.03 onward.  
self.has_DIG = ('DIG' in self._licenses() or LooseVersion('A.03') <= LooseVersion(idn['firmware'])) and self.is_34465A_34470A
```

Then we can get rid of the model checks on line 87 and 123.

vhartong: Looking at the table above, the DIG can be either activated via license or the new firmware so that's ok I think. The other models don't have either the option for DIG or MEM, so the model checks are also good. 
Now the DIG activates special triggering and more values for this. The MEM only activates the pretrigger to go up to 2e6 for pretriggering. I will implement this and let's have a second look at this from that point.

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -438,7 +444,8 @@ def __init__(self, name: str, address: str, silent: bool = False,
...
+        self.has_MEM = 'MEM' in self._licenses()
```
vhartong: Adding the has_MEM since this will allow 2e6 counts in pretrigger

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
vhartong: Now DIG is either enabled by the license or the newer firmware. The MEM will only be enabled if the license is added to the device. This will only enable more counts in the pretrigger. The other changes come with either DIG license or new firmware.

astafan8: Nice!

jenshnielsen:
We have learned from keysight that ```*OPT?``` which returns options is a better way to check for options as that will always work (as opposed to the license command) I suggest that we add a private function similar to _licenses() that does this? and use that in the has_dig and has_mem checks. That will be much more future proof

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -756,6 +767,17 @@ def _licenses(self) -> Sequence[str]:
             return licenses_list
         return tuple()
 
+    def _options(self) -> Sequence[str]:
```
jenshnielsen: Please don't use Abstract types in return position. We know this is a tuple so the correct type is `Tuple[str, ...]`

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -438,7 +444,12 @@ def __init__(self, name: str, address: str, silent: bool = False,
         ####################################
         # Instrument specifications
 
-        self.has_DIG = 'DIG' in self._licenses()
+        options = self._options()
+        self.has_DIG = self.is_34465A_34470A and (
+            'DIG' in options
+            or LooseVersion('A.03') <= LooseVersion(idn['firmware'])
```
jenshnielsen: Do we still need this version check?

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -172,12 +173,17 @@ def __init__(self, parent: '_Keysight_344xxA', name: str, **kwargs):
             option) or 2,000,000 readings (with the MEM option)"""))
 
         if self.parent.has_DIG:
+            if self.parent.has_MEM:
+                _max_pretrig_count = 2e6
```
jenshnielsen: this should really be int(2e6) as 2e6 is a float

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -172,12 +173,17 @@ def __init__(self, parent: '_Keysight_344xxA', name: str, **kwargs):
...
+            else:
+                _max_pretrig_count = 5e4
+
```
jenshnielsen: The way that I read the above comment in the count parameter is that you have a total of 50_000 or 2_000_000 points if you use pretrigger not that amount of points in the pretrigger but I have not tested this my self

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
astafan8: yes, because  as per our testing `options` does NOT return DIG for 3.0 firmware :)

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
astafan8: done. good catch!

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
astafan8: done

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
vhartong: This is correct. From the manual:
afbeelding

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
vhartong: Correct. And the _max_pretrig_count below should be int(5e4)

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
jenshnielsen: That is disappointing. Please document in a comment

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
jenshnielsen: The way I read that thing from the manual is that the maximum for count is 50_000 or 2_000_000 if pretrigger is enabled at all. This does not say anything about the maximum number of pretrigger samples (which I agree must be the same or lower)

astafan8: ok, we read the manual again, and here's the outcome:
- the maximum allowed number of pretrigger samples ('pretrigger_count') is always equal to the current sample count ('sample.count') minus 1.
- the maximum allowed number of sample counts ('sample.count') depends on a model, on whether pretrigger samples are used or not, and the MEM option availability.
  - for the models in question (those which can have DIG), if pretriggerring is used, the maximum allowed sample count in 50k without MEM option, and 2000k with MEM option.

This last sub-bullet point is not implemented for the 'sample.count' parameter. I think its out of scope of this PR. As for the maximums for the pretrigger count, i think we can keep them as is for now, and in the bright future properly implement the fact that with the change of 'sample.count' the validator for the pretrigger_count changes as well.

@jenshnielsen would you agree?

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
astafan8: yes, done.

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
jenshnielsen: Yes that sounds good. 
Can we:

* move the -1 into the definition of the _max_pretrig_count variable so the name is true. 
* Add the fact that max_pretrigger_count is max_count -1 to the docs of pretrigger_count with a note to read the docs of count?

astafan8: yes! i'll do both right now!

PR #1847, opened 2019-11-28
Run pip with --no-index in installation_info

mgunyho:
We're running Qcodes on a lab computer with no internet access. Running `pip list` looks online for the package index. This takes 20 seconds to time out, which makes `start_all_logging()` annoyingly slow. This fixes that.


jenshnielsen:
but please fix the style issue pointed out by codacy

mgunyho:
Done. I made it True instead of the default False to catch errors more clearly.

jenshnielsen:
Thanks, I think check=True is good no need to revert anything

PR #1848, opened 2019-11-28
Release notes 0.8.1

Dominik-Vogel:


jenshnielsen:
I suggest that we are nice and also back port https://github.com/QCoDeS/Qcodes/pull/1847

mgunyho:
To be precise, `start_all_logging` didn't previously raise an exception, but rather it just stalled for 20 seconds before pip timed out and continued normally.

Dominik-Vogel:
Thank you @mgunyho ! I didn't worry about changing it, as it such a minor detail.

PR #1850, opened 2019-11-28
Call `start_all_logging` on qcodes import

Dominik-Vogel:
Adding the call to `start_all_logging` to the `qcodes.__init__` file.
This PR provide a new option in  `qcodesrc.json` to enable this feature (defaults to false). It will also automatically enabled if a workstation id and location id has been defined.

Furthermore, it is tested if we are in a pytest session. It is bad practice to change anything dependent on whether it is a test situation (this lesson we learned from an anonymous automobile company). Here one could argue that we have not been testing with logging enabled in any case, so that the effective behavior is not altered.

I checked if the logging was triggered by pylint and mypy. Neither do trigger logging.

I tried to incorporate Jens suggestion about marking the test environment in conftest.py, but this failed because qcodes gets imported before entering conftest.
I added a condition for the instrumentation key and refactored the pr.


Dominik-Vogel:
This PR is ready for review. I have, however, not figured out how the cyclic import works:
`qcodes.__init__`  imports `qcodes.logger.start_all_logging` and `qcodes.logger` in turn imports `qcodes`. Same for `add_to_spyder_uml`.

qcodes/__init__.py
```diff
@@ -5,13 +5,38 @@
 # config
 
 from qcodes.config import Config
+from qcodes.logger import start_all_logging
 from qcodes.utils.helpers import add_to_spyder_UMR_excludelist
 from .version import __version__
 
+config: Config = Config()
+
+# start logging if work_station is configured or
```
astafan8:
```suggestion
```

Dominik-Vogel: Thanks!

Dominik-Vogel: I will add the documentation.

qcodes/logger/logger.py
```diff
@@ -293,6 +293,36 @@ def start_all_logging() -> None:
...
+    ``config.logger.start_logging_on_import`` is ``True``.
+    """
+    def start_logging_on_import() -> bool:
+        config = qc.config
+        return (
+            config.GUID_components.location != 0
+            and config.GUID_components.work_station != 0
+            and config.telemetry.instrumentation_key != \
+                "00000000-0000-0000-0000-000000000000"
+            or config.logger.start_logging_on_import
```
jenshnielsen: I would still consider if it makes sense to make `start_logging_on_import` an enum with 3 states. `Always`, `Never` and `if_setup_for_telemetry`

Dominik-Vogel: Yes that is right, sorry Jens, I forgot about this point. I am on it.

jenshnielsen:
@Dominik-Vogel There are some (unintentional i guess) changes to to the docs templates in this pr

Dominik-Vogel:
Thank you for spotting this. I'll fix it.

qcodes/logger/logger.py
```diff
@@ -293,6 +293,54 @@ def start_all_logging() -> None:
...
+                    "00000000-0000-0000-0000-000000000000"
+            )
+        else:
+            raise RuntimeError('Error in qcodesrc validation.')
+
+    def running_in_test_or_tool() -> bool:
+        import sys
+        tools = (
+            'pytest.py', 'pytest',
+            '_jb_pytest_runner.py', 'testlauncher.py'
```
jenshnielsen: It might be nice to document where these are coming from ? I guess the _jb_pytest_runner is pycharm but what is testlauncher?

PR #1852, opened 2019-11-28
Add identification component to Station

astafan8:
### Description

This is an idea to add identification that is used in GUID components to the Station as well, so that these ID's are also openly available from snapshot.

Discussion is open :)

### ToDo:
- [ ] tests

### Questions:
- [ ] naming of the class, fields, etc
- [ ] should identification be a proper "component" in the station, analogous to the yaml config? Or can it be a completely new root key in the snapshot dictionary? or else? there reason it is now made a component is that to re-use and conform to the concept that "a station just contains metadatable (meaning snapshot-able) components".


astafan8:
@QCoDeS/core would you mind providing your perspective on the questions mentioned in the description of the PR? :)

astafan8:
this hasn't been proved needed or useful, hence closing

PR #1861, opened 2019-12-04
fix the lakeshore325 sensor status bug, and add two missing commas in the driver

liangosc:
For the while loop, `comp` will become [0] in the end and `len(comp)` will always be 1, so the while loop will not finish.

There was also a comma (,) missing in the setter of `resistance` parameter of `Model_325.Model_325_Heater class`, and another missing in the setter of `curve_index` parameter of the same class.

- I changed the condition for the while loop to be `len(comp)>1`
- I also added a unit test to test the `_get_sum_terms` method, which is used in `decode_sensor_status` method
-  I added the two missing commas in this PR.

@sohailc


qcodes/tests/drivers/test_lakeshore_325.py
```diff
@@ -0,0 +1,27 @@
+import numpy as np
+from hypothesis import given
+import hypothesis.strategies as st
+from qcodes.instrument_drivers.Lakeshore import Model_325
+
+COMPONENTS = [0, 1, 16, 32, 64, 128]
```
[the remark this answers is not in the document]
liangosc: This is much better. Thanks!

The `list` doesn't add anything, can I just use the following?
`COMPONENTS = sorted(Model_325.Model_325_Sensor.sensor_status_codes.keys())`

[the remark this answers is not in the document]
astafan8: sure, remove the `list` :)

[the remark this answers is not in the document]
liangosc: I'll work on this. Don't merge yet.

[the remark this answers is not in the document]
liangosc: Done. The `IntFlag` is quite convenient. Thanks for the suggestion!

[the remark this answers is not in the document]
liangosc: I removed the dictionary `sensor_status_codes` from the sensor class because it is not used in the decoder anymore, as a result, I added the dictionary in the test function. (no way to call the IntFlag object in the decoder)

qcodes/instrument_drivers/Lakeshore/Model_325.py
```diff
@@ -278,41 +269,23 @@ def __init__(self, parent: 'Model_325', name: str, inp: str) -> None:
...
+            sensor_units_zero = 64
+            temp_overrange = 32
+            temp_underrange = 16
+            invalid_reading = 1
+
+        if sum_of_codes == 0:
+            return 'OK'
+        sensor_status = []
+        for st in Status:
+            if st <= sum_of_codes:
```
astafan8: this `if` stuff should not be needed because the `sum_of_codes` can be converted to a `Status` instance, and then checked against containing separate flags in it, like this (the check for `0` is still needed unfortunately):
```python
total_status = Status(sum_of_codes)
if total_status == 0:
    return 'OK'
status_messages = [st.name.replace('_', ' ') for st in Status if st in total_status]
return ', '.join(status_messages)
```

Moreover, I read that although `IntFlag` exists, preference should be given to use of `Flag` and the reason for this is that `Flag` is less forgiving to `sum_of_codes` - for example, `Flag` will raise an exception on `total_status = Status(sum_of_codes)` if `sum_of_codes` contains codes that are not in the enumeration, for example, if `sum_of_codes == 16+2`.

qcodes/instrument_drivers/Lakeshore/Model_325.py
```diff
@@ -278,41 +269,23 @@ def __init__(self, parent: 'Model_325', name: str, inp: str) -> None:
...
-            while len(comp):
-                c = comp[0]
-                number -= c
-                terms.append(c)
-
-                comp = comp[comp <= number]
-
-        return terms
+    def decode_sensor_status(sum_of_codes: int) -> str:
+        class Status(IntFlag):
```
astafan8: i wonder, why is it a private enum? i think it's OK to have it "public", at least outside this method so that it can be used in tests directly and can be referred to in code if needed elsewhere :)

qcodes/instrument_drivers/Lakeshore/Model_325.py
```diff
@@ -471,7 +444,7 @@ def __init__(self, parent: 'Model_325', name: str, loop: int) -> None:
         self.add_parameter(
             "resistance",
             get_cmd=f"HTRRES? {self._loop}",
-            set_cmd=f"HTRRES {self._loop} {{}}",
+            set_cmd=f"HTRRES {self._loop}, {{}}",
```
astafan8: oh) this should've been a separate PR but it's ok, just please mention this change in the description (and perhaps in the title)

qcodes/instrument_drivers/Lakeshore/Model_325.py
sohailc: This is really interesting, I never knew about the Flags. You can do stuff like : 

```python
from enum import IntFlag

class Status(IntFlag):
    sensor_units_overrang = 128
    sensor_units_zero = 64
    temp_overrange = 32
    temp_underrange = 16
    invalid_reading = 1

total_status = Status(128)
Status.sensor_units_overrang in Status(128)  # Returns True 
Status.sensor_units_zero in Status(128)  # Returns False 
```

liangosc: Wow, very nice use of `IntFlag`, I changed the `decode_sensor_status` method accordingly.

qcodes/instrument_drivers/Lakeshore/Model_325.py
liangosc: I have updated the description and title.

qcodes/instrument_drivers/Lakeshore/Model_325.py
liangosc: I made the `Status` "public" so that it can be called elsewhere (I called it in the test). However, I have to keep the `decode_sensor_status` as "staticmethod" because I want to call it directly in the test without defining an equipment.

So `def decode_sensor_status(sum_of_codes: int, status=Status) -> str:`
instead of  `def decode_sensor_status(self, sum_of_codes: int) -> str:`

qcodes/tests/drivers/test_lakeshore_325.py
```diff
@@ -0,0 +1,29 @@
...
+def test_decode_sensor_status(list_of_codes):
+    """
+    The sensor status is one of the status codes, or a sum thereof. Multiple
+    status are possible as they are not necessarily mutually exclusive.
+    The static method 'decode_sensor_status' in the Model_325_Sensor class can
+    decode the status into status code(s).
+    """
+    codes = [code.name.replace('_', ' ') for code in list_of_codes[::-1]]
+    codes_message = ', '.join(codes)
+    sum_of_codes = STATUS(sum(list_of_codes))
```
astafan8: this is incorrect. `decode_sensor_status` takes an `int` as its `sum_of_codes` argument, not `Status` type, hence in the test `decode_sensor_status` should be called with an `int`, not a `Status` instance.
```suggestion
    sum_of_codes = int(sum(list_of_codes))
```

qcodes/instrument_drivers/Lakeshore/Model_325.py
```diff
@@ -278,41 +276,14 @@ def __init__(self, parent: 'Model_325', name: str, inp: str) -> None:
...
-                comp = comp[comp <= number]
-
-        return terms
+    def decode_sensor_status(sum_of_codes: int, status=Status) -> str:
+        total_status = status(sum_of_codes)
+        if sum_of_codes == 0:
+            return 'OK'
+        status_messages = [st.name.replace('_', ' ') for st in status
+                           if st in total_status]
+        return ", ".join(status_messages)
```
astafan8: Why is this `, status=Status` argument needed? the static method can access the "class attribute `Status`" without any problems. Also, any external code can access the "class attrbiute `Status`" via the class (not instance): `Model_325_Sensor.Status` - that's it :)

i leave it up to you, but i think it's also fine to declare the `Status` of the module level, not necessarily inside the class.

qcodes/instrument_drivers/Lakeshore/Model_325.py
```diff
@@ -1,4 +1,4 @@
-import numpy as np
+from enum import IntFlag
```
astafan8: @liangosc so what do you think about use of `IntFlag` vs `Flag` classes? I read in the python docs (or somewhere else?) that although `IntFlag` exists, preference should be given to use of `Flag` and the reason for this is that `Flag` is less forgiving to integer values it can represent - for example, if `Status` was a subclass of `Flag`, it'd will raise an exception on `total_status = Status(sum_of_codes)` if `sum_of_codes` contains codes that are not in the enumeration, for example, if `sum_of_codes == 16+2`. imho, it's a valuable feature in this case. What's your opinion?

qcodes/tests/drivers/test_lakeshore_325.py
liangosc: Good catch!

qcodes/instrument_drivers/Lakeshore/Model_325.py
liangosc: I have made it module level.

qcodes/instrument_drivers/Lakeshore/Model_325.py
liangosc: I agree it''ll be nice to raise an exception when the sum contains codes that are not in the enumeration -- one situation is, the manufacturer may add more error codes in the future and `Flag` will be able to capture it.

However, I like the fact that `IntFlag` supports `==, <, >`. It doesn't seem we have used this feature in the code, but we may have: in the test function, we won't be able to do `st.sampled_from(list(STATUS))` if STATUS is `Flag` type. (It may still work after some modification though.)

qcodes/instrument_drivers/Lakeshore/Model_325.py
```diff
@@ -278,41 +277,14 @@ def __init__(self, parent: 'Model_325', name: str, inp: str) -> None:
...
-            while len(comp):
-                c = comp[0]
-                number -= c
-                terms.append(c)
-
-                comp = comp[comp <= number]
-
-        return terms
+    def decode_sensor_status(sum_of_codes: int) -> str:
+        total_status = Status(int(sum_of_codes))
```
astafan8: what is the conversion to `int` needed for here if the input type is already `int`?
```suggestion
        total_status = Status(sum_of_codes)
```

liangosc: The input from the machine was actually `str`, not `int`. I changed the `get_parser` of the parameter `status` to a lambda function `get_parser=lambda status: self.decode_sensor_status(int(status))`, so that the parser can convert the input to int and call the decoder.

PR #1876, opened 2019-12-12
fix: Keithley_2400 READ? format may be inconsistent

lychee512:
Fixes 'bugginess' mentioned in #488 

- Added code to ensure Keithley 2400 instrument sends responses to READ? in correct format in the driver __init__ method.
- Modify 'output' parameter to use create_on_off_val_mapping instead of int.

Breaking changes: 'output' parameter type change could possibly break existing code that depends on 'output' being an int.

@WilliamHPNielsen

### DIFF
--- a/qcodes/instrument_drivers/tektronix/Keithley_2400.py
+++ b/qcodes/instrument_drivers/tektronix/Keithley_2400.py
@@ -1,5 +1,6 @@
 from qcodes import VisaInstrument
 from qcodes.utils.validators import Strings, Enum
+from qcodes.utils.helpers import create_on_off_val_mapping
 
 
 class Keithley_2400(VisaInstrument):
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/tektronix/Keithley_2400.py
```diff
@@ -97,6 +98,7 @@ def __init__(self, name, address, **kwargs):
                            docstring="Measure resistance from current and voltage "
                                      "Note that it is an error to read current "
                                      "and voltage with output off")
+        self.write(':TRIG:COUN 1;:FORM:ELEM VOLT,CURR')
```
jenshnielsen: Is this the same mode that reset would put the instrument in? Could you document (for instance in a comment) what this does

lychee512: Comment added in the driver itself, and pushed in a new commit. Did you mean to comment just here instead? I didn't realise there was a github comment feature!

jenshnielsen: Thanks

astafan8: both these commands aught to be represented as qcodes parameters (or at least methods), in which case this explanation wouldn't be needed. but we can leave this to another PR.

PR #1877, opened 2019-12-13
Feature Rohde-Schwarz.ZNB: adjust timeout automatically to measurement time during data acquisition

LGruenhaupt:
@astafan8 as we briefly discussed last week:

VNA measurements regularly take longer than the default timeout of 5s.
To avoid having to adjust the timeout every time globally, query the sweep time from the instrument and use ` with vna.timeout.set_to(sweep time +1s)` around the data acquisition. Additional 1 s is to allow for some communication overhead.
This is done in method _get_sweep_data of class ZNBChannel

- adjust timeout during data acquisition in method _get_sweep_data of class ZNBChannel
- some minor formatting changes


LGruenhaupt:
@astafan8 
As discussed I implemented the following changes:
- made sweep_time a parameter, 
- move the added time to allow for some communication overhead into a private argument of the channel class, 
- use the with statement to set the status of the channel to 1 during data acquisition,
- fix the style, which hopefully is now conforming with the Codacy quality review :)

PR #1884, opened 2020-01-03
QCoDeS: Keithley System Switch Instrument Driver

GateBuilder:
This PR introduces the instrument driver for Keithley System Switch. While the main driver is ready and in working condition (parties who are in urgent need for a driver can in fact make use of the current version), followings should be done before general review process and hopefully merging:
 
- [x] Add general validator functions and exception handling
- [x] Add tests
- [x] Add example notebook

@jenshnielsen @Dominik-Vogel @astafan8


qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,525 @@
...
+        """
+        return self.ask(f"channel.getforbidden('{val}')")
+
+    def clear_forbidden_channels(self, val: str) -> None:
+        """
+        Clears the list of channels that are forbidden to close.
+
+        Args:
+            val: A string representing the channels that will no longer
+            be listed as forbidden to close.
```
jenshnielsen: Missing indent here

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,600 @@
...
+            val: A string representing the channels,
+                backplane relays or channel patterns to be queried.
+        """
+        if val == "":
+            raise InvalidValue('Argument cannot be an empty string.')
+        if not self._validator(val):
+            raise InvalidValue(f'{val} is not a valid specifier. '
+                               'The specifier should be channels, channel '
+                               'ranges, slots, backplane relays or "allslots".')
+        return self.ask(f"channel.getclose('{val}')")
```
jdwatson: It might be worth parsing the 'nil' returned if no channels are closed to None - I can imagine this would be a nicer check in a routing situation so you could write something like:
if smatrix.get_closed_channels("slot1") is not None:
     foo

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,600 @@
...
+        if val in slots:
+            raise InvalidValue("Slots cannot be exclusively closed.")
+        if val == "":
+            raise InvalidValue('Argument cannot be an empty string.')
+        if not self._validator(val):
+            raise InvalidValue(f'{val} is not a valid specifier. '
+                               'The specifier should be channels or channel '
+                               'ranges and associated backplane relays.')
+        self.write(f"channel.exclusiveslotclose('{val}')")
+
```
jdwatson: Another potentially useful convenience function would be something like connect all rows in a slot to a given column or connect all columns in a slot to a given row.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,600 @@
...
+        self.connect_message()
+
+    def _reset_channel(self, val: str) -> None:
+        if not self._validator(val):
+            raise InvalidValue(f'{val} is not a valid specifier. '
+                               'The specifier should be channels, channel '
+                               'ranges, slots, backplane relays or "allslots".')
+        self.write(f"channel.reset('{val}')")
+
+    def _open_channel(self, val: str) -> None:
```
jdwatson: A general comment on the open/close/reset methods - strings and lists of strings (which are actually still just strings like "1101,1102") seem like they might be kind of cumbersome to use.  Could one instead use ints and allow for the user to also pass tuples of ints?  This feels like an easier way to programmatically specify channels to open/close.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,600 @@
...
+                           docstring='Resets the specified channels to '
+                                     'factory default settings.')
+
+        self.add_parameter('open_channel',
+                           get_cmd=None,
+                           set_cmd=self._open_channel,
+                           docstring='Opens the specified channels and '
+                                     'backplane relays.')
+
+        self.add_parameter('close_channel',
```
jdwatson: I tested the set command with this function on the backplane relays, and it does not raise an exception if the interlock for that backplane relay is disabled.  It would probably be a good idea to raise this exception or at least issue a warning so that the user knows that the relay is not in the desired state.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
GateBuilder: The problem here, I cannot (I mean should not) change the input types of a specific function(s). Either I should do it for the whole driver or not for the sake of clarity and coherency. Additionally, I do not see any advantage of passing 1111 instead of "1111", which differs by 2 chars which shall be auto-completed in any modern text editor, thus only one additional key stroke. Also, if user would like to do slot wise operations he/she has to pass strings anyway. 

Aside from above,  I am not in favor of this as : we have to cast every integer to string, a list of integers to strings etc as instrument only accepts strings. I do not know how much overhead this may bring.

However, if you certainly feel like this should be the way, I will do implement the changes driver-wise.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+                                     'before any channels close. If it is set '
+                                     'to make before break, it is ensured that '
+                                     'all channels close before any channels '
+                                     'open. If it is off, channels open and '
+                                     'close simultaneously.',
+                           vals=vals.Enum('BREAK_BEFORE_MAKE',
+                                          'MAKE_BEFORE_BREAK',
+                                          'OFF'))
+
+        self.add_parameter('gpib_enable',
```
astafan8: ;)
```suggestion
        self.add_parameter('gpib_enabled',
```

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+                                                                 ))
+
+        self.add_parameter('gpib_address',
+                           get_cmd=self._get_gpib_address,
+                           get_parser=int,
+                           set_cmd=self._set_gpib_address,
+                           docstring='Sets and gets the GPIB address.',
+                           vals=vals.Ints(1, 30))
+
+        self.add_parameter('lan_enable',
```
astafan8: ;)
```suggestion
        self.add_parameter('lan_enabled',
```

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+        kind of matrix, as well.
+        """
+        idn = self.get_idn()
+        cards = self.get_switch_cards()
+        states = self.get_interlock_state()
+
+        con_msg = ('Connected to: {vendor} {model} SYSTEM SWITCH '
+                   '(serial:{serial}, firmware:{firmware})'.format(**idn))
+        print(con_msg)
+        self.log.info(f"Connected to instrument: {idn}")
```
astafan8: these three lines can be substituted with `super().connect_message()`, right?

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+
+    def save_setup(self, val: Optional[str] = None) -> None:
+        """
+        Saves the present setup.
+
+        Args:
+            val: An optional string representing the path and the file name
+                to which the setup shall be saved on a USB flash drive. If not
+                provided, the setup will be saved to the nonvolatile memory
+                of the instrument. Any previous saves shall be overwritten.
```
astafan8: "shall" sounds more like "wish/expectation" whereas the docstring should state the "fact", imho
```suggestion
                of the instrument, any previous saves will be overwritten.
```

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+        cannot be energized.
+        """
+        slot_id = self._get_slot_id()
+        interlock_status = {0: 'Interlock is disengaged',
+                            1: 'Interlock is engaged'}
+        states: List[Dict[str, str]] = []
+        for i in slot_id:
+            state = self.ask(f'slot[{int(i)}].interlock.state')
+            state_dict = {'slot_no': i,
+                          'state': interlock_status[int(float(state))]}
```
astafan8: is the `int(state)` not enough because the instrument returns `1.0` instead of `1`? :)

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+        for element in specifiers:
+            if element in backplanes:
+                raise InvalidValue(f'{val} cannot be a analog backplane relay.')
+        if not self._validator(val):
+            raise InvalidValue(f'{val} is not a valid specifier. '
+                               'The specifier should be channels, channel '
+                               'ranges, slots, or "allslots".')
+        return self.ask(f"channel.getbackplane('{val}')")
+
+    def _get_slot_id(self) -> List[str]:
```
astafan8: shall it be plural because it returns a list of ids?
```suggestion
    def _get_slot_ids(self) -> List[str]:
```

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+        """
+        Returns the slot ids of the installed cards.
+        """
+        cards = self.get_switch_cards()
+        slot_id = []
+        for _, item in enumerate(cards):
+            slot_id.append('{slot_no}'.format(**item))
```
astafan8: i find these lines a bit too complicated. Just for my understanding, is this what is happening here `slot_id = [f'{card['slot_no']}' for card in cards]`?

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+        return slot_id
+
+    def _get_slot_name(self) -> List[str]:
```
astafan8:
```suggestion
    def _get_slot_names(self) -> List[str]:
```

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+        slot_names = [f'slot{x}' for x in slot_id]
+        return slot_names
+
+    def _get_number_of_rows(self) -> List[int]:
+        """
+        Returns the total number of rows of the installed cards.
+        """
+        slot_id = self._get_slot_id()
+        total_number_of_rows = [int(float(self.ask(
+            f'slot[{int(i)}].rows.matrix'))) for i in slot_id]
```
astafan8: for readability?
```python
total_number_of_rows = [
    int(float(self.ask(f'slot[{int(i)}].rows.matrix')))
    for i in slot_id
]
```

astafan8: `int(i)` is unnecessary, right?

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+        return total_number_of_rows
+
+    def _get_number_of_columns(self) -> List[int]:
+        """
+        Returns the total number of columns of the installed cards.
+        """
+        slot_id = self._get_slot_id()
+        total_number_of_columns = [int(float(self.ask(
+            f'slot[{int(i)}].columns.matrix'))) for i in slot_id]
+        return total_number_of_columns
```
astafan8: same comments as for `_get_number_of_rows` method

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+
+    def _get_rows(self) -> List[List[str]]:
+        """
+        Returns the elements of each row.
+        """
+        total_number_of_rows = self._get_number_of_rows()
+        row_list = []
+        for _, item in enumerate(total_number_of_rows):
```
astafan8: there's no need for `enumerate` here (please also fix this in other methods)
```suggestion
        for n_rows_in_slot in total_number_of_rows:
```

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+    def _get_columns(self) -> List[List[str]]:
+        """
+        Returns the elements of each column.
+        """
+        total_number_of_columns = self._get_number_of_columns()
+        column_list = []
+        for _, item in enumerate(total_number_of_columns):
+            columns_in_each_slot = []
+            for i in range(1, item + 1):
+                if i < 10:
```
astafan8: there's no need for `if i < 10`, the formatting string will do the trick - `f'{i:02d}'`. (there are a occasions of this in other methods as well)

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
astafan8: should this helper method really return strings? I saw that in maany places the conversion to `int` is done, so perhaps returning ints is more optimal?

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+        slot_id = self._get_slot_id()
+        row_list = self._get_rows()
+        column_list = self._get_columns()
+        matrix_channels = []
+        for i, slot in enumerate(slot_id):
+            for element in itertools.product(slot, row_list[i], column_list[i]):
+                matrix_channels.append(''.join(element))
+        return matrix_channels
+
+    def get_channels_by_slot(self, val: int) -> List[str]:
```
astafan8: for readability?
```suggestion
    def get_channels_by_slot(self, slot_id: int) -> List[str]:
```

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+        backplane_common_number = '9'
+        backplane_relay_common_numbers = ['11', '12', '13', '14', '15', '16']
+        slot_id = self._get_slot_id()
+        analog_backplane_relays = []
+        for element in itertools.product(slot_id, backplane_common_number,
+                                         backplane_relay_common_numbers):
+            analog_backplane_relays.append(''.join(element))
+        return analog_backplane_relays
+
+    def connect_row_to_columns(self, slot_id: int, row_id: int,
```
astafan8: i think since there are these `connect_*` convenience methods, it make sense to also have `disconnect_*` methods. At least in the drivers for other switch matrices in qcodes repo `connect` method also comes with a matching `disconnect`. Would you agree?

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
astafan8: @jdwatson do you think that `connect_column_to_rows`/`connect_row_to_columns` fulfill the request? I think they do (expecially if `disconnect` methods will be added).

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+    def _set_lan_status(self, val: Union[str, bool]) -> None:
+        self.write(f'comm.lan.enable = {val}')
+
+    def _get_gpib_address(self) -> int:
+        return int(float(self.ask('gpib.address')))
+
+    def _set_gpib_address(self, val: int) -> None:
+        self.write(f'gpib.address = {val}')
+
+    def get_closed_channels(self, val: str) -> Union[str, None]:
```
astafan8:
```suggestion
    def get_closed_channels(self, val: str) -> Optional[str]:
```

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+        """
+        Queries for the closed channels.
+
+        Args:
+            val: A string representing the channels,
+                backplane relays or channel patterns to be queried.
+        """
+        if val == "":
+            raise InvalidValue('Argument cannot be an empty string.')
```
astafan8: isn't this captured by the `_validator`?

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+        if not self._validator(val):
+            raise InvalidValue(f'{val} is not a valid specifier. '
+                               'The specifier should be channels, channel '
+                               'ranges, slots, backplane relays or "allslots".')
+        data = self.ask(f"channel.getclose('{val}')")
+        if data == 'nil':
+            return None
+        return data
```
astafan8: i'm a bit confused here. This method returns a potentially-comma-separated string of "specifiers".  However, helper methods like `get_channels` return a `List` of "specifiers", thus avoiding the comma-separated-ness. Shall the return type be unified across methods? And if yes, I'd vote for "list of 'specifiers'" in order to hide the comma-separated-ness from users, and also already provide them with and iterable thing (the `List`) that does not require the users to `.split(',')` it first.

docs/examples/driver_examples/QCoDeS example with Keithley 3706A System Switch.ipynb
```diff
@@ -0,0 +1,2413 @@
...
+    "These channels cannot be closed:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 51,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "smatrix.close_channel(\"1101\")"
```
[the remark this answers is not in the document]
GateBuilder: That is the thing nothing happens, which is handled by the instrument, simply you can cannot close it. Thus, no need for an exception. But I can put a warning.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
+import itertools
+import warnings
+from typing import List, Dict, Optional, Union, Tuple
+
```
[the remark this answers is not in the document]
GateBuilder: At start, I implemented a channels subclass. Then I figured that all the channels are in fact controlled and manipulated via the same command. Therefore, I removed the channels subclass but let the name stay the same as the driver indeed is all about channels.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
GateBuilder: No, because "SYSTEM SWITCH" is a custom addition.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
GateBuilder: Instrument returns a string "1.0".

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
GateBuilder: I used slot_id as the general name for the available slots within the method. nevertheless I changed val to slot_no.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
GateBuilder: Disconnect methods would be the replica of the connect methods with one line diff, thus bad code. I intentionally let these functions to return (which is not needed for it to function) the list of connected channels (i.e. opened channels) so that user pass these lists to close_channels methods directly without modification to disconnect. 

However, what can be done is that same functions can be used for both connect and disconnect via a boolean thus avoiding replication?

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
GateBuilder: No because this is an edge case where in fact the instrument returns a silent error code, thus implemented individually here.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
astafan8: ok, sad but true

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
astafan8: hmm, do we care about this custom addition? :)

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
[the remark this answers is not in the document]
astafan8: ok, then as per convention of calling qcodes isntrument driver modules, the name should only contain the brand and the model (unless really necessary), so please remove the `*_Matrix_channels` part.

docs/examples/driver_examples/QCoDeS example with Keithley 3706A System Switch.ipynb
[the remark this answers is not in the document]
astafan8: > But I can put a warning.

a warning seems to be useful becuase imho such a behavior might hide unwanted "attempts to close forbidden channels".

And if a forbidden channel is attempted to be closed, doesn't the instrument put at least some error message in its error queue that we can read? or smth similar?

docs/examples/driver_examples/QCoDeS example with Keithley 3706A System Switch.ipynb
```diff
@@ -0,0 +1,2413 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 73,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "C:\\Users\\a-alhard\\miniconda3\\envs\\qcodes\\lib\\site-packages\\ipykernel_launcher.py:1: UserWarning: The hardware interlock in Slot 1 is disengaged. The corresponding analog backplane relays cannot be energized.\n",
```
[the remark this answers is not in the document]
GateBuilder: Yes, since there is no way in this tutorial that I can engage any slot. But also this information already printed out immediately after connection. Anyways, I will add a single sentence.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+    def __init__(self, name: str, address: str, **kwargs) -> None:
+        """
+        Args:
+            name: Name to use internally in QCoDeS
+            address: VISA resource address
+        """
+        super().__init__(name, address, terminator='\n', **kwargs)
+
+        self.add_parameter('reset_channel',
```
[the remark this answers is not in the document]
GateBuilder: As these are the most fundamental channel operations that the switch provides, I thought that they should be in snapshot.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
```diff
@@ -0,0 +1,709 @@
...
+                           set_cmd=self._set_exclusive_slot_close,
+                           docstring='Closes the specified channels on the '
+                                     'associated slots abd opens any other '
+                                     'channels if they are not specified by '
+                                     'the parameter.')
+
+        self.add_parameter('channel_connect_rule',
+                           get_cmd=self._get_channel_connect_rule,
+                           set_cmd=self._set_channel_connect_rule,
+                           docstring='Controls the connection rule for closing '
```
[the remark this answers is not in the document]
GateBuilder: thanks! great advice.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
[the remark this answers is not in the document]
astafan8: > they should be in snapshot

this is a great intention, but if I `open_channel` and then `close_channel` for same channels, snapshot will contain that channel id in both open and close parameters - is this useful? yes, one can look at the timestamps, but why making it so complicated.

so, i suggest for simlicity just make them methods and not worry about the snapshot. anyway, some kind of "all_current_connections" is a better candidate for a parameter that gets snapshotted to preserve the state of connecetions before the measurement is run.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
GateBuilder: Here I prefer the current implementation --the logic flow-- as it is more readable, controllable and understandable than compact format method. Plus, there will be no performance gain.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
GateBuilder: No, return of channel lists depend on strings as itertools are used.

qcodes/instrument_drivers/tektronix/Keithley_3706A_Matrix_channels.py
astafan8: this is not about performance, this is about "not reinventing the wheel" and using existing proven/tested methods.

alright, at least we have a test that covers this code, right?

qcodes/instrument_drivers/tektronix/Keithley_3706A.py
```diff
@@ -0,0 +1,742 @@
...
+        backplane_common_number = '9'
+        backplane_relay_common_numbers = ['11', '12', '13', '14', '15', '16']
+        slot_id = self._get_slot_ids()
+        analog_backplane_relays = []
+        for element in itertools.product(slot_id, backplane_common_number,
+                                         backplane_relay_common_numbers):
+            analog_backplane_relays.append(''.join(element))
+        return analog_backplane_relays
+
+    def connect_or_disconnect_row_to_columns(self, action: str,
```
astafan8: could you explain the choice for the API being "one method for connect/disconnect actions with first argument being the flag"? as opposed to, for example, "two separate methods, one for connect, other for disconnect" (removing the need for the flag argument and making the method names shorter and more direct).

GateBuilder: I already explained above, while responding to the relevant query, that creating a separate method will be a code repetition. If we assume for a moment that we have two functions for one to connect and for other to disconnect, these function will differ, effectively, only with a single line, i.e., execution of connect/disconnect. Thus, the latter implementation, IMHO, would decrease the code quality. With the current implementation, we have the desired functionality without duplication. I also added the usage of these functions to the example notebook, so users should be able to benefit from them in no time.

vyaskhyati2019:
@astafan8 @GateBuilder this has been in PR review since Jan 3rd. Anything we can do to support turnaround?

qcodes/instrument_drivers/tektronix/Keithley_3706A.py
jenshnielsen: I suggest that we make 2 individual methods that both call this one with the appropriate flat set and then make this method private. That will not give code repetition and create a cleaner api

jenshnielsen:
Pipelines passed but the update did not propagate back due to a server side issue at github https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=8376&view=results

PR #1887, opened 2020-01-06
Try to improve Azure pipelines speed on windows.

jenshnielsen:
As suggested here https://github.com/microsoft/azure-pipelines-tasks/issues/8456 it may be faster to install the env to the d drive


jenshnielsen:
This did not speed anything up. Seems to actually make it slower

PR #1891, opened 2020-01-13
Upgrade pyqt5 to 5.12 which has support for python 3.8

jenshnielsen:
This will hopefully enable us to run qt related tests on python 3.8 and eventually build the docs too. 

Pyqt 5.13 and 5.14 have been released but are not yet supported by spyder

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ matplotlib==3.1.*
 pyqtgraph==0.10.0
 PyVISA==1.10.1
 h5py==2.9.0
-PyQt5==5.9.*;python_version<'3.8' # there are currently no pyqt5 packages build for 3.8 on pypi
+PyQt5==5.12.*
 QtPy
 jsonschema
 ruamel.yaml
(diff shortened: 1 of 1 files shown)


jenshnielsen:
This does mean that the pyqt version used in conda and pip builds will diverge as conda still ships 5.9

astafan8:
> This does mean that the pyqt version used in conda and pip builds will diverge as conda still ships 5.9

shall we wait for `conda`? or do we prioritize availability on `python 3.8`?

jenshnielsen:
L

> 
> 
> > This does mean that the pyqt version used in conda and pip builds will diverge as conda still ships 5.9
> 
> shall we wait for `conda`? or do we prioritize availability on `python 3.8`?

Lets see if this actually works. I am getting tired of waiting on conda. This seems to be very low priority for them

jenshnielsen:
Well this segfaulted with all 3 python versions so ...

astafan8:
> Well this segfaulted with all 3 python versions so ...

i'm not sure with what smiley to reply :)

jenshnielsen:
It seems like we are seeing segfaults with pyqt5.12/13/14 for a simple matplotlib plot. It's not clear if this is a qt bug or specifically related to travis and the xvfb setup that we use

PR #1895, opened 2020-01-19
Move doNd of Wrappers to the Core

GateBuilder:
This PR moves the doNd functions of qdev-wrappers to the core with the corresponding history and the related tests.

@Dominik-Vogel @astafan8 @jenshnielsen @WilliamHPNielsen


qcodes/utils/ndim_analysis/dataset/doNd.py
```diff
@@ -0,0 +1,309 @@
...
+    return _handle_plotting(datasaver, do_plot)
+
+
+
+
+
+def do1d(
+    param_set: _BaseParameter, start: float, stop: float,
+    num_points: int, delay: float,
+    *param_meas: ParamMeasT,
```
ThorvaldLarsen: Is it possible to pass in a function that can break the measurement? I might want to just look at the result of the latest measurement and if it passes some threshold break the measurement. Creating the function to do the check can easily be done outside the doNd functions but what should the flag/function return be that correctly stops the measurement and exists with save of dataset and figures?

qcodes/utils/ndim_analysis/dataset/doNd.py
```diff
@@ -0,0 +1,309 @@
...
+    experiment_name = datasaver._dataset.exp_name
+    sample_name = datasaver._dataset.sample_name
+    storage_dir = os.path.join(mainfolder, experiment_name, sample_name)
+    os.makedirs(storage_dir, exist_ok=True)
+    png_dir = os.path.join(storage_dir, 'png')
+    pdf_dif = os.path.join(storage_dir, 'pdf')
+    os.makedirs(png_dir, exist_ok=True)
+    os.makedirs(pdf_dif, exist_ok=True)
+    save_pdf = True
+    save_png = True
```
ThorvaldLarsen: Is it possible to add these to the config file so one can choose to save either png, pdf, or both?

qcodes/utils/ndim_analysis/dataset/doNd.py
```diff
@@ -286,7 +286,7 @@ def _handle_plotting(
     return res
 
 
-def _create_plots(data: DataSet) -> Tuple[DataSet, list, list]:
+def get_plots(data: DataSet) -> Tuple[DataSet, list, list]:
```
Dominik-Vogel: I am not very happy with this name, because it suggests that the plots already exist, and it does not communicate well that it will generate plots on disk, if specified.
Maybe we can simply call it `plot`, which is well defined in the context of the module and is very short.
We should also add all parameters that can be passed into `plot_dataset`.

Dominik-Vogel:
@GateBuilder is there a special reason you decided to put the donds into a folder called `ndim_analysis`? I don't really see how this is related to analysis.

qcodes/utils/ndim_analysis/dataset/doNd.py
GateBuilder: Sure. The name was just a placeholder until such a review.

Dominik-Vogel:
I did not mean to question the `ndim` part. Maybe a potential home could be the root of utils? It feels to me that this is quite an important util.

astafan8:
> I am open to any suggestions, though.

allow me to add a new suggestion - `qcodes/dataset/dond.py` where `dond` may potentially be a package. as an extenstion to `measurements.py` that uses the tools of `qcodes.dataset.*`, i think donds can nicely live in `qcodes.dataset.*`. we can go even further, and put them in qcodes itself, `qcodes.dond.*` :) What do you think?

and just to respond to other suggestions mentioned before. the word analysis in ndim_analysis compltetely distracts me and i wouldn't know why dond, smth realted to a measurement-related utility would be in some ndim_analysis module. I also don't think that we should polllute utils with all the utils possible, in my head utils are subtly relevant to the main package and are not really its significant parts - donds are more of an extension of the package (qcodes) that are so close to it that deserve to be in it (not as separate qcodes-dond repo/packgage), but should be visible (unlike the utils dumpster).

Dominik-Vogel:
@astafan8 you are speaking right from my heart! I like your idea way better than mine. I would however not go that far to put them in the main qcodes module (if you disagree, I'll try to find and argument ;-)). I think it would be good to give them their own subpackage=folder, because that file is already filled enough and I can easily see how it will keep on growing. I wish we had done the same for the station, because migrating is not so much fun.

jenshnielsen:
I don't have any strong opinion other than that they should **not** be called analysis. The fact that they can be used for some lightweight analysis does not change the fact that they are primarily intended to do measurements with

qcodes/utils/dataset/doNd.py
```diff
@@ -299,13 +298,19 @@ def _handle_plotting(
     return res
 
 
-def get_plots(data: DataSet) -> Tuple[DataSet, list, list]:
+def plot_data(data: DataSet, save_pdf: bool = True,
```
Dominik-Vogel: Shall we call this public function  simply `plot`? The 'data' part of the name does not give any additional information if you ask me.

docs/examples/DataSet/Implementing_doND_using_the_dataset.ipynb
```diff
@@ -201,71 +247,134 @@
...
-   "execution_count": 32,
+   "cell_type": "markdown",
    "metadata": {},
-   "outputs": [],
    "source": [
-    "def do2d(param_set1, start1, stop1, num_points1, delay1, \n",
-    "         param_set2, start2, stop2, num_points2, delay2,\n",
-    "         *param_meas):\n",
-    "    # And then run an experiment\n",
+    "Noe that `do_plot = True` if the default value and we only added it for the sake of explicitness. The `doNd` utility, in fact, uses an internal public function called `plot_data` to handle data visualization, which by default saves the figures in both pdf and png formats, as well. Therefore, users, if needed, can only do their measurements with the option `do_plot = False` and then pass the relevant dataset to `plot_data` when visualization is necessary and choose to save figures in a desired format.\n",
```
Dominik-Vogel: typo: noe

qcodes/utils/dataset/doNd.py
GateBuilder: Sure, will do.

docs/examples/DataSet/Implementing_doND_using_the_dataset.ipynb
GateBuilder: Thanks.

jenshnielsen:
I suggest that we change the notebook so that its no longer called implementing... but rather something like "using the dond functions for running experiments ...` There will also be the need to change some of the text slightly to reflect this

PR #1896, opened 2020-01-20
Feature: Out-of-thread data saving

WilliamHPNielsen:
This PR introduces a simple mechanism for writing data to the database in a separate background thread. The `DataSet` objects gets a `Queue` to hold results and a `Write(Thread)` to consume the queue and write to the DB. The writer thread holds its own connection.

Design choices:

- The measurement context manager waits for the writing to finish as it exits
  - In particular: the waiting happens as the `DataSet` is marked complete

- Add out-of-thread data saving per measurement

@QCoDeS/core 

I have manually goofed around with `SIGINT`s and I haven't seen anything worrisome.


qcodes/dataset/data_set.py
```diff
@@ -205,6 +206,36 @@ def _clean_up(self) -> None:
...
+            item = self.queue.get()
+            if item['keys'] == 'stop':
+                self.keep_writing = False
+                self.conn.close()
+            else:
+                self.write_results(item['keys'], item['values'])
+            self.queue.task_done()
+
+    def write_results(self, keys: List[str],
+                      values: List[List[Any]]) -> None:
```
jenshnielsen: I suggest using Sequence as the type for input arguments to make it clear that they are not modified

qcodes/dataset/data_set.py
```diff
@@ -255,6 +286,7 @@ def __init__(self, path_to_db: str = None,
         self.subscribers: Dict[str, _Subscriber] = {}
         self._interdeps: InterDependencies_
         self._parent_dataset_links: List[Link]
+        self._data_write_queue: Queue = Queue()
```
jenshnielsen: I am surprised that this type annotation is needed.

qcodes/dataset/data_set.py
```diff
@@ -753,6 +799,38 @@ def add_results(self, results: Sequence[Mapping[str, VALUE]]) -> int:
                            values)
         return len_before_add
 
+    def add_result_to_queue(self,
```
jenshnielsen: What will happen is someone calls `add_result_to_query` for a dataset that is not created with `start_bg_writer`=True ?

qcodes/dataset/measurements.py
```diff
@@ -672,6 +703,8 @@ def __exit__(self,  # type: ignore[no-untyped-def]
 
             # and finally mark the dataset as closed, thus
             # finishing the measurement
+            # Note that the completion of a dataset entails waiting for the
+            # write thread to terminate (iff the write thread has been started)
```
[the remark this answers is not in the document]
jenshnielsen: iff is normally shorthand for if and only if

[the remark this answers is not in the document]
astafan8: ohhhh, ok, then nevermind

qcodes/dataset/data_set.py
astafan8: this comment made me wonder whether the "queue" part should be explicit on datasaver's and dataset's public interface (as these methods like `add_result_to_queue`) or should this be hidden behind the existing methods like `add_results`?

qcodes/dataset/measurements.py
```diff
@@ -1190,7 +1223,7 @@ def add_subscriber(self: T,
 
         return self
 
-    def run(self) -> Runner:
+    def run(self, write_in_background: bool = False) -> Runner:
         """
         Returns the context manager for the experimental run
```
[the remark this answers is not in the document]
WilliamHPNielsen: Thanks!

qcodes/tests/dataset/test_datasaver.py
```diff
@@ -121,4 +131,5 @@ def test_saving_numeric_values_as_text(numeric_type):
         with pytest.raises(ValueError, match=msg):
             data_saver.add_result((p.name, value))
     finally:
+        test_set.mark_completed()
```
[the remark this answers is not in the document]
WilliamHPNielsen: `write_in_background` is `False` to keep the behaviour unchanged from what it was before, i.e. to see that if we disable the new feature, we really get back to how the code was before.

Good catch regarding the completion. It was needed before 79c6798 but just stayed around. I'll remove for clarity.

qcodes/tests/dataset/test_string_data.py
```diff
@@ -44,7 +44,8 @@ def test_string_via_datasaver(experiment):
     idps = InterDependencies_(standalones=(p,))
 
     data_saver = DataSaver(
-        dataset=test_set, write_period=0, interdeps=idps)
+        dataset=test_set, write_period=0, interdeps=idps,
+        write_in_background=False)
```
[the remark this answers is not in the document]
WilliamHPNielsen: Same response as above.

qcodes/dataset/data_set.py
WilliamHPNielsen: The signature of `write_results` here is made to match the signature of `insert_many_values` (source: https://github.com/QCoDeS/Qcodes/blob/61eb02f460a3777fa7d247d85201cab78970cabb/qcodes/dataset/sqlite/query_helpers.py#L151). Should we change both?

jenshnielsen: Yes, I think those are from times where we knew less of mypy

WilliamHPNielsen:
From an offline discussion, the following strategy emerged:

- On the `DataSet`, we put both direct writes and queue-putting under `add_result`. The decision will be based on whether the writer thread is alive.
- On the `DataSaver` side, most of the newly added code can disappear. The only thing the `write_in_background` flag will do is to set the write period to zero and to mark the dataset as started with the flag to spawn the write thread.

qcodes/dataset/data_set.py
WilliamHPNielsen: Done.

qcodes/dataset/data_set.py
WilliamHPNielsen: The two bullet points in the comment below addressing the concerns here have now been implemented.

qcodes/dataset/data_set.py
```diff
@@ -667,8 +709,15 @@ def mark_completed(self) -> None:
             raise RuntimeError('Can not mark DataSet as complete before it '
                                'has been marked as started.')
         self.completed = True
+        self._perform_completion_actions()
```
jenshnielsen: Logically it seems like these two lines should be reordered. Not sure that it matters

qcodes/dataset/data_set.py
```diff
@@ -748,11 +794,45 @@ def add_results(self, results: Sequence[Mapping[str, VALUE]]) -> int:
         expected_keys = frozenset.union(*[frozenset(d) for d in results])
         values = [[d.get(k, None) for k in expected_keys] for d in results]
 
-        len_before_add = length(self.conn, self.table_name)
+        # len_before_add = length(self.conn, self.table_name)
```
jenshnielsen: delete?

qcodes/dataset/data_set.py
WilliamHPNielsen: Good point!

qcodes/dataset/data_set.py
WilliamHPNielsen: Whoops, haha 😄

PR #1898, opened 2020-01-21
Add numpy_stubs to type checking

jenshnielsen:
And fix the last few issues for that to work.


test_requirements.txt
```diff
@@ -10,4 +10,5 @@ codecov
 asv
 gitpython
 pylint
-attrs>=19.2.0 # needed for hypothesis >=4.38.1 but not correctly honored by pip 19.2.3
\ No newline at end of file
+attrs>=19.2.0 # needed for hypothesis >=4.38.1 but not correctly honored by pip 19.2.3
+git+https://github.com/numpy/numpy-stubs.git@f3c6315738489983f5f37e1477ac68373d71b470 # no release yet so fix to a known working version
```
astafan8: just for my understanding, if this is not installed, then the type checking won't simply be as rigorous, right?

jenshnielsen: Yes that is indeed the case. Basically if this is not installed all numpy types will be equivalent to Any

PR #1899, opened 2020-01-21
add initialised_database_at contextmanager

basnijholt:
Often one does
```python
db_location = qc.config["core"]["db_location"]
try:
    qc.initialise_or_create_database_at(db_file)
    some_operations_here(...)
finally:
    qc.config["core"]["db_location"] = db_location
```
this can be replaced with:
```python
with initialised_database_at(db_file):
    some_operations_here(...)
```
which is much cleaner!

I am using this in another code, however, @astafan8 recommended me to create a PR 👍


basnijholt:
Sounds good! I've amended my commit and force pushed.

PR #1906, opened 2020-01-22
Add new features to keithley 2450

liangosc:
- add reset() method to the main module,
- add nplc parameter to the sense module,
- add user delay to the sense module,
- add delay, user delay, and auto delay setting to the source module.

The newly added user delay to both sense and source module can be used in the trigger model. However, the trigger model does not exist in the qcodes driver yet. So my next step is to write the trigger module for the driver, and a new example notebook for how to perform triggering.

Speaking of example notebook, the new features added in the PR do not really fit in the existing example notebook (methods like the reset() are well self-explained). I'll try to include as many as I can in the upcoming example notebook for trigger module.

@sohailc

### DIFF
--- a/qcodes/instrument_drivers/tektronix/Keithley_2450.py
+++ b/qcodes/instrument_drivers/tektronix/Keithley_2450.py
@@ -87,6 +87,20 @@ class Sense2450(InstrumentChannel):
             parameter_class=ParameterWithSetpoints
         )
 
+        self.add_parameter(
+            "nplc",
+            get_cmd=f":SENSe:{self._proper_function}:NPLCycles?",
+            set_cmd=f":SENSe:{self._proper_function}:NPLCycles {{}}",
+            vals=Numbers(0.001, 10)
+        )
+
+        self.add_parameter(
+            "delay",
+            get_cmd=f":SENSe:{self._proper_function}:DELay:USER1?",
+            set_cmd=f":SENSe:{self._proper_function}:DELay:USER1 {{}}",
+            vals=Numbers(0, 1e4)
+        )
+
     def _measure(self) -> str:
         if not self.parent.output_enabled():
             raise RuntimeError("Output needs to be on for a measurement")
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -109,6 +131,22 @@ def clear_trace(self) -> None:
...
+
+    def set_user_number(self, value) -> None:
+        self._user_number = value
+
+    def get_user_delay(self) -> str:
+        get_cmd = f":SENSe:{self._proper_function}:DELay:USER" \
+                  f"{self.user_number()}?"
+        return self.ask(get_cmd)
+
+    def set_user_delay(self, value) -> None:
```
astafan8: do these methods need to be public? if this dfunctionality is supposed to be accessed via parameters, then please make these methods private .

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -109,6 +131,22 @@ def clear_trace(self) -> None:
         """
         self.write(":TRACe:CLEar")
 
+    def get_user_number(self) -> int:
+        return self._user_number
```
astafan8: the functionality can be implemented via `set_cmd=None, get_cmd=None` without the need for a private attribute.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
liangosc: Good point. User should only call the parameters. I have set them to private.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
liangosc: Nice! I was hoping there was a simpler way to do it. I have implemented as you suggested.

liangosc: I also removed the self._user_number=1 in the __init__ because the user number should be a required input parameter, not with a default value of 1.

PR #1907, opened 2020-01-27
Complete typing of logging module

jenshnielsen:
and enforce this via mypy config


jenshnielsen:
While doing this I looked up the right types for `__exit__` methods in context managers so apply this to other context managers that are missing it.

PR #1908, opened 2020-01-28
Add types to monitor 

jenshnielsen:

### DIFF
--- a/qcodes/monitor/monitor.py
+++ b/qcodes/monitor/monitor.py
@@ -26,7 +26,7 @@ import os
 import time
 import json
 from contextlib import suppress
-from typing import Dict, Any, Optional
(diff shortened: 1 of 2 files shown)


qcodes/monitor/monitor.py
```diff
@@ -40,7 +40,8 @@
 from qcodes.instrument.parameter import Parameter
 
 
-def _get_all_tasks():
+def _get_all_tasks() -> Callable[[Optional[asyncio.AbstractEventLoop]],
+                                 Set[asyncio.Task]]:
```
jenshnielsen: This now trips up python 3.6. We should probably rewrite this with `sys.version_info` as suggested by the mypy docs https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks

PR #1913, opened 2020-02-03
Complete the typing of the qcodes station 

jenshnielsen:
@QCoDeS/core This should finaly be ready for review. 

I did not see a clean way to deprecate the access of one specific key in the `station.components` dict so this comes with a breaking change that `station.components[`config`]` is no longer there. Let me know if you have other suggestions


qcodes/station.py
```diff
@@ -118,16 +123,16 @@ def __init__(self, *components: Metadatable,
         if default:
             Station.default = self
 
-        self.components: Dict[str, Metadatable] = {}
+        self.components: Dict[str, Union[Metadatable, 'StationConfig']] = {}
```
astafan8:
```suggestion
        self.components: Dict[str, Union[Metadatable, StationConfig]] = {}
```

qcodes/station.py
```diff
@@ -76,6 +77,10 @@ class ValidationWarning(Warning):
     pass
 
 
+class StationConfig(UserDict):
```
astafan8: I'd vote for making first a quick PR that fixes this (or just in this PR), and then we won't need to have the ugly union in the station.components
```suggestion
class StationConfig(Metadatable):
```

jenshnielsen: @Dominik-Vogel Do you have any opinion on this? I am happy to make the change but I may be missing the context

jenshnielsen: I would more lean towards removing config from components and having it as a specific separate entry in the station and update snapshot_base to incorporate it

astafan8: > I would more lean towards removing config from components and having it as a specific separate entry in the station and update snapshot_base to incorporate it

agree, let's not pollute "components".

Dominik-Vogel: I am also all for that suggestion.

PR #1914, opened 2020-02-03
Unique name for instrument logger

jenshnielsen:
This ensures that each instrument will have it's own logger so that messages can easily be filtered by instrument instance

An instrument will have a logger with a name of the form `module.path.to.instrument.classname-instancename`
and in the same way a channel will have a logger with a name of the form `module.path.to.instrument.classname-instancename.channelclassname-channelinstancename`

@Dominik-Vogel


jenshnielsen:
If we go that way there should be some tests added

qcodes/instrument/base.py
```diff
@@ -58,8 +58,8 @@ def __init__(self, name: str,
 
         # This is needed for snapshot method to work
         self._meta_attrs = ['name']
-
-        self.log = get_instrument_logger(self, __name__)
+        log_name = ".".join(self._logger_name_parts)
```
astafan8: so this delimiter is to connect parent instrument and the channel, right? then i'd suggest to not use `.` as delimiter becuase it may get confused with the module/classname.
```suggestion
        log_name = "--".join(self._logger_name_parts)
```

jenshnielsen: I don't think that will work. We explicit want the channel logger to be a child of the instrument logger. As I understand it that depends on using `.` as a separator

jenshnielsen: But I will check

jenshnielsen: A quick manual test verified that this is indeed the case. To establish the child-parent relationship the names must be separated by `.`. This should be incorporated into the tests

astafan8: This is unfortunate :( but great to know and yes , shouldbe tested for :)

jenshnielsen:
This is now replaced by https://github.com/QCoDeS/Qcodes/pull/5297 which will ensure that the instrument_name is an extra attribute on the logrecord from the instrument LoggerAdapter

PR #1918, opened 2020-02-10
Improvements to 344xx driver

spauka:
This change implements a fix to the 344xx driver that stops the DMM from cycling through all its measurement modes on a snapshot/initialization.

- Don't include measurements in a snapshot that would cause the measurement mode to change.
- Make ask abstract in the AbstractInstrument class to fix a bunch of pylint errors.

### DIFF
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -2,7 +2,7 @@
 import time
 import weakref
 import logging
-from abc import ABC
+from abc import ABC, abstractmethod
 from typing import Sequence, Optional, Dict, Union, Callable, Any, List, \
     TYPE_CHECKING, cast, Type
 
@@ -395,6 +395,7 @@ class AbstractInstrument(ABC):
     """ABC that is useful for defining mixin classes for Instrument class"""
     log: 'InstrumentLoggerAdapter'  # instrument logging
 
+    @abstractmethod
     def ask(self, cmd: str) -> str:
         pass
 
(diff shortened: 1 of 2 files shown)


jenshnielsen:
@spauka Thanks, I think this looks fine. One question is would it worth the trobuble to switch which parameters are snapshotted depending on the sense mode? The yokogava gs200 driver does something along those lines as far as I remember. In general you don't worry about pylint errors that requires markup inline to fix. It's not a requirement that codacy comes back clean. We specifically only run pylint on the diff of a pr for this reason. 

Here it's specifically caused by the way that get and set are automatically constructed from parameters.

spauka:
@jenshnielsen Thanks for checking :) Regarding snapshotting, the previously read value is still stored in the snapshot as part of the cache, however, it is not updated when connecting to the instrument for the first time or when performing a snapshot, i.e.:
```ipynb
In[12]: dmm.volt()
Out[12]: 0.00179960528

In[13]: snap = station.snapshot(update=True)

In[14]: snap["instruments"]["dmm1"]["parameters"]["volt"]["value"]
Out[14]: 0.00179960528
```

It is not clear that it would be desirable to read a new value anyway, since depending on the aperture/NPLC setting of the DMM, it could be a long running operation.

The logic in the GS200 driver was slightly different, since we wanted to prevent someone from trying to set the current when in voltage mode and vice versa, hence we needed to keep track of the state of the instrument.

However, let me know if there's any changes you would like made, I can implement a cached mode if you like or revert linting fixes?

jenshnielsen:
Hmn yes I don't think it makes sense to have any of these values in the snapshot. But perhaps we should just add a small comment above the first of them explaining why they are excluded for future reference?

spauka:
Agreed and done :)

PR #1922, opened 2020-02-11
Fix type of timestamp in monitor

### LINKED ISSUE #1921: QCoDeS monitor broken on master
The qcodes monitor is broken on the latest master. The error is occuring somewhere in the javascript as the data is still correctly sent over the websocket.

### Steps to reproduce
1. Create a monitor in QCoDeS.
2. Launch the monitor using `python -m qcodes.monitor.monitor`

### Expected behaviour
QCoDeS monitor displays in webbrowser.

### Actual behaviour
I get an error message in both Firefox and IE. (I don't have a copy of chrome to test on my experimental PC.

The error is: 
```
Status:error:I ran into the following problems at _.parameters[0].parameters[34].ts: Expecting null but instead got: "1581406601.743032" Expecting a Float but instead got: "1581406601.743032"
```

The issue seems to have crept in between revisions e5c5324 and 1ce7a23.

I've attached a sample response from the websocket in the working state and the broken state, where you can see that the timestamp is being sent as a string instead of a float...

[websocket_brokensample.txt](https://github.com/QCoDeS/Qcodes/files/4184596/websocket_brokensample.txt)
[websocket_workingsample.txt](https://github.com/QCoDeS/Qcodes/files/4184597/websocket_workingsample.txt)

### System
Windows 10
QCodes Master 33d4e71

spauka:
Fixes #1921 .

- Corrects type of timestamp in the monitor

This could also be fixed in the web frontend code, but I can't seem to find that anywhere...

@jenshnielsen

### DIFF
--- a/qcodes/monitor/monitor.py
+++ b/qcodes/monitor/monitor.py
@@ -33,6 +33,7 @@ import asyncio
 from asyncio import CancelledError
 from threading import Thread, Event
 
+import socketserver
 import webbrowser
 import websockets
 
@@ -266,7 +267,7 @@ if __name__ == "__main__":
     os.chdir(STATIC_DIR)
     try:
         log.info("Starting HTTP Server at http://localhost:%i", SERVER_PORT)
-        with http.server.HTTPServer(("", SERVER_PORT),
+        with socketserver.TCPServer(("", SERVER_PORT),
                                     http.server.SimpleHTTPRequestHandler) as httpd:
             log.debug("serving directory %s", STATIC_DIR)
             webbrowser.open("http://localhost:{}".format(SERVER_PORT))
--- a/qcodes/monitor/monitor.py
+++ b/qcodes/monitor/monitor.py
@@ -26,14 +26,13 @@ import os
 import time
 import json
 from contextlib import suppress
-from typing import Dict, Any, Optional, Sequence, Callable, Awaitable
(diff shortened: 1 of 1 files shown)


jenshnielsen:
Thanks @spauka Just left one question about the changes to socket server.

qcodes/monitor/monitor.py
```diff
@@ -267,7 +266,7 @@ def show() -> None:
     os.chdir(STATIC_DIR)
     try:
         log.info("Starting HTTP Server at http://localhost:%i", SERVER_PORT)
-        with socketserver.TCPServer(("", SERVER_PORT),
```
jenshnielsen: I don't think this change is needed. There is a bug in typing shed with mypy 0.761 that breaks the types around here. I have submitted a pr for that here so it should be fixed in the next version. https://github.com/python/typeshed/pull/3579/files this is why qcodes is checked with mypy 0.750 btw

spauka: Haha sounds good 😄 do you think it's better to revert this change? I am happy for you to do that, or I can too :P

jenshnielsen: Yes if you can confirm that you only made the change to keep mypy happy lets revert it. I am happy to do so

spauka: Awesome, I just left work so feel free :+1:

PR #1927, opened 2020-02-18
Changelog for 0.11.0 release

astafan8:


jenshnielsen:
We need to either delete the 0.12 file or include it in the index to pass the docs build

PR #1929, opened 2020-02-25
Fix: Keysight_34465A overload value 9.9e37 to be np.inf instead

lakhotiaharshit:
In case the DMM tries to measure the voltage value outside the its range, it overloads and throws 9.9e37 instead of throwing an error. Here we substitute this number with np.nan. 

Side note: The error query does not throw any error even though the instrument display shows in bold that instrument is overloaded. 

 - [x] Test on the instrument  

To reproduce the error: 

@jenshnielsen @astafan8

### DIFF
--- a/qcodes/tests/drivers/test_keysight_34465a.py
+++ b/qcodes/tests/drivers/test_keysight_34465a.py
@@ -17,15 +17,11 @@ def driver():
     try:
         yield keysight_sim
     finally:
-        keysight_sim.close()
+        Keysight_34465A.close_all()
 
 
 @pytest.fixture(scope='function')
-def driver_volt(val_volt):
-    keysight_sim = Keysight_34465A('keysight_34465A_sim',
-                                   address='GPIB::1::INSTR',
-                                   visalib=visalib)
-
+def driver_volt(driver, val_volt):
     def get_ask_with_read_mock(original_ask, read_value):
         def ask_with_read_mock(cmd: str) -> str:
             if (cmd == "READ?") or (cmd == "FETCH?"):
@@ -34,11 +30,8 @@ def driver_volt(val_volt):
                 return original_ask(cmd)
         return ask_with_read_mock
 
-    keysight_sim.ask = get_ask_with_read_mock(keysight_sim.ask, val_volt)
-    try:
-        yield keysight_sim
-    finally:
-        keysight_sim.close()
+    driver.ask = get_ask_with_read_mock(driver.ask, val_volt)
+    yield driver
 
 
 def test_init(driver):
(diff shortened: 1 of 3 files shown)


jenshnielsen:
thanks @lakhotiaharshit 

Does this work if you read out a buffer with multiple datapoints?

jenshnielsen:
What happens with negative voltages. Is that even supported. Does it return a negative value?

lakhotiaharshit:
> thanks @lakhotiaharshit
> Does this work if you read out a buffer with multiple datapoints?

Yes, I think I have fixed that part in the code. I am also planning to test it on the actual instrument in the evening to have the final confirmation. Do not know about the negative voltage though.

lakhotiaharshit:
> What happens with negative voltages. Is that even supported. Does it return a negative value?

Yes, fixed the PR to support negative voltage overload also.

jenshnielsen:
So the question that I didn't really raise but is relevant when you consider positive and negative values is. Should this not return +- inf rather than nan ?

jenshnielsen:
Looks great but something seems to be up with the tests.

lakhotiaharshit:
> Looks great but something seems to be up with the tests.

Individually all tests are running fine however when we run together all the test after the test below are failing. I do not know the reason at the moment.

> @pytest.mark.parametrize("val_volt", ['10, 9.9e37, -9.9e37']) 
def test_get_timetrace(driver_volt):
    driver_volt.timetrace_npts(3)
    voltage = driver_volt.timetrace.get() 




    assert (voltage == np.array([10.0, np.inf, -np.inf])).all()

qcodes/tests/drivers/test_keysight_34465a.py
```diff
@@ -11,6 +13,28 @@ def driver():
...
+        yield keysight_sim
+    finally:
+        keysight_sim.close()
+
+
+@pytest.fixture(scope='function')
+def driver_volt(val_volt):
+    keysight_sim = Keysight_34465A('keysight_34465A_sim',
+                                   address='GPIB::1::INSTR',
+                                   visalib=visalib)
```
GateBuilder: keysight_sim is already defined, this should not be here I guess. Probably will fix the problem. @lakhotiaharshit

lakhotiaharshit: Did not helped. But thanks to your suggestion, I removed some code duplication.

lakhotiaharshit:
Following Mikhail's nice suggestion. I have ignored that particular test at the moment. The is marked not to run and a small reason is added. I will open a new card in qcodes debt so that at later point we can fix that test.

PR #1934, opened 2020-03-11
ZI add support for DAQ module

jenshnielsen:
This replaces #978 for reference


jenshnielsen:
This has been deprecated and replaced by the drivers in https://github.com/zhinst/zhinst-qcodes/

PR #1940, opened 2020-03-16
Use pytest retry to reexecute known flaky test 

jenshnielsen:

### DIFF
--- a/qcodes/tests/dataset/test_measurement_context_manager.py
+++ b/qcodes/tests/dataset/test_measurement_context_manager.py
@@ -707,6 +707,7 @@ def test_subscribers_called_at_exiting_context_if_queue_is_not_empty(experiment,
     assert collected_x_vals == given_x_vals
 
 
+@pytest.mark.flaky(reruns=5)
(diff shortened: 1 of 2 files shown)


jenshnielsen:
@astafan8 This uses the decorator to mark that test as explicitly known flaky. We could also do this in a different way and pass this as a cmd option for all tests

astafan8:
> This uses the decorator to mark that test as explicitly known flaky.

for that particular test i do prefer the decorator approach. Shall we mark all the subscriber-related tests as flaky? And also all the tests that generate tooooo much data with hypothesis (i think those are mostly for parameters)?

> We could also do this in a different way and pass this as a cmd option for all tests

My opinion is that I wouldn't want to rerun failed tests that failed NOT due to flaky-ness. But perhaps the overhead of rerunning failed tests is not that big, and it's just more convenient to use the cmd option, i don't know. Let's perhaps see how it goes for now and make changes later?

jenshnielsen:
@astafan8 I suggest merging this now and seeing which other tests are flaky before marking them as such

PR #1941, opened 2020-03-16
Deprecate get_data and get_values methods in dataset

GateBuilder:
This PR deprecates the methods `get_data` and `get_values` from dataset in favor of `get_parameter_data`.

@Dominik-Vogel


qcodes/dataset/data_set.py
```diff
@@ -857,6 +857,8 @@ def _validate_parameters(*params: Union[str, ParamSpec, _BaseParameter]
                 valid_param_names.append(maybeParam)
         return valid_param_names
 
+    @deprecate('This method has been deprecated.', 'Use `get_parameter_data`'
```
lakhotiaharshit: I think the reason part "This method has been deprecated" is redundant. The output to user will be like " The function get_data is deprecated, because **This method has been deprecated**." We can only go ahead with giving alternative.

qcodes/dataset/sqlite/queries.py
```diff
@@ -99,6 +99,8 @@ def _build_data_query(table_name: str,
     return query
```
lakhotiaharshit: Since we are putting deprecation warnings qcodes.dataset.data_set. The same warning in queries is not needed.

GateBuilder: Well I am not sure, I just deprecated when I see one. Having two warnings will not be a problem, if it tails, IMHO.

qcodes/dataset/data_set.py
GateBuilder: Good catch let me rephrase.

astafan8:
> do you know of a fast way of looking into docs for the usage? Maybe I can use grep..Not sure.

@GateBuilder just use "find everywhere" feature of your editor, both pycharm and vs code have it. grep is also OK, but i prefer pycharm's tool for convenience.

since we are deprecating these, we want to make sure that the docs don't mention/recommend its usage (also, as per our discussion offline).

PR #1943, opened 2020-03-16
[DRAFT]: Add binder for notebooks

lakhotiaharshit:
Add binder (mybinder.org) for the qcodes notebooks so that they can easily executed. Helps users to quickly interact with new features, APIs, and tutorials. 

DRAFT because I am still playing around with the binder to look for the best settings

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -441,3 +441,15 @@ numfig = True
 nbsphinx_kernel_name = 'python3'
 # always execute notebooks.
 nbsphinx_execute = 'always'
+
+# Binder configuration (WRITE MORE DETAILS ONCE FIXED)
+sphinx_gallery_conf = {
+  'binder': {
+     'org': 'qcodes',
+     'repo': 'https://github.com/qcodes/qcodes',
+     'branch': 'master',
+     'binderhub_url': 'https://mybinder.org',
+     'notebooks_dir': 'notebooks',
+     'use_jupyter_lab': True,
+     }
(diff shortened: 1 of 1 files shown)


jenshnielsen:
I don't know the best way to do this it may very whell be sphinx gallery but I would probably have tried something like https://github.com/spatialaudio/nbsphinx/blob/master/doc/conf.py#L39 which should render like the header in this notebook https://nbsphinx.readthedocs.io/en/0.5.1/configuring-kernels.html

docs/conf.py
```diff
@@ -71,6 +71,52 @@
...
 # source_encoding = 'utf-8-sig'
 
+# Add link to Binder in Prolog (WRITE MORE DETAILS ONCE FIXED)
+# -- Get version information  ----------------------------
+try:
+    from subprocess import check_output
+    release = qcodes.__version__
+
+    if 'rc' in release:
+        release = release.split('+')[0]
```
jenshnielsen: This seems very specific to rcs. What about alphas or betas? I think it is better to use something like packageing https://packaging.pypa.io/en/latest/version/ to extract the part of the version that we want

requirements.txt
```diff
@@ -17,4 +17,5 @@ websockets==8.1
 tabulate
 applicationinsights
 requirements-parser
+packaging
```
jenshnielsen: Should this be in docs_reqirements.txt?

PR #1944, opened 2020-03-17
code simplifications

basnijholt:
Just some small changes that I had to make while browsing the code.


qcodes/dataset/experiment_container.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, path_to_db: Optional[str]=None,
         max_id = len(get_experiments(self.conn))
 
         if exp_id is not None:
-            if exp_id not in range(1, max_id+1):
+            if not isinstance(exp_id, int) or not (1 <= exp_id <= max_id):
```
jenshnielsen: This does not relly seem like a simlification at this stage

basnijholt: It's an O(1) instead of O(N) operation and it's more explicit what's being tested (I think).

I could revert it if you still want.

jenshnielsen: I don't think this is O(N) as range is a generator in python3 and no longer a list. 

```
In [15]: max_id = 100_000_000

In [16]: exp_id = np.random.randint(max_id)

In [17]: exp_id
Out[17]: 93130756

In [22]: %timeit isinstance(exp_id, int) and (1 <= exp_id <= max_id)
97.6 ns ± 0.313 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [24]: %timeit exp_id in range(100_000_000)
250 ns ± 8.78 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [25]: %timeit exp_id in list(range(100_000_000))
2.98 s ± 9.85 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

In any case the performance difference is not important in real life senarious.

basnijholt: Fair enough, the speed up definitely not worth it! I will change it back.

PR #1949, opened 2020-03-19
Add note that measurements do not nest

GateBuilder:
We add a note to the relevant notebook to clear out that measurement context manager do not nest.
@QCoDeS/core


docs/examples/DataSet/Performing-measurements-using-qcodes-parameters-and-dataset.ipynb
```diff
@@ -14,7 +14,9 @@
     "This notebook shows some ways of performing different measurements using \n",
     "QCoDeS parameters and the [DataSet](DataSet-class-walkthrough.ipynb) via a powerful ``Measurement`` context manager. Here, it is assumed that the reader has some degree of familiarity with fundamental objects and methods of QCoDeS.\n",
     "\n",
-    "Let us start with necessary imports:"
+    "Users should not confuse the ``Measurement`` context manager of QCoDeS with that of conventional Python. In particular, here in the context of experiments, one cannot nest multiple managers to perform related or independent measurements. This is a particular consequence of how data is stored and managed.\n",
```
jenshnielsen: I would probably reformulate this a bit. There are plenty of other context managers in python that it makes no sense to nest

PR #1951, opened 2020-03-19
Throw warning if nested measurements are used

lakhotiaharshit:
This makes nested loops runtime error. 

To do:
 - [x] Add tests


qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -245,6 +245,24 @@ def test_log_includes_extra_info(caplog, meas_with_registered_param):
 
     assert "some extra info" in caplog.text
 
+def test_nested_measurement_throws_error(experiment, DAC, DMM):
```
astafan8: remove unused fixtures :)
```suggestion
def test_nested_measurement_throws_error(experiment):
```

qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -245,6 +245,24 @@ def test_log_includes_extra_info(caplog, meas_with_registered_param):
...
+    # pytest.raises(Exception): does not work because it does not allow
+    # the state of _is_entered to be changed to False when context manager
+    # ends. Hence all the test after this one fails.
+    try:
+        with meas1.run():
+            with meas2.run():
+                pass
+            pass
+    except RuntimeError:
+        return True
```
astafan8: in order to mimic the behavior of `pytest.raises` we'd need more, smth like this:
```suggestion
    exception_raised = False

    try:
        with meas1.run():
            with meas2.run():
                pass
            pass
    except RuntimeError as e:
        if str(e) == 'Nested measurements are not supported':
            exception_raised = True
        else:
            raise e

    assert exception_raised, "Expected exception about nesting measurements wasn't raised"
```

lakhotiaharshit: Tests are still failing. So pytest.raises(Excpetion) is not the only reason, although not using it fixes the few tests. I do not completely understand why things are failing so I will look into this again afterwards.

lakhotiaharshit:
@astafan8 @jenshnielsen This is ready for second review. Thanks !

qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -428,6 +428,24 @@ def test_mixing_array_and_numeric(DAC, bg_writing):
                              (DAC.ch2, np.array([DAC.ch2(), DAC.ch1()])))
 
 
+@pytest.mark.usefixtures("experiment")
+def test_nested_measurement_throws_error(DAC, caplog):
```
astafan8, line 432:
```suggestion
def test_nested_measurement_raises_warning(DAC, caplog):
```

qcodes/tests/dataset/test_measurement_context_manager.py
```diff
@@ -428,6 +428,24 @@ def test_mixing_array_and_numeric(DAC, bg_writing):
...
+
+    meas2 = Measurement()
+    meas2.register_parameter(DAC.ch1, paramtype='numeric')
+    meas2.register_parameter(DAC.ch2, paramtype='array')
+
+    with meas1.run():
+        with meas2.run():
+            pass
+        pass
+    assert "Nested measurements are not supported. This will become an " \
```
astafan8, line 445: perhaps `with pytest.warns(..` ?

PR #1953, opened 2020-03-20
Add note on limitations in delegate parameter

jenshnielsen:


qcodes/instrument/parameter.py
```diff
@@ -1235,9 +1235,16 @@ class DelegateParameter(Parameter):
     base class without overwriting properties of the source: for example to
     set a different scaling factor and unit on the :class:`.DelegateParameter`
     without changing those in the source parameter
+
+    Note:
+        DelegateParameter only supports mappings between the
+        :class:`.DelegateParameter` and:class:`.Parameter` that are invertible
```
astafan8:
```suggestion
        :class:`.DelegateParameter` and :class:`.Parameter` that are invertible
```

PR #1956, opened 2020-03-23
Better attribute documentation for parameter

jenshnielsen:
Inline the documentation of attributes (and make them actual documental attributes by taking them out of the init) Name is not an attribute but a property (and already documented as such)

### DIFF
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -411,20 +411,6 @@ class Instrument(InstrumentBase, AbstractInstrument):
         metadata: additional static metadata to add to this
             instrument's JSON snapshot.
 
-
-    Attributes:
-        name (str): an identifier for this instrument, particularly for
-            attaching it to a Station.
-
-        parameters (Dict[Parameter]): All the parameters supported by this
-            instrument. Usually populated via ``add_parameter``
-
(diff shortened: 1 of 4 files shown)


qcodes/instrument/base.py
```diff
@@ -30,30 +30,32 @@ class InstrumentBase(Metadatable, DelegateAttributes):
...
-
-        parameters (Dict[Parameter]): All the parameters supported by this
-            instrument. Usually populated via ``add_parameter``.
-
-        functions (Dict[Function]): All the functions supported by this
-            instrument. Usually populated via ``add_function``.
-        submodules (Dict[Metadatable]): All the submodules of this instrument
-            such as channel lists or logical groupings of parameters.
-            Usually populated via ``add_submodule``.
+    functions: Dict[str, Function] = {}
```
astafan8: i wouldn't want `functions` and `submodules` to be perceived as class attributes as opposed to instance attributes. So i'd prefer that they are documented the same ways as `parameters` - in the `__init__`. Alternatively, perhaps, it's possible to declare the attribute at the class level (with typing and docstring) but not give it any value?

jenshnielsen: Yes, this pr should really have been draft. It's meant to show different options. They should all be done the same way

PR #1957, opened 2020-03-23
Keithley s46 Correct paramter that incorrectly overwrites get and set

jenshnielsen:
Owerwriting get and set in this way will prevent the caching and snapshot of the parameter
from working correctly and should never be done in qcodes


jenshnielsen:
@sohailc Can you please have a look at this change as this looks like your driver

jenshnielsen:
@sohailc Any comments?

PR #1961, opened 2020-03-24
Better snapshot of station.config (and other UserDicts)

jenshnielsen:
This was falling back to `str() repr` which is not great for reloading. This fixes that by serializing the data member of the userdict which is a regural dict

### DIFF
--- a/qcodes/tests/test_json.py
+++ b/qcodes/tests/test_json.py
@@ -1,8 +1,12 @@
 from unittest import TestCase
 import numpy as np
 import json
+from collections import UserDict
 
 from qcodes.utils.helpers import NumpyJSONEncoder
+
+class SomeUserDict(UserDict):
+    pass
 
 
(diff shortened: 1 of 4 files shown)


qcodes/tests/test_json.py
```diff
@@ -42,7 +48,8 @@ def test_numpy_good(self):
             'length': 45.23,
             'points': [12, 24, 48],
             'RapunzelNumber': {'__dtype__': 'complex', 're': 4.89, 'im': 0.11},
-            'verisimilitude': {'__dtype__': 'complex', 're': 0, 'im': 1}
+            'verisimilitude': {'__dtype__': 'complex', 're': 0, 'im': 1},
+            'myuserdict': {'a': 1}
```
jenshnielsen: Prior to this fix that would have been a string as in  `'myuserdict': "{'a': 1}"`

PR #1967, opened 2020-03-30
Interactive widget to explore datasets in Jupyter Notebooks

basnijholt:
With this widget, one can view qcodes experiments interactively.


qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+            row["Variables"] = expandable_dict(variables, tab, ds)
+            row["MSMT Time"] = label(ds.completed_timestamp() or "")
+            rows.append(row)
+
+    grid = GridspecLayout(n_rows=len(rows), n_columns=len(header_names))
+
+    empty_label = label("")
+    for i, row in enumerate(rows):
+        for j, name in enumerate(header_names):
+            grid[i, j] = row.get(name, empty_label)
```
astafan8, line 411: i tried to use the widget on a large DB with ~300 datasets and i noticed that the order in which datasets appeared was not following the run id, the left most column. So it got me thinking that we'd actually want some kind of sorting. We could leave this for later (because, for example, we could swtich to using SQL queries to populate the widget which might help with the sorting it, where efficient, on sql side).

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+    for exp in qcodes.experiments():
+        tooltip = f"{exp.name}#{exp.sample_name}@{exp.path_to_db}"
+
+        for ds in exp.data_sets():
+            coords, variables = _get_coords_and_vars(ds)
+            row = {}
+            row["Run ID"] = label(str(ds.run_id))
+            row["Name"] = label(ds.name)
+            row["Experiment"] = button(
+                f"#{exp.exp_id}", "warning", tooltip=tooltip
```
astafan8, line 398: i have a general comment about button styles - they are great, but sometimes not consistent. i think it would help users if the colors were indicating "what can be done with this 'button'" - nothing vs it-will-expand vs something-will-close vs ... :) for example - experiment id is orange but nothing happens when you click it, "open snapshot" is orange but an actual new tab gets opened when you click it, "back" button is orange but when you click it the section collapses, etc etc.

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+    header = {n: button(n, "info") for n in header_names}
+    rows = [header]
+    for exp in qcodes.experiments():
+        tooltip = f"{exp.name}#{exp.sample_name}@{exp.path_to_db}"
+
+        for ds in exp.data_sets():
+            coords, variables = _get_coords_and_vars(ds)
+            row = {}
+            row["Run ID"] = label(str(ds.run_id))
+            row["Name"] = label(ds.name)
```
astafan8, line 396: sometimes these names are long, and the widget truncates them with "..." at the end - do ipywidgets have a different way for displaying such that the full name is still possible to see? perhaps a hover reveals full name, or it's wraped or else?

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+
+def _plot_ds(ds: DataSet) -> None:
+    try:
+        # `get_data_by_id` might fail
+        nplots = len(get_data_by_id(ds.run_id))  # TODO: might be a better way
+        nrows = math.ceil(nplots / 2) if nplots != 1 else 1
+        ncols = 2 if nplots != 1 else 1
+        fig, axes = plt.subplots(nrows, ncols, figsize=(6 * ncols, 4 * nrows))
+        # `plot_dataset` might also fail.
+        plot_dataset(ds, axes=axes.flatten())
```
astafan8, line 173: `plot_dataset` has `**kwargs` which are passed internally to `plt.subplots`, so i suggest to:
* leave the rows/cols logic to `plot_dataset`, OR apply it perhaps after the `plot_dataset` returns the axes if really needed?
* pass the figure created here to `plot_dataset` explicitly - when i played with the widget i somehow managed to screw the place where the plots should appear, so they started showing up not in the tab but elsewhere. So i guess the remedy is to create a new explicit figure her eand pass it to `plot_dataset` (if that somehow doesn't work, let's improve `plot_dataset` to fix that)

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+        fig.tight_layout()
+        plt.show(fig)
+    except Exception as e:
+        print(e)  # TODO: print complete traceback
```
astafan8, line 177: this already happens in `_do_in_tab`->`on_click` shall we not duplicate the error handling and leave it to the "higher level", that is to the tab?

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+                if which == "plot":
+                    _plot_ds(ds)
+                elif which == "snapshot":
+                    snapshot = ds.snapshot
+                    if snapshot is not None:
+                        display(nested_dict_browser(snapshot))
+                    else:
+                        print("This dataset has no snapshot")
+            except Exception as e:
+                print(e)  # TODO: print complete traceback
```
astafan8, line 231: yes, this would be very welcome :) `traceback.print_exc` perhaps?

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+
+def create_tab(do_display: bool = True) -> Tab:
+    """Creates a `ipywidgets.Tab` which can display outputs in its tabs."""
+    tab = Tab(children=(Output(),))
+
+    tab.set_title(0, "Info")
+    if do_display:
+        display(tab)
+
+    with tab.children[-1]:
```
astafan8, line 245: this is a bit hacky, it'd be cleaner to have the instance of output and do `with output` as opposed to `with <last child of tab>`

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+
+    # Header
+    title = " ► ".join(nested_keys)
+    grid[0, :-1] = button(title, "success")
+    up_click = update(nested_keys[:-1])
+    grid[0, -1] = button("↰", "info", up_click)
+
+    # Body
+
+    i = 1
```
astafan8, line 125: what is "i"? :) i'm basically just asking for more explicit names than I, k, v, k_, v_

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+    nested_keys: Sequence[str], table: Dict[Any, Any], box: Box
+) -> Callable[[Button], None]:
+    def update_box(_: Button) -> None:
+        box.children = (_nested_dict_browser(nested_keys, table, box),)
+
+    return update_box
+
+
+def _nested_dict_browser(
+    nested_keys: Sequence[str],
```
astafan8, line 84: would be great to have a small docstring for this method. especially, what are the nested keys?

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+    table: Dict[Any, Any],
+    box: Box,
+    max_nrows: int = 30,
+) -> GridspecLayout:
+    """Generates a `GridspecLayout` of the ``nested_keys`` in ``table`` which is
```
astafan8, line 89: i guess `table` is the same as `dict` in `_nested_dict_browser`? some uniformity with naming would help :)

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+def _update_nested_dict_browser(
+    nested_keys: Sequence[str], table: Dict[Any, Any], box: Box
+) -> Callable[[Button], None]:
+    def update_box(_: Button) -> None:
+        box.children = (_nested_dict_browser(nested_keys, table, box),)
+
+    return update_box
+
+
+def _nested_dict_browser(
```
astafan8, line 83: this is probably something for later, but it would help the widget to be more expressive if some of the "dict"s would implement more convenient ways to display themselves, for example, some qcodes parameters - i tried initiate this long time ago here https://github.com/QCoDeS/Qcodes/pull/1215/files#diff-ce80bf817fb0af0f47ae2069c8954f98R63 :)

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+        "Coordinates",
+        "Variables",
+        "MSMT Time",
+        "Notes",
+    ]
+
+    header = {n: button(n, "info") for n in header_names}
+    rows = [header]
+    for exp in qcodes.experiments():
+        tooltip = f"{exp.name}#{exp.sample_name}@{exp.path_to_db}"
```
astafan8, line 390: this widget should absolutely promote use of DataSet.sample_name, DataSet.exp_name and DataSet.name. it does a great job with the latter one but not with the former two. So let's have it :) unfortunately i don't see any other options for this rather than having three columns, and getting rid of the "experiment id" column (the id itself could become part of a ~~tooltip~~ expandable button or smth).

astafan8, line 390: wrt `path_to_db` - depending on the usage of the widget, it might iether make sense to put to the top of the table, or, in case we allow listing arbitrary datasets, path_to_db could be part of that "expandable identity info" that i mentioned in a few other comments.

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+    if do_display:
+        display(tab)
+
+    with tab.children[-1]:
+        # Prints it in the Output inside the tab.
+        print("Plots and snapshots will show up here!")
+    return tab
+
+
+def editable_metadata(ds: DataSet) -> Box:
```
astafan8, line 251: note that due to the value the `DataSet.metadata` is implemented, changes to the notes made in the widget would be persistet but won't be visible to any other isntances of the same dataset elsewhere unless that dataset gets reloaded. This is just a note, it's unrelated to the widget but related to the `DataSet.metadata` feature which qcodes will improve at some point.

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+
+    header = {n: button(n, "info") for n in header_names}
+    rows = [header]
+    for exp in qcodes.experiments():
+        tooltip = f"{exp.name}#{exp.sample_name}@{exp.path_to_db}"
+
+        for ds in exp.data_sets():
+            coords, variables = _get_coords_and_vars(ds)
+            row = {}
+            row["Run ID"] = label(str(ds.run_id))
```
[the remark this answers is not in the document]
basnijholt: Fixed in c86e53a14.

qcodes/interactive_widget.py
basnijholt: Fixed in 23b6b3523. I renamed everything to `nested_dict` because it's more obvious what that is.

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+                disabled=False,
+                layout=Layout(height="auto", width="auto"),
+            )
+            save_button = button(
+                "",
+                "danger",
+                on_click=_save_button(box, ds),
+                button_kwargs=dict(icon="save"),
+            )
+            box.children = (text_input, save_button)
```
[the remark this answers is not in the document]
basnijholt: close-button mov
Fixed in 56e2764887529299414664ccb08e4c69b64cc0fa.

qcodes/interactive_widget.py
basnijholt: Fixed in 172dcf3c08b91afbad8d06b9b7e259535f5516ff.
run-id-exp-button mov

basnijholt: I've added it to the Experiment button in 72e07e1a2271aa2a904569138af4aa02bf7bb34d.

qcodes/interactive_widget.py
basnijholt: I've made all expandable buttons green in a8562a47316420dda54fc5c4ef2bcb15d951d5c9.

image

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+            row["Name"] = label(ds.name)
+            row["Experiment"] = button(
+                f"#{exp.exp_id}", "warning", tooltip=tooltip
+            )
+            row["Notes"] = editable_metadata(ds)
+            row["Coordinates"] = expandable_dict(coords, tab, ds)
+            row["Variables"] = expandable_dict(variables, tab, ds)
+            row["MSMT Time"] = label(ds.completed_timestamp() or "")
+            rows.append(row)
+
```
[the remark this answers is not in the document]
basnijholt: Fixed.

qcodes/interactive_widget.py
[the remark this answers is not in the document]
basnijholt: run-id-exp-button mov
Guid is copyable now.

qcodes/interactive_widget.py
basnijholt: Right, I agree this is a good idea but out of the scope of this PR (I think). Therefore, I will resolve this conversation now. Reopen if you do want to revisit it.

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+    for i, row in enumerate(rows):
+        for j, name in enumerate(header_names):
+            grid[i, j] = row.get(name, empty_label)
+
+    grid.layout.grid_template_rows = "auto " * len(rows)
+    grid.layout.grid_template_columns = "auto " * len(header_names)
+    return grid
+
+
+def experiments_widget(db: Optional[str] = None) -> VBox:
```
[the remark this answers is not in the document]
basnijholt: This makes sense!
I've implemented it in b150574211da502c44b2e5cd6a117dfff18724e7.
image

qcodes/interactive_widget.py
basnijholt: OK, makes sense. I don't think I could do anything about it really. There is an explanation in the `experiments_widget`'s doc-string.

qcodes/interactive_widget.py
basnijholt: In 0757dd7ac I have renamed `i` -> `row_index`.

About the other variables, `k` and `v` are keys and values from an arbitrary dict and I don't think I can do much better. I think `(k, v)` are a commonly used notation for `(key, value)` pairs. However, if you want I can `k -> key` and `v -> value`?

qcodes/interactive_widget.py
basnijholt: I improved it in b91b7ece1.

qcodes/interactive_widget.py
astafan8: > I have renamed i -> row_index

great!

key/value would be better but it's fine as is :)

qcodes/interactive_widget.py
basnijholt: Fixed in d7ae9906a.

qcodes/interactive_widget.py
basnijholt: Fixed in d7ae9906a.

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+            row = {}
+            row["Run ID"] = label(str(ds.run_id))
+            row["Name"] = label(ds.name)
+            row["Experiment"] = button(
+                f"#{exp.exp_id}", "warning", tooltip=tooltip
+            )
+            row["Notes"] = editable_metadata(ds)
+            row["Coordinates"] = expandable_dict(coords, tab, ds)
+            row["Variables"] = expandable_dict(variables, tab, ds)
+            row["MSMT Time"] = label(ds.completed_timestamp() or "")
```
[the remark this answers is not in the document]
basnijholt: I have changed it to `run_timestamp` because it always exists, whereas `completed_timestamp` might now exist.
I have also made it into a button that upon clicking reveals:
image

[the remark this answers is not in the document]
basnijholt: Fixed in f4c7d5dcf.

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+            coords, variables = _get_coords_and_vars(ds)
+            row = {}
+            row["Run ID"] = label(str(ds.run_id))
+            row["Name"] = label(ds.name)
+            row["Experiment"] = button(
+                f"#{exp.exp_id}", "warning", tooltip=tooltip
+            )
+            row["Notes"] = editable_metadata(ds)
+            row["Coordinates"] = expandable_dict(coords, tab, ds)
+            row["Variables"] = expandable_dict(variables, tab, ds)
```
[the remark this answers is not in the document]
basnijholt: Fixed in b6d1f6333 (and previous commits).

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+                layout=Layout(height="300px", width="auto"),
+            )
+            plot_button = button(
+                "Plot",
+                "warning",
+                on_click=_do_in_tab(tab, ds, "plot"),
+                button_kwargs=dict(icon="line-chart"),
+            )
+            snapshot_button = button(
+                "Open snapshot",
```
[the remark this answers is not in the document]
basnijholt: I fixed this in 9c034d50f.

As discussed on Teams snapshot and plot have its own column now:
image

qcodes/interactive_widget.py
```diff
@@ -0,0 +1,436 @@
...
+
+    box = VBox([], layout=Layout(height="auto", width="auto"))
+    box.children = (_changeable_button(dct, box),)
+    return box
+
+
+def _get_coords_and_vars(ds: DataSet) -> Tuple[Dict[str, Any], Dict[str, Any]]:
+    coordinates = {}
+    variables = {}
+    for p, spec in ds.paramspecs.items():
```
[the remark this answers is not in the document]
basnijholt: In dc6311f8e I've changed the function to use `ds.description.interdeps`.

Do you have any other suggestions?

qcodes/interactive_widget.py
basnijholt: In 1c162386d, I sort Data_Sets on run_id when data_sets isn't explicitly passed.

qcodes/interactive_widget.py
basnijholt: In ef12dcbdc, I use `with output` instead of `with tab.children[-1]`.

qcodes/interactive_widget.py
basnijholt: In f8aad59d4, I add DataSet.name to the expanded Experiment button text area for the case where the Name col is too long.

qcodes/interactive_widget.py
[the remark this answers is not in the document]
astafan8: not specifically. the use of interdeps is good enough for the start.

qcodes/interactive_widget.py
astafan8: ok, good.

i know from plottr/inspectr experiences that users like the descending sorting by timestamp (newest first). perhaps we should add an option for that? or leave to some later PR?

qcodes/interactive_widget.py
basnijholt: In 49e4147e5, I let plot_dataset figure out the fig sizes and axes.

qcodes/interactive_widget.py
basnijholt: In c1f503ebc, I implement sorting by timestamp, run_id, or None:
image

astafan8 (approved):

PR #1969, opened 2020-04-07
add types to init functions of concrete keysight drivers

jenshnielsen:
These simply match the types in the base class. So it seems like a good idea to add them to ensure that the end user can get type checking

### DIFF
--- a/qcodes/instrument_drivers/Keysight/Keysight_34465A_submodules.py
+++ b/qcodes/instrument_drivers/Keysight/Keysight_34465A_submodules.py
@@ -1,3 +1,4 @@
+from typing import Any
 from .private.Keysight_344xxA_submodules import _Keysight_344xxA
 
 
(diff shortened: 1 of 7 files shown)


qcodes/instrument_drivers/Keysight/Keysight_34465A_submodules.py
```diff
@@ -1,10 +1,11 @@
+from typing import Any
 from .private.Keysight_344xxA_submodules import _Keysight_344xxA
 
 
 class Keysight_34465A(_Keysight_344xxA):
```
astafan8, line 5: shall _Keysight_344xxA also be typed?

jenshnielsen: Sure. It's only missing a few Any on kwargs and a single None return

jenshnielsen:
Added disallow_untyped_defs  for both the subclasses and the base class

PR #1970, opened 2020-04-10
Add new methods to keithley 2450 driver

liangosc:
(still need to add code examples to the notebook)

The following features are added:
Sense:
	(new param) auto_zero
	(new param) auto_zero_once
	(new param) count
	(new method) fetch
	add buffer name and extra elements for measurements (both single measurement and sweep)
	(new method) make and delete buffer
	(new method) number of readings in the buffer
	(new method) choose which buffer to read from and the data fields
	(new method) reset and get buffer size
	(new method) trigger trace

Source:
	(new param) read_back
	added more inputs var to the sweep arguments
	modified the sweep_start

Main class:
	(new param) power_line_frequency

@sohailc


sohailc:
@liangosc Good work!

docs/examples/driver_examples/Qcodes example with keithley 2450.ipynb
```diff
@@ -26,20 +28,496 @@
...
+   ],
+   "source": [
+    "print(f\"By default, {keithley.sense.count()} measurement will be made.\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This can be changed by the \"count()\" method:"
```
astafan8, line 297: :)
```suggestion
    "This can be changed by the \"count()\" parameter:"
```

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -108,27 +108,101 @@ def __init__(self, parent: 'Keithley2450', name: str, proper_function: str) -> N
...
+    ) -> str:
+        """
+        This method requests the latest reading from a reading buffer.
+
+        Args:
+            buffer_name: buffer name, default is 'defbuffer1'
+            buffer_elements: available options are "DATE", "FORMatted",
+                "FRACtional", "READing", "RELative", "SEConds",
+                "SOURce", "SOURFORMatted", "SOURSTATus",
+                "SOURUNIT", "STATus","TIME", "TSTamp", "UNIT"
```
astafan8, line 149: it'd be great if the driver would check the validity of the "fields"/"elements" that the user is entering here because a typo will result in an inconvenient visa io error, i guess? it'd be better if user was getting an exception with a clear message if he/she mistypes the field/element name.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -108,27 +108,101 @@ def __init__(self, parent: 'Keithley2450', name: str, proper_function: str) -> N
...
+        delete an user defined data buffer
+        """
+        self.write(f":TRACe:DELete '{buffer_name}'")
+
+    def clear_trace(self, buffer_name: str = "defbuffer1") -> None:
         """
         Clear the data buffer
         """
-        self.write(":TRACe:CLEar")
+        self.write(f":TRACe:CLEar '{buffer_name}'")
```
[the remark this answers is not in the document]
liangosc: If user calls it without buffer name, the system will assume the user is using defbuffer1, same as the default value of the clear_trace() function. I fail to see how this will affect existing code.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -108,27 +108,101 @@ def __init__(self, parent: 'Keithley2450', name: str, proper_function: str) -> N
...
+            get_cmd=f":SENSe:{self._proper_function}:AZERo?",
+            set_cmd=f":SENSe:{self._proper_function}:AZERo {{}}",
+            val_mapping=create_on_off_val_mapping(on_val="1", off_val="0"),
+            docstring="This command enables or disables automatic updates to"
+                      "the internal reference measurements (autozero) of the"
+                      "instrument."
+        )
+
+        self.add_parameter(
+            'auto_zero_once',
```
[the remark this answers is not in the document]
liangosc: A new method is defined, removed the parameter.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -108,27 +108,101 @@ def __init__(self, parent: 'Keithley2450', name: str, proper_function: str) -> N
             vals=Numbers(0, 1e4)
         )
 
-    def _measure(self) -> str:
+        self.add_parameter(
+            'auto_zero',
```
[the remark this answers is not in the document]
liangosc: changed.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -367,6 +499,14 @@ def __init__(self, name: str, address: str, **kwargs) -> None:
             val_mapping=create_on_off_val_mapping(on_val="1", off_val="0")
         )
 
+        self.add_parameter(
+            "power_line_frequency",
```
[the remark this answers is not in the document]
liangosc: Good point. `line_frequency` it is.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -140,6 +214,47 @@ def _set_user_delay(self, value) -> None:
...
+
+        Args:
+            buffer_name: the default buffers (defbuffer1 or defbuffer2) or the
+                name of a user-defined buffer.
+        Returns:
+             number of readings in the specified reading buffer.
+        """
+        return self.ask(f":TRACe:ACTual? '{buffer_name}'")
+
+    def set_buffer_and_fields_to_read(
```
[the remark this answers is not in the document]
liangosc: read_settings?

docs/examples/driver_examples/Qcodes example with keithley 2450.ipynb
```diff
@@ -6,6 +6,8 @@
    "source": [
     "# QCoDeS Example with Tektronix Keithley 2450 Source Meter\n",
     "\n",
+    "Update: new methods related to the user-defined reading buffer have been added to the driver. New examples have been added to previous code example.\n",
```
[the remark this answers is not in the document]
liangosc: you mean to put the new examples after the existing examples? That makes sense.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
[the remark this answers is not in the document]
astafan8: thanks for the explanation. that's indeed shouldn't be a problem now

docs/examples/driver_examples/Qcodes example with keithley 2450.ipynb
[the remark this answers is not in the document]
astafan8: > you mean to put the new examples after the existing examples?

yes :) especially because "new examples" explain a more advanced functionality than the "older examples"

qcodes/instrument_drivers/tektronix/Keithley_2450.py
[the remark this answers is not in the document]
liangosc: A new approach is used, and this function is removed. (similar one is added to the new Buffer class, with different name).

qcodes/instrument_drivers/tektronix/Keithley_2450.py
liangosc: Good point. Now the driver is checking if the user input is correct or not, with feedback.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -83,8 +217,7 @@ def __init__(self, parent: 'Keithley2450', name: str, proper_function: str) -> N
             label=self._proper_function,
             get_cmd=self._measure_sweep,
             unit=unit,
-            vals=Arrays(shape=(self.parent.npts,)),
-            parameter_class=ParameterWithSetpoints
```
liangosc, line 87: I should not remove this. Trying to fix it now....

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -459,6 +681,27 @@ def sense(self) -> Sense2450:
...
+        self._buffer_name = name
+
+    def get_buffer_name(self) -> str:
+        return self._buffer_name
+
+    def set_buffer_elements(self, elements: list) -> None:
+        self._buffer_elements = elements if len(elements) > 0 else None
+
+    def get_buffer_elements(self) -> Optional[list]:
+        return self._buffer_elements
```
astafan8, line 703: same for this parameter, it can be a `set_cmd=None, get_cmd=None` parameter as well with a validator for "list of values from an enumeration" where that enumeration will be list of keys of the "elements" dict.

Similar for the ""elements"" parameter of the buffer - the long validation logic can be taken care by a validator, and the set_cmd/get_cmd's only functionality would be to update the `parent.buffer_elements` parameter with the "elements" selected for that buffer.

docs/examples/driver_examples/Qcodes example with keithley 2450.ipynb
```diff
@@ -335,12 +340,361 @@
...
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "buffer_name = 'userbuff1'\n",
+    "buffer_size = 100\n",
+    "with keithley.buffer(buffer_name, buffer_size) as buff1:\n",
+    "    buff1.elements(['time', 'date', 'measurement', 'source_value_formatted'])\n",
+    "    keithley.source.sweep_setup(0, 1E-6, 10, buffer_name=buffer_name)\n",
```
astafan8, line 403: it is a bit weird that `sweep_setup` is not aware of the current buffer, so that `buffer_name` needs to be passed. is there a specific reason for that?

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -1,11 +1,156 @@
 import numpy as np
-from typing import cast, Dict, Union
+from typing import cast, Dict, Union, Optional
 
 from qcodes import VisaInstrument, InstrumentChannel, ParameterWithSetpoints
 from qcodes.utils.validators import Enum, Numbers, Arrays, Ints
 from qcodes.utils.helpers import create_on_off_val_mapping
 
 
+class NewParameterWithSetpoints(ParameterWithSetpoints):
```
astafan8, line 9: please use a more specific name :)

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -459,6 +681,27 @@ def sense(self) -> Sense2450:
...
+            style: str = ''
+    ) -> Buffer2450:
+        self._buffer_name = name
+        return Buffer2450(parent=self, name=name, size=size, style=style)
+
+    def set_buffer_name(self, name: str) -> None:
+        self._buffer_name = name
+
+    def get_buffer_name(self) -> str:
+        return self._buffer_name
```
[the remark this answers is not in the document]
liangosc: I was trying to store the buffer name in "self._buffer_name" but looks like it's redundant. Fixed, thanks.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -1,11 +1,156 @@
...
+        "measurement": "READing",
+        "relative_time": "RELative",
+        "seconds": "SEConds",
+        "source_value": "SOURce",
+        "source_value_formatted": "SOURFORMatted",
+        "source_value_status": "SOURSTATus",
+        "source_value_unit": "SOURUNIT",
+        "measurement_status": "STATus",
+        "time": "TIME",
+        "timestampe": "TSTamp",
```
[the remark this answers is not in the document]
liangosc: typo fixed.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
liangosc: I was trying to use `buffer_elements([])` to reset the variable `self._buffer_elements` to be None. But I guess `[]` will be fine. Fixed.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
liangosc: Totally agree. Second try: `ParameterWithSetpointsAllType`. Thoughts?

docs/examples/driver_examples/Qcodes example with keithley 2450.ipynb
liangosc: It it weird -- I totally agree with you here. There is no "current buffer" for the instrument, the user has to tell the system which buffer the reading is from every time. (by default it's always "defbuff01")

docs/examples/driver_examples/Qcodes example with keithley 2450.ipynb
liangosc: No there anymore

docs/examples/driver_examples/Qcodes example with keithley 2450.ipynb
```diff
@@ -18,19 +20,22 @@
     "import qcodes as qc \n",
     "from qcodes.instrument_drivers.tektronix.Keithley_2450 import Keithley2450\n",
     "from qcodes.dataset import initialise_database, Measurement, new_experiment\n",
-    "from qcodes.dataset.plotting import plot_by_id\n"
+    "from qcodes.dataset.plotting import plot_by_id\n",
```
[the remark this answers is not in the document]
liangosc: fixed.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -1,11 +1,171 @@
...
+        self.add_parameter(
+            "elements",
+            get_cmd=None,
+            set_cmd=None,
+            vals=Lists(Enum('date', 'measurement_formatted',
+                            'fractional_seconds', 'measurement',
+                            'relative_time', 'seconds', 'source_value',
+                            'source_value_formatted', 'source_value_status',
+                            'source_value_unit', 'measurement_status', 'time',
+                            'timestamp', 'measurement_unit')),
```
liangosc, line 97: Is there a way to map each element in the list? `mapping=buffer_elements ` won't work here.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -1,11 +1,171 @@
...
+        if (self.elements() is None) or (len(self.elements()) == 0):
+            return self.ask(f":FETCh? '{self.buffer_name}'")
+        fetch_elements = [
+            self.buffer_elements[element] for element in self.elements()
+        ]
+        return self.ask(
+            f":FETCh? '{self.buffer_name}', {','.join(fetch_elements)}"
+        )
+
+    def get_data(self, start_idx: int, end_idx: int) -> list:
```
liangosc, line 125: I put all the ":TRACE" related cmd under Buffer class. Maybe I should rewrite the `_measure_sweep()` with this `get_data()` method.

docs/examples/driver_examples/Qcodes example with keithley 2450.ipynb
astafan8: hm.... shall the example at least re-use the `buff` instance for it, for example `, buffer_name=buff1.short_name`? or perhaps even further, ``sweep_setup`` is capable of taking a buffer instance, like `sweep_setup(..., buffer=buff1)`?

qcodes/instrument_drivers/tektronix/Keithley_2450.py
astafan8: > map each element in the list
do you mean that you would like to have a `val_mapping` kind of behavior but allowing for "lists of these elements", right? if so, indeed, val_mapping wouldn't work for this but you could use `get_parser` and `set_parser` arguments to pass callables that implement the conversion, for example:
```python
def from_name_to_scpi(element_names: List[str]) -> List[str]:
    return [buffer_elements[element] for element in element_names]

def from_scpi_to_name(element_scpis: List[str]) -> List[str]:
    # assuming the inverted buffer_elements dict is already prepared :)
    return [inverted_buffer_elements_dict[element] for element in element_scpis]

# both of these could just be lambas, but it's up to you.

####################

# and then in add_parameter:
   ...,
   set_parser=from_name_to_scpi,
   get_parser=from_scpi_to_name,
   ...
    )
```

btw, in order not to repeat things, `buffer_elements.keys()` can simple be used: `vals=Lists(Enum(self.buffer_elements.keys()))`.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
astafan8: > Maybe I should rewrite the _measure_sweep() with this get_data() method.

that sounds like a good idea - removing duplication and properly using the buffer infra :)

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -603,6 +618,10 @@ def __init__(self, name: str, address: str, **kwargs) -> None:
                 Sense2450(self, "sense", proper_sense_function)
             )
 
+        self.buffer_name('defbuffer1')
+        self.add_submodule(
+            f"_buffer_{self.buffer_name()}",
+            Buffer2450(parent=self, name=self.buffer_name()))
```
astafan8, line 624: can't the `buffer` method be re-used here?

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -229,7 +244,7 @@ def __init__(self, parent: 'Keithley2450', name: str, proper_function: str) -> N
             get_cmd=self._measure_sweep,
             unit=unit,
             vals=Arrays(shape=(self.parent.npts,)),
-            parameter_class=NewParameterWithSetpoints
+            parameter_class=ParameterWithSetpointsAllType
```
astafan8, line 247: sorry, i forgot - what are the setpoints of this parameter-with-setpoints? is it up to the user to decide/define?

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -2,18 +2,19 @@
...
+from qcodes.utils.validators import Enum, Numbers, Arrays, Ints, Lists
 from qcodes.utils.helpers import create_on_off_val_mapping
 
 
-class NewParameterWithSetpoints(ParameterWithSetpoints):
+class ParameterWithSetpointsAllType(ParameterWithSetpoints):
     """
-    The newly added "_raw_value" will include extra fields for the data, in
-    addition to the numerical values, for the "sweep" parameter.
+    While the parent class ParameterWithSetpoints only support numerical data,
```
astafan8, line 11: > only support numerical data

this is not true. parameter with setpoints supports any kind of arrays, be it numeric, or string, or complex, etc. From how i understand the code, the docstring should say that `_raw_value` attribute is meant to be filled with some sort of "full data" while the parameters `get_raw`/`get_cmd` returns only a subset of that data, and it's up to the implementation of the `get_raw`/`get_cmd` to set `_raw_value` to this "full data" (i don't know how to write this better).

So, i think the `..AllType` suffix misses the point - and the point is to return on `.get()` the "important data" while on `get_selected` "more" data is accessible from that last `.get()` invocation. Could you try another iteration with the names? :)

also, it'd be great to have a better name for the `_raw_value`, raw value is something meaningful in the world of Parameters. Moreover, it's a bit unclear why get **selected** returns **raw** value, what does the **get** return then? :)

qcodes/instrument_drivers/tektronix/Keithley_2450.py
liangosc: `vals=Lists(Enum(self.buffer_elements.keys()))` I don't think this one would work, nor ` vals=Lists(Enum(list(self.buffer_elements.keys())))` -- `Enum` takes only individual variables, can't take a list of vars. There'll be a TypeError: TypeError: unhashable type: 'dict_keys', or
TypeError: unhashable type: 'list' for the latter.

liangosc: The `set_parser` and `get_parser` works like a charm~. Thanks!

qcodes/instrument_drivers/tektronix/Keithley_2450.py
liangosc: I added an extra argument to the `get_data` method, and rewrite the `_measure_sweep`. Now less code repetition, works the same.

docs/examples/driver_examples/Qcodes example with keithley 2450.ipynb
liangosc: Good point, I changed to `buffer_name=buff1.buffer_name`. Let's not push it further, because I want to provide the user the simpler way -- use the user-defined buffer name.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
liangosc: But in the class definition of `ParameterWithSetpoints`, there is:
>         if not isinstance(vals, Arrays):
>           raise ValueError(f"A ParameterWithSetpoints must have an Arrays "
>                             f"validator got {type(vals)}")
where `Arrays` is "numerical numpy arrays of numeric types"

As for the naming game we are playing... how about `_raw_value` -> `_user_selected_data`, and the new class name `ParameterWithSetpoints**Customized**`?

qcodes/instrument_drivers/tektronix/Keithley_2450.py
liangosc: The "setpoints" will be defined by the user, with start, end, step count, etc... when defining the sense and source function, there are:
> sense.sweep.setpoints = (self.source.sweep_axis,)

and

>self.sense.sweep.setpoints = (source.sweep_axis,)

where the sweep_aixs is:

>     def get_sweep_axis(self) -> np.ndarray:
>         if self._sweep_arguments == {}:
>             raise ValueError(
>                 "Please setup the sweep before getting values of this parameter"
>             )
> 
>         return np.linspace(
>             start=self._sweep_arguments["start"],
>             stop=self._sweep_arguments["stop"],
>             num=self._sweep_arguments["step_count"]
>         )

qcodes/instrument_drivers/tektronix/Keithley_2450.py
```diff
@@ -108,27 +294,59 @@ def __init__(self, parent: 'Keithley2450', name: str, proper_function: str) -> N
...
+            val_mapping=create_on_off_val_mapping(on_val="1", off_val="0"),
+            docstring="This command enables or disables automatic updates to"
+                      "the internal reference measurements (autozero) of the"
+                      "instrument."
+        )
+
+        self.add_parameter(
+            'count',
+            get_cmd=":SENSe:COUNt?",
+            set_cmd=":SENSe:COUNt {{}}",
```
astafan8, line 310: no need for double braces here
```suggestion
            set_cmd=":SENSe:COUNt {}",
```

liangosc: removed extra braces.

qcodes/instrument_drivers/tektronix/Keithley_2450.py
liangosc: Yes that works.

liangosc:
Please put this PR on hold because the author wants to double check the backwards compatibility of the new driver, with physical access to the Keithley 2450.

liangosc:
> Please put this PR on hold because the author wants to double check the backwards compatibility of the new driver, with physical access to the Keithley 2450.

liangosc:
Oops I closed it by accident... (I was going to cancel the comment, but didn't realize it's "close and comment" not "cancel the comment"). So I had to reopen it.

Anyway, I tested the driver in this PR and everything works including the examples in the notebook. Please review.

PR #1973, opened 2020-04-14
Add note in 15_min_qcodes notebook that overwriting get, set is an error.

lakhotiaharshit:
Added a small paragraph stating not to use get, set in subclass of _BaseParameter instead use get_raw and set_raw


docs/examples/15_minutes_to_QCoDeS.ipynb
```diff
@@ -55,7 +55,19 @@
    "cell_type": "code",
    "execution_count": 1,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "ename": "ModuleNotFoundError",
+     "evalue": "No module named 'qcodes'",
+     "output_type": "error",
```
jenshnielsen, line 62: Looks like something strange happened when you executed the notebook. Can you reexecute?

jenshnielsen:
We should also fix the place where we actually do this in the notebook 

```
dmm.v1.get = partial(customgetter, dac)
```
by creating a subclass of the dmm with the new get_raw command

docs/examples/15_minutes_to_QCoDeS.ipynb
lakhotiaharshit: I also realized something had gone wrong when PR showed few hundred lines are deleted.

docs/examples/DataSet/Working-With-Pandas-and-XArray.ipynb
```diff
@@ -185,14 +195,77 @@
...
+   "execution_count": 7,
    "metadata": {},
    "outputs": [
     {
-     "name": "stdout",
+     "name": "stderr",
      "output_type": "stream",
      "text": [
-      "Starting experimental run with id: 125\n"
+      "2020-04-15 10:25:37,830 ¦ qcodes.dataset.sqlite.connection ¦ ERROR ¦ connection ¦ atomic ¦ 106 ¦ Rolling back due to unhandled exception\n",
```
jenshnielsen, line 205: Could you rerun this one ?

lakhotiaharshit: Thanks for catching this. It happened most likely because I was running all notebooks at once.

PR #1976, opened 2020-04-16
Add settable/gettable attribute to parameters

jenshnielsen:
This should enable any user to detect if a parameter can be get/set without having to use getattr/hasattr introspection on the parameter.


qcodes/tests/test_parameter.py
```diff
@@ -1762,3 +1763,42 @@ def test_unknown_args_to_baseparameter_warns():
...
+            self._value = initial_value
+            super().__init__(*args, **kwargs)
+
+        def get_raw(self):
+            return self._value
+
+        def set_raw(self, value):
+            self._value = value
+
+    a = GetSetParam('foo', None, initial_value=1)
```
astafan8, line 1796: took me a while to realize that `None` here is an instrument because it doesn't have a default value in _BaseParameter but has in Parameter :)

qcodes/tests/test_parameter.py
```diff
@@ -1762,3 +1763,42 @@ def test_unknown_args_to_baseparameter_warns():
...
+
+    assert a.gettable is expected_gettable
+    assert a.settable is expected_settable
+
+
+def test_gettable_settable_attributes_with_get_set_raw():
+    """Test that parameters that have get_raw,set_raw are
+    listed as gettable/settable and reverse."""
+
+    class GetSetParam(_BaseParameter):
```
astafan8, line 1785: for completeness, shall we also add a subclass of a Parameter with get_raw/set_raw? it's great that Parameter already respects that _BaseParameter CAN create get/set methods (and we might have a test for that as well perhaps), but perhaps it's better to be a bit extra safe here?

qcodes/instrument/parameter.py
```diff
@@ -900,6 +904,9 @@ class Parameter(_BaseParameter):
        :meth:`set_raw` methods are automatically wrapped to provide ``get`` and
        ``set``.
 
+    To detect if a parameter is gettable or settable check the attributes ``gettable``
+    and ``settable`` on the parameter.
```
astafan8, line 908: do we have `hasattr(..)` calls in qcodes? perhaps we should convert those to use these attributes already?

qcodes/instrument/parameter.py
```diff
@@ -302,8 +302,10 @@ def __init__(self, name: str,
             and not getattr(self.get_raw,
                             '__qcodes_is_abstract_method__', False)
         )
+        self.gettable = False
         if implements_get_raw:
             self.get = self._wrap_get(self.get_raw)
+            self.gettable = True
```
astafan8, line 308: it's `gettable` as opposed to `is_gettable` for conciseness? other options would be `has_get` or `can_get` (not a fan of this one).

qcodes/tests/test_parameter.py
jenshnielsen: Fine to put in the keyword

qcodes/instrument/parameter.py
jenshnielsen: I have no strong oppinion on this any name if fine with me

qcodes/instrument/parameter.py
jenshnielsen: Converted the use internally in parameter to use this. This breaks the S46 driver due to https://github.com/QCoDeS/Qcodes/pull/1957/files that monkey patches the get and the set functions.

jenshnielsen:
blocked on #1957  for now

qcodes/instrument/parameter.py
astafan8: then we leave as is.

PR #1977, opened 2020-04-16
Remove deprecated community drivers that are in qcodes_contrib_drivers

astafan8:
* and their notebooks
- [x] and tests
- [x] and mypy exclusions
- [x] and docs build exclusions
- [x] and note in the changelog for the un-released


docs/changes/0.14.0.rst
```diff
@@ -0,0 +1,43 @@
...
+
+* ..
+
+Deprecations:
+_____________
+
+* Community drivers that were shipped with ``QCoDeS``, and were deprecated
+  and moved to
+  `qcodes_contrib_drivers <https://github.com/QCoDeS/Qcodes_contrib_drivers>`_
+  some months ago, are now removed from ``QCoDeS``.
```
jenshnielsen, line 32: Maybe this should be a breaking change and not a deprecation since the deprecation happened earlier?

docs/changes/0.14.0.rst
```diff
@@ -0,0 +1,43 @@
+Changelog for QCoDeS 0.14.0
```
jenshnielsen, line 1: Maybe note that this is currently unreleased

PR #1983, opened 2020-04-16
Remove deprecated zmq_helpers.py with Publisher, UnboundedPublisher

astafan8:


jenshnielsen:
Need to remove from the docs via this file https://github.com/QCoDeS/Qcodes/blob/master/docs/api/utils/zmq_helpers.rst

docs/api/utils/zmq_helpers.rst
```diff
@@ -1,5 +0,0 @@
-qcodes.utils.zmq_helpers
-------------------------
-
-.. automodule:: qcodes.utils.zmq_helpers
-   :members:
```
jenshnielsen, line 5: Sorry you will also need to remove the ref to this from the index file in the same dir

astafan8: no, it's on me, should've done the `grep` homework :)

PR #1988, opened 2020-04-20
Use pytest fixtures for tmp dir creation in tests

jenshnielsen:
This should prevent the unable to unlink tmpdir because pytest has a reference to it failures that we see from time to time. https://docs.pytest.org/en/latest/tmpdir.html#temporary-directories-and-files

There are 3 context managers that I have still not changed. These should probably eventually be rewritten as pytest fixtures. 

The pytest fixtures will only create one tmp_path pr test so in tests that requires 2 distinct directories I have used the factory functions to generate another one

### DIFF
--- a/qcodes/tests/dataset/temporary_databases.py
+++ b/qcodes/tests/dataset/temporary_databases.py
@@ -1,3 +1,4 @@
+import tempfile
 import gc
 import os
 from contextlib import contextmanager
@@ -111,13 +112,14 @@ def temporarily_copied_DB(tmp_path, filepath: str, **kwargs):
     Kwargs:
         kwargs to be passed to connect
     """
-    dbname_new = str(tmp_path / 'temp.db')
-    shutil.copy2(filepath, dbname_new)
+    with tempfile.TemporaryDirectory() as tmpdir:
+        dbname_new = os.path.join(tmpdir, 'temp.db')
+        shutil.copy2(filepath, dbname_new)
 
-    conn = connect(dbname_new, **kwargs)
+        conn = connect(dbname_new, **kwargs)
 
-    try:
-        yield conn
+        try:
+            yield conn
 
-    finally:
-        conn.close()
+        finally:
+            conn.close()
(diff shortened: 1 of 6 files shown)


qcodes/tests/dataset/temporary_databases.py
```diff
@@ -15,71 +14,69 @@
...
+            qc.config["core"]["db_debug"] = False
+        initialise_database()
+        yield
+    finally:
+        # there is a very real chance that the tests will leave open
+        # connections to the database. These will have gone out of scope at
+        # this stage but a gc collection may not have run. The gc
+        # collection ensures that all connections belonging to now out of
+        # scope objects will be closed
+        gc.collect()
```
jenshnielsen, line 36: We can perhaps drop these now?

astafan8: we could try. but as per my understanding this really ensures that connections that are "out of scope" are closed, which allows us to detect those that may still be open (due to some stupid bug or bad design)

jenshnielsen: Yes I will leave this for now

astafan8:
> There are 3 context managers that I have still not changed. These should probably eventually be rewritten as pytest fixtures.

in `test_config.py`? i just looked at them, i think that they can all be safely be rewritten with tmp_path. the more interesting one, `config_file_context`, that one seems to be a candidate for a "fixture factory" (if i remember correctly, pytest supports that)

jenshnielsen:
> 
> 
> > There are 3 context managers that I have still not changed. These should probably eventually be rewritten as pytest fixtures.
> 
> in `test_config.py`? i just looked at them, i think that they can all be safely be rewritten with tmp_path. the more interesting one, `config_file_context`, that one seems to be a candidate for a "fixture factory" (if i remember correctly, pytest supports that)

Maybe but they all failed the first time round. I suggest doing them in a separarate pr since this one is working now

PR #1989, opened 2020-04-20
Make parameter name and short name immutable

lakhotiaharshit:
Makes name and short name read only property.


qcodes/instrument/parameter.py
```diff
@@ -257,8 +257,8 @@ def __init__(self, name: str,
             warnings.warn(f"_BaseParameter got unexpected kwargs: {kwargs}."
                           f" These are unused and will be discarded. This"
                           f" will be an error in the future.")
-        self.name = str(name)
-        self.short_name = str(name)
```
jenshnielsen, line 261: I don't think we need an interal property for both short_name and name ? Could we not just have name point to _short_name

qcodes/instrument/parameter.py
```diff
@@ -339,6 +339,16 @@ def __init__(self, name: str,
         # to perform a validation on get
         self._validate_on_get = False
 
+    @property
```
jenshnielsen, line 342: For readablilty should we put these next to full_name ?

qcodes/instrument/parameter.py
lakhotiaharshit: Yes. Done as suggested

qcodes/instrument/parameter.py
lakhotiaharshit: Yes. Done as suggested

qcodes/instrument_drivers/devices.py
```diff
@@ -57,9 +57,9 @@ def __init__(self,
             self.label = "{}_attenuated".format(self.v1.label)
 
         if name:
-            self.name = name
+            self._short_name = name
         else:
-            self.name = "{}_attenuated".format(self.v1.name)
```
jenshnielsen, line 62: I think we can avoid this by just passing the name to the super method. There is no reason why this should be set here

qcodes/instrument/parameter.py
```diff
@@ -797,6 +796,16 @@ def inter_delay(self, inter_delay: float) -> None:
                 'inter_delay ({}) must not be negative'.format(inter_delay))
         self._inter_delay = inter_delay
 
+    @property
+    def name(self) -> str:
+        """Name of the parameter"""
```
jenshnielsen, line 801: Should we clarify that this is identical to short name.

qcodes/instrument/parameter.py
```diff
@@ -2283,9 +2292,9 @@ def __init__(self,
                  unit: str = None) -> None:
         # Set the name
         if name:
-            self.name = name
+            self._short_name = name
         else:
-            self.name = "{}_scaled".format(output.name)
+            self._short_name = "{}_scaled".format(output.name)
```
astafan8, line 2297: so for this particular parameter the self.name/label/unit assignments seem to be done by mistake because anyway the values are passed to super.init below, and the super will take care of assigning it properly. so i suggest that this is fixed (here, or in a separate PR), so that "making name a readonly property" does not have impact on this parameter class.

qcodes/instrument/parameter.py
```diff
@@ -797,6 +796,16 @@ def inter_delay(self, inter_delay: float) -> None:
...
         self._inter_delay = inter_delay
 
+    @property
+    def name(self) -> str:
+        """Name of the parameter"""
+        return self._short_name
+
+    @property
+    def short_name(self) -> str:
+        """Short name of the parameter"""
```
jenshnielsen, line 806: Lets clarif that this is the name without any instrument attached

qcodes/instrument_drivers/stanford_research/SR86x.py
```diff
@@ -37,7 +37,6 @@ def __init__(self, name: str, instrument: 'SR86x') ->None:
                          docstring='Holds an acquired (part of the) data '
                                    'buffer of one channel.')
 
-        self.name = name
```
jenshnielsen, line 40: We should probably check qcodes_contrib_drivers for this pattern before merging

qcodes/instrument_drivers/devices.py
lakhotiaharshit: Thanks. Done

qcodes/instrument_drivers/stanford_research/SR86x.py
lakhotiaharshit: Yes, I did not think of checking that. Thanks for the input.

qcodes/instrument/parameter.py
lakhotiaharshit: Done

qcodes/instrument_drivers/stanford_research/SR86x.py
lakhotiaharshit: I checked qcodes_contrib_driver and did not find any driver assigning new name to `self.name` .

qcodes/instrument/parameter.py
lakhotiaharshit: This is fixed now.

jenshnielsen:
This is probably a breaking change that we should document already now in the changelog

PR #1990, opened 2020-04-21
fix for slack module when a user has not send any messages

peendebak:
The slack module generates an error when retreiving the latest timestamp if a user has not posted any messages. This PR assigns a None value in that case.

@astafan8


peendebak:
@astafan8 There is an error from codacy 'Import outside toplevel (qcodes.utils.slack)' that I do not know how to resolve. The import needs to be inside the patch environment

jenshnielsen:
@peendebak Codacy is a linting tool and marked as not required so feel free to ignore it when it does not make sense to make the change

jenshnielsen:
@peendebak could you either merge master into this branch or give admins rights to commit to the branch so that I can do it

peendebak:
> @peendebak could you either merge master into this branch or give admins rights to commit to the branch so that I can do it

@jenshnielsen Done

PR #1994, opened 2020-05-06
Do not overwrite get in GroupParameter

lakhotiaharshit:
Do not overwrite get at init of GroupParameter

@jenshnielsen @astafan8

PR #1995, opened 2020-05-06
Group Parameter | set: update cache | get(update): only if get_cmd is not None

lakhotiaharshit:
This PR does the following:
 - Setting value on group should update cache.
 - While getting value if get_cmd is none then just give the raw_value. 

To do after review:
- [x] ~~Add change log for this and~~ #1994 PR

### DIFF
--- a/qcodes/instrument/group_parameter.py
+++ b/qcodes/instrument/group_parameter.py
@@ -8,7 +8,7 @@ should be of type :class:`GroupParameter`
 from collections import OrderedDict
 from typing import List, Union, Callable, Dict, Any, Optional
 
-from qcodes.instrument.parameter import Parameter
+from qcodes.instrument.parameter import Parameter, ParamRawDataType
 from qcodes import Instrument
 
 
@@ -55,7 +55,8 @@ class GroupParameter(Parameter):
         self._initial_value = initial_value
         super().__init__(name, instrument=instrument, **kwargs)
 
-    def get_raw(self, parameter_value: Any = ParamRawDataType) -> ParamRawDataType:
+    def get_raw(self, parameter_value: ParamRawDataType = None) -> \
+            ParamRawDataType:
         return parameter_value if parameter_value is not None else \
             self._get_raw_value()
 
--- a/qcodes/instrument/group_parameter.py
+++ b/qcodes/instrument/group_parameter.py
@@ -60,7 +60,7 @@ class GroupParameter(Parameter):
         return parameter_value if parameter_value is not None else \
             self._get_raw_value()
 
-    def _get_raw_value(self) -> Any:
+    def _get_raw_value(self) -> ParamRawDataType:
         if self.group is None:
             raise RuntimeError("Trying to get Group value but no "
                                "group defined")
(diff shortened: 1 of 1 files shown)


qcodes/instrument/group_parameter.py
```diff
@@ -221,7 +221,10 @@ def set(self, set_parameter: GroupParameter, value: Any) -> None:
         """
         if any((p.get_latest() is None) for p in self.parameters.values()):
             self.update()
-        calling_dict = {name: p.raw_value
+        for _, p in self.parameters.items():
+            print(f'while setting raw value{p.cache.raw_value} and type'
```
jenshnielsen, line 225: remove print statement?

lakhotiaharshit: yes, stupid mistake. Thanks for catching

qcodes/instrument/group_parameter.py
```diff
@@ -239,15 +239,20 @@ def _set_from_dict(self, calling_dict: Dict[str, Any]) -> None:
...
         Update the values of all the parameters within the group by calling
         the ``get_cmd``.
         """
-        if self.instrument is None:
-            raise RuntimeError("Trying to update GroupParameter not attached "
-                               "to any instrument.")
-        ret = self.get_parser(self.instrument.ask(self.get_cmd))
-        for name, p in list(self.parameters.items()):
-            p.get(parameter_value=ret[name])
+        if self.get_cmd is not None:
```
jenshnielsen, line 251: It seems dangerous that this will silently do nothing if self.get_cmd is none. Could you elaborate on how this should work?

lakhotiaharshit: The get_raw of the parameter first `updates` all the parameter of the group and then outputs the updated raw_value from the updated cache. In case the parameter does not have a get_cmd the GroupParameter was failing. With the above implementation if the get_cmd is None the GroupParameter falls back to taking cached value.

lakhotiaharshit: Had a discussion offline. Agreed to revert it back as this seems dangerous to both of us.

jenshnielsen:
This looks good to me but it should probably have a test

qcodes/instrument/group_parameter.py
```diff
@@ -239,15 +239,19 @@ def _set_from_dict(self, calling_dict: Dict[str, Any]) -> None:
...
         the ``get_cmd``.
         """
         if self.instrument is None:
-            raise RuntimeError("Trying to update GroupParameter not attached "
-                               "to any instrument.")
+            raise RuntimeError("Trying to update GroupParameter not "
+                               "attached to any instrument.")
+
         ret = self.get_parser(self.instrument.ask(self.get_cmd))
         for name, p in list(self.parameters.items()):
```
jenshnielsen, line 256: It seems to me like this logic should just be replaced by setting the cache of the parameter. But that can be done in a different pr

jenshnielsen:
@astafan8 @lakhotiaharshit Do you need this one for your b1500 work ? I can put some work into it tomorrow if that is useful?

astafan8:
@jenshnielsen this basically what decided to do after our offline conversation https://github.com/QCoDeS/Qcodes/compare/master...astafan8:group-param-cache

astafan8:
> Do you need this one for your b1500 work ? 

not strictkyl needed for b1500 but would be great to fix

PR #1996, opened 2020-05-06
added range function into keysight submodule

ianhuezo:
- Adding a method to increase the range of a keysight instrument
- accesses and searches the ranges and implements them into the setter for the range method

@jenshnielsen @astafan8

### DIFF
--- a/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
+++ b/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
@@ -1,7 +1,7 @@
 import textwrap
 from contextlib import ExitStack
 from functools import partial
-from typing import Sequence, Tuple, Any, Optional, Union
+from typing import Sequence, Tuple, Any, Optional
 from distutils.version import LooseVersion
 from bisect import bisect_left
 
@@ -910,7 +910,7 @@ class _Keysight_344xxA(KeysightErrorQueueMixin, VisaInstrument):
 
     def increase_range(
         self,
-        range_value: Optional[Union[float, int]] = None,
+        range_value: Optional[float] = None,
         increase_by: int = 1
     ) -> None:
         """
@@ -939,7 +939,7 @@ class _Keysight_344xxA(KeysightErrorQueueMixin, VisaInstrument):
 
     def decrease_range(
         self,
-        range_value: Optional[Union[float, int]] = None,
+        range_value: Optional[float] = None,
         decrease_by: int = -1
     ) -> None:
         """
--- a/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
+++ b/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
@@ -909,7 +909,9 @@ class _Keysight_344xxA(KeysightErrorQueueMixin, VisaInstrument):
         self.range.get()
 
     def increase_range(
-        self, range_value: Optional[Union[float, int]] = None, increase_by: int = 1
+        self, 
+        range_value: Optional[Union[float, int]] = None,
+        increase_by: int = 1
     ) -> None:
         """
         Increases the voltage range by a certain amount with default of 1.
@@ -918,8 +920,8 @@ class _Keysight_344xxA(KeysightErrorQueueMixin, VisaInstrument):
         Args:
             range_value : The desired voltage range needed.  Expressed by power
                 of 10^x range from -3 to 10
-            increase_by: How much to increase range by, default behavior is by a step
-                of one.
+            increase_by: How much to increase range by, default behavior
+                 is by a step of one.
 
         """
         if increase_by < 1:
--- a/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
+++ b/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
@@ -918,7 +918,7 @@ class _Keysight_344xxA(KeysightErrorQueueMixin, VisaInstrument):
         If limit is reached, the max range is used.
 
         Args:
-            range_value : The desired voltage range needed.  Expressed by power
+            range_value: The desired voltage range needed.  Expressed by power
                 of 10^x range from -3 to 10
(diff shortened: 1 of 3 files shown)


ianhuezo:
I'll add a couple of tests

jenshnielsen:
Thanks @ianhuezo looks good. Tests would be very welcome. Perhaps we should also add a matching decrease_range method?

ianhuezo:
I'll add a decrease range function too.

ianhuezo:
@jenshnielsen Something interesting was coming out of my test when I queried an index <4 for
```
driver_range_user = driver.ranges[3]
driver.decrease_range(driver_range_user, -2)
assert driver.range() == driver.ranges[1] #where the value is supposed to be 0.01 but errors
#results in error
```
Is it meant to error on a range at driver.range(0.01)?

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -907,6 +908,64 @@ def autorange_once(self) -> None:
         self.write('SENSe:VOLTage:DC:RANGe:AUTO ONCE')
         self.range.get()
 
+    def increase_range(
+        self,
+        range_value: Optional[Union[float, int]] = None,
```
jenshnielsen, line 913: This can be simplified to `Optional[float]` the hierarchy of numeric types in python is such that a float value is allowed to be an int too

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -907,6 +908,64 @@ def autorange_once(self) -> None:
...
+        increase_by: int = 1
+    ) -> None:
+        """
+        Increases the voltage range by a certain amount with default of 1.
+        If limit is reached, the max range is used.
+
+        Args:
+            range_value : The desired voltage range needed.  Expressed by power
```
jenshnielsen, line 921: We can remove the space before the :

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -907,6 +908,64 @@ def autorange_once(self) -> None:
...
+
+        index = bisect_left(self.ranges, current_range)  # binary search
+        if index + increase_by < len(self.ranges):
+            self.range(self.ranges[index + increase_by])
+        else:
+            self.range(self.ranges[-1])
+
+    def decrease_range(
+        self,
+        range_value: Optional[Union[float, int]] = None,
```
jenshnielsen, line 942: same

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -907,6 +908,64 @@ def autorange_once(self) -> None:
...
+        decrease_by: int = -1
+    ) -> None:
+        """
+        Decrease the voltage range by a certain amount with default of -1.
+        If limit is reached, the min range is used.
+
+        Args:
+            range_value : The desired voltage range needed.  Expressed by power
```
jenshnielsen, line 950: same

jenshnielsen:
@ianhuezo not sure what is going on there. Can you capture the traceback? That might give a clue

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -907,6 +908,64 @@ def autorange_once(self) -> None:
...
+                of 10^x range from -3 to 10
+            increase_by: How much to increase range by, default behavior
+                 is by a step of one.
+
+        """
+        if increase_by < 1:
+            raise ValueError("The steps must be increasing in value")
+
+        current_range = self.range.get()
```
jenshnielsen, line 930: Lets move this call to get inside the if statement. That way we can avoid calling get if a user has supplied the range

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -907,6 +908,64 @@ def autorange_once(self) -> None:
...
+            range_value : The desired voltage range needed.  Expressed by power
+                of 10^x range from -3 to 10
+            decrease_by: How much to decrease range by, default behavior
+                 is by a step of one.
+
+        """
+        if decrease_by > -1:
+            raise ValueError("The steps must be decreasing in value")
+
+        current_range = self.range.get()
```
jenshnielsen, line 959: same

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
ianhuezo: That makes sense.  I'll change that.

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
ianhuezo: I used an if-else for the statement.  Is that ok?

jenshnielsen:
@ianhuezo the problem was that the ranges specified in the simulation yaml file were incorrect. I have pushed a fix.

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -907,6 +908,68 @@ def autorange_once(self) -> None:
...
+            range_value: The desired voltage range needed.  Expressed by power
+                of 10^x range from -3 to 10
+            increase_by: How much to increase range by, default behavior
+                 is by a step of one.
+
+        """
+        if increase_by < 1:
+            raise ValueError("The steps must be increasing in value")
+
+        current_range = 1.0 #default to 10 ** 0
```
jenshnielsen, line 930: This line has no effect right as we either use the if or else branch?

ianhuezo: Should be good to go now.

PR #1997, opened 2020-05-07
Additions to B1500 driver for Capacitance measurements

lakhotiaharshit:
Contains the ongoing work for the driver. The description will be updated soon. 

The plan is to implement the most necessary stuff first. However some of the things must immediately follow after the necessary stuff to comply with QCoDeS standards. To not hinder the process the followup items will be taken up most likely in another PR. 
Necessary 
- [ ] Add tests 
Follow up Do: 
- [ ] Hold: validator with 10ms step can also be added
- [ ] Delay:  0.1 ms reso validator can be added
- [ ] adc_coeff = validator must depend on adc_mode and change to 1 to 1023 if not in PLC mode 
- [ ] CVMeasurement(MultiParamater) Setpoint units and label must depend on the mode.


qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_base.py
```diff
@@ -273,3 +275,140 @@ def error_message(self, mode: Optional[Union[constants.ERRX.Mode,
...
+    
+
+    def setup_staircase_sweep(
+        self,
+        v_start: float,
+        v_end: float,
+        N_steps: int,
+        post_sweep_voltage_val: int = constants.WMDCV.Post.STOP,        
+        sweep_chan = 1,
+        measure_chan_list = [1],
```
jenshnielsen, line 287: It normally a bad idea to use mutable default arguments. I suggest using a tuple here 

https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

lakhotiaharshit: Thanks! I will keep this in mind. As a side note, I will create a new PR for SMU driver only and abandon this one. The CMU part is already done in #2001.

jenshnielsen:
@lakhotiaharshit Is this abandoned and should it be closed?

lakhotiaharshit:
This is work is already been taken care of in PR #2028 and in #2001

PR #1998, opened 2020-05-10
Add get_cmd for voltage_dc, voltage_ac and frequency in KeysightB1520A

lakhotiaharshit:
This PR fixes following 2 issues:
1) is_enabled functions was giving error when called before enabling channels. This is fixed by checking if activated_channels is not empty string. 
2) Adds get_cmd to voltage_dc, voltage_ac, frequency

The above two fixes allows user to use the `step` attribute of the parameter to slowly raise/decrease a parameter. Before these fixes the user was getting warning that parameter is jumping to designated value because starting value of the parameter is not found is there was no get_cmd available. 

@astafan8 @jenshnielsen

### DIFF
--- a/qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
+++ b/qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
@@ -76,7 +76,7 @@ _pattern_lrn = re.compile(
 )
 
 
-def parse_dcv_measurement_response(response: str) -> dict:
+def parse_dcv_measurement_response(response: str) -> Dict[str, Union[str, float]]:
     """
     Extract status, channel number, value  and accompanying metadata from
     the string and return them as a dictionary.
--- a/qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
+++ b/qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
@@ -68,6 +68,32 @@ def parse_module_query_response(response: str) -> Dict[SlotNr, str]:
     }
 
 
+# pattern to match dcv experiment
+_pattern_lrn = re.compile(
+    r"(?P<status_dc>\w{1,3})(?P<chnr_dc>\w),(?P<voltage_dc>\d{1,3}.\d{1,4});"
+    r"(?P<status_ac>\w{1,3})(?P<chnr_ac>\w),(?P<voltage_ac>\d{1,3}.\d{1,4});"
+    r"(?P<status_fc>\w{1,2})(?P<chnr_fc>\w),(?P<frequency>\d{1,6}.\d{1,4})"
+)
+
+
+def parse_dcv_measurement_response(response: str) -> dict:
+    """
+    Extract status, channel number, value  and accompanying metadata from
+    the string and return them as a dictionary.
+
+    Args:
+        response: Response str to *LRN? measurement query For the MFCMU.
+    """
+
+    match = re.match(_pattern_lrn, response)
+    if match is None:
+        raise ValueError(f"{response!r} didn't match {_pattern_lrn!r} pattern")
+
+    dd = match.groupdict()
+    d = cast(Dict[str, Union[str, float]], dd)
+    return d
+
+
 # Pattern to match the spot measurement response against
 _pattern = re.compile(
     r"((?P<status>\w)(?P<chnr>\w)(?P<dtype>\w))?"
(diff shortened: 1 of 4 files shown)


qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1520A.py
```diff
@@ -89,6 +92,22 @@ def _set_voltage_ac(self, value: float) -> None:
 
         self.write(msg.message)
 
+    def _get_dcv(self):
+        if self.is_enabled():
+            msg = MessageBuilder().lrn_query(self.channels[0])
+            response = self.ask(msg.message)
+            d = parse_dcv_measurement_response(response)
+        return d or None
```
astafan8, line 100:
```suggestion
    def _get_dcv(self) -> Optional[Dict]:
        d: Optional[Dict] = None
        if self.is_enabled():
            msg = MessageBuilder().lrn_query(self.channels[0])
            response = self.ask(msg.message)
            d = parse_dcv_measurement_response(response)
        return d
```

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1520A.py
```diff
@@ -89,6 +92,22 @@ def _set_voltage_ac(self, value: float) -> None:
...
+
+    def _get_voltage_dc(self) -> Optional[float]:
+        return float(self._get_dcv()['voltage_dc'])
```
astafan8, line 103:
```suggestion
    def _get_voltage_dc(self) -> Optional[float]:
        dcv = self._get_dcv()
        return float(dcv['voltage_dc']) if dcv is not None else None
```

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1520A.py
astafan8, line 100: does it make sense to allow returning ``None``? i'd say if `is_enabled` is `False`, then just raise an error. This will also simplify the get methods for the individual parameters

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
```diff
@@ -222,6 +248,6 @@ def is_enabled(self) -> bool:
         activated_channels = re.sub(r"[^,\d]", "", response).split(",")
 
         is_enabled = set(self.channels).issubset(
-            int(x) for x in activated_channels
+            int(x) for x in activated_channels if x is not ''
```
astafan8, line 251: thanks for fixing this, noticed it last week as well! :)

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
```diff
@@ -68,6 +68,32 @@ def parse_module_query_response(response: str) -> Dict[SlotNr, str]:
...
 
+# pattern to match dcv experiment
+_pattern_lrn = re.compile(
+    r"(?P<status_dc>\w{1,3})(?P<chnr_dc>\w),(?P<voltage_dc>\d{1,3}.\d{1,4});"
+    r"(?P<status_ac>\w{1,3})(?P<chnr_ac>\w),(?P<voltage_ac>\d{1,3}.\d{1,4});"
+    r"(?P<status_fc>\w{1,2})(?P<chnr_fc>\w),(?P<frequency>\d{1,6}.\d{1,4})"
+)
+
+
+def parse_dcv_measurement_response(response: str) -> dict:
```
astafan8, line 79:
```suggestion
def parse_dcv_measurement_response(response: str) -> Dict[str, Union[str, float]]:
```

astafan8: ... or shall we be a bit fancier and return a `TypedDict`? then mypy will be able to check for dict keys correctness

jenshnielsen: +1 for typeddict :)

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1520A.py
jenshnielsen: Yes either that or remove the Optional from the type

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1520A.py
lakhotiaharshit: Alternate suggestion accepted

PR #1999, opened 2020-05-11
Add schedule for nightly runs at Azure pipelines

GateBuilder:
This PR introduces Nightly runs for CI at Azure pipelines. Currently it is configured to run exactly at midnight and to run for branches `master` and `releases`. Please let me know if you would like to have a different config.

@jenshnielsen


jenshnielsen:
It seems like we currently have sufficently speedy jobs that we can run them for all commits/prs and do not need nightly jobs

PR #2005, opened 2020-05-15
Allow Delegate parameter to change source and support a None source

jenshnielsen:
There are a few open questions. Especially around how we handle traits such as gettable/settable/snapshot_get being inherited from the source parameter or alternatively set specifically on the Delegated parameter


qcodes/instrument/parameter.py
```diff
@@ -1292,33 +1305,60 @@ def _update_with(self, *,
...
+
+    def _set_properties_from_source(self, source: Optional[Parameter]) -> None:
+        if source is None:
+            self.gettable = False
+            self.settable = False
+            self._snapshot_get = False
+        else:
+            self.gettable = source.gettable
+            self.settable = source.settable
+            self._snapshot_get = source._snapshot_get
```
jenshnielsen, line 1345: Is this what we want? Should the parameter always follow its source or should we allow this to be overwritten in the Delgate

qcodes/instrument/parameter.py
```diff
@@ -1292,33 +1305,60 @@ def _update_with(self, *,
...
+    @source.setter
+    def source(self, source: Optional[Parameter]) -> None:
+        self._set_properties_from_source(source)
+        self._source: Optional[Parameter] = source
+
+    def _set_properties_from_source(self, source: Optional[Parameter]) -> None:
+        if source is None:
+            self.gettable = False
+            self.settable = False
+            self._snapshot_get = False
```
astafan8, line 1341:
```suggestion
            self._snapshot_value = False
```

qcodes/instrument/parameter.py
astafan8: >Should the parameter always follow its source

yes, as discused offline

qcodes/instrument/parameter.py
```diff
@@ -1292,33 +1305,60 @@ def _update_with(self, *,
...
+        else:
+            self.gettable = source.gettable
+            self.settable = source.settable
+            self._snapshot_get = source._snapshot_get
+            # TODO should remapping the source change the unit/label
+            # perhaps only if the unit/label has not been supplied as
```
astafan8, line 1347: yes, for unit and label only

qcodes/instrument/parameter.py
```diff
@@ -1308,25 +1309,38 @@ def raw_value(self) -> ParamRawDataType:
             This bug will not be fixed since the `raw_value` property will be
             removed soon.
             """
-            return self._parameter.source.cache._value
+            if self._parameter.source is None:
+                return None
+            return self._parameter.source.cache.get(get_if_invalid=False)
```
jenshnielsen, line 1321: @astafan8 I am going to suggest leaving the Cache as is. The only other alternative that I can see is to make the Cache member of the parameter optional. That is a significanly more intrusive change.

astafan8: that's fine, we can get to this later

qcodes/instrument/parameter.py
```diff
@@ -943,8 +951,8 @@ class Parameter(_BaseParameter):
         snapshot_get: ``False`` prevents any update to the
             parameter during a snapshot, even if the snapshot was called with
             ``update=True``, for example, if it takes too long to update,
-            or if the parameter is only meant for measurements hence its value
-            in the snapshot may not always make sense. Default True.
+            or if the parameter is only meant for measurements hence calling
+            get on in during snapshot may be an error. Default True.
```
astafan8, line 955:
```suggestion
            get on it during snapshot may be an error. Default True.
```

qcodes/instrument/parameter.py
```diff
@@ -1272,15 +1282,28 @@ def validate(self, value: ParamDataType) -> None:
...
     the provided arguments.
 
     The reason for using a :class:`DelegateParameter` instead of the
     source parameter is to provide all the functionality of the Parameter
     base class without overwriting properties of the source: for example to
     set a different scaling factor and unit on the :class:`.DelegateParameter`
-    without changing those in the source parameter
+    without changing those in the source parameter.
+
+    The :class:`DelegateParameter` supports chancing the `source`
```
astafan8, line 1295:
```suggestion
    The :class:`DelegateParameter` supports changing the `source`
```

qcodes/instrument/parameter.py
```diff
@@ -1272,15 +1282,28 @@ def validate(self, value: ParamDataType) -> None:
...
+    :class:`Parameter`. :py:attr:`~gettable`, :py:attr:`~settable` and
+    :py:attr:`snapshot_value` properties automatically follows the source
```
astafan8, line 1297:
```suggestion
    :py:attr:`snapshot_value` properties automatically follow the source
```

qcodes/instrument/parameter.py
```diff
@@ -1272,15 +1282,28 @@ def validate(self, value: ParamDataType) -> None:
...
+    parameter. If source is set to ``None`` :py:attr:`~gettable` and
+    :py:attr:`~settable` will always be ``False``. It is therefore an error
+    to call get and set on a :class:`DelegateParameter` without a `source`.
+    Note that a parameter without a source can be snapshotted correctly.
+
+    :py:attr:`.unit` and :py:attr:`.label` can either be set when constructing
+    a :class:`DelegateParameter` or inherited from the source
+    :class:`Parameter` If inherited they will automatically change when
```
astafan8, line 1305:
```suggestion
    :class:`Parameter`. If inherited they will automatically change when
```

qcodes/instrument/parameter.py
```diff
@@ -1349,35 +1381,89 @@ def _update_with(self, *,
...
+        self._attr_inherit = {"label": {"fixed": False,
+                                        "default_none_value": name},
+                              "unit": {"fixed": False,
+                                       "default_none_value": ""}}
+
+        for attr, attr_props in self._attr_inherit.items():
+            if attr in kwargs:
+                attr_props["fixed"] = True
+            else:
+                attr_props["fixed"] = True
```
astafan8, line 1403: typo-bug?
```suggestion
                attr_props["fixed"] = False
```

qcodes/instrument/parameter.py
```diff
@@ -1349,35 +1381,89 @@ def _update_with(self, *,
...
                 raise KeyError(f'It is not allowed to set "{cmd}" of a '
                                f'DelegateParameter because the one of the '
                                f'source parameter is supposed to be used.')
+        if source is None and ("initial_cache_value" in kwargs
+                               or "initial_value" in kwargs):
+            raise KeyError("It is not allowed to supply 'initial_value'"
+                           " or 'initial_cache_value' "
+                           "without a source.")
+
         initial_cache_value = kwargs.pop("initial_cache_value", None)
```
astafan8, line 1419: this is not related to this PR but i wonder why we don't do the same for the initial_value...

qcodes/instrument/parameter.py
```diff
@@ -1349,35 +1381,89 @@ def _update_with(self, *,
...
         if initial_cache_value is not None:
             self.cache.set(initial_cache_value)
 
-    # Disable the warnings until MultiParameter has been
-    # replaced and name/label/unit can live in _BaseParameter
+    @property
+    def source(self) -> Optional[Parameter]:
+        """
+        The source parameter that this :class:`DelegateParameter` is bound to
+        or ``None`` if the source is unbound.
```
astafan8, line 1434: "if the source is unbound" or "if this :class:`DelegateParameter` is unbound"? not sure what's correct

qcodes/instrument/parameter.py
```diff
@@ -1349,35 +1381,89 @@ def _update_with(self, *,
...
-    def __init__(self, name: str, source: Parameter, *args: Any,
+    def __init__(self, name: str, source: Optional[Parameter], *args: Any,
                  **kwargs: Any):
-        self.source = source
 
-        for ka, param in zip(('unit', 'label', 'snapshot_value'),
-                             ('unit', 'label', '_snapshot_value')):
-            kwargs[ka] = kwargs.get(ka, getattr(self.source, param))
+        self._attr_inherit = {"label": {"fixed": False,
+                                        "default_none_value": name},
```
astafan8, line 1395: default_none_value means "the value of the attr to be used when the source is None", right? if so, then perhaps a bit more specific name such as "value_when_unbound" or "value_when_without_source" (i think i prefer this so far) or "value_when_none_source" or "source_none_value"?

qcodes/instrument/parameter.py
```diff
@@ -1387,9 +1473,14 @@ def snapshot_base(self, update: Optional[bool] = True,
...
-            {'source_parameter': self.source.snapshot(update=update)}
-        )
+        if self.source is None:
+            snapshot.update(
+                {'source_parameter': None}
+            )
+        else:
+            snapshot.update(
+                {'source_parameter': self.source.snapshot(update=update)}
+            )
```
astafan8, line 1490: perhaps smth like this in order not to duplicate the "source_parameter" key?
```suggestion
        source_parameter_snapshot = None if self.source is None \
            else self.source.snapshot(update=update)
        snapshot.update(
            {'source_parameter': source_parameter_snapshot}
        )
```

qcodes/tests/test_delegate_parameter.py
```diff
@@ -294,3 +296,115 @@ def test_setting_initial_cache_delegate_parameter():
...
+    assert snapshot == updated_snapshot
+
+
+def _assert_delegate_cache_none_source(delegate_param):
+    with pytest.raises(TypeError):
+        delegate_param.cache.set(1)
+    with pytest.raises(TypeError):
+        delegate_param.cache.get()
+    with pytest.raises(TypeError):
+        delegate_param.cache.raw_value()
```
astafan8, line 372: it's a property, right? not a method
```suggestion
        delegate_param.cache.raw_value
```

qcodes/instrument/parameter.py
```diff
@@ -1308,29 +1331,45 @@ def raw_value(self) -> ParamRawDataType:
...
+                                "that delegates to None")
             self._parameter.validate(value)
             self._parameter.source.cache.set(
                 self._parameter._from_value_to_raw_value(value))
 
+        def _set_from_raw_value(self, value: ParamRawDataType) -> None:
+            if self._parameter.source is None:
+                raise TypeError("Cannot set the cache of a DelegateParameter "
+                                "that delegates to None")
+            self._parameter.validate(value)
```
astafan8, line 1370: as per my memory, _set_from_raw_value was made with an assumtion that it will be used in a "get" context (as opposed to "set"). this is why in _BaseParameter, _set_from_raw_value 
 validates the value only if _validate_on_get is True. I don't think that this assumtion is a good one. So either the _validate_on_get  condition needs to be gone but then we need to be careful about validation values in group parameter case, or add the _validate_on_get  condition below for consistency just "for now":
```python
if self._parameter._validate_on_get:
    self._parameter.validate(value)
```

what do you think?

qcodes/instrument/parameter.py
jenshnielsen: Sure

qcodes/instrument/parameter.py
jenshnielsen: Hmn yes. Seems like the tests are not strict enough then

qcodes/instrument/parameter.py
jenshnielsen: Because initial_value will call set on the parameter which wil call set on the underlying parameter which the cache can read from. It is all covered in tests :) However inital_cache_value will only set the value in the `_Cache` that is replaced with the `DelegateCache` below

astafan8: yes, make sense, thanks!

qcodes/instrument/parameter.py
jenshnielsen: Fixed and added tests

qcodes/instrument/parameter.py
jenshnielsen: Actually this never matters. It calls set on the base parameter which will always do a validate. In anycase it should validate against the value and not the raw_value. So lets perhaps just remove this validator

qcodes/instrument/parameter.py
jenshnielsen: Not crazy about the line break but sure

PR #2007, opened 2020-05-17
Added fixed frequency readout mode (CW mode) to ZNB driver

GitJaap:
The ZNB20 VNA driver currently supports only the 'linear' sweep mode. This is by default set when instantiating a channel. For things like two tone spectroscopy we would then hack a linear frequency trace with 1 Hz span such that we could do readout at a single point. 

It turns out the VNA has internal support for fixed frequency readout so I added that to the driver. I added it in a way that the current way the VNA is used should still function as it did, but you now have the option to switch to Continuous wave mode (CW mode). In CW mode there is no AVG setting(defaults to 1) and the integration time per point is set by 1/bandwidth. This allows for doing measurements of at the minimum 1us(max BW=1MHz) integration time at a fixed frequency and doing time resolved traces with the VNA or a crude version of single shot measurements. Also two tone can be faster (depending on how you implemented it with the vna before) since this reduces overhead (e.g simulating extremely small frequency stepping). And a small bonus: now you can go to 100k points instead of the 30k that is the limit in linear frequency sweep mode. 

For this added three parameters similar to the existing ones:
- trace_fixed_frequency -> measures the raw real and imag part of the signal and sets time as the x-axis.
- point_fixed_frequency -> basically averages the trace_fixed_frequency into a single point which is useful if you don't care about the individual I-Q outcomes but you just want the average for e.g two-tone spectroscopy.
- point_fixed_frequency_mag_phase -> same as point_fixed_frequency but returns the mag/phase instead of I-Q if you are more used to that. 

@mention one core developer
- @lakhotiaharshit @astafan8 Could you have a look if this is in any way according to the qcodes protocol and if you have suggestions for additions. 
- Also, I some example code, but where should i put that preferably?

## Example
``` python
pwr=-20
npts=700
bw=1e5
frequency=4.823e9

exp = new_experiment("VNA fixed freq P=%d dBM,N=%d,BW=%d KHz" % ( pwr,npts,bw/1e3), sample_name)
meas = Measurement()

meas.register_parameter(vna.channels.S21.trace_fixed_frequency)

#setup vna CW mode
vna.S21.setup_cw_sweep()
#set vna params
vna.S21.cw_frequency(frequency)
vna.S21.npts(npts)
vna.S21.bandwidth(bw)
vna.S21.power(pwr)
vna.rf_on()

with meas.run() as datasaver:
    get_v = vna.S21.trace_fixed_frequency.get()
    datasaver.add_result((vna.S21.trace_fixed_frequency, get_v))
    
vna.S21.power.set(-60)
#reset the vna back to linear mode so you can use if for frequency traces again
vna.S21.setup_lin_sweep()
vna.rf_off()
```
## result:
image

 Note the time axis is here determined by npts/BW, but there is a little overhead so for very fast measurements one should note that the time might not perfectly correspond to the time the point was taken. I tested it with %timeit and the overhead was about 3 to 5ms minimum, but I'm not sure where that happens (communication or internally.) I'm guessing in communication since when you increase npts the overhead scales sub-linearly.


abargerbos:
@astafan8 We are currently using this fork on 3 (that I know of) measurement setups in Delft (QT3, K1, K2), it has been a very useful addition to the driver!

astafan8:
@abargerbos @GitJaap thanks for reminding, I'll have a look at it!

LGruenhaupt:
Hi @astafan8, 
just pinging you again to ask about the status of this. As Arno already mentioned, we are running this fork still on three setups (K1, K2/QT6, QT3). Is there some additional work to be done before this gets pulled into the qcodes master branch?

astafan8:
@LGruenhaupt thanks for reminding! we'll sync with @Akshita07 and let you know :)

Akshita07:
@LGruenhaupt thank you for reminding. This PR will be merged in the qcodes master the upcoming week.
@astafan8 please have a final look at the driver code. For the example notebook, I'll create a separate PR so that this merge is not stalled more.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -270,9 +391,39 @@ def __init__(self, parent: 'ZNB', name: str, channel: int,
                            get_cmd=f'SENS{n}:SWE:TIME?',
                            get_parser=float,
                            unit='s')
+        # Allow switching the default linear VNA sweep type to other types.
+        # note that at the moment only the linear and CW_Point modes have
+        # supporting measurement parameters
```
jenshnielsen, line 396: I suggest that these comment are rewritten as docstrings for the parameters

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -346,25 +497,42 @@ def _set_stop(self, val: float):
...
+        channel = self._instrument_channel
+        self.write(f'SENS{channel}:SWE:TYPE {val}')
+
+    def _set_cw_frequency(self, val: float):
+        channel = self._instrument_channel
+        self.write(f'SENS{channel}:FREQ:CW {val:.7f}')
 
-    def update_traces(self):
-        """ updates start, stop and npts of all trace parameters"""
+    def update_lin_traces(self):
```
jenshnielsen, line 531: renaming this is an api break. We can add back `update_traces` as a deprecated method that simply calls `update_lin_traces`

jenshnielsen (commented):
Added a few small comments. This will also need full types for all the new functions that were added

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -1,16 +1,143 @@
...
+    See (https://www.rohde-schwarz.com/webhelp/ZNB_ZNBT_HTML_UserManual_en
+    /ZNB_ZNBT_HTML_UserManual_en.htm) under GUI reference -> sweep softtool
+    -> sweep type tab -> CW mode
+    """
+
+    def __init__(self, name: str, instrument: "ZNBChannel",
+                 npts: int, bandwidth: int, channel: int,
+                 check_cw_sweep_first: bool = True) -> None:
+        super().__init__(name, names=("", ""), shapes=((), ()))
+        self._instrument = instrument
```
jenshnielsen, line 30: Please pass the instrument to the super class and let that asign to `self._instrument` rather than repeating the work here. 

Multiple of these below in other parameters too

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -1,16 +1,143 @@
...
+    def set_cw_sweep(self, npts: int, bandwidth: int) -> None:
+        """
+        Needed to update config of the software parameter on sweep change
+        time setpoints tuple as needs to be hashable for look up.
+        Note: This is similar to the set_sweep functions of the frequency
+        sweep parameters.
+        """
+        t = tuple(np.linspace(0, npts/bandwidth, num=npts))
+        self.setpoints = ((t,), (t,))
+        self.shapes = ((npts,), (npts,))
```
jenshnielsen, line 59: How are you supposed to set these back to the default.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -1,16 +1,143 @@
...
+            (f'{instrument.short_name}_time',),
+            (f'{instrument.short_name}_time',)
+        )
+
+    def set_cw_sweep(self, npts: int, bandwidth: int) -> None:
+        """
+        Needed to update config of the software parameter on sweep change
+        time setpoints tuple as needs to be hashable for look up.
+        Note: This is similar to the set_sweep functions of the frequency
+        sweep parameters.
```
jenshnielsen, line 55: This docstring is very hard to read

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -1,16 +1,143 @@
...
+    Sweep that returns the real (I) and imaginary (Q) parts of the VNA response.
+    Requires the use of the sweep type to be set to continuous wave mode.
+    See (https://www.rohde-schwarz.com/webhelp/ZNB_ZNBT_HTML_UserManual_en
+    /ZNB_ZNBT_HTML_UserManual_en.htm) under GUI reference -> sweep softtool
+    -> sweep type tab -> CW mode
+    """
+
+    def __init__(self, name: str, instrument: "ZNBChannel",
+                 npts: int, bandwidth: int, channel: int,
+                 check_cw_sweep_first: bool = True) -> None:
```
jenshnielsen, line 28: How is this supposed to be used? None of the places where this parameter is added passes this flag. I suggest that we make a manual parameter on the instrument called `check_cw_sweep_first` that we check when calling `get_raw` rather than having a flag here

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -177,18 +304,19 @@ def __init__(self, parent: 'ZNB', name: str, channel: int,
...
+                           set_cmd=self._set_bandwidth,
                            get_parser=int,
                            vals=vals.Enum(
                                *np.append(10 ** 6,
                                           np.kron([1, 1.5, 2, 3, 5, 7],
-                                                  10 ** np.arange(6))))
+                                                  10 ** np.arange(6)))),
+                           docstring="There is an 'increased bandwidth' "
+                                     "option' (p. 4 of manual) that does not "
+                                     "get taken into account here."
```
jenshnielsen, line 319: Please start the docstring with explaining what this actually is. 
Something like. 
```
The bandwidth used for sensing. Note There is an 'increased bandwidth option' (p. 4 of manual) that does not get taken into account here.
```

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -270,9 +398,44 @@ def __init__(self, parent: 'ZNB', name: str, channel: int,
...
+                                        'Power': 'POW\n',
+                                        'CW_Time': 'CW\n',
+                                        'CW_Point': 'POIN\n',
+                                        'Segmented': 'SEGM\n',
+                                        },
+                           docstring="Allows switching the default linear VNA "
+                                     "sweep type to other types. Note that "
+                                     "at the moment only the linear and "
+                                     "CW_Point modes have supporting "
+                                     "measurement parameters."
```
jenshnielsen, line 415: please formulate to directly describe what this is

```
The sweep type used when ...  . Note Note that 
                                     at the moment only the linear and
                                     CW_Point modes have supporting 
                                     measurement parameters.
```

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -270,9 +398,44 @@ def __init__(self, parent: 'ZNB', name: str, channel: int,
...
+                           )
+        self.add_parameter(name='cw_frequency',
+                           get_cmd=f'SENS{n}:FREQ:CW?',
+                           set_cmd=self._set_cw_frequency,
+                           get_parser=float,
+                           vals=vals.Numbers(self._parent._min_freq + 0.5,
+                                             self._parent._max_freq - 10),
+                           docstring="Similar to center, except that this is "
+                                     "used when VNA sweep type is set to "
+                                     "CW_Point mode."
```
jenshnielsen, line 425: Please make this docstring stand on its own rather than haing to read another docstring to understand what this does

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -270,9 +398,44 @@ def __init__(self, parent: 'ZNB', name: str, channel: int,
...
+                           )
+        self.add_parameter(name='trace_fixed_frequency',
+                           npts=self.npts(),
+                           bandwidth=self.bandwidth(),
+                           channel=self._instrument_channel,
+                           parameter_class=FixedFrequencyTraceIQ)
+        # Also add an averaged version of trace_fixed_frequency
```
jenshnielsen, line 432: remove this comment I think

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
jenshnielsen: Or perhaps the defaults are just wrong and should be removed. I guess this always has a npts axis ?

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -346,25 +509,49 @@ def _set_stop(self, val: float) -> None:
...
+    def _set_sweep_type(self, val: str) -> None:
+        channel = self._instrument_channel
+        self.write(f'SENS{channel}:SWE:TYPE {val}')
 
+    def _set_cw_frequency(self, val: float) -> None:
+        channel = self._instrument_channel
+        self.write(f'SENS{channel}:FREQ:CW {val:.7f}')
+
+    @deprecate(reason='the method name has been updated to have method names '
+                      'consistent for different modes',
```
jenshnielsen, line 544: I think we can simply write `the method has been renamed`

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -419,6 +623,79 @@ def _get_sweep_data(self, force_polar: bool = False) -> np.ndarray:
                 self.root_instrument.cont_meas_on()
         return data
 
+    def setup_cw_sweep(self) -> None:
+        """
+        Sets CW mode.
```
jenshnielsen, line 628: This docstring does not tell a lot. Could we explain what CW mode is?

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -419,6 +623,79 @@ def _get_sweep_data(self, force_polar: bool = False) -> np.ndarray:
...
+        """
+
+        # set the channel type to single point msmt
+        self.sweep_type('CW_Point')
+        # turn off average on the VNA since we want single point sweeps.
+        self.write(f'SENS{self._instrument_channel}:AVER:STAT OFF')
```
jenshnielsen, line 634: It seems better it this was a boolean paramter. Averaging_enabled or some such thing

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -419,6 +623,79 @@ def _get_sweep_data(self, force_polar: bool = False) -> np.ndarray:
...
+        # This format is required for getting both real and imaginary parts.
+        self.format('Complex')
+        # Set the sweep time to auto such that it sets the delay to zero
+        # between each point (e.g msmt speed is optimized). Note that if one
+        # would like to do a time sweep with time > npts/bandwidth, this is
+        # where the delay would be set, but in general we want to measure as
+        # fast as possible without artificial delays.
+        self.write(f'SENS{self._instrument_channel}:SWE:TIME:AUTO ON')
```
jenshnielsen, line 642: Same. This really should be a parameter

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -419,6 +623,79 @@ def _get_sweep_data(self, force_polar: bool = False) -> np.ndarray:
...
+        # Set cont measurement off here so we don't have to send that command
+        # while measuring later.
+        self.root_instrument.cont_meas_off()
+
+    def setup_lin_sweep(self) -> None:
+        """
+        Function in order to revert setup_CW_sweep and go back to lin sweep
+        mode.
```
astafan8, line 650: the docstring of this method does not need to know about "reverting" to some other mode etc etc - just keep it simple to what it does and that's it :) it will be less effort to maintain and understand. smth like this:
```suggestion
        Setup the instrument into linear sweep mode.
```

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -419,6 +623,79 @@ def _get_sweep_data(self, force_polar: bool = False) -> np.ndarray:
...
+        # It is possible that the instrument and QCoDeS disagree about
+        # which parameter is measured on this channel.
+        instrument_parameter = self.vna_parameter()
+        if instrument_parameter != self._vna_parameter:
+            raise RuntimeError("Invalid parameter. Tried to measure "
+                               f"{self._vna_parameter} "
+                               f"got {instrument_parameter}")
+
+        # Turn off average on the VNA since we want single point sweeps.
+        self.write(f'SENS{self._instrument_channel}:AVER:STAT OFF')
```
astafan8, line 677: as Jens mentioned above - this should be a parameter

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -419,6 +623,79 @@ def _get_sweep_data(self, force_polar: bool = False) -> np.ndarray:
...
+    def _get_cw_data(self, check_cw_sweep_first: bool = True) \
+            -> Tuple[np.ndarray, np.ndarray]:
+        # Make the checking optional such that we can do super fast sweeps as
+        # well, skipping the overhead of the other commands.
+        if check_cw_sweep_first:
+            self._check_cw_sweep()
+
+        with self.status.set_to(1):
+            self.write(f'INIT{self._instrument_channel}:IMM; *WAI')
+            data_str = self.ask(f'CALC{self._instrument_channel}:DATA? SDAT')
```
astafan8, line 692: is there a need to wrap this (i don't know which part though, the `*WAI` or `:DATA?``..) into `with self.root_insrtument.timeout.set_to(<smth>):` so that if taking the trace takes more time than the default VISA command timeout, VISA will patiently wait for the measurement and the data transfer to finish before raising a VisaTimeoutError? @LGruenhaupt @abargerbos

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -1,16 +1,143 @@
...
+    /ZNB_ZNBT_HTML_UserManual_en.htm) under GUI reference -> sweep softtool
+    -> sweep type tab -> CW mode
+    """
+
+    def __init__(self, name: str, instrument: "ZNBChannel",
+                 npts: int, bandwidth: int, channel: int,
+                 check_cw_sweep_first: bool = True) -> None:
+        super().__init__(name, names=("", ""), shapes=((), ()))
+        self._instrument = instrument
+        self.set_cw_sweep(npts, bandwidth)
```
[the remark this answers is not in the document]
Akshita07: no, they do not.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -436,6 +436,10 @@ def __init__(self, parent: 'ZNB', name: str, channel: int,
                            parameter_class=FixedFrequencyPointIQ)
         self.add_parameter(name='point_fixed_frequency_mag_phase',
                            parameter_class=FixedFrequencyPointMagPhase)
+        self.add_parameter(name='enable_averaging',
```
astafan8, line 439: https://qcodes.github.io/Qcodes/examples/writing_drivers/Creating-Instrument-Drivers.html#On/Off-parameters ;)

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -1,16 +1,143 @@
...
+        Gets the magnitude and phase of the mean of the raw real and imaginary
+        part of the data. If the flag `check_cw_sweep_first` is set to `True`
+        for the instrument then at the cost of a few ms overhead checks if
+        the vna is setup correctly.
+        """
+        assert isinstance(self.instrument, ZNBChannel)
+        i, q = self.instrument._get_cw_data(self._check_cw_sweep_first)
+        s = np.mean(i) + 1j*np.mean(q)
+        return np.abs(s), np.angle(s)
+
```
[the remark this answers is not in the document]
Akshita07: Yes, I think so too. FixedFrequencyPointMagPhase and FixedFrequencyPointIQ can inherit from FixedFrequencyTraceIQ.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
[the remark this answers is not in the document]
astafan8: sorry, i couldn't understand - so should they call this method as well or not? :)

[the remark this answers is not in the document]
Akshita07: I need to check the manual for the actual functioning.  But from the code it seems it's not been used for other parameters. Otherwise user code would have been giving errors already. Will check and update :)

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
Akshita07: thanks :)

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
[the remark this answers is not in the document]
GitJaap: This function set_cw_sweep is added similar to what the set_sweep function does in the regular frequency trace. the reason for this function is that if you change the parameter npts in the VNA or the bandwidth, the setpoints for this parameter have to change accordingly. 

For the other two averaged versions of this method it is not necessary since their setpoints are always just (),(). IT therefore shouldn't be called there.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -1,16 +1,143 @@
...
+    -> sweep type tab -> CW mode
+    """
+
+    def __init__(self, name: str, instrument: "ZNBChannel",
+                 npts: int, bandwidth: int, channel: int,
+                 check_cw_sweep_first: bool = True) -> None:
+        super().__init__(name, names=("", ""), shapes=((), ()))
+        self._instrument = instrument
+        self.set_cw_sweep(npts, bandwidth)
+        self._channel = channel
```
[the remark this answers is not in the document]
GitJaap: i think it is indeed unused, but then this holds also for the existing frequencysweep parameter of the VNA. I copied this in case it was used somewhere that was unclear. We can try to remove it and see.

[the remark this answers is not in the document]
astafan8: > in case it was used somewhere that was unclear

would the following approach work better: see if it is used for the case of this parameter, and based on that decide to keep it or not, and leave the existing frequencysweep parameter "as is"?

-------

copy-paste-driver-development is fine, as long as after the copy-paste the pasted code is actually refined, adapted, etc ;)

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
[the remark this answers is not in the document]
astafan8: > For the other two averaged versions of this method it is not necessary since their setpoints are always just (),()

aaa! thanks for the explanation! ( i totally misread the docstrings, my bad ).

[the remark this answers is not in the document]
astafan8: and this also means that the other comment suggesting an inheritance is wrong. inheritance is not the way here. but perhaps extracting the "meat" of get_raw into smth the instrument owns, could be an improvement.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
GitJaap: Agreed, added that https://github.com/QCoDeS/Qcodes/pull/2007/commits/0b021a7e72ea0bbdd1fa9e1335c5443fdb9b9e57

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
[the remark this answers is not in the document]
GitJaap: see the conversation above, it could in principle inherit from the FixedFrequencyTraceIQ, but actually most of the difference is contained in the setpoints set in the init. All shared functionality is already in the top instrument _get_cw_data function anyway so it think this is fine to keep it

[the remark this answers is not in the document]
astafan8: perfecto!

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -635,15 +643,15 @@ def setup_cw_sweep(self) -> None:
...
-        self.write(f'SENS{self._instrument_channel}:AVER:STAT OFF')
+        self.enable_averaging(False)
         # This format is required for getting both real and imaginary parts.
         self.format('Complex')
         # Set the sweep time to auto such that it sets the delay to zero
         # between each point (e.g msmt speed is optimized). Note that if one
         # would like to do a time sweep with time > npts/bandwidth, this is
         # where the delay would be set, but in general we want to measure as
         # fast as possible without artificial delays.
-        self.write(f'SENS{self._instrument_channel}:SWE:TIME:AUTO ON')
```
GitJaap: this is not the same as enable averaging right?

astafan8: indeed, it should've been a parameter of its own (and the comment above can move to be part this parameter's docstring)

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -1,16 +1,143 @@
...
+        self._instrument = instrument
+        self.set_cw_sweep(npts, bandwidth)
+        self._channel = channel
+        self._check_cw_sweep_first = check_cw_sweep_first
+        self.names = ('I', 'Q')
+        self.labels = (f'{instrument.short_name} I',
+                       f'{instrument.short_name} Q')
+        self.units = ('', '')
+        # TODO the question is if the time really corresponds to the time. The
+        # manual specifies there might be some additional overhead delay
```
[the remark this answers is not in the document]
GitJaap: https://github.com/QCoDeS/Qcodes/pull/2007/commits/065223a8165550af0cddf43f733c83be636fb08c

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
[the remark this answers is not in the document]
GitJaap: 1288fef20562c682f58fce256d64e976165ab054

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
GitJaap: yeah this could be done, i remember i tried to remove as much unnecessary commands out of the raw get in order to optimize speed of measurements (asking self.sweep_time() is another visa command), since this mode is used for very short measurements(us to ms)

GitJaap:
@astafan8  @Akshita07 @LGruenhaupt @abargerbos @jenshnielsen Fixed the issues mentioned above. I will test it on QT3 today if it isn't broken after all the additions and then it should be ready imo.

Thanks for the comments and efforts!

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
astafan8: @gitjaap so does this mean that it's not possible to set the bandwidth and npts to values that will make the [measurement] + [dataobtaining from the isntrument to PC] time get comparable to the default visa timeout (10 seconds, i think)?

GitJaap:
ValueError                                Traceback (most recent call last)
<ipython-input-117-cf0ea34e99af> in <module>
----> 1 vna = ZNB.ZNB('VNA', 'USB0::0x0AAD::0x01A1::101971::INSTR')
      2 vna.clear_channels()
      3 channel_name = 'S21'
      4 vna.add_channel(channel_name)
      5 vna.channels.S21.format('dB')

D:\Code\qcodes_dev\qcodes\instrument_drivers\rohde_schwarz\ZNB.py in __init__(self, name, address, init_s_params, reset_channels, **kwargs)
    827                 for j in range(1, num_ports + 1):
    828                     ch_name = 'S' + str(i) + str(j)
--> 829                     self.add_channel(ch_name)
    830             self.channels.lock()
    831             self.display_sij_split()

D:\Code\qcodes_dev\qcodes\instrument_drivers\rohde_schwarz\ZNB.py in add_channel(self, channel_name, **kwargs)
    845     def add_channel(self, channel_name: str, **kwargs: Any) -> None:
    846         i_channel = len(self.channels) + 1
--> 847         channel = self.CHANNEL_CLASS(self, channel_name, i_channel, **kwargs)
    848         self.channels.append(channel)
    849         if i_channel == 1:

D:\Code\qcodes_dev\qcodes\instrument_drivers\rohde_schwarz\ZNB.py in __init__(self, parent, name, channel, vna_parameter, existing_trace_to_bind_to)
    383                            npts=self.npts(),
    384                            channel=n,
--> 385                            parameter_class=FrequencySweepMagPhase)
    386         self.add_parameter(name='trace',
    387                            start=self.start(),

D:\Code\qcodes_dev\qcodes\instrument\base.py in add_parameter(self, name, parameter_class, **kwargs)
     99         if name in self.parameters:
    100             raise KeyError(f'Duplicate parameter name {name}')
--> 101         param = parameter_class(name=name, instrument=self, **kwargs)
    102         self.parameters[name] = param
    103 

D:\Code\qcodes_dev\qcodes\instrument_drivers\rohde_schwarz\ZNB.py in __init__(self, name, instrument, start, stop, npts, channel)
    157                              (f'{instrument.short_name} frequency',),
    158                          ),
--> 159                          shapes=((), (),))
    160         self.setpoint_names=(
    161             (f'{instrument.short_name}_frequency',),

D:\Code\qcodes_dev\qcodes\instrument\parameter.py in __init__(self, name, names, shapes, instrument, labels, units, setpoints, setpoint_names, setpoint_labels, setpoint_units, docstring, snapshot_get, snapshot_value, snapshot_exclude, metadata)
   1829 
   1830         metadata: Extra information to include with the
-> 1831             JSON snapshot of the parameter.
   1832     """
   1833 

ValueError: setpoint_labels must be a tuple of tuples of strings


I have tried to make the setpoints_labels a tuple of tuples of strings, but it doesn't work. I remember this from long ago and that might be the reason why i used to set them not in the super_init, but after the init was gone. @astafan8 @Akshita07 any fix? This breaks the driver right now.

astafan8:
@GitJaap instead of passing emppty shapes, perhaps, `shapes = ((npts,), (npts,))` needs to be passed?

GitJaap:
@astafan8 yup that worked!

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -375,6 +709,23 @@ def update_traces(self) -> None:
                 except AttributeError:
                     pass
 
+    def update_cw_traces(self) -> None:
+        """
+        Updates the bandwidth and npts of all fixed frequency (CW) traces.
+
+        TODO:
+        - This updating seems like a dirty solution looping over all possible
+        parameters.
```
astafan8, line 718: no value in this comment/todo here.
```suggestion
```

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
GitJaap: Hey its definitely possible, and the instrument will timeout at a time which is the vna.timeout(time), its just that i dont see a reason to change the timeout set by the user.

The reason is I for example sometimes added or forgot a zero in the npts and it suddenly takes a very long time. Then if the timeout is overwritten you also dont timeout before you see the result, isnt that why you have a timeout option?

Akshita07: I am on it.

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -397,28 +747,118 @@ def _get_sweep_data(self, force_polar: bool = False) -> np.ndarray:
...
+        Checks if all required settings are met to be able to measure in
+        CW_point mode. Similar to what is done in get_sweep_data
+        """
+        if self.sweep_type() != "CW_Point":
+            raise RuntimeError(
+                f"Sweep type is not set to continuous wave "
+                f"mode, instead it is: {self.sweep_type()}"
+            )
+
+        if not self._parent.rf_power():
```
astafan8, line 822: perhaps?
```suggestion
        if not self.root_instrument.rf_power():
```

astafan8 (approved):

PR #2008, opened 2020-05-18
Refactor parameter cache.get

jenshnielsen:
This IMHO makes the logic a bit more clear. Functinality it should be the same. 
I have changed a case in the if/else statments that should never happen to an error from silently returning a not up to date value. 

- [x] Should there be a warning if get_if_invalid=False and the cache is invalid? (mayby but not in this pr)


jenshnielsen:
@astafan8 I think this should implement the changes we talked about. Do you want to have a look

PR #2012, opened 2020-05-18
Add explicit method to Invalidate cache

jenshnielsen:
So that there is a public interface to mark the cache invalid and force a new call to get. Build on top 0f #2008 that should be merged first

TODO:

- [x] Tests
- [x] Docstring


qcodes/tests/parameter/test_parameter_cache.py
```diff
@@ -330,3 +330,29 @@ def test_set_latest_works_for_plain_memory_parameter(p, value, raw_value):
...
+
+    param.cache.invalidate()
+
+    param.cache.get(get_if_invalid=True)
+    assert param._get_count == 2
+
+    param.cache.invalidate()
+
+    param.cache.get(get_if_invalid=False)
+    assert param._get_count == 2
```
jenshnielsen, line 351: Is this really the behaviour we want. Should this raise or warn?

jenshnielsen:
@astafan8 I think this is ready for review

qcodes/instrument/parameter.py
```diff
@@ -2107,19 +2156,21 @@ def _timestamp_expired(self) -> bool:
     def get(self, get_if_invalid: bool = True) -> ParamDataType:
         """
         Return cached value if time since get was less than ``max_val_age``,
-        otherwise perform ``get()`` on the parameter and return result. A
+        or the parameter was explicitly marked invalid.
```
astafan8, line 2159:
```suggestion
        or the parameter's cache was explicitly marked invalid.
```

qcodes/instrument/parameter.py
```diff
@@ -2107,19 +2156,21 @@ def _timestamp_expired(self) -> bool:
...
         ``get()`` will also be performed if the parameter has never been
         captured but only if ``get_if_invalid`` argument is ``True``.
 
         Args:
             get_if_invalid: if set to ``True``, ``get()`` on a parameter
                 will be performed in case the cached value is invalid (for
-                example, due to ``max_val_age``, or because the parameter has
-                never been captured)
+                example, due to ``max_val_age``, because the parameter has
+                never been captured, or because the parameter was marked
```
astafan8, line 2168:
```suggestion
                never been captured, or because the parameter's cache was marked
```

qcodes/instrument/parameter.py
```diff
@@ -2038,6 +2071,21 @@ def max_val_age(self) -> Optional[float]:
...
+    def valid(self) -> bool:
+        """
+        Returns True if the cache is expected be be valid.
+        """
+        return not self._timestamp_expired() and self._marked_valid
+
+    def invalidate(self) -> None:
+        """
+        Call this method to mark the cache invalid.
+        If the cache is invalid the next call to `cache.get()` attempt
```
astafan8, line 2084:
```suggestion
        If the cache is invalid the next call to `cache.get()` will attempt
```

qcodes/instrument/parameter.py
```diff
@@ -2038,6 +2071,21 @@ def max_val_age(self) -> Optional[float]:
         """
         return self._max_val_age
 
+    @property
+    def valid(self) -> bool:
+        """
+        Returns True if the cache is expected be be valid.
```
astafan8, line 2077: "expected" to? can't we just say "if the cache is valid"?
```suggestion
        Returns True if the cache is expected to be valid.
```

qcodes/instrument/parameter.py
```diff
@@ -1387,6 +1388,30 @@ def timestamp(self) -> Optional[datetime]:
                 return None
             return self._parameter.source.cache.timestamp
 
+        @property
+        def valid(self) -> bool:
```
astafan8, line 1392: shouldn;t the delegate cache totally forward this to the source parameter? and if there's no source parameter, then just be "valid" all the time? yeah, the meaning of "valid" is not clear when there's no source parameter... perhaps "valid" is a safer option for now

qcodes/tests/parameter/test_delegate_parameter.py
```diff
@@ -460,3 +459,20 @@ def test_delegate_parameter_fixed_label_unit_unchanged():
...
+    d = DelegateParameter('test_delegate_parameter', p,
+                          initial_cache_value=value)
+    assert p._get_count == 0
+    assert d.cache.get() == value
+    assert p._get_count == 0
+    d.cache.invalidate()
+    p.cache.invalidate()
+    # since both the cache or the
+    # delegate and param is marked invalid
+    # this should trigger a get on the parameter
```
astafan8, line 476: shouldn't it trigger from a single invalidate() call on either of the parameters? delegate cache is just proxying to the soruce pararmater cache, right?

qcodes/tests/parameter/test_parameter_cache.py
astafan8: are you asking if we should differentiate between the explicit invalidation via `invalidate` call and other invalidations like the max_val_age? i'd say that for the clean-ness of the interface, if cache is invalid, it is simply invalid - hence i agree with this test.

qcodes/instrument/parameter.py
jenshnielsen: Yes not sure why I did it like this originally

jenshnielsen:
@astafan8 ready for another review

astafan8 (approved):

PR #2015, opened 2020-05-19
make it an error to supply get/set_cmd if get/set_raw is not abstract

jenshnielsen:
It makes no sense to both implement get_raw and supply a get_cmd. Which one should be used. So lets make this an error. There are already 2 subclasses that does the same. DelegateParameter and GroupParameter

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -883,6 +883,10 @@ class Parameter(_BaseParameter):
        :meth:`set_raw` methods are automatically wrapped to provide ``get`` and
        ``set``.
 
+    It is an error to do both 1 and 2. E.g supply a ``get_cmd``/``set_cmd``
+     and implement ``get_raw``/``set_raw``
+
+
     To detect if a parameter is gettable or settable check the attributes
     ``gettable`` and ``settable`` on the parameter.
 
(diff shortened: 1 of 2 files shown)


qcodes/instrument/parameter.py
```diff
@@ -1020,8 +1024,11 @@ def __init__(self, name: str,
             self.gettable = True
             self.get = self._wrap_get(self.get_raw)
 
-
-        if not hasattr(self, 'set') and set_cmd is not False:
+        if self.settable and set_cmd is not None:
```
jenshnielsen, line 1027: We probably need to allow set_cmd is False too to allow the default value to subclass

qcodes/tests/test_parameter.py
```diff
@@ -72,6 +72,19 @@ def set(self, value):
...
+    """ Parameter that implements get and set raw"""
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def get_raw(self):
+        return self.cache._raw_value
+
+    def set_raw(self, raw_value):
+        value = self._from_raw_value_to_value(raw_value)
+        self.cache.set(value)
```
astafan8, line 85:
```suggestion
        pass
```

PR #2016, opened 2020-05-20
Prevent save config to home in tests

GateBuilder:
This PR removes redundant and unnecessary execution of `save_to_home` function in tests. 
The functionality of the corresponding context managers are not effected and they do not need this command. Although I tested their functionality without this attribute, please confirm that this is indeed the case.

@jenshnielsen


qcodes/tests/dataset/test_database_creation_and_upgrading.py
```diff
@@ -36,9 +35,7 @@
 from qcodes.dataset.sqlite.queries import get_run_description, update_GUIDs
 from qcodes.dataset.sqlite.query_helpers import is_column_in_table, one
 from qcodes.tests.common import error_caused_by
-from qcodes.tests.dataset.temporary_databases import (empty_temp_db,
-                                                      experiment,
-                                                      temporarily_copied_DB)
```
jenshnielsen, line 41: Because of the way that pytest fixtures work all of these imports are needed

jenshnielsen:
@GateBuilder I suspect but I am not sure that these context managers does this because there is a bunch of places in the code where a new config is generated by calling `Config()` these really should read the config from the default qcodes.config instead.  We may need to change those. The way to check I think is to see if the location and work_station not set warning is triggered when running the tests after all the tests passes

jenshnielsen:
For refererence the config instance creation below all look suspicious 

```
qcodes\tests\test_config.py
177:        qcodes.config = Config()
224:        self.conf = Config()
322:        cfg = Config()
341:    cfg = Config()
363:        cfg = Config()

qcodes\tests\dataset\test_guids.py
24:    ocfg: DotDict = Config().current_config
30:        cfg = Config()
41:        cfg = Config()
66:    orig_cfg = Config().current_config
73:        cfg = Config().current_config
86:    orig_cfg = Config().current_config
93:        cfg = Config().current_config
129:        cfg = Config()

qcodes\tests\dataset\test_database_creation_and_upgrading.py
51:    cfg = Config()

qcodes\tests\dataset\test_database_extract_runs.py
514:    cfg = Config()

qcodes\dataset\sqlite\queries.py
1596:    cfg = Config()

qcodes\dataset\guids.py
28:    cfg = Config()
84:    cfg = Config()
109:    cfg = Config()
```

jenshnielsen:
These calls to save_to_home could also be problematic if they are called by tests 

``
qcodes\dataset\guids.py
102:    cfg.save_to_home()
128:    cfg.save_to_home()
```

**Comment by jenshnielsen** [comment]
@GateBuilder great lets hope that is the case for all of them :)

**Comment by jenshnielsen** [comment]
I think we should land this as is if passing and then revisit the remaining ones

File: qcodes/tests/dataset/test_database_creation_and_upgrading.py [hunk 2]
```diff
@@ -48,19 +47,16 @@
...
+    old_cfg = deepcopy(cfg)
     cfg['GUID_components']['location'] = location
     cfg['GUID_components']['work_station'] = work_station
-    cfg.save_to_home()
 
     try:
         yield
 
     finally:
         cfg.current_config = old_cfg
```
jenshnielsen, line 59: This is actually wrong. Not a bug introduced here but exposed by rewriting. The `old_cfg` is a `Class Config` but `cfg.current_config` is a `DotDict`

qcodes/tests/dataset/test_database_creation_and_upgrading.py
```diff
@@ -616,25 +613,17 @@ def test_update_existing_guids(caplog):
             for record, lvl in zip(caplog.records, expected_levels):
```
jenshnielsen, line 613: Then we just need a `print(record)` here

qcodes/dataset/sqlite/queries.py
```diff
@@ -1593,7 +1593,7 @@ def update_GUIDs(conn: ConnectionPlus) -> None:
 
     log.info('Commencing update of all GUIDs in database')
 
-    cfg = Config()
+    cfg = qc.config.current_config
```
jenshnielsen, line 1596: This should just be `qc.config`

qcodes/dataset/sqlite/queries.py
```diff
@@ -1593,7 +1593,10 @@ def update_GUIDs(conn: ConnectionPlus) -> None:
 
     log.info('Commencing update of all GUIDs in database')
 
-    cfg = Config()
+    cfg = qc.config
+
+    if cfg is None:
```
jenshnielsen, line 1598: I don't think this is needed. `qc.config.current_config` can be None since it has the type `Optional[DotDict]` but qc.config is always a `Config` object

GateBuilder: Yes I forgot to remove that. I will push once the local run of tests finish.

PR #2023, opened 2020-05-27
B1500: Transient fix for converting str as '-0.-1' to float 

lakhotiaharshit:
Keysight B1500 responds with a string containing decimal numbers but with signs on both sides of the decimal. For ex: '-0.-1' this PR add function to covert such strings to float.

Once this is fixed in Keysight firmware the commits in this PR can be reverted.

### DIFF
--- a/qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
+++ b/qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
@@ -140,6 +140,20 @@ def parse_dcorr_query_response(response: str) -> _DCORRResponse:
                           secondary=float(secondary))
 
 
+def fixed_negative_float(response: str) -> float:
+    """
+    Keysight sometimes responds for ex. '-0.-1' as an output when you input
+    '-0.1'. This fixed_float function can convert such strings also float.
+    """
+    if len(response.split('.')) != 2:
+        raise Exception('String must of format a.b')
+
+    number, decimal = response.split('.')
+    decimal = decimal.replace("-", "")
+    output = ".".join([number, decimal])
+    return float(output)
+
+
 _dcorr_labels_units_map = {
     constants.DCORR.Mode.Cp_G: dict(
         primary=dict(label='Cp', unit='F'),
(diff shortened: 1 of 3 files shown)


qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1520A.py
```diff
@@ -598,8 +598,8 @@ def _get_sweep_steps_parser(response: str) -> Dict[str, Union[int, float]]:
         out_dict = cast(Dict[str, Union[int, float]], out_str)
```
jenshnielsen, line 598: I would make outdict a typed_dict with actual types of the 5 different fields. 

Alternatively just avoiding copying out_str to out_dict would also make it cleaner and fix the mypy issue
```
out_dict: Dict[str, Union[int, float]] = {}
out_dict['chan'] = int(out_str['chan'])
...
out_dict['sweep_start'] = fixed_negative_float(out_str['sweep_end'])
...
```

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
```diff
@@ -140,6 +140,20 @@ def parse_dcorr_query_response(response: str) -> _DCORRResponse:
                           secondary=float(secondary))
 
 
+def fixed_negative_float(response: str) -> float:
+    """
+    Keysight sometimes responds for ex. '-0.-1' as an output when you input
+    '-0.1'. This fixed_float function can convert such strings also float.
+    """
+    if len(response.split('.')) != 2:
+        raise Exception('String must of format a.b')
```
jenshnielsen, line 149: Are we sure that this will never have to handle an integer string? I could imagine that comming from the keysight too.

jenshnielsen: This should probably return a ValueError to match `float` ?

```
In [3]: float("a")
ValueError                                Traceback (most recent call last)
<ipython-input-3-a1ff3dbf6f4f> in <module>
----> 1 float("a")

ValueError: could not convert string to float: 'a'
```

PR #2025, opened 2020-05-29
Add Zurich Instruments drivers 

YakBizzarro:
This PR add support for Zurich Instruments devices as drivers inside of QCodes.

The actual implementation is in https://github.com/zhinst/zhinst-qcodes/ as described in
https://blogs.zhinst.com/andrea/2020/05/24/control-your-measurements-with-qcodes-and-labber/
Here we propose to add stubs in Qcodes to help user to find and install such drivers.

We propose deprecation of the existing ZIUHF, since its functionality is replaced by this PR.

A new optional installation option, 'ZurichInstruments' is introduced. That add the package zhinst-qcodes as optional dependency of Qcodes, if the user wish so. 

@jenshnielsen

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,8 @@ extras = {
     'MatPlot': ('matplotlib', '2.2.3'),
     'QtPlot': ('pyqtgraph', '0.10.0'),
     'coverage tests': ('coverage', '4.0'),
-    'Slack': ('slacker', '0.9.42')
+    'Slack': ('slacker', '0.9.42'),
+    'ZurichInstruments': ('zhinst-qcodes', '0.1.1')
 }
 extras_require = {k: '>='.join(v) for k, v in extras.items()}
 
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -14,7 +14,8 @@ def readme():
     'MatPlot': ('matplotlib', '2.2.3'),
     'QtPlot': ('pyqtgraph', '0.10.0'),
     'coverage tests': ('coverage', '4.0'),
-    'Slack': ('slacker', '0.9.42')
+    'Slack': ('slacker', '0.9.42'),
+    'ZurichInstruments': ('zhinst-qcodes', '0.1.1')
```
jenshnielsen, line 18: I would prefer not to depend on the ZI modules. There are many users that do not currently use ZI drivers. Can we perhaps replace the raise of an Import error with a warning.

YakBizzarro: This is an optional dependency, it's installed only if the user ask explicitly for it with something like 'pip install qcodes[ZurichInstrument]'. There is anyway an import check in the individual drivers.

jenshnielsen: Sorry I was too quick. This does indeed look fine to me

jenshnielsen:
@astafan8 Please have a look

YakBizzarro:
thank you!

PR #2026, opened 2020-05-29
make sure that new lines are handled correctly on windows

jenshnielsen:
As suggested here https://www.edwardthomson.com/blog/git_for_windows_line_endings.html
I had this problem again for some reason on #2005

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 qcodes/_version.py export-subst
+* text=auto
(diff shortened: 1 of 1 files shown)


jenshnielsen:
As suggested here https://jessitron.com/2019/11/11/line-endings-in-git/ its probably best to not convert bat and cmd file like the build script for the docs

jenshnielsen:
For future reference I had somehow manged to check the file with Windows line ending. It was not possible to convert it until I explicitly disabled safecrlf

PR #2027, opened 2020-05-29
B1500 testing improvements

astafan8:
Small improvements, mostly related to testing of the driver:
* Use pyvisa-sim to test error handling of B1500
* Add dialogs to B1500 yaml for commands sent in the B1500 tests
* Use pytest finalizer to close instrument instances in B1500 tests
* Remove unused attribute


qcodes/tests/drivers/keysight_b1500/b1500_driver_tests/test_b1500.py
```diff
@@ -199,10 +199,5 @@ def test_self_calibration_failed(b1500):
 
 
 def test_error_message(b1500):
-    mock_ask = MagicMock()
-    b1500.ask = mock_ask
-    mock_ask.return_value = '0,"No Error."'
-
     response = b1500.error_message()
     assert '0,"No Error."' == response
-    mock_ask.assert_called_once_with(f'ERRX?')
```
jenshnielsen, line 208: I guess we are ok with loosing the counter on ask being called?

qcodes/instrument/sims/keysight_b1500.yaml
```diff
@@ -64,6 +64,13 @@ devices:
         r: "B1517A,0;B1517A,0;B1520A,0;0,0;0,0;B1530A,0;0,0;0,0;0,0;0,0"
       - q: "TI 1,-16"
         r: "NAI+000.000E-06"
+      - q: "WTDCV 0,0,0,0,0"
+      - q: "WDCV 3,1,0,0,1"
+      - q: "ACT 2,1"
+      - q: "IMP 101"
+      - q: "LMN 0"
+      - q: "MT 0,0.1,7"
+      - q: "FMT 1,0"
```
jenshnielsen, line 73: How did it work before without these?

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1520A.py
```diff
@@ -814,11 +814,12 @@ def __init__(self, name, instrument, **kwargs):
             setpoint_units=(('V',),) * 2,
             **kwargs)
         self._instrument = instrument
-        self.data = _FMTResponse(None, None, None, None)
+
         self.param1 = _FMTResponse(None, None, None, None)
```
jenshnielsen, line 818: I just learned that instance attributes can be documented like this https://github.com/QCoDeS/Qcodes/pull/2005/files#diff-7ab35e1b5092c8dd21f4bb9883f61a76R1094 should we document these 4?

qcodes/tests/drivers/keysight_b1500/b1500_driver_tests/test_b1500.py
```diff
@@ -16,7 +16,9 @@
 
 
 @pytest.fixture
-def b1500():
+def b1500(caplog, request):
```
jenshnielsen, line 19: What is the reason for using the caplog fixture?

qcodes/instrument/sims/keysight_b1500.yaml
astafan8: since these are called with `write`, nothing was "returned", so from `.write(` call point of view everything was fine. However, since these commands were not known to the yaml file, they generated "command error" errors in the error queue of the visa resource (which mimics what the instrument does as well), so if you were to check the error queue after you "write", then you would see the "problem". and the latter we plan to do as a later improvememnt.

qcodes/tests/drivers/keysight_b1500/b1500_driver_tests/test_b1500.py
astafan8: yep. not only the error_message should just work, but init of the instrument (via fixture) should not accumulate errors in the queue

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1520A.py
astafan8: sure

PR #2031, opened 2020-06-03
Group Parameter improvements 

jenshnielsen:
Replaces #1995

* Correctly keeps cache up to date when getting and setting the group parameters
* Respect mapping between raw_value and value for groupparameters


qcodes/instrument/group_parameter.py
```diff
@@ -55,23 +55,18 @@ def __init__(self,
         self._initial_value = initial_value
```
jenshnielsen, line 55: `self.group` should be a read only property

qcodes/instrument/group_parameter.py
```diff
@@ -210,27 +204,27 @@ def parser(ret_str: str) -> Dict[str, Any]:
 
         return parser
 
-    def set(self, set_parameter: GroupParameter, value: Any) -> None:
+    def set(self, set_parameter: GroupParameter, raw_value: Any) -> None:
         """
         Sets the value of the given parameter within a group to the given
```
jenshnielsen, line 209: I think this method is super confusing. It does not actually set all of the group and it takes a raw value which is confusing. I vote for deprecating it and replacing with a private method that does the same.

astafan8: wholeheartedly support!

qcodes/instrument/group_parameter.py
astafan8: agree!

qcodes/instrument/group_parameter.py
```diff
@@ -160,19 +163,19 @@ def __init__(self,
...
-        self.instrument = parameters[0].root_instrument
+        self._instrument = parameters[0].root_instrument
 
-        self.set_cmd = set_cmd
-        self.get_cmd = get_cmd
+        self._set_cmd = set_cmd
+        self._get_cmd = get_cmd
 
         if get_parser:
             self.get_parser = get_parser
```
jenshnielsen, line 181: Hmn `get_parser` and `set_parser` should mayby also be private perhaps with a read only property. However they are exposed as attributes on `Parameter` so I have left them the same here

jenshnielsen:
@lakhotiaharshit and @astafan8 This exposes some inconsitencies in the b15xx drivers especially around the handling of float/int values. After discussing with @lakhotiaharshit I have only done the minimal changes here to fix the tests but it seems likely that more work is needed

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1520A.py
```diff
@@ -63,7 +63,7 @@ def __init__(self, parent: 'B1520A', name: str, **kwargs: Any):
         self.post_sweep_voltage_condition.cache.set(constants.WMDCV.Post.START)
 
         self.add_parameter(name='hold',
-                           initial_value=0,
+                           initial_value=0.0,
```
jenshnielsen, line 66: The tests expect these values to be 0.0 initially so we change them to this

qcodes/tests/drivers/keysight_b1500/b1500_driver_tests/test_b1520a_cmu.py
```diff
@@ -226,8 +226,8 @@ def test_run_sweep(cmu):
     cmu.adc_coef(5)
     cmu.run_sweep()
     mainframe.write.assert_has_calls([
-        call('WDCV 3,1,-1.0,1.0,5'),
-        call('WDCV 3,1,-1.0,1.0,5'),
```
jenshnielsen, line 230: I am not sure how this ever worked before. The value of 1.0 for end is set in the second call and should therefor not be used in the first call. Rather the default value of 0.0 should be used. Similar for the value of 5.0 which is set in the third call

qcodes/instrument/group_parameter.py
astafan8: > I have left them the same here

good

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1520A.py
astafan8: ok

qcodes/tests/drivers/keysight_b1500/b1500_driver_tests/test_b1520a_cmu.py
astafan8: > I am not sure how this ever worked before

me neither :( but great catching all of this!

lakhotiaharshit: Earlier implementation was definitely wrong. I had missed it and the test passed by fluke. Now, this is correct. Thanks for catching.

PR #2032, opened 2020-06-03
Use in memory config and do not save to home in guids

GateBuilder:
This PR removes the usage of `Config()` in favor of `qcode.config` and removes `save_to_home()` calls in guids.py. This PR is necessary for [#2016](https://github.com/QCoDeS/Qcodes/pull/2016) to pass. With these changes test_database_creation_and_upgrading.py passes with and without xdist-parallel without the need of a temp config.

@jenshnielsen

### DIFF
--- a/qcodes/dataset/guids.py
+++ b/qcodes/dataset/guids.py
@@ -99,6 +99,7 @@ def set_guid_location_code() -> None:
         return
 
     cfg['GUID_components']['location'] = location
+    cfg.save_to_home()
 
 
 def set_guid_work_station_code() -> None:
@@ -124,6 +125,7 @@ def set_guid_work_station_code() -> None:
         return
 
     cfg['GUID_components']['work_station'] = work_station
+    cfg.save_to_home()
 
 
 def filter_guids_by_parts(guids: Sequence[str],
--- a/qcodes/dataset/guids.py
+++ b/qcodes/dataset/guids.py
@@ -4,7 +4,7 @@ import re
 
 import numpy as np
 
-from qcodes.configuration import Config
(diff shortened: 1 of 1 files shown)


qcodes/dataset/guids.py
```diff
@@ -99,14 +99,13 @@ def set_guid_location_code() -> None:
         return
 
     cfg['GUID_components']['location'] = location
-    cfg.save_to_home()
```
jenshnielsen, line 102: This one should really call save_to_home as it is meant to run interactively when the user sets the config

qcodes/dataset/guids.py
```diff
@@ -81,7 +81,7 @@ def set_guid_location_code() -> None:
     """
     Interactive function to set the location code.
     """
-    cfg = Config()
```
jenshnielsen, line 84: Good

qcodes/dataset/guids.py
```diff
@@ -125,7 +124,6 @@ def set_guid_work_station_code() -> None:
         return
 
     cfg['GUID_components']['work_station'] = work_station
-    cfg.save_to_home()
```
jenshnielsen, line 128: Same here I think

qcodes/dataset/guids.py
astafan8: yes

qcodes/dataset/guids.py
astafan8: yes

jenshnielsen:
@GateBuilder Yes that sounds good

PR #2036, opened 2020-06-09
use pyqtgraph 0.11 in recommended env

jenshnielsen:
Fixes deprecation warnings with 0.10 among other things


jenshnielsen:
> 
> 
> shall we change it in setup.py extras as well?

Yes.

PR #2037, opened 2020-06-09
ensure that links and dataset does not share guid before testing

jenshnielsen:
In `test_parent_dataset_links_invalid_input` we have seen ocational falilures since the merge of #2016 
I suspect this is due to the links and dataset being created within the same ms and therefore having the same guid. This extra assertion will confirm or disprove this hypothesis

### DIFF
--- a/qcodes/tests/dataset/test_dataset_basic.py
+++ b/qcodes/tests/dataset/test_dataset_basic.py
@@ -680,6 +680,9 @@ def test_parent_dataset_links_invalid_input():
 
     ds = DataSet()
 
+    for link in links:
+        assert link.head != ds.guid
+
     match = re.escape('Invalid input. Did not receive a list of Links')
     with pytest.raises(ValueError, match=match):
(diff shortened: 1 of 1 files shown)


qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -680,6 +680,9 @@ def test_parent_dataset_links_invalid_input():
 
     ds = DataSet()
 
+    for link in links:
+        assert link.head != ds.guid
```
astafan8, line 684: an idea - could this be biting us https://github.com/QCoDeS/Qcodes/blob/55a8d16bf479e2210833163d664b9ce6cd6cd17a/qcodes/dataset/guids.py#L40 ?

jenshnielsen: Yes that is indeed what I suspect :) Just want to add this to confirm that the issue is indeed identical guids

PR #2038, opened 2020-06-09
generate links with a timestamp in the past to avoid collisions

jenshnielsen:
https://travis-ci.com/github/QCoDeS/Qcodes/jobs/346553423 confirms that the issue is indeed due to timestamp collisions so lets ensure that links are generated with a timestamp in the past

### DIFF
--- a/qcodes/tests/dataset/test_links.py
+++ b/qcodes/tests/dataset/test_links.py
@@ -1,5 +1,7 @@
 import re
 import json
+from datetime import datetime, timedelta
+import random
 from typing import List
 
 import pytest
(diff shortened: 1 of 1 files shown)


qcodes/tests/dataset/test_links.py
```diff
@@ -16,15 +18,24 @@ def generate_some_links(N: int) -> List[Link]:
     """
     Generate N links with the same head
     """
+    def _timestamps() -> int:
```
astafan8, line 21:
```suggestion
    def _timestamp() -> int:
```

PR #2040, opened 2020-06-10
Run tests in parallel

GateBuilder:
This PR introduces changes to run test suit in parallel with pytest-xdist.

@jenshnielsen


jenshnielsen:
Please write a note in the changelog that the math module has moved

jenshnielsen:
There are some issues with the code coverage that we need to fix

.travis.yml
```diff
@@ -71,7 +71,8 @@ script:
       pip uninstall -y qcodes
       pip install .
     - cd qcodes
-    - py.test --cov=qcodes --cov-report xml --cov-config=.coveragerc
+    - py.test -m "not serial" --cov=qcodes --cov-report xml --cov-config=.coveragerc
+    - py.test -n 1 -m serial --cov=qcodes --cov-report xml --cov-config=.coveragerc
```
jenshnielsen, line 75: You probably also need to change the file names here and below (line 102)

jenshnielsen:
@astafan8 Do you want to have a final look?

PR #2041, opened 2020-06-10
Improved error handling of B1500 driver

lakhotiaharshit:
Each time a parameter is set the instrument is queried for errors. If an error is found user is presented with a runtime error with an error message.


qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_base.py
```diff
@@ -64,6 +63,17 @@ def __init__(self, name, address, **kwargs):
 
         self.connect_message()
 
+    def write(self, cmd):
```
astafan8, line 66: anything we need to do for "ask" calls?

lakhotiaharshit: I think no. 1) Will only be dangerous only if instrument throws sensible value but store some error in the buffer. For all the measurements we have written, we use 'XE' command. Its output already contains which data is normal "N" and which one has problems by storing strings such as "G", "V" ... in compliance setting. Hence from the measurement side, we are safe. From the setting point of view, we are already checking the error while setting it. So maybe we are ok if we do not check it again while trying to read the settings.

astafan8: great!

qcodes/instrument/sims/keysight_b1500.yaml
```diff
@@ -4,7 +4,7 @@ devices:
     error:
       error_queue:
         - q: 'ERRX?'
-          default: '0,"No Error."'
+          default: '+0,"No Error."'
           command_error: '1,"Command error"'
```
astafan8, line 8:
```suggestion
          command_error: '+1,"Command error"'
```

jenshnielsen:
remember to backport this to the release branch

jenshnielsen:
I created a label for that in case there are more

lakhotiaharshit:
> great! (waiting for last confirmation of success of last testing on the instrument )

The test passed. We successfully ran a CV mobility protocol. Also, did some tests to check if the error message is thrown with wrong settings. Good to go!

PR #2043, opened 2020-06-11
Add pytest sugar for nicer test output

jenshnielsen:
I am not sure how well this works in CI but it gives much better output locally. So perhaps it also help here?


jenshnielsen:
Hmn travis makes a mess of the output and pipelines does not use it. So lets just stick with installing this locally if you like

PR #2045, opened 2020-06-12
Move the last few tests out of test_parameter

jenshnielsen:
This shoud be the last pr for now. As some point we could convert the remaining tests to pytest from unittest testcases but that has lower priority


jenshnielsen:
For whatever reason the webhook for pipelines does not work correctly but the job did run and the tests are passing https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=10756&view=results

PR #2047, opened 2020-06-12
B1520A: CVSweepMeasurement class works with all impedance models

lakhotiaharshit:
Earlier driver supported only Cp_D (Capacitance, Dissipation) impedance model. This was hardcoded in the driver. This PR removes these constraints and supports all possible impedance models.

@astafan8 @jenshnielsen


qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1520A.py
```diff
@@ -840,13 +841,15 @@ def __init__(self, name: str, instrument: B1520A, **kwargs):
...
-            raise Exception('Run sweep only supports Cp_D impedance model')
-
         if not self.instrument.setup_fnc_already_run:
             raise Exception('Sweep setup has not yet been run successfully')
 
+        model = self.instrument.impedance_model()
+        names, labels, units = get_name_label_unit_of_impedance_model(model)
+        self.names = names
+        self.labels = labels
+        self.units = units
```
astafan8, line 851: this also works
```suggestion
        self.names, self.labels, self.units = get_name_label_unit_of_impedance_model(model)
```

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1520A.py
```diff
@@ -840,13 +841,15 @@ def __init__(self, name: str, instrument: B1520A, **kwargs):
...
     def get_raw(self) -> Tuple[Tuple[float, ...], Tuple[float, ...]]:
-        model = self.instrument.impedance_model()
-        if model != constants.IMP.MeasurementMode.Cp_D:
-            raise Exception('Run sweep only supports Cp_D impedance model')
-
         if not self.instrument.setup_fnc_already_run:
             raise Exception('Sweep setup has not yet been run successfully')
 
+        model = self.instrument.impedance_model()
+        names, labels, units = get_name_label_unit_of_impedance_model(model)
```
astafan8, line 848: setting name/label/unit in get is too late - users need the correct name/label/unit when they register the parameter. I'd suggest for this multiparameter to have a method along the lines of "update_name_label_unit_of_impedance_model" which will be called once "impedance_mode" value is updated (and the same method can be called on the __init__ of the multiparameter).

This is also needs to be noted in the example notebook - that impedance model has to be set BEFORE registering this parameter.

qcodes/instrument_drivers/Keysight/keysightb1500/constants.py
```diff
@@ -618,6 +618,42 @@ class MeasurementMode(IntEnum):
...
+        Lp = 'H'
+        Ls = 'H'
+        X = 'ohm'
+        B = 'S'
+        THETA_RAD = 'radian'
+        THETA_DEG = 'degree'
+        D = 'unit'
+        Q = 'unit'
+        Rp = 'ohm'
+        Rs = 'ohm'
```
astafan8, line 655: keep forgetting - are using "ohm" or "ohms" across qcodes, in other drivers?

qcodes/instrument_drivers/Keysight/keysightb1500/constants.py
```diff
@@ -618,6 +618,42 @@ class MeasurementMode(IntEnum):
...
+        Cp = 'F'
+        Cs = 'F'
+        Lp = 'H'
+        Ls = 'H'
+        X = 'ohm'
+        B = 'S'
+        THETA_RAD = 'radian'
+        THETA_DEG = 'degree'
+        D = 'unit'
+        Q = 'unit'
```
astafan8, line 653: unit-less parameters better have empty string as a unit
```suggestion
        D = ''
        Q = ''
```

lakhotiaharshit: That leaves things ambiguous. Was unit known and not set or was it really unitless? I personally prefer having 'unit' if things are unitless and 'arb. unit' if things are an arbitrary unit. If you strongly feel otherwise, I can change it to ''

qcodes/instrument_drivers/Keysight/keysightb1500/constants.py
lakhotiaharshit: I saw some drivers using ohms. I will change it to ohms then. But did not understand why ohm(s) was used in other drivers?

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
```diff
@@ -187,6 +187,26 @@ def format_dcorr_response(r: _DCORRResponse) -> str:
...
+    param2 = '_'.join(params[1:])
+
+    label = (constants.IMP.Name[param1].value,
+             constants.IMP.Name[param2].value)
+
+    unit = (constants.IMP.Unit[param1].value,
+            constants.IMP.Unit[param2].value)
+
+    name = ('_'.join(label[0].lower().split(' ')),
+            '_'.join(label[1].lower().split(' ')))
```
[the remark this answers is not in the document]
lakhotiaharshit: done

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1520A.py
lakhotiaharshit: done

qcodes/instrument_drivers/Keysight/keysightb1500/constants.py
astafan8: > That leaves things ambiguous

agree. However, using `''` for unit-less parameter is consistent with other drivers and the parameter's docstring itself:

```
#: The unit of measure. Use ``''`` for unitless.
self.unit = unit if unit is not None else ''
```

So i believe we have to use `''` to be consistent.

qcodes/instrument_drivers/Keysight/keysightb1500/constants.py
astafan8: > why ohm(s) was used in other drivers

not sure i understand the question. some instruments do measure resistance or allow to set smth such as a resistance. wrt ohm vs ohms vs Ohm vs Ohms vs <big omega> - i don't know, qcodes doesn't dictate a standard for the units yet.

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1520A.py
lakhotiaharshit: Please have a look on the current implementation.

lakhotiaharshit:
> Looks great! one question - does the output of the "capacitance" parameter also change with the change of the impedance model? if so, shall we also update the label/unit of that parameter upon model change? or we just restrict that parameter to the Cp_D model and do the enhancement in another PR? (i think i like the latter for now)

I think this is not the part of this PR. Even before this PR user can change the impedance model and get the capacitance; capacitance parameter was not restrained to Cp_D. I think this whole problem should be fixed in a separate PR.

PR #2058, opened 2020-06-22
Remove deprecated default measurement from station

jenshnielsen:

### DIFF
--- a/qcodes/loops.py
+++ b/qcodes/loops.py
@@ -7,17 +7,11 @@ The general scheme is:
 delays
 
 2. activate the loop (which changes it to an ActiveLoop object),
-or omit this step to use the default measurement as given by the
-Loop.set_measurement class method.
 
 3. run it with the .run method, which creates a DataSet to hold the data,
 and defines how and where to save the data.
 
 Some examples:
-
-- set default measurements for later Loop's to use
-
->>> Loop.set_measurement(param1, param2, param3)
 
 - 1D sweep, using the default measurement set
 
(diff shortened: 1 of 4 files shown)


qcodes/loops.py
```diff
@@ -242,20 +224,6 @@ def validate_actions(*actions):
                             'objects. `Loop` objects are OK too, except in '
                             'Station default measurements.')
 
-    def run(self, *args, **kwargs):
-        """
```
jenshnielsen, line 246: Unfortunatly run and run_temp on a loop (as opposed to the activeloop) were not deprecated along with default measurements. They should however not be possible to use without getting a deprecation warning. Should we deprecated these first and then remove or is this fine?

astafan8: at a first glance i'd agree with "removing these methods right now in this PR".

at a second glance, i wonder if we can keep these run/run_temp methods - so far as i see in order to do that we need to find a new place for "default measurement", i think a class variable on the Loop class would do the job (and since none of this is documented, i guess, it's not a big deal to implement). Another approach would be to pull the `*default` into args of the methods. Anyway, of course, the main question is "who will benefit" from keeping these two methods, which i don't have a good answer to.

astafan8: ok, read a bit around, ActiveLoop is actually the one that is being "run", so indeed, we can remove these methods safely without breaking anyone's workflows.

jenshnielsen: I did actually try putting them on the loop. That works for the 1D case but changes the semantics for the 2d and higher case since it is unclear which of the loops you should bind the default measurements to.

PR #2062, opened 2020-06-23
Add cache reprecentation of dataset

jenshnielsen:
Very much work in progress.


qcodes/dataset/data_set.py
```diff
@@ -282,6 +282,8 @@ def __init__(self, path_to_db: str = None,
         self._interdeps: InterDependencies_
         self._parent_dataset_links: List[Link]
         self._data_write_queue: Queue = Queue()
+        self.data = []
```
ianhuezo, line 285: I noticed that the notebook shows the type as a list into a dict(after the value is cached).  Is this intentional for ordering at the beginning?

qcodes/dataset/data_set.py
```diff
@@ -890,6 +892,21 @@ def get_data(self,
         return get_data(self.conn, self.table_name, valid_param_names,
                         start, end)
 
+    def cache_parameter_data(self):
```
hbarovertwo, line 895: We came across a situation where we wanted to get access to the datasaver dataset during measurement loops, and flushing every loop helped with at least having it readily available. Is the intention here to have an easily accessible container for the data, without too much effort? This is Mason by the way 👍

qcodes/dataset/data_set.py
jenshnielsen: @ianhuezo Yes that should probably just be None type

qcodes/dataset/data_set.py
jenshnielsen: @hbarovertwo This tries to do two things but only one of them is implemented at the moment. I think I forgot to mention that. 

1) Make it possible to read the data gradually from disk as it is beeing written in a different process.  
2) Automatically keep a cache of the data in the process that is writing the the data. This is not currently implemented. 
   That cache can be used for inprocess liveplot or send over a socked or some such thing.

qcodes/dataset/data_set_cache.py
```diff
@@ -0,0 +1,99 @@
...
+        If the dataset is marked completed and data has already been loaded
+        no load will be performed.
+        """
+        if self._loaded_from_completed_ds:
+            return
+        self._dataset._completed = completed(self._dataset.conn, self._dataset.run_id)
+        if self._dataset.completed:
+            self._loaded_from_completed_ds = True
+
+        parameters = get_non_dependencies(self._dataset.conn, self._dataset.table_name)
```
astafan8, line 41: `get_non_dependencies` seems to be used only in this file and is inside calling `get_interdeps_from_result_table_name`. i'd suggest to get rid of `get_non_dependencies` and instead of it's call here execute `parameters = tuple(ps.name for ps in interdeps.non_dependencies)` on the `interdeps` returned by `get_interdeps_from_result_table_name`

qcodes/tests/dataset/test_data_set_cache.py
```diff
@@ -0,0 +1,297 @@
...
+                                 (param, param.get()))
+            datasaver.flush_data_to_database(block=True)
+            data = dataset.cache.data()
+            n_rows_written = i+1
+            if array_used:
+                shape = (n_rows_written, param.shape[0])
+            else:
+                shape = (n_rows_written * param.shape[0],)
+            if storage_type != 'array':
+                # TODO the expansion both for cache and get_parameter_data is buggy iff all types are array
```
astafan8, line 189: still todo? what is the bugginess?

qcodes/tests/dataset/test_data_set_cache.py
```diff
@@ -0,0 +1,297 @@
...
+            data = dataset.cache.data()
+            n_rows_written = i+1
+
+            if array_used:
+                expected_shape = (n_rows_written,) + param.shape
+            else:
+                expected_shape = n_rows_written * np.prod(param.shape)
+            assert data[param.full_name][param.full_name].shape == expected_shape
+            if storage_type != 'array':
+                # TODO the expansion both for cache and get_parameter_data is buggy iff all types are array
```
astafan8, line 269: same todo

qcodes/tests/dataset/test_data_set_cache.py
```diff
@@ -0,0 +1,297 @@
...
+                assert data[param.full_name][outer_param.full_name].shape == expected_shape
+            for setpoint_name in param.setpoint_full_names:
+                assert data[param.full_name][setpoint_name].shape == expected_shape
+            _assert_parameter_data_is_identical(dataset.get_parameter_data(),
+                                                data)
+    _assert_parameter_data_is_identical(dataset.get_parameter_data(),
+                                        dataset.cache.data())
+
+
```
astafan8, line 278: i couldn't see if there is a test where dependent parameters z1 and z2 are both dependent on x and y but, say, z1 is written for 10 values of x, and z2 is written for 20 values of x? did i miss it or does it not exist? if it doesn't, it seems like a worthwhile test to ensure that tracking rows per param tree works well.

qcodes/tests/instrument_mocks.py
```diff
@@ -332,6 +335,47 @@ def get_raw(self):
         return items
 
 
+
+class Multi2DSetPointParam2Sizes(MultiParameter):
```
astafan8, line 339: awesome!!!

jenshnielsen: Should update the docstring

qcodes/dataset/data_set_cache.py
```diff
@@ -0,0 +1,99 @@
...
+    def _merge_data_dicts_inner(existing_data: Dict[str, np.ndarray],
+                                new_data: Dict[str, np.ndarray]) -> Dict[str, np.ndarray]:
+        merged_data = {}
+        parameters = set(existing_data.keys()) | set(new_data.keys())
+
+        for parameter in parameters:
+            existing_values = existing_data.get(parameter)
+            new_values = new_data.get(parameter)
+            if existing_values is not None and new_values is not None:
+                merged_data[parameter] = np.append(existing_values, new_values, axis=0)
```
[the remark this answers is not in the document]
jenshnielsen: There is no such thing as appending in place. The numpy array has a fixed size if you append you will have to reallocate no matter what. 

Once we have shape  meta data we can allocate the correct size up front and fill

qcodes/tests/dataset/test_data_set_cache.py
jenshnielsen: The expansion is incorrect. This is not a new bug but just revealed by my testing. I will create an issue for it

qcodes/dataset/data_set_cache.py
jenshnielsen: Sure we can do this. As far as I remember this is not new code. I simply extracted part of a function into its own

qcodes/tests/dataset/test_data_set_cache.py
astafan8: :(

qcodes/tests/dataset/test_data_set_cache.py
jenshnielsen: Its there indirectly due to the multiparameters etc but lets just add it explicit too

qcodes/dataset/data_set.py
```diff
@@ -938,42 +945,52 @@ def get_data_as_pandas_dataframe(self,
...
+        return df
+
+    @staticmethod
+    def _generate_pandas_index(data: Dict[str, numpy.ndarray]) -> Union["pd.Index", "pd.MultiIndex"]:
+        import pandas as pd
+        keys = list(data.keys())
+        if len(data) <= 1:
+            index = None
+        elif len(data) == 2:
+            index = pd.Index(data[keys[1]].ravel(), name=keys[1])
```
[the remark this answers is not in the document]
jenshnielsen: The only thing that has changed here is that the code has been broken up into a few more funtions. I can revert that and get rid of `_get_data_for_one_param_tree` if that makes it easier to read? For now I have put in an explicit test that the name of the first parameter is as expected and raisa a value error if not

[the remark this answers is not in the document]
astafan8: well, i do like the `_get_data_for_one_param_tree `, it did make it easier to read, so let's leave it, i think :)

> I have put in an explicit test that the name of the first parameter is as expected and raisa a value error if not

that's perhaps good enough for now.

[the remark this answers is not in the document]
astafan8: could you also add a comment or smth to `_generate_pandas_index ` that `keys[0]` is always the dependent parameter?

qcodes/dataset/data_set_cache.py
```diff
@@ -0,0 +1,102 @@
...
+        """
+        Loads data from the dataset into the cache.
+        If new data has been added to the dataset since the last time 
+        this method was called, calling this method again would load 
+        that new portion of the data and append to the already loaded data.
+        If the dataset is marked completed and data has already been loaded
+        no load will be performed.
+        """
+        if self._loaded_from_completed_ds:
+            return
```
astafan8, line 40: wow, codecov tells that we don't have a test for this case :)

jenshnielsen: Yes, rewrote the tests to ensure that we also trigger this

PR #2066, opened 2020-06-30
Remove slave master references

WilliamHPNielsen:
This PR deprecates the use of "slave" as an argument name for a function and as a class name and variable name for magnet power supplies. We have one un-addressed master/slave reference in the code, but that is in `versioneer.py`, which should be addressed upstream.

I have played it safe by deprecation, but perhaps we can more aggressively simply remove/rename?

@QCoDeS/core

### DIFF
--- a/qcodes/instrument_drivers/ZI/ZIUHFLI.py
+++ b/qcodes/instrument_drivers/ZI/ZIUHFLI.py
@@ -1565,9 +1565,6 @@ class ZIUHFLI(Instrument):
                            val_mapping = {'ON': 1, 'OFF': 0},
                            vals=vals.Enum('ON', 'OFF'))
 
-        # TODO: make scope_trig_holdoffmode change according to
-        # parameter off scope_holdoff_seconds
-        # and scope_holdoff_events
         self.add_parameter('scope_trig_holdoffmode',
                             label="Scope trigger holdoff mode",
                             set_cmd=partial(self._setter, 'scopes', 0,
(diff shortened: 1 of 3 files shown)


qcodes/instrument_drivers/ZI/ZIUHFLI.py
```diff
@@ -1565,7 +1565,8 @@ def __init__(self, name: str, device_ID: str, **kwargs) -> None:
                            val_mapping = {'ON': 1, 'OFF': 0},
                            vals=vals.Enum('ON', 'OFF'))
 
-        # make this a slave parameter off scope_holdoff_seconds
+        # TODO: make scope_trig_holdoffmode change according to
+        # parameter off scope_holdoff_seconds
         # and scope_holdoff_events
         self.add_parameter('scope_trig_holdoffmode',
```
jenshnielsen, line 1571: I suggest removing todo since this is anyway deprecated so it will never be implemented

qcodes/instrument_drivers/oxford/MercuryiPS_VISA.py
```diff
@@ -206,6 +207,16 @@ def _param_setter(self, set_cmd: str, value: Union[float, str]) -> None:
         #  the intended value
 
 
+class MercurySlavePS(MercuryWorkerPS):
+    """
+    Deprecated class
+    """
+    def __init__(self, parent: VisaInstrument, name: str, UID: str) -> None:
+        issue_deprecation_warning("MercurySlavePS class", "",
```
astafan8, line 215: i believe that `deprecate` decorator works on classes, so this one can be just made into a `pass` class.

PR #2070, opened 2020-07-03
Move dataset related fixtures to conftest.

jenshnielsen:
So that we can avoid the chain of unused imports in test files


jenshnielsen:
> not super cool that dataset.conftest stuff is used in test modules/packages level(s) up, but it's for another PR.
Indeed some of the fixtures should be pushed up one more level but I want to wait for another pr to do that

PR #2074, opened 2020-07-07
B1500: Output NaN instead of 199.999e99

lakhotiaharshit:
Instead of passing dummy value 199.999e99 due to power compliance being reached in sampling measurements or measurement being aborted, NaN value is output.

Some relevant lines from Manual:
"Measurement data is over the measurement range. Or the sweep measurement was aborted by the automatic stop function or power compliance. D will be 199.999E+99 (no meaning)."

"The B1500 returns the data measured before any abort condition is detected. Dummy data 199.999E+99 will be returned for the data after abort."

@astafan8 @jenshnielsen

### DIFF
--- a/qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
+++ b/qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
@@ -243,6 +243,24 @@ def get_measurement_summary(status_array: np.ndarray) -> str:
 
     return summary
 
+
+def convert_dummy_val_to_nan(param: _FMTResponse):
+    """
+    Converts dummy value to NaN. Instrument may output dummy value (
+    199.999E+99) if measurement data is over the measurement range. Or the
+    sweep measurement was aborted by the automatic stop function or power
+    compliance. Or if any abort condition is detected. Dummy data
+    199.999E+99 will be returned for the data after abort."
+
+    Args:
+        param: This must be of type named tuple _FMTResponse.
+
+    """
+    for index, value in enumerate(param.status):
+        if param.value[index] == 199.999e99:
+            param.value[index] = float('nan')
+
+
 class B1500Module(InstrumentChannel):
     """Base class for all modules of B1500 Parameter Analyzer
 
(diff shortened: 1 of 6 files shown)


qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
```diff
@@ -243,6 +243,24 @@ def get_measurement_summary(status_array: np.ndarray) -> str:
...
+    199.999E+99) if measurement data is over the measurement range. Or the
+    sweep measurement was aborted by the automatic stop function or power
+    compliance. Or if any abort condition is detected. Dummy data
+    199.999E+99 will be returned for the data after abort."
+
+    Args:
+        param: This must be of type named tuple _FMTResponse.
+
+    """
+    for index, value in enumerate(param.value):
```
astafan8, line 259: not sure if we care about performance, or if my suggestion would even help, but we might be able to save some time by converting to numpy array, using logical slicing there, and then covnerting back to a list:
```python
values = np.array(param.value)
values[values > 199.999e99] = float('nan')
param.value = values.tolist()
```

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
```diff
@@ -243,6 +243,24 @@ def get_measurement_summary(status_array: np.ndarray) -> str:
...
+        if value == 199.999e99:
```
astafan8, line 260: i believe that JohnW used a more generous condition for nan substitution in measurements - ` > 1e99`. shall we do that? using `>` would also let us skip possible floating-point precision issues.

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
lakhotiaharshit: Since these are namedtuple I cannot assign values to its attribute value. However if this is faster I can define a new namedtuple replacing the value of 'value'. 
> param = param._replace(value = values.tolist()) 

Does this solution sound ok to you?

astafan8: nah, let's not use private methods :) not worth it

PR #2076, opened 2020-07-12
Add support for pytest 6

jenshnielsen:
To be released very soon (rc1 is out) 

This changes the logger slightly so that there are now 2 default loggers for pytest. 
It also adds types to pytest :) So fix one error related to that

### DIFF
--- a/qcodes/tests/common.py
+++ b/qcodes/tests/common.py
@@ -11,6 +11,7 @@ import qcodes
 from qcodes.utils.metadata import Metadatable
 from qcodes.configuration import Config, DotDict
 
+from _pytest._code.code import ExceptionChainRepr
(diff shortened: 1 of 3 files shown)


qcodes/tests/common.py
```diff
@@ -11,6 +11,7 @@
 from qcodes.utils.metadata import Metadatable
 from qcodes.configuration import Config, DotDict
 
+from _pytest._code.code import ExceptionChainRepr
```
jenshnielsen, line 14: I would like to avoid this private import. But since excinfo.getrepr can return two different types I don't see any way around this that does not involve either comparing the types or ignoring the type checking error. The later seems worse

PR #2079, opened 2020-07-16
Add formatting to the log messages of Telemetry

lakhotiaharshit:
Currently telemetry only captures messages which are not formatted to contain logger name and other things. Here we make it similar to logging of file handler and console handler

@jenshnielsen

### DIFF
--- a/qcodes/logger/logger.py
+++ b/qcodes/logger/logger.py
@@ -70,6 +70,18 @@ def get_formatter() -> logging.Formatter:
     return logging.Formatter(format_string)
 
 
+def get_formatter_for_telemetry() -> logging.Formatter:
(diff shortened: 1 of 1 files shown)


lakhotiaharshit:
> Thanks

This may be a hasty solution. Most of the information from the get_formatter ss also repeater in custom dimensions. I will check if I can just add a logger name in custom dimensions of telemetry. That may be better solution.

lakhotiaharshit:
Now message looks like this.
> qcodes.dataset.measurements ¦ _register_parameter ¦ Registered dac_ch1 in the Measurement.

jenshnielsen:
Looks good to me

PR #2088, opened 2020-07-17
Update v0.16.0 changelog

WilliamHPNielsen:
This is the proposed changelog for release `v0.16.0`

@QCoDeS/core

### DIFF
--- a/docs/changes/0.16.0.rst
+++ b/docs/changes/0.16.0.rst
@@ -1,7 +1,48 @@
-Changelog for QCoDeS 0.16.0 (not released yet)
+Changelog for QCoDeS 0.16.0
 ==============================================
+
+The July 2020 release of QCoDeS.
+
+Note: this will be the **last** QCoDeS release to support python 3.6
+
 
 Breaking Changes:
 _________________
 
-* ``math`` has been renamed as ``math_utils`` (#2040).
+* ``math`` has been renamed to ``math_utils`` (#2040).
+* The deprecated `default_measurement` of the `Station` is now removed (#2058)
+* The deprecated methods `get_data` and `get_values` of the `DataSet` have now been removed (#2067)
+* The `get_setpoints` method of the `DataSet` has been deprecated (#2068)
+
+
+New drivers:
+____________
+
+
+Improved drivers:
+_________________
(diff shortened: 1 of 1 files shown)


docs/changes/0.16.0.rst
```diff
@@ -1,7 +1,48 @@
...
+
+Improved:
+_________
+* All master-slave references from drivers have been removed or deprecated (#2066)
+* QCoDeS now comes with a code of conduct (#2078)
+
+Under the hood:
+_______________
+
+* Better execution of the test suite and hardening of tests (#2016, #2037, #2038, #2039, #2040, #2042, #2045, #2056, #2057, #2063, #2064, #2070, #2073, #2077, #2076)
```
jenshnielsen, line 44: Missing support for upcomming pytest 6 and explicit dependency on packaging via https://github.com/QCoDeS/Qcodes/pull/2076

docs/changes/0.16.0.rst
```diff
@@ -1,7 +1,48 @@
...
+* Update of dependencies to `mypy` 0.782 (#2034, #2075) and `pyqtgraph` 0.11.0 (#2036)
+* Include a new dependency: `typing_extensions` (#2048)
```
jenshnielsen, line 46: This is also mentioned in 0.15 was that wrong?

docs/changes/0.16.0.rst
jenshnielsen: I would also highlight that the merge of #2016 etc enable the execution of the test suite in parallel. This is a significant win for all developers. 

Furthermore it is also importatant to note that this now enables consistent handling of in memory config such that config values are now always consistently used from the in memory condig at `qcodes.config`. This also has the slight breaking change that the config is not reread from disk so a restart may be needed if you change the config on disk

docs/changes/0.16.0.rst
WilliamHPNielsen: Good question and good catch. I did `git checkout tags/v0.15.0` just now and `typing_extensions` **is** a part of `setup.py` for that version already. I wonder why it showed up in the list of merges since `v0.15.0` then? Anyway, I'll remove it from this changelog.

PR #2097, opened 2020-07-28
Fix: convert spyder kernels site print statement to logging message

peendebak:
In the spyder environment for each new console a message is printed about the site kernels. This PR converts the message to a logging statement and adds the text spyder to make it more informative

@astafan8


astafan8:
needed to reopen the PR in order to try to fix a hickup from travis - info that builds passed does not reach github

PR #2098, opened 2020-07-28
feat: Add a query to get database runs information in a nice python dictionary

edumur:
- Allow to get all runs info in only 3 sqlite requests instead of looping through all runs and making few request per run.

Checklist:
- Is the code ok with the rest of qcodes.
- I had no idea how to write test for that and other functions in the same file do not have test which didn't help me.
- Maybe the code is not safe for the database?

@astafan8


astafan8:
@edumur sorry for delay in reacting to this PR. i'll work on getting it in soon

PR #2104, opened 2020-08-10
remove conda ruamel package to hopofully fix pipelines ci build

jenshnielsen:
this no longer works well together with the pip version of the package

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -19,7 +19,6 @@ dependencies:
  - spyder
  - pyzmq
  - wrapt
- - pyyaml
  - lxml
  - scipy
  - gitpython
--- a/environment.yml
+++ b/environment.yml
@@ -18,7 +18,6 @@ dependencies:
  - pytest-runner
  - spyder
  - pyzmq
- - ruamel_yaml
  - wrapt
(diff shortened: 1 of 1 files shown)


environment.yml
```diff
@@ -18,7 +18,6 @@ dependencies:
  - pytest-runner
  - spyder
  - pyzmq
- - ruamel_yaml
  - wrapt
  - pyyaml
```
astafan8, line 22: what about this pyyaml ? :)

environment.yml
```diff
@@ -18,7 +18,6 @@ dependencies:
  - pytest-runner
  - spyder
  - pyzmq
- - ruamel_yaml
```
astafan8, line 21: we don't need to include it in the "pip" section becuase it's a requirement in qcodes's setup file, right?

jenshnielsen: No but that is true for a lot of package in both the env file and the requirements

environment.yml
jenshnielsen: Not really related to this pr but it seems to be unused so lets remove it

PR #2112, opened 2020-08-13
Move write cache to from datasaver to dataset

jenshnielsen:
First step in creating a more clean interface between the dataset and the datasaver

### DIFF
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -753,14 +753,8 @@ class DataSet(Sized):
         It is an error to add results to a completed :class:`.DataSet`.
         """
 
-        if self.pristine:
-            raise RuntimeError('This DataSet has not been marked as started. '
-                               'Please mark the DataSet as started before '
-                               'adding results to it.')
+        self._raise_if_not_writable()
 
-        if self.completed:
-            raise CompletedError('This DataSet is complete, no further '
-                                 'results can be added to it.')
         try:
             parameters = [self._interdeps._id_to_paramspec[name]
                           for name in results]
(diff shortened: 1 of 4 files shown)


qcodes/dataset/data_set.py
```diff
@@ -1167,6 +1176,194 @@ def __repr__(self) -> str:
 
         return "\n".join(out)
 
+    def _enqueue_results(
```
jenshnielsen, line 1179: Perhaps we should make this a public api?

qcodes/dataset/measurements.py
```diff
@@ -669,8 +478,7 @@ def __enter__(self) -> DataSaver:
         self.datasaver = DataSaver(
                             dataset=self.ds,
                             write_period=self.write_period,
-                            interdeps=self._interdependencies,
-                            write_in_background=self._write_in_background)
```
jenshnielsen, line 673: This is tecnially an API break but I don't think anyone is inializing the datasaver directly

astafan8: exactly. as per our docs, inititlzng DataSaver is NOT a pulbic API :)

qcodes/dataset/data_set.py
astafan8: that's probably fair, espcially if it's guarded by checking what state the dataset is in - enqueue should only work in the "running" state similar to add_results.

On the other hand, the interfaces of "add_results" and "enqueue results" differ significantly - add_results does not know about qcodes parameters and the fact that things need to unpacked, which enqueue_results knows. So, i would refrain from making it public just for now.

astafan8: on the other hand, there's also a publci `add_result_to_queue` :) 

UPDATE: had a look, and that method is NOT used by any code :) moreover, it's behavior is totally included in the behavior of `add_result`. So, either in this or another PR, we need to remove this `add_result_to_queue` method.

astafan8: my conclusion - let's keep `_enqueue_results` private for now, and regardless of that guard it with checks for the "running" state of the DataSet.

qcodes/dataset/data_set.py
```diff
@@ -1167,6 +1176,194 @@ def __repr__(self) -> str:
...
+                if value.shape:
+                    res_list += [{param.name: number} for number in value]
+                else:
+                    res_list += [{param.name: complex(value)}]
+            else:
+                res_list += [{param.name: value}]
+
+        return res_list
+
+    def flush_data_to_database(self, block: bool = False) -> None:
```
astafan8, line 1336: shall we keep this private? my thinking is if it's private now, then if we decide to change/remove/move it, we don't need to do deprecation cycles and stuff.

PR #2115, opened 2020-08-14
limit pytest xdist to version 1.x

jenshnielsen:
Will hopefully fix the test failure in #2114


jenshnielsen:
This is due to pytest-cov not supporting the changes in pytest-xdist 2.0.0 should be fixed once a new version of pytest-cov is out
see https://github.com/pytest-dev/pytest-cov/issues/411

PR #2124, opened 2020-08-19
Only create one background writer for multiple datasets

jenshnielsen:
This will eventually make it safe to have multiple datasets writen at the same time and therefor to nest measurement context managers. However there are a number of open questions in the design

* This binds the background thread as a class variable in the Dataset class. Alternatively this could be a module level class (or something else?)
* All datasets share one queue. This is the simplest implementation of the background writer as multiple queues would require 
handling that one queue is empty but anther one is full. (Likely done with a timeout on get and cycle between the queus) 
This seems like it would be hard to ensure optimal performance in this way. However a single queue makes it hard to block on there not being any data left to write for a specific dataset
* To get the tests to pass this requires ensuring that flush is always called with block=True otherwise the background writer may not be done when the context exits. Technically this could happen even without this change but seems to be more likely to happen now.


qcodes/dataset/data_set.py
```diff
@@ -818,18 +875,30 @@ def add_result_to_queue(self,
...
+
+    def _ensure_dataset_written(self) -> None:
+        writer_status = self._writer_status
+        if writer_status is None:
+            raise ValueError("Dataset writer has not been setup")
+
+        if writer_status.write_in_background:
+            writer_status.data_write_queue.put({'keys': 'finalize', 'values': self.run_id})
+            while self.run_id in writer_status.active_datasets:
+                time.sleep(1e-3)
```
jenshnielsen, line 893: One could replace this with a threading.event to signal the dataset writting has been complted

qcodes/dataset/data_set.py
```diff
@@ -1333,26 +1402,31 @@ def _flush_data_to_database(self, block: bool = False) -> None:
...
                 else:
                     log.warning(f'Could not commit to database; {e}')
         else:
             log.debug('No results to flush')
 
-        if self._bg_writer.is_alive() and block:
+        if writer_status.write_in_background and block:
             log.debug(f"Waiting for write queue to empty.")
-            self._data_write_queue.join()
+            writer_status.data_write_queue.join()
```
jenshnielsen, line 1429: This now bloks on all datasets being written and not just the relevant one. I think this is acceptable and makes the code a lot cleaner

qcodes/dataset/data_set.py
```diff
@@ -710,8 +745,23 @@ def _perform_start_actions(self, start_bg_writer: bool) -> None:
         pdl_str = links_to_str(self._parent_dataset_links)
         update_parent_datasets(self.conn, self.run_id, pdl_str)
 
+        writer_status = self._writer_status
+
+        write_in_backgrond_status = writer_status.write_in_background
```
astafan8, line 750: typo - groUnd
```suggestion
        write_in_background_status = writer_status.write_in_background
```

qcodes/dataset/data_set.py
astafan8: would that be better because one doesn't need to define the time.sleep? or because the code would be more straightforward to read?

i am personally ok with the sleep here

qcodes/dataset/data_set.py
astafan8: i agree. especially because one can't mix out-of-thread and main-thread writing for the same DB file.

qcodes/dataset/data_set.py
```diff
@@ -1333,26 +1395,29 @@ def _flush_data_to_database(self, block: bool = False) -> None:
...
                 else:
                     log.warning(f'Could not commit to database; {e}')
         else:
             log.debug('No results to flush')
 
-        if self._bg_writer.is_alive() and block:
+        if writer_status.write_in_background and block:
             log.debug(f"Waiting for write queue to empty.")
-            self._data_write_queue.join()
+            writer_status.data_write_queue.join()
```
astafan8, line 1420: one question - perhaps i missed - from the DataSaver/RUnner side we never block on the flushing, even when exiting the measurement context manager (see here https://github.com/QCoDeS/Qcodes/blob/84648a52b79b3af7b48237f5062bd41302126fe8/qcodes/dataset/measurements.py#L491) - is that good? i was thinking that likely when exiting the outermost measurement context manager, the user really expects ALL the data to be successfully written? or is this an unreasonable ask?

qcodes/tests/dataset/test_concurrent_datasets.py
```diff
@@ -0,0 +1,60 @@
...
+    exp = new_experiment("test", "test1", conn=empty_temp_db_connection)
+    ds1 = DataSet(conn=empty_temp_db_connection)
+    ds1.mark_started(start_bg_writer=True)
+
+    ds2 = DataSet(conn=empty_temp_db_connection)
+    with pytest.raises(RuntimeError, match="All datasets written"):
+        ds2.mark_started(start_bg_writer=False)
+
+
+def test_background_after_foregroud_raises(empty_temp_db_connection):
```
astafan8, line 20:
```suggestion
def test_background_after_foreground_raises(empty_temp_db_connection):
```

astafan8 (approved):

astafan8:
perhaps we should add a few cells to the measurement context manager notebook showing that nested measurements work and mentioning that they work with the background thread and that mixing is not allowed?

qcodes/dataset/data_set.py
jenshnielsen: Mostly since the sleep time is a bit random. If its too long this will potentially be very slow if too small we are wasting cpu cycles in this busy loop but the current value seems to work well enough

astafan8: ok. then it's fine for now, i guess

astafan8: the sleeping time can be made a attribute/class-variable ;)

qcodes/dataset/data_set.py
jenshnielsen: So this actually happens another way round. The ``__exit__`` method of the Runner context manager calls ``_ensure_dataset_written`` which will use the finalize message to ensure that a given dataset is written before the context manager can exit.

qcodes/dataset/data_set.py
jenshnielsen: Sure why not.

qcodes/dataset/data_set.py
astafan8: aaah!! that's amazing ! :)

PR #2132, opened 2020-08-27
Simplify instrument destructor

jenshnielsen:
Currently `__del__` first tries to remove the instrument from the instances list and then calls close which does the same + more.
It seems better to just call close. This also enables us to get rid of a subclassing of `__del__` in IPInstrument that does just that

### DIFF
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -504,7 +504,7 @@ class Instrument(InstrumentBase, AbstractInstrument):
         try:
             self.close()
         except BaseException:
-            log.exception("Failed to close instrument in destructor")
+            pass
 
     def close(self) -> None:
         """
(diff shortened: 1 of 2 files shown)


qcodes/instrument/base.py
```diff
@@ -502,12 +502,9 @@ def __repr__(self) -> str:
...
         """Close the instrument and remove its instance record."""
         try:
-            wr = weakref.ref(self)
-            if wr in getattr(self, '_instances', []):
-                self._instances.remove(wr)
             self.close()
-        except:
-            pass
+        except BaseException:
+            log.exception("Failed to close instrument in destructor")
```
jenshnielsen, line 507: I am not sure if logging from a ``_del__`` method is safe? Perhaps we should just ommit this

astafan8: agree to omit. close already has logging

astafan8 (approved):

PR #2138, opened 2020-09-03
Add trigger, and program feature to Yokogawa GS200, and digitizer feature to Keithley 7510

liangosc:
First of all, this PR includes changes to **two** instruments.

The reason why I want to include two changes in one:
The changes made to the Yokogawa GS200 will add the ability of programming source data pattern, and sending out external triggers. Hence, it'll be great if I can measure the pattern with a dmm which can receive the trigger.

The changes made to the Keithley 7510 will add the ability of receiving external trigger, and performing a "digitize" measurement, which is ideal for high speed measurement.

It feels "incomplete" if I include the new changes to each instrument in its existing QCoDeS example, so I combine two together in this one PR, with one new example notebook, but I can always split this PR into two separate ones if necessary.

For the Yokogawa GS200:
- A "GS200_Program" class is added, so that the user can program the source data pattern. Each step in the program is executed in order, and each step has a fixed execution time. User can also set the slope time, and generate step responses, ramp responses, and various waveforms such as triangular waveforms.
- BNC trigger in and out are added as parameters.

For the Keithley 7510:
- The "Buffer" class is added, same as the one for Keithley 2450, and several related parameter/methods are also added to the driver.
- A new "DigitizeSense" class is added. The digitize sense function is similar to the normal sense, but has some of its own method. In other words, the "digitize sense" and "sense" each has some unique methods, so I decided to make a new class, instead of subclass the "Sense" class.

@sohailc


qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -161,6 +161,62 @@ def update_measurement_enabled(self, unit: str,
...
+class GS200Program(InstrumentChannel):
+    """
+    """
+    def __init__(self, parent: 'GS200', name: str) -> None:
+        super().__init__(parent, name)
+        self._repeat = 1
+        self._file_name = None
+
+        self.add_parameter("interval",
+                           label="the program interval time",
```
jenshnielsen, line 173: do these parameters have a unit (seconds) ?

jenshnielsen: Or any validators that can be set

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -161,6 +161,62 @@ def update_measurement_enabled(self, unit: str,
...
+                           vals=Enum('normal', 'mend'))
+
+        self.add_parameter("save",
+                           label="save the program",
+                           set_cmd=":PROG:SAVE {}")
+
+        self.add_parameter("load",
+                           label="load the program",
+                           get_cmd=":PROG:LOAD?",
+                           set_cmd=":PROG:LOAD {}")
```
jenshnielsen, line 195: Could you add some explanation of how load and save works ? it is not clear from the docscrings and the examples. Can we validate the input for save and load/

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -302,6 +356,29 @@ def __init__(self, name: str, address: str, terminator: str = "\n",
...
+        self.add_parameter("BNC_out",
+                           label="BNC trigger out",
+                           get_cmd=":ROUT:BNCO?",
+                           set_cmd=":ROUT:BNCO {}",
+                           vals=Enum("trigger", "output", "ready"),
+                           docstring="Sets or queries the output BNC signal")
+
+        self.add_parameter("BNC_in",
+                           label="BNC trigger in",
+                           get_cmd=":ROUT:BNCO?",
```
jenshnielsen, line 370: Is this really ROUT? Same as above ?

qcodes/instrument_drivers/tektronix/keithley_7510.py
```diff
@@ -1,10 +1,201 @@
...
+            get_cmd=f":TRACe:ACTual? '{self.buffer_name}'",
+            get_parser=int,
+            docstring="To get the number of readings in the reading buffer."
+        )
+
+        self.add_parameter(
+            "last_index",
+            get_cmd=f":TRACe:ACTual:END? '{self.buffer_name}'",
+            get_parser=int,
+            docstring="To get the last index in the reading buffer."
```
jenshnielsen, line 86: Is this the index of the last point read or the last point allocated in the buffer

qcodes/instrument_drivers/tektronix/keithley_7510.py
```diff
@@ -1,10 +1,201 @@
...
+            docstring="List of buffer elements to read."
+        )
+
+        self.add_parameter(
+            "fill_mode",
+            get_cmd=":TRACe:FILL:MODE?",
+            set_cmd=":TRACe:FILL:MODE {}",
+            docstring="if a reading buffer is filled continuously or is filled"
+                      " once and stops"
+        )
```
jenshnielsen, line 112: can we add an enum validator to this one ?

qcodes/instrument_drivers/tektronix/keithley_7510.py
```diff
@@ -1,10 +1,201 @@
...
+
+    def from_name_to_scpi(self, element_names: List[str]) -> List[str]:
```
jenshnielsen, line 114: Is the user expected to use these two methods or should they be private?

qcodes/instrument_drivers/tektronix/keithley_7510.py
```diff
@@ -225,12 +521,81 @@ def __init__(self, name: str, address: str, terminator='\n', **kwargs):
...
+                for key, value in DigitizeSense7510.function_modes.items()
+            },
+            docstring="Add digitize sense functions."
+        )
+
+        self.add_parameter(
+            "buffer_name",
+            get_cmd=None,
+            set_cmd=None,
+            docstring="name of the reading buffer in using."
```
jenshnielsen, line 539: in using ?

qcodes/instrument_drivers/tektronix/keithley_7510.py
```diff
@@ -225,12 +521,81 @@ def __init__(self, name: str, address: str, terminator='\n', **kwargs):
...
 
+        self.add_parameter(
+            "digi_sense_function",
+            set_cmd=":DIGitize:FUNCtion {}",
+            get_cmd=":DIGitize:FUNCtion?",
+            val_mapping={
+                key: value["name"]
+                for key, value in DigitizeSense7510.function_modes.items()
+            },
+            docstring="Add digitize sense functions."
```
jenshnielsen, line 532: It's hard to understand from this docstring what this parameter does

jenshnielsen:
There does not seem to be an example of reading the buffer into a qcodes parameter. Is this supported? It is an esential feature for a qcodes driver to be able to capture the data as a parameter so it can be storred in a dataset easily

liangosc (commented):
> There does not seem to be an example of reading the buffer into a qcodes parameter. Is this supported? It is an esential feature for a qcodes driver to be able to capture the data as a parameter so it can be storred in a dataset easily

I have made `data` a parameter, and the usage is as following:
> buffer.data_start = 1
buffer.data_end = total_data_points
v=buffer.data()

the type of `v` is a list, with float-type elements.

User can specify several different elements, such as time, measurement, unit...
>buffer.elements(['measurement', 'time', 'measurement_unit'])
data=buffer.data()

In this case, the data will be a 3xN matrix/array, of which all the elements are str-type.

I didn't add the `vals=Arrays(...)` nor the `parameter_class`, for the `data` parameter, because:
1. the shape of the data is not fixed, it depends on how many elements the user wants;
2. the `parameter_class=...` part can only be added if there is the `vals=Arrays(..)`
suggestion? @jenshnielsen 

(Maybe I can just save everything -- all total 14 elements)

qcodes/instrument_drivers/tektronix/keithley_7510.py
```diff
@@ -1,10 +1,205 @@
...
+
+    def __init__(
+            self,
+            parent: 'Keithley7510',
+            name: str,
+            size: Optional[int] = None,
+            style: str = ''
+    ) -> None:
+        super().__init__(parent, name)
+        self.buffer_name = name
```
jenshnielsen, line 47: Why is `self.buffer_name` needed. It seems like this will always be equivalent to self.short_name which is already defined on all instruments

jenshnielsen:
> (Maybe I can just save everything -- all total 14 elements)
I think you can set this up dynamically from the parameters that determine the size. Have a look at 
https://qcodes.github.io/Qcodes/examples/Parameters/Simple-Example-of-ParameterWithSetpoints.html?highlight=parameterwithsetpoints and see if it is possible to do something similar to the parameter with setpoints

qcodes/instrument_drivers/tektronix/keithley_7510.py
liangosc: It turns out `self.short_name` works. (not `self.name` though)

I was not sure if I should use `self.name` or `self.short_name`, so I created my own... :D

liangosc:
> > (Maybe I can just save everything -- all total 14 elements)
> 
> I think you can set this up dynamically from the parameters that determine the size. Have a look at
> https://qcodes.github.io/Qcodes/examples/Parameters/Simple-Example-of-ParameterWithSetpoints.html?highlight=parameterwithsetpoints and see if it is possible to do something similar to the parameter with setpoints
> 
> EDIT:
> 
> It would be nice to have it as a ParameterWithSetpoints so that one can register what setpoints the axes corresponds to

I tried to match what they did in the example, and here it is... 

My concerns:
1. a `dummy_setpoints` variable is used -- is it necessary?
2. I have to remove the datatype validation because the result may not be `Array` (it can be list of string)
3. I have to add a extra `update" step to populate the `data` parameter

All those extra works just to make the `data` a `parameter`. Suggestion? @jenshnielsen

jenshnielsen:
@liangosc It is not just making a parameter :) Making a parameter is esential to the fuction of the instrument as a qcodes instrument. 

We can probably loosen the requirement and make it an array of strings.

The best construction for the setpoints would be to make it easy to pass in another parameter which acts as the setpoints in this example proably from the yokogawa. I will try to write something to show you what I mean

qcodes/instrument_drivers/tektronix/keithley_7510.py
```diff
@@ -1,10 +1,367 @@
...
+class DataArray7510(MultiParameter):
+    """
+    Data class when user selected more than one field for data output.
+    """
+    value = None
+
+    def __init__(self,
+                 names: tuple,
+                 shapes: tuple,
+                 setpoints: tuple,
```
jenshnielsen, line 20: Please write these out to match the super class 

```
                 names: Sequence[str],
                 shapes: Sequence[Sequence[int]],
                 setpoints: Optional[Sequence[Sequence]],
```

qcodes/instrument_drivers/tektronix/keithley_7510.py
```diff
@@ -1,10 +1,367 @@
...
+            parent: 'Keithley7510',
+            name: str,
+            size: Optional[int] = None,
+            style: str = ''
+    ) -> None:
+        super().__init__(parent, name)
+        self._size = size
+        self.style = style
+        self.data_start = 1  # first index of the data to be returned
+        self.data_end = 1    # last index of the data to be returned
```
jenshnielsen, line 109: Please make data_end and data_start parameters with get_cmd and set_cmd = False.

qcodes/instrument_drivers/tektronix/keithley_7510.py
```diff
@@ -1,10 +1,367 @@
...
+
+        self.add_parameter(
+            "n_pts",
+            label="total n for the setpoints",
+            unit="",
+            initial_value=10,
+            get_cmd=None,
+            set_cmd=None,
+            get_parser=int
+        )
```
jenshnielsen, line 190: When data_start and and data_end are made parameters this can simply be made a parameter that can be get but not set and calculates the number of setpoints from the other paramters

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -316,6 +376,47 @@ def __init__(self, name: str, address: str, terminator: str = "\n",
...
+        )
+
+        self.add_parameter(
+            "setpoints_stop",
+            initial_value=0,
+            unit="",
+            label="stop value for the setpoints",
+            get_cmd=None,
+            set_cmd=None
+        )
```
jenshnielsen, line 395: It does not seem like these setpoints parameters are currently used ?

qcodes/instrument_drivers/tektronix/keithley_7510.py
```diff
@@ -1,10 +1,367 @@
...
+    A Parameter class which can includes string values.
+    """
+    def validate(self, value: Any) -> None:
+        """
+        Overwrites the standard ``validate`` method to only check the the
+        parameter has consistent shape with its setpoints.
+
+        Note: the validation of the datatype is removed!
+        """
+        self.validate_consistent_shape()
```
jenshnielsen, line 70: Why is this hack needed?

qcodes/instrument_drivers/tektronix/keithley_7510.py
liangosc: fixed.

qcodes/instrument_drivers/tektronix/keithley_7510.py
liangosc: They are parameters now, with get/set_cmd=`None` (I have been using `None` for other places, but it has to be `False`, please let me know)

qcodes/instrument_drivers/tektronix/keithley_7510.py
liangosc: Good point. I have made the changes.

qcodes/instrument_drivers/yokogawa/GS200.py
liangosc: You are right, they are not currently used, so I decided to remove them from this PR.

qcodes/instrument_drivers/tektronix/keithley_7510.py
```diff
@@ -1,10 +1,367 @@
...
+                         shapes=shapes,
+                         setpoints=setpoints,
+                         **kwargs)
+        for param_name in self.names:
+            self.__dict__.update({param_name: []})
+
+    def set_raw(self, value: tuple) -> None:
+        self.value = value
+        for i in range(len(self.names)):
+            setattr(self, self.names[i], value[i])
```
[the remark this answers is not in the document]
liangosc: It's actually used in the `_get_data` method. When there are more than one elements are selected, the `set_raw` function is used to set the value.

qcodes/instrument_drivers/tektronix/keithley_7510.py
liangosc: For some of the "elements", for example, "measurement_formatted", the return values would have letters in addition of numbers: `['+00.010 V', '+00.011 V'...]`. It can not be converted to numbers, and has to be kept as list of str, which won't pass the validation check of `ParameterWithSetpoints`.

This is only for the situation when only one element is selected, because for more than one, `MultiParameter` works with list of str. Unfortunately, `MultiParameter` has to have more than one values, so for 1-elements, I have to hack the `ParameterWithSetpoints`.

Suggestions?

jenshnielsen: I don't understand why you say that. There is no requirement that the Multiparameter must have more than one value

docs/examples/driver_examples/QCoDeS Example with Yokogawa GS200 and Keithley 7510.ipynb
```diff
@@ -0,0 +1,1829 @@
...
+    "    if 'No error' in dmm_error:\n",
+    "        break\n",
+    "    print(dmm_error)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 4. Make the measurement"
```
jenshnielsen, line 770: Please rewrite this section so that we only show how to make measurements with qcodes. We do not want to promote manual measurements with this driver that are different from every other driver

qcodes/instrument_drivers/tektronix/keithley_7510.py
liangosc: You are absolutely right. Now only the `MultiParameter` is used, no need for `ParameterWithSetpoints`.

qcodes/instrument_drivers/tektronix/keithley_7510.py
```diff
@@ -1,10 +1,372 @@
...
+            elements = ['measurement']
+        else:
+            elements = self.elements()
+        n_elements = len(elements)
+
+        units = tuple(elements_units[element] for element in elements)
+        processed_data = dict.fromkeys(elements)
+        for i, (element, unit) in enumerate(zip(elements, units)):
+            if unit == 'str':
+                processed_data[element] = all_data[i::n_elements]
```
jenshnielsen, line 330: This then needs to be converted to the right type? As in to numpy array of floats ?

qcodes/instrument_drivers/tektronix/keithley_7510.py
```diff
@@ -1,10 +1,372 @@
...
+            shapes=((self.n_pts(),),) * n_elements,
+            units=units,
+            setpoints=((self.setpoints(),),) * n_elements,
+            setpoint_units=((self.setpoints.unit,),) * n_elements,
+        )
+        data._data = tuple(
+            tuple(processed_data[element]) for element in elements
+        )
+        for i in range(len(data.names)):
+            setattr(data, data.names[i], processed_data[data.names[i]])
```
jenshnielsen, line 347: Why do we need this?

qcodes/instrument_drivers/tektronix/keithley_7510.py
liangosc: Should I try to converted the non-numerical data to floats, like the datetime "10/20/2020 06:11:11.977197606"?

qcodes/instrument_drivers/tektronix/keithley_7510.py
liangosc: To be honest, we want to access the data in the following way:
>data = buffer.data()
buffer.elements(['element1', 'element2']
data1 = data.element1   <---- this
data2 = data.element2    <---- and this. By doing so, the python lint would be able to check if "element1" is an attribute of object "data".

In addition to the QCoDeS way:
>dataset = datasaver.dataset
data = dataset.get_parameter_data()
data1 = data['element1']
data2 = data['element2']

docs/examples/driver_examples/QCoDeS Example with Yokogawa GS200 and Keithley 7510.ipynb
liangosc: The make measurement section has been rewritten.

qcodes/instrument_drivers/tektronix/keithley_7510.py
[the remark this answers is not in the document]
liangosc: Update: there is no `set_raw` method anymore.

qcodes/instrument_drivers/tektronix/keithley_7510.py
liangosc: Update: all the non-numerical data is left as string.

jenshnielsen (approved):

PR #1493, opened 2019-02-27
[WIP] Tests: Complex Arrays

jenshnielsen:
Test that complex values can be stored correctly in array data types. Sits on top of #1489 which should be merged before this

PR #2140, opened 2020-09-07
Remove warning about complex numbers in R&S ZNB driver

mgunyho:
As of https://github.com/QCoDeS/Qcodes/pull/1493, this warning seems redundant. We've checked in an experiment that the measured complex values get stored and can be read out/plotted properly.

@astafan8


astafan8 (approved):
thanks a lot @mgunyho !!!!! 👍

Issue #2003, opened 2020-05-14
Yokogawa GS210 output mode cache always set to 'VOLT' upon initializing


abargerbos:
### Steps to reproduce
1. load Yokogawa GS210 instrument  
2. set yoko.source_mode('CURR')
3. close connection, reload instrument
4. try calling yoko.current_range()

### Expected behaviour
What should happen is that the current range is returned.

### Actual behaviour
What happens instead is a value error:
('Cannot get/set CURR settings while in VOLT mode', 'getting yoko_current_range')
This is because the driver checks the mode of the command (a command like current_range can only be used in the 'CURR' mode) not against the actual source mode (gotten from yoko.source_mode()) but against a cached value yoko._cached_mode. However, in the init yoko._cached_mode is always set to 'VOLT', even when yoko.source_mode is 'CURR'. 

This is probably fixed by changing line 198 in the driver to self._cached_mode = self.source_mode(), but I was not 100% sure that this is the desired solution? 

### System
**operating system**
Windows 10
**qcodes version**
0.13.0 (pip installed)

astafan8:
@abargerbos i think indeed the fix that you are suggesting is a step in the right direction - could you submit a PR and then we can take it from there?

WilliamHPNielsen:
@Akshita07 or @FarBo, this might be an issue for you to look into?

FarBo:
> @Akshita07 or @FarBo, this might be an issue for you to look into?

Definitely.

FarBo:
@abargerbos 
I am working on some fixes. I will test those fixes and make a PR. But, the fix is something like what you proposed.

FarBo:
@abargerbos
Have you seen this notebook:
https://qcodes.github.io/Qcodes/examples/driver_examples/Qcodes%20example%20with%20Yokogawa%20GS2xx.html?highlight=source_mode
With the current driver, you should be able to switch between modes as shown in the notebook. Could you try to follow the notebook to see if you can successfully switch to the current mode and run yoko.current_range() without any issue?
The driver can be improved, but I would like to know if the current driver is capable of switching between modes.

PR #2143, opened 2020-09-10
Yokogawa GS200 source_mode bug fix

FarBo:
This PR Fixes issue #2003.
The local parameter _cached_mode, which was storing the mode of the instrument is removed. source_mode.get_latest() is used instead to capture the latest mode of the instrument. These changes have fixed the source_mode bug, when someone wants to close the instrument and initialize it again instead of turning off and on. 

Note that the functionality of this PR has been tested with the instrument.


qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -191,11 +191,10 @@ def __init__(self, name: str, address: str, terminator: str = "\n",
...
 
-        # When getting the mode internally in the driver, look up the mode as
-        # recorded by the _cached_mode property, instead of calling source
-        # _mode(). This will prevent frequent VISA calls to the instrument.
-        # Calling _set_source_mode will change the chased value.
-        self._cached_mode = "VOLT"
+        # The source_mode attribute of the instrument. The default mode is
+        # VOLT when the instrument is truned on. The mode can be switched
+        # between VOLT and CURR.
+        self.source_mode()
```
jenshnielsen, line 197: Typo `truned on`

As it is this comment is not very useful. It should explain why we get the mode here. This is because at connect time it is what ever the user set it to the last time or voltage if it has not been set.

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -471,9 +470,9 @@ def _assert_mode(self, mode: str) -> None:
         Args:
             mode: "CURR" or "VOLT"
         """
-        if self._cached_mode != mode:
+        if self.source_mode() != mode:
```
jenshnielsen, line 473: why does this not use get_latest ? 

The docstring is also wrong (There is no such thing as check)

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -191,11 +191,10 @@ def __init__(self, name: str, address: str, terminator: str = "\n",
...
-        # _mode(). This will prevent frequent VISA calls to the instrument.
-        # Calling _set_source_mode will change the chased value.
-        self._cached_mode = "VOLT"
+        # The source_mode attribute of the instrument. The default mode is
+        # VOLT when the instrument is truned on. The mode can be switched
+        # between VOLT and CURR.
+        self.source_mode()
 
         # We want to cache the range value so communication with the instrument
         # only happens when the set the range. Getting the range always returns
```
jenshnielsen, line 200: Can we also get rid of cached_range_value just below and `_auto_range ` for the same reasons (use the cache within the parameter)

qcodes/instrument_drivers/yokogawa/GS200.py
jenshnielsen: The code from line 232 to 247 still assumes that we are in voltage mode when starting up. This is wrong. The snapshot_exlude needs to be set depending on the mode we are in.

jenshnielsen: Same is true for voltage_range and current_range around line 204

qcodes/instrument_drivers/yokogawa/GS200.py
FarBo: get_latest is added.
The docstring could be something like this: To raise Error when the active source_mode of the instrument is not the same as the desired mode of the user.

qcodes/instrument_drivers/yokogawa/GS200.py
FarBo: I think this is better:
This attribute returns the latest user-set mode (VOLT or CURR) upon initialization of the instrument. It returns VOLT as the default mode of the instrument if nothing is set.

jenshnielsen: I would suggest something like

`we need to get the value of source_mode here as we cannot rely on the default value
as the user may have changed the value before we connect to the instrument (in a previous session or via the 
frontpanel)`

We do not need to document that this returns the value. That is obvious from how parameters work. We only need to document why we need to update it here

qcodes/instrument_drivers/yokogawa/GS200.py
jenshnielsen: I would just delete the second line of the docstring. The first line is fine

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -191,38 +191,43 @@ def __init__(self, name: str, address: str, terminator: str = "\n",
...
+        # instrument (in a previous session or via the frontpanel).
+        self.source_mode()
+
+        self.add_parameter('range',
+                           label='Source Range',
+                           get_cmd=':SOUR:RANG?',
+                           set_cmd=partial(self._set_range, self.source_mode()),
+                           vals=Enum(1e-3, 10e-3, 100e-3, 200e-3, 1e0,
+                           10e0, 30e0),
+                           )
```
jenshnielsen, line 205: I don't think that it is a good idea to add this range parameter. Now there are two different parameters that reprecent the same value (In voltage mode voltage_range and range) furthermore this does not correctly do the logic that voltage and current range does

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -191,38 +191,43 @@ def __init__(self, name: str, address: str, terminator: str = "\n",
...
 
         self.add_parameter('voltage_range',
                            label='Voltage Source Range',
                            unit='V',
                            get_cmd=partial(self._get_range, "VOLT"),
                            set_cmd=partial(self._set_range, "VOLT"),
-                           vals=Enum(10e-3, 100e-3, 1e0, 10e0, 30e0))
+                           vals=Enum(10e-3, 100e-3, 1e0, 10e0, 30e0),
+                           snapshot_exclude=(True if self.source_mode() ==
+                           'CURR' else False))
```
jenshnielsen, line 219: Since this and the matching logic statement for current are both reused I would refactor them into variables.

`exclude_voltage_snapshot = True if self.source_mode() =='CURR' else False`

or just
`exclude_voltage_snapshot = self.source_mode() =='CURR'`

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -191,38 +191,43 @@ def __init__(self, name: str, address: str, terminator: str = "\n",
...
                            get_cmd=partial(self._get_range, "CURR"),
                            set_cmd=partial(self._set_range, "CURR"),
                            vals=Enum(1e-3, 10e-3, 100e-3, 200e-3),
-                           snapshot_exclude=True
+                           snapshot_exclude=(True if self.source_mode() ==
+                           'VOLT' else False)
                            )
 
-        # This is changed through the source_mode interface
-        self.range = self.voltage_range
```
jenshnielsen, line 224: Rather than replacing this with a new parameter range should be an alias for voltage range if we are in voltage mode and an alias for current range if we are in current mode

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -234,22 +239,20 @@ def __init__(self, name: str, address: str, terminator: str = "\n",
...
                            unit='I',
                            set_cmd=partial(self._get_set_output, "CURR"),
                            get_cmd=partial(self._get_set_output, "CURR"),
-                           snapshot_exclude=True
+                           snapshot_exclude=(True if self.source_mode() ==
+                           'VOLT' else False)
                            )
 
-        # This is changed through the source_mode interface
-        self.output_level = self.voltage
```
jenshnielsen, line 251: You can't remove this. It is used within _ramp_source. Rather it needs to be set to voltage or current depending on the mode that we are in

FarBo: That's correct.

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -441,10 +444,10 @@ def _update_measurement_module(self, source_mode: str = None,
             return
 
         if source_mode is None:
-            source_mode = self._cached_mode
+            source_mode = self.source_mode.get_latest()
         # Get source range if auto-range is off
         if source_range is None and not self.auto_range():
-            source_range = self.range()
+            source_range = float(self.range.get_latest())
```
jenshnielsen, line 450: This replaces a call to the instrument with the cache. Is that safe

qcodes/instrument_drivers/yokogawa/GS200.py
FarBo: The only reason for adding this range parameter is to use it in _set_output method. This method checks if we are in correct range to set the output value and raise error if the value is out of range. In the original code, a local _cache_range_value served this checking, which we wanted to get rid of that. So, what do you suggest for the replacement of _cache_range_value?

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -393,25 +396,25 @@ def _set_output(self, output_level: float) -> None:
...
         # Check we are not trying to set an out of range value
-        if self._cached_range_value is None or abs(output_level)\
+        if self.range.get_latest() is None or abs(output_level)\
                 > abs(self_range):
             # Check that the range hasn't changed
             if not auto_enabled:
                 # Update range
-                self.range()
-                self_range = self._cached_range_value
+                self.range.get_latest()
```
jenshnielsen, line 416: This also replaces a call to the insturment with the cache. Is that safe?

qcodes/instrument_drivers/yokogawa/GS200.py
jenshnielsen: You are confusing two different things.

`self.range` is an alias for either `self.voltage_range` or self.current_range` depending on the mode. You might think that is it safe to replace a alias with a new definition of the parameter with the same command but it is not since the snapshot will be out of date e.g. if you set the voltage_range the snapshot of range will not update and the other way around. 

Futher to that there was an internal cache of the range value stored in _cache_range_value. That is unneeded and error prone since both `voltage_range` and `current_range` have their own internal cache and the alias to them called `range` therefor also does.

Rather than just creating an alias as the orginal code does it would be better to use a DelegateParameter https://qcodes.github.io/Qcodes/api/parameters/parameter.html?qcodes.instrument.parameter.DelegateParameter#qcodes.instrument.parameter.DelegateParameter for `range` and change its source as needed

qcodes/instrument_drivers/yokogawa/GS200.py
FarBo: Self_range updates to the latest range value of the instrument and prevents setting out of range output value. I tested it with all range modes and source modes on the instrument.

jenshnielsen: `self_range = float(self.range.get_latest())` does not communicate with the instrument. It calls the cache. 

Furthermore the line `self.range.get_latest()` does nothing as it just reads the value from the cache and throws it away

qcodes/instrument_drivers/yokogawa/GS200.py
FarBo: > self.range is an alias for either self.voltage_range or self.current_range
True. The problem of this approach is that it always assumes we initialize the instrument in the voltage mode.

> Futher to that there was an internal cache of the range value stored in _cache_range_value...
This is fair.

> Rather than just creating an alias as the orginal code does it would be better to use a DelegateParameter...
I will try this method

qcodes/instrument_drivers/yokogawa/GS200.py
FarBo: Correct

qcodes/instrument_drivers/yokogawa/GS200.py
FarBo: @jenshnielsen 
Do you think it is a good idea to define the output_level as a delegate parameter like the range?

jenshnielsen: Yes 👍

qcodes/instrument_drivers/yokogawa/GS200.py
FarBo: @jenshnielsen 
Regarding auto_range, as you probably know, the instrument does not support this functionality. So, I don't know how to initialize the instrument with the previous state of auto_mode (True or False). However, I can replace self._auto_range=False with initial_cache_value=False in the auto_range parameter.
Is there a way to remember the previous state or in this case, the logic of initializing auto_mode in False state is inescapable?

jenshnielsen: @farbo since the autorange is not build into the instrument we can freely pick the default and set it with `initial_cache_value`
as you suggest. We should not try to keep state for things that are not set on the insturment.

FarBo: Good point. Thanks

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -3,6 +3,7 @@
 
 from qcodes import VisaInstrument, InstrumentChannel
 from qcodes.utils.validators import Numbers, Bool, Enum, Ints
+from qcodes import DelegateParameter
```
jenshnielsen, line 6: Please use the full path for DelegateParameter.

We try to not use the shortcuts within qcodes as it can result in circular dependencies. 
I think the right import is 
`
from qcodes.instruments.parameter import DelegateParameter
`

jenshnielsen: For bonus points fix the imports of VisaInstrument and InsturmentChannel above

qcodes/instrument_drivers/yokogawa/GS200.py
```diff
@@ -517,6 +518,7 @@ def _set_source_mode(self, mode: str) -> None:
 
         self.write("SOUR:FUNC {}".format(mode))
         # Update the measurement mode
+        self.source_mode.cache.set(mode)
```
jenshnielsen, line 521: Sorry one last thing. we should put a comment explaining why this is needed. 
```
# we set the cache here since `_update_measurement_module` needs the current value which would otherwise only be set after this method exits
```

jenshnielsen:
@FarBo is this ready to merge ?

jenshnielsen (approved):

FarBo:
@jenshnielsen 

I believe it is ready to be merged.

PR #2148, opened 2020-09-11
fix docs and test requirements to minor version

jenshnielsen:
and let dependabot handle keeping them up to date

Note that ~= 1.0.2 matches all version >=1.0.2  but < 1.1.0 (e.g. all patch version on the 1.0 branch)

https://pip.pypa.io/en/stable/user_guide/#understanding-your-error-message


jenshnielsen:
It seems like IPython has already dropped support for python 3.6 so if we want to do this we should either first remove our python 3.6 jobs or be less strict with the ipython versions we require

astafan8:
> first remove our python 3.6 jobs

voting for this :)

astafan8 (approved):

PR #2150, opened 2020-09-11
Fix Keysight B1500 Measure Range Bug

Akshita07:
- Add separate methods to configure measuring Voltage and Current ranges
- Deprecate measure_config method
- Update and add relevant tests

@astafan8


qcodes/tests/drivers/keysight_b1500/b1500_driver_tests/test_b1517a_smu.py
```diff
@@ -37,6 +38,7 @@ def test_snapshot():
     smu.use_high_speed_adc()
     smu.source_config(output_range=VOutputRange.AUTO)
     smu.measure_config(measure_range=IMeasRange.AUTO)
+    smu.measure_config(measure_range=VMeasRange.AUTO)
```
astafan8, line 41: general tests should not use deprecated API. deprecated API should have it's own test for which the deprecation warning is explicitly silenced.

qcodes/tests/drivers/keysight_b1500/b1500_driver_tests/test_b1517a_smu.py
```diff
@@ -89,10 +89,14 @@ def test_v_measure_range_config_raises_type_error(smu):
 
 def test_v_measure_range_config_raises_invalid_range_error(smu):
     msg = re.escape("15000 voltage measurement range is invalid for the "
-                    "device. Use required voltage measurement range constant "
-                    "from any of the following - AUTO, MIN_0V5, MIN_2V, "
-                    "MIN_5V, MIN_20V, MIN_40V, MIN_100V, FIX_0V5, FIX_2V, "
-                    "FIX_5V, FIX_20V, FIX_40V, FIX_100V")
+                    "device. Valid ranges are [<VMeasRange.AUTO: 0>, "
```
astafan8, line 92: asserting for the list of these ranges is perhaps overkill :)

astafan8 (approved):

PR #2157, opened 2020-09-15
install pyqt for python 3.8

jenshnielsen:
Will make the coverage less confusing as python 3.7 and 3.8 should hit all the same tests


astafan8 (approved):

jenshnielsen:
@astafan8 Unfortunatly it does not work (some problem with XVFB needs investigation)

jenshnielsen:
@astafan8 this now works with installing `libxkbcommon-x11` which seems to now has become a dependency for qt after 5.13 
I also bumped the linux version on travis and made the docs build with python 3.8. Can you have a look again?

astafan8 (approved):
eys!

PR #2164, opened 2020-09-18
Constrain all requirements and their dependencies in requirements.txt

jenshnielsen:
This is a proposal to make the requirements more robust. 
Like this all requirements are constrained to a specific minor version and a minimum patch version.

Dependabot will be responsible for keeping everything up to date.

 The file (apart from the last few pyqt related lines is generated with pip freeze after installing qcodes)

I also added broadbean as a dependency since it is unconditionally imported in several drivers


astafan8 (approved):
great approach!

PR #2171, opened 2020-09-18
Increase Hypothesis deadline when running in CI

jenshnielsen:
We seem to be getting more hypothesis deadline timeouts since we upgraded hypothesis. It seems like the simplest solution is to increase the deadline from the default value of 200 ms to something more resonable like 1000 

This pr aims to do that


astafan8 (approved):
looks very neat with these profiles :)

PR #2178, opened 2020-09-21
use pytest.ini settings when testing from within python

jenshnielsen:
Using pytest.ini settings should give a nice speedup as it activates parallel tesing and the other should ensure a better timeout setting for hypothesis

EDIT: The ci profile cannot be found in an external run so this hardcodes the same values 
EDIT2: Note this requires passing the non serial flag when running tests in parallel in the same way as the regular tests does

@astafan8


jenshnielsen:
The tests that are normally marked as serial needs to be executed in serial here too.

jenshnielsen:
I will split out the hypothesis related changes so we can improve stability of the build pipeline

astafan8 (approved):

jenshnielsen:
replaced by github actions pr

PR #2184, opened 2020-09-21
handling TypeError in write_raw (pyvisa 1.11)

ctschnur:
After updating qcodes to the following version,  
```
$ pip show qcodes
Name: qcodes
Version: 0.17.0+139.g11417d944
Summary: Python-based data acquisition framework developed by the Copenhagen / Delft / Sydney / Microsoft quantum computing consortium
Home-page: https://github.com/QCoDeS/Qcodes
[...]
```
I was not able to initialize a keysight device any more (using the driver in `qcodes/instrument_drivers/Keysight/N52xx.py`), 
since it gave me a `TypeError`, more specifically: 

```
  File "qcodes_keysight_example.py", line 43, in <module>
    pna = N5245A("VNA2", "TCPIP0::maip-franck::hislip0,4880::INSTR",
  File "qcodes_keysight_example.py", line 32, in __init__
    super().__init__(name, address,
  File "c:\users\nanospin\misc\qcodes\qcodes\instrument_drivers\Keysight\N52xx.py", line 422, in __init__
    trace1 = self.traces[0]
  File "c:\users\nanospin\misc\qcodes\qcodes\instrument_drivers\Keysight\N52xx.py", line 471, in traces
    trace_num = self.select_trace_by_name(trace_name)
  File "c:\users\nanospin\misc\qcodes\qcodes\instrument_drivers\Keysight\N52xx.py", line 504, in select_trace_by_name
    self.write(f"CALC:PAR:SEL '{trace_name}'")
  File "c:\users\nanospin\misc\qcodes\qcodes\instrument\base.py", line 712, in write
    raise e
  File "c:\users\nanospin\misc\qcodes\qcodes\instrument\base.py", line 708, in write
    self.write_raw(cmd)
  File "c:\users\nanospin\misc\qcodes\qcodes\instrument\visa.py", line 212, in write_raw
    nr_bytes_written, ret_code = self.visa_handle.write(cmd)
TypeError: ('cannot unpack non-iterable int object', 'writing "CALC:PAR:SEL \'CH1_S11_1\'" to <N5245A: VNA2>')
```

So I tracked it down and indeed, in my case `self.visa_handle.write(cmd)` returned only an `int`, not a tuple!
```
> c:\users\nanospin\appdata\local\continuum\anaconda3\envs\qcodes_sandbox\lib\site-packages\pyvisa\ctwrapper\functions.py(2797)write()
-> return return_count.value, ret
(Pdb) ll
2772    def write(library, session, data):
2773        """Write data to device or interface synchronously.
2774
2775        Corresponds to viWrite function of the VISA library.
2776
2777        Parameters
2778        ----------
2779        library : ctypes.WinDLL or ctypes.CDLL
2780            ctypes wrapped library.
2781        session : VISASession
2782            Unique logical identifier to a session.
2783        data : bytes
2784            Data to be written.
2785
2786        Returns
2787        -------
2788        int
2789            Number of bytes actually transferred
2790        constants.StatusCode
2791            Return value of the library call.
2792
2793        """
2794        return_count = ViUInt32()
2795        # [ViSession, ViBuf, ViUInt32, ViPUInt32]
2796        ret = library.viWrite(session, data, len(data), byref(return_count))
2797 ->     return return_count.value, ret
(Pdb) return_count.value
25
```
I then edited `write_raw` in `qcodes/instrument/visa.py` to account for this case (and not break): 
```
 def write_raw(self, cmd: str) -> None:
        """
        Low-level interface to ``visa_handle.write``.

        Args:
            cmd: The command to send to the instrument.
        """
            self.visa_log.debug(f"Writing: {cmd}")
            response = self.visa_handle.write(cmd)
            nr_bytes_written = None

            if type(response) is tuple and len(response) == 2:
                nr_bytes_written = response[0]
                ret_code = response[1]
                self.check_error(ret_code)
            else:
                nr_bytes_written = response
```
I minimally altered the code to circumvent this error. Now the connection to the device works again.

### DIFF
--- a/qcodes/instrument/visa.py
+++ b/qcodes/instrument/visa.py
@@ -210,27 +210,10 @@ class VisaInstrument(Instrument):
         with DelayedKeyboardInterrupt():
             self.visa_log.debug(f"Writing: {cmd}")
             response = self.visa_handle.write(cmd)
-            self.check_error(ret_code)
 
-    def write_raw(self, cmd: str) -> None:
-        """
-        Low-level interface to ``visa_handle.write``.
-
-        Args:
-            cmd: The command to send to the instrument.
-        """
-        with DelayedKeyboardInterrupt():
-            self.visa_log.debug(f"Writing: {cmd}")
-            response = self.visa_handle.write(cmd)
-            nr_bytes_written = None
-
-            if type(response) is tuple and len(response) == 2:
-                nr_bytes_written = response[0]
+            if isinstance(response, tuple) and len(response) == 2:
                 ret_code = response[1]
                 self.check_error(ret_code)
-            else:
-                nr_bytes_written = response
-
 
     def ask_raw(self, cmd: str) -> str:
         """
--- a/qcodes/instrument/visa.py
+++ b/qcodes/instrument/visa.py
@@ -209,8 +209,28 @@ class VisaInstrument(Instrument):
         """
         with DelayedKeyboardInterrupt():
             self.visa_log.debug(f"Writing: {cmd}")
-            nr_bytes_written, ret_code = self.visa_handle.write(cmd)
+            response = self.visa_handle.write(cmd)
             self.check_error(ret_code)
+
+    def write_raw(self, cmd: str) -> None:
+        """
+        Low-level interface to ``visa_handle.write``.
+
+        Args:
+            cmd: The command to send to the instrument.
+        """
+        with DelayedKeyboardInterrupt():
+            self.visa_log.debug(f"Writing: {cmd}")
+            response = self.visa_handle.write(cmd)
+            nr_bytes_written = None
+
+            if type(response) is tuple and len(response) == 2:
+                nr_bytes_written = response[0]
+                ret_code = response[1]
+                self.check_error(ret_code)
+            else:
+                nr_bytes_written = response
+
 
     def ask_raw(self, cmd: str) -> str:
         """


jenshnielsen:
@ctschnur thanks for the patch. Can you check the version of pyvisa that you are using. I believe this is due to changes in pyvisa 1.11. QCoDeS only works with pyvisa <= 1.10 for now. We will make sure to take your patch into account when adding support for 1.11

ctschnur:
Thanks, I was indeed using a newer pyvisa version: 
```
$ pip show pyvisa
Name: PyVISA
Version: 1.11.0
Summary: Python VISA bindings for GPIB, RS232, TCPIP and USB instruments
Home-page: https://github.com/pyvisa/pyvisa
```

In this newer version, the following function returns just an `int`: 
```
> c:\users\nanospin\appdata\local\continuum\anaconda3\envs\qcodes_sandbox\lib\site-packages\pyvisa\resources\messagebased.py(199)write()
-> return count
(Pdb) ll
159         def write(
160             self,
161             message: str,
162             termination: Optional[str] = None,
163             encoding: Optional[str] = None,
164         ) -> int:
165             """Write a string message to the device.
166
167             The write_termination is always appended to it.
168
169             Parameters
170             ----------
171             message : str
172                 The message to be sent.
173             termination : Optional[str], optional
174                 Alternative character termination to use. If None, the value of
175                 write_termination is used. Defaults to None.
176             encoding : Optional[str], optional
177                 Alternative encoding to use to turn str into bytes. If None, the
178                 value of encoding is used. Defaults to None.
179
180             Returns
181             -------
182             int
183                 Number of bytes written.
184
185             """
186             term = self._write_termination if termination is None else termination
187             enco = self._encoding if encoding is None else encoding
188
189             if term:
190                 if message.endswith(term):
191                     warnings.warn(
192                         "write message already ends with " "termination characters",
193                         stacklevel=2,
194                     )
195                 message += term
196
197             count = self.write_raw(message.encode(enco))
198
199  ->         return count
(Pdb) count
25
(Pdb)
```
so that `self.visa_handle.write(cmd)` also returns just an `int`.

jenshnielsen:
I suggest that you downgrade pyvisa to 1.10.1 then. 
I don't think we can merge your pr as is. Significantly more work is required to properly support 1.11 including fixing testing with pyvisa-sim that also depends on the return code. But we will use your patch as a starting point when we get to add support for 1.11

PR #2187, opened 2020-09-22
rework 2d tests to sweep over two separate parameters

jenshnielsen:
Sweeping over the same parameter in both dimensions makes lille sense and should probably be an error.
@astafan8


astafan8 (approved):
this looks to me as an oversight when implementing the tests :) great!

> Sweeping over the same parameter in both dimensions makes lille sense and should probably be an error.

I agree. Not sure if it's worth implementing an error for it, but it definitely does not make sense to me at all :)

jenshnielsen:
> I agree. Not sure if it's worth implementing an error for it, but it definitely does not make sense to me at all :)

Ok, lets leave out the error for now

PR #2198, opened 2020-09-24
remove pyzmq from dependencies in setup.py

jenshnielsen:
Since it is not used directly by qcodes


astafan8 (approved):

jenshnielsen:
As pointed out by @basnijholt

PR #2199, opened 2020-09-24
move mypy.ini and pytest.ini to setup.cfg

basnijholt:
Reduces the number of files in the root of the repo.


jenshnielsen:
I have not been following the development of this part of the packaging that closely but are things not supposed to mov to pyproject.toml eventually?

astafan8:
latest i know (been questing this qute recently) is that setup.cfg should be used only for super simple/trivial declarations. everything is indeed moving to toml, pytest since version 6 supports is completely. mypy still doesn't support toml properly so with that we should wait (or merge it into setup.cfg for now).

astafan8:
on the other hand, if this just works, then it's fine. at later point we'll just migrate from single file setup.cfg to pyproject.toml and that's it

setup.cfg
```diff
@@ -1,3 +1,73 @@
+[pytest]
```
jenshnielsen, line 1: should this not be `[tool:pytest]` https://docs.pytest.org/en/stable/customize.html#setup-cfg as below

basnijholt: You are correct, I've fixed it.

jenshnielsen: There isanother section `[tool:pytest]` below

jenshnielsen (approved):

astafan8 (approved):

PR #2201, opened 2020-09-25
Add precommit hook to ensure consistent line endings

jenshnielsen:
and remove the check for the same using pylint


jenshnielsen:
This unfortunatly does not understand that .bat files should have windows style line endings

astafan8 (approved):

jenshnielsen:
This does not seem like the way forward. The gitattributes should take care of this

PR #2200, opened 2020-09-25
add pyupgrade to pre-commit

basnijholt:
As a followup from https://github.com/QCoDeS/Qcodes/pull/1945.

This adds pyupgrade as pre-commit filter.

This PR is blocked by
- #2201
- #2202
- #2203
- #2204
- #2205
- #2206
- #2207
- #2208
- #2209
- #2210
- #2211

PR #2207, opened 2020-09-25
pyupgrade qcodes/tests

basnijholt:
Applying `pyupgrade` to a fraction of all files as suggested in https://github.com/QCoDeS/Qcodes/pull/2200#issuecomment-698800644.

This is in preparation for adding `pyupgrade` as a `pre-commit` filter https://github.com/QCoDeS/Qcodes/pull/2200.

### DIFF
--- a/qcodes/tests/drivers/test_tektronix_AWG70000A.py
+++ b/qcodes/tests/drivers/test_tektronix_AWG70000A.py
@@ -118,7 +118,7 @@ def test_SML_successful_generation_vary_length(N):
     ejs = [0]*N
     ejt = [0]*N
     goto = [0]*N
-    wfm_names = [['pos{}ch{}'.format(pos, ch)
+    wfm_names = [[f'pos{pos}ch{ch}'
                   for ch in range(1, 3)] for pos in range(N)]
 
     seqname = 'seq'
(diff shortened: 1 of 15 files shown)


qcodes/tests/drivers/test_tektronix_AWG70000A.py
```diff
@@ -185,7 +185,7 @@ def test_seqxfile_from_fs(forged_sequence):
     path_to_schema = auxfiles.__file__.replace('__init__.py',
                                                'awgSeqDataSets.xsd')
 
-    with open(path_to_schema, 'r') as fid:
+    with open(path_to_schema) as fid:
```
jenshnielsen, line 188: While this is identical I am not convinsed that removing arguments just because they are the default value is an improvment

jenshnielsen (approved):

qcodes/tests/drivers/test_tektronix_AWG70000A.py
astafan8: agree with Jens.

basnijholt: > I am not convinced that removing arguments just because they are the default value is an improvement

It doesn't always do this, just with `open` AFAIK (where I think it's fine).

Unfortunately, this option isn't configurable.

PR #2215, opened 2020-09-26
Port travis ci to Github actions and split into multiple jobs

jenshnielsen:
This should eventually replace the travis workflow.
TODO 

- [x] caching of pip dependencies
- [x] generation of dbs for upgrade script
- [x] Docs build (should probabably be done in a different build)
- [x] test with qcodes.test() (should probably also be a different build)
- [x] refactor to reusable components
- [x] coverage upload
- [x] upload docs artifacts
- [x] Ensure triggers match (branch, pr, tag) Now matches azure pipelines
- [x] Should we port over the run of asv or delete it? (no can run it locally to hunt down regressions, and prehaps we should setup a dedi
- [x] Should we port over the pylint check for line endings. May not be needed now that we set gitattributes
- [x] Deadlock (maybe solved by switching to -n=1?) There seems to be a deadlock with --n=0 and a faliure on the subscriber related tests. This is worked around by now by not useing -n=0. We now use pytest rerunfailure on all these tests. In the long run we can probably use this to get rid of the run these test in serial.


jenshnielsen:
It seems like the job hangs on test failure some times. https://github.com/QCoDeS/Qcodes/runs/1170320665?check_suite_focus=true Not clear why

astafan8:
>  Should we port over the run of asv or delete it?

for 2 years we haven't used it, so "concervating" (or simply removing - i have to learn to let go ;) ) our asv setup is a good idea. one thing about this asv is that you get much more benefit if you host the results of asv runs somewhere, as in, if you collect and accumulate the statistics. otherwise, it's to some extent a waste of ci resources. one the flipside, the current setup does allow to just run asv locally and let it find a "commit" that introduces a performance regression.

astafan8:
> Should we port over the pylint check for line endings. May not be needed now that we set gitattributes

indeed, no need to port.

jenshnielsen:
Note this is waiting on a travis job that will never get run since .travis.yml has been deleted. Once this is merged we can remove the required status for travis and add it for github actions

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,60 @@
...
+          ${{ runner.os }}-pip-
+
+    - name: install requirements.txt
+      run: |
+        pip install -r requirements.txt --use-feature=2020-resolver
+    - name: install docs_requirements.txt
+      run: |
+        pip install -r docs_requirements.txt --use-feature=2020-resolver
+    - name: install qcodes
+      run: pip install . --use-feature=2020-resolver
```
astafan8, line 50: is it a useful idea to extract common steps across the "workflows" into some kind of "install ubunu, python, pip, qcodes minimum" "step grouping" thing? is that even possible? azure pipelines had these "templates", smth where the pipeline can take steps from

.github/workflows/pytest.yaml
```diff
@@ -0,0 +1,69 @@
...
+        pip install -r requirements.txt --use-feature=2020-resolver
+    - name: install test_requirements.txt
+      run: |
+        pip install -r test_requirements.txt --use-feature=2020-resolver
+    - name: generate test dbs
+      uses: ./.github/actions/generate-test-dbs
+    - name: install qcodes
+      run: pip install . --use-feature=2020-resolver
+    - name: Run Mypy
+      run: mypy qcodes
```
astafan8, line 56: i haven't read about actions enough, hence a quick question - if mypy step fails, will the subsequent test steps be executed?

also, what do we want? i remember myself prefering to run tests despite mypy problems being useful for getting an overview of errors which is potentially useful for reducing number of pushes and hence ci triggers ))

.github/actions/generate-test-dbs/action.yml
```diff
@@ -0,0 +1,27 @@
...
+description: "Generate qcodes dbs used for upgrade tests"
+runs:
+  using: "composite"
+  steps:
+    - name: generate test dbs
+      run: |
+        pip install -e .
+        pwd
+        cd ..
+        git clone https://github.com/QCoDeS/qcodes_generate_test_db.git
```
astafan8, line 11: is it more beneficial to use action's checkout (or clone?) step instead of `git clone ...` manually?

.github/actions/generate-test-dbs/action.yml
```diff
@@ -0,0 +1,27 @@
...
+        python generate_version_0.py
+        python generate_version_1.py
+        python generate_version_2.py
+        python generate_version_3.py
+        python generate_version_4a.py
+        python generate_version_4.py
+        python generate_version_5.py
+        python generate_version_6.py
+        python generate_version_7.py
+        python generate_version_8.py
```
astafan8, line 22: is it worth the effort to use `actions/cache@v2` to maintain a cache of DB files as well?

astafan8 (approved):
great to see this!

just left a few curiosity questions

also - when do you think we'll migrate from azure pipelines as well? :)

.github/actions/generate-test-dbs/action.yml
jenshnielsen: Might savew 40 sec or so. I would say its a nice upgrade but can be done separately.

.github/actions/generate-test-dbs/action.yml
jenshnielsen: It's not currently possible to use actions from whinin a composable run step as I understand. So I suggest to leave as is for now

.github/workflows/pytest.yaml
jenshnielsen: As far as I understand it will terminate immediately.

.github/workflows/docs.yaml
jenshnielsen: As far as I understand it is possbile to create composable actions of one or more run stesps but not to put actions into other composable actions. This limits a bit what we can do. E.g. that is why the

jenshnielsen:
@astafan8 Lets discuss Azure pipelines offline

.github/workflows/pytest.yaml
astafan8: ok, we can look at it later if needed

PR #2220, opened 2020-09-28
Add allow_changes option to set_to method, and add another context manager for restoring current value

mgunyho:
This PR adds the following changes and new features related to the Parameter `set_to` method:

- Add the `allow_changes` option (False by default), which makes a parameter not settable within the context
- Add another context manager, `restore_at_exit`, which is the same as `p.set_to(p.cache())`.
- Ensure that the parameter value gets properly reverted to its original value when exiting the context when it's modified inside but was not changed when entering
- Remove usage of the old `get_latest` method

This issue came up originally when I noticed that the parameter value was not restored in the following code:

```py
p = qc.Parameter("foo", set_cmd=None, get_cmd=None)
p(42)
with p.set_to(42):  # same as current value
    p(123)
p()  # expected 42 but is 123
```
This behavior is also fixed by this PR (when using `allow_changes=True`)


jenshnielsen:
To me it seems like 

```
with p.set_to(42):
	p(123) <- trying to set within a set to context should raise
```
Should simply be an error:
As the set_to context means within this block the value of this parameter is whatever it is set to and nothing else

jenshnielsen:
There could be a usecase for what you descibe but that context manager should imho be called something else

mgunyho:
I come across such situations where the parameter is modified inside the context very often when messing around with experiments. Typically I'll have a jupyter cell with something like

```py
with ExitStack() as stack:
    #stack.enter_context(param1.set_to(some_value1))
    stack.enter_context(param2.set_to(some_value2))
    #stack.enter_context(param3.set_to(some_value3))
    
    do_ndim_sweep(
        #param1,
        param2,
        param3,
        some_values
    )
```
I may have many lines with `set_to` and many different sweep parameters that I'm commenting in and out while trying to figure out why something doesn't work. Often this results in sweeping over some parameter which is also set using a context manager, but I think that should be fine. The sweep function might also temporarily alter some parameter values using more context managers (for example to calibrate something). Maybe this is slight abuse of the context manager, but at least for me it would be quite considerably less useful if the value couldn't be changed inside.

I also think it would be quite complicated to implement this behavior, if it would need to modify the `set` method of the parameter.

mgunyho:
Maybe there could be another context manager, like "lock" or "freeze" that would temporarily prevent modifications to the parameter. But for me the current use case is important.

jenshnielsen:
In my oppinion `set_to` impliezes freeze so I would much rather add a new context mananger for you use case with a better name 

Making the parameter temorarily non settable is very easy you will just need to change the `_settable` flag on it

mgunyho:
Thinking about it a bit more, I could come up with reasonable use cases for all four combinations of entering the context with/without changing the value and preventing/allowing it to change inside.

Maybe `set_to` could be extended so that it can be called without any value (same as entering with the current value), and with a kwarg `freeze` (maybe with the default value of True), which would prevent changing. This way we don't need four different context managers for all cases. The common feature in all of these is that the parameter value gets returned to its original value after exiting (which is not the case with the current implementation).

astafan8:
hey @mgunyho , we had an offline discussion with @jenshnielsen and we decided to suggest the following:
1. improve the behavior of `set_to` such that inside the context it is not possible to set the parameter to any different value - this is justified by the care about users, even when they fiddle around with code of the measurement, qcodes still should be there to prevent unnoticed mistakes - i hope you'll agree that it's easier for the user to think that inside the `set_to` context the value of the parameter is NOT supposed to change.
2. in order to enable your use case of allowing users to change the value of the "set-to" parameter, let's indeed introduce a flag - we suggest that flag to be called `allow_changes` that is set to `False` by default -- so `set_to(new_value, allow_changes: bool = False)` - would you agree?
3. the use case of only reverting __the current value__ of the parameter can be nicely achieved via `with p.set_to(p.cache()): ...` - is that good enough?

mgunyho:
Hi

> it's easier for the user to think that inside the set_to context the value of the parameter is NOT supposed to change

That's fair enough, although at least we have not automatically assumed so in our use until now.


> we suggest that flag to be called `allow_changes`

I would prefer the name `freeze` or `lock`, because it's less typing, but still quite well understandable in my opinion. But in either case the flag is a good idea, and it's ok that the default value is to not allow changes.

> `with p.set_to(p.cache()):`

I would prefer the version with no args. This version might be hard to read if the variable name is long. Also consider the case I discussed above, with `ExitStack()`. Then, if I want to add a new `set_to` statement by duplicating an existing line, I would have to change the variable name in two places, and this can be prone to errors.


I worked a bit on implementing this functionality, and have now pushed it to [this branch](https://github.com/mgunyho/Qcodes/tree/parameter_set_to_freeze). Should I add those changes to this PR or create a new one? The flag is now called `freeze`, but I can rename it. I've also implemented the no-args version, but that can be left out if you really think that it's not good.

astafan8:
> I would prefer the version with no args. This version might be hard to read if the variable name is long. Also consider the case I discussed above, with ExitStack(). Then, if I want to add a new set_to statement by duplicating an existing line, I would have to change the variable name in two places, and this can be prone to errors.

This is fair. Then let's take the middle ground of having a new context manager called `.reset_at_exit(allow_changes=False)` (the same flag is included)?

> I worked a bit on implementing this functionality, and have now pushed it to this branch. Should I add those changes to this PR or create a new one?

Thank you! that's great! i think it's more convenient to use this pull-request, just push those changes to the branch that is attached to this PR (or re-target this PR to that other branch, if that's more convenient)

> The flag is now called freeze, but I can rename it.

yes, please, `allow_changes` (it is more descriptive of what happens) :)

> I've also implemented the no-args version, but that can be left out if you really think that it's not good.

yeah, `with p.set_to():` it's less readable/intuitive of what is going on. So, instead of having the no-args version, let's go for another context manager - `.reset_at_exit(allow_changes=True)`.

mgunyho:
Alright, I'll update this branch.

For our use case, where we may want to set a parameter temporarily for the duration of a sweep, but it might be altered during the sweep to e.g. do some calibration, having two different context managers would then require having twice as many lines, which is again susceptible to typos. But maybe that's quite rare even in our case, we'll have to see.

Do you think `reset_at_exit` is the best possible name for the method? For a new user not familiar with the `with` statement, it might be unclear that something is being _exited_ at the end of the block. How about something like `keep_value`?

mgunyho:
I've updated the PR description.

Some TODOs/questions:

1. Are the tests extensive enough? In particular, there's no test for `reset_at_exit` when the parameter cache is not initialized, but I'm not sure how to clear the parameter cache.
1. These features should be mentioned in the documentation/example notebooks, I haven't done that yet. If you have any particular spot in mind where these could be mentioned, let me know.
1. Should the `_original_value` be set only when entering the context? Now, it is in principle possible to get some weird behavior if the parameter value is changed between creating and entering the context manager (although this is very much of a fringe case): 
    ```py
    p = qc.Parameter("foo", set_cmd=None, get_cmd=None)
    p(3)
    ctx = p.set_to(5)

    p(123)
    with ctx:
        assert p() == 5
    
    assert p() == 3  # although someone may assume that p() == 123 here
    ```
    The same goes for `_parameter_was_settable`.

mgunyho:
Another small thing that came to mind: if parameter B is a delegate of A, and I enter a context where I make B not settable, it is still possible to modify the value of A and thus the value of B. But maybe that's a problem for another PR (if it's a problem at all). At first thought it feels like trying to implement something like this could be quite complicated in general for multiple nested DelegateParameters.

astafan8:
> Do you think reset_at_exit is the best possible name for the method? For a new user not familiar with the with statement, it might be unclear that something is being exited at the end of the block. How about something like keep_value?

well, users that are "not familiar with the with statement" but are using python should take their time to get familiar with one of the most fantastic features of python - the context managers and the `with` statement :) next, of course `reset_at_exit` is likely not the best possible name, however it is very explicit in expressing the behavior that the user should expect - the value of the parameter will be reset at the exit of the context manager. the `keep_value` option doe snot explicitly mention that the parameter's value will be reset once the context is exited, moreover, "keep value" may confuse users to think that inside the context they should not change the value of the parameter which is opposite from your feature request, and is definitely possible with allow_changes=True. So, keep_value is not better than reset_at_exit. I don't have any other splendid ideas, so we might as well just stick with reset_at_exit?

> These features should be mentioned in the documentation/example notebooks, I haven't done that yet. If you have any particular spot in mind where these could be mentioned, let me know.

great question! I think that for this PR it is good enough make sure that set_to and reset_at_exit have clear docstrings explaining the behavior and usage.

> Should the _original_value be set only when entering the context? Now, it is in principle possible to get some weird behavior if the parameter value is changed between creating and entering the context manager (although this is very much of a fringe case):

I think this is a great test case! indeed, the "original_value" should be taken at the moment the context is entered, and NOT at the moment the context manager is initiated. So please feel free to add this exact test, and modify the implementation to capture the original value at the enter of the context manager.

> Another small thing that came to mind: if parameter B is a delegate of A, and I enter a context where I make B not settable, it is still possible to modify the value of A and thus the value of B. But maybe that's a problem for another PR (if it's a problem at all). At first thought it feels like trying to implement something like this could be quite complicated in general for multiple nested DelegateParameters.

Again, great call! However, let's keep this for another PR, this needs a bit of thinking, and let's not pollute the discussion inside this particular PR.

astafan8 (commented):
just reviewed the PR - it looks very good! looking forward to the changes after our discussion in the comments

mgunyho:
I've added the changes about entering the context.

You're correct, the `with` statement is important, and the more people learn about it the better (although from my experience it takes a while before beginners understand its usefulness).

How about `restore_at_exit`? "Reset" could imply something like resetting to the initial value of the parameter.

astafan8:
> How about restore_at_exit?

yes, that's probably better. my english is not that good so i had to google differences in meanings of the words.

mgunyho:
Alright, to me this is now good to merge. Let me know if there's something to change.

astafan8 (approved):

jenshnielsen (approved):
Thanks @mgunyho

PR #2222, opened 2020-09-28
Remove serial jobs on linux

jenshnielsen:
This removes serial jobs from all linux runners where it seems like there is no issue running all the tests in parallel


qcodes/tests/dataset/test_nested_measurements.py
```diff
@@ -167,6 +167,7 @@ def subscriber(results: List[Tuple[VALUE]], length: int,
     return subscriber
 
 
+@pytest.mark.flaky(reruns=5)
 @pytest.mark.serial
```
astafan8, line 171: we should alsao remove this marker in this PR, right?

jenshnielsen: Yes if this actually works (as you can see from the outcome it probably does not)

jenshnielsen: Since the serial jobs are still needed for azure pipelines on windows this will be left as is

azure-pipelines.yml
```diff
@@ -95,8 +95,7 @@ jobs:
       - bash: |
           source activate qcodes &&
           cd qcodes &&
-          pytest -m "not serial" --junitxml=test-parallel-results.xml --cov=qcodes --cov-report=xml --cov-config=.coveragerc --hypothesis-profile ci &&
-          pytest -n 0 --dist no -m serial --junitxml=test-serial-results.xml --cov-append --cov=qcodes --cov-report=xml --cov-config=.coveragerc --hypothesis-profile ci
+          pytest --junitxml=test-parallel-results.xml --cov=qcodes --cov-report=xml --cov-config=.coveragerc --hypothesis-profile ci &&
```
astafan8, line 98: it should be the other way around , right? the windows part should still run serial/not-serial

jenshnielsen: Yes don't know how that happened :(

jenshnielsen:
@astafan8 ok to merge?

astafan8 (approved):

PR #2232, opened 2020-10-03
Rewrite parameter tests as pytest style tests.

jenshnielsen:
Mostly done automatically by unittest2pytest

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -3,6 +3,7 @@ pytest-cov~=2.10.1
 pytest~=6.1.0
 pytest-rerunfailures~=9.1.1
 pytest-xdist~=2.1.0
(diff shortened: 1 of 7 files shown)


test_requirements.txt
```diff
@@ -3,6 +3,7 @@ pytest-cov~=2.10.1
 pytest~=6.1.0
 pytest-rerunfailures~=9.1.1
 pytest-xdist~=2.1.0
+pytest-mock~=3.3.1
```
astafan8, line 6: is it not part of pytest already since some version?

astafan8 (approved):

test_requirements.txt
jenshnielsen: I don't think so

PR #2235, opened 2020-10-03
convert helpers tests to pytest style and break up into individual files

jenshnielsen:
And integrate the json tests from test helpers with the onces from test_json


qcodes/tests/helpers/test_strip_atters.py
```diff
@@ -0,0 +1,57 @@
+import pytest
```
astafan8, line 1: typo in the file name - test_strip_attrs

astafan8 (approved):

PR #2239, opened 2020-10-05
Fix some of the warnings triggered by numpy 1.19

jenshnielsen:
Only use instantiable types as dtypes when creating arrays. Using the non concrete types triggers a warning and will be an error in the future


qcodes/utils/validators.py
```diff
@@ -737,9 +737,9 @@ def __init__(self, min_value: Optional[numbertypes] = None,
     def valid_values(self) -> Tuple[np.ndarray]:
         shape = self.shape
         if shape is None:
-            return (np.array([self._min_value], dtype=self.valid_types[0]),)
+            return (np.array([self._min_value], dtype=np.int32),)
```
astafan8, line 740: do i understand correctly that the code was already assuming that the min/max values are integers regardless of whether valid_types argument was passed or not, and this PR does not fix that problem, it only gets rid of the numpy warning?

astafan8 (approved):

qcodes/utils/validators.py
jenshnielsen: The changes makes no attempt of fixing this

jenshnielsen: But yes I should think of a better way to do this. Will probably have to hard code a fall back for non concrete types

PR #2251, opened 2020-10-07
Run github action with python 3.9

jenshnielsen:
This seems to be working fine. A few packages needed to be build from source but they are cached so the speed impact is minimal.
@astafan8


.github/workflows/docs.yaml
```diff
@@ -38,7 +40,7 @@ jobs:
         path: ${{ steps.prepare-cache-pip.outputs.cache-dir }}
         key: ${{ runner.os }}-${{ matrix.python-version }}-pip-${{ hashFiles('**/requirements.txt') }}-${{ hashFiles('**/docs_requirements.txt') }}-${{ hashFiles('**/setup.py') }}
         restore-keys: |
-          ${{ runner.os }}-pip-
+          ${{ runner.os }}-${{ matrix.python-version }}-pip-
```
astafan8, line 43: that's a good one!

.github/workflows/docs.yaml
```diff
@@ -25,6 +25,8 @@ jobs:
       uses: ./.github/actions/setup-ubuntu-latest-xvfb
     - name: install pandoc ubuntu
       run: sudo apt install pandoc
+    - name: install libhdf5-dev ubuntu
+      run: sudo apt install libhdf5-dev
```
astafan8, line 29: so docs need only hdf5 while the tests need a few more packages for dealing with XML? interesting

astafan8 (approved):
python 3.9 - ready! :)

.github/workflows/docs.yaml
jenshnielsen: Yes some of the awg tests uses lxml which needs these extra packages to be build from source

PR #2253, opened 2020-10-07
Driver for Keysight E4980A LCR meter

liangosc:
- Driver for Keysight E4980A LCR meter
- QCoDeS example notebook for the new driver

@sohailc


qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -0,0 +1,333 @@
...
+            docstring="Executes SHORT correction based on all frequency points."
+        )
+
+        self.add_parameter(
+            "short_state",
+            get_cmd=":CORRection:SHORt:STATe?",
+            set_cmd=":CORRection:SHORt:STATe {}",
+            val_mapping={
+                'off': 0,
+                'on': 1,
```
astafan8, line 193: please use create_on_off_val_mapping function :)

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -0,0 +1,333 @@
...
+        self.write(f":FUNCtion:IMPedance {measurement_function.name}")
+
+    def clear_status(self) -> None:
+        """
+        Clears the following:
+            • Error Queue
+            • Status Byte Register
+            • Standard Event Status Register
+            • Operation Status Event Register
+            • Questionable Status Event Register (No Query)
```
astafan8, line 325: let's avoid unicode symbols? so that sphinx can render this as a bullet point list in the docs

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -0,0 +1,333 @@
...
+            get_cmd=self._get_complex_impedance,
+        )
+
+        self.add_parameter(
+            "measurement_function",
+            get_cmd=":FUNCtion:IMPedance?",
+            set_cmd=self._set_measurement
+        )
+
+        self.add_parameter(
```
astafan8, line 256: in this intricate arrangement of parameters - how would users use them with qcodes Measurement? what should they pass to register_parameter ?

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
liangosc: Now it's using `create_on_off_val_mapping`. Thanks~

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
liangosc: Good point. I have removed the "dots", and I "think" I fixed all the other sphinx error.

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -0,0 +1,313 @@
...
+            get_cmd=":FUNCtion:IMPedance:RANGe?",
+            set_cmd=":FUNCtion:IMPedance:RANGe {}",
+            unit='Ohm',
+            vals=Enum(0.1, 1, 10, 100, 300, 1000, 3000, 10000, 30000, 100000),
+            docstring="Selects the impedance measurement range, also turns "
+                      "the auto range function OFF."
+        )
+
+        self.add_parameter(
+            "system_errors",
```
astafan8, line 236: please make this a bound method as opposed to a parameter - "getting errors" is not a feature that is meant to be wrapped into a parameter

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -0,0 +1,313 @@
...
+        Returns the impedance in the format of (R, X), where R is the
+        resistance, and X is the reactance.
+        """
+        measurement = self.ask(":FETCH:IMPedance:CORRected?")
+        r, x = [float(n) for n in measurement.split(",")]
+        measurement_pair = MeasurementPair(
+            name="RX",
+            names=("resistance", "reactance"),
+            units=("Ohm", "Ohm")
+        )
```
astafan8, line 272: shall the carefully made definitions from E4980AMeasurements be used here? :)
```suggestion
        measurement_pair = E4980AMeasurements.RX
```

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -0,0 +1,313 @@
...
+
+    @property
+    def measure_impedance(self) -> MeasurementPair:
+        return self._get_complex_impedance()
+
+    @property
+    def measurement(self) -> MeasurementPair:
+        return self._measurement()
+
+    def _get_complex_impedance(self) -> MeasurementPair:
```
astafan8, line 261: what is the difference between using `measure_impedance` and `measurement` with RX? specifically, is there a reason for them having slightly different instrument commands? if not, then why not let `measure_impedance` resuse `measurement` in its implementation?

astafan8 (commented):
last batch of comments :)

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
liangosc: Fixed. now it's a method instead of a parameter.

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
liangosc: I had the same thoughts but decided not to do it because: if I do `measurement_pair = E4980AMeasurements.RX`, and take a measurement, the class `E4980AMeasurements` will be modified -- the `RX` instance will have a pair of measurement value instead of (0.0). No big deal because it won't affect anything, but I don't like the fact it's muted.

astafan8: ok, that's fine

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
liangosc: They are doing the same thing. The reason I have the stand alone impedance measurement here is to provide a shortcut without change the "main" measurement function. I guess this is why the instrument have both at the same time.

astafan8 (approved):

PR #2263, opened 2020-10-09
Notebook: ParameterWithSetpoints for buffered sweep

jenshnielsen:
An example of how buffered sweep could be done with parameter with setpoints


astafan8 (approved):
The more use-case-driver tutorials - the better :)

Remarks:
* the notebook lacks description of what it tries to explain, it seems like only people who requested this example will understand it :( Could we a bit more description to the top of the notebook?
* i think the notebook will read better if DummyArray class is not there, and ParameterWithSetpoints is used explicitly.
* it's unclear why one would change the setpoints of PwS on the DMM, there's a cell saying "or even change it".

PR #2264, opened 2020-10-09
Use dataset shape when loading data with get_parameter_data and dataset cache

jenshnielsen:
Incomple list of TODOS

- [x] Make sure that there are tests for the correct behaviour when shape do not match actual data. Including too much data captured and datasets that are stopped early.


qcodes/dataset/sqlite/queries.py
```diff
@@ -176,11 +176,52 @@ def get_parameter_data(conn: ConnectionPlus,
...
+        if shape is not None:
+            total_len_shape = np.prod(shape)
+            for name, paramdata in one_param_output.items():
+                total_data_shape = np.prod(paramdata.shape)
+                if total_data_shape == total_len_shape:
+                    one_param_output[name] = paramdata.reshape(shape)
+                else:
+                    log.warning(f"tried to set data shape from metadata when "
+                                f"loading but found inconsistent lengths "
+                                f"{total_data_shape} and {total_len_shape}")
```
jenshnielsen, line 221: Make sure that this branch is covered

qcodes/tests/dataset/test_doNd.py
```diff
@@ -563,3 +590,15 @@ def test_do2d_additional_setpoints_shape(
...
+@pytest.mark.usefixtures("temp_exp", "temp_db")
+@given(num_points=hst.integers(min_value=1, max_value=100))
+def test_do1d_sweep(_param_set, _param, num_points):
+
+    start = 0
+    stop = 1
+    delay = 0
+
+    results = do1d(_param_set, start, stop, num_points, delay, _param, do_plot=False)
+    results[0]._shape = {'simple_parameter': (num_points, )}
```
jenshnielsen, line 604: Remove this test?

docs/examples/DataSet/Performing-measurements-using-qcodes-parameters-and-dataset.ipynb
```diff
@@ -1527,6 +1528,79 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Specifying shape of measurement\n",
+    "As the context manager allows you to generate data in any shape with the only restriction being that you supply values for both dependent and independent parameters together it cannot know if the date is to be measured on a grid and as a consequence the Numpy array of data loaded from the dataset may not be in the shape that you expect. `plot_dataset` and `get_data_as_pandas_dataframe`  contains logic that can detect the shape of the data measured at load time. However, if you know the shape of the measurement that you are going to perform up front you may chose to specify it before initializing the measurement. \n",
+    "\n",
+    "`dataset.get_parameter_data` and `dataset.cache.data` automatically makes use of this information to return shaped data when loaded from the database.\n",
```
astafan8, line 1538: i believe it is out of scope for this PR to rework plot_dataset to rely on shape metadata whenever possible, right? because at the moment it ALWAYS guesses because it relies on get_data_by_id which in turn does rely on get_parameter_data however it explicitly flattens any shapes, see here https://github.com/qcodes/Qcodes/blob/master/qcodes/dataset/data_export.py#L90

qcodes/tests/dataset/test_data_set_cache.py
```diff
@@ -331,6 +333,150 @@ def test_cache_complex_array_param_in_1d(experiment, DAC, channel_array_instrume
...
+        expected_shapes,
+        n_points_measured,
+        param_data_trees,
+        cache_expected_too_small=False
+):
+    assert sorted(cache_data_trees.keys()) == sorted(expected_shapes.keys())
+    for outer_key, cache_data_tree in cache_data_trees.items():
+        exshape = expected_shapes[outer_key]
+        if len(exshape) > 2:
+            array_shape = np.prod(exshape[2:])
```
astafan8, line 557: is this "array" "shape"? it seems like a "size" (in numpy terms) without the first axis which is likely the measured parameter? or i am misunderstaanding?
```suggestion
            array_shape = np.prod(exshape[2:])
```

qcodes/tests/dataset/test_data_set_cache.py
```diff
@@ -331,6 +333,150 @@ def test_cache_complex_array_param_in_1d(experiment, DAC, channel_array_instrume
...
+        else:
+            array_shape = 1
+
+        for inner_key, cache_data in cache_data_tree.items():
+            if not cache_expected_too_small:
+                assert cache_data.shape == exshape
```
astafan8, line 563: do i understand it right?
```suggestion
            if not cache_expected_too_small:
                assert cache_data.shape == exshape
            else:
                assert cache_data.size > np.prod(exshape)  # sanity check of cache_expected_too_small flag usage
```

qcodes/tests/dataset/test_data_set_cache.py
```diff
@@ -331,6 +333,150 @@ def test_cache_complex_array_param_in_1d(experiment, DAC, channel_array_instrume
...
+def _assert_completed_cache_is_as_expected(
+        cache_data_trees,
+        param_data_trees,
+        flatten=False):
+    for outer_key, cache_data_tree in cache_data_trees.items():
+        for inner_key, cache_data in cache_data_tree.items():
+            if flatten:
+                assert_array_equal(
+                    cache_data.flatten(),
+                    param_data_trees[outer_key][inner_key].flatten()
```
astafan8, line 537: if the cache is too small, then indeed the data will be flat (right?), so then my question is why forcefully flatten them again here?

qcodes/tests/dataset/test_data_set_cache.py
```diff
@@ -331,6 +333,150 @@ def test_cache_complex_array_param_in_1d(experiment, DAC, channel_array_instrume
...
+        'dummy_channel_inst_ChanA_dummy_parameter_with_setpoints': (n_points_outer, n_points_inner, pws_n_points),
+        'dummy_channel_inst_ChanA_dummy_parameter_with_setpoints_complex': (n_points_outer, n_points_inner, pws_n_points)
+    }
+
+    if not cache_too_small:
+        assert meas._shapes == expected_shapes
+
+    with meas.run(write_in_background=bg_writing) as datasaver:
+        dataset = datasaver.dataset
+        _assert_parameter_data_is_identical(dataset.get_parameter_data(), dataset.cache.data())
```
astafan8, line 497: right?
```suggestion
        # Check that parameter data and cache data are indential for empty datasets
        _assert_parameter_data_is_identical(dataset.get_parameter_data(), dataset.cache.data())
```

qcodes/tests/dataset/test_data_set_cache.py
```diff
@@ -331,6 +333,150 @@ def test_cache_complex_array_param_in_1d(experiment, DAC, channel_array_instrume
...
+                meas_vals = [(param, param.get()) for param in meas_parameters]
+
+                datasaver.add_result((DAC.ch1, v1),
+                                     (DAC.ch2, v2),
+                                     *meas_vals)
+                datasaver.flush_data_to_database(block=True)
+                param_data_trees = dataset.get_parameter_data()
+                cache_data_trees = dataset.cache.data()
+
+                _assert_partial_cache_is_as_expected(
```
astafan8, line 513: does this mean that parameter data and cache data are NOT identical during the measurement (otherwise i'd expect _assert_parameter_data_is_identical to be called as well)? why is that, it seems i misunderstood something?

qcodes/tests/dataset/test_data_set_cache.py
```diff
@@ -331,6 +333,150 @@ def test_cache_complex_array_param_in_1d(experiment, DAC, channel_array_instrume
...
+                    cache_data_trees,
+                    expected_shapes,
+                    n_points_measured,
+                    param_data_trees,
+                    cache_too_small
+                )
+    cache_data_trees = dataset.cache.data()
+    param_data_trees = dataset.get_parameter_data()
+
```
astafan8, line 522: are they not identical when the measurement is complete?
```suggestion
    _assert_parameter_data_is_identical(param_data_trees, cache_data_trees)
```

qcodes/tests/dataset/test_data_set_cache.py
```diff
@@ -331,6 +333,150 @@ def test_cache_complex_array_param_in_1d(experiment, DAC, channel_array_instrume
                                         dataset.cache.data())
 
 
+@pytest.mark.parametrize("bg_writing", [True, False])
+@pytest.mark.parametrize("cache_too_small", [True, False])
```
astafan8, line 427: it seems that we need a parameterization here for an interrupted dataset. i think expectations are the same

docs/examples/DataSet/Performing-measurements-using-qcodes-parameters-and-dataset.ipynb
```diff
@@ -1527,6 +1528,79 @@
...
+   "source": [
+    "n_points_1 = 100\n",
+    "n_points_2 = 200\n",
+    "\n",
+    "meas_with_shape = Measurement(exp=exp)\n",
+    "meas_with_shape.register_parameter(dac.ch1)  # register the first independent parameter\n",
+    "meas_with_shape.register_parameter(dac.ch2)  # register the second independent parameter\n",
+    "meas_with_shape.register_parameter(dmm.v2, setpoints=(dac.ch1, dac.ch2))  # now register the dependent oone\n",
+    "\n",
+    "meas_with_shape.set_shapes(detect_shape_of_measurement((dmm.v2,), (n_points_1, n_points_2)))\n",
```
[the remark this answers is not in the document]
jenshnielsen: I would not put that in. I do not want any users to be encouraged to try to do this them self.

qcodes/tests/dataset/test_data_set_cache.py
jenshnielsen: Yes I thought that was obvious but sure

docs/examples/DataSet/Performing-measurements-using-qcodes-parameters-and-dataset.ipynb
```diff
@@ -1527,6 +1528,79 @@
...
+   "metadata": {},
+   "source": [
+    "### Specifying shape of measurement\n",
+    "As the context manager allows you to generate data in any shape with the only restriction being that you supply values for both dependent and independent parameters together it cannot know if the date is to be measured on a grid and as a consequence the Numpy array of data loaded from the dataset may not be in the shape that you expect. `plot_dataset` and `get_data_as_pandas_dataframe`  contains logic that can detect the shape of the data measured at load time. However, if you know the shape of the measurement that you are going to perform up front you may chose to specify it before initializing the measurement. \n",
+    "\n",
+    "`dataset.get_parameter_data` and `dataset.cache.data` automatically makes use of this information to return shaped data when loaded from the database.\n",
+    "\n",
+    "Note that if you use the doNd functions demonstrated in [Using doNd functions for performing measurements](Using_the_doNd_functions_for_running_experiments.ipynb) this will happen automatically.\n",
+    "\n",
+    "In the example below we show how the shape can be specified manually.\n"
```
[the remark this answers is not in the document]
jenshnielsen: I think that goes without saying

qcodes/dataset/sqlite/queries.py
```diff
@@ -177,11 +177,55 @@ def get_parameter_data(conn: ConnectionPlus,
...
+
+    one_param_output, _ = get_parameter_data_for_one_paramtree(
+        conn,
+        table_name,
+        rundescriber,
+        output_param,
+        start,
+        end
+    )
+    if rundescriber.shapes is not None:
```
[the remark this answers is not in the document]
jenshnielsen: That is correct

qcodes/dataset/sqlite/queries.py
```diff
@@ -225,6 +269,13 @@ def get_parameter_data_for_one_paramtree(
 
     for paramspec, column_data in zip(paramspecs, res_t):
         try:
+            if paramspec.type == "numeric":
+                # there is no reliable way to
+                # tell the difference between a float and and int loaded
+                # from sqlite numeric columns so always fall back to float
+                dtype: Optional[type] = np.float64
+            else:
```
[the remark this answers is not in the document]
jenshnielsen: Nope this is only requied because there is no way to tell the difference between float and ints storred as the numeric type in sqlite. Complex is storred as its own type. 

For numeric you could have (1,1.5.,2) and so on as the data read back. You will therefor get an array of integers if you only look at the first data point. Which can happen as you read the data back.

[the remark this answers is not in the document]
jenshnielsen: No see above

qcodes/tests/dataset/test_data_set_cache.py
jenshnielsen: The cache is only too small once you have measured more than the size of the cache. One could extend the test to separate those two tests

qcodes/tests/dataset/test_data_set_cache.py
jenshnielsen: Nope this is not always true, Only after a number of measurements have been performed

qcodes/tests/dataset/test_data_set_cache.py
jenshnielsen: Yes this is the part of the shape that comes from array/multiparameter/parameterwithsetpoints

qcodes/tests/dataset/test_data_set_cache.py
jenshnielsen: Yes that is obviously true. The way that the code is currently written 

get_parameter_data will attempt to reshape the data is and only if the full data set has the correct shape. It will never contain empty unfilled data points. 

The cache on the other hand will return an preshaped array with empty values

qcodes/dataset/sqlite/queries.py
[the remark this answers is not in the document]
astafan8: aah, ok, tjhanks

qcodes/tests/dataset/test_data_set_cache.py
astafan8: ehm... i mean "once you have measured more than the size of the cache", calling the cache will already give a flat array, right? hence cache_data that is passed to this assertion will already be flat, right? or not? (if yes, then why flatten it again?)

qcodes/tests/dataset/test_data_set_cache.py
astafan8: ah, yes, indeed, sorry

qcodes/tests/dataset/test_data_set_cache.py
astafan8: > The cache is only too small once you have measured more than the size of the cache

aaaah, now got it, indeed, thanks!

qcodes/tests/dataset/test_data_set_cache.py
astafan8: > The way that the code is currently written get_parameter_data will attempt to reshape the data is and only if the full data set has the correct shape

aha. i didn't realize this fully. what is the reason to have this difference? to make people completely move to using dataset.cache.data and never come to get_parameter_data unless they need controlled fine-grained data access? :)

jenshnielsen: It seems to me that changing get_parameter_data in the way that cache is implemented would be an API break of sorts which is why I left it as is but perhaps we should not worry about that

qcodes/dataset/data_set_cache.py
```diff
@@ -27,7 +25,17 @@ def __init__(self, dataset: 'DataSet'):
...
         #: number of rows read per parameter tree (by the name of the dependent parameter)
         self._read_status: Dict[str, int] = {}
+        self._write_status: Dict[str, Optional[int]] = {}
         self._loaded_from_completed_ds = False
+        self._rundescriber: Optional[RunDescriber] = None
+
+    @property
+    def rundescriber(self) -> RunDescriber:
+        if self._rundescriber is None:
+            self._rundescriber = get_rundescriber_from_result_table_name(
```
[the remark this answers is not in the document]
jenshnielsen: Right this is an artifact of the way the original code was written (while the dataset also changed at the same time)

docs/examples/DataSet/Performing-measurements-using-qcodes-parameters-and-dataset.ipynb
jenshnielsen: Yes that should indeed be fixed but in a different pr. We should be able to clean up that code significnantly (and perhaps even share the guess code between plottr and qcodes)

qcodes/dataset/sqlite/queries.py
```diff
@@ -177,11 +177,55 @@ def get_parameter_data(conn: ConnectionPlus,
...
+        shape = rundescriber.shapes.get(output_param)
+
+        if shape is not None:
+            total_len_shape = np.prod(shape)
+            for name, paramdata in one_param_output.items():
+                total_data_shape = np.prod(paramdata.shape)
+                if total_data_shape == total_len_shape:
+                    one_param_output[name] = paramdata.reshape(shape)
+                else:
+                    log.warning(f"Tried to set data shape for {name} in "
```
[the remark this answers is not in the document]
jenshnielsen: Perhaps we should only warn when dataset > cache

qcodes/tests/dataset/test_data_set_cache.py
astafan8: Agree, let's keep it as it is now

qcodes/dataset/sqlite/queries.py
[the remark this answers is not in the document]
astafan8: > we should only warn when dataset > cache

yes, i think we should "warn IF shape info available AND dataset > cache"

jenshnielsen:
@astafan8 I think this addressed all the questions raised. Do you want to have another look?

qcodes/dataset/data_set.py
```diff
@@ -944,10 +944,20 @@ def get_parameter_data(
...
         parameters as keys and values consisting of dictionaries with the
         names of the parameters and its dependencies as keys and numpy arrays
-        of the data as values. If some of the parameters are stored as arrays
+        of the data as values. If the dataset has a shape recorded
+        in its metadata and the number of datapoints recorded matches the
+        expected number of points the data will be returned as numpy arrays
+        in this shape. If there are less datapoints recorded than expected
+        from the metadata the dataset will be returned as is. This could happen
+        if you call `get_parameter_data` on an incomplete dataset. See
+        :py:class:`.DataSetCache.data` for an implementation that returns the
```
astafan8, line 953: would `DataSet.cache.data` work as a better reference? i'm just a bit worried that users would get confused in a "oh, i need some other class, oh dear" way. perhaps i'm too careful)

qcodes/dataset/data_set_cache.py
```diff
@@ -27,7 +25,14 @@ def __init__(self, dataset: 'DataSet'):
         self._data: ParameterData = {}
         #: number of rows read per parameter tree (by the name of the dependent parameter)
         self._read_status: Dict[str, int] = {}
+        #: number of rows written per parameter tree (by the name of the dependent parameter)
+        self._write_status: Dict[str, Optional[int]] = {}
         self._loaded_from_completed_ds = False
+        self._rundescriber: Optional[RunDescriber] = None
```
astafan8, line 31: not needed now that rundescriber property relies on `_dataset`
```suggestion
```

qcodes/dataset/sqlite/queries.py
```diff
@@ -1736,3 +1791,138 @@ def remove_trigger(conn: ConnectionPlus, trigger_id: str) -> None:
...
+def _create_new_data_dict(new_values: np.ndarray,
+                          shape: Optional[Tuple[int, ...]]
+                          ) -> Tuple[np.ndarray, int]:
+    if shape is None:
+        return new_values, new_values.size
+    else:
+        n_values = new_values.size
+        data = np.zeros(shape, dtype=new_values.dtype)
+
+        # TODO what are good empty values for non floats
```
astafan8, line 1894:
```suggestion
```

qcodes/dataset/sqlite/queries.py
```diff
@@ -1736,3 +1791,138 @@ def remove_trigger(conn: ConnectionPlus, trigger_id: str) -> None:
...
+        subtree_parameters = set(existing_data.keys()) | set(new_data.keys())
+        new_write_status: Optional[int]
+
+        for subtree_param in subtree_parameters:
+            # todo one should check that all write_status values are
+            # the same within one parameter_tree
+            existing_values = existing_data.get(subtree_param)
+            new_values = new_data.get(subtree_param)
+            if existing_values is not None and new_values is not None:
+                array_allocated = shape is not None
```
astafan8, line 1862: this variable is unused?

astafan8 (approved):

PR #2269, opened 2020-10-12
Widget only list completed

jenshnielsen:
The experiments widget assumes that run_timestamp_raw is set to not None 
This is not the case for pristine runs that have not been started.

It seems like the widget should only show completed runs so this pr filters the non completed runs away. 
Note we could also just filter away the pristine runs and show started runs? 

replaces #2268 since mypy .790 now checks the signature of keys given to sorted

### DIFF
--- a/qcodes/interactive_widget.py
+++ b/qcodes/interactive_widget.py
@@ -9,7 +9,7 @@ import traceback
 from datetime import datetime
 from functools import partial, reduce
 from typing import Any, Callable, Dict, Optional, Sequence, Iterable, \
-    TYPE_CHECKING
+    TYPE_CHECKING, cast
 
 import matplotlib.pyplot as plt
 from IPython.core.display import display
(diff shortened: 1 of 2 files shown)


astafan8 (commented):
> It seems like the widget should only show completed runs

i am of a different opinion: i think it's fair for the widget to "show what there is". what do you think @basnijholt ? the fact that the pristine datasets can't be sorted by timestamp is a fact of life, i guess.

jenshnielsen:
@astafan8 What would you then prefer? Catch the exception and return an unstored list. Sort all prisine runs after (or before) the runs that have a time stamp

astafan8:
> What would you then prefer? Catch the exception and return an unstored list. Sort all prisine runs after (or before) the runs that have a time stamp

hmm, i now tend to agree that showing pristine runs is generally useless - the only reason to do that is so that the user has a "feedback" if he created a new run and something bad happened that kept the run pristine. So, perhaps, indeed, filtering out the pristine ones from the table, and if those exist, then add a note somewhere "The N pristine DataSets are not shown on the table".

btw, what is the current behavior? they are shown at the end of the table? (and mypy is not happy, right?)

jenshnielsen:
I belive the current behaviour is to throw if there are runs without a `raw_run_timestamp`

```
In [2]: sorted([1.,2.,3., None])
<ipython-input-2-8e52d67a9c97> in <module>
----> 1 sorted([1.,2.,3., None])

TypeError: '<' not supported between instances of 'NoneType' and 'float'
```

jenshnielsen:
replaced by #2275

PR #2276, opened 2020-10-13
Legacy: ensure that datasets have unique names to avoid potential race condition

jenshnielsen:
These tests randomly fail with data set already existing when the name is not set. 
Try to fix this by giving all of them a unique name


astafan8 (approved):
interesting, sounds fair :)

jenshnielsen:
( a better solution would be to rewrite these to use a per test temporary directory but that is out of scope for the moment since these are legacy tests and are still using unittest rather than pytest )

PR #2278, opened 2020-10-13
Keysight B1511B Driver

Akshita07:
- Add a new driver and tests for Keysight B1511B
- Improve Keysight B1517A for valid voltage and current output ranges

@astafan8


qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1517A.py
```diff
@@ -21,6 +21,7 @@
 
 if TYPE_CHECKING:
     from .KeysightB1500_base import KeysightB1500
+    from .KeysightB1511B import B1511B
```
astafan8, line 24: there's not need for this anymore because B1511B inherits from B1517A, this places where B1517A is needed, B1511B is allowed automatically.

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_base.py
```diff
@@ -437,7 +440,7 @@ class IVSweepMeasurement(MultiParameter, StatusMixin):
         instrument: Instrument to which this parameter communicates to.
     """
 
-    def __init__(self, name: str, instrument: B1517A, **kwargs):
+    def __init__(self, name: str, instrument: Union[B1511B, B1517A], **kwargs):
```
astafan8, line 443: should not be needed now that B1511B inherits from B1517A (in the whole PR)

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1511B.py
```diff
@@ -0,0 +1,110 @@
...
+                                            IMeasRange.MIN_100pA,
+                                            IMeasRange.FIX_1pA,
+                                            IMeasRange.FIX_10pA,
+                                            IMeasRange.FIX_100pA]
+            self._valid_i_output_ranges = self._valid_i_output_ranges + \
+                                          [IOutputRange.MIN_1pA,
+                                           IOutputRange.MIN_10pA,
+                                           IOutputRange.MIN_100pA]
+
+    def source_config(
```
astafan8, line 70: why can't this method belong to the parent class? and B1511B will then just define the valid_* lists (as it already does it ints init) without the need for modifying the source_config method?

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_base.py
```diff
@@ -129,7 +130,9 @@ def from_model_name(model: str, slot_nr: int, parent: 'KeysightB1500',
         Returns:
             A specific instance of :class:`.B1500Module`
         """
-        if model == "B1517A":
+        if model == "B1511B":
+            return B1511B(slot_nr=slot_nr, parent=parent, name=name)
```
astafan8, line 134: in order to test that this works, we can change the pyvisa-sim yaml file to specify that one of the slots is B1511B as opposed to B1517A

astafan8 (commented):
very close to be done :)

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1511B.py
```diff
@@ -0,0 +1,63 @@
...
+        parent: mainframe B1500 instance that this module belongs to
+        name: Name of the instrument instance to create. If `None`
+            (Default), then the name is autogenerated from the instrument
+            class.
+        slot_nr: Slot number of this module (not channel number)
+        asu_present: Flag to acknowledge ASU presence
+    """
+
+    def __init__(self, parent: 'KeysightB1500', name: Optional[str],
+                 slot_nr: int, asu_present: bool = False, **kwargs):
```
astafan8, line 25: assuming ASU is present, how would users make sure asu_present argument here is set to True?

in the current implementation, user does not instantiate modules explicitly, instead B1500 class does that. So in the current implementation there's no way the user can instantiate B1500 with B1511 and ASU present.

I see at least two options: 1) make asu_present an argument to B1500 (and pass it all the way to B1511 init) so that the user can tell B1500 at init time that ASU is present, 2) make asu_present an attribute (or property) of B1511 so that the user can tell B1500/B1511 AFTER init time if ASU is present.

Option 2 seems to me the least amount of effort which is great since the users that we know of don't use ASU.

astafan8 (approved):

PR #2291, opened 2020-10-15
AMI430 remove has_current_rating that does not work with new firmware versions

jenshnielsen:


astafan8 (approved):
and the "new firmware versions" are 2.62 and the 3.xx?

jenshnielsen:
It is newer than some version older than 2.62 but I am not sure which version. @FarBo will verify

FarBo:
@jenshnielsen 

I set has_current_rating: true in the station file of a magnet running 2.62. This set caused snapshot warning for current_rating and field_rating in the load time of the instrument. To make sure that these two parameters are not present for this firmware, I called them, which returned me error. All means that the changes in this PR is correct.

jenshnielsen:
@FarBo thanks for confirming. @astafan8 Do you agree this is the correct solution then ?

astafan8 (approved):
I agree! thanks @FarBo and @jenshnielsen !

PR #2297, opened 2020-10-16
remove pylint and its deps from test_requirements

jenshnielsen:
We are currently not using pylint in ci so no need to install it. 
Pylint is used via codecov


astafan8 (approved):
fair.

PR #2301, opened 2020-10-18
Alazar drivers cleanup

jenshnielsen:
* Remove redundant get_cmd from all parameters. (This is the default anyway)
* Remove redundant set_cmd from TraceParameters. TraceParamter implements set_raw so there is no need for this.
* Rewrite loops over channels and trigger engines to be more explicit.
* Make dll_path an argument
* Remove deprecated config method from alazar class

This should be tested on an actual device


qcodes/instrument_drivers/AlazarTech/ATS9360.py
```diff
@@ -77,55 +72,48 @@ def __init__(self, name, **kwargs):
...
                            parameter_class=TraceParameter,
                            label='Decimation',
                            unit=None,
                            initial_value=1,
                            vals=validators.Ints(0, 100000))
 
-        for i in ['1', '2']:
-            self.add_parameter(name='coupling' + i,
-                               get_cmd=None,
+        for i in range(1, self.channels+1):
```
FarBo, line 88: I think channels is missing to be defined.

jenshnielsen: No the default value of 2 is set in the super class `AlazarTech_ATS`

qcodes/instrument_drivers/AlazarTech/ATS9870.py
```diff
@@ -96,16 +100,16 @@ def __init__(self, name, **kwargs):
...
                                             2.0: 11,
                                             4.0: 12})
-            self.add_parameter(name='impedance' + i,
+            self.add_parameter(name=f'impedance{i}',
                                parameter_class=TraceParameter,
-                               label='Impedance channel ' + i,
+                               label=f'Impedance channel {i}',
                                unit='Ohm',
                                initial_value=50,
                                val_mapping={1000000: 1,
```
FarBo, line 108: This is not related to this PR, but we don't have 1000000 as impedance for this card. the card has a fixed 50 ohm impedance. I missed this in the original driver review, but it is not harmful and is nice to be corrected now.

FarBo, line 108: This is not related to this PR, but we don't have impedance of 1000000 ohm for this card. It only supports a fixed 50 ohm. Since this parameter is not settable, it is fine but maybe it is nice to just delete this line.

FarBo (approved):

jenshnielsen:
@astafan8 could you have a look too?

qcodes/instrument_drivers/AlazarTech/ATS9870.py
```diff
@@ -77,16 +81,16 @@ def __init__(self, name, **kwargs):
                            initial_value=0,
                            vals=validators.Ints(0, 100000))
 
-        for i in ['1', '2']:
-            self.add_parameter(name='coupling' + i,
+        for i in range(1, self.channels+1):
```
astafan8, line 84:
```suggestion
        for i in range(1, self.channels + 1):
```

astafan8 (approved):
great, much cleaner!

PR #2304, opened 2020-10-19
Updates to allow pyvisa 1.11 dependency

trevormorgan:
Several updates are needed to use pyvisa 1.11 in qcodes. 

- ` import visa` is deprecated  in favor of `import pyvisa as visa`
- pyvisa backend using of `-` is deprecated 
  - qcodes forked pyvisa-sim repo has merge updates from upstream to change backend naming schema to `pyvisa_sim`. This PR needs to land before this one for testing to pass. https://github.com/QCoDeS/pyvisa-sim/pull/5
- `write_raw` low-level interface to `visa_handle.write` update needed to to changes in pyvisa event handler 
  - This section could use some more eyes as I think there should be a smarter way to handle write error events. 

Closes #2184


qcodes/instrument/visa.py
```diff
@@ -209,8 +209,11 @@ def write_raw(self, cmd: str) -> None:
         """
         with DelayedKeyboardInterrupt():
             self.visa_log.debug(f"Writing: {cmd}")
-            nr_bytes_written, ret_code = self.visa_handle.write(cmd)
-            self.check_error(ret_code)
+            response = self.visa_handle.write(cmd)
+
+            if isinstance(response, tuple) and len(response) == 2:
+                ret_code = response[1]
+                self.check_error(ret_code)
```
jenshnielsen, line 216: I think we should just drop this code block. Evidently this was never meant to return the status code

qcodes/tests/test_wrong_address.py
```diff
@@ -12,15 +12,15 @@
 
 
 def test_wrong_address():
-    wrong_address = 'GPIB::2::INSTR'
+    wrong_address = 'GPIB0::2::0::INSTR'
```
jenshnielsen, line 15: Is the secondary address (the last 0) really needed ? I suspect if that is also true for real instruments that could break a lot of experiments.

qcodes/instrument/visa.py
astafan8: return - no, but check the error and raise a VisaIOError on it - yes, right? or am i misunderstanding something?

astafan8 (commented):
@trevormorgan so this PR works with both pyvisa 1.11 and 1.10.x?

qcodes/instrument/visa.py
jenshnielsen: @astafan8 I mean pyvisa was never meant to return the error code to here

astafan8: yes, thanks (also clarified this in an offline conversation)

trevormorgan: okay great, I was wondering this as well.  I wasn't exactly sure if the `return_handler` took care of this so it's great for the confirmation.

astafan8 (approved):

qcodes/tests/test_wrong_address.py
MatthieuDartiailh: This is the normalized form of the address and both forms should behave in an identical manner for all implementation of VISA.

qcodes/tests/test_visa.py
```diff
@@ -51,11 +52,9 @@ def write(self, cmd):
             raise ValueError('be more positive!')
 
         if num == 0:
-            ret_code = visa.constants.VI_ERROR_TMO
-        else:
-            ret_code = 0
+            raise visa.VisaIOError(visa.constants.VI_ERROR_TMO)
```
trevormorgan, line 55: The test `test_ask_write_local` was failing after removing the `ret_code` and `error_check` from `visa.py`. I don't think my solution to get the test to pass is actually testing how visa's resource handle is used. I wanted to get your thoughts @jenshnielsen and @astafan8 before I merge this PR.

qcodes/instrument/visa.py
```diff
@@ -181,25 +185,6 @@ def close(self) -> None:
             self.visa_handle.close()
         super().close()
 
-    def check_error(self, ret_code: int) -> None:
```
jenshnielsen, line 184: Maybe we should leave this around deprecated for a few releases @astafan8 what do you think?

qcodes/tests/test_visa.py
jenshnielsen: I think this is fine. It looks like the intention of this mock all the visa handle out and only test qcodes handling of the visa commands. The tests are not of particulary high quality but I don't think they are made any worse by doing this

qcodes/instrument/visa.py
astafan8: yes, we should

astafan8: done

astafan8 (approved):

PR #2309, opened 2020-10-19
Try publishing docs for all versions using sphinx-multiversion

astafan8:
[`sphinx-multiversion`](https://holzhaus.github.io/sphinx-multiversion) package DOES work, and produces the following (which is great!):

MicrosoftTeams-image (1)

Pros:
* the package does not rely on almost any internals of sphinx, it calls it via "python -m sphinx ..." - should not have problems with version compatibilities, etc.
* the package mainly implements injecting the information on the "version" of the docs that are being built, and injecting html templates that use that information at the docs built time (templates you have to provide yourself, but the documentation provides minimal ones) - very flexible and simple

Cons:
* docs for all the versions are build in the same python environment (different dirs, and checking out exact tags etc is done perfectly) which means:		
  * executable docs will likely have FAILURES
  * the assumption is that all "your" older docs CAN build with whatever sphinx version that you have in your python environment

The con above is basically ruining it for us because we have a lot of executable jupyter notebooks.
    
My personal conclusion of this little research is that we could use the ideas/approaches of sphinx-multiversion to set up our own thing that does what sphinx-multiversion does + takes care of creating correct python environments with correct version of qcodes and docs dependencies.


astafan8:
closing as this is not really needed for now and also see the cons in the description

PR #2315, opened 2020-10-20
Remove deprecated keithley 2600 driver

jenshnielsen:
Note that I had to update two notebooks that used it. I do not have the hardware to test these notebooks.

This has been deprecated for 3 years


astafan8 (approved):
i think it's fine to not test on the notebook on the instrument because the the change accounts for the difference in the drivers

PR #2317, opened 2020-10-20
Use the full names when registering multiparameters

jenshnielsen:
For some reason multiparameters where unlike all other paramters not using the full name. This is both confusing and problematic if you have to multiparameters with the same short name (e.g. from different channels) 

@astafan8 Any suggestions for other ways of doing it?

Note This is a breaking change but I don't see any way around it.


jenshnielsen:
The changes to not use `str(parameter)` are actually only for consistency. The real issue is the fix 
from `self._register_parameter(multiparameter.names[i],` to `self._register_parameter(multiparameter.full_names[i],`

astafan8:
> The changes to not use str(parameter) are actually only for consistency

that's fine, it's just that when reviewing i had to look and see what str(parameter) does to confirm the consistency ;)

PR #2324, opened 2020-10-21
Keithley3706A: Convert _get_slot_id to a public method

trevormorgan:
This method would useful upstream as a public method. This PR is simply converting `_get_slot_id` to `get_slot_id`


trevormorgan:
Closing this PR as I can use a public facing method that already exists to get this information.

PR #2326, opened 2020-10-22
Don't exclude instrument drivers from codecov

jenshnielsen:
While we don't have good coverage for these I do not see that as a reason to not measure coverage of them


astafan8 (approved):
totally agree

PR #2327, opened 2020-10-22
Add traceback of exception during measurement to dataset's metadata

astafan8:
The metadata key is called "exception".

questions/todos:
- [x] is the name of the metadata key good? - yes, "measurement_exception"
- [x] is storing a string good enough or shall we store the exception data in a more structured way? and if so, how? - string formatted by python's traceback module should be good enough for a start
- [x] add a test :) - done


jenshnielsen:
Cool, this has some suggestions of how to serialize that may be useful https://stackoverflow.com/questions/45240549/how-to-serialize-an-exception/45241491

jenshnielsen:
I don't have a better idea for the key name

jenshnielsen (approved):

PR #2341, opened 2020-10-24
Add full type checking of Qdac drivers

jenshnielsen:
Should be merged after #2336 and not backported. This probably should be tested manually on both devices


qcodes/instrument_drivers/QDev/QDac_channels.py
```diff
@@ -354,35 +375,33 @@ def _set_vrange(self, chan, switchint):
...
-    def _num_verbose(self, s):
+    def _num_verbose(self, s: str) -> float:
         """
         turn a return value from the QDac into a number.
         If the QDac is in verbose mode, this involves stripping off the
         value descriptor.
         """
         if self.verbose.get_latest():
-            s = s.split[': '][-1]
+            s = s.split(': ')[-1]
```
jenshnielsen, line 395: is this correct. The original code looks like it never worked?

qcodes/instrument_drivers/QDevil/QDevil_QDAC.py
```diff
@@ -700,7 +712,7 @@ def _setslope(self, chan, slope):
             try:
                 fg = self._assigned_fgs[chan]
                 self._assigned_fgs[chan].t_end = 0
-                self._assigned_triggers.pop(fg)
+                self._assigned_triggers.pop(fg.fg)
```
jenshnielsen, line 715: This is a change but as far as I can see _assigned_triggers does not contain Generator objects but integers so this code has probably never worked correctly

qcodes/instrument_drivers/QDevil/QDevil_QDAC.py
```diff
@@ -938,35 +956,43 @@ def ramp_voltages(self, channellist, v_startlist, v_endlist, ramptime):
...
                             slow_steps=1, fast_steps=steps)
 
-    def ramp_voltages_2d(self, slow_chans, slow_vstart, slow_vend,
-                         fast_chans, fast_vstart, fast_vend,
-                         step_length, slow_steps, fast_steps):
+    def ramp_voltages_2d(
+            self,
+            slow_chans: Sequence[int],
+            slow_vstart: Sequence[float],
+            slow_vend: Sequence[float],
```
jenshnielsen, line 963: This was documented as List[int] but surely the voltages must be list of floats? This is also needed for functions calling this to check correctly

qcodes/instrument_drivers/QDev/QDac_channels.py
```diff
@@ -251,11 +265,15 @@ def __init__(self,
...
+    def snapshot_base(
+            self,
+            update: Optional[bool] = False,
+            params_to_skip_update: Optional[Sequence[str]] = None
+    ) -> Dict:
+        update_currents = self._update_currents and update is True
         if update:
             self._update_cache(readcurrents=update_currents)
-        self._get_status_performed = True
+            self._get_status_performed = True
```
astafan8, line 276: this is now correct with the indentation, right?

jenshnielsen: Yes, the intention is that the status call is only done if update is true. This will enable you correctly update a single parameter if its marked as invalid with a regular snapshot call

qcodes/instrument_drivers/QDevil/QDevil_QDAC.py
```diff
@@ -49,9 +52,9 @@ class Waveform:
     all_waveforms = [sine, square, triangle, staircase]
 
 
-class Generator():
+class Generator:
     #  Class used in the internal book keeping of generators
-    def __init__(self, generator_number):
+    def __init__(self, generator_number: int):
         self.fg = generator_number
         self. t_end = 9.9e9
```
astafan8, line 59: this line seems to have this extra space - did't know that this happens to be valid python :)
```suggestion
        self.t_end = 9.9e9
```

astafan8 (approved):
I approve as long as the change has been tested on the insturment (including the verbose mode :) )

qcodes/instrument_drivers/QDevil/QDevil_QDAC.py
jpsecher: AFAICS, `fg` is an integer (representing a function generator), so `fg.fg` does not make sense, and the original `pop(fg)` seems right.

qcodes/instrument_drivers/QDev/QDac_channels.py
jpsecher: Yeah, `split[...]` do definitely not work.

qcodes/instrument_drivers/QDevil/QDevil_QDAC.py
jpsecher: Yes, the voltages are floats.

qcodes/instrument_drivers/QDevil/QDevil_QDAC.py
jenshnielsen: If you look at line 882 you can see that the variables in self._assigned_fgs are instances of the Generator Class and not integers. 
As far as I can see you therefore need to do `fg.fg` to get the generator number which is what is used within `self._assigned_triggers` 

See 
 https://github.com/QCoDeS/Qcodes/blob/0813f163cb3dbf5ffbbff9bc64edf235c13a1f5d/qcodes/instrument_drivers/QDevil/QDevil_QDAC.py#L882

jpsecher: Yes, you're right.  I guess I should work on making the code easier to understand :-/

jenshnielsen: Yes that would be great. These dictionaries are a bit confusing in their types

FarBo:
I will test it with the instrument and approve it

FarBo:
@jenshnielsen 

I tested this PR for Qdevil. It seems that it is working without any issue (tested instrument version is 1.07). For testing _assigned_triggers, the only requirement is to ramp_voltage for more than 1 channel, otherwise triggers is 1 and _assigned_triggers will not be populated. Then, setting slope to 'Inf' for each channel results in popping that channel out of the _assigned_triggers.

@astafan8 
What should be done for verbose mode? The comments in the driver is not that informative to what to do. Especially in the QDac docstring, it says "The driver assumes that the instrument is ALWAYS in verbose mode OFF
    and sets this as part of the initialization, so please do not change this"

jenshnielsen:
@FarBo thanks for testing

jenshnielsen:
Perhaps @astafan8 meant `update_currents=True` ?

FarBo:
update_currents=True queries the instrument for all channels and return responses, which can be seen as DEBUG messages in the load time.

FarBo (approved):
Both Qdac and Qdevil tested with the instruments. I approve this PR, unless @astafan8 explicitly explains what he wants to be tested in verbose mode.

astafan8:
> unless @astafan8 explicitly explains what he wants to be tested in verbose mode.

The verbose mode allows one to test this change (the square brackets got turned into curly brackets):

https://github.com/QCoDeS/Qcodes/blob/36aeac67a27dd9630577705b531988e080c0cad7/qcodes/instrument_drivers/QDevil/QDevil_QDAC.py#L573-L575

> Especially in the QDac docstring, it says "The driver assumes that the instrument is ALWAYS in verbose mode OFF
and sets this as part of the initialization, so please do not change this"

well, what can i say.... then no need to test the functionality that is not even supposed to be used by anyone apparently... :)

astafan8 (approved):

PR #2343, opened 2020-10-26
change log for patch 0.19.1 release

trevormorgan:
Change log for patch release needed to fix qdac driver with pyvisa 1.11


jenshnielsen:
@trevormorgan it looks like this removed the changelog that I started for the next release

jenshnielsen (approved):

PR #2361, opened 2020-10-27
Mypy Disallow implicit optional

jenshnielsen:
See https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-optional for details

IMHO it is much cleaner to explicitly annotate Optional


astafan8 (approved):
> IMHO it is much cleaner to explicitly annotate Optional

Tottally support that! i like the explicitness :)

PR #2363, opened 2020-10-27
Update environment.yml versions to reflect those in requirements files

astafan8:
We should probably just keep this up to date with the latest versions. (dependabot does not support conda envs i think)


jenshnielsen:
If done that way pyvisa will be install from pip instead of conda. We should probably just keep this up to date with the latest versions. (dependabot does not support conda envs i think)

jenshnielsen (approved):

astafan8:
> ResolvePackageNotFound: 
>  - conda-forge::pyvisa=1.11.1

and of course conda forge is behind :)

jenshnielsen:
Seems like there are some issues with the packagaing

jenshnielsen:
@astafan8 we should probably move it to a pip install then for now

astafan8:
> we should probably move it to a pip install then for now

i think this is fair to do anyway. i din't think that pyvisa is such a popular package that the community makes it timely available on conda.

> I exepct that this will only require us to modify the use of enums for constans in the clear function.

yeah, that's probably a good thing :) and then release qcodes 0.19.2 ? :)

PR #2365, opened 2020-10-28
Add full types to Stanford drivers

jenshnielsen:
And

* fix formatting of some docstrings. 
* Correct use of instrument in `SR86xBufferReadout` parameter


qcodes/instrument_drivers/stanford_research/SR560.py
```diff
@@ -16,36 +16,41 @@ class VoltageParameter(MultiParameter):
     ``VoltageParameter.get()`` returns ``(voltage_raw, voltage)``
 
     Args:
-        measured_param (Parameter): a gettable parameter returning the
```
trevormorgan, line 19: I like how the doc strings were typed :)

jenshnielsen: We usually remove the types from docstrings once they are added to the signature since they are often wrong and there is no way to check them for correctness in the signatures

qcodes/instrument_drivers/stanford_research/SR560.py
```diff
@@ -16,36 +16,41 @@ class VoltageParameter(MultiParameter):
...
-        v_amp_ins (SR560): an SR560 instance where you manually
+        v_amp_ins: an SR560 instance where you manually
             maintain the present settings of the real SR560 amp.
 
             Note: it should be possible to use other voltage preamps, if they
             define parameters ``gain`` (V_out / V_in) and ``invert``
             (bool, output is inverted)
 
-        name (str): the name of the current output. Default 'curr'.
+        name: the name of the current output. Default 'curr'.
```
astafan8, line 28: the default is actually 'volt' :)
```suggestion
        name: the name of the current output. Default 'volt'.
```

astafan8 (approved):

qcodes/instrument_drivers/stanford_research/SR560.py
jenshnielsen: Nice find

PR #2366, opened 2020-10-28
Update contact to new qcodes alias 

trevormorgan:
Update to new qcodes support distribution list email.

@jenshnielsen these were the only references I could find. Do you know if there are any others floating around in the docs?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,8 +46,8 @@ setup(name='qcodes',
       cmdclass=versioneer.get_cmdclass(),
       use_2to3=False,
 
-      maintainer='Jens H Nielsen',
-      maintainer_email='Jens.Nielsen@microsoft.com',
+      maintainer='QCoDeS Core Developers',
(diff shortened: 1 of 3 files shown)


jenshnielsen:
The other one I was thinking of is `maintainer_email` in `setup.py`

setup.py
```diff
@@ -46,8 +46,8 @@ def readme():
       cmdclass=versioneer.get_cmdclass(),
       use_2to3=False,
 
-      maintainer='Jens H Nielsen',
-      maintainer_email='Jens.Nielsen@microsoft.com',
+      maintainer='QCoDeS Core Developers',
```
trevormorgan, line 49: @jenshnielsen I'm happy to leave you has the maintainer if you think an actual name here is more appropriate.

jenshnielsen: No I think this is good.

jenshnielsen (approved):

PR #2367, opened 2020-10-28
Add full types to keysight b1500 driver

jenshnielsen:
@astafan8 Note that there are still a few untyped functions in message builder. These are unused and their tests are stubs so it is hard to guess what the intended types are. Perhaps you know better

### DIFF
--- a/qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1517A.py
+++ b/qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1517A.py
@@ -271,7 +271,7 @@ class IVSweeper(InstrumentChannel):
         0.001 to 100 for UHVU
                            """))
 
-    def _set_sweep_mode(self, value) -> None:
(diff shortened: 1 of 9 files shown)


qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1517A.py
```diff
@@ -396,14 +397,56 @@ def _get_post_sweep_voltage_condition(self) -> int:
...
+            self,
+            name: Literal['sweep_steps']
+    ) -> int: ...
+
+    @overload
+    def _get_sweep_steps_parameters(
+            self,
+            name: Literal['current_compliance',
+                          'power_compliance']
+    ) -> Optional[float]: ...
```
astafan8, line 436: this is so cool!

astafan8 (approved):
> Note that there are still a few untyped functions in message builder. These are unused and their tests are stubs so it is hard to guess what the intended types are. Perhaps you know better

let's defer the typing of those to the time when those methods are actually implemented and used :)

PR #2371, opened 2020-10-28
Pyvisa back support

trevormorgan:
PR for back support of pyvisa 1.10.1

-revert config files to min version of 1.10.1 for pyvisa
- update `device_clear` method to use flush operation based on pyvisa version >1.11 or <=1.11


environment.yml
```diff
@@ -7,6 +7,7 @@ dependencies:
  - numpy=1.17.5
  - matplotlib=3.3.2
  - conda-forge::pyqtgraph=0.11.0
+ - conda-forge::pyvisa=1.10.1
```
trevormorgan, line 10: Did we want to revert to conda or should we keep it as pip?

qcodes/instrument/visa.py
```diff
@@ -139,12 +141,22 @@ def device_clear(self) -> None:
...
-            self.visa_handle.clear()
+
+            # This can be dropped once we drop support for qcodes <0.19
+            # and therefore pyvisa 1.10
+            flush_operation = cast(
+                Any,
+                vi_const.VI_READ_BUF | vi_const.VI_WRITE_BUF_DISCARD
+            )
+        self.visa_handle.flush(flush_operation)
+        self.visa_handle.clear()
```
trevormorgan, line 159: Is it okay to flush and clear or do we still need the logic to only flush if `isinstance(self.visa_handle, pyvisa.resources.SerialInstrument)` ?

jenshnielsen: I think the flush operation only appies to serial instruments

qcodes/instrument/visa.py
```diff
@@ -139,12 +141,22 @@ def device_clear(self) -> None:
         if self.visabackend == 'sim':
             return
 
-        if isinstance(self.visa_handle, pyvisa.resources.SerialInstrument):
-            self.visa_handle.flush(
-                vi_const.BufferOperation.discard_read_buffer_no_io | vi_const.BufferOperation.discard_write_buffer
+        assert isinstance(self.visa_handle, pyvisa.resources.SerialInstrument)
```
jenshnielsen, line 144: This assert will not in general be true. This code can be called for all instrument types.

requirements.txt
```diff
@@ -61,7 +61,7 @@ pyparsing~=2.4.7
 pyrsistent~=0.17.3
 python-dateutil~=2.8.1
 pytz~=2020.1
-PyVISA~=1.11.1
+PyVISA~=1.10.1
```
jenshnielsen, line 64: This one should be left as is. We still want to recommend and test with 1.11.1

trevormorgan: okay thanks I wasn't too sure about changing this. I'll revert.

environment.yml
astafan8: pyvisa 1.11.x is not avialble yet at conda-forge, so in #2363 we decided to just use pip for pyvisa.

qcodes/instrument/visa.py
```diff
@@ -139,12 +141,22 @@ def device_clear(self) -> None:
...
+        if pyvisa_is_1_11_or_higher:
+            flush_operation = (
+                    vi_const.BufferOperation.discard_read_buffer
+                    | vi_const.BufferOperation.discard_write_buffer
             )
         else:
-            self.visa_handle.clear()
+
+            # This can be dropped once we drop support for qcodes <0.19
+            # and therefore pyvisa 1.10
```
astafan8, line 153:
```suggestion
            # This can be dropped once we drop support for pyvisa 1.10
```

environment.yml
jenshnielsen: but if we support 1.10.1 again we could revert to using conda forge ?

qcodes/instrument/visa.py
astafan8: indeed, flush needs to be kept inside `if isinstance(self.visa_handle, pyvisa.resources.SerialInstrument):` block

environment.yml
astafan8: we could, but then folks using the environment.yml won't be using the latest greatest pyvisa :)

qcodes/instrument/visa.py
trevormorgan: thanks for confirming

environment.yml
trevormorgan: okay let's keep as pip for 1.11 then.

jenshnielsen (approved):

PR #2375, opened 2020-10-29
Type oxford Mercury VISA driver and Triton driver. Deprecate old Mercury driver

jenshnielsen:
@astafan8 and @WilliamHPNielsen Do you see any good reason not to deprecate the old Mercury driver


astafan8 (approved):
> Do you see any good reason not to deprecate the old Mercury driver

no. my hope would be that folks migrated to the visa one already. and the IP one is saying "beta version" in the docstring and looks less maintained than the visa one.

PR #2376, opened 2020-10-29
Fix error when calling mark_complted twice

jenshnielsen:
The code would try to remove a non existing element from a set. 
Fix this by both checking for the element first and never do the complete action for an already complted ds

Basic test included

@liangosc


astafan8 (approved):
wow, nice catch!

jenshnielsen:
@liangosc found the bug :)

PR #2377, opened 2020-10-29
Update old driver notebooks (Agilent 34400A)

Akshita07:
Fixes legacy code use in old example driver notebooks.

- Remove Loop usage in Agilent 34400A notebook
- Use get_idn method for getting information in Oxford Triton notebook

To do:
- [x] Test Agilent 34400A notebook with instrument


jenshnielsen:
For the Agilent 34400A notebook I think we should:

* Get rid of that Timer class. It does not really have anything to do with this instrument and just creates more confusion.
* Only create one station. 
* Write a bit of text explaining what we are doing.

docs/examples/driver_examples/Qcodes example with Agilent 34400A.ipynb
```diff
@@ -7,312 +7,121 @@
     "# QCoDeS Example with Agilent 34400A"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This notebook shows how to work with Agilent 34400A instrument qcodes driver and conduct measurements. For testing purposes of this notebook two HP 34401A instruments are connected to a Keithley 2600 equipment as shown in the image below."
```
astafan8, line 14: it's a really great idea to have a photos of the setup, HOWEVER i'd suggest to be more descriptive and clear:
- the reader may not know that HP and Agilent are the same thing
- the reader may not know that 34401A CAN work with 34400A driver
- the cabling on the photo is not explicit enough, as in, the cables are black, and curved and mixed up with each other, and hence less easy to comprehend. I suggest to either make a new photo, from the front panel of all the instruments, make sure that the cables are visible, use decent resolution so that front panels of the instruments are clear (for seeing model names and names of ports); OR draw a single diagram of the connections
- i think readers will appreciate a short summary of the demo measurement NEXT to the photo, just to describe the intent, and then get into details of how it's done and how that demonstrates the usage of Agilent 34400A

jenshnielsen: > the cabling on the photo is not explicit enough, as in, the cables are black, and curved and mixed up with each other, and hence less easy to comprehend. I suggest to either make a new photo, from the front panel of all the instruments, make sure that the cables are visible, use decent resolution so that front panels of the instruments are clear (for seeing model names and names of ports); OR draw a single diagram of the connections

Further more the breakout box on the bottom of the Keithley is custom made so we should make it clear which back panel connection this is

Akshita07: I have replaced the image with a diagram and will add a back panel pic of keithley showing connection points soon.

Akshita07: updated the notebook with an img for custom made connector and how it's connected at the back panel of keithley 2600 equipment.

astafan8 (approved):
Great work on the diagram and the explanations!

PR #2396, opened 2020-11-02
Add types to Alazar drivers

jenshnielsen:
This should probably be tests on an actual device


qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -580,7 +597,7 @@ def clear_buffers(self) -> None:
         self.log.debug("buffers cleared")
         self.buffer_list = []
 
-    def signal_to_volt(self, channel: int, signal: int) -> float:
+    def signal_to_volt(self, channel: int, signal: float) -> float:
```
astafan8, line 600: just double checking if signal is really a float - "convert a value from a buffer" might mean that the value is actually integer, because alazar "measures" integers, and then we convert them to float volts, but i don't remember for sure.

it does not mean though that this method can't be called on integer-values floats or any floats, but i think it was meant to be called on integers that alazar puts into the buffer.

qcodes/instrument_drivers/AlazarTech/ATS_acquisition_controllers.py
```diff
@@ -123,12 +133,16 @@ def post_acquire(self):
             #return [alazar.signal_to_volt(1, res1[0] + 127.5),
             #        alazar.signal_to_volt(2, res2[0] + 127.5),
             #        res1[1], res2[1],
-            #        (res1[1] - res2[1]) % 360]
+            #        (r es1[1] - res2[1]) % 360]
+            # todo can we fix this type error by making contoller generic
+            # in the output type and subclassing.
+            # need to check if that would result in an error for qdev wrappers
+            # that are not typed and do not specialize
```
astafan8, line 140: still a todo?

qcodes/instrument_drivers/AlazarTech/utils.py
```diff
@@ -20,18 +25,22 @@ class TraceParameter(Parameter):
...
     @property
     def synced_to_card(self) -> bool:
         """True if the parameter value has been synced to the instrument"""
         return self._synced_to_card
 
-    def set_raw(self, value):
-        self._instrument._parameters_synced = False
+    def set_raw(self, value: ParamRawDataType) -> None:
+        if TYPE_CHECKING:
+            instrument = cast("AlazarTech_ATS", self._instrument)
```
astafan8, line 42: if a string is used in the cast, is it really needed to be inside the `if TYPE_CHECKING` guard?

astafan8 (approved):

qcodes/instrument_drivers/AlazarTech/ATS.py
jenshnielsen: I changed this to float since we are actually passing a float to this method from `Demodulation_AcquisitionController` Note that the center point will always be a float (127.5) which is the reason for this issue

qcodes/instrument_drivers/AlazarTech/ATS_acquisition_controllers.py
jenshnielsen: No forgot to remove

qcodes/instrument_drivers/AlazarTech/ATS.py
astafan8: oh... ok :)

FarBo (approved):
I tested this PR with ATS9360 using both qdev-wrapper notebook and qcodes instrument_driver notebook. I didn't get any issue running those notebooks using the files in this PR. I also did some small checks outside of those notebook for querying some attributes and methods.
The only thing that I couldn't test was testing in non-windows platforms, which I think it's tested before by our friend shawn, if I'm not wrong :)

PR #2397, opened 2020-11-02
Reverse logic for which modules are typed

jenshnielsen:
This way a module must explicitly be marked as not typed. This should help ensure that all new code is marked typed.

And add the remaing types to `__init__.py`


astafan8 (approved):
Brilliant!

PR #2404, opened 2020-11-05
Add full types to utils.helpers

jenshnielsen:


qcodes/utils/helpers.py
```diff
@@ -397,8 +396,8 @@ def __getattr__(self, key):
             "'{}' object and its delegates have no attribute '{}'".format(
                 self.__class__.__name__, key))
 
-    def __dir__(self):
-        names = super().__dir__()
+    def __dir__(self) -> List[str]:
```
astafan8, line 399: `sorted` returns an iterable, right?
```suggestion
    def __dir__(self) -> Iterable[str]:
```

astafan8 (approved):

qcodes/utils/helpers.py
jenshnielsen: No it always returns a list https://github.com/python/typeshed/blob/master/stdlib/3/builtins.pyi#L1205

astafan8: woooooow! i sohuld've known :)

jenshnielsen: I think this is because you cant sort anything without looking at all the elements so its hard to make it a real Iterable

astafan8: indeed, this makes total sense.

PR #2417, opened 2020-11-09
Add types to delaykeyboardinterrupt module

jenshnielsen:

### DIFF
--- a/qcodes/utils/delaykeyboardinterrupt.py
+++ b/qcodes/utils/delaykeyboardinterrupt.py
@@ -1,5 +1,5 @@
 from typing import Optional, Type
-from types import TracebackType
+from types import TracebackType, FrameType
 import signal
 import logging
 import threading
@@ -28,7 +28,7 @@ class DelayedKeyboardInterrupt:
         elif is_default_sig_handler:
             log.debug("Not on main thread cannot intercept interrupts")
 
-    def handler(self, sig, frame):
+    def handler(self, sig: signal.Signals, frame: FrameType) -> None:
         self.signal_received = (sig, frame)
         print("Received SIGINT, Will interrupt at first suitable time. "
               "Send second SIGINT to interrupt immediately.")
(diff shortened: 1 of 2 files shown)


qcodes/utils/delaykeyboardinterrupt.py
```diff
@@ -47,8 +47,10 @@ def forceful_handler(sig, frame):
...
     def __exit__(self, exception_type: Optional[Type[BaseException]],
                  value: Optional[BaseException],
-                 traceback: Optional[TracebackType]):
+                 traceback: Optional[TracebackType]) -> None:
         if self.old_handler is not None:
             signal.signal(signal.SIGINT, self.old_handler)
         if self.signal_received is not None:
-            self.old_handler(*self.signal_received)
+            if (self.old_handler is not None and
+                    not isinstance(self.old_handler, int)):
```
astafan8, line 55: so if the old handler is `signal.SIG_DFL` then i understand that it's not possible to call it in this way `self.old_handler(*self.signal_received)` but perhaps we still should ask the `signal`'s machinery to somehow "simply perform the default function for the signal" (https://docs.python.org/3/library/signal.html#signal.SIG_DFL) ?

i have negligible clue of what i am talking about, btw :)

astafan8 (approved):

qcodes/utils/delaykeyboardinterrupt.py
jenshnielsen: I thoght about this. I would say that we should not do that. As the code currently is the user could have set another handler before entering the context so that calling the default handler would be wrong. I don't really understand when and how `signal.signal` would return None or an `int` rather than the old handler?

astafan8: then let's keep it as is :) as i said, i know almost nothing and also don't know "when and how signal.signal would return None or an int rather than the old handler" :)

PR #2418, opened 2020-11-10
Add support h5py 3 to legacy dataset

jenshnielsen:
Strings that used to be encoded as bytestrings are now returned as regular strings so calling decode on them is an error. 

This code is not the prettiest but given the legacy nature of this code I think that is ok. 

Also fixed a small bug where an undefined variable was used in an error message in the same file


qcodes/data/hdf5_format.py
```diff
@@ -307,7 +321,7 @@ def _read_list_group(self, entry_point, list_type):
         elif list_type == 'list':
             item = [d[k] for k in sorted(d.keys())]
         else:
-            raise Exception('type %s not supported' % type(item))
+            raise Exception('type %s not supported' % type(list_type))
```
astafan8, line 324: as far as i read the code `list_type` is actually a string, so perhaps the type call is not needed
```suggestion
            raise Exception('type %s not supported' % list_type)
```

astafan8 (approved):

PR #2429, opened 2020-11-12
upgrade versioneer to 0.19

jenshnielsen:
New version cleans up some code and drops support for python 2.


qcodes/__init__.py
```diff
@@ -17,8 +17,6 @@
 # instrument list and running monitor
 add_to_spyder_UMR_excludelist('qcodes')
 
-from qcodes.version import __version__
```
astafan8, line 20: why removing this ? :( i thought that it's a common thing to have the root of the package have `__version__`?

astafan8 (approved):

qcodes/__init__.py
jenshnielsen: Look at the whole file :) It's there twice

astafan8: phew :) :) :) indeed, twice is too many ;)

PR #2443, opened 2020-11-18
Verify that reqiremenets are compatible with Spyder

jenshnielsen:
And a few cleanups 

* Set the same version of parso compatible with Spyder
* Always use pyqt5.12 (spyder depends on pyqtwebengine which depends on a version of pyqt5 newer that 5.9)


science_requirements.txt
```diff
@@ -0,0 +1,117 @@
...
+QtAwesome==1.0.1
+qtconsole==4.7.7
+QtPy==1.9.0
+regex==2020.11.13
+requests==2.25.0
+rope==0.18.0
+six==1.15.0
+snowballstemmer==2.0.0
+sortedcontainers==2.3.0
+Sphinx==3.3.1
```
astafan8, line 94: spyder depends on shpinx and jupyter? wow))

astafan8 (approved):

science_requirements.txt
jenshnielsen: Yes this is literally the output of `pip freeze`

PR #2451, opened 2020-11-19
updated zenodo for v0.20.1

FarBo:
Updated zenodo links


jenshnielsen (approved):
You mean 0.20.1 right

PR #2453, opened 2020-11-20
Add explicit generics to most modules outside tests and drivers

jenshnielsen:


qcodes/dataset/sqlite/queries.py
```diff
@@ -1594,7 +1594,9 @@ def get_metadata(conn: ConnectionPlus, tag: str, table_name: str) -> str:
                             "result_table_name", table_name)
 
 
-def get_metadata_from_run_id(conn: ConnectionPlus, run_id: int) -> Dict:
+def get_metadata_from_run_id(
+        conn: ConnectionPlus, run_id: int
+) -> Dict[Any, Any]:
```
astafan8, line 1599: i guess it's a bit too tedious to make it Dict[str, Any] because sqlite can potentially return not only strings?

qcodes/dataset/sqlite/queries.py
```diff
@@ -1764,7 +1766,7 @@ def _both_zero(run_id: int,
 
         log.info(f'Succesfully updated run number {run_id}.')
 
-    actions: Dict[Tuple[bool, bool], Callable]
+    actions: Dict[Tuple[bool, bool], Callable[..., Any]]
```
astafan8, line 1769: there are the arguments `run_id, conn, guid_comps`
```suggestion
    actions: Dict[Tuple[bool, bool], Callable[[int, ConnectionPlus, Dict[str, Any]], None]]
```

qcodes/instrument/base.py
```diff
@@ -413,12 +413,12 @@ class Instrument(InstrumentBase, AbstractInstrument):
 
     shared_kwargs = ()
 
-    _all_instruments: Dict[str, weakref.ref] = {}
+    _all_instruments: "Dict[str, weakref.ref[Any]]" = {}
     _type = None
-    _instances: List[weakref.ref] = []
+    _instances: "List[weakref.ref[Any]]" = []
```
astafan8, line 418: why not weakref.ref[Instrument] ?

qcodes/instrument/parameter.py
```diff
@@ -1120,7 +1121,7 @@ def __init__(self, name: str,
                             " set_raw is an error.")
         elif not self.settable and set_cmd is not False:
             if set_cmd is None:
-                self.set_raw: Callable = lambda x: x
+                self.set_raw: Callable[..., Any] = lambda x: x
```
astafan8, line 1124: this lambda does not actually conform to the type... :)
```suggestion
                self.set_raw: Callable[[ParamRawDataType, Any kwargs ...], None] = lambda x: x
```

astafan8 (approved):
all good, but couldn't stop myself leaving a few comments))

jenshnielsen:
@astafan8 Actually constraining the types is a non goal of this pr. I will put your suggestions in another pr.

qcodes/instrument/parameter.py
jenshnielsen: Why not? Your suggestion is not valid syntax. There is not as far as I know any good way to type something like a callable that takes on specific argument but the rest can be dynamic.

astafan8: that's sad. but ok, `Callable[...` it will be then

PR #2455, opened 2020-11-20
azure pipelines use v2 coverage tool

jenshnielsen:
See https://github.com/microsoft/azure-pipelines-tasks/pull/11144 for changes


jenshnielsen:
Apparently that task exists in the repo but is not yet released :(

PR #2475, opened 2020-11-25
Add complete types to Validators

jenshnielsen:
To make this possible make the validator generic in the valid type which requires a few more modifications to typecheck nicely

### DIFF
--- a/qcodes/utils/validators.py
+++ b/qcodes/utils/validators.py
@@ -25,7 +25,9 @@ shape_type = Union[int, TCallable[[], int]]
 shape_tuple_type = Optional[Tuple[shape_type, ...]]
 
 
-def validate_all(*args: Tuple["Validator[Any]", Any], context: str = '') -> None:
+def validate_all(*args: Tuple["Validator[Any]", Any],
+                 context: str = ''
+                 ) -> None:
     """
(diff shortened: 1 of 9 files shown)


qcodes/utils/validators.py
```diff
@@ -193,7 +196,7 @@ def __init__(self, min_length: int = 0,
                             'no smaller than min_length')
         self._valid_values = ('.' * min_length,)
 
-    def validate(self, value: str, context: str = ''):
+    def validate(self, value: str, context: str = '') -> None:
```
astafan8, line 199: why do we type the value in validate methods to the type we expect? isn't the point that the type of value is always Any?

qcodes/utils/validators.py
```diff
@@ -823,15 +825,15 @@ def __repr__(self) -> str:
                                               self.shape_unevaluated)
 
 
-class Lists(Validator):
+class Lists(Validator[TList[Anything]]):
```
astafan8, line 828: ?
```suggestion
class Lists(Validator[TList[Validator[Any]]]):
```

astafan8 (approved):

qcodes/utils/validators.py
astafan8: that makes sense

PR #2478, opened 2020-11-26
Changed the default do_plot behaviour to False in doNd and make it configurable

FarBo:
The default behavior of do_plot is changed to False, which I think it is preferred by most users. Also, it may solve the figure.max_open_warning issue, when users do a lot of plotting simultaneously.


astafan8 (approved):
looks good!

would you mind starting a changelog for the next release and mention this as a breaking change?

jenshnielsen:
I agree. We should point users towards interactive plots such as plottr rather than doing manual plots like this. @astafan8 @WilliamHPNielsen @Akshita07 what do you think

jenshnielsen (approved):

ThorvaldLarsen:
@FarBo @jenshnielsen @astafan8 this might be out of scope for this PR but maybe it would be worthwhile to rethink the implementation of doNd functions slightly (see comments in this PR https://github.com/QCoDeS/Qcodes/pull/2478). Many setups are run with the need for slightly different default values for the doNd functions. Is there a way to allow the user to _choose_ the defaults at runtime rather than having to request a change in qcodes itself? It gets quite cumbersome if one needs to pass in arguments at every doNd function call just because the default is not what you want.

jenshnielsen:
@ThorvaldLarsen 

Updated the pr to still make it false by default but enable the user to change the default value locally via the qcodesrc config file

astafan8 (approved):

FarBo:
This is much more a proper way.

PR #2488, opened 2020-12-01
plotting: don't scale empty units

jenshnielsen:
It seems to me that we should not generate plot labels like. `My label (1*10^6)` if the unit is empty.


qcodes/utils/plotting.py
```diff
@@ -317,6 +317,7 @@ def find_scale_and_prefix(data: np.ndarray, unit: str) -> Tuple[str, int]:
     The units for which unit prefixes are added can be found in
     ``_UNITS_FOR_RESCALING``. For all other units
     the prefix is an exponential scaling factor e.g. ``10^3``.
+    If no unit is given (e.g. an empty string) no scaling is performed.
```
WilliamHPNielsen, line 320: I think this "e.g." should be "i.e.". Is that possible?

astafan8:
> not generate plot labels like. My label (1*10^6) if the unit is empty

i don't know, perhaps we should check with users because if we don't scale, then the values on the axes will become like 2.5e6, right? or will matplotlib nicely put 2.5 on the tick label and 1e6 on a side of the axis?

trevormorgan (approved):

PR #2489, opened 2020-12-01
Add kwarg for experiment id to doNd functions

ThorvaldLarsen:
doNd function will currently _always_ use the latest created experiment ID in the database. This PR adds kwarg to explicitly choose the experiment ID used for a given measurement.

### DIFF
--- a/qcodes/utils/dataset/doNd.py
+++ b/qcodes/utils/dataset/doNd.py
@@ -90,6 +90,7 @@ def _catch_keyboard_interrupts() -> Iterator[Callable[[], bool]]:
 def do0d(
         *param_meas: ParamMeasT,
         write_period: Optional[float] = None,
+        exp: Optional[Experiment] = None,
         do_plot: bool = True
         ) -> AxesTupleListWithDataSet:
     """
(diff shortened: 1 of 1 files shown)


jenshnielsen:
@ThorvaldLarsen I was thinking a bit about how to make the dond more flexible in general. Another solution that I thought of would be to allow it to take an optional measurement. That would also allow you to tweak the write period on the measurement (and with a small tweak if the measurement should be written in the background. What do you think about that ?

ThorvaldLarsen:
It does seem like something in that direction could be a much nicer approach. When making this I was starting to worry about how many inputs the doNd function will have.

One concern I have with passing in a measure object though is that I think one would need to create a new measurement object for every doNd call. Or is there a way to create a measure object just once that can then be used in all subsequent calls?
That is can I do:
`meas = qc.Measurement(my_kwargs)`
`do1d = functools.partial(do1d, meas=meas)`

With this PR I was thinking to do 
`do1d = functools.partial(do1d, exp=exp)` and `do1d = functools.partial(do1d.func, exp=exp)` for updating the experiment id used. Havent tested this yet so not sure if this works to change the experiment (and other default kwargs) for do1d at runtime.

ThorvaldLarsen:
I guess what I am asking for is to be allowed to change the defaults of doNd function on the fly without having to add these arguments for every doNd function call. Maybe this is better achieved by making them into a class instead that can hold attributes with the doNd functions as method taking attributes as default values?

jenshnielsen:
replaced by #2526

PR #2499, opened 2020-12-07
Add support for type checking with upcomming Numpy 1.20 release. 

jenshnielsen:
Checked locally with numpy 1.20 rc1

* Some numpy types are now generic so give them an explicit Any. Use types in strings to support older numpy versions
* np.int, np.float and np.complex which are aliases for the build in python types are deprecated. 
* Use the dtype constructor explicitly when needed
* Various fixes for interaction between the python types and numpy types

EDIT:
while this does run with older versions of numpy it does not type check. Not sure what the best solution is.
We can add ignores to those lines or just wait until numpy 1.20 is officially released. Or we could disable the non explicit generics for now.

### DIFF
--- a/qcodes/dataset/data_export.py
+++ b/qcodes/dataset/data_export.py
@@ -384,8 +384,8 @@ def reshape_2D_data(x: np.ndarray, y: np.ndarray, z: np.ndarray
         z_to_plot = np.full((ny, nx), '', dtype=z.dtype)
     else:
         z_to_plot = np.full((ny, nx), np.nan)
-    x_index = np.zeros_like(x, dtype=np.int)
-    y_index = np.zeros_like(y, dtype=np.int)
+    x_index = np.zeros_like(x, dtype=int)
+    y_index = np.zeros_like(y, dtype=int)
     for i, xval in enumerate(xrow):
         x_index[np.where(x == xval)[0]] = i
     for i, yval in enumerate(yrow):
--- a/qcodes/dataset/data_export.py
+++ b/qcodes/dataset/data_export.py
@@ -122,12 +122,12 @@ def _all_steps_multiples_of_min_step(rows: np.ndarray) -> bool:
         The answer to the question
     """
 
-    steps: List[np.ndarray] = []
+    steps_list: List[np.ndarray] = []
     for row in rows:
         # TODO: What is an appropriate precision?
-        steps += list(np.unique(np.diff(row).round(decimals=15)))
+        steps_list += list(np.unique(np.diff(row).round(decimals=15)))
 
-    steps = np.unique(steps)
+    steps = np.unique(steps_list)
     remainders = np.mod(steps[1:]/steps[0], 1)
 
     # TODO: What are reasonable tolerances for allclose?
--- a/qcodes/dataset/data_export.py
+++ b/qcodes/dataset/data_export.py
@@ -384,8 +384,8 @@ def reshape_2D_data(x: np.ndarray, y: np.ndarray, z: np.ndarray
         z_to_plot = np.full((ny, nx), '', dtype=z.dtype)
     else:
         z_to_plot = np.full((ny, nx), np.nan)
-    x_index = np.zeros_like(x, dtype=int)
-    y_index = np.zeros_like(y, dtype=int)
+    x_index = np.zeros_like(x, dtype=np.dtype(int))
+    y_index = np.zeros_like(y, dtype=np.dtype(int))
     for i, xval in enumerate(xrow):
         x_index[np.where(x == xval)[0]] = i
     for i, yval in enumerate(yrow):
--- a/qcodes/dataset/data_export.py
+++ b/qcodes/dataset/data_export.py
@@ -384,8 +384,8 @@ def reshape_2D_data(x: np.ndarray, y: np.ndarray, z: np.ndarray
         z_to_plot = np.full((ny, nx), '', dtype=z.dtype)
     else:
         z_to_plot = np.full((ny, nx), np.nan)
-    x_index = np.zeros_like(x, dtype=np.dtype(int))
-    y_index = np.zeros_like(y, dtype=np.dtype(int))
+    x_index = np.zeros_like(x, dtype=np.dtype(np.int_))
+    y_index = np.zeros_like(y, dtype=np.dtype(np.int_))
     for i, xval in enumerate(xrow):
         x_index[np.where(x == xval)[0]] = i
     for i, yval in enumerate(yrow):
--- a/qcodes/dataset/sqlite/database.py
+++ b/qcodes/dataset/sqlite/database.py
@@ -14,11 +14,16 @@ import numpy as np
 from numpy import ndarray
 
 from qcodes.dataset.sqlite.connection import ConnectionPlus
-from qcodes.dataset.sqlite.db_upgrades import _latest_available_version, \
-    get_user_version, perform_db_upgrade
+from qcodes.dataset.sqlite.db_upgrades import (
+    _latest_available_version,
+    get_user_version,
+    perform_db_upgrade
+)
 from qcodes.dataset.sqlite.initial_schema import init_db
 import qcodes
-from qcodes.utils.types import numpy_ints, numpy_floats, complex_types, complex_type_union
+from qcodes.utils.types import (
+    numpy_ints, numpy_floats, complex_types, complex_type_union
+)
 
 
 # utility function to allow sqlite/numpy type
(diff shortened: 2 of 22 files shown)


qcodes/dataset/sqlite/queries.py
```diff
@@ -318,7 +318,7 @@ def _expand_data_to_arrays(data: List[List[Any]], paramspecs: Sequence[ParamSpec
             for element in numeric_elms:
                 row[element] = np.full_like(row[first_array_element],
                                             row[element],
-                                            dtype=np.float)
+                                            dtype=np.double)
```
jenshnielsen, line 321: Probably cleaner `np.dtype(np.float64)`

astafan8: because of the explicit "64"?

qcodes/dataset/sqlite/database.py
```diff
@@ -150,14 +150,14 @@ def connect(name: str, debug: bool = False,
 
     # Make sure numpy ints and floats types are inserted properly
     for numpy_int in [
-        np.int, np.int8, np.int16, np.int32, np.int64,
+        np.int_, np.int8, np.int16, np.int32, np.int64,
```
astafan8, line 153: don't we need to add `int` similar to floats below? also, do i understand correctly that the current behavior (despite that names) is that python's builtin types are ALSO going through these adapters because np.int and np.float include int and float?

qcodes/dataset/sqlite/queries.py
astafan8: analogiues to complex128 below, i'd suport float64 here.

qcodes/tests/helpers/test_json_encoder.py
```diff
@@ -51,7 +51,7 @@ def test_numpy_int_types():
 def test_numpy_float_types():
     e = NumpyJSONEncoder()
 
-    numpy_floats = (np.float, np.float_, np.float16, np.float32,
+    numpy_floats = (np.float_, np.float_, np.float16, np.float32,
```
astafan8, line 54: duplicate
```suggestion
    numpy_floats = (np.float_, np.float16, np.float32,
```

qcodes/tests/validators/test_complex.py
```diff
@@ -11,7 +11,7 @@ def test_complex(complex_val):
     n = ComplexNumbers()
     assert str(n) == '<Complex Number>'
     n.validate(complex_val)
-    n.validate(np.complex(complex_val))
+    n.validate(np.complex_(complex_val))
```
astafan8, line 14: is there a value in this? shou]ldn't we just rather use python's `complex`?

astafan8 (commented):
> while this does run with older versions of numpy it does not type check

you mean that it doesn't typecheck with "old verions of numpy" (and the current version of stubs)? then it sounds like the best thing to do is to wait for 1.20 to be released, merge this PR immediately then, and never worry about 1.19 and before ever again.

qcodes/dataset/sqlite/queries.py
jenshnielsen: Yes, Double is always equivalent to 64 bit as far as I know.

qcodes/dataset/sqlite/database.py
jenshnielsen: We actually need the regular (non numpy) float type to go through the adapter to support correct handling of NaN. There is no such requirement for int.

qcodes/dataset/data_export.py
```diff
@@ -384,8 +384,8 @@ def reshape_2D_data(x: np.ndarray, y: np.ndarray, z: np.ndarray
         z_to_plot = np.full((ny, nx), '', dtype=z.dtype)
     else:
         z_to_plot = np.full((ny, nx), np.nan)
-    x_index = np.zeros_like(x, dtype=np.int)
-    y_index = np.zeros_like(y, dtype=np.int)
+    x_index = np.zeros_like(x, dtype=int)
+    y_index = np.zeros_like(y, dtype=int)
```
jenshnielsen, line 388: Perhaps it's cleaner to use the numpy int type: `np.int_` or a sized one `np.int32`

qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -495,7 +495,7 @@ def test_numpy_ints(dataset):
     dataset.mark_started()
 
     numpy_ints = [
-        np.int, np.int8, np.int16, np.int32, np.int64,
+        np.int_, np.int8, np.int16, np.int32, np.int64,
         np.uint, np.uint8, np.uint16, np.uint32, np.uint64
     ]
```
jenshnielsen, line 500: import these from types.py ?

qcodes/tests/dataset/test_dataset_basic.py
```diff
@@ -514,7 +514,7 @@ def test_numpy_floats(dataset):
     dataset.set_interdependencies(idps)
     dataset.mark_started()
 
-    numpy_floats = [np.float, np.float16, np.float32, np.float64]
+    numpy_floats = [np.float_, np.float16, np.float32, np.float64]
     results = [{"y": tp(1.2)} for tp in numpy_floats]
```
jenshnielsen, line 518: As above  import these from types.py ?

qcodes/tests/helpers/test_json_encoder.py
```diff
@@ -39,7 +39,7 @@ def test_complex_types():
 def test_numpy_int_types():
     e = NumpyJSONEncoder()
 
-    numpy_ints = (np.int, np.int_, np.int8, np.int16, np.int32,
+    numpy_ints = (np.int_, np.int_, np.int8, np.int16, np.int32,
```
jenshnielsen, line 42: import these from types.py ?

qcodes/tests/helpers/test_json_encoder.py
```diff
@@ -51,7 +51,7 @@ def test_numpy_int_types():
 def test_numpy_float_types():
     e = NumpyJSONEncoder()
 
-    numpy_floats = (np.float, np.float_, np.float16, np.float32,
+    numpy_floats = (np.float_, np.float16, np.float32,
                     np.float64)
```
jenshnielsen, line 56: import these from types.py ?

qcodes/utils/types.py
```diff
@@ -8,22 +8,22 @@
 
 numpy_concrete_ints = (np.int8, np.int16, np.int32, np.int64,
                        np.uint8, np.uint16, np.uint32, np.uint64)
-numpy_non_concrete_ints_instantiable = (np.int, np.int_, np.uint)
+numpy_non_concrete_ints_instantiable = (np.int_, np.uint)
 numpy_non_concrete_ints = \
     numpy_non_concrete_ints_instantiable + (np.integer,)
 
 numpy_concrete_floats = (np.float16, np.float32, np.float64)
```
jenshnielsen, line 15: add `np.double` which is normally (always?) an alias for `np.float64`   

Also consider adding `np.longdouble` which is a `80` or `128` bit float

qcodes/utils/types.py
```diff
@@ -8,22 +8,22 @@
...
-numpy_non_concrete_floats_instantiable = (np.float, np.float_)
+numpy_non_concrete_floats_instantiable = (np.float_, )
 numpy_non_concrete_floats = \
     numpy_non_concrete_floats_instantiable + (np.floating,)
 
 numpy_concrete_complex = (np.complex64, np.complex128)
-numpy_non_concrete_complex = (np.complex, np.complex_, np.complexfloating)
+numpy_non_concrete_complex = (np.complex_, np.complexfloating)
```
jenshnielsen, line 21: Add a numpy_non_concrete_complex_instantiable type alias also

qcodes/instrument_drivers/agilent/E8267C.py
```diff
@@ -84,9 +84,9 @@ def __init__(self, name: str, address: str, **kwargs: Any) -> None:
 
     # functions to convert between rad and deg
     @staticmethod
-    def deg_to_rad(angle_deg: numbertypes) -> np.float:
+    def deg_to_rad(angle_deg: numbertypes) -> "np.floating[Any]":
```
jenshnielsen, line 87: Consider replacing with something not Any https://numpy.org/devdocs/reference/typing.html#number-precision

jenshnielsen:
@astafan8 I realized that we can just get this passing with older numpy versions by removing the stubs. I think this is ready for review but I would like to merge https://github.com/QCoDeS/Qcodes/pull/2569 first so we ensure that this does not break any older numpy versions

astafan8:
@jenshnielsen ok, good! is it worth waiting until the stubs start matching the types of numpy 1.20? or is the issue somewhere else? I guess since the stubs are not an official released project (right?), we'd better just stick to numpy 1.20 since it's the latest nunmpy that will alweays be installed when we run mypy in CI?

jenshnielsen:
@astafan8 The stubs are deprecated and will never be updated. The types in numpy 1.20 are the official source of truth

PR #2501, opened 2020-12-08
throw in test if not all guids unique

jenshnielsen:
This test generates a lot of datasets quickly after each other. There is a chance that these may not have unique guids

### DIFF
--- a/qcodes/tests/dataset/test_database_extract_runs.py
+++ b/qcodes/tests/dataset/test_database_extract_runs.py
@@ -472,6 +472,12 @@ def test_combine_runs(two_empty_temp_db_connections,
     source_2_datasets = [DataSet(conn=source_conn_2,
                                  exp_id=source_2_exp.exp_id) for i in range(10)]
 
+    guids_1 = set(dataset.guid for dataset in source_1_datasets)
+    guids_2 = set(dataset.guid for dataset in source_2_datasets)
+
+    guids = guids_1 | guids_2
+    assert len(guids) == 20
+
     source_all_datasets = source_1_datasets + source_2_datasets
 
     shuffled_datasets = source_all_datasets.copy()


qcodes/tests/dataset/test_database_extract_runs.py
```diff
@@ -472,6 +472,12 @@ def test_combine_runs(two_empty_temp_db_connections,
     source_2_datasets = [DataSet(conn=source_conn_2,
                                  exp_id=source_2_exp.exp_id) for i in range(10)]
 
+    guids_1 = {dataset.guid for dataset in source_1_datasets}
+    guids_2 = {dataset.guid for dataset in source_2_datasets}
+
+    guids = guids_1 | guids_2
+    assert len(guids) == 20
```
astafan8, line 479: shall we just do a "retry" if this happens? don't know if pytest supports such a feautre. another way is to make the list of datasets a hypothesys strategy that throws away examples with duplicate guids

jenshnielsen: IMHO no this must never happen. We should fix the underlying implementation to ensure that this cannot happen

jenshnielsen: Rather than putting this here we should perhaps just add a new test that does only this. Generates ~ 100 guids and ensure that they are all unique

trevormorgan: Shall I put in a test to generate 100 guids and check that they're all unique then to get this PR closed? Or did we want to keep it open to solve the underlying problem?

astafan8: > keep it open to solve the underlying problem

i'd vote for this

trevormorgan: Do either of you have an idea on why this happens?

jenshnielsen:
folded into https://github.com/QCoDeS/Qcodes/pull/4763

PR #2502, opened 2020-12-08
Add DC bias and measurement time params to Keysight E4980A driver 

Akshita07:
- Adds DC Bias parameters
- Adds measurement time parameter
- Adds method to check installed option number

To Do:
- [x] Check and add valid voltage range
- [x] Update meas time parameter (implementation in progress)
- [x] Test updates on actual instrument

@astafan8


qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -293,6 +320,59 @@ def _set_measurement(self,
...
+            raise RuntimeError("Option 001 is not installed. Hence, DC Bias "
+                               "voltage level cannot be set.")
+
+        self.write(f":BIAS:VOLTage[:LEVel] {val}")
+
+    def _get_dc_bias_voltage_level(self) -> float:
+        """
+        Gets DC Bias voltage level.
+        """
+        if self.option_number() != "001":
```
astafan8, line 357: shall we perhaps not add these parameters if the option is not installed? for an example of this from another driver look for ".has_DIG" attribute on keysight 34465A driver.

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -293,6 +320,59 @@ def _set_measurement(self,
...
+        return float(self.ask(":BIAS:VOLTage[:LEVel]?"))
+
+    def _set_measurement_time(self, duration: str, avg_rate: int) -> None:
+        """
+        Sets measurement time and averaging rate. Measurement time duration
+        can be short, medium or long.
+        """
+        self.write(f":APERture {duration},{avg_rate}")
+
+    def option_number(self) -> str:
```
astafan8, line 370: shall we also tech the driver to decode vavrious return values of the "option" and create boolean attributes (or readonly protperties) for them?

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -293,6 +320,59 @@ def _set_measurement(self,
...
+                               "get DC Bias voltage level")
+
+        return float(self.ask(":BIAS:VOLTage[:LEVel]?"))
+
+    def _set_measurement_time(self, duration: str, avg_rate: int) -> None:
+        """
+        Sets measurement time and averaging rate. Measurement time duration
+        can be short, medium or long.
+        """
+        self.write(f":APERture {duration},{avg_rate}")
```
astafan8, line 368: looks like a candidate for a group parameter since two values are set/get by single command

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -236,6 +237,32 @@ def __init__(self,
...
+            unit="V",
+            docstring="Sets the DC bias voltage. Setting does not implicitly "
+                      "turn the DC bias ON."
+        )
+
+        self.add_parameter(
+            "measurement_time",
+            get_cmd=":APERture?",
+            set_cmd=self._set_measurement_time,
+            docstring="Sets the measurement time mode and the averaging rate."
```
astafan8, line 263: typo?
```suggestion
            docstring="Sets the measurement time and the averaging rate."
```

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
Akshita07: Updated as discussed.

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
Akshita07: No, it is not typo. It is the mode which is being set.

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -1,9 +1,14 @@
...
 
 from qcodes import VisaInstrument, InstrumentChannel
 from qcodes.instrument.parameter import MultiParameter
 from qcodes.utils.helpers import create_on_off_val_mapping
-from qcodes.utils.validators import Enum, Numbers, Bool
+from qcodes.utils.validators import Enum, Numbers, Bool, Ints
+from qcodes.instrument.group_parameter import GroupParameter, Group
+
+
+logger = logging.getLogger(__name__)
```
jenshnielsen, line 11: The instrument already has a logger attached at `self.log`

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -236,6 +248,53 @@ def __init__(self,
                       "the auto range function OFF."
         )
```
jenshnielsen, line 250: Now some (all?) of these parameters should only be added if has_option_001 is True right?

Akshita07: No, these parameters are still there but with different allowed ranges. All the implemented parameters take ranges which are allowed with option 001.

jenshnielsen: Ok that just doesn't match your original implementation nor the comment on @astafan8 s comment. 

If the ranges depend on the option they should be set according to the option so they are always correct.

Akshita07: Yes, my original implementation was not correct. I had misunderstood how the options are making a difference. My comment 'Updated as discussed' reflects an offline discussion I had will @astafan8 regarding updating all parameters according to options.

Sure, I'll update the parameters accordingly.

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -231,11 +254,71 @@ def __init__(self,
...
+            unit="V",
+            vals=self._dc_bias_v_level_range,
+            docstring="Sets the DC bias voltage. Setting does not "
+                      "implicitly turn the DC bias ON."
+        )
+
+        self.add_parameter(
+            "meas_time_mode",
+            initial_value="MED",
+            vals=Enum("SHOR", "MED", "LONG"),
```
astafan8, line 287: i'd recommend to decouple the qcodes driver from the specifics of the VISA command values and use val_mapping here:
```
val_mapping={"short": "SHOR", "medium": "MED", "long": "LONG"}
```
this makes the code that uses the driver more readable. and removes the need for explaining the values in the docstring (i'm referring to the `It can be SHOR (short), MED (medium - default) or LONG.`  sentence :) )

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -231,11 +254,71 @@ def __init__(self,
...
+            parameter_class=GroupParameter,
+            docstring="Measurement time mode holds the mode of measurement "
```
astafan8, line 289: speaking of comments and documentation: `Measurement time mode holds the mode of measurement time` is a redundant information, right? so either there's no need for a doctrsing for the case of this parameter or some more relevant and useful information needs to be added, would you agree?

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -231,11 +254,71 @@ def __init__(self,
...
+
+        self.add_parameter(
+            "averaging_rate",
+            initial_value=1,
+            vals=Ints(1, 256),
+            parameter_class=GroupParameter,
+            docstring="Averaging rate for the measurement."
+        )
+
+        self._set_measurement_time = Group(
```
astafan8, line 302: a better, more specific name? also, it's just a parameter group as opposed to some cool private method that "sets measurement time", right?
```suggestion
        self._aperture_group = Group(
```

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -231,11 +254,71 @@ def __init__(self,
...
+        if self.has_firmware_a_02_10_or_above:
+            self.add_parameter(
+                "dc_bias_range_auto",
+                get_cmd=":BIAS:RANGe:AUTO?",
+                set_cmd=":BIAS:RANGe:AUTO {}",
+                vals=Bool(),
+                val_mapping=create_on_off_val_mapping(on_val="1",
+                                                      off_val="0"),
+                docstring="Enables DC Bias range AUTO setting. When DC bias "
+                          "range is FIX (AUTO disabled), '#' is displayed in "
```
astafan8, line 318: i'm a bit confused with this docstring. The driver does not have a parameter that specifies the range for the DC bias, so if i do `dc_bias_range_auto(False)`, how will i be able to set the "fixed" range for the DC bias? even, should i be able to set a specific "fixed" range for the DC bias?

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -231,11 +254,71 @@ def __init__(self,
...
+        self._set_measurement_time = Group(
+            [self.meas_time_mode,
+             self.averaging_rate],
+            set_cmd=":APERture {meas_time_mode},{averaging_rate}",
+            get_cmd=":APERture?"
+        )
+
+        if self.has_firmware_a_02_10_or_above:
+            self.add_parameter(
+                "dc_bias_range_auto",
```
astafan8, line 311: perhaps `dc_bias_autorange_enabled`? naming is important, right?

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -293,6 +376,14 @@ def _set_measurement(self,
         self._measurement_pair = measurement_pair
         self.write(f":FUNCtion:IMPedance {measurement_pair.name}")
 
+    def _options(self) -> Tuple[str, ...]:
+        """
+        Returns installed options numbers. The options are Power/DC Bias
```
astafan8, line 381: right? "possible"? or are these more options?
```suggestion
        Returns installed options numbers. The possible options are Power/DC Bias
```

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -184,6 +186,26 @@ def __init__(self,
         """
         super().__init__(name, address, terminator=terminator, **kwargs)
 
+        idn = self.IDN.get()
+
+        self.has_firmware_a_02_10_or_above = LooseVersion(idn["firmware"]) >=\
+                                             LooseVersion("A.02.10")
+
+        self.has_option_001 = '001' in self._options()
```
astafan8, line 194: what are the possible outputs of the options method? is it one of `"", "001", "002"` or else? or are combinations possible?

i just wonder whether we should define these "001"s are contants as opposed to describing them in plain text in the docstring of _options method

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
Akshita07: yes

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
Akshita07: Yes, there are more options possible

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
Akshita07: combinations are possible, options method returns a tuple of installed options

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
astafan8: should then the sentence be reworded so that it's clear that it mentions only two options while there are actually more options?

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
astafan8: ok, then it's good as is :)

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
astafan8: discussed offline that "FIX" is not something that the instrument "has", it's just capitalized word "fix" (probably should've been fixed, with 'ed'), so all is good here. I would still suggest to reword it a bit like below:
```suggestion
                          "range is fixed (not "AUTO"), '#' is displayed in "
```

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -231,11 +254,68 @@ def __init__(self,
             get_cmd=":FUNCtion:IMPedance:RANGe?",
             set_cmd=":FUNCtion:IMPedance:RANGe {}",
             unit='Ohm',
-            vals=Enum(0.1, 1, 10, 100, 300, 1000, 3000, 10000, 30000, 100000),
+            vals=self._imp_range,
             docstring="Selects the impedance measurement range, also turns "
                       "the auto range function OFF."
```
astafan8, line 259: dang, i just noticed this! is this "auto range" related to the DC bias "autorange_enabled"? or is it related to impedance? (if it's the latter, then all is good, it's just that apparently the driver is missing a parameter that lets the user do "auto ranging" of the impeadance measurement; if it's the former, then we should change the set command of this parameter to call get() on the dc_bias_autorange_enabled parameter so that it gets updated when a new impedance range is being set)

astafan8 (approved):
Looks great!

Just one small question/note, hopefully without the need for any adjustments

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
Akshita07: This parameter sets the range for impedance measurement. There is another command which sets auto range for impedance - `:FUNCtion:IMPedance:RANGe:AUTO {ON|OFF|1|0}` which is not implemented in this driver.

PR #2507, opened 2020-12-10
feature: add native xarray support and concatenation of DataFrames/DataArrays

nikhartman:
- add `get_data_as_xarray` for xarray, similar to `get_data_as_pandas_dataframe`
- add a kwarg to `get_data_as_pandas_dataframe` to concatenate individual DataFrames (similar feature in `get_data_as_xarray`)
    - print a warning if independent parameters are not consistent across all dependent parameters when trying to use `concat`

@WilliamHPNielsen per our discussion earlier.


docs_requirements.txt
```diff
@@ -40,6 +40,7 @@ notebook==6.1.5
 numpy==1.19.3
 packaging==20.7
 pandas==1.1.5
+xarray==0.16.2
 pandocfilters==1.4.3
```
jenshnielsen, line 44: Please sort .txt files alphabetically. Xarray is already listed under X in this file :)

trevormorgan: will do!

qcodes/dataset/data_set.py
```diff
@@ -1024,6 +1074,82 @@ def _load_to_dataframes(self, datadict: ParameterData) -> Dict[str, "pd.DataFram
...
+        datadict = self.get_parameter_data(*params,
+                                           start=start,
+                                           end=end)
+
+        data_arrs = {}
+        for name, subdict in datadict.items():
+            index = self._generate_pandas_index(subdict)
+            arr = self._data_to_dataframe(subdict, index).to_xarray()[name]
+            arr.attrs["label"] = self.paramspecs[name].label
+            arr.attrs["unit"] = self.paramspecs[name].unit
```
samantha-ho, line 1134: Why not include all of the paramspec elements, rather than just label and unit?

nikhartman: Those were the minimum sweet of things I could imagine wanting. By all means, add more.

qcodes/dataset/data_set.py
```diff
@@ -1141,14 +1141,10 @@ def get_data_as_xarray(self,
             warnings.warn('Independent parameter setpoints are not equal. Check concatenated output carefully.')
 
         xds = xr.Dataset(data_arrs)
-        for dim in xds.dims:
```
nikhartman: Why remove this dimension info?

trevormorgan: I might be mistaken but I think this is handled by `xds = xr.Dataset(data_arrs)`. At least when I'm viewing a create xarray dataset object after this change the dimension info still shows up and looks right. ''

qcodes/dataset/data_set.py
```diff
@@ -961,22 +999,39 @@ def get_data_as_pandas_dataframe(self,
                 ParamSpec objects. If no parameters are supplied data for
                 all parameters that are not a dependency of another
                 parameter will be returned.
+            concat: if True individual DataFrames are concatenated along columns
+                and a single DataFrame is returned for the entire DataSet
```
jenshnielsen, line 1003: Having (public) methods that return different types is imho a code smell and should almost always be avoided. 

I think its better to have 2 different methods for this. If we really want to go this way we should use overload so that 
it is clear from the type system which type we get back

qcodes/dataset/data_set.py
```diff
@@ -1024,6 +1079,84 @@ def _load_to_dataframes(self, datadict: ParameterData) -> Dict[str, "pd.DataFram
...
         return dfs
 
+    def get_data_as_xarray(self,
+                           *params: Union[str,
+                                          ParamSpec,
+                                          _BaseParameter],
+                           concat: Optional[bool] = False,
+                           start: Optional[int] = None,
+                           end: Optional[int] = None) -> \
+            Union[Mapping[Hashable, "xr.DataArray"], "xr.Dataset"]:
```
jenshnielsen, line 1089: Why Hashable here. Is this not always a string?. I would also suggest never to use Mapping or any other generic type in return position. We know this is s dict

qcodes/dataset/data_set.py
jenshnielsen: I would suggest leaving this method as is and probably deprecate and then create 4 new methods.
for pandas.

`to_pandas_dataframe`
`to_pandas_dataframe_dict`

for xarray
`to_xarray_dataset
`to_xarray_dataarray_dict`
or something similar

qcodes/dataset/data_set.py
```diff
@@ -929,13 +936,44 @@ def get_parameter_data(
...
+
+        def _get_setpoints(dd: ParameterData) -> Iterator[Dict[str, numpy.ndarray]]:
+
+            for dep_name, param_dict in dd.items():
+                out = {
+                    name: vals for name, vals in param_dict.items() if name != dep_name
+                }
+                yield out
+
+        iterator = _get_setpoints(datadict)
```
jenshnielsen, line 960: sp_iterator perhaps?

requirements.txt
```diff
@@ -87,6 +87,7 @@ websockets~=8.1
 widgetsnbextension~=3.5.1
 wincertstore~=0.2
 wrapt~=1.12.1
+xarray~=0.16.2
```
jenshnielsen, line 90: xarray should also be added as a real dependency to setup.py

jenshnielsen:
I think we should also update https://qcodes.github.io/Qcodes/examples/DataSet/Working-With-Pandas-and-XArray.html?highlight=xarray

qcodes/dataset/data_set.py
nikhartman: I think that suggestion to deprecate and add new methods is a good one. It always bothered me that `get_data_as_pandas_dataframe` didn't actually return a dataframe. Having a function where the `concat=True` behavior is default is great (but maybe the warning about setpoints needs to stay?). Making a distinction between `xarray.Dataset` and `xarray.DataArray` in the two suggested xarray functions is also nice.

qcodes/dataset/data_set.py
trevormorgan: You'd think `Dict[str, xr.DataArray]` would be correct but really struggled to get typing to pass. I think they're using generic types for xarray based on this [PR ](https://github.com/pydata/xarray/pull/2929#issue-274455843).

jenshnielsen: I am confused how that can matter. We create the variable that this is the type of here 
```
data_arrs: MutableMapping[Hashable, xr.DataArray] = {}
```

jenshnielsen: I guess the problem is 
```
qcodes\dataset\data_set.py:1150:26: error: Argument 1 to "Dataset" has incompatible type "Dict[str, DataArray]"; expected "Mapping[Hashable, Any]"
```

That should imho not be solved by changing our types. I think a better solution is to simply cast before passing to that function. 

The problem is that Mapping is not covariant in the first type so even if str is considered a subtype of Hashable
Dict[str, Any] is not a subtype of Dict[Hashable, Any] for that reason I would normally not think it worth the trouble of using Hashable for the key type until https://github.com/python/mypy/issues/1114 is resolved

qcodes/dataset/data_set.py
trevormorgan: good point Jens and Nik, having separate methods will be much cleaner.

qcodes/dataset/data_set.py
trevormorgan: Ahh that's right the problem is with xarray's `Dataset` type being incompatible with `Dict[str, DataArray]`. Thanks for explaining the core of the issue, I was getting pretty confused why a subtype of Hashable was not working. 

I'll cast before passing into the function and leave a note about mypy#1114 typing#445.

qcodes/dataset/data_set.py
```diff
@@ -949,22 +950,17 @@ def _parameter_data_identical(param_dict_a: Dict[str, numpy.ndarray],
...
-        def _get_setpoints(dd: ParameterData) -> Iterator[Dict[str, numpy.ndarray]]:
+        def _sp_iterator(dd: ParameterData) -> Iterator[Dict[str, numpy.ndarray]]:
 
             for dep_name, param_dict in dd.items():
                 out = {
                     name: vals for name, vals in param_dict.items() if name != dep_name
                 }
                 yield out
 
-        iterator = _get_setpoints(datadict)
```
jenshnielsen: I actually meant having a variable called iterator is not very descriptive

trevormorgan: ahh thanks, agreed. I'll get that name change in.

qcodes/dataset/data_set.py
```diff
@@ -1292,7 +1572,7 @@ def _finalize_res_dict_numeric_text_or_complex(
 
         return res_list
 
-    @staticmethod
+    @ staticmethod
```
jenshnielsen, line 1575: These spaces before staticmethod look odd to me

trevormorgan: bleh I tried to set auto formatting on vscode to remove white spaces and every time I save it puts a space in from of # and @. Thanks for pointing this out, I'll fix and turn off the auto formatter.

jenshnielsen: That seems less than helpful ;)

jenshnielsen (approved):

PR #2511, opened 2020-12-10
Add config options for dataset write_in_background and write_periode

jenshnielsen:
Make it possible to set these globally. 

Along the way

* Remove unsupported plottr subscriber. 

@ThorvaldLarsen


jenshnielsen:
Both the documentation on config in general and the write options should be improved. I will do that in a subsequent pr

astafan8 (approved):

ThorvaldLarsen:
@jenshnielsen @astafan8 this is a great solution to changes in doNd defaults in my opinion that allows everyone the settings they want rather than being stuck with the defaults in qcodes core as what most (vocal) users would like at any given time.

PR #2512, opened 2020-12-10
Updating doNd notebook with comparing Measurement context manager with qcodes doNd 

FarBo:
This PR compares two things in 4 sections:
Comparing 1D and 2D sweeps when z parameter depends on x and y parameters sweeps (This is not the goal of this PR anymore)
Comparing the use of Measurement context manager with qcodes doNd.

Update: After careful consideration, it is decided to extend [this existing notebook](https://qcodes.github.io/Qcodes/examples/DataSet/Using_the_doNd_functions_for_running_experiments.html) with the content of Measurement.


docs/examples/DataSet/Deeper dive on qcodes doNd functionalities with a focus on arrayed sweeping benefits.ipynb
```diff
@@ -0,0 +1,455 @@
...
+    "# and start recording\n",
+    "t1 = time.time()\n",
+    "with meas.run() as datasaver:\n",
+    "    for xval, yval, zval in very_simple_2d_sweep():\n",
+    "        datasaver.add_result(\n",
+    "            ('x', xval),\n",
+    "            ('y', yval),\n",
+    "            ('z', zval),\n",
+    "        )\n",
+    "        time.sleep(0.1)\n",
```
astafan8, line 118: i have a question about the "timings comparison". I clearly don't udnerstand something about this example notebook :)

so, if the goal is to "simulate" a measurement that take 0.1s per point vs a buffered measurement that takes 0.1s per buffer (array of y), then i'm not sure i get the point of this notebook because this comparison can be done without any code, right?

now, if the point is to demonstrate how to transition from point-per-point measurement to buffer-per-buffer measurement code using Measurement() and doNd(), then the notebook clearly lacks clarity on explaining its point. It should be very clear to the user what is being compared and what is demonstrated and how, otherwise i read this notebook as proving the obvious in a too abstract way.

astafan8, line 118: i think these sleep times can be made 20/50/100 times smaller so that the notebook runs faster but still the difference in measurement times would be visible, am i wrong with this suggestion?

also, i don't think we want to wait for 2x298sec for this notebook to execute on CI...

docs/examples/DataSet/Deeper dive on qcodes doNd functionalities with a focus on arrayed sweeping benefits.ipynb
```diff
@@ -0,0 +1,455 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "xvals = np.linspace(-5, 5, 51)\n",
+    "yvals = np.linspace(-5, 5, 51)\n",
+    "xx, yy = np.meshgrid(xvals, yvals, indexing='ij')\n",
+    "zz = np.cos(xx) * np.sin(yy)\n",
+    "\n",
+    "def very_simple_2d_sweep():\n",
```
astafan8, line 90: i think that extracting the "loops" into helper functions makes the narrative more confusing than it helps. i think that this is a good case where a comment "very_simple_2d_sweep" next to the "loops" inside the measurement context manager will actually work better than the function. Moreover, perhaps these names/comments should be more specific? for example, point_by_point_x_y_sweep and point_by_point_x_sweep_with_y_buffered_sweep (along these lines but much better)?

docs/examples/DataSet/Deeper dive on qcodes doNd functionalities with a focus on arrayed sweeping benefits.ipynb
```diff
@@ -0,0 +1,455 @@
...
+   "source": [
+    "# Deeper dive on qcodes doNd functionalities with a focus on arrayed sweeping benefits"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this example notebook, we want to show how the qcodes measurement can be simplified by using qcodes doNd functions. Also, we test 2-dimentional sweeping with a naive point by point sweeping over x and y axis and measure z axis values compared to sweeping over x axis and considering y axis as an array of numbers to measure z axis values. The array method can be useful in buffered data acquisitions from instruments that have buffer to write data in it (this notebook is not aiming to show how to setup buffer). Therefore, the measurement time can be enhanced significantly."
+   ]
```
astafan8, line 15: i do suggest to explain/describe what a buffered data acquistion is meant here broader. "instruments that have buffer to write data in it" reads more as instrument's implementation detail rather than approach where you tell the instrument what to sweep and it returns you arrays as opposed to setting/getting parmaeters value by value.

docs/examples/DataSet/Deeper dive on qcodes doNd functionalities with a focus on arrayed sweeping benefits.ipynb
```diff
@@ -0,0 +1,455 @@
...
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.7.9"
+  },
+  "nbsphinx": {
+   "execute": "never"
```
jenshnielsen, line 450: Why is execution disabled?

astafan8: cause it's slow :) in one of the comments i'm wondering if it can be sped up to always execute

docs/examples/DataSet/Deeper dive on qcodes doNd functionalities with a focus on arrayed sweeping benefits.ipynb
```diff
@@ -0,0 +1,455 @@
+{
```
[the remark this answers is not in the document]
jenshnielsen: A parameter that returns an Array should be either an array parameter or a parameter with setpoints so the shape can be specified

docs/examples/DataSet/Deeper dive on qcodes doNd functionalities with a focus on arrayed sweeping benefits.ipynb
FarBo: Actually, your question is because of that the initial goal of the notebook was to show the difference between Measurement and doNd. The point by point comparison with buffered added later, so that's might be the reason you can't find cohesion in the notebook. The main point is to show the use case of do1d and do2d compared to Measurement. So, timing and all buffered stuff were the last minute addition to have extra context in these comparisons.

That's said, do you think I should only focus on the Measurement and doNd or what?

docs/examples/DataSet/Deeper dive on qcodes doNd functionalities with a focus on arrayed sweeping benefits.ipynb
FarBo: That's right. It takes nearly 10 minutes to run the notebook.

jenshnielsen: As discussed above we should probably fix the long execution instead. 

In the future please document in the git log message why execution is disabled

docs/examples/DataSet/Deeper dive on qcodes doNd functionalities with a focus on arrayed sweeping benefits.ipynb
astafan8: then it sounds like there should be two notebooks - one about Measurement vs doNd, and another one about buffered acquisitions (which would include a section showing that buffered acquisition with doNd functions also "works"). Would you agree?

FarBo: Sounds fine with me. I will change this notebook scope to only cover Measurement VS doNd. I will try to add sections to cover all doNd functions compared to Measurement. We are lacking do0d in this notebook.

FarBo:
@jenshnielsen 
@astafan8 

Could you look at the modifications of the notebook. If you think, it is acceptable in its current form, we can have it in the release.

astafan8 (commented):
`Meaurement took 41.3532292842865 seconds` - this is too long. Can we reduce this to ~seconds while still making a point? do we even need the sleep() calls now that "timings of buffered aquisition" is not in scope?

FarBo:
> this is too long.
This is the minimum sleep. I kept this because earlier in the notebook, I promoted using plottr to see the live plot. For having that, We need sleep time. One thing that I can do is to give the sleep for the outer loop. Then It takes 5 to 6 seconds.

docs/examples/DataSet/Comparing qcodes Measurement context manager with qcodes doNd functions.ipynb
```diff
@@ -0,0 +1,443 @@
...
+      "Starting experimental run with id: 1. \n",
+      "Meaurement took 5.572435140609741 seconds\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Setting up parameters\n",
+    "x = Parameter('x', set_cmd=None)\n",
+    "y = Parameter('y', get_cmd=None)\n",
+    "z = Parameter('z', get_cmd=None)\n",
```
jenshnielsen, line 114: To make this clearer we should use the same parameters in all examples. E.g. add the lambda functions here and reuse these parameters throughout the examples

FarBo: This is a good idea. But, there should be two sets of parameters, one for 1D and the other one for 2D because parameters are slightly different in two section (only x is the same in all cases). I can segregate them by saying x1, y1, z1 and x2, y2 and z2.  What do you think?

jenshnielsen: I suggest that you create all the parameters up front such that you have something like the following:
Otherwise y will change from a independent to a dependent parameter.
```
x = Parameter('x', set_cmd=None)
y = Parameter('y', set_cmd=None)
z1 = Parameter('z1', get_cmd=lambda: np.sin(x())) 
z3 = Parameter('z2', get_cmd=lambda: np.cos(x())
z3 = Parameter('z1', get_cmd=lambda: np.cos(x())*np.sin(y())) 
```

FarBo: This is much cleaner, thanks.

jenshnielsen: @FarBo if this has been done please feel free to press the resolve conversation button ;) That makes it easier to see if this is ready for a new review

FarBo:
@jenshnielsen 
My bad not setting the conversations resolved. Thanks for reminding it. I think all comments have been addressed now, hopefully :).

jenshnielsen:
I am a bit confused as how this compares to the [existing dond notebook](https://qcodes.github.io/Qcodes/examples/DataSet/Using_the_doNd_functions_for_running_experiments.html). There seems to be a lot of duplicated content. Perhaps it is better to extend that notebook to explain how the experiments in that notebook can be done using the existing measurement framwork?

FarBo:
The modified version of this notebook caused this duplication. In the current format of this notebook, It's reasonable to extend the other notebook with Measurement comparison. It might need an updated title such as : **Using doNd functions in comparison to Measurement for performing measurements** or something similar.

docs/examples/DataSet/Using_the_doNd_functions_for_running_experiments.ipynb
```diff
@@ -225,79 +439,92 @@
...
     "\n",
-    "Internally, the `doNd` functions return a tuple consists of the dataset obtained in the measurement, the boolean value correponding to `do_plot` and a callable which checks whether the experiment is interupted by the user. Thus, if the `plot` shall be used explicitly after the measurement, one has to pass the first argument of the return value. "
+    "Internally, the `doNd` functions return a tuple consists of the dataset obtained in the measurement, the boolean value correponding to `do_plot` and a callable which checks whether the experiment is interupted by the user. Thus, if the `plot` shall be used explicitly after the measurement, one has to pass the first argument of the return value.\n",
+    "\n",
+    "Moreover, it is possible to pass an already existing experiment as an argument to `doNd` constructor. This is usually a convenient way as users may prefer not to create a new experiment before running a measurement with `doNd`s. "
```
jenshnielsen, line 464: Suggest something like

"
As with the Measurement context manager it is possible to pass an explicit experiment to the `doNd` functions. This makes it possible to switch between experiments when using the `doNd` functions

jenshnielsen (commented):
Left some comments inline. The overall structure is good but we need to work a bit on the writing

docs/examples/DataSet/Using_the_doNd_functions_for_running_experiments.ipynb
```diff
@@ -103,38 +109,140 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Starting experimental run with id: 2. \n"
      ]
     }
    ],
    "source": [
+    "# Creating an experiment\n",
+    "exp2 = load_or_create_experiment('1D_do1d', sample_name='no sample')\n",
```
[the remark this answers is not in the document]
FarBo: This is only because I want to use to different experiment names in the last section of the notebook. Otherwise, the plots title will be wrong.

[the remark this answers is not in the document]
jenshnielsen: Just use a generic experiment name that does not refer to either but just the notebook as a whole.

docs/examples/DataSet/Using_the_doNd_functions_for_running_experiments.ipynb
```diff
@@ -225,79 +439,92 @@
...
    "source": [
-    "Note that `do_plot = True` is the default value and we only added it for the sake of explicitness. The `doNd` utility, in fact, uses an internal public function called `plot` to handle data visualization, which by default saves the figures in both pdf and png formats, as well. Therefore, users, if needed, can only do their measurements with the option `do_plot = False` and then pass the relevant dataset to `plot` when visualization is necessary and choose to save figures in a desired format.\n",
+    "### Handling plot and experiment with doNd"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The `doNd` utility, in fact, uses an internal public function called `plot` to handle data visualization, which by default saves the figures in both pdf and png formats, as well. Therefore, users, if needed, can only do their measurements leaving to set `do_plot`, which by default is False, then pass the relevant dataset to `plot` when visualization is necessary and choose to save figures in a desired format.\n",
```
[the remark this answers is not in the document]
FarBo: This sentence was there already :). I only deleted the part, which was mentioning the do_plot is True by default.

docs/examples/DataSet/Using_the_doNd_functions_for_running_experiments.ipynb
```diff
@@ -225,79 +439,92 @@
...
     "\n",
-    "Internally, the `doNd` functions return a tuple consists of the dataset obtained in the measurement, the boolean value correponding to `do_plot` and a callable which checks whether the experiment is interupted by the user. Thus, if the `plot` shall be used explicitly after the measurement, one has to pass the first argument of the return value. "
+    "Internally, the `doNd` functions return a tuple consists of the dataset obtained in the measurement, the boolean value correponding to `do_plot` and a callable which checks whether the experiment is interupted by the user. Thus, if the `plot` shall be used explicitly after the measurement, one has to pass the first argument of the return value.\n",
```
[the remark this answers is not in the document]
FarBo: This was also a part of the old notebook and I didn't touch it.

docs/examples/DataSet/Using_the_doNd_functions_for_running_experiments.ipynb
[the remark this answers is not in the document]
FarBo: Sure.

docs/examples/DataSet/Using_the_doNd_functions_for_running_experiments.ipynb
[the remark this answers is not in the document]
jenshnielsen: I know but when you rewrite a notebook you should never leave wrong things their just because they were already there

docs/examples/DataSet/Using_the_doNd_functions_for_running_experiments.ipynb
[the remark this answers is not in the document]
jenshnielsen: I know but you still need to fix it

FarBo:
@jenshnielsen 
@astafan8 

After some back and forth, we decided to improve the already existed doNd notebook with comparing doNds with Measurement instead of creating a new notebook. 
I think Jens already reviewed it. So, you may have a look (hopefully the final one :) ).

astafan8 (approved):
Looks good! and the execution time of the whole notebook is on the order of 1-2 seconds, right?

FarBo:
The execution time may be about 15 seconds for the whole notebook.

docs/examples/DataSet/Using_the_doNd_functions_for_running_experiments.ipynb
```diff
@@ -161,33 +281,136 @@
...
      "text": [
-      "Starting experimental run with id: 16. \n"
+      "Starting experimental run with id: 4. \n"
      ]
     }
    ],
    "source": [
-    "result = do2d(dac.ch1, -1, 1, 20, 0.01, dac.ch2, -1, 1, 20, 0.01, dmm.v1, dmm.v2)"
+    "# Setting up Measurement\n",
+    "meas = Measurement(exp)\n",
```
jenshnielsen, line 327: please remove exp here too

docs/examples/DataSet/Using_the_doNd_functions_for_running_experiments.ipynb
```diff
@@ -91,50 +83,178 @@
...
      "output_type": "stream",
      "text": [
-      "Starting experimental run with id: 14. \n"
+      "Starting experimental run with id: 1. \n"
      ]
     }
    ],
    "source": [
+    "# Setting up Measurement\n",
+    "meas = Measurement(exp)\n",
```
jenshnielsen, line 134: Please remove exp here too

jenshnielsen (approved):

PR #2520, opened 2020-12-11
Updated HP8753D notebook with doNd and its driver timeout  

FarBo:
In this PR:
1. We replaced the legacy Loop method for measurement with the qcodes doNd method for this instrument notebook
2. A picture of the test setup added to the notebook.
3. Updated the function run_N_times in the instrument driver with the timeout context manager, and increased the timeout offset to 2 seconds to avoid timeout error, which was happening with 1 second.


jenshnielsen:
The 2d plot does not look as I would expect. How come the area between -60 and -50 dbm looks different from the rest ?

FarBo:
@jenshnielsen 
It's surprising for me as well. Just note that the old notebook with the loop is returning exactly this result.
It could be the instrument is not behaving correctly or something in the driver.

jenshnielsen:
@FarBo Perhaps its just noise since the power is so low. If you have acess to the instrument perhaps update the example to sweep from -50dbm to -10 dbm or so

FarBo:
@jenshnielsen 
That's a good idea. I will try to test it on Monday and hopefully we can merge it before the next release.

jenshnielsen (approved):

PR #2526, opened 2020-12-14
Add kwarg for experiment id to doNd functions (With tests)

jenshnielsen:
Replaces #2489 with tests and docstrings included


astafan8 (approved):
Great!

I wonder if we should also add the "name" kwarg, so that folks can pass a "name" to Measurement which changes the ``name`` of the DataSet. If yes, could be even this PR.

PR #2540, opened 2020-12-16
filter urllib3.connection related warnings from console logger

jenshnielsen:
opencensus seems to sometime have invalid certificates that will trigger warnings from `urllib3.connections` 

@astafan8 The default Filter only supports filtering by name. Here I have just opted for the simple solution. We could also add a subclass that allows us to filter by message. What do you think?

### DIFF
--- a/qcodes/logger/logger.py
+++ b/qcodes/logger/logger.py
@@ -60,10 +60,16 @@ telemetry_handler: Optional[AzureLogHandler] = None
 
 
 _opencensus_filter = logging.Filter(name="opencensus")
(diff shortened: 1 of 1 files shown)


FarBo:
@jenshnielsen 
This PR should work fine.
Your suggestion for filtering out by message is not the replacement for this filtering by name, right?
If so, that would be also useful for filtering out specific unwanted messages.

astafan8 (approved):
Im happy to agree to the simple solution for now. Let’s see if really need to filter by message from experience/data :)

FarBo (approved):

PR #2560, opened 2020-12-25
Build docs with Sphinx 3.4.1

jenshnielsen:
Sphinx now extract instance attribute docs from the super class so remove duplicated attributes


jenshnielsen:
Local testing seems to indicate that search is broken with 3.4.1. Need to investigate before merging

astafan8 (approved):

jenshnielsen:
Looks like 3.4.2 broke something that was working in 3.4.1 so once https://github.com/QCoDeS/Qcodes/pull/2589 is merged we could revert this back to 3.4.1 or wait for a fix (in 3.4.3). It seems likely this issue will be fixed by https://github.com/QCoDeS/Qcodes/pull/2589

PR #2561, opened 2020-12-25
Add checks for signal current level and voltage level queries in Keysight E4980A driver

Akshita07:
- Checks for signal current level and voltage level queries

To Do:
- [x] Test on actual instrument


qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -386,6 +384,32 @@ def _set_measurement(self,
...
         self.write(f":FUNCtion:IMPedance {measurement_pair.name}")
 
+    def _get_voltage_level(self) -> float:
+        """
+        Gets voltage level if signal is set with voltage level parameter
+        otherwise raises an error.
+        """
+        try:
+            v_level = self.ask(":VOLTage:LEVel?")
+        except Exception as e:
```
jenshnielsen, line 394: Normally it is best not to capture general exceptions like this but only a specific type

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -386,6 +385,32 @@ def _set_measurement(self,
         self._measurement_pair = measurement_pair
         self.write(f":FUNCtion:IMPedance {measurement_pair.name}")
 
+    def _get_voltage_level(self) -> float:
+        """
+        Gets voltage level if signal is set with voltage level parameter
+        otherwise raises an error.
+        """
+        try:
```
jenshnielsen, line 393: Would it make sense to check the cached value of the `current_level` parameter before trying to set the voltage level parameter?

trevormorgan: That seems reasonable, a unit test of this would be good too if this is added.

Akshita07: I though about checking cached value of the `current_level` param but we cannot raise an error in that case because otherwise if say signal is set using current level param and someone tries to set now voltage level he will never be able to do so with error raise in place. Though we can log a warning in this scenario.

jenshnielsen: I am not sure I understand. 

I expected the work flow to be:

```
set current level none zero
do some measurements
set current level back to zero
set voltage level none zero
do some other measurements
```

And then in case the user forgets to do `set current level back to zero` an error will be raised when the user tries to `set voltage level none zero`

Akshita07: We can restrict the instrument to the workflow which you have mentioned with  a check but instrument allows to set the signal using voltage level if set with current level previously.

Akshita07: I will check again with instrument and make the changes.

jenshnielsen: Ok could I make another suggestion?

Add a manual parameter that does not communicate with the instrument that keeps track of which mode we are in.
By default this will be unset. 

When setting the current_level to something none zero this will set the mode to current mode and the other way around if setting voltage level.

Now if you try to get the voltage while in current mode we can raise an error and the other way around. 

If we assume that users are not touching the front panel will this not work?

Akshita07: Yes, this will work perfectly. Thanks for the suggestion Jens :)

jenshnielsen:
@astafan8 do you want to have the honors of doing the final review

astafan8 (approved):
thank you @jenshnielsen :) @Akshita07 it looks great :)

PR #2567, opened 2020-12-28
upgrade jedi and parso

jenshnielsen:
Replaces #2564 and #2514 

These two packages need to be kept in sync


jenshnielsen:
python-language-server and Spyder does not support Jedi 0.18

astafan8 (approved):

PR #2569, opened 2020-12-29
Add tests with minimal requirements

jenshnielsen:
* Add minimum version of all reqirementes in setup.py ( Upgrade a few of them. In general I have upgraded them such that there are python 3.7 wheels available)
* Add test requirements as extra requirements in setup.py
* Use requirements-builder to genereate min-requirements.txt and install from there

Note that this bumps the minimal version of pyvisa to 1.11.0 since there is no released version of pyvisa-sim that currently works with pyvisa 1.10 and our tests


astafan8 (approved):

jenshnielsen:
remove workarounds for pyvisa 1.10

PR #2571, opened 2020-12-30
use ubuntu 20.04 in github jobs

jenshnielsen:
Github actions warns that ubuntu-latest will soon be ubuntu 20.04 so lest make that explicit


jenshnielsen:
Ubuntu 20.04 seems to work fine so we can safely leave this as is

PR #2580, opened 2021-01-03
fix typos: lenght to length

jenshnielsen:
Seems to be a somewhat common typo in our code so lets fix it


astafan8 (approved):
great :) one day there will be a commit hook or a spellingbot or something ;)

PR #2582, opened 2021-01-03
Deprecate unused semi private functions modify_values and modify_many_values and remove unimplemented add_parameter method

jenshnielsen:


astafan8 (approved):
Great!

PR #2612, opened 2021-01-13
Set signal mode and snapshot_exclude attributes of volt and curr level params when needed (Keysight E4980A)

Akshita07:
- Set signal mode at driver initialization
- Set snapshot_exclude attribute of volt and curr level params

@jenshnielsen

### DIFF
--- a/qcodes/instrument_drivers/Keysight/keysight_e4980a.py
+++ b/qcodes/instrument_drivers/Keysight/keysight_e4980a.py
@@ -1,5 +1,6 @@
 from typing import Tuple, Sequence, cast, Any, Union
 from distutils.version import LooseVersion
+from pyvisa.errors import VisaIOError
 
 from qcodes import VisaInstrument, InstrumentChannel
 from qcodes.instrument.parameter import (MultiParameter, ParamRawDataType,
@@ -337,6 +338,7 @@ class KeysightE4980A(VisaInstrument):
             "_correction",
             Correction4980A(self, "correction")
         )
+        self._set_signal_mode_on_driver_initialization()
         self.connect_message()
 
     @property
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -433,6 +435,31 @@ def _get_current_level(self) -> float:
...
+        Sets signal mode on driver initialization
+        """
+        if self._is_signal_mode_voltage_on_driver_initialization():
+            self.signal_mode("Voltage")
+            self.voltage_level.snapshot_exclude = False
+            self.current_level.snapshot_exclude = True
+        else:
+            self.signal_mode("Current")
+            self.voltage_level.snapshot_exclude = True
+            self.current_level.snapshot_exclude = False
```
jenshnielsen, line 461: Don't we need to change which parameters are excluded when we change mode in `_set_voltage_level` and `_set_current_level` too

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -433,6 +435,31 @@ def _get_current_level(self) -> float:
...
+        """
+        Checks if signal is set with voltage_level param at instrument driver
+        initialization
+        """
+        assert self.signal_mode() is None
+        try:
+            self.voltage_level()
+            return True
+        except VisaIOError:
+            return False
```
jenshnielsen, line 448: In this case I guess we must be in current mode. So perhaps we should ensure that the mode is set correctly too. That could be done by either calling `self.voltage_level` or simply setting the value of signal_mode to `current`

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
Akshita07: Yes, we do. Will set these there as well.

jenshnielsen (approved):

PR #2626, opened 2021-01-18
Galil DMC-41x3 Motor Controller (Stepper motors) Driver

Akshita07:
Galil DMC-41x3 motor controller can be used with Janis probe for moving the measurement arm. Currently, the movement of this arm happens manually. Aim for this driver is to automate this process as much as possible.

- Adds driver for stepper motors

To Do:
- [x] Add Galil Instrument Base Class
- [x] Driver for the DMC-4133 model to control movement in 3 axis - A, B and C
- [x] Test on instrument

@jenshnielsen @astafan8


docs/examples/driver_examples/QCoDeS example with Galil DMC4133 Controller.ipynb
```diff
@@ -0,0 +1,895 @@
...
+    "arm.set_arm_kinematics()  # sets default values of arm speed to be 100 micro meters per second,\n",
+    "                          # acceleration and deceleration to be 2048 micro meters per second square"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d73781eb",
+   "metadata": {},
+   "source": [
+    "Manually move the motors and take the needle head to the position where you want to set the origin. This will be the left bottom corner of the chip and run following commands."
```
astafan8, line 194: i think it's probably fine to talk about needles, but for a general audience it would help to know what kind of setting we are talking about here - what is a needed, what chip is in mind, etc etc.

docs/examples/driver_examples/QCoDeS example with Galil DMC4133 Controller.ipynb
```diff
@@ -0,0 +1,895 @@
...
+   "source": [
+    "# QCoDeS example with Galil DMC4133 Controller"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "86f9a772",
+   "metadata": {},
+   "source": [
+    "Purpose of this notebook is to demonstrate how Galil DMC4133 Controller driver along with Arm class can be used for running measurements while controlling the arm head simulteneously. Before begining,\n",
```
astafan8, line 16: @Akshita07 i am missing the description of the "lower level" control capabilities in this notebook, assuming that the arm class is not used. would it be fair to add that description? the reason i think it's very beneficial is that the "lower level" capabilities that the arm class hides could also be very useful. imagine when a user wants to control galil motors with qcodes but the arm class does not match the use-case, hence the arm class can't be used, but all the rest of the implemented features can, right? if so, then let's also present them in the notebook.

docs/examples/driver_examples/QCoDeS example with Galil DMC4133 Controller.ipynb
```diff
@@ -0,0 +1,895 @@
...
+    "exp = load_or_create_experiment(experiment_name='galil_controller_testing',\n",
+    "                                sample_name=\"no sample1\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "407f0f45",
+   "metadata": {},
+   "source": [
+    "Our arm is set up and the measurement is set up. Fun the following block for measurement as we are at the 1st row."
```
astafan8, line 659: typos :) 
```suggestion
    "Our arm is set up and the measurement is set up. Run the following block for measurement as we are at the 1st row."
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        """
+        Initializes and opens the connection to the Galil Motion Controller
+
+        Args:
+            name: name for the instrument
+            address: address of the controller
+        """
+        super().__init__(name=name, **kwargs)
+        self.g = gclib.py()
+        self.address = address
```
astafan8, line 41: should these be private attributes? or is it intended for users to be able to quickly get to gclib or the address?

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        self.open()
+
+    def open(self) -> None:
+        """
+        Open connection to Galil motion controller. This method assumes that
+        the initial mapping of Galil motion controller's hardware's mapping
```
astafan8, line 47: typo? or not?
```suggestion
        the initial mapping of Galil motion controller's hardware's
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        to an IP address is done using GDK and the IP address in burned in.
```
astafan8, line 48: "is"? and "burned in" - is that how they call it? :)
```suggestion
        to an IP address is done using GDK and the IP address is burned in.
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        Initializes and opens the connection to the Galil Motion Controller
+
+        Args:
+            name: name for the instrument
+            address: address of the controller
+        """
+        super().__init__(name=name, **kwargs)
+        self.g = gclib.py()
+        self.address = address
+        self.open()
```
astafan8, line 42: shall we have connect message?:)
```suggestion
        self.open()
        self.connect_message()
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+
+    def close(self) -> None:
+        """
+        Close connection to the instrument
+        """
+        self.g.GClose()
+
+    def motion_complete(self, axes: str) -> None:
+        """
+        Wait for motion to complete for given axes
```
astafan8, line 101: is my understanding correct?
```suggestion
        Wait for motion to complete for given axes, e.g. ``"A"``
        for axis A, ``"ABC"`` for axes A, B, and C
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        self._vector_position_validator = Ints(min_value=-2147483648,
+                                               max_value=2147483647)
+
+        self.add_parameter(
+            "coordinate_system",
+            get_cmd="CA ?",
+            get_parser=self._parse_coordinate_system_active,
+            set_cmd="CA {}",
+            vals=Enum("S", "T"),
+            docstring="sets coordinate system for the motion",
```
astafan8, line 133: shall the docstring also explain the values of S and T?

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        )
+
+        self.add_parameter(
+            "clear_sequence",
+            get_cmd=None,
+            set_cmd="CS {}",
```
astafan8, line 139: should this rather be a bound method, not a parameter?

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        )
+
+        self.add_parameter(
+            "vector_acceleration",
+            get_cmd="VA ?",
+            get_parser=lambda s: int(float(s)),
+            set_cmd="VA {}",
+            vals=Multiples(min_value=1024, max_value=1073740800, divisor=1024),
+            unit="counts/sec2",
+            docstring="sets and gets the defined vector's " "acceleration",
```
astafan8, line 152:
```suggestion
            docstring="sets and gets the defined vector's acceleration",
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        )
+
+        self.add_parameter(
+            "vector_deceleration",
+            get_cmd="VD ?",
+            get_parser=lambda s: int(float(s)),
+            set_cmd="VD {}",
+            vals=Multiples(min_value=1024, max_value=1073740800, divisor=1024),
+            unit="counts/sec2",
+            docstring="sets and gets the defined vector's " "deceleration",
```
astafan8, line 162:
```suggestion
            docstring="sets and gets the defined vector's deceleration",
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+            set_cmd="VS {}",
+            vals=Multiples(min_value=2, max_value=15000000, divisor=2),
+            unit="counts/sec",
+            docstring="sets and gets defined vector's speed",
+        )
+
+    @staticmethod
+    def _parse_coordinate_system_active(val: str) -> str:
+        """
+        parses the the current active coordinate system
```
astafan8, line 178:
```suggestion
        parses the current active coordinate system
```

docs/examples/driver_examples/QCoDeS example with Galil DMC4133 Controller.ipynb
```diff
@@ -0,0 +1,895 @@
...
+    "1. Make sure that you have `gclib` package installed in your environment. If not, then follow the instructions [here](https://www.galil.com/sw/pub/all/doc/gclib/html/python.html) for installation.\n",
+    "2. Make sure that the controller is connected to your PC through an Ethernet cable and the configuration is set as according to these instructions on Windows operating system.\n",
+    "    \n",
+    "        a. Go to Control Panel -> Network and Internet -> Network Connections and select the appropriate network adapter.\n",
+    "        b. Next go to the Properties of that adapter, and then Properties for Internet Protocol Version 4 (TCP/IPv4).\n",
+    "        c. Select \"Use the following IP address\" and add an IP address and Subnet. (If the Galil has an IP address of 10.10.10.100 burned in, you would need a PC IP address of something like 10.10.10.1 with a subnet of 255.255.255.0.)\n",
+    "\n",
+    "Once, the connection to the DMC4133 Controller is established, we can begin with necessary imports and the calibration process. So, let us begin!"
+   ]
+  },
```
astafan8, line 27: and would still be nice to have a few words about the vector mode :) at least that "this is submodule that implements the relevant commands, refer to the manual for more info"

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+    def after_seq_motion(self) -> None:
+        """
+        wait till motion ends
+        """
+        self.write("AM S")
+
+
+class Motor(InstrumentChannel):
+    """
+    Class to control motors independently
```
astafan8, line 223:
```suggestion
    Class to control a single motor (independent of possible other motors)
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        """
+        self.write(f"SP{self._axis}={val}")
+
+    def _set_relative_position(self, val: str) -> None:
+        """
+        sets relative position
+        """
+        self.write(f"PR{self._axis}={val}")
+
+    def off(self) -> None:
```
astafan8, line 371: should there be a corresponding "on"?

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+
+    def servo_here(self) -> None:
+        """
+        servo at the motor
+        """
+        self.write(f"SH {self._axis}")
+
+    def begin(self) -> None:
+        """
+        begins motion of the motor and waits until motor stops moving
```
astafan8, line 395: right?
```suggestion
        begins motion of the motor (returns immediately)
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        """
+        self.write(f"BG {self._axis}")
+
+    def wait_till_motor_motion_complete(self) -> None:
+        """
+        this method on call waits for motion to complete on the motor
```
astafan8, line 401:
```suggestion
        wait for motion on the motor to complete
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        """
+        try:
+            while int(float(self.ask(f"MG _BG{self._axis}"))):
```
astafan8, line 404: let's extract it into a method, e.g. is_in_motion or any better name. and also use that in the wait method of the controller class. this exposes more motor API, and thus capabilities, to users.
```suggestion
            while self.is_in_motion():
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+                pass
+        except KeyboardInterrupt:
+            self.root_instrument.abort()
+            self.off()
+
+    def error_magnitude(self) -> float:
+        """
+        gives the magnitude of error, in drive step counts, for axes
```
astafan8, line 412:
```suggestion
        gives the magnitude of error, in drive step counts, for this motor
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+
+        a step count is directly proportional to the micro-stepping
```
astafan8, line 414: do we need an additional conveience method that return this value in more useful units? perhaps, in quadrature counts? or is that not useful/relevant/possible?

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+            "command",
+        )
+
+        self._set_default_update_time()
+        self.add_submodule("motor_a", Motor(self, "A"))
+        self.add_submodule("motor_b", Motor(self, "B"))
+        self.add_submodule("motor_c", Motor(self, "C"))
+        self.add_submodule("plane_ab", VectorMode(self, "AB"))
+        self.add_submodule("plane_bc", VectorMode(self, "BC"))
+        self.add_submodule("plane_ac", VectorMode(self, "AC"))
```
astafan8, line 470: this assumes that 3 motors are connected to the controller, right? Is there a way we can ask controller what motors does it have? that would be best. otherwise, perhaps we need an argument to the constructor to let the user specify what motors are available (as a list, e.g. `['a','b','c']`) and then generate the submodules only for those.

that would also affect the methods like `_get_absolute_position`.

additionally, perhaps it makes sense to make a channel list out of motors.

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+            ):
+                pass
+        except KeyboardInterrupt:
+            self.abort()
+            self.motors_off()
+
+
+class Arm:
+    """
+    Module to control probe arm
```
astafan8, line 554: this docstring would be much more helpful if it wouls include information on what kind of arm/chip is assumed here.

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        self.right_top_position: Tuple[int, int, int]
+
+        # motion directions
+        self._a: np.ndarray  # right_top - left_bottom
+        self._b: np.ndarray  # left_top - left_bottom
+        self._c: np.ndarray  # right_top - left_top
+        self._n: np.ndarray
+        self.norm_a: float
+        self.norm_b: float
+        self.norm_c: float
```
astafan8, line 578: should the norms be public attributes?

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+
+        # eqn of the chip plane
+        self._plane_eqn: np.ndarray
+
+        # current vars
+        self.current_row: Optional[int] = None
+        self.current_pad: Optional[int] = None
```
astafan8, line 585: should these be read-only properties as opposed to public attributes?

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        self.pads: int
+        self.inter_row_dis: float
+        self.inter_pad_dis: float
+
+        # arm kinematics
+        self.speed: int
+        self.acceleration: int
+        self.deceleration: int
+
+        self._arm_pick_up_dis: int
```
astafan8, line 598: right?
```suggestion
        self._arm_pick_up_distance: int
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+
+        self._target: np.ndarray
+
+    def set_arm_kinematics(
+        self,
+        speed: int = 100,
+        acceleration: int = 2048,
+        deceleration: int = 2048
+    ) -> None:
```
astafan8, line 607: we should mention the units, perhaps in a docstring?

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+
+        if acceleration % 256 != 0 or deceleration % 256 != 0:
```
astafan8, line 609: is this condition in the units of micro meters or quandrature counts?

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        self.current_row: Optional[int] = None
+        self.current_pad: Optional[int] = None
+
+        # chip details
+        self.rows: int
+        self.pads: int
+        self.inter_row_dis: float
+        self.inter_pad_dis: float
+
+        # arm kinematics
```
astafan8, line 593:
```suggestion
        # arm kinematics in the units of quandrature counts
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+        self.speed: int
+        self.acceleration: int
+        self.deceleration: int
```
astafan8, line 596: what is the intention here - should these attrbiutes be public and thus easily changable or should the user be only able to set them via the set arm kinematics method?

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+                "Acceleration and deceleration must be a multiple of 256."
+            )
+
+        self.speed = _convert_micro_meter_to_quadrature_counts(speed)
+        self.acceleration = \
+            _convert_micro_meter_to_quadrature_counts(acceleration)
+        self.deceleration = \
+            _convert_micro_meter_to_quadrature_counts(deceleration)
+
+    def set_pick_up_distance(self, distance: float = 3000) -> None:
```
astafan8, line 620: please note the units for all the methods on this class. perhaps in docstrings

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+
+        Args:
+            controller: an instance of DMC4133Controller
+        """
+        self.controller = controller
+
+        # initialization
+        self.left_bottom_position: Tuple[int, int, int]
+        self.left_top_position: Tuple[int, int, int]
+        self.right_top_position: Tuple[int, int, int]
```
astafan8, line 569: should these be user-writable or not?

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+
+        a = np.array(
+            tuple(
+                map(
+                    lambda i, j: i - j,
+                    self.right_top_position,
+                    self.left_bottom_position,
+                )
+            )
+        )
```
astafan8, line 652: shall we let numpy do the vectorized algebra for us ? :) or am i missing something here?
```suggestion
        a = np.asarray(self.right_top_position) - np.asarray(self.left_bottom_position)
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+
+        pos = self.controller.absolute_position()
+        self.left_top_position = (pos["A"], pos["B"], pos["C"])
+
+    def set_right_top_position(self) -> None:
+
+        pos = self.controller.absolute_position()
+        self.right_top_position = (pos["A"], pos["B"], pos["C"])
+
+        self._calculate_ortho_vector()
```
astafan8, line 640: i'd add this call to the other to set_*_position methods to relieve the user from knowing the right order. This can be achieved by making the _calculate_ortho_vector method "do nothing" unless all 3 `*_position` attributes are set

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+
+        if np.dot(self._plane_eqn, self._target) < 0:
+            raise RuntimeError(
+                f"Cannot move to {self._target[:3]}. Target location is below "
+                f"chip plane."
+            )
+
+        sp_a = int(np.floor(abs(rel_vec[0]) * speed))
+        if sp_a % 2 != 0:
+            sp_a += 1
```
astafan8, line 754: is it better to have higher or lower speeds? :) i guess "1" is such a small difference that either is fine. and the speed limit is so high that noone would even want to use that, right?
```suggestion
            sp_a -= 1
```

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,964 @@
...
+
+        self._pick_up()
+
+        motion_vec = -1 * self._a
+        self._setup_motion(rel_vec=motion_vec, d=self.norm_a, speed=self.speed)
+        self._move()
+        self.current_row = 1
+        self.current_pad = 1
+
+        self._put_down()
```
astafan8, line 828: it's not critical but (in all methods that change row/pad) i'd recommend to set row/pad last so that it's ensured that the motions were successfully executed. otherwise if put down fails, current row and pad are also changed which is pre-mature :)
```suggestion
        self._put_down()

        self.current_row = 1
        self.current_pad = 1
```

docs/examples/driver_examples/QCoDeS example with Galil DMC4133 Controller.ipynb
```diff
@@ -0,0 +1,895 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "9c1b17cc",
+   "metadata": {},
+   "source": [
+    "From now on all the motor movements will be controlled by the driver commands.\n",
+    "\n",
+    "Next step is to set reverse limits for for all three motors. First take the motor C to the extreme reverse position which you want to set as the reverse limit with following command."
```
astafan8, line 234: what about forward limits?

docs/examples/driver_examples/QCoDeS example with Galil DMC4133 Controller.ipynb
Akshita07: ok, will add this information

docs/examples/driver_examples/QCoDeS example with Galil DMC4133 Controller.ipynb
Akshita07: Yes, that is correct. All the low level features can be used without arm class also. Ok, I will add a section for it in the beginning of the notebook.

docs/examples/driver_examples/QCoDeS example with Galil DMC4133 Controller.ipynb
Akshita07: :)

qcodes/instrument_drivers/Galil/dmc_41x3.py
Akshita07: Ah yes, address should be a private attribute but gclib attribute is for users to directly access other functionalities of the python wrapper which are not provided in our driver.

qcodes/instrument_drivers/Galil/dmc_41x3.py
Akshita07: yes

qcodes/instrument_drivers/Galil/dmc_41x3.py
Akshita07: yes :)

qcodes/instrument_drivers/Galil/dmc_41x3.py
Akshita07: yes

qcodes/instrument_drivers/Galil/dmc_41x3.py
Akshita07: ok

docs/examples/driver_examples/QCoDeS example with Galil DMC4133 Controller.ipynb
Akshita07: ok

qcodes/instrument_drivers/Galil/dmc_41x3.py
Akshita07: servo_here is actually the "on" command. I have used the same terminology as in the manual for it. :)

qcodes/instrument_drivers/Galil/dmc_41x3.py
Akshita07: yes :)

qcodes/instrument_drivers/Galil/dmc_41x3.py
Akshita07: fixed

qcodes/instrument_drivers/Galil/dmc_41x3.py
astafan8: ah, ok, good :)

qcodes/instrument_drivers/Galil/dmc_41x3.py
astafan8: ah, i see, so the model of the controller defines the number of motors attached, i see.

>  we can read the IDN and get the number of motors connected from the message as if 4 motors are connected the message would read that it is a DMC4143 controller

i like this idea! but if it's too much of an extra effort, then let's not do it in this PR.

Akshita07: Ok, then will leave it for another PR and keep the current implementation

docs/examples/driver_examples/QCoDeS example with Galil DMC4133 Controller.ipynb
astafan8: i see, makes then. then i'd explain this in the notebook (e.g. add this exact comment to the notebook). and even despite the lack of sense users still want to be extra careful and set the forward limits, they still can, right?

Akshita07: Ok, I will add this info to the notebook and yes, forward limits can be set.

Akshita07: Updated as discussed

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,1033 @@
...
+            self.abort()
+            self.motors_off()
+
+
+class Arm:
+    """
+    Module to control probe arm. It is assumed that the chip to be probed has
+    one or more rows with each row having one or more pads. Design of the
+    chip to be probed should resemble the following:
+            |--------------------------|
```
astafan8, line 568: for this to render nicely in the docs perhaps the ``::`` should be used https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#rst-literal-blocks

qcodes/instrument_drivers/Galil/dmc_41x3.py
```diff
@@ -0,0 +1,1033 @@
...
+                f"Pad num: {num} is out of range. Pad number start from 1 "
+                f"and max number is {self.pads}."
+            )
+
+        sign = 1
+        assert self._current_pad is not None
+        d = (num - self._current_pad) * self.inter_pad_distance
+
+        if d == 0:
+            raise RuntimeError(
```
astafan8, line 983: it would be good to ask users whether they like a runtime exception here or just let the method return if the current pad is already the desired one. My hypothesis is that an exception is not desirable, but i might be wrong :)

Akshita07: ok, will check this :)

qcodes/instrument_drivers/Galil/dmc_41x3.py
Akshita07: updated

qcodes/instrument_drivers/Galil/dmc_41x3.py
Akshita07: I checked this, and preference is for letting the users know that are in fact they are on the row that they want to move to. So, current implementation can stay.

astafan8: perfect :)

trevormorgan (approved):

PR #2647, opened 2021-01-25
Add support for Mypy 0.800

jenshnielsen:
Replaces https://github.com/QCoDeS/Qcodes/pull/2645

* Mypy now requires that you use the same name for a named tuple variable and its name. Apparently that has 
always been recommended. 
* Handle that `msvcrt` module is now correctly detected on Windows but missing on other platforms


astafan8 (approved):
awesome!!!!

PR #2648, opened 2021-01-25
Add bias volt param to Basel SP983C

Akshita07:
- Adds bias volt param 

This addition of delegate parameter will help in ability to bias device under test via Basel SP983c preamp.

@jenshnielsen @astafan8 @trevormorgan


qcodes/instrument_drivers/basel/sp983c.py
```diff
@@ -38,6 +39,12 @@ def __init__(self, name: str, **kwargs: Any):
                            vals=Enum(30., 100., 300., 1e3, 3e3, 10e3, 30e3,
                                      100e3, 1e6))
 
+        self.add_parameter('bias_volt',
+                           initial_value=0,
+                           unit='mV',
```
trevormorgan, line 44: The source instruments to supply the bias voltage will be with units volts so think we should change this as unit V and change the vals to match.

qcodes/instrument_drivers/basel/sp983c.py
```diff
@@ -38,6 +39,13 @@ def __init__(self, name: str, **kwargs: Any):
                            vals=Enum(30., 100., 300., 1e3, 3e3, 10e3, 30e3,
                                      100e3, 1e6))
 
+        self.add_parameter('bias_volt',
```
astafan8, line 42: does this name correctly represent the "abstraction", as in, does the basel preamp know that this parameter is used for biasing, as opposed to something else? in other words, is the name of this parameter generic enough?

astafan8, line 42: shall we avoid shortenings for readability sake ? :)
```suggestion
        self.add_parameter('bias_voltage',
```

qcodes/instrument_drivers/basel/sp983c.py
```diff
@@ -38,6 +39,13 @@ def __init__(self, name: str, **kwargs: Any):
...
+                           initial_value=0,
+                           label="Bias Voltage",
+                           unit='V',
+                           vals=Numbers(-0.1, 0.1),
```
astafan8, line 46: these limits, are they dictated by the pre-amp?

qcodes/instrument_drivers/basel/sp983c.py
Akshita07: instrument uses the word 'offset' instead of bias.. should it be changes to that?

qcodes/instrument_drivers/basel/sp983c.py
Akshita07: yes

qcodes/instrument_drivers/basel/sp983c.py
astafan8: yes, i'd say it's better to reflect the term that the instrument uses because the cases on the user side may different and likely not for all of them the word "bias" would make sense.

qcodes/instrument_drivers/basel/sp983c.py
```diff
@@ -1,7 +1,8 @@
 from typing import Any, Optional, Dict
 
+from qcodes import Parameter, DelegateParameter
 from qcodes.instrument.base import Instrument
-from qcodes.utils.validators import Enum
+from qcodes.utils.validators import Enum, Numbers
 
 
 class SP983C(Instrument):
```
trevormorgan, line 8: Before you merge can you update the doc string with information about the input_offset_voltage parameter? Thanks!

trevormorgan (approved):

qcodes/instrument_drivers/basel/sp983c.py
jenshnielsen: It would probably also be nice to add an example notebook that shows this

qcodes/instrument_drivers/basel/sp983c.py
```diff
@@ -13,12 +13,20 @@ class SP983C(Instrument):
...
     for the the remote control interface (SP 983a). It is the responsibility of
     the user to ensure that values set here are in accordance with the values
     set on the instrument.
+
+    Args:
+        name
+        input_offset_voltage: (Optional) A source input offset voltage
+            parameter. The range for input is -10 to 10 Volts and it is
+            user's responsibility to ensure this. This source parameter is
+            used to set offset voltage parameter of the preamp.
```
jenshnielsen, line 29: I would add something like "The parameter should represent an voltage source that is connected to the "Offset Input volgate" connector of the SP983C

docs/examples/driver_examples/Qcodes example with Basel SP983c Preamp.ipynb
```diff
@@ -0,0 +1,314 @@
...
+   "source": [
+    "# Qcodes example with Basel SP983c Preamp"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "assigned-cartridge",
+   "metadata": {},
+   "source": [
+    "This notebook shows how to setup an input offset voltage source for the Basel SP983c Preamp. The range of input offset voltage is -10 to 10 Volts. This input offset voltage is used for offsetting the voltage by the preamp in range -0.1 to 1 Volts. Let's begin!"
```
jenshnielsen, line 16: I think the notebook should be extended to not only explain this but how the preamp works in general.

qcodes/instrument_drivers/basel/sp983c.py
jenshnielsen: Are we sure this is True for all models. E.g. here there are listed 4 models https://www.baspi.ch/low-noise-high-stab-itov-conv

2 of them have a +-0.1 v offset and 2 of them have a +- 1 v offset

docs/examples/driver_examples/Qcodes example with Basel SP983c Preamp.ipynb
Akshita07: Ok, will update it.

astafan8 (approved):

qcodes/instrument_drivers/basel/sp983c.py
Akshita07: Yes, the current implementation uses ranges for two model of SP983c. We are not sure how the scaling happens for SP983c01 models and hence not implemented.

PR #2649, opened 2021-01-25
Keysight B1500: support measuring any number of channels during IV sweep

astafan8:
Before this, Keysight B1500 driver was able to perform an IV sweep strictly for two channels, that was the most common case. This PR extends the support to as many channels as the measurement mode and the instrument dictates. For example, the instrument supports measuring up to 10 channels in the sweep mode (during an IV sweep).

- [ ] add docstrings
- [x] update the IV sweep section of the example notebook


qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_base.py
```diff
@@ -465,14 +465,95 @@ def __init__(self, name: str, instrument: B1517A, **kwargs: Any):
...
         measurement_mode = self.instrument.get_measurement_mode()
-        if len(measurement_mode['channels']) != 2:
-            raise ValueError('Two measurement channels are needed, one for '
-                             'gate current and other for source drain '
-                             'current.')
+        channels = measurement_mode['channels']
 
-        smu = self.instrument.by_channel[measurement_mode['channels'][0]]
+        if names is None:
+            names = [f"param{n+1}" for n in range(len(channels))]
```
jenshnielsen, line 477: should this also be param{n+1}_current to match the label?

astafan8: that would be backwards incompatible :(

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_base.py
```diff
@@ -499,22 +580,30 @@ def get_raw(self) -> Tuple[Tuple[float, ...], Tuple[float, ...]]:
...
+        parsed_data = fmt_response_base_parser(raw_data)
 
-        self.shapes = ((len(self.source_voltage.value),),) * 2
-        self.setpoints = ((self.source_voltage.value,),) * 2
+        n_items_per_data_point = 4  # == len(_FMTResponse(None, None, None, None))
+        n_all_data_channels = n_channels + 1  # because sourced voltage values are also returned
+
+        for n in range(n_channels):
+            single_channel_data = _FMTResponse(
+                *[parsed_data[i][n::n_all_data_channels] for i in range(0, n_items_per_data_point)])
```
jenshnielsen, line 594: If breaking just after the ( I would also put ) on a new line to make it more clear where the call to _FMTResponse ends. Alternatively I would extract the list generator into a variable

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_base.py
```diff
@@ -499,22 +580,30 @@ def get_raw(self) -> Tuple[Tuple[float, ...], Tuple[float, ...]]:
...
+            convert_dummy_val_to_nan(single_channel_data)
+            setattr(self, f"param{n+1}", single_channel_data)
+
+        channel_values_to_return = tuple(
+            getattr(self, f"param{n + 1}").value
+            for n in range(n_channels)
+        )
+
+        self.source_voltage = _FMTResponse(
+            *[parsed_data[i][n_channels::n_all_data_channels] for i in range(0, n_items_per_data_point)])
```
jenshnielsen, line 604: Same as above

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_base.py
astafan8: done

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_base.py
astafan8: done

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_base.py
```diff
@@ -465,14 +465,95 @@ def __init__(self, name: str, instrument: B1517A, **kwargs: Any):
...
+        if names is None:
+            names = [f"param{n+1}" for n in range(len(channels))]
+            if labels is None:
+                labels = [f"Param{n + 1} Current" for n in range(len(channels))]
+
+        if labels is None:
+            labels = tuple(names)
+
+        if units is None:
+            units = ['A'] * len(names)
```
trevormorgan, line 485: Does this mean if no channel parameters are specified then the default channel is in current mode?

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_base.py
```diff
@@ -465,14 +465,95 @@ def __init__(self, name: str, instrument: B1517A, **kwargs: Any):
...
+    def set_setpoint_name_label_and_unit(
+            self,
+            name: Optional[str] = None,
+            label: Optional[str] = None,
+            unit: Optional[str] = None
+    ) -> None:
+        # number of measured parameters of this MultiParameter
+        n_names = len(self.names)
+
+        name = name if name is not None else "Voltage"
```
trevormorgan, line 522: Similarly here if the setpoint specifics aren't specified then the default mode is for souring voltage right? That seems fine to me but maybe add a quick note about  that in the doc string.

trevormorgan (approved):

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_base.py
astafan8: yes, that's actually the only way this IV sweep works, it only sweeps voltage. it's left specifyable in case someone wants to be more specific for some unknown reason :)

> add a quick note about that in the doc string

will definitely do!

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_base.py
astafan8: yes, that's actually the only way this IV sweep works, it only measures currents. it's left specifyable in case someone wants to be more specific for some unknown reason :)

PR #2654, opened 2021-01-26
Refactor data merging when loading from disk

jenshnielsen:
This will eventually enable the addend logic to be used with other sources of data


astafan8 (approved):
Looks indeed cleaner with this separation :)

and i do agree to NOT preserve the original signature of append_shaped_parameter_data_to_existing_arrays for backwards compatibility.

PR #2665, opened 2021-01-28
Separate conda from pip in CI

WilliamHPNielsen:
This PR will eventually separate conda from pip in our CI and in doing so render the Azure pipelines CI obsolete.

- [x] Verify in CI that qcodes can be installed with (almost) all dependencies from conda forge
- [x] Verify in CI that such an installation works, i.e. passes the test suite
- [x] Do the above on windows and ubuntu
- [x] Remove similar tests from azure pipelines flow

For another PR, the azure pipelines can be retired

@jenshnielsen, thoughts? Additional check items you'd like to see?


environment_forge_full.yml
```diff
@@ -0,0 +1,41 @@
...
+ - wrapt
+ - lxml
+ - scipy
+ - gitpython
+ - pandas
+ - testpath>=0.4.4
+ - tqdm
+ - tabulate
+ - pandoc
+ - contextlib2 # for broadbean
```
jenshnielsen, line 32: That comment (and probably the whole line can be removed)

WilliamHPNielsen: Good point.

jenshnielsen:
Looks good. To completely replace azure pipelines we should in addition do the following:

- [ ] Run mypy on windows
- [ ] Build docs on windows
- [ ] Push docs from github actions to gh-pages 
- [ ] Push to pypi from github pages

environment_forge_full.yml
```diff
@@ -0,0 +1,40 @@
+name: qcodesforge
+channels:
+ - conda-forge
+dependencies:
+ - python=3.7.9
+ - pip
+ - deepdiff
+ - numpy=1.19.5
```
jenshnielsen, line 8: Since there is no way for dependabot to automatically update the conda deps I think we should leave all of them unpinned except python. I would probably also move this file to within the .github dir to make it clear that it is not intended for end users.

.github/workflows/conda-forge-install-and-test.yaml
```diff
@@ -0,0 +1,42 @@
...
+        auto-activate-base: false
+    - name: install qcodes
+      shell: bash -l {0}
+      run: |
+        conda info
+        pip install . --no-deps
+    - name: run tests
+      shell: bash -l {0}
+      run: |
+        pytest -n 1
```
jenshnielsen, line 42: i would try with `pytest -n 0 --dist no` which runs all tests in the main process. `-n 1` spawns exactly one process that runs all tests

jenshnielsen:
@WilliamHPNielsen this now reveals that the issue is due to legacy/test_plots segfaulting. These tests use qt and do therefor require a functional xserver (which the other jobs set up)

.github/workflows/conda-forge-install-and-test.yaml
```diff
@@ -0,0 +1,48 @@
...
+      if: runner.os == 'Linux'
+    - name: install dependencies
+      uses: conda-incubator/setup-miniconda@v2
+      with:
+        activate-environment: qcodesforge
+        environment-file: .github/static/environment_forge_full.yml
+        auto-activate-base: false
+    - name: generate test dbs
+      uses: ./.github/actions/generate-test-dbs
+      # todo is it an issue that this pip installs qcodes with deps?
```
jenshnielsen, line 39: I think we should pull the pip install logic out of that script and simply require that the user has installed qcodes editable before running this.

astafan8: or we create a job that creates the DBs, puts then as some artifact and passes that artifact to this job :)

jenshnielsen: @astafan8 Indeed that would be better but that is out of scope for this pr

jenshnielsen: Furthermore it is not obvious that you can share data between jobs that are defined in different workflow files https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts#passing-data-between-jobs-in-a-workflow

jenshnielsen:
@astafan8 @WilliamHPNielsen @trevormorgan I think this is ready for review. I will move final things (the docs build on windows, the push to gh pages and the push to pypi) in another pr.

.github/actions/generate-test-dbs/action.yml
```diff
@@ -5,8 +5,6 @@ runs:
   steps:
     - name: generate test dbs
       run: |
```
astafan8, line 7: shall we add a comment in this action that it's expected that qcodes is installed in editable mode before this action can be run?

astafan8 (approved):
Looks good! looking forward to the subsequent PR that removes the need for azure pipelines alltogether ;)

.github/actions/generate-test-dbs/action.yml
jenshnielsen: Nah, the action will fail since there is check inside the generate scripts that raises if this is not the case

PR #2667, opened 2021-01-28
Commit data directy to dataset cache

jenshnielsen:


qcodes/dataset/data_set.py
```diff
@@ -292,7 +292,8 @@ def __init__(self, path_to_db: Optional[str] = None,
                  specs: Optional[SpecsOrInterDeps] = None,
                  values: Optional[VALUES] = None,
                  metadata: Optional[Mapping[str, Any]] = None,
-                 shapes: Optional[Shapes] = None) -> None:
+                 shapes: Optional[Shapes] = None,
+                 in_memory_cache: bool = True) -> None:
```
astafan8, line 296: it's missing its documentation in the docstring

qcodes/dataset/data_set.py
```diff
@@ -1491,8 +1501,35 @@ def _enqueue_results(
...
+        if self._in_memory_cache:
+            self.cache.add_data(new_results)
+
+    @staticmethod
+    def _reshape_array_for_cache(
+            param: ParamSpecBase,
+            param_data: numpy.ndarray
+    ) -> numpy.ndarray:
+        """
+        Shape cache data so it matches data read from dd.
```
astafan8, line 1519:
```suggestion
        Shape cache data so it matches data read from database.
```

qcodes/dataset/data_set.py
```diff
@@ -1491,8 +1501,35 @@ def _enqueue_results(
...
+        Add an extra singleton dim to array data and flatten
+        non array data into a linear array.
```
astafan8, line 1521:
```suggestion
        This means:
        
        - Add an extra singleton dim to array data
        - flatten non array data into a linear array.
```

qcodes/dataset/data_set.py
```diff
@@ -1491,8 +1501,35 @@ def _enqueue_results(
 
         if standalones:
             stdln_dict = {st: result_dict[st] for st in standalones}
+            for st in standalones:
+                new_results[st.name] = {
+                    st.name: self._reshape_array_for_cache(st, result_dict[st])
+                }
             self._results += self._finalize_res_dict_standalones(stdln_dict)
 
+        if self._in_memory_cache:
```
astafan8, line 1510: shall the calls to _reshape_array_for_cache also be done with `if self._in_memory_cache` guard? wouldn't look beautiful, but it is what it is, i guess.

qcodes/dataset/data_set.py
```diff
@@ -1807,6 +1846,8 @@ def new_data_set(name: str,
         specs: list of parameters to create this dataset with
         values: the values to associate with the parameters
         metadata: the metadata to associate with the dataset
+        in_memory_cache: Should measured data be keep in memory
+            and available as part of the `dataset.cache` object.
```
astafan8, line 1850:
```suggestion
        in_memory_cache: Should measured data be kept in memory
            and available as part of the `dataset.cache` object, as opposed
            to being persisted
```

qcodes/dataset/data_set_cache.py
```diff
@@ -49,13 +50,20 @@ def load_data_from_db(self) -> None:
         If the dataset is marked completed and data has already been loaded
         no load will be performed.
         """
+        if self.live:
+            raise RuntimeError("Cannot read data into this cache from the "
+                               "database as the dataset is build "
+                               "directly from measured data.")
```
astafan8, line 56: would this be a clearer message? or do we not want to mention "in memory" at all?
```suggestion
            raise RuntimeError("Cannot load data into this cache from the "
                               "database because this dataset is being built "
                               "in-memory.")
```

qcodes/dataset/data_set_cache.py
```diff
@@ -66,6 +74,16 @@ def load_data_from_db(self) -> None:
...
             self._data
         )
+        if old_write_status != self._write_status:
+            self._live = False
+
+    @property
+    def live(self) -> Optional[bool]:
+        """
+        If true this cache has been produced by appending data as measured.
+        If false the data has been read from disk.
```
astafan8, line 84:
```suggestion
        If false the data has been read from disk.
        If None, then the cache does not yet have any data.
```

qcodes/dataset/data_set_cache.py
```diff
@@ -66,6 +74,16 @@ def load_data_from_db(self) -> None:
             self._read_status,
             self._data
         )
+        if old_write_status != self._write_status:
+            self._live = False
+
+    @property
+    def live(self) -> Optional[bool]:
```
astafan8, line 81: it's very interesting that with this implementation the cache can be loaded from the database and THEN extra data can be added to it (and once that is done, reading from database is not possible anymore regardless of whether there is new data there or not). What would we use this for? If we are not going to use it, i wonder if we should restrict the cache to always be either live or not-live, as opposed to give him a change to become live after being not live :)

qcodes/dataset/data_set_cache.py
```diff
@@ -252,7 +294,20 @@ def _insert_into_data_dict(
...
     if shape is None or write_status is None:
-        return np.append(existing_values, new_values, axis=0), None
+        try:
+            data = np.append(existing_values, new_values, axis=0)
+        except ValueError:
+            # we cannot append into a ragged array so make that manually
+            n_existing = existing_values.shape[0]
+            n_new = new_values.shape[0]
+            n_rows = n_existing + n_new
+            data = np.ndarray((n_rows,), dtype=object)
```
astafan8, line 304: why is the type of the existing values not re-used? or will numpy do it automatically behind the scenes?
```suggestion
            data = np.ndarray((n_rows,), dtype=existing_values.dtype)
```

qcodes/dataset/descriptions/detect_shapes.py
```diff
@@ -55,7 +55,10 @@ def detect_shape_of_measurement(
     shapes: Dict[str, Tuple[int, ...]] = {}
 
     for param_name in array_shapes.keys():
-        shapes[param_name] = tuple(loop_shape) + array_shapes[param_name]
+        total_shape = tuple(loop_shape) + array_shapes[param_name]
+        if total_shape == ():
+            total_shape = (1,)
```
astafan8, line 60: so this means we had a bug before that scalar parameters would get `()` as opposed to `(1,)` - do i understand this correctly? if so, i wonder if the line `array_shapes[param.full_name] = ()` above is correct and should be changed to `... = (1,)`? or i'm totally misunderstanding the why here.

qcodes/dataset/measurements.py
```diff
@@ -499,10 +501,13 @@ def __enter__(self) -> DataSaver:
         # next set up the "datasaver"
         if self.experiment is not None:
             self.ds = qc.new_data_set(
-                self.name, self.experiment.exp_id, conn=self.experiment.conn
+                self.name, self.experiment.exp_id, conn=self.experiment.conn,
```
astafan8, line 504: missed the cache flag here
```suggestion
                self.name, self.experiment.exp_id,
                conn=self.experiment.conn,
                in_memory_cache=self._in_memory_cache
```

qcodes/dataset/measurements.py
```diff
@@ -1113,6 +1119,8 @@ def run(self, write_in_background: Optional[bool] = None) -> Runner:
                 main thread that is executing the context manager.
                 By default the setting for write in background will be
                 read from the ``qcodesrc.json`` config file.
+            in_memory_cache: Should measured data be keep in memory
+                and available as part of the `dataset.cache` object.
```
astafan8, line 1123:
```suggestion
            in_memory_cache: Should measured data be kept in memory
                and available as part of the `dataset.cache` object, as opposed
                to be persisted.
```

qcodes/tests/dataset/test_data_set_cache.py
```diff
@@ -122,19 +128,29 @@ def test_cache_1d(experiment, DAC, DMM, n_points, bg_writing,
                                                     shaped_partial=set_shape)
     _assert_parameter_data_is_identical(dataset1.get_parameter_data(),
                                         dataset1.cache.data())
-    assert dataset1.cache._loaded_from_completed_ds is True
+    if in_memory_cache is False:
+        assert dataset1.cache._loaded_from_completed_ds is True
+    else:
+        # todo enable live cache to set this
```
astafan8, line 134: this comment can go, right?

qcodes/tests/dataset/test_data_set_cache.py
```diff
@@ -122,19 +128,29 @@ def test_cache_1d(experiment, DAC, DMM, n_points, bg_writing,
...
+        assert dataset1.cache._loaded_from_completed_ds is False
```
astafan8, line 135: do i understand it correctly?
```suggestion
        assert dataset1.cache._loaded_from_completed_ds is False
    assert dataset1.cache.live is in_memory_cache
```

qcodes/dataset/data_set.py
jenshnielsen: No not as opposed in addition

qcodes/dataset/data_set_cache.py
jenshnielsen: Because this is a ragged array as written in the comment.

qcodes/dataset/measurements.py
jenshnielsen: Again no

qcodes/dataset/data_set_cache.py
astafan8: ah, indeed.

qcodes/dataset/data_set.py
astafan8: aaaaaaaaa........ so stupid of me - so in this PR the cache is maintained in memory and we don't yet "stop writing to the DB", right?

qcodes/dataset/data_set_cache.py
jenshnielsen: I think you have misunderstood the logic. 

It can be loaded from the db and if that happens live is set to False and you can continue loading from the db until all data has been loaded from the dataset as signaled by the dataset being marked completed. 

Alternatively you can start building the cache from live data in which case it is an error to try to load from the db.

qcodes/dataset/descriptions/detect_shapes.py
jenshnielsen: I don't think so. This is really only cover the case where you measure a 0d experiment with a scalar parameter so that both `loop_shape` and `array_shapes[param_name]` are empty. If you set array_shapes to (1,) you would expand 1d sweeps of scalar parameters from (n,) to (n,1) which I don't think we want

qcodes/dataset/data_set.py
jenshnielsen: I thought about that and my initial assertion was that the ugly code was worse than the performance hit suffered from doing this but happy to change my mind

qcodes/tests/dataset/test_data_set_cache.py
jenshnielsen: I would like to eventually find a way for the live cache to mark the dataset as completed

qcodes/dataset/data_set_cache.py
jenshnielsen: to clarify this we should probably raise in live is false when calling `add_data`

astafan8: thanks for the explanation! perhaps it's worth adding it to the docstring of the cache?

> raise in live is false when calling add_data
yes :)

qcodes/dataset/data_set.py
astafan8: perfecet!

qcodes/dataset/descriptions/detect_shapes.py
astafan8: ah, 0d, of course.

qcodes/dataset/data_set.py
astafan8: i think that the current ugliness of the code can't get significantly worse ;) however, not calling _reshape_array_for_cache  if _in_memory_cache is False makes the ugly code clearer. otherwise, the way it is now, it is ugly and confusing, as opposed to just ugly. would you agree to this reasoning?

qcodes/tests/dataset/test_data_set_cache.py
astafan8: oh, hmmm. why would we want that? at the first sight seems to me like increase of coupling between dataset and its cache and i don't see what this brings. could you share more?

qcodes/dataset/data_set.py
jenshnielsen: Not sure that I agree but lets try it

qcodes/tests/dataset/test_data_set_cache.py
jenshnielsen: Mostly I would like that the dataset could be marked completed even if not written to disk

qcodes/dataset/data_set_cache.py
jenshnielsen: both are done

jenshnielsen:
@astafan8 I think all the issues should be addressed

astafan8 (approved):

PR #2674, opened 2021-01-29
Remove left over usage of deprecated  get_data_as_pandas_dataframe

jenshnielsen:
@trevormorgan


astafan8 (approved):

qcodes/dataset/data_set.py
```diff
@@ -1335,7 +1335,7 @@ def write_data_to_text_file(self, path: str,
                                in a single file but no filename provided.
         """
         import pandas as pd
-        dfdict = self.get_data_as_pandas_dataframe()
+        dfdict = self.to_pandas_dataframe_dict()
```
astafan8, line 1338: for the sake of this, should we do a patch release? ( i prefer not, saves some time ;) )

jenshnielsen: I suggest that we nominate this to a patch release and include it if there are more things worth doing a patch release for

astafan8: agree. done https://github.com/QCoDeS/Qcodes/milestone/16

jenshnielsen: Cool (I did the same by adding the backport needed label)

trevormorgan: dang sorry I thought I caught all of these. Thanks for getting a fix in.

trevormorgan: Did CI not catch this because deprecated warnings aren't treated as failures?

trevormorgan (approved):

qcodes/dataset/data_set.py
jenshnielsen: Yes it would be very welcome to cleanup the last few warnings in the test suite and run with warnings as errors. That would have caught the call to the deprecated api. However, it would still not catch the use of the deprecated api in doc strings and example notebooks so its allways a good idea to search for calls to an api across the codebase when deprecating it

trevormorgan: I'll put in a note to look into cleaning up these last few warnings in CI so we can flag as errors going forward. In the future I'll make sure to search the code base when deprecating as well.

PR #2690, opened 2021-02-04
automatic validator for agilent E8527D based on options

simonzihlmann:
-determine the validators for the frequency and power parameter of the Agilent E8527D microwave source based on the installed options.

@jenshnielsen

### DIFF
--- a/qcodes/instrument_drivers/agilent/E8527D.py
+++ b/qcodes/instrument_drivers/agilent/E8527D.py
@@ -10,17 +10,65 @@ from qcodes.utils.helpers import create_on_off_val_mapping
 class Agilent_E8527D(VisaInstrument):
     """
     This is the QCoDeS driver for the Agilent_E8527D signal generator.
-
     This driver will most likely work for multiple Agilent sources.
-
     This driver does not contain all commands available for the E8527D but
     only the ones most commonly used.
     """
     def __init__(self, name: str, address: str,
-                 step_attenuator: bool = False,
                  terminator: str = '\n',
                  **kwargs: Any) -> None:
         super().__init__(name, address, terminator=terminator, **kwargs)
+
+        # Query installed options
+        self._options = self.ask_raw('DIAG:CPU:INFO:OPT:DET?')
+
+        # Determine installed frequency option
+        f_option = None
+        for f_option in ['513', '520', '521', '532', '540', '550', '567']:
+            if f_option in self._options:
+                frequency_option = f_option
+        if f_option == None:
+            raise RuntimeError('Could not determine the frequency option')
+
+        # convert installed frequency option to frequency ranges, based on: 
+        # https://www.keysight.com/us/en/assets/7018-01233/configuration-guides/5989-1325.pdf
+        # the frequency range here is the max range and not the specified (calibrated) one
+        frequency_options_dict = {
+            "513": (100e3, 13e9),
+            "520": (100e3, 20e9),
+            "521": (10e6, 20e9),
+            "532": (100e3, 31.8e9),
+            "540": (100e3, 40e9),
+            "550": (100e3, 50e9),
+            "567": (100e3, 70e9)
+            }
+
+        # assign min and max frequencies
+        self._min_freq: float
+        self._max_freq: float
+        self._min_freq, self._max_freq = frequency_options_dict[frequency_option]
+
+        # Based on installed frequency option and presence/absence of step attenuator (option '1E1')
+        # determine power range based on: https://www.keysight.com/us/en/assets/7018-01211/data-sheets/5989-0698.pdf
+
+        self._min_power: float
+        self._max_power: float
+        
+        if '1E1' in self._options:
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/agilent/E8527D.py
```diff
@@ -10,26 +10,75 @@
...
+        # Query installed options
+        self._options = self.ask_raw('DIAG:CPU:INFO:OPT:DET?')
+
+        # Determine installed frequency option
+        f_option = None
+        for f_option in ['513', '520', '521', '532', '540', '550', '567']:
+            if f_option in self._options:
+                frequency_option = f_option
+        if f_option == None:
+            raise RuntimeError('Could not determine the frequency option')
```
jenshnielsen, line 31: This looks a bit wrong to me. f_option can never be None but frequency_option can be unset.
You could do something like the following:

```python
frequency_option = None
for f_option in ['513', '520', '521', '532', '540', '550', '567']:
    if f_option in self._options:
        frequency_option = f_option
if frequency_option is None:
    raise RuntimeError('Could not determine the frequency option')
```

jenshnielsen:
@simonzihlmann thanks apart from the one comment this looks good. You will also need to re add the new line at the end of the file to make the CI happy

qcodes/instrument_drivers/agilent/E8527D.py
```diff
@@ -10,26 +10,75 @@
...
     """
     This is the QCoDeS driver for the Agilent_E8527D signal generator.
-
     This driver will most likely work for multiple Agilent sources.
-
     This driver does not contain all commands available for the E8527D but
     only the ones most commonly used.
     """
     def __init__(self, name: str, address: str,
-                 step_attenuator: bool = False,
```
jenshnielsen, line 20: I am not sure if we can just remove this step_attenuator since that may break some other users use of the code

simonzihlmann: ok. as you wish. But it will not be used within the driver, it will only be an argument of the init function.

jenshnielsen: I suggest that you do the following to avoid breaking users code.

keep step_attenuator and add something like the following:

```
                 step_attenuator: Optional[bool] = None,
                 terminator: str = '\n',
                 **kwargs: Any) -> None:

    if step_attenuator is not None:
         warnings.warn("step_attenuator argument to E8527D is deprecated and has no effect. It will be removed in the future.)
```

qcodes/instrument_drivers/agilent/E8527D.py
```diff
@@ -1,3 +1,4 @@
+import warnings
 from typing import Any
```
jenshnielsen, line 2: you will now also need to import Optional from typing

qcodes/instrument_drivers/agilent/E8527D.py
```diff
@@ -10,26 +11,86 @@
...
+
+        # Query installed options
+        self._options = self.ask_raw('DIAG:CPU:INFO:OPT:DET?')
+
+        # Determine installed frequency option
+        frequency_option = None
+        for f_option in ['513', '520', '521', '532', '540', '550', '567']:
+            if f_option in self._options:
+                frequency_option = f_option
+        if frequency_option == None:
```
jenshnielsen, line 37: It is considered better python style to use is rather than == when comparing against None, True or False

jenshnielsen (approved):

qcodes/instrument_drivers/agilent/E8527D.py
simonzihlmann: why? could you please explain me? I already saw this on the mypy test. Does it has to do that the "frequency_option" could be either a boolean or a str depending on the if statement?

simonzihlmann: ahh, I guess it is for "step_attenuator: Optional[bool] "

jenshnielsen: No, this is because we changed the type of `step_attenuator` so it now can be either None or True or False by using Optional on line 19. 

I suggested this such that we change the default value for `step_attenuator` to None such that if any user explicitly sets it to any of the previously allowed values (True or False) they will trigger the warning that it is unused and they should stop setting it. 

We are trying to type almost all the code within QCoDeS as this helps catch many types of errors in the code. If you are interested in learning more I recommend reading the introduction to the MyPy docs here https://mypy.readthedocs.io/en/stable/getting_started.html#getting-started 

Hope that makes sense :)

jenshnielsen: > ahh, I guess it is for "step_attenuator: Optional[bool] "
Yes I'm sorry I did not see this before I wrote my reply

simonzihlmann: thanks!

PR #2692, opened 2021-02-05
Fix station snapshot of parameters with snapshot_exclude

### LINKED ISSUE #2671: station.snapshot_base doesn't respect parameter.snapshot_exclude
I noticed the following behavior when directly adding parameters to a station:

```py
from qcodes import Station, Parameter

station = Station()

station.add_component(Parameter("foo", snapshot_exclude=True))

station.snapshot_base()
```
This issues a warning "foo is used in the snapshot while it should be excluded from the snapshot", and the parameter indeed appears in the snapshot even though it shouldn't.

This is a simple fix, it should just be a an additional condition [here](https://github.com/QCoDeS/Qcodes/blob/85f56f312ad1d784de72a1b816022d5a1650803c/qcodes/station.py#L184). I don't have time to write a proper PR with tests for this right now, but I can submit something later if this isn't fixed by then.

mgunyho:
Fixes #2671 

However, there's still one issue: the line

```
station.add_component(excluded_parameter)
```

in the test still issues a warning, because `add_component` takes a snapshot, because of `update_snapshot=True`. How should this be handled?

I first thought that maybe it's the caller's responsibility to pass `update_snapshot=False` when adding an excluded parameter like this, but this won't work then when initializing the Station as `Station(param1, param2, excluded_param, param4...)`. 

So maybe there could be a similar check in `add_component` for `isinstance(component, Parameter) and not component.snapshot_exclude`, but that also feels a bit clumsy.

@astafan8


jenshnielsen:
@mgunyho thanks for the pr. 

> in the test still issues a warning, because add_component takes a snapshot, because of update_snapshot=True. How should this be handled?
> 
> I first thought that maybe it's the caller's responsibility to pass update_snapshot=False when adding an excluded parameter like this, but this won't work then when initializing the Station as Station(param1, param2, excluded_param, param4...).
> 
> So maybe there could be a similar check in add_component for isinstance(component, Parameter) and not component.snapshot_exclude, but that also feels a bit clumsy.
> 

I don't see a better solution than checking this in `add_component` too. You are probably seeing these issues since the usecase of directly adding parameters to the station is not particularly commonly used. In most cases users will add an instrument to the station and the snapshot of the instrument will take care of the filtering.

mgunyho:
Alright, I added it now. It's true that this is a bit of an edge case. We are using some custom Station subclasses where this came up.

astafan8:
@mgunyho custom station classes? :) that's interesting! could you share what these classes do and what is their goal? (we could also do it on slack if you prefer)

mgunyho:
@astafan8 We use station subclasses for a couple of purposes.

- The subclass is specific to an experiment, and it has a `load_instruments` method that loads a specific set of instruments from the config. It also initializes some metainstruments (that require loading physical instruments before they can be instantiated), and also some experiment-wide "global" parameters that are not bound to any instrument but only to the station (that's where this issue came up).
- We have a complicated custom version of the `doNd` sweep function as a method of the subclass. The subclass has an `experiment` attribute so we don't always have to do `station.do_sweep(..., exp=exp)`.

I'm not sure if we are doing things in the most optimal way, but it seems to work reasonably well for now.

astafan8:
@mgunyho thank you for sharign! the first point makes sense and is an itneresting case; the folks i'm working with usually use separate yml files and anyway load all instruments from them. For the second point, also fair, however, this will be better solved (and we have it on our radar, just haven't got to doing it) by introducing a concept/variable of active experiment that you set once, and doNds and what not can pick it up (at the moment, donds and also Measuremnt class pick the "last" experiment which may well be the "active" one but also may not).

jenshnielsen (approved):

PR #2693, opened 2021-02-05
Modernize packaging by moving most static packaging data to setup.cfg

jenshnielsen:
Inspired by http://ivory.idyll.org/blog/2021-transition-to-pyproject.toml-example.html 

At the moment it seems like we cannot fully implement pep518 as versioneer does not support it but this moves everything else


astafan8 (approved):
super awesome!!!!

PR #2699, opened 2021-02-09
Add a measurement param for KeysightE4980A driver

Akshita07:
Fixes taking measurement with KeysightE4980A using doNd methods.

- Add a measurement parameter

@astafan8 @jenshnielsen

### DIFF
--- a/qcodes/instrument_drivers/Keysight/keysight_e4980a.py
+++ b/qcodes/instrument_drivers/Keysight/keysight_e4980a.py
@@ -334,6 +334,15 @@ class KeysightE4980A(VisaInstrument):
                       "set."
         )
 
+        self.add_parameter(
+            "CPG_meas",
+            get_cmd=self._get_data,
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -334,6 +334,17 @@ def __init__(self,
                       "set."
         )
 
+        self.add_parameter(
+            "CPG_meas",
+            get_cmd=self._get_data,
+            names=("capacitance", "conductance"),
+            units=("F", "S"),
+            parameter_class=MultiParameter,
```
jenshnielsen, line 342: I don't think this will work. I would subclass MultiParameter and add the _get_data as a get_raw method

Akshita07:
A measurements class is already implemented which can be used to achieve what this PR intended to accomplish. Hence, it is not needed.

PR #2700, opened 2021-02-09
Guid-helpers: Use ast instead of regex

Dominik-Vogel:
The regular expression used to read guids from a string represantation of a list/tuple of guids was flawed and managed to sneak around the test, by working on 0, 1, and 2 guids.
In this PR I am using the `ast.parse` instead, which makes it a lot easier and error proof.
Thanks @astafan8 !


qcodes/dataset/guid_helpers.py
```diff
@@ -77,16 +77,10 @@ def guids_from_list_str(s: str) -> Optional[Tuple[str, ...]]:
...
-    open_parens = r"[\[\(\{]"
-    close_parens = r"[\]\)\}]"
-    m = re.match(
-        fr"^\s*{open_parens}?"
-        fr"(?:{captured_guid},)*{captured_guid}?"
-        fr"{close_parens}?\s*$",
-        s,
-    )
-    if m is None:
+    parsed = (ast.parse(s, mode='eval')).body
```
astafan8, line 80: why `parse` as oppsoed to `literal_eval`? for simple and builtin data types, i read in the docs that literal_eval is the safe and recommended way, and you get a python object from it that you can safely perform operations on, as opposed to working with an ast itself.
```suggestion
    parsed = (ast.literal_eval(s, mode='eval')).body
```

qcodes/dataset/guid_helpers.py
```diff
@@ -77,16 +77,10 @@ def guids_from_list_str(s: str) -> Optional[Tuple[str, ...]]:
...
+    if not isinstance(parsed, (ast.List, ast.Tuple, ast.Set)):
         return None
-    return tuple(v for v in m.groups() if v is not None)
+    if not all([isinstance(e, ast.Constant) for e in parsed.elts]):
+        return None
+    return tuple (v.value for v in parsed.elts)
```
astafan8, line 85: weird unnecessary space ;)
```suggestion
    return tuple(v.value for v in parsed.elts)
```

qcodes/dataset/guid_helpers.py
```diff
@@ -77,16 +77,10 @@ def guids_from_list_str(s: str) -> Optional[Tuple[str, ...]]:
...
-        fr"(?:{captured_guid},)*{captured_guid}?"
-        fr"{close_parens}?\s*$",
-        s,
-    )
-    if m is None:
+    parsed = (ast.parse(s, mode='eval')).body
+    if not isinstance(parsed, (ast.List, ast.Tuple, ast.Set)):
         return None
-    return tuple(v for v in m.groups() if v is not None)
+    if not all([isinstance(e, ast.Constant) for e in parsed.elts]):
```
astafan8, line 83: list is not necessary
```suggestion
    if not all(isinstance(e, ast.Constant) for e in parsed.elts):
```

astafan8 (approved):
Awesome! :)

qcodes/dataset/guid_helpers.py
astafan8: it's ok like this :)

qcodes/dataset/guid_helpers.py
```diff
@@ -77,16 +78,30 @@ def guids_from_list_str(s: str) -> Optional[Tuple[str, ...]]:
...
+    if isinstance(parsed, ast.Str):
+        if len(parsed.s) > 0:
+            return (parsed.s,)
+        else:
+            return tuple()
+
+    if not isinstance(parsed, (ast.List, ast.Tuple, ast.Set)):
+        return None
+
+    if not all(isinstance(e, ast.Constant) for e in parsed.elts):
```
jenshnielsen, line 103: Looks like the current code fails since this is not a list of ast.Constant but a list of ast.Str which have a .s rather than a .v member

astafan8: indeed! thanks!

PR #2707, opened 2021-02-10
SP983c Basel IV amplifier remote control

ThorvaldLarsen:
Add a driver for the Basel SP983 amplifier for remote control. The string interaction is somewhat weird for the driver but even so the current commands can likely be written in a cleaner way.

If an error occurs the instrument can get stuck in a loop if an error occurs where one has to ping the instrument multiple times to clear the error message from the VISA log. Would be good to discuss with a developer the best way to handle this case. Driver documentation is sparse but can be found here: [link](https://2d82ac02-eae7-42d1-a874-f7d09fc59cf0.filesusr.com/ugd/68eb62_730add610f284337b3010b040a80446a.pdf).

Need to retest driver from the branch.


astafan8:
thanks @ThorvaldLarsen for the PR! apart from what you mentioned, we should also look into adding the remote control capability and an option to the existing basel preamp driver (or at least, make sure the set of parameters are exactly the same between the "virtual" and remote-controlled driver).

qcodes/instrument_drivers/basel/sp983c_remote.py
```diff
@@ -0,0 +1,104 @@
...
+from qcodes.utils import validators as vals
+
+
+class SP983A(VisaInstrument):
+    """
+    A driver for Basel Preamp's (SP983C) Remote Instrument - Model SP983A.
+
+    Args:
+        name
+        address
```
astafan8, line 15: missing explanation of arguments?

qcodes/instrument_drivers/basel/sp983c_remote.py
```diff
@@ -0,0 +1,104 @@
...
+            used to set offset voltage parameter of the preamp and the
+            source parameter should represent a voltage source that is
+            connected to the "Offset Input Volgate" connector of the SP983C.
+    """
+    def __init__(self,
+                 name: str,
+                 address: str,
+                 input_offset_voltage: Optional[Parameter] = None,
+                 **kwargs: Any) -> None:
+        super().__init__(name, address, terminator="\r\n", **kwargs)
```
astafan8, line 28: please set the terminator's default value in the `def init`, as an argument with the  default value, because this way the terminator can't be changed (if needed) without changing the driver code itself, an i'm not sure that we want that. i think we want to provide the corect default, but we also should give the users he flexibility to use a different terminator if they have to for an unknown reason

qcodes/instrument_drivers/basel/sp983c_remote.py
```diff
@@ -0,0 +1,104 @@
...
+            scale=100,
+            source=input_offset_voltage,
+            parameter_class=DelegateParameter)
+
+    def get_idn(self) -> Dict[str, Optional[str]]:
+        vendor = 'Physics Basel'
+        model = 'SP 983A'
+        serial = None
+        firmware = None
+        return {'vendor': vendor, 'model': model,
```
astafan8, line 72: is it true that there's no way to get this information with the remote controller?

qcodes/instrument_drivers/basel/sp983c_remote.py
```diff
@@ -0,0 +1,104 @@
...
+            get_parser=self._parse_filter_value,
+            set_cmd=self._set_filter,
+            val_mapping={30: '30', 100: '100', 300: '300', 1000: '1k',
+                         3000: '3k', 10e3: '10k', 30e3: '30k',
+                         100e3: '100k', 1e6: 'FULL'},
+        )
+        self.add_parameter(
+            "overload_status",
+            label="Overload Status",
+            set_cmd=None,
```
astafan8, line 54: if this does not have a set command, shall we disallow setting it altogether?
```suggestion
            set_cmd=False,
```

qcodes/instrument_drivers/basel/sp983c_remote.py
```diff
@@ -0,0 +1,104 @@
...
+import numpy as np
+
+from qcodes.instrument.visa import VisaInstrument
+from qcodes.instrument.parameter import Parameter, DelegateParameter
+from qcodes.utils import validators as vals
+
+
+class SP983A(VisaInstrument):
+    """
+    A driver for Basel Preamp's (SP983C) Remote Instrument - Model SP983A.
```
astafan8, line 11: since we have a notebook for the other basel driver, we probably should mention the remote one there as well --> https://qcodes.github.io/Qcodes/examples/driver_examples/Qcodes%20example%20with%20Basel%20SP983c%20Preamp.html?highlight=basel

FarBo:
@Akshita07 
Will you handle the raised comments by @astafan8 ?

Akshita07:
@FarBo Yes, I'll address them tomorrow. Thanks Farzad.

qcodes/instrument_drivers/basel/sp983c_remote.py
Akshita07: yes, in the manual there is no command mentioned to get this information from the remote controller.

qcodes/instrument_drivers/basel/sp983c_remote.py
Akshita07: added the explanation

qcodes/instrument_drivers/basel/sp983c_remote.py
Akshita07: updated as suggested

qcodes/instrument_drivers/basel/sp983c_remote.py
Akshita07: added remote's information to the notebook

astafan8 (approved):

FarBo (approved):

docs/examples/driver_examples/Qcodes example with Basel SP983c Preamp.ipynb
```diff
@@ -466,13 +466,66 @@
...
+   "source": [
+    "from qcodes.instrument_drivers.basel.sp983c_remote import SP983A"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "junior-transportation",
+   "metadata": {},
+   "source": [
+    "This driver can be used in exactly the same way as Basel Preamp driver mentioned above with addition of checking overload status. Initialization happens as below."
```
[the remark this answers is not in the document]
Akshita07: The information has been added regarding the remote's driver that it replaces the basel preamp driver when remote is being used with the preamp.

PR #2714, opened 2021-02-11
Sync environment_forge_full with setup.cfg

astafan8:


.github/static/environment_forge_full.yml
```diff
@@ -24,22 +26,21 @@ dependencies:
...
  - pytest-xdist>=2.0.0
  - pytest-mock>=3.0.0
- - pyzmq
  - lxml>=4.3.0
  - gitpython>=3.0.6
- - testpath
+ # extra relevant requirements from setup.cfg
+ - pyqtgraph>=0.11.0
+ # required for building docs
  - pandoc
```
jenshnielsen, line 41: We are not currently building the docs with Anaconda so we could remove this

.github/static/environment_forge_full.yml
```diff
@@ -24,22 +26,21 @@ dependencies:
...
  - pytest>=6.0.0
  - pytest-rerunfailures>=5.0.0
  - pytest-xdist>=2.0.0
  - pytest-mock>=3.0.0
- - pyzmq
  - lxml>=4.3.0
  - gitpython>=3.0.6
- - testpath
+ # extra relevant requirements from setup.cfg
+ - pyqtgraph>=0.11.0
```
jenshnielsen, line 39: does this automatically install pyqt (I guess we should so that we can run the pyqtgraph related tests(

astafan8: yes, we get qt-5.12.9 and pyqt-5.12.3 ..... oh dear

is this good enough, or shall we pin pyqt or qt or lese?

.github/static/environment_forge_full.yml
astafan8: indeed, let me remove it

.github/static/environment_forge_full.yml
jenshnielsen: I think that is fine for now. (I would argue that that dependency is wrong since pyqtgraph can use pyside(2) too)

jenshnielsen (approved):

PR #2715, opened 2021-02-11
keep github actions up to date with dependabot

jenshnielsen:


astafan8 (approved):
is this for the external github actions steps that we use/re-use? awesome!

jenshnielsen:
> 
> 
> is this for the external github actions steps that we use/re-use? awesome!

Yes indeed that is the idea. Was recommended in the docs of the chocolaty one

PR #2726, opened 2021-02-12
Push docs from gh pages

jenshnielsen:


.github/workflows/docs.yaml
```diff
@@ -79,3 +79,11 @@ jobs:
       with:
         name: docs_${{ matrix.python-version }}_${{ matrix.os }}
         path: ${{ github.workspace }}/docs/_build/html
+    - name: Deploy to gh pages
+      uses: JamesIves/github-pages-deploy-action@4.0.0
+      with:
+        branch: gh-pages
+        folder: ${{ github.workspace }}/docs/_build/html
+        dry-run: true  # remove
+        clean: true
```
astafan8, line 88: for consistency :) maybe we can pass something measningful to the "???" which used to be build number of some kind in azure pipelines
```suggestion
        clean: true
        git-config-email: "bot"
        git-config-name: "Documentation Bot"
        commit-message: "Generated gh-pages for ???"
```

astafan8 (approved):
Looks great! github actions make it so much easier :)

.github/workflows/docs.yaml
jenshnielsen: I think it already sets a better commit message. "Deploying to gh-pages from @ QCoDeS/Qcodes@40f6dd483f4850e2305fa2cafab42e0f1b5e337d"

I will set the user

PR #2733, opened 2021-02-15
fetch full git history when building docs so tag can be included in docs build info

jenshnielsen:
Currently the docs renders the version number as 0+unknown. This is a quick fix to get the right version number back. 
It may be worth investigating using a tree or a blob less clone to speed this up. https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/

@astafan8


astafan8 (approved):
as a quickfix - great!

PR #2734, opened 2021-02-15
Add optional tqdm progress bar to doNd functions

### LINKED ISSUE #2728: Include tqdm progress bar in doNd functions
As the title suggests, using the progress bar offered by e.g. `tqdm.notebook` is a most simple but extremely useful feature for using `doNd` to run long measurements.

astafan8:
* using tqdm.auto to determine whether console or jupyter-notebook-widget-like progress bar should be used.
* new entry in qcodes config called dond_show_progress so that passing argument to doNds is not necessary

Closes #2728 .

@gaberber does this address the need? Could you test and verify?


qcodes/utils/dataset/doNd.py
```diff
@@ -312,7 +320,14 @@ def do1d(
     # reimplemented from scratch
     with _catch_keyboard_interrupts() as interrupted, meas.run() as datasaver:
         additional_setpoints_data = _process_params_meas(additional_setpoints)
-        for set_point in np.linspace(start, stop, num_points):
+        setpoints = np.linspace(start, stop, num_points)
+
+        # flush to prevent unflushed print's to visually interrupt tqdm bar
+        # updates
+        sys.stdout.flush()
+        sys.stderr.flush()
```
jenshnielsen, line 328: How come we need this? I guess this still does not resolve the issue if something is printed within the dond function?

jenshnielsen (approved):

qcodes/utils/dataset/doNd.py
astafan8: yes, exactly. Measurement when starts it print "starting run with id: 42", and that sometimes can get flushed after tqdm started also printing. hence we flush first, and then tqdm can start printing.

PR #2735, opened 2021-02-15
attempt blob-less clone for docs build

astafan8:


astafan8:
ok, let's just wait until actions/checkout implement the blobless clone....

PR #2746, opened 2021-02-17
bump pylint, astroid and lazy-object-proxy

jenshnielsen:
Combine these since they must be upgraded together.


jenshnielsen:
This will require a new version of pylint

astafan8 (approved):

PR #2747, opened 2021-02-17
Update install instructions

jenshnielsen:
Update to the readme:

Remove Azure badge
Use the correct target for docs build

### DIFF
--- a/docs/start/index.rst
+++ b/docs/start/index.rst
@@ -11,10 +11,10 @@ Requirements
 ------------
 
 You need a working python 3.x installation to be able to use QCoDeS. We highly
-recommend installing Anaconda, which takes care of installing Python and
+recommend installing Miniconda, which takes care of installing Python and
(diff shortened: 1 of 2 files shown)


astafan8 (approved):
great update to the guide!

docs/start/index.rst
```diff
@@ -11,10 +11,10 @@ Requirements
 ------------
 
 You need a working python 3.x installation to be able to use QCoDeS. We highly
-recommend installing Anaconda, which takes care of installing Python and
+recommend installing Miniconda, which takes care of installing Python and
 managing packages. In the following it will be assumed that you use Anaconda.
```
FarBo, line 15: Anaconda should be changed to Miniconda here as well

FarBo:
This is a great update. 
I am just thinking if we should recommend installing between pip or conda-forge, or not?

PR #2763, opened 2021-02-22
Add pandas and xarray methods to dataset.cache matching the dataset

jenshnielsen:
Tests to be done via https://github.com/QCoDeS/Qcodes/pull/2771


trevormorgan (approved):

astafan8:
>  Update and improve tests

@jenshnielsen ended up adding one test with parametrization per export format only for 2D data. Feel free to adjust as needed to save some test execution time. some of them are failing, sorry, didn't have time to dig into why

astafan8 (approved):

PR #2768, opened 2021-02-23
Simplify science req tests

jenshnielsen:
Don't explicitly constrain transitive dependencies of spyder. We don't need this is we install the science requirements in the same step as the regular requirements as pip will automatically flag any problematic overlap


astafan8 (approved):
AWESOMEEE!!!!! :)

PR #2774, opened 2021-02-25
add dataset.export_data to save as netcdf file

guenp:
Fixes #issuenumber.

- Add an `export_data` method to `DataSet` that allows users to export data to a supported file format. The goal is not to save data "live" (i.e. per row) but only at the end of a measurement. I am open to a different name for this method if that makes more sense.
- Addition of "export_type", "export_prefix" and "export_path variables in the qcodes config under "dataset". These variables contain respectively the data type to export in (if this is set to `null`, export is disabled), the prefix to add to the file and the path to export the file to.

@jenshnielsen

### DIFF
--- a/qcodes/configuration/qcodesrc.json
+++ b/qcodes/configuration/qcodesrc.json
@@ -67,7 +67,10 @@
         "write_in_background": false,
         "write_period": 5.0,
         "dond_plot": false,
-        "dond_show_progress": false
+        "dond_show_progress": false,
+        "export_type": null,
+        "export_prefix": "qcodes_",
+        "export_path": "~"
     },
     "telemetry":
     {
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -6,6 +6,8 @@ import warnings
 import os
 import time
 import uuid
+import enum
+
 from dataclasses import dataclass
 from queue import Empty, Queue
 from threading import Thread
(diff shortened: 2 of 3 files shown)


guenp:
@jenshnielsen to reduce boilerplate for the user would it be an idea to add a call to `DataSet.export_data()` to the `__exit__` method of the `Runner`? Or would we want to add it to `.flush_data_to_database()`?

qcodes/configuration/qcodesrc.json
```diff
@@ -67,7 +67,10 @@
         "write_in_background": false,
         "write_period": 5.0,
         "dond_plot": false,
-        "dond_show_progress": false
+        "dond_show_progress": false,
+        "export_type": null,
+        "export_prefix": "qcodes_",
+        "export_path": "~"
```
jenshnielsen, line 73: You will also need to update the schema file next to this one with the new fields and their types

qcodes/dataset/data_set.py
```diff
@@ -1693,6 +1695,29 @@ def _flush_data_to_database(self, block: bool = False) -> None:
...
+        xarr_dataset = self.to_xarray_dataset()
+        xarr_dataset.to_netcdf(path=path)
+
+    def export_data(self) -> None:
+        """Export data to disk with file name {prefix}{run_id}.{ext}.
+        Values for the export type, path and prefix are set in the qcodes "dataset" config.
+        """
+
+        from qcodes.dataset.data_export import DataExportType, get_data_export_type, get_data_export_path, get_data_export_prefix
+        data_export_type = get_data_export_type()
```
jenshnielsen, line 1711: Personally I would make these input arguments too and then use the config iff no input argument is set

jenshnielsen (commented):
I think putting it in the `__exit__` of the runner would be the best option. flush_data_to_database is intended to be called at any time to just ensure that all currently captured data has been flushed to disk

guenp:
I renamed the branch as per the [contributing guidelines](https://qcodes.github.io/Qcodes/community/contributing.html) and will reopen this PR.

PR #2816, opened 2021-03-12
Abstract parameter

sohailc:
Abstract parameters allow us to create abstract instrument types which are guaranteed to have certain parameters present. For instance, this will allow us to create a unified interface for all voltage sources.

Usually, when a parameter is added to an instrument and that instrument is sub classed and a parameter of the same name is added in the subclass as well, this will raise an exception. This issue is circumvented by specifying in the base class that the parameter being added there is of type AbstractParameter.

A catch is that, if a unit is specified in the base class and a unit is specified in the sub class, these units must match.


qcodes/instrument/parameter.py
```diff
@@ -2766,3 +2766,29 @@ def expand_setpoints_helper(parameter: ParameterWithSetpoints,
         data = results
     res.append((parameter, data))
     return res
+
+
+class AbstractParameter(Parameter):
```
astafan8, line 2771: does it have to inherit from Parameter? this feature will be more useful if it inherits from _BaseParameter becasue then abstract definitions of ParameterWithSetpoints and MultiParameter may be supported

qcodes/instrument/parameter.py
```diff
@@ -2766,3 +2766,29 @@ def expand_setpoints_helper(parameter: ParameterWithSetpoints,
...
+    def __init__(self, name: str,
+                 instrument: Optional['InstrumentBase'] = None,
+                 **kwargs: Any):
+        """
+        An abstract parameter is meant to define an abstract interface,
+        much like abc.abstractmethod. This is useful for defining instrument
```
astafan8, line 2777: question - unlike the ABC infrastructure, the current implementation of the AbstractParameter is such that it is not enforced to be "implemented" at the class instantiation time. the only thing that there is is the set/get which raise exception. questions are:
- is this good enough for defining "abstract instrument interfaces"?
- if not, can we come up with something better (that perhaps does not use metaclasses ;) )?

qcodes/instrument/parameter.py
jenshnielsen: Agreed perhaps we should consider making AbstractParameter the baseclass for parameter

jenshnielsen: _BaseParameter already implements `get_raw` and `set_raw` that will raise a NotIMplementedError if called. 

I think it would be better if we simply rename `_BaseParameter` to `AbstractParameter` (perhaps retaining `_BaseParameter` as an alias for now and use that everywhere.

qcodes/instrument/parameter.py
jenshnielsen: I think this is hard to do in a clean way since there are so many use-cases of parameter.

Normally one would make `get_raw` and `set_raw` abstract methods and then 
enforce that it is illegal to instantiate a class where these are not overwritten but we have to support the following usecases. 

* get only parameter. `set_raw` is not implemented
* set only parameter. `get_rwa` is not implemented
* set/get cmd passed as a string to `__init__`

qcodes/instrument/parameter.py
sohailc: Oh these are great suggestions! I will modify my PR accordingly

jenshnielsen:
So one potential problem is the promotion of unit to the baseclass. Currently `_BaseParmeter` does not have a unit attribute since multiparameter does not have a unit (it has a units tuple) I am not sure how we best deal with this but perhaps we should Inject AbstractParameter between _BaseParameter and Parameter such that all classes that have a unit are subclasses of abstract parameter

sohailc:
@jenshnielsen I'm not sure if I understand the problem. `MultiParameter` does not have a `unit` argument so its super class is always instantiated with `unit=None`. So an instance of `MultiParameter` will always have the `unit` attribute set to `None`. I fail to see the problem with this.

jenshnielsen:
@sohailc This is an api break. You have demonstrated that your self by the change that you had to make to test_multparamter.py 

Not having an attribute is not the same as having an None attribute. Now perhaps we don't care about that api break but at the very least it is something that we need to think carefully about and not something that one can trivially do. 

The semantics of unit being None is unclear. What does that mean. I could see it both meaning unitless and meaning that the thing does not have a unit

sohailc:
I see your point. I will rewrite this so that `AbstractParameter` is `_BaseParameter` + unit. Then `MultiParameter` inherits from the latter and everything else from the former. 

I do have one observation though. We already state in our current documentation that a parameter with an empty string `''` is unitless. So it stands to reason that a parameter with a None unit is something that does not have a unit.

qcodes/instrument/parameter.py
```diff
@@ -1541,7 +1558,7 @@ def snapshot_base(self, update: Optional[bool] = True,
         return snapshot
 
 
-class ArrayParameter(_BaseParameter):
+class ArrayParameter(AbstractParameter):
     """
```
jenshnielsen, line 1562: ArrayParameter should also forward its unit to the `AbstractParameter` class rather than setting the attribute it self

qcodes/tests/parameter/test_abstract_parameter.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+
+
+def test_basic():
+    """
+    We should be able to use the instrument like any other
+    """
+    source = MyVoltageSource("name")
+    source.voltage.set(2.3)
+    assert source.voltage.get() == 2.3
```
jenshnielsen, line 50: These tests will leave open instruments around. I suggest adding a finalizer that will close the instrument https://docs.pytest.org/en/stable/fixture.html#adding-finalizers-directly  or make the instrument a fixture that yields the instrument and closes it after the test

qcodes/instrument/base.py
```diff
@@ -93,11 +93,26 @@ def add_parameter(self, name: str,
             **kwargs: Constructor arguments for ``parameter_class``.
 
         Raises:
-            KeyError: If this instrument already has a parameter with this
-                name.
-        """
-        if name in self.parameters:
-            raise KeyError(f'Duplicate parameter name {name}')
+            ValueError: If this instrument already has a parameter with this
```
jenshnielsen, line 96: Chancing this from a KeyError to a ValueError is technically an API break but I am not sure we care? @astafan8 What do you think ?

jenshnielsen:
I don't really like that it is possible to instantiate a Instrument that has non overwriten abstract parameters and you will only learn about that once you try to call the parameters get or set methods. 

It's not however completely trivial how to prevent this from happening. One idea would be to verify that all parameters in self.parameters are non abstract. However this requires a method to run after `__init__` It seems hard to do that without a metaclass https://stackoverflow.com/questions/16017397/injecting-function-call-after-init-with-decorator which we probably want to avoid.

sohailc:
How about if a class containing Abstract paramters is itself decorated. It is not allowed for a class to have abstract parameters at init, except with it is decorated with `@AbstractDriver` or something.

astafan8:
> How about if a class containing Abstract paramters is itself decorated. It is not allowed for a class to have abstract parameters at init, except with it is decorated with @AbstractDriver or something.

Could you expand on this? especially how does it prevent instantiation of a real instrument driver class that still has non overwriten abstract parameters?

sohailc:
@astafan8 Never mind, that particular idea won't work. What we can do is have an explicit init argument in `AbstractInstrument` which tells the class whether it is an abstract driver or not, e.g. `abstract_driver=True`. We can add abstract parameters if and only if this parameter is True. Furthermore, this parameter is False by default.

sohailc:
this is too much work... I'm abandoning this. Maybe we will revisit this in the future.

PR #2832, opened 2021-03-18
Updating SR830 driver with adding autorange method and its example notebook

FarBo:
This PR:

Updates SR830 driver with the autorange function, which was introduced in the driver's example notebook.
Updates SR830 driver notebook with replacing old Measure with do0d and autorange function with the new implementated method in the driver.


qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -695,3 +696,27 @@ def _set_sensitivity(self, s: float) -> int:
...
         else:
             return self._CURR_TO_N[s]
+
+    def autorange(self, max_changes: int=1) -> None:
+        """
+        Automatically changes the sensitivity of the instrument according
+        to the R value and defined max_changes.
+        Args:
+            max_changes: Maximum number of steps allowing the function to
+            automatically change the sensitivity (default is 1). Actual
```
jenshnielsen, line 706: Indent these 3 lines that are continuation by 4 more spaces

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -695,3 +696,27 @@ def _set_sensitivity(self, s: float) -> int:
...
+            steps before reaching the limit in the function might be more
+            or less than this maximum.
```
jenshnielsen, line 708: How can it be more ?

FarBo: Depends on the R value and the autorange implementation here, the sensitivity can change a finite number of steps before reaching to the limit that it doesn't change. So if you limit the max_changes steps below that headroom, it means that the actual change is more than what you set here by this max_changes.

docs/examples/driver_examples/Qcodes example with Stanford SR830.ipynb
```diff
@@ -118,55 +124,99 @@
...
-   "metadata": {
-    "collapsed": true
-   },
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "sr.buffer_SR(512)  # Sample rate (Hz)"
+   ]
+  },
+  {
```
jenshnielsen, line 185: Is it worth it to also show how the buffer can be filled by using an external trigger (I assume that it can) rather than using a time series?

qcodes/instrument_drivers/stanford_research/SR830.py
jenshnielsen: Ok so perhaps what you mean is `the actual number of steps needed to change to the optimal sensitivity may be more or less than the maximum`?

The number of steps performed is always less than or equal to the maximum right?

jenshnielsen (approved):

docs/examples/driver_examples/Qcodes example with Stanford SR830.ipynb
FarBo: Alright, but you mean a software trigger, right?

qcodes/instrument_drivers/stanford_research/SR830.py
FarBo: > Ok so perhaps what you mean
Yes, it is what I mean. I will rephrase it.

> The number of steps performed
True

docs/examples/driver_examples/Qcodes example with Stanford SR830.ipynb
jenshnielsen: Does it not have a hardware trigger ? Anyway its fine to wait with this. Perhaps just mention that there are other options will be good enough for now

FarBo: There is a trig button in its front panel. Unfortunately, it is not easy to find one and work with it today. So we can mention the trigger in the notebook for now.

FarBo: How about something like this:
Here we fill the buffer for one second as shown below. In a real use, a trigger can be sent to fill the buffer either on the instrument or a software trigger using the driver:

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -695,3 +696,27 @@ def _set_sensitivity(self, s: float) -> int:
             return self._VOLT_TO_N[s]
         else:
             return self._CURR_TO_N[s]
+
+    def autorange(self, max_changes: int = 1) -> None:
+        """
+        Automatically changes the sensitivity of the instrument according to
+        the R value and defined max_changes.
```
jenshnielsen, line 703: You need a blank line before the `Args:` line to get the docs to build correctly

FarBo: OOOPS
I missed that. Fix is comming :)
Thanks

PR #2834, opened 2021-03-19
tests: cleanup warning handling 

jenshnielsen:
* use pytests warning fixture 
* Port some tests to pytest style

This should eventually help us get to a point where warnings can become errors for testing

### DIFF
--- a/qcodes/tests/test_deprecate.py
+++ b/qcodes/tests/test_deprecate.py
@@ -1,18 +1,10 @@
 import warnings
-from contextlib import contextmanager
 import pytest
 
 
 from qcodes.utils.deprecate import (
     deprecate, issue_deprecation_warning, _catch_deprecation_warnings,
-    assert_not_deprecated, assert_deprecated)
-
-
-@contextmanager
-def _ignore_warnings():
-    with warnings.catch_warnings():
-        warnings.simplefilter("ignore")
-        yield
+    assert_not_deprecated, assert_deprecated, QCoDeSDeprecationWarning)
 
 
 def test_assert_deprecated_raises():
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -88,6 +88,8 @@ markers = serial
 ; Filter google rcp related deprecation warnings.
 ; See https://github.com/googleapis/python-api-common-protos/issues/23
 filterwarnings =
+    error
+    ignore:Exception ignored in
```
astafan8, line 92: these two are so generic :) it a plan to remove them before merge or ?

jenshnielsen: I think you have misunderstood how this works. The first line means that all warnings that do not match the ignore lines will be an error so it should be generic. It would indeed be nice to make the next one less generic but I am not sure it is possible since the next char in the error string is a : and there seems to be no way to escape that in this ignore dsl

astafan8: indeed, sorry for misunderstanding the first line. for the second line, the docs say:
> message is a string containing a regular expression that the start of the warning message must match. The expression is compiled to always be case-insensitive

so perhaps escpaing the colon in a regexp way, or making a regexp that ignores the "beginning of the string" default behavior (say `*.`) and reflects the actual error message would work?

trevormorgan:
@jenshnielsen Should we try to land this before 0.26.0 release? I can take a look at the errors for python 3.9

jenshnielsen:
@trevormorgan We could at least land the improvements to the tests to use the correct context managers. I think there is a test within the monitor module that needs significant work on all versions of python so it may take some time to get all warnings enabled

setup.cfg
jenshnielsen: Removed this from the pr

jenshnielsen:
@trevormorgan and @astafan8 Retargeted this to just cleanup some warning handling in tests so this should be ready to land now

astafan8 (approved):

PR #2837, opened 2021-03-22
Remove hardcoded acquire mode and allow users to set acquire mode for trace on Infiniium driver

Akshita07:
- Move setting of acquire_mode to prepare_curvedata method of RawTrace param

To Do:
- [ ] Test driver

@astafan8 @jenshnielsen


jenshnielsen (approved):

qcodes/instrument_drivers/Keysight/Infiniium.py
```diff
@@ -72,6 +72,11 @@ def prepare_curvedata(self) -> None:
         assert isinstance(root_instrument, Infiniium)
         root_instrument.trace_ready = True
 
+        # set up the instrument
+        # ---------------------------------------------------------------------
+        # realtime mode: only one trigger is used
+        instr._parent.acquire_mode(acquire_mode)
```
jenshnielsen, line 78: Cant you use `root_instrument` as above ?

qcodes/instrument_drivers/Keysight/Infiniium.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, name: str, instrument: Instrument, channel: int):
                          )
         self._channel = channel
 
-    def prepare_curvedata(self) -> None:
+    def prepare_curvedata(self, acquire_mode: str = 'RTIMe') -> None:
         """
         Prepare the scope for returning curve data
         """
```
jenshnielsen, line 47: Should we document the acquire_mode argument?

qcodes/instrument_drivers/Keysight/Infiniium.py
Akshita07: Yes, updating now

qcodes/instrument_drivers/Keysight/Infiniium.py
```diff
@@ -44,6 +44,12 @@ def __init__(self, name: str, instrument: Instrument, channel: int):
     def prepare_curvedata(self, acquire_mode: str = 'RTIMe') -> None:
         """
         Prepare the scope for returning curve data
+
+        Args:
+            acquire_mode: This argument sets the `acquire_mode` param on the
+            Infiniium oscilloscope instrument for the measurement. Default
+            value is `RTIMe` which means realtime mode and only one trigger
+            is used.
```
jenshnielsen, line 52: Please indent all lines that are does not start with a parameter a further 4 spaces. 

```
            acquire_mode: This argument sets the `acquire_mode` param on the
                Infiniium oscilloscope instrument for the measurement. Default
                value is `RTIMe` which means realtime mode and only one trigger
                is used.
```

jenshnielsen: Otherwise it ends up looking like this:

Screenshot 2021-03-22 110235

Akshita07: ok

guenp:
@Akshita07 I just spoke with Rachpon about this. He would like for the acquire mode to only be set *once*, perhaps in the constructor, but not set before acquiring data. Would it be possible to set it in the constructor instead (just to the default value) and then just have the user call `dso.acquire_mode("my_mode")` if they want to?

Akshita07:
@guenp Sure we can try to do this and we have to make sure that any changes do not break current user's code. This was the main reason for me asking what other changes to the driver Rachpon has to make to the driver for his measurements. I will go through the manual and check if the RawTrace parameter can handle all acquire_modes and will reach out to Rachpon as well to understand better.

jenshnielsen:
I agree that we should do this. It is a breaking change but it is a good one. We should however make sure that this is mentioned in the changelog as a breaking change

qcodes/instrument_drivers/Keysight/Infiniium.py
```diff
@@ -366,6 +367,7 @@ class Infiniium(VisaInstrument):
 
     def __init__(self, name: str, address: str,
                  timeout: float = 20,
+                 acquire_mode: str = 'RTIMe',
```
jenshnielsen, line 370: I don't see the need to actually pass this to init function of the scope since it is already a parameter on the scope. If you want to set it to a default value (which I don't think we should) one can simply pass that as initial_value to `add_parameter`

Akshita07: Yes, you are right. We do not need to pass this value to init function. But since `RawTrace` class had previously set `acquire_mode` parameter to this default value, we do need to set it as initial_value. Is there better way to handle it?

jenshnielsen: I am not sure if we need that. It will be a breaking change no matter what. If we set the initial value then users may be surprised that the mode changes when they connect to the scope and setting it once on init will anyway not match the previous behavior

guenp: @jenshnielsen previously the acquire mode was set to a hard-coded value at acquisition time, which I think is equally surprising especially if you set the value manually and it then changes when you connect via qcodes. Perhaps this acquire mode should just default to None and only be set in the constructor if the user specifies it?

jenshnielsen: > @jenshnielsen previously the acquire mode was set to a hard-coded value at acquisition time, which I think is equally surprising especially if you set the value manually and it then changes when you connect via qcodes. Perhaps this acquire mode should just default to None and only be set in the constructor if the user specifies it?

Sorry I think you are misunderstanding what the issue at discussion is now. 
We all agree that this should not be set as part of `RawTrace` `get_raw` method. Neither should it be set as part of `prepare_curve_data` 

The problem is if it should be part of the constructor of the Infiniium class at all. I am arguing that it should **not** since that would allow it to change the value when connecting to the instrument which is confusing and error prone and not how most qcodes drivers work. Furthermore it is not needed at all since it is already a parameter on the instrument and the user can simply set it as such. 

```
inst = Infiniium(name='name', addr='addr)
inst.acquisition_mode('somemode') 
```

Akshita07: @jenshnielsen , @guenp has let me know that to run this driver, they just comment out the `acquire_mode` call in the `RawTrace` `get_raw` method without any other change to the driver. In this case, I think if we remove `acquire_mode` call from all places (both `RawTrace` and Infiniium constructor) we should be fine. I will implement this now and will test this with help of @guenp and Rachpon next to next week when I am back.

jenshnielsen:
@Akshita07 I think this is simple enough that we can merge without testing.

@sohailc Please make sure that this change to how the scope works is called out as a breaking change in the changelog

PR #2838, opened 2021-03-22
Extra explanation for Experiment and database in 15_minutes notebook

FarBo:
This PR:

1. Adds some extra explanation exphasasing on Experiment creation before Measurement.
2. Mentioning config file change is temporary for the current kernel
3. Adds explanation about running Experiment creates or load database.

### DIFF
--- a/docs/examples/15_minutes_to_QCoDeS.ipynb
+++ b/docs/examples/15_minutes_to_QCoDeS.ipynb
@@ -63,13 +63,13 @@
      "text": [
       "Logging hadn't been started.\n",
       "Activating auto-logging. Current session state plus future input saved.\n",
-      "Filename       : C:\\Users\\Jens-work\\.qcodes\\logs\\command_history.log\n",
+      "Filename       : C:\\Users\\a-fbonabi\\.qcodes\\logs\\command_history.log\n",
       "Mode           : append\n",
       "Output logging : True\n",
       "Raw input log  : False\n",
       "Timestamping   : True\n",
       "State          : active\n",
-      "Qcodes Logfile : C:\\Users\\Jens-work\\.qcodes\\logs\\210129-24288-qcodes.log\n"
+      "Qcodes Logfile : C:\\Users\\a-fbonabi\\.qcodes\\logs\\210322-13432-qcodes.log\n"
      ]
     }
    ],
@@ -129,13 +129,13 @@
      "output_type": "stream",
      "text": [
       "Activating auto-logging. Current session state plus future input saved.\n",
-      "Filename       : C:\\Users\\Jens-work\\.qcodes\\logs\\command_history.log\n",
+      "Filename       : C:\\Users\\a-fbonabi\\.qcodes\\logs\\command_history.log\n",
       "Mode           : append\n",
       "Output logging : True\n",
       "Raw input log  : False\n",
       "Timestamping   : True\n",
       "State          : active\n",
-      "Qcodes Logfile : C:\\Users\\Jens-work\\.qcodes\\logs\\210129-24288-qcodes.log\n"
+      "Qcodes Logfile : C:\\Users\\a-fbonabi\\.qcodes\\logs\\210322-13432-qcodes.log\n"
      ]
     }
    ],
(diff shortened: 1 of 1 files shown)


jenshnielsen (approved):

docs/examples/15_minutes_to_QCoDeS.ipynb
```diff
@@ -788,7 +804,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "After initialising the database we create the `Experiment` object. This object contains the name of the experiment and the sample, and the path of the database. You can  use `load_or_create_experiment` to find and return an experiment with the given experiment and sample name if it already exists, or create one if not found."
+    "After initialising the database we create the `Experiment` object. Note that if there is no database created or loaded before creating or loading an `Experiment`, by running `Experiment` modules that are related to creating an experiment, a database will be created or loaded (if already exists) according to the QCoDeS config. `Experiment` object contains the name of the experiment and the sample, and the path of the database. You can  use `load_or_create_experiment` to find and return an experiment with the given experiment and sample name if it already exists, or create one if not found."
```
astafan8, line 807: the "by running `Experiment` modules" is confusing - `Experiment` is class, what modules are meant?

i think this should be explained simpler and better - i suggest step-by-step explanation. first, just instroduce the experiment as it was before, and then in a paragraph-per-idea way explain what happens if there's an empty database without an experiment, or if there's no database but someone creates an experiment object.

FarBo: Sure. I added a little more clarifications. Please take a look and let me know if sounds reasonable.

I had a hard time typing initialize as initialise, though :)

docs/examples/15_minutes_to_QCoDeS.ipynb
```diff
@@ -858,7 +890,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Without creating or loading an `Experiment`, it is not possible to run any QCoDeS `Measurement`."
+    "It worths mentioning that what happens if we initilise a database but not create or load an experiment before running a QCoDeS `Measurement`. One of two following outcomes would happen: 1. if the initialised database is empty, then the `Measurement` will throw error because it could not find any `Experiment` in it; 2. if the databse has one or more `Experiment` created in it, then the `Measurement` will be performed and it inherits from the latest `Experiment` name and run number in the database."
```
astafan8, line 893: recommendation:
``then the `Measurement` will be performed and it inherits from the latest `Experiment` name and run number in the database.`` --> ``then creating a `Measurement` object will automatically pick up the latest `Experiment` from the database.``

otherwise the reader will be asking "what name?" "what run number if we were talking about the experiment", etc :)

FarBo:
I will perform a bit restructuring in this notebook as I think the current flow can be improved a bit. Then, It could be easier to convey what I want to add in this PR. 
So please do not merge it until then.

FarBo:
@astafan8 
You may look at the new changes and see if you see any improvement :)

docs/examples/15_minutes_to_QCoDeS.ipynb
```diff
@@ -700,51 +716,96 @@
...
+   "source": [
+    "initialise_or_create_database_at(\"~/experiments.db\")"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Current location of database"
+    "Note that it is user's responsibility to provide the correct absolute path for the existing database. The notation of the path may differ with respect to the operating system. The method ``initialise_or_create_database_at`` makes sure that your QCoDeS session is connected to the referred database. If the database file does not exist, it will be created at the provided path:"
```
astafan8, line 762: the path does not have to be absolute :) but it's the users responsibility to provide a path :)

astafan8 (approved):
It's much better, thanks a lot @FarBo ! (just one little comment)

docs/examples/15_minutes_to_QCoDeS.ipynb
FarBo: Yes, I don't know why it was there in the first place, though absolute is a nice vodka :)

PR #2841, opened 2021-03-23
use new version of sphinx json schema with string escape

jenshnielsen:
as my fix has made it into an upstream release we can revert back to that


astafan8 (approved):
Great oss interaction!!!!!!!!

PR #2846, opened 2021-03-25
add snapshot and other relevant metadata to xarray export

jenshnielsen:
@astafan8 

In principle this means that the qcodes dataset should be reconstructable from the exported data. 

Note that I have left out a few things on purpose.

The `run_id` since this is anyway not conserved when exporting
completed flag since this can be constructed from the completed time stamp 
exp_id since this is really specific to the specific db

- [ ] Add some tests

### DIFF
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -1246,6 +1246,7 @@ class DataSet(Sized):
                 paramspec_dict = self.paramspecs[str(dim)]._to_dict()
                 xrdataset.coords[str(dim)].attrs.update(paramspec_dict.items())
 
+        xrdataset.attrs["name"] = self.name
         xrdataset.attrs["sample_name"] = self.sample_name
         xrdataset.attrs["exp_name"] = self.exp_name
         xrdataset.attrs["guid"] = self.guid
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -1253,7 +1253,7 @@ class DataSet(Sized):
         xrdataset.attrs["run_timestamp"] = self.run_timestamp_raw
         xrdataset.attrs["completed_timestamp"] = self.completed_timestamp_raw
         xrdataset.attrs["parameters"] = self.parameters
-        xrdataset.attrs["run_description"] = self.description
+        xrdataset.attrs["run_description"] = serial.to_json_for_storage(self.description)
         xrdataset.attrs["snapshot"] = self.snapshot
         xrdataset.attrs["captured_run_id"] = self.captured_run_id
         xrdataset.attrs["captured_counter"] = self.captured_counter
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -1248,6 +1248,17 @@ class DataSet(Sized):
 
         xrdataset.attrs["sample_name"] = self.sample_name
         xrdataset.attrs["exp_name"] = self.exp_name
+        xrdataset.attrs["guid"] = self.guid
+        xrdataset.attrs["run_timestamp"] = self.run_timestamp_raw
+        xrdataset.attrs["completed_timestamp"] = self.completed_timestamp_raw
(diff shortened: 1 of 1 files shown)


qcodes/dataset/data_set.py
```diff
@@ -1248,6 +1248,17 @@ def _load_to_xarray_dataset(self, data: ParameterData) -> "xr.Dataset":
...
+        xrdataset.attrs["guid"] = self.guid
+        xrdataset.attrs["run_timestamp"] = self.run_timestamp_raw
+        xrdataset.attrs["completed_timestamp"] = self.completed_timestamp_raw
+        xrdataset.attrs["parameters"] = self.parameters
+        xrdataset.attrs["run_description"] = self.description
+        xrdataset.attrs["snapshot"] = self.snapshot
+        xrdataset.attrs["captured_run_id"] = self.captured_run_id
+        xrdataset.attrs["captured_counter"] = self.captured_counter
+
+        for metadata_tag, metadata in self._metadata.keys():
```
astafan8: Shall we not mix metadata with other keys and store metadata dict under “metadata” key?

qcodes/dataset/data_set.py
```diff
@@ -1248,6 +1248,17 @@ def _load_to_xarray_dataset(self, data: ParameterData) -> "xr.Dataset":
 
         xrdataset.attrs["sample_name"] = self.sample_name
         xrdataset.attrs["exp_name"] = self.exp_name
+        xrdataset.attrs["guid"] = self.guid
+        xrdataset.attrs["run_timestamp"] = self.run_timestamp_raw
+        xrdataset.attrs["completed_timestamp"] = self.completed_timestamp_raw
+        xrdataset.attrs["parameters"] = self.parameters
```
astafan8, line 1254: Parameters should redundant while run description is here? Or is it for convenience?

qcodes/dataset/data_set.py
```diff
@@ -1248,6 +1248,17 @@ def _load_to_xarray_dataset(self, data: ParameterData) -> "xr.Dataset":
...
+        xrdataset.attrs["run_description"] = self.description
```
astafan8, line 1255: Shall this be the run described object or it’s dict representation or actually it’s json string? What is more safe/stable?

qcodes/dataset/data_set.py
```diff
@@ -1248,6 +1248,17 @@ def _load_to_xarray_dataset(self, data: ParameterData) -> "xr.Dataset":
 
         xrdataset.attrs["sample_name"] = self.sample_name
         xrdataset.attrs["exp_name"] = self.exp_name
+        xrdataset.attrs["guid"] = self.guid
```
astafan8, line 1251: We also should export name of the dataset

qcodes/dataset/data_set.py
jenshnielsen: Perhaps, I could see this go either way

qcodes/dataset/data_set.py
jenshnielsen: It should probably be the json repr

qcodes/dataset/data_set.py
jenshnielsen: Not really sure this layout matches the way its stored in the dataset but I could also see the argument for nesting it one more level. I don't think we should call it metadata however since all of these columns are really metadata. Its more like `user added custom metadata` but what is a good name for that?

astafan8: agree with you! (and "this layout matches the way its stored in the dataset" i think is not relevant really). what about `extra_metadata` ? :)

jenshnielsen: So it looks like attrs do not allow None elements. We should normalize any none out of fixed columns (easy enough I think since we can use json str for description and snapshot but we also need to recursively normalize the metadata field

astafan8: or we remove the None-valued metadata items?

jenshnielsen: well that is one form of normalization

PR #2854, opened 2021-03-26
export more metadata to xarray dataset

jenshnielsen:
Replaces https://github.com/QCoDeS/Qcodes/pull/2846

@astafan8 

Compared to #2846 

We are using the string formatted time stamps. I can see that being useful for the user but perhaps we should also retain the raw timestamps to make it simpler to reconstruct a qcodes ds in the future. 

the extra metadata is now exported to a separate dict. No need to check for None since the qcodes ds does not allow that anyway. 

I also did a few more cleanups. Move all the export tests to a separate file and cleaned them up a little.

Added some docstrings for the changes in #2848


qcodes/dataset/data_set.py
```diff
@@ -1247,15 +1255,25 @@ def _load_to_xarray_dataset(self, data: ParameterData) -> "xr.Dataset":
...
+            "snapshot": self.snapshot_raw or "null",
             "guid": self.guid,
             "run_timestamp": self.run_timestamp() or "",
             "completed_timestamp": self.completed_timestamp() or "",
-            "run_id": self.run_id
+            "captured_run_id": self.captured_run_id,
+            "captured_counter": self.captured_counter,
+            "run_id": self.run_id,
+            "run_description": serial.to_json_for_storage(self.description)
         })
```
astafan8, line 1270: adding the raw, integer/float, represenation of timestamp. i think it's cleaner to NOT add an attibute if it's "None" other than add it as "null" (where a numeric value is expected) or as None-like value ("0" or "-1").
```suggestion
        if self.run_timestamp_raw is not None:
            xrdataset.attrs["run_timestamp_raw"]: self.run_timestamp_raw
        if self.completed_timestamp_raw is not None:
            xrdataset.attrs["completed_timestamp_raw"]: self.completed_timestamp_raw
```

qcodes/tests/dataset/test_dataset_export.py
```diff
@@ -0,0 +1,230 @@
...
+
+    _assert_xarray_metadata_is_as_expected(ds, mock_dataset)
+
+
+def _assert_xarray_metadata_is_as_expected(xarray_ds, qc_dataset):
+
+    assert xarray_ds.name == qc_dataset.name
+    assert xarray_ds.sample_name == qc_dataset.sample_name
+    assert xarray_ds.exp_name == qc_dataset.exp_name
+    assert xarray_ds.snapshot == json.dumps(qc_dataset.snapshot)
```
astafan8, line 223:
```suggestion
    assert xarray_ds.snapshot == qc_dataset.snapshot_raw
```

qcodes/dataset/data_set.py
```diff
@@ -1247,15 +1255,25 @@ def _load_to_xarray_dataset(self, data: ParameterData) -> "xr.Dataset":
                 xrdataset.coords[str(dim)].attrs.update(paramspec_dict.items())
 
         xrdataset.attrs.update({
```
astafan8, line 1257: note that this PR only adds attributes to xarray DATASET, but we also have the xarray dataarrays, `_load_to_xarray_dataarray_dict`, shall we add some attributes to them as well? i'm thinking at least name,exp_name,sample_name,guid,run_description (or even fancier, run_description of the contents of the dataarray only ;) )

qcodes/dataset/data_set.py
```diff
@@ -1254,6 +1254,14 @@ def _load_to_xarray_dataset(self, data: ParameterData) -> "xr.Dataset":
                 paramspec_dict = self.paramspecs[str(dim)]._to_dict()
                 xrdataset.coords[str(dim)].attrs.update(paramspec_dict.items())
 
+        self._add_metadata_to_xarray(xrdataset)
```
astafan8, line 1257: but this way wouldn't we have duplicated snapshot in xrdataset and all it's xrdatarrays ?

astafan8 (approved):

PR #2863, opened 2021-03-29
upgrade importlib-metadata

jenshnielsen:
Apparently 3.9.0 has some serious packaging issues


astafan8 (approved):
should we add importlib-metadata!=3.9.0 to setup.cfg?

jenshnielsen:
Nah I don't think it matters for our usecase

PR #2894, opened 2021-04-06
Driver sr830 update

RasmusBC59:
Adds class ChannelTrace, to replace class ChannelBuffer

The ChannelTrace should make it easier to do buffered acquisition.
 
added: Classes
- ChannelTrace(ParameterWithSetpoints)
- GeneratedSetPoints(Parameter)

added Parameters:
- sweep_start: Parameter
- sweep_stop: Parameter
- sweep_n_points: Parameter
- sweep_setpoints: GeneratedSetPoints
- ch{ch}_datatrace: ChannelTrace (to replace ch{ch}_databuffer)

added functions:
- set_sweep_parameters
 

- add example of using ch{ch}_datatrace
- depreciate ch{ch}_databuffer
- do we need sweep_start, sweep_stop and sweep_n_points to be parameters of the instruments?

@jenshnielsen


qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -3,13 +3,73 @@
...
+                             'channels 1 and 2.')
+
+        if not isinstance(self.root_instrument, SR830):
+            raise ValueError('Invalid parent instrument. ChannelBuffer '
+                             'can only live on an SR830.')
+
+        self.channel = channel
+        self.update_unit()
+
+    def update_unit(self):
```
jenshnielsen, line 40: Should type this as `-> None:` since it does not return any thing

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -432,6 +474,55 @@ def parse_offset_get(s: str) -> Tuple[float, int]:
...
+                               set_cmd=partial(self._set_ch_display, ch),
+                               vals=Strings())
+            self.add_parameter(f'ch{ch}_databuffer',
+                               channel=ch,
+                               parameter_class=ChannelBuffer)
+            self.add_parameter(f'ch{ch}_datatrace',
+                               channel=ch,
+                               vals=Arrays(shape=(self.buffer_npts.get,)),
+                               setpoints=(self.sweep_setpoints,),
+                               parameter_class=ChannelTrace)
```
jenshnielsen, line 524: Perhaps we should consider if it makes sense to replace this with a InstrumentChannel for each channel. That will be a breaking change but it would bring the instrument more in line with Other instruments. @astafan8 what do you think

astafan8: yes, i had the same feeling when reading the PR. we could leave this driver as is, and make another one next to it in a file called SR830_channels.py, similar to some other drivers.

RasmusBC59:
@astafan8 yes. If you use a trigger, the unit of the setpoints should be the unit of the parameter you are sweeping, else the unit is time

qcodes/instrument_drivers/stanford_research/SR830.py
RasmusBC59: Yes of course.
I have 3 mypy erros  I don't understand:
qcodes\instrument_drivers\stanford_research\SR830.py:18:5: error: Function is missing a type annotation for one or more arguments
qcodes\instrument_drivers\stanford_research\SR830.py:754:9: error: "_BaseParameter" has no attribute "update_unit"
qcodes\instrument_drivers\stanford_research\SR830.py:848:5: error: Function is missing a type annotation for one or more arguments

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -3,13 +3,73 @@
...
 from typing import Tuple
 import time
 
 
+class ChannelTrace(ParameterWithSetpoints):
+    """
+    Parameter class for the two channel buffers
+    """
+
+    def __init__(self, name: str, channel: int, **kwargs) -> None:
```
jenshnielsen, line 18: To fix the mypy error here you should type **kwargs as `Any`

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -721,3 +816,59 @@ def autorange_once() -> bool:
...
+        elif sweep_param.label is not None:
+            self.sweep_setpoints.label = sweep_param.label
+
+
+class GeneratedSetPoints(Parameter):
+    """
+    A parameter that generates a setpoint array from start, stop and num points
+    parameters.
+    """
+    def __init__(self, startparam, stopparam, numpointsparam, *args, **kwargs):
```
jenshnielsen, line 847:
```suggestion
    def __init__(self, startparam: Parameter, stopparam: Parameter, numpointsparam: Parameter, *args: Any, **kwargs: Any):
```

jenshnielsen: you probably also need to import Any from typing

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -656,6 +747,10 @@ def _set_ch_display(self, channel: int, disp: str) -> None:
         ratio_val = int(self.ask(f'DDEF ? {channel}').split(',')[1])
         self.write(f'DDEF {channel}, {disp_int}, {ratio_val}')
         self._buffer_ready = False
+        # we update the unit of the datatrace
+        # according to the choice of channel
+        params = self.parameters
+        params[f'ch{channel}_datatrace'].update_unit()
```
jenshnielsen, line 753:
```suggestion
        dataparam = self.parameters[f'ch{channel}_datatrace']
        assert isinstance(dataparam,  GeneratedSetPoints)
        dataparam.update_unit()
```

Mypy cannot see that this parameter is a GeneratedSetPoints parameter so we can help it by asserting that as shown here

RasmusBC59: thanks it worked (replacing GeneratedSetPoints by ChannelTrace)
I was confused because I had tryed:
        params = self.parameters
        assert isinstance(params[f'ch{channel}_datatrace'], ChannelTrace)
        params[f'ch{channel}_datatrace'].update_unit()
which did not work.

Now I am left with mypy errors for the init function of GeneratedSetPoint and ChannelTrace

jenshnielsen: Yes you have to make that a variable before using the assert. Mypy will not look through the indirection of the dict lookup

qcodes/instrument_drivers/stanford_research/SR830.py
RasmusBC59: done thanks

qcodes/instrument_drivers/stanford_research/SR830.py
RasmusBC59: Should I start making a version using Channels?
The use of channels is only relevant if you write to the buffer, so I would prefer to call it SR830_buffered.py 
Are you sure we want to have two drivers for the same instrument?
It will create some confusion for the user of which one to use. 
Could we add it as a non-breaking change, but with duplicated functionality?  
If you want we could have a short call discussing the best strategy?

RasmusBC59: Hi @jenshnielsen  and @astafan8 have you concluded anything regarding if I should change the current driver to include channels or I should make a new one?

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -1,15 +1,76 @@
...
+        params = self.root_instrument.parameters
+        if params[f'ch{self.channel}_ratio'].get() != 'none':
+            self.unit = '%'
+        else:
+            disp = params[f'ch{self.channel}_display'].get()
+            if disp == 'Phase':
+                self.unit = 'deg'
+            else:
+                self.unit = 'V'
+                self.label = self.root_instrument.name + '_' + disp
```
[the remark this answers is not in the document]
RasmusBC59: > Using the instrument in the label
The Scientist often uses more, than one SR830 at a time. Thus, having the instrument  as part of  the label helps the user to instantly decode the plot. It was after a request from the people I was helping, but maybe it is to focused one there use case, and it would be better to just give people the change to supply a custom label.

> should the channel number be also part of the label:
The channel can be inferred from “disp”.  The instrument does not really have channels but two displays, so including channels in the label would be confusing for someone who have not read the driver.  The two displays can not show the same output.

qcodes/instrument_drivers/stanford_research/SR830.py
jenshnielsen: I think that is fine for now. It would be relative little effort to copy and convert this to a channel version

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -721,3 +819,69 @@ def autorange_once() -> bool:
...
+    """
+    def __init__(self, startparam: 'Parameter', stopparam: 'Parameter',
+                 numpointsparam: 'Parameter', *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self._startparam = startparam
+        self._stopparam = stopparam
+        self._numpointsparam = numpointsparam
+        self.update_units_if_time()
+
+    def update_units_if_time(self) -> None:
```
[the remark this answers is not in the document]
RasmusBC59: I agree

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -721,3 +819,69 @@ def autorange_once() -> bool:
...
+        assert isinstance(self.root_instrument, SR830)
+        SR = self.root_instrument.buffer_SR.get()
+        if SR != 'Trigger':
+            self.unit = 's'
+            self.label = 'Time'
+
+    def get_raw(self) -> ParamRawDataType:
+        assert isinstance(self.root_instrument, SR830)
+        SR = self.root_instrument.buffer_SR.get()
+        N = self.root_instrument.buffer_npts.get()
```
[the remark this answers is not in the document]
RasmusBC59: Yes I agree

qcodes/instrument_drivers/stanford_research/SR830.py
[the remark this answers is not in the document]
astafan8: thanks for the explanation! and i support this:

> it would be better to just give people the change to supply a custom label

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -721,3 +819,69 @@ def autorange_once() -> bool:
...
+        if SR == 'Trigger':
```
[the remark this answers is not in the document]
RasmusBC59: I think I it is the most frequent case, but I see you point. This was also what I was hinting in the pull request description. “Do we need all the sweep parameters”.
 I think I will make a more slim and flexible version. Then we can always go back.

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -1,15 +1,76 @@
...
+        if N == 0:
+            raise ValueError('No points stored in SR830 data buffer.'
+                             ' Can not poll anything.')
+
+        # poll raw binary data
+        self.root_instrument.write(f'TRCL ? {self.channel}, 0, {N}')
+        rawdata = self.root_instrument.visa_handle.read_raw()
+
+        # parse it
+        realdata = np.fromstring(rawdata, dtype='<i2')
```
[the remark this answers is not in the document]
RasmusBC59: I don’t have any strong feeling about this; if you thing the query_binary_data method  I will try it out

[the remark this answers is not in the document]
astafan8: nah, if this works, then let's not bother. it works and the code is clear. my comment was more to mention the query method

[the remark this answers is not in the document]
jenshnielsen: read_raw will only read up to a given chunk size so using `query_binary_values` is definitely less error prone. In this example it may not matter but as I understand it `read_raw` really never should be used in production code since it disables many of the safeties of the more high level functions

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -721,3 +798,56 @@ def autorange_once() -> bool:
...
+        update the unit to "s" and label to "Time";
+        otherwise do nothing
+        """
+        assert isinstance(self.root_instrument, SR830)
+        SR = self.root_instrument.buffer_SR.get()
+        if SR != 'Trigger':
+            self.unit = 's'
+            self.label = 'Time'
+
+    def set_raw(self, value: Iterable[Union[float, int]]) -> None:
```
jenshnielsen, line 841: There is no need to use `Union[float,int]` that is fully equivalent to `float` as int is considered a subtype of float by mypy

jenshnielsen (approved):

astafan8 (approved):

PR #2910, opened 2021-04-09
Significant performance improvement in Tektronix5014 driver

sldesnoo-Delft:
- For the conversion of 4 waveforms with 1e6 Sa the time is reduced from 890 ms to 310 ms.
- The change in _pack_waveform gives a reduction of 180 ms. 
- The change in _pack_record gives a reduction of 400 ms.


qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -1683,11 +1688,10 @@ def _pack_waveform(
...
                             ' Only 0 and 1 are allowed')
 
-        wflen = len(wf)
-        packed_wf = np.zeros(wflen, dtype=np.uint16)
-        packed_wf += np.uint16(np.round(wf * 8191) + 8191 +
-                               np.round(16384 * m1) +
-                               np.round(32768 * m2))
+        # Note: np.trunc is an order of magnitude faster then np.round
+        packed_wf = np.trunc(16384 * m1 + 32768 * m2
+                             + wf * 8191 + 8191.5).astype(np.uint16)
```
astafan8, line 1693: is it correct that 8191 changed to 8191.5?

qcodes/instrument_drivers/tektronix/AWG5014.py
```diff
@@ -1683,11 +1688,10 @@ def _pack_waveform(
             raise TypeError('Marker 2 contains invalid values.' +
                             ' Only 0 and 1 are allowed')
 
-        wflen = len(wf)
-        packed_wf = np.zeros(wflen, dtype=np.uint16)
-        packed_wf += np.uint16(np.round(wf * 8191) + 8191 +
-                               np.round(16384 * m1) +
-                               np.round(32768 * m2))
+        # Note: np.trunc is an order of magnitude faster then np.round
```
astafan8, line 1691: it seems to me that this note is only relevant in the context of the diff, and if that is true, shall we remove it?

jenshnielsen: I think it's fine to keep. My instinct would be to use round here so an explanation is nice. 

Perhaps we could rephrase it a bit to say. 
```
# Note: we use np.trunc here rather than np.round as it is an order of magnitude faster
```

qcodes/instrument_drivers/tektronix/AWG5014.py
sldesnoo-Delft: It is 8191.5, because np.round has been replaced by np.trunc. For positive numbers this results in the same rounding with a very small exception. Python and numpy round half towards the nearest even value. So, 1.5 becomes 2 and 2.5 also becomes 2. That's what makes the Python and numpy implementations slow.
Added 0.5 and truncating rounds the half to the higher integer. So, 1.5 becomes 2 and 2.5 becomes 3.

astafan8: thank you for explaining!

qcodes/instrument_drivers/tektronix/AWG5014.py
astafan8: done

astafan8 (approved):
@jenshnielsen i think this looks good now, shall we merge?

PR #2918, opened 2021-04-12
plot_dataset avoid loading from db if in cache

jenshnielsen:
This should both provide a speed boost for plotting data live and and make plot_dataset work for data not stored in the db


astafan8 (approved):

qcodes/dataset/data_export.py
```diff
@@ -69,11 +69,16 @@ def get_data_by_id(run_id: int) -> \
...
+
 
+def _get_data_from_ds(ds: DataSet) -> List[List[Dict[str, Union[str, np.ndarray]]]]:
     dependent_parameters: Tuple[ParamSpecBase, ...] = ds.dependent_parameters
 
-    parameter_data = ds.get_parameter_data(
-        *[ps.name for ps in dependent_parameters])
+    all_data = ds.cache.data()
+
+    parameter_data = {ps.name: all_data[ps.name] for ps in dependent_parameters}
```
jenshnielsen, line 81: This is not the same behavior when `dependent_parameters == ()` as get_parameter_data treats `()` as `None` which means all parameters

jenshnielsen:
@astafan8 

Two issues were discovered. 

* When dependent parameters was empty all data was loaded. I have restored this behavior but perhaps we should fix the tests instead? 
* The measurement context manager tests for `get_data_by_id` implicitly relies on the order of keys in the output of get_parameter_data. This is not guaranteed to be preserved in the cache. Rather than fixing these tests I have removed them since arguably they have nothing to do with the measurement context manager

astafan8 (approved):

PR #2927, opened 2021-04-14
Add support for Keysight 34411A

janekfleper:
-Correctly implement the Keysight 3441XA series. A few settings for the 34410A were wrong/missing and the 34411A was not considered at all
-Set the correct DC voltage ranges for all models supported by the _Keysight_344xxA class


qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -475,15 +483,10 @@ def __init__(self, name: str, address: str, silent: bool = False,
             PLCs['34465A'] = [0.001, 0.002, 0.006] + PLCs['34465A']
             PLCs['34470A'] = [0.001, 0.002, 0.006] + PLCs['34470A']
 
-        ranges = {'34410A': [10**n for n in range(3, 10)],  # 100 to 1 G
-                  '34460A': [10**n for n in range(-3, 9)],  # 1 m to 100 M
-                  '34461A': [10**n for n in range(-3, 9)],  # 1 m to 100 M
-                  '34465A': [10**n for n in range(-3, 10)],  # 1 m to 1 G
-                  '34470A': [10**n for n in range(-3, 10)],  # 1 m to 1 G
-                  }
```
jenshnielsen, line 483: Were these ranges incorrect?

janekfleper: The list `self.ranges` was only used for the DC voltage range parameter, and in that context they were incorrect. All DMMs share the same voltage ranges, 100mV to 1kV.
Given that 34460A/34461A end at 100M and 34465A/34470A end at 1G would suggest that the range was (also) meant for the resistance measurement? The lower limit on the other hand fits neither the resistance nor the voltage. Maybe someone tried to define the absolute ranges over all possible measurements?

jenshnielsen: Thanks, I agree lets just remove them

jenshnielsen:
@astafan8 I think this looks fine to merge what do you think?

astafan8 (approved):

Issue #2924, opened 2021-04-14
Station should be able to load multiple yaml files


guenp:
Currently, `Station.load_config_file` can only load a single YAML file. Even when running said method a second time on a different yaml file, it does not append the data from the new file and keeps the original yaml file in memory. For systems with many instruments and parameters it would be very useful if instead of a single file, `Station.load_config_file` would be able to load multiple yaml files and combine them.

Example usage as proposed by @samantha-ho:

```python
station.load_config_file([foo.yaml, bar.yaml, baz.yaml])
```

CC @jenshnielsen, as discussed in our meeting last week.

FarBo:
@guenp 
Thank you for openning this issue.
We will add the suggested feature to qcodes :)

PR #2928, opened 2021-04-15
DelegateInstrument and InstrumentGroup drivers

guenp:
In use cases where an experiment requires using the same instruments for several devices on a chip, it would be useful to be able to specify an "alias" for one or more instrument parameters and map them to abstract device terminals that can be easily swapped in the instrument YAML. Likewise, it can be useful to specify different `setter` or `getter` methods depending on how an instrument is used, and specify this in the instrument yaml rather than having to write a custom instrument driver.

To give an example for the latter, one might want to use the driver for a superconducting magnet, and instead of blocking execution during the ramp time, by default ramp the magnet with `set_field(value, block=False)`.

Example usage:
```yaml
instruments:
  field_X:
    type: qcodes.instrument_drivers.american_magnetics.AMI430.AMI430
    address: 192.168.0.1
    init:
      port: 31337

  field:
    type: qcodes.instruments.delegate.DelegateInstrument
    init:
      parameters:
        X:
          - field_X.field
        ramp_rate:
          - field_X.ramp_rate
      set_defaults_on_load: true
      default_values:
        ramp_rate: 0.1
      setters:
        X:
          method: field_X.set_field
          block: false
```

Here, `field.ramp_rate` would be set to `0.1 T/min`, `field.X()` would be an alias to `field_X.field()`, `field.X(1.)` would be an alias to `field_X.set_field(1., block=False)`.

This PR contains the following instrument drivers:
- `MetaInstrument`: class for creating a meta instrument that aliases one or more parameter endpoints from real instruments.
- `MetaChannelInstrument`: Meta instrument that auto generates aliases for a given ChannelList
- `Chip`: Meta instrument for a Chip that contains several `NanoDevice` submodules
- `NanoDevice`: Meta instrument for a quantum device on a chip

Open questions:
- Would it be useful for the meta instruments be specified in the same yaml file or a separate one to enforce loading order? Currently, loading meta instruments requires one to pass the station; it would perhaps be more intuitive to be able to simply run `station.load_chip()`, and have `station` as an optional input argument. (See also: #2924)
- How to deal with devices that are connected to instruments via a software-controlled switch? Currently, `NanoDeviceParameter.channel` is optional (in the UTs I used wildcard names but these can be set to channel names for routing purposes).

CC @jenshnielsen @astafan8


jenshnielsen:
@guenp Sorry I did not find time to look at this today. I will have a look tomorrow

guenp:
@jenshnielsen  no worries, the PR is quite large so am happy to split it up if that makes reviewing a bit easier

qcodes/instrument/meta/meta_parameter.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+    def __init__(
+        self,
+        name: str,
+        endpoints: Tuple[Parameter],
+        endpoint_names: Tuple[str],
+        unit: str = None,
+        setter: callable = None,
+        **kwargs
+    ):
```
jenshnielsen, line 57: Why does the metaparameter not expose a label?  In this case label will simply be the same as name if there is nothing supplied. But this means that it is not possible to git it a nicer formatted label. 

name should be without spaces etc for use in generation of identifiers but label can be formatted with spaces etc

qcodes/dataset/measurements.py
```diff
@@ -773,6 +774,12 @@ def register_parameter(
                                      parameter.unit,
                                      setpoints,
                                      basis, paramtype)
+        elif isinstance(parameter, MetaParameter):
+            self._register_parameter(parameter.full_name,
+                                     parameter.name,
```
jenshnielsen, line 779: Perhaps we should give MetaParameter a label to match regular parameters

requirements.txt
```diff
@@ -74,6 +74,7 @@ ruamel.yaml.clib~=0.2.2
 schema~=0.7.4
 Send2Trash~=1.5.0
 six~=1.15.0
+scipy~=1.6.2
```
jenshnielsen, line 77: Why Scipy here. QCoDeS does not depend on scipy directly (as stated in setup.cfg) I would prefer to keep this that way since scipy is a somewhat big dependency

qcodes/instrument/meta/nano_device.py
```diff
@@ -0,0 +1,152 @@
...
+        Args:
+            name: Name of sub parameter
+            endpoint: Endpoint parameter to connect to
+
+        Returns:
+            Parameter: The parameter instance
+        """
+        assert not hasattr(
+            self, name
+        ), f"Duplicate parameter name {name}."
```
jenshnielsen, line 62: It is generally a bad idea to throw a bare assertion. I think this should be a KeyError to match the standard parameter

qcodes/instrument/meta/meta_channel_instrument.py
```diff
@@ -0,0 +1,45 @@
+from typing import Dict, List, Any
+from qcodes.station import Station
+from qcodes.instrument.meta.meta_instrument import MetaInstrument
+
+
+class MetaChannelInstrument(MetaInstrument):
+    """
+    Meta instrument that auto generates aliases for a given ChannelList
```
jenshnielsen, line 8: Please expand this to clarify what is an alias and how it is meant to be used

qcodes/instrument/meta/meta_instrument.py
```diff
@@ -0,0 +1,185 @@
...
+import logging
+
+_log = logging.getLogger(__name__)
+
+
+class MetaInstrument(InstrumentBase):
+    """MetaInstrument class for creating a meta instrument that aliases one or more 
+    parameter endpoints from real instruments.
+
+    Example usage in instrument YAML:
```
jenshnielsen, line 19: This docstring could do with a bit more detail. It is hard to see from the current docstring how this is meant to be called

qcodes/instrument/meta/meta_instrument.py
```diff
@@ -0,0 +1,185 @@
...
+            method: field_X.set_field
+            block: false
+        units:
+            X: T
+            ramp_rate: T/min
+
+    Args:
+        name: Instrument name
+        station: Real instrument station to connect the endpoints to.
+        aliases: Aliases and the endpoints they connect to.
```
jenshnielsen, line 43: For example this should say something like. A mapping from name of an alias parameter to the sequence of endpoints that it connects to.

qcodes/instrument/meta/meta_instrument.py
```diff
@@ -0,0 +1,185 @@
...
+            station: Measurement station
+            path: Relative path to parse
+        """
+        def _parse_path(parent, elem):
+            child = getattr(parent, elem[0])
+            if len(elem) == 1:
+                return child
+            return _parse_path(child, elem[1:])
+
+        return _parse_path(station, path.split("."))
```
jenshnielsen, line 68: I find that putting the static method between the init method and the docstring significantly reduces readability since one has to scroll to compare the docstring with the signature

qcodes/instrument/meta/meta_instrument.py
```diff
@@ -0,0 +1,185 @@
...
+    def _endpoint_names(endpoints: List[Parameter]):
+        """Get the endpoint names"""
+        endpoint_names = [_e.name for _e in endpoints]
+        if len(endpoint_names) != len(set(endpoint_names)):
+            endpoint_names = [
+                f"{_e}{n}" for n, _e in enumerate(endpoint_names)
+            ]
+        return endpoint_names
+
+    def _add_parameter(
```
jenshnielsen, line 147: perhaps `_create_and_add_parameter` to clarify what more than add_parameter it does

qcodes/instrument/meta/meta_parameter.py
jenshnielsen: Label is part of the Parameter interface but not the `_Baseparameter` interface but I think it makes sense to add it here

qcodes/instrument/meta/meta_parameter.py
```diff
@@ -0,0 +1,174 @@
...
+            self, name
+        ), f"Duplicate parameter name {name}."
+        parameter = cls(
+            name=name,
+            endpoints=(endpoint,),
+            endpoint_names=(endpoint.name,),
+            instrument=None
+        )
+        setattr(self, name, parameter)
+        return parameter
```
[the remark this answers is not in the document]
jenshnielsen: Alternatively this should use the Metadatable interface to register parameters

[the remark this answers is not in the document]
jenshnielsen: This happens since Parameter was never intended to have sub components in the first place. I unfortunately think that using parameters for this purpose will result in more issues down the line but we can try it out now and see what works out

qcodes/instrument/meta/meta_parameter.py
```diff
@@ -0,0 +1,174 @@
...
+            return self._setter(value)
+        elif self.group is not None:
+            self.group.set_parameters(value)
+        else:
+            return self._endpoints[0](value)
+    
+    def _set_override(self, value: ParamDataType = None, **kwargs):
+        """Overide set method such that values can optionally be set using only kwargs"""
+        if value is None:
+            self._set(kwargs)
```
jenshnielsen, line 156: Using kwargs without expanding them looks odd and is not general something that I would expect to see but perhaps it makes sense here?

jenshnielsen:
Mostly I think it looks like a useful addition and now is a good time to start adding something like this to qcodes/

There are some api questions that we will likely have to answer down the line but perhaps
it makes sense to merge this model as is and learn from its usage 

* To which extend does the MetaParameter interface enable us to map device to instrument parameters that require some form of conversion. Scaling, offset transform etc
* Are parameters the right type for the thing that maps to multiple endpoints?

Smaller things to improve before merge:

* The docstrings are a bit more terse than I like. I think we should put some effort into how the usage of the individual classes could be done better.
* Think about naming. There is a lot of use of the word meta which in it self is not that descriptive
* There are a bunch of typing improvements that could be made.  I am happy to take care of those.
* We should have a worked through example

qcodes/instrument/meta/meta_parameter.py
```diff
@@ -0,0 +1,174 @@
...
+        self.unit = unit if unit is not None else ''
+        endpoint_names = endpoint_names or [_e.name for _e in endpoints]
+        self._group = None
+
+        # Generate sub parameters if >1 endpoints,
+        # e.g. my_meta_param.X, my_meta_param.Y
+        if len(endpoints) > 1:
+            self._parameters = self._sub_parameters(endpoints, endpoint_names)
+        self._set = self.set
+        self.set = self._set_override
```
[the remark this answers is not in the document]
guenp: Yeah if I could just override `.set()` I would have but unfortunately that's not possible because of how this is implemented:
https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument/parameter.py#L329

I suggest we just change that, and implement `set` as a method on `Parameter`. I a not sure if that would break anything else though.

qcodes/instrument/meta/meta_parameter.py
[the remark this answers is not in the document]
guenp: That is a fair point, I think it would be more consistent to create a new parameter called e.g. `lockin_frequency` rather than `lockin.frequency`.

qcodes/instrument/meta/meta_parameter.py
```diff
@@ -0,0 +1,174 @@
...
+        return self._namedtuple(*[_p.get() for _p in self.parameters.values()])
+
+    def _set_from_dict(self, calling_dict: Dict[str, ParamRawDataType]) -> None:
+        for name, p in list(self.parameters.items()):
+            p.set(calling_dict[name])
+
+
+class MetaParameter(_BaseParameter):
+    """
+    Meta parameter that returns one or more endpoint parameter values
```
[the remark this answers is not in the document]
guenp: I'm fine with consolidating the two in this PR.

qcodes/instrument/meta/meta_parameter.py
guenp: Sounds good, thanks for the suggestion.

requirements.txt
guenp: I used `scipy` to do interpolation, but since it's just interpolation between two datapoints I suppose I could just use `numpy` for that.

qcodes/instrument/meta/meta_parameter.py
guenp: Here it's mean to support:
`my_group_param(foo="bar", spam="eggs")`
which then calls
`my_group._set_from_dict({"foo": "bar", "spam": "eggs"})`

requirements.txt
jenshnielsen: But that is only in the tests right ?

guenp: Yes that's right, I've removed it from here but left the dependency in test_requirements.txt.

qcodes/instrument/meta/meta_parameter.py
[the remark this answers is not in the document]
guenp: @jenshnielsen I renamed this to `DelegateGroupParameter` and it's now both using `Group` and `DelegateParameter`. Everything else is essentially the same... please take a look and let me know if this makes sense.

qcodes/instrument/meta/meta_parameter.py
```diff
@@ -0,0 +1,174 @@
...
+        parameters: Tuple['MetaParameter'],
+        parameter_names: Tuple[str] = None,
+        **kwargs
+    ):
+        super().__init__(parameters=list(parameters), single_instrument=False, **kwargs)
+        parameter_names = parameter_names or [_e.name for _e in parameters]
+        self._namedtuple = namedtuple(name, parameter_names)
+
+    def get_parameters(self) -> Any:
+        return self._namedtuple(*[_p.get() for _p in self.parameters.values()])
```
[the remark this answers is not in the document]
guenp: Not sure I understand the question. This returns a `namedtuple`, where the values are the sub parameter values. Could you clarify what you mean with metagroup?

guenp:
@jenshnielsen , thanks for your review.

> To which extend does the MetaParameter interface enable us to map device to instrument parameters that require some form of conversion. Scaling, offset transform etc

Right now it does not, but I like the idea of being able to specify some kind of template/equation and subbing those with values from one or more delegate parameters.

> Are parameters the right type for the thing that maps to multiple endpoints?

Perhaps not. I guess the question is, why would you want to group parameters in the first place? The main reason I wanted to group them is because that way I can pass it to a function that creates a measurement, and automatically gets the endpoint ("source") parameters and measures all of them as a function of the sweep value. So for that purpose we don't need it to be a parameter. But if we go for the above feature (e.g. be able to specify a linear combination of delegate parameters) such that the .get() method returns a single value then I think it does make sense to make it a parameter. However as @astafan8 suggested I think it will be good to keep things as simple as possible at this stage.

> The docstrings are a bit more terse than I like. I think we should put some effort into how the usage of the individual classes could be done better.

I've improved some of the docstrings but kept the ones on the methods that I think we still need to discuss open for now. Same for the example notebooks. If it's okay with you I'd like to first resolve the above discussion before writing more detailed docstrings.

> Think about naming. There is a lot of use of the word meta which in it self is not that descriptive

Agreed. I renamed things to `DelegateInstrument`,  `Chip` and `Device`, I hope that is more clear

> There are a bunch of typing improvements that could be made. I am happy to take care of those.

Tak for det! I'm also happy to take a stab at that once we nail down the design.

> We should have a worked through example

I added a couple notebooks to start, I will expand those/add more explanation to them once we nail down some of the details on the Meta/DelegateGroupParameter.

guenp:
@jenshnielsen @trevormorgan @astafan8 Based on feedback so far I've made a bunch of changes to this draft, so let me explain a bit what the main concepts here to make it easier to review.

The main goal is to create a new instrument driver that makes it possible to create delegate parameters on-the-fly. A delegate parameter is an instrument parameter that "delegates" to another target instrument parameter, ie. it is fully virtual and its only job is to defer "get" and "set" requests to another instrument parameter.

This PR adds an instrument driver `DelegateInstrument` that lets the user specify custom parameters that point to the desired target instrument parameters. To support being able to create a parameter that points to more than one target instrument parameter, a few new concepts are introduced:
* `DelegateGroup`: A subclass of `Group` that groups together one or more `DelegateParameter`s. For each target instrument parameter, a `DelegateParameter` is created and added as a parameter to the `DelegateInstrument` instance with the name `<group name>_<parameter name>`. To address all of them in tandem, the `DelegateGroup` has a `.set` and `.get` method that can optionally be customized by passing a function to the `setter` and/or `getter` argument in the constructor.
* `GroupedParameter`: Whereas the `DelegateGroup` represents the group of `DelegateParameter`s, the `GroupParameter` is a new parameter introduced to be able to address the group as a single parameter, such that the newly created instrument will have a parameter that can be used in measurements. The formatting of the output can be customized through the `formatter` input argument to the constructor.

Additionally, there is the following class:
* `InstrumentGroup`: This is an instrument driver with the sole purpose of creating one or more subinstruments/submodules of a given class. This can be used to represent a "Chip" with multiple "Devices" (`DelegateInstrument`s)


For examples, please review the below notebooks:
* [QCodes example with DelegateInstrument.ipynb](https://github.com/QCoDeS/Qcodes/blob/64680037a0faf9614dfc1839dd7b8168cffafd8e/docs/examples/driver_examples/QCoDeS%20example%20with%20DelegateInstrument.ipynb)
* [https://github.com/QCoDeS/Qcodes/blob/64680037a0faf9614dfc1839dd7b8168cffafd8e/docs/examples/driver_examples/QCoDeS%20example%20with%20InstrumentGroup%20and%20DelegateInstrument.ipynb](QCodes example with InstrumentGroup)

jenshnielsen:
Thanks @guenp I did not get as far looking at this today as I had hoped. I will have a look tomorrow. Did you see that I opened a pr with some typing changes against your prs branch ?

guenp:
@jenshnielsen Thank you!

guenp:
@jenshnielsen I tried to resolve the typing errors (thanks for fixing a large amount of them!), the only ones I got stuck on were these:
```python
$ mypy qcodes
qcodes/instrument/group_parameter.py:306:16: error: Incompatible return value type (got "OrderedDict[str, GroupParameter]", expected "OrderedDict[str, DelegateParameter]")
qcodes/instrument/delegate/grouped_parameter.py:82:24: error: Argument "parameters" to "__init__" of "Group" has incompatible type "Sequence[DelegateParameter]"; expected "Sequence[GroupParameter]"
qcodes/instrument/delegate/grouped_parameter.py:93:67: error: Argument 2 to "zip" has incompatible type "Sequence[DelegateParameter]"; expected "Iterable[GroupParameter]"
qcodes/instrument/delegate/delegate_instrument.py:210:24: error: Argument "parameters" to "DelegateGroup" has incompatible type "List[_BaseParameter]"; expected "Sequence[DelegateParameter]"
Found 4 errors in 3 files (checked 440 source files)
```

What's your suggested solution for these? I don't think it's a good idea to make `DelegateParameter` a child class of `GroupParameter`.

guenp:
@jenshnielsen I was able to mitigate the mypy errors, would you mind taking another pass at reviewing? Thanks!

jenshnielsen:
@guenp Thanks, I will have a look. I had proposed a different solution in https://github.com/guenp/Qcodes/pull/3 but I think this looks fine too. I have made a pr https://github.com/guenp/Qcodes/pull/4 to add this module to the api docs and fix some .rst (restructured text) issues so it formats correctly

guenp:
Thanks @jenshnielsen ! Sorry I hadn't seen your other PR guenp#3. I've merged guenp#4.

qcodes/instrument/delegate/delegate_instrument.py
```diff
@@ -0,0 +1,251 @@
...
+    .. code-block:: python
+
+        field_X.set_field(value, block=False),
+
+    as opposed to ``field.X.field.set()`` which ramps with ``block=True``.
+
+    Args:
+        name: Instrument name
+        station: Real instrument station that is used to get the endpoint
+            parameters.
```
jenshnielsen, line 72:
```suggestion
        station: Station containing the real instrument that is used to get the endpoint
            parameters. 
```

perhaps?

qcodes/instrument/delegate/delegate_instrument.py
```diff
@@ -0,0 +1,251 @@
...
+        parameters: A mapping from the name of a parameter to the sequence
+            of source parameters that it points to.
+        initial_values: Default values to set on the delegate instrument's
+            parameters. Defaults to None.
```
jenshnielsen, line 76: Should this explain that None is equivalent to not setting anything?

qcodes/instrument/delegate/grouped_parameter.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+class DelegateGroupParameter(DelegateParameter, GroupParameter):
+
+    def __init__(
+        self,
+        name: str,
+        source: Optional[Parameter],
+        instrument: Optional['InstrumentBase'] = None,
+        initial_value: Union[float, int, str, None] = None
```
jenshnielsen, line 37: You can remove int from this union as int is considered a subtype of float

jenshnielsen:
@guenp Left a few minor comment otherwise I think this is almost ready to merge. @Dominik-Vogel had a point about extending the schema validation for the station yaml files but I am not sure about the details

Dominik-Vogel:
Dear @guenp ,
I like the general idea of that PR and I have been long thinking about something similar when I originally implemented the station yaml file, but hadn't found such a convincing solution. (I like the idea of consolidating with the DelegateParameter). I have only one point to add:
There is a yaml schema for the station-yaml files, which needs updating with your additions. The schema can become quite useful if you use it with a language server that supports that, like VScode with a plugin. See here in the qcodes documentation:
https://qcodes.github.io/Qcodes/examples/Station.html#Setting-up-Visual-Studio-Code-to-work-with-station-files
The schema file is here:
https://github.com/QCoDeS/Qcodes/blob/master/qcodes/dist/schemas/station-template.schema.json

qcodes/instrument/delegate/delegate_instrument.py
```diff
@@ -0,0 +1,251 @@
...
+        def _parse_path(parent: Any, elem: Sequence[str]) -> Any:
+            child = getattr(parent, elem[0])
+            if len(elem) == 1:
+                return child
+            return _parse_path(child, elem[1:])
+
+        return _parse_path(parent, path.split("."))
+
+    def set_initial_values(self, dry_run: bool = False) -> None:
+        """Set parameter initial values on meta instrument
```
trevormorgan, line 127:
```suggestion
        """Set parameter initial values on delegate instrument
```

qcodes/instrument/delegate/delegate_instrument.py
```diff
@@ -0,0 +1,251 @@
...
+        group_name: str,
+        station: Station,
+        paths: List[str],
+        setter: Optional[Dict[str, Any]] = None,
+        getter: Optional[Callable[..., Any]] = None,
+        formatter: Optional[Callable[..., Any]] = None,
+        unit: Optional[str] = None,
+        **kwargs: Any
+    ) -> None:
+        """Create meta parameter that links to a given set of paths
```
trevormorgan, line 213:
```suggestion
        """Create delegate parameter that links to a given set of paths
```

qcodes/instrument/delegate/grouped_parameter.py
```diff
@@ -0,0 +1,190 @@
...
+    Parameter,
+    _BaseParameter
+)
+
+import logging
+
+_log = logging.getLogger(__name__)
+
+
+class DelegateGroup(Group):
```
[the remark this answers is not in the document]
trevormorgan: Actually ignore this comment, I think its fine.

trevormorgan (approved):
@guenp just had some minor comments where I saw `meta` being used in some doc strings. PR looks good and I think we can merge when you're ready.

qcodes/tests/instrument_mocks.py
```diff
@@ -698,3 +699,185 @@ def snapshot_base(self, update: Optional[bool] = True,
...
+        """
+        while True:
+            _time = yield
+            if _time is None:
+                _time = 0.0
+
+            yield float(self._field_ramp_fcn(_time))
+
+
+class MockLockin(Instrument):
```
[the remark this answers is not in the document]
trevormorgan: This seems fine to me, don't worry about trying to use pyvisa-sim,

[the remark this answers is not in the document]
jenshnielsen: Yes, I think pyvisa-sim makes sense for a real instrument where you want to test that instruments visa api. Here I think we want to test something slightly different where it is easier to not mock a real instrument

guenp:
@jenshnielsen, @trevormorgan, @Dominik-Vogel thanks for taking a look!
I have read the docs and template but I don't think I need to add anything; `DelegateInstrument` is just a new instrument driver so all of the arguments are added under `init`. Please correct me if I'm wrong though!

PR #2930, opened 2021-04-15
Added the functionality of getting guids from multiple run ids and directly from database

FarBo:
This PR:
Enbale users to pass databse and an iterable of run ids in the function and get a list of guids for those ids.
It uses get_guid_from_run_id to retrieve the guids and run_exists to check run ids exist in the database and raise Error if not, and tell the user which run ids are not available.

As a side improvement: 
Added the return doc to get_guid_from_run_id

- [x] Add a test file


qcodes/dataset/sqlite/queries.py
```diff
@@ -829,10 +831,41 @@ def get_guid_from_run_id(conn: ConnectionPlus, run_id: int) -> str:
...
+    Args:
+        db_path: The path to the database file.
+        run_ids: An integer iterable of run ids to get their guids.
+
+    Returns:
+        A list of guids belonging to the inserted run_ids.
+    """
+
+    guids: List[str] = []
+    for run_id in run_ids:
```
jenshnielsen, line 857:
```suggestion
    with connect(db_path) as conn:
        for run_id in run_ids:
            if run_exists(conn=conn, run_id=run_id):
            run_id_guid = get_guid_from_run_id(conn=conn,
                                               run_id=run_id)
            guids.append(run_id_guid)
        else:
            raise RuntimeError(f'run id {run_id} and above are not'
                               ' available in the database')
```

FarBo: ValueError                                Traceback (most recent call last)
<ipython-input-2-6a39f4a81d5d> in <module>
----> 1 get_guid_from_multiple_run_ids(db_path='doNd.db', run_ids=[1, 2, 3, 4, 5])

c:\users\a-fbonabi\qcodes\qcodes\dataset\sqlite\queries.py in get_guid_from_multiple_run_ids(db_path, run_ids)
    858     for run_id in run_ids:
    859         with connect(db_path) as conn:
--> 860             if run_exists(conn=conn, run_id=run_id):
    861                 run_id_guid = get_guid_from_run_id(conn=conn,
    862                                                     run_id=run_id)

c:\users\a-fbonabi\qcodes\qcodes\dataset\sqlite\queries.py in run_exists(conn, run_id)
   1064     );
   1065     """
-> 1066     res: sqlite3.Row = atomic_transaction(conn, query, run_id).fetchone()
   1067     return bool(res[0])
   1068 

c:\users\a-fbonabi\qcodes\qcodes\dataset\sqlite\connection.py in atomic_transaction(conn, sql, *args)
    156 
    157     """
--> 158     with atomic(conn) as atomic_conn:
    159         c = transaction(atomic_conn, sql, *args)
    160     return c

c:\users\a-fbonabi\anaconda3\envs\qcodes\lib\contextlib.py in __enter__(self)
    111         del self.args, self.kwds, self.func
    112         try:
--> 113             return next(self.gen)
    114         except StopIteration:
    115             raise RuntimeError("generator didn't yield") from None

c:\users\a-fbonabi\qcodes\qcodes\dataset\sqlite\connection.py in atomic(conn)
     82     with DelayedKeyboardInterrupt():
     83         if not isinstance(conn, ConnectionPlus):
---> 84             raise ValueError('atomic context manager only accepts '
     85                              'ConnectionPlus database connection objects.')
     86 

ValueError: atomic context manager only accepts ConnectionPlus database connection objects.

jenshnielsen: Ok that is unfortunate. If we cannot use a context manager we will have to rely on a try finally clause to handle this. 
But in any case we should both ensure that we only open one connection and make sure that it is probably closed at the end of the function. Lets also write a test for this function so we are sure it works correctly

```
try:
   conn =connnect(dbpath)
   rest of the code here



finally:
    conn.close()
```

FarBo: Yes. We should do this.

qcodes/dataset/sqlite/queries.py
```diff
@@ -829,10 +831,47 @@ def get_guid_from_run_id(conn: ConnectionPlus, run_id: int) -> str:
...
+    """
+
+    guids: List[str] = []
+
+    try:
+        conn = connect(db_path)
+        for run_id in run_ids:
+            if run_exists(conn=conn, run_id=run_id):
+                run_id_guid = get_guid_from_run_id(conn=conn,
+                                                   run_id=run_id)
```
jenshnielsen, line 863: Perhaps we should use captured run_id ? and get_guid_from_run_spec` ?

jenshnielsen: @astafan8 what do you think?

qcodes/dataset/sqlite/queries.py
```diff
@@ -829,10 +831,47 @@ def get_guid_from_run_id(conn: ConnectionPlus, run_id: int) -> str:
...
+                guids.append(run_id_guid)
+            else:
+                raise RuntimeError(f'run id {run_id} and above do not'
+                                   ' exist in the database')
```
jenshnielsen, line 867: What does above mean in this context. I think `run id {run_id} and above do not exist in the database` would be fine

qcodes/dataset/sqlite/queries.py
```diff
@@ -829,10 +831,47 @@ def get_guid_from_run_id(conn: ConnectionPlus, run_id: int) -> str:
...
+    """
+    Expanding the functionality of get_guid_from_run_id to retrive guids
+    from database and for multiple run ids.
+
+    Args:
+        db_path: The path to the database file.
+        run_ids: An integer iterable of run ids to get their guids.
+
+    Returns:
+        A list of guids belonging to the inserted run_ids.
```
jenshnielsen, line 853: "A list of guids for the supplied run_ids"

qcodes/dataset/sqlite/queries.py
FarBo: It is an extra (maybe not so necessary) explanation to say if for example run_id=3 doesn't exist, more than 3 also not available. 

you mean this is fine, right?:
run id {run_id} does not exist in the database

jenshnielsen: Yes I think that's fine. Otherwise we could iterate through all of them to collect all the missing ones before raising but I don't think that is worth the effort.

```
missing = []
...
else:
    missing.append(run_id)
...
if len(missing) > 0
   raise RuntimeError(f"{missing} run_id(s) do not exist ...")
```

FarBo: Sure.  Also there might be a large database with a lot of runs that iterating over those is just wasting resources.

qcodes/dataset/sqlite/queries.py
astafan8: @FarBo please correct me if the following is wrong, but it seems to me that the users of this function have the following problem: they were working with run_ids (not captured_run_ids), and then they chose a set of run_ids because of what they want to do next, and they just need the GUIDs of those runs, hence using run_id here is actually what the users want. If the runs the users work on happen to be from a number of different databases, i don't see them working with the captured_run_id as they may not uniquely identify a run, hence my feeling is that they work with run_id. So, i think that using run_ids here is fine.

However, i do think that we should make it explicit in the docstring - that it's `run_id` and not `captured_run_id`.

qcodes/dataset/sqlite/queries.py
```diff
@@ -829,10 +831,47 @@ def get_guid_from_run_id(conn: ConnectionPlus, run_id: int) -> str:
...
         conn: database connection
         run_id: id of the run
+
+    Returns:
+        The guid of the run_id.
     """
     return select_one_where(conn, "runs", "guid", "run_id", run_id)
 
 
+def get_guid_from_multiple_run_ids(db_path: Union[str, Path],
```
astafan8, line 841: plural of guid, right?
```suggestion
def get_guids_from_multiple_run_ids(db_path: Union[str, Path],
```

qcodes/tests/dataset/test_guid_helpers.py
```diff
@@ -77,3 +81,39 @@ def test_many_guids_from_list_str() -> None:
...
+                        p1(v)
+                        datasaver.add_result((p1, cast(float, p1())),
+                                             (p2, cast(float, p2())))
+                guid = datasaver.dataset.guid
+                guids.append(guid)
+            datasaver.flush_data_to_database(block=True)
+        return guids
+
+    path = tmp_path/'dbfile2.db'
+    path.parent.mkdir(exist_ok=True, parents=True)
```
astafan8, line 111: why creating `tmp_path`? doesn't pytest ensure that it exists?

qcodes/tests/dataset/test_guid_helpers.py
```diff
@@ -77,3 +81,39 @@ def test_many_guids_from_list_str() -> None:
...
+            # Meaure for 2 times to get 2 run ids and 2 guids
+            for run in range(2):
+                with meas.run() as datasaver:
+                    for v in np.linspace(0*run, 2*run, 50):
+                        p1(v)
+                        datasaver.add_result((p1, cast(float, p1())),
+                                             (p2, cast(float, p2())))
+                guid = datasaver.dataset.guid
+                guids.append(guid)
+            datasaver.flush_data_to_database(block=True)
```
astafan8, line 107: this is not needed because it happens already upon exit of the `with meas.run() as datasaver` context

astafan8 (commented):
a few minor things, and then ready for merge :)

qcodes/tests/dataset/test_guid_helpers.py
FarBo: OK. I will check it.

qcodes/tests/dataset/test_guid_helpers.py
FarBo: Also remove it and test if it works

qcodes/dataset/sqlite/queries.py
FarBo: It can be guid if only one run provided, but yes, plural makes more sense :)

FarBo:
@astafan8 
`@FarBo please correct me if the following is wrong`
That's correct. I will add some explanation in the docstring about this

FarBo:
@astafan8 
I think I addressed all of your points.
Appreciate if sign the PR if everything looks fine :)

astafan8 (approved):

PR #2934, opened 2021-04-16
Don't write extra metadata for xarray in a nested dict 

### LINKED ISSUE #2922: Dataset with nested metadata throws TypeError when exporting to NetCDF file
### Steps to reproduce
1. Prepare a Dataset
2. Add metadata to the dataset where the value is a `dict`
3. Export the dataset to netCDF

### Expected behaviour
The data is exported to netCDF where values that are `dict`s are either skipped or converted to `str`

### Actual behaviour
```python
<ipython-input-155-35b4f9ff4165> in <module>
----> 1 ds.export(export_type="netcdf", path=output_dir, prefix=file_prefix)

~/repos/Qcodes/qcodes/dataset/data_set.py in export(self, export_type, path, prefix)
   1711             )
   1712 
-> 1713         self._export_path = self._export_data(
   1714             export_type=export_type,
   1715             path=path,

~/repos/Qcodes/qcodes/dataset/data_set.py in _export_data(self, export_type, path, prefix)
   1675             file_name = self._export_file_name(
   1676                 prefix=prefix, export_type=DataExportType.NETCDF)
-> 1677             return self._export_as_netcdf(path=path, file_name=file_name)
   1678 
   1679         elif DataExportType.CSV == export_type:

~/repos/Qcodes/qcodes/dataset/data_set.py in _export_as_netcdf(self, path, file_name)
   1642         file_path = os.path.join(path, file_name)
   1643         xarr_dataset = self.to_xarray_dataset()
-> 1644         xarr_dataset.to_netcdf(path=file_path)
   1645         return path
   1646 

~/anaconda3/envs/bringup/lib/python3.9/site-packages/xarray/core/dataset.py in to_netcdf(self, path, mode, format, group, engine, encoding, unlimited_dims, compute, invalid_netcdf)
   1687         from ..backends.api import to_netcdf
   1688 
-> 1689         return to_netcdf(
   1690             self,
   1691             path,

~/anaconda3/envs/bringup/lib/python3.9/site-packages/xarray/backends/api.py in to_netcdf(dataset, path_or_file, mode, format, group, engine, encoding, unlimited_dims, compute, multifile, invalid_netcdf)
   1057     # validate Dataset keys, DataArray names, and attr keys/values
   1058     _validate_dataset_names(dataset)
-> 1059     _validate_attrs(dataset)
   1060 
   1061     try:

~/anaconda3/envs/bringup/lib/python3.9/site-packages/xarray/backends/api.py in _validate_attrs(dataset)
    229     # Check attrs on the dataset itself
    230     for k, v in dataset.attrs.items():
--> 231         check_attr(k, v)
    232 
    233     # Check attrs on each variable within the dataset

~/anaconda3/envs/bringup/lib/python3.9/site-packages/xarray/backends/api.py in check_attr(name, value)
    220 
    221         if not isinstance(value, (str, Number, np.ndarray, np.number, list, tuple)):
--> 222             raise TypeError(
    223                 f"Invalid value for attr {name!r}: {value!r} must be a number, "
    224                 "a string, an ndarray or a list/tuple of "

TypeError: Invalid value for attr 'extra_metadata': {
```

### System
**operating system**
Ubuntu 20.04.1 LTS on WSL2
**qcodes branch**
`master`

Tagging @jenshnielsen. What would you recommend as a best solution here - drop the nested data or convert it to `str`? We currently convert the snapshot to `str` as well so to be consistent this may be the best way to go.

jenshnielsen:
Since this prevents export to netcdf

Fixes #2922


guenp:
@jana-d FYI, the `"extra_metadata"` tag is going away in favor of supporting netCDF data export.

qcodes/tests/dataset/test_dataset_export.py
```diff
@@ -230,6 +230,21 @@ def test_export_to_xarray_ds_dict_extra_metadata(mock_dataset):
...
+        "foo": {
+            "bar": {
+                "baz": "test"
+            },
+            "spam": [1, 2, 3],
+        }
+    }
+    mock_dataset.add_metadata("foo_metadata", json.dumps(nt_metadata))
+    mock_dataset.export(export_type="netcdf", path=str(tmp_path))
+
```
guenp, line 246: I think it would be good to test loading the file here and asserting that the attribute values are indeed what you'd expect.

guenp (approved):
Thank you for fixing this bug! I have one minor suggestion to make the UT more complete but otherwise LGTM.

jenshnielsen:
@jana-d @guenp I meant to say that this is one proposed solution. We could also merge all the data in extra_metadata (from all tags into one json string) but I think this would be cleaner? @astafan8 Any opinion

jana-d:
I think the proposed solution is good, there is no need for the extra_metadata tag. Also agree that not merging everything into one json string might be cleaner, but happy either way.

astafan8 (approved):
Agree with this solution as well.

the point of using the extra_metadata key to group "all aadditional metadata" was to separate it from the must-have/-be attributes, but seems to give more headache than benefit

PR #2936, opened 2021-04-16
Added merge_yamls function to station

FarBo:
This PR:

Adds the functionality of merging multiple yaml files into one temporary yaml file.
We may want discuss to keep the function as it is or improve it. Then, we can modify and test this function and modify the Station notebook accordingly.

@jenshnielsen 
@Akshita07 
@astafan8 
@Dominik-Vogel 
@samantha-ho

### DIFF
--- a/qcodes/station.py
+++ b/qcodes/station.py
@@ -36,6 +36,11 @@ from qcodes.instrument.parameter import (
 import qcodes.utils.validators as validators
 from qcodes.monitor.monitor import Monitor
 
+import tempfile
+from io import StringIO
+from collections import deque
+import ruamel.yaml
+from pathlib import Path
 
 log = logging.getLogger(__name__)
 
@@ -684,3 +689,37 @@ def update_config_schema(
                 [qcodes.instrument_drivers] + additional_instrument_modules)
         ))
     )
+
+
+def merge_yamls(*yamls: Union[str, Path]) -> IO[str]:
+    if len(yamls) == 1:
+        return open(yamls[0], "r")
+
+    top_key = "instruments"
+    yaml = ruamel.yaml.YAML()
+
+    deq: deque = deque()
+
+    # Load the yaml files and add to deque in reverse entry order
+    for filepath in yamls[::-1]:
+        with open(filepath, "r") as file_pointer:
+            deq.append(yaml.load(file_pointer))
+
+    # Add the top key entries from filepath n to filepath n-1 to ... filepath 1.
+    while len(deq) > 1:
+        data2, data1 = deq[0], deq[1]
+        for entry in data2[top_key]:
+            if entry not in data1[top_key].keys():
+                data1[top_key].update({entry: data2[top_key][entry]})
+            else:
+                raise KeyError(
+                    f"duplicate key `{entry}` detected among files: {','.join(map(str, yamls))}"
+                )
+        deq.popleft()
+
+    # Dump to a temp file so it can be read by load_config
+    merged_yaml_temp_file = tempfile.TemporaryFile("w+")
+    yaml.dump(data1, merged_yaml_temp_file)
+    merged_yaml_temp_file.seek(0)
+
+    return merged_yaml_temp_file


qcodes/station.py
```diff
@@ -36,6 +36,12 @@
 import qcodes.utils.validators as validators
 from qcodes.monitor.monitor import Monitor
 
+import tempfile
+from io import StringIO
+from collections import deque
+import ruamel.yaml
+from pathlib import Path
+from typing.Deque import deque as Tdeque
```
jenshnielsen, line 44: This import looks strange. Are you sure it should not just be?

```
from typing import Deque as Tdeque
```

FarBo: I tried that first but my local mypy fails:
qcodes\station.py:702:10: error: Missing type parameters for generic type "Tdeque"

jenshnielsen: But this is not much better since it fails with 

```
ModuleNotFoundError: No module named 'typing.Deque'; 'typing' is not a package
```

;)

What that error message means is that  Deque is a generic type like List so you should write
```
Tdeque[Any]
```
Since the definition of load is `(Union[Path, StreamTextType]) -> Any`

FarBo: For an unknown reason it passed my local mypy, although it is weird :).

I was looking for correct typing of the output, and Any should definitely work, so
"Tdeque[Any]" -> passes the mypy

qcodes/station.py
```diff
@@ -684,3 +690,37 @@ def update_schema_file(
...
+            if entry not in data1[top_key].keys():
+                data1[top_key].update({entry: data2[top_key][entry]})
+            else:
+                raise KeyError(
+                    f"duplicate key `{entry}` detected among files: {','.join(map(str, yamls))}"
+                )
+        deq.popleft()
+
+    # Dump to a temp file so it can be read by load_config
+    merged_yaml_temp_file = tempfile.TemporaryFile("w+")
```
jenshnielsen, line 722: Perhaps we should update this to use StringIO so we avoid writing to disk for the merge https://docs.python.org/3/library/io.html#io.StringIO

jenshnielsen:
@FarBo we also need to test this and imho we should also modify the station it self to take multiple yaml files and merge them

FarBo:
@jenshnielsen 
Testing is a good idea. But first we need to know how we want to change the Station itself.
from changing the station, you mean accepting multiple config_file in the init and automatically merge them into one using this merge_yamls?

jenshnielsen:
> Testing is a good idea. But first we need to know how we want to change the Station itself.

Testing is mandatory ;). And we can also test this function standalone. 

> from changing the station, you mean accepting multiple config_file in the init and automatically merge them into one using this merge_yamls?

Yes indeed something like that

astafan8:
> > from changing the station, you mean accepting multiple config_file in the init and automatically merge them into one using this merge_yamls?
> 
> Yes indeed something like that

looking forward to the implementation :)

FarBo:
@jenshnielsen 
@astafan8 

Please take a look at the modifications for haveing merge_yaml integration into Station class.
Jens,
I modified it further with more logics after the chat we had, because tests were not passing. I am using load_config_file if one yaml is provided and load_config if many. We also need to load the config if it is None, so I am using load_config_file for this case.

I am looking forward to receive your feedback, then I will approach writing tests for new features and update docs.

qcodes/station.py
```diff
@@ -132,12 +140,13 @@ def __init__(self, *components: Metadatable,
             self.add_component(item, update_snapshot=update_snapshot)
 
         self.use_monitor = use_monitor
-        self.config_file = config_file
 
         self._added_methods: List[str] = []
         self._monitor_parameters: List[Parameter] = []
 
-        self.load_config_file(self.config_file)
+        self.config_file = config_file
```
jenshnielsen, line 147: I suggest that we normalize the config_file here such that config file is always a List of strings. 
E.g. if None we should make s
```
if config_file is None:
  self.config_file = [ ]
if isinstance(config_file, str)
   self.config_file = [str,] 
else:
   self.config_file = config_file
```
That way we can write all the other code in the class such that we can always assume that self.config_file is a list

qcodes/station.py
```diff
@@ -113,7 +121,7 @@ class Station(Metadatable, DelegateAttributes):
     A user dict representing the YAML file that the station was loaded from"""
 
     def __init__(self, *components: Metadatable,
-                 config_file: Optional[str] = None,
+                 config_file: Optional[Union[str, Sequence[str]]] = None,
```
trevormorgan, line 124: should we consider making this input arg to be called `config_files`? Not sure how much extra work that would cause to change throughout the code base.

qcodes/station.py
```diff
@@ -313,6 +321,28 @@ def get_config_file_path(
         with open(path) as f:
             self.load_config(f)
 
+    def load_config_files(self,
+                          filenames: Optional[Union[str, Sequence[str]]]
```
trevormorgan, line 325: any reason why we shouldn't just make this a positional argument `*filenames` ?

qcodes/station.py
```diff
@@ -684,3 +716,45 @@ def update_schema_file(
                 [qcodes.instrument_drivers] + additional_instrument_modules)
         ))
     )
+
+
+def _merge_yamls(*yamls: Union[str, Path]) -> IO[str]:
```
trevormorgan, line 721: Should we consider not making this a private method and add an option to write to disk? I could see it being useful as a tool to consolidate many yaml files.  If we do that maybe pull into it's own file in utils.

trevormorgan:
@Dominik-Vogel  regarding your comment in the Delegate Instrument PR. 

> Dear @guenp ,
> I like the general idea of that PR and I have been long thinking about something similar when I originally implemented the station yaml file, but hadn't found such a convincing solution. (I like the idea of consolidating with the DelegateParameter). I have only one point to add:
> There is a yaml schema for the station-yaml files, which needs updating with your additions. The schema can become quite useful if you use it with a language server that supports that, like VScode with a plugin. See here in the qcodes documentation:
> https://qcodes.github.io/Qcodes/examples/Station.html#Setting-up-Visual-Studio-Code-to-work-with-station-files
> The schema file is here:
> https://github.com/QCoDeS/Qcodes/blob/master/qcodes/dist/schemas/station-template.schema.json

Any unforeseen issues with how we merge yaml files here and the yaml schema that you mentioned?

qcodes/station.py
FarBo: I also think writing to disk could be potentially useful. Perhaps having a boolean argument like write_to_disk with default of False.

qcodes/station.py
Akshita07: I am not sure about this change as I think this is a breaking change. We can consider it when it absolutely becomes necessary and for now probably keeping the argument name same is ok.

qcodes/station.py
Akshita07: @trevormorgan do you mean to making it a public method as it is already a private function? Though for the time being I think it can be private and may be we can think of making this public when the need arises.

jenshnielsen:
@trevormorgan I don't think so. The actual validation happens after the file has been merged as far as I can see. The linter within vs code may work less well for certain ways of splitting the file but I think that is a very minor issue

qcodes/station.py
```diff
@@ -313,6 +326,27 @@ def get_config_file_path(
...
+        instantiation with the :meth:`load_instrument` method.
+
+        Additionally the shortcut methods ``load_<instrument_name>`` will be
+        updated.
+        """
+        if len(filenames) == 1 and filenames[0] is None:
+            self.load_config_file()
+        else:
+            with _merge_yamls(*filenames[0]) as yamls:
+                self.load_config(yamls)
```
jenshnielsen, line 348: This still has the same issue as before. If there is only one config file this is passed through load_config_file which will expand the path to the config file but if there is more than one config file the path is passed directly to _merge_yaml without being expanded.

Akshita07: Yes, `get_config_file_path` is getting the file path in load_config_file method. Fixing this for new method as well.

qcodes/station.py
```diff
@@ -333,7 +334,7 @@ def load_config_file(self, filename: Optional[str] = None) -> None:
             self.load_config(f)
 
     def load_config_files(self,
-                          *filenames: List[Union[str, Path]]
```
jenshnielsen: Why did this change from using `*filenames` to not. We do want to use `*` args here

Akshita07: Yes, I am working on it. Tests were failing with `*` and I am figuring out why.

jenshnielsen: Ok, I would suggest that you work on that by running the tests locally before committing and pushing

Akshita07: Ok, will do that going forward.

Akshita07:
@jenshnielsen I have made all the changes from my side in the source code including a test. Will update the station notebook now. Please have a look when time allows.

qcodes/station.py
```diff
@@ -90,7 +94,11 @@ class Station(Metadatable, DelegateAttributes):
     Args:
         *components: components to add immediately to the
             Station. Can be added later via ``self.add_component``.
-        config_file: Path to YAML file to load the station config from.
+        config_file: Path to YAML files to load the station config from.
+            - If only one yaml file needed to be loaded, it should be passed
+            as a string, e.g., '~/station.yaml.
+            - If more than one yaml file needed, they should be supplied as
+            a sequence of strings, e.g. ['~/station1.yaml', ~/station2.yaml'].
```
jenshnielsen, line 101: Please check if this renders correctly as a bulleted list in the docs

qcodes/station.py
```diff
@@ -313,6 +333,41 @@ def get_config_file_path(
         with open(path) as f:
             self.load_config(f)
 
+    def load_config_files(self,
+                          *filenames: Sequence[str]
```
jenshnielsen, line 337: This is not the correct way to type a *args thing. It should be the type of the content 
```suggestion
                          *filenames: str
```

qcodes/station.py
```diff
@@ -132,12 +140,19 @@ def __init__(self, *components: Metadatable,
...
         self._monitor_parameters: List[Parameter] = []
 
-        self.load_config_file(self.config_file)
+        if config_file is None:
+            self.config_file = []
+        elif isinstance(config_file, str):
+            self.config_file = [config_file, ]
+        else:
+            assert isinstance(config_file, List)
+            self.config_file = config_file
```
jenshnielsen, line 153:
```suggestion
            self.config_file = list(config_file)
```
Otherwise this will fail for a tuple and other sequences

Akshita07:
@jenshnielsen last review comment is resolved and the station notebook is updated. I think after a review of the updated notebook, this PR is ready to be merged.

docs/examples/Station.ipynb
```diff
@@ -519,26 +519,42 @@
...
    "source": [
-    "In this section, we shall briefly describe the usage of `Station` with a YAML configuration file. For more details, we refer to the docstrings of the methods of the `Station` class.\n",
+    "In this section, we shall briefly describe the usage of `Station` with YAML configuration files. For more details, we refer to the docstrings of the methods of the `Station` class.\n",
     "\n",
-    "A `Station` with a YAML configuration file can be created by passing the file name (or file name with absolute path) to `Station`s constructor. File name and location resolution also takes into account related entries in the `'station'` section of the QCoDeS configuration. We refer to corresponding documentation for more information.\n",
+    "A `Station` can be created with or without one or more YAML configuration files. With YAML configuration files, `Station` can be created by passing the file names (or file names with absolute path) to `Station`s constructor. File names and location resolution also takes into account related entries in the `'station'` section of the QCoDeS configuration. We refer to corresponding documentation for more information.\n",
```
FarBo, line 531: to `Station`s constructor. --?-> to `Station`'s constructor/

docs/examples/Station.ipynb
```diff
@@ -519,26 +519,42 @@
...
     "```\n",
     "\n",
-    "Alternatively, `load_config_file` method can be called on an already instantiated station to load the configuration file.\n",
+    "or\n",
+    "\n",
+    "```python\n",
+    "station = Station(config_file=['example_1.yaml', 'example_2.yaml'])\n",
+    "```\n",
+    "\n",
+    "Alternatively, `load_config_file` or `load_config_files` methods can be called on an already instantiated station to load the configuration files.\n",
```
FarBo, line 543: on an already instantiated station --?--> on an already instantiated Station

docs/examples/Station.ipynb
```diff
@@ -519,26 +519,42 @@
...
+    "\n",
+    "* `load_config_file` method can be used for loading 1 configuration file.\n",
+    "* `load_config_files` method can be used for loading 1 or more configuration files. If multiple configuration files are provided, this method first merges them into one and then loads.\n",
```
FarBo, line 546: Shouldn't we mention that the merge is temporary and not physically saved in the disk?

FarBo, line 546: this method first merges them into one and then loads.  --?--> this method first merges them into one temporary YAML file and then loads that into the station.

Akshita07: Sure, I'll mention it

trevormorgan (approved):
Took one last look through the PR and looked at the updates in the notebook. Looks good from my end. Nice job @Akshita07 and @FarBo!

qcodes/station.py
```diff
@@ -684,3 +742,48 @@ def update_schema_file(
                 [qcodes.instrument_drivers] + additional_instrument_modules)
         ))
     )
+
+
+def _merge_yamls(*yamls: Union[str, Path]) -> IO[str]:
```
astafan8, line 747: didn't we discuss before that this IO output is unnecessary and a simple string output would work perfectly and make the usage of this function easier? :)
```suggestion
def _merge_yamls(*yamls: Union[str, Path]) -> str:
```

qcodes/station.py
```diff
@@ -313,6 +334,41 @@ def get_config_file_path(
...
+                assert isinstance(filename, str)
+                path = self._get_config_file_path(filename)
+
+                if path is None and filename is not None:
+                    raise FileNotFoundError(path)
+
+                paths.append(path)
+
+            if None in paths:
+                raise RuntimeWarning("Couldn't find path to one or more of the "
```
astafan8, line 366: isn't this case taken care of by ``raise FileNotFoundError(path)`` line earlier?

qcodes/station.py
Akshita07: Yes, we discussed it but according to my understanding it was not decided to definitely make it a string. Hence, I tried making it a str but got some errors so reverted it back. But you are right, I'll work on it and do the updates.

qcodes/station.py
Akshita07: yes, you are right. removing it.

astafan8 (approved):

PR #2937, opened 2021-04-16
contribution guidelines: remove no longer relevant sections and make the sections on naming somewhat less strict

jenshnielsen:
Start updating the contribution guide. 
Remove no longer relevant sections and make the sections on naming somewhat less strict:

There is a lot more updating that could be done. Specifically I think we should perhaps write something about pyvisa-sim and example notebooks (or at least add some links) 

@astafan8 @trevormorgan @samantha-ho @sohailc  @FarBo @Akshita07 @Dominik-Vogel 
Please have a look and see if there is anything you would like to change


guenp:
This looks great, thanks for simplifying! I would also perhaps add a section on how to write tests using mock instruments.

CONTRIBUTING.rst
```diff
@@ -359,11 +252,6 @@ Pull requests
    (``- [ ] take over the world``) to let others know what more to
    expect in the near future.
 
--  There are a number of emoji that have specific meanings within our
-   github conversations. The most important one is :dancer: which means
```
trevormorgan, line 363: is it because there is no longer a dancer emoji and only 👯 :). Fun concept but probably hard to get everyone to commit to it.

jenshnielsen: It's more that there is actually a approve button now so we don't need to pick a emoji but good question what happened to the dancer ❓

jenshnielsen:
> This looks great, thanks for simplifying! I would also perhaps add a section on how to write tests using mock instruments.

Agreed (my reference to pyvisa-sim was intended to cover something similar)

CONTRIBUTING.rst
```diff
@@ -248,11 +148,6 @@ born familiar with git, and everybody makes mistakes.
 -  Write the docs, following the other documentation files (.rst) in the
    repo.
```
astafan8, line 149:
```suggestion
-  Write the docs, following the other documentation files (.rst) in the
   repo as an example. Or write the docs in the form of example IPython
   notebook (there are many of those in our docs as well).
```

astafan8 (approved):
I think that for now this is a good cleanup.

Shall the PR be renamed to "contribution guidelines: remove no longer relevant sections and make the sections on naming somewhat less strict" ? :)

CONTRIBUTING.rst
```diff
@@ -263,27 +158,31 @@ good or fail and where it fails.
...
+      and analysis. You can break that up into an acquisition test and an
+      analysis by saving the intermediate state, namely the data file, in
+      the test directory. Use it to compare to the output of the
+      acquisition test, and as the input for the analysis test.
+
+-  Refer to QCoDeS documentation on how to implement tests for the 
+   instrument drivers.
+
+   -  We have not yet settled on a framework for testing real hardware.
+      For some tests we use `pyvisa-sim <https://github.com/pyvisa/pyvisa-sim>`__
```
astafan8, line 178: @jenshnielsen @guenp i added a short notion of pyvisa-sim here, can be improved/expanded in the future. I also refer the readers to qcodes docs, becasue for example pyvisa-sim use is explained there https://qcodes.github.io/Qcodes/examples/writing_drivers/Creating-Simulated-PyVISA-Instruments.html . do you agree with this minimal improvement?

jenshnielsen: Yes I think that is good for now :)

PR #2938, opened 2021-04-17
Add conversion between qcodes legacy DataSet and xarray format

peendebak:

- Add conversion methods between the QCoDeS legacy dataset and xarray Dataset

A minimal example:
```
from qcodes.tests.legacy.data_mocks import DataSet1D, DataSet2D
qd = DataSet1D()
xds = qd.to_xarray()  
print(xds)
qds = DataSet.from_xarray(xds)
```

@jenshnielsen The goal of this PR is to allow conversion between the easy cases (e.g. 1D, 2D). Since the two datasets have slightly different formats, we do not cover all cases. Can you comment on whether this functionality can be added to qcodes and if the structure of the PR is good?


jenshnielsen:
@peendebak Thanks, I will have a look. Happy to merge this to QCoDeS. I have edited the description to clarify that this is about the legacy `data.dataset` and not the current QCoDeS dataset.

qcodes/data/data_array.py
```diff
@@ -529,3 +532,50 @@ def fraction_complete(self):
...
+    else:
+        if data_array.set_arrays:
+            data_dictionary["dims"] = tuple([a.array_id for a in data_array.set_arrays])
+            data_dictionary["depends_on"] = data_dictionary["dims"]
+        data_dictionary["data"] = data_array.ndarray
+
+    return data_dictionary
+
+
+if __name__ == '__main__':
```
astafan8, line 578: let's remove this if block, it's contents are in the tests (as far as i understand), and qcodes submodules are not really meant to be executable))

astafan8 (approved):
@peendebak thanks for the contribution (and the typing, and the tests)! if this helps for the users of the legacy DataSet, let's get it in :)

qcodes/data/data_array.py
peendebak: Done!

peendebak:
@astafan8 I addressed the comments, and added some more tests. The build seems to be failing due to a general error.

astafan8:
> The build seems to be failing due to a general error.

@peendebak  what do you mean with "general error"? the error here https://github.com/QCoDeS/Qcodes/pull/2938/checks?check_run_id=2399670141 seems to be a typo in the assert method call:
```
>       self.asserEqual(qd.default_parameter_array().shape, xarray_dataset.tmin.shape)
E       AttributeError: 'TestDataSet' object has no attribute 'asserEqual'
```
it should be assertEqual, with "t" :)

PR #2949, opened 2021-04-21
AMI430 3D driver: allow instantiating from names of existing AMI430 instances, next to from AMI430 instances only

astafan8:
Useful when defining AMI430_3D in station YAML config file (**after** defining the individual AMI430 axes)

as requested by @jdwatson.

also: Make error message explicit when instrument with name does not exist.

### DIFF
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -625,6 +625,8 @@ class Instrument(InstrumentBase, AbstractInstrument):
             TypeError: If a specific class was requested but a different
                 type was found.
         """
+        if name not in cls._all_instruments:
+            raise KeyError(f"Instrument with name {name} does not exist")
         ins = cls._all_instruments[name]()
 
         if ins is None:
--- a/qcodes/instrument_drivers/american_magnetics/AMI430.py
+++ b/qcodes/instrument_drivers/american_magnetics/AMI430.py
@@ -2,7 +2,8 @@ import collections
 import logging
 import time
 from functools import partial
-from typing import Union, Iterable, Callable, Any, Optional, Tuple, List, Sequence, TypeVar
+from typing import Union, Iterable, Callable, Any, Optional, Tuple, List, \
+    Sequence, TypeVar, cast
 import numbers
 import warnings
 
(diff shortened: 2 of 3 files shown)


qcodes/instrument/base.py
```diff
@@ -625,6 +625,8 @@ def find_instrument(cls, name: str,
             TypeError: If a specific class was requested but a different
                 type was found.
```
jenshnielsen, line 626: Update docstring to document that a KeyError is raised if not found

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -462,9 +463,9 @@ def _update_units(
...
                  name: str,
-                 instrument_x: AMI430,
-                 instrument_y: AMI430,
-                 instrument_z: AMI430,
+                 instrument_x: Union[AMI430, str],
+                 instrument_y: Union[AMI430, str],
+                 instrument_z: Union[AMI430, str],
                  field_limit: Union[numbers.Real,
                                     Iterable[CartesianFieldLimitFunction]],
                  **kwargs: Any):
```
jenshnielsen, line 471: Should this have a docstring describing the use of other instruments as str etc ...?

qcodes/tests/drivers/test_ami430.py
```diff
@@ -107,6 +107,60 @@ def ami430():
...
+
+    with pytest.raises(
+            ValueError, match="instrument_z argument is neither of those"
+    ):
+        AMI430_3D(
+            "AMI430-3D",
+            mag_x.name, mag_y, badly_typed_instrument_z_argument,
+            field_limit
+        )
+
```
jenshnielsen, line 162: Should we also add a test for a badly named instrument. E.g. a string but not the name of an AMI

jenshnielsen (approved):
Looks good to me. Left a few questions inline. Is this worth documenting in the example notebook ?

qcodes/instrument/base.py
astafan8: it's already documented:
> KeyError: If no instrument of that name was found, or if its reference is invalid (dead).

qcodes/instrument_drivers/american_magnetics/AMI430.py
astafan8: abolutely, i forgot, will do in a subsequent PR

qcodes/tests/drivers/test_ami430.py
astafan8: yes. i'll add a variation of the test_instantiation_from_name_of_nonexistent_instrument for that

PR #2960, opened 2021-04-27
try changing formatting for env markers differently

jenshnielsen:
Perhaps this will enable dependabot to parse them correctly?
Since it has no problem with the pywin32 related env markers


jenshnielsen:
This does not seem to have any effect. Most likely dependabot uses python 3.8 or newer to parse the requirements files so reqs that are marked as 3.7 only are ignored

PR #2972, opened 2021-05-03
Support websocket 9

jenshnielsen:
Import server apis from their new location in the legacy module. While they moved to a legacy module
that does not mean that they are deprecated just yet so no need to change the use. See https://websockets.readthedocs.io/en/stable/changelog.html#id4 for more details

replaces https://github.com/QCoDeS/Qcodes/pull/2970

Also removed a no longer needed import shim for python < 3.7


jenshnielsen:
This is likely blocked on a conda forge release of websockets 9 for now. See https://github.com/conda-forge/websockets-feedstock/pull/19

astafan8 (approved):

PR #2979, opened 2021-05-06
Add darker to precommit hooks

jenshnielsen:
This adds darker https://github.com/akaihola/darker as a pre commit hook. Darker basically works as a incremental version of black that only reformat the code that has been changed in a given pr


jenshnielsen:
This does not actually work correctly due to https://github.com/akaihola/darker/issues/130 

So this should not be merged until a new version with that fix has been released

trevormorgan:
cool find! Sounds like this is good to go with the work around.

trevormorgan (approved):

PR #2980, opened 2021-05-06
[Bug fix] Return parameter data with same type as before applying scale or offset 

Akshita07:
At the moment, on applying scale or offset, parameter data return type changes to tuple if the original data type was any iterable data type. Desired behaviour is to return data with same type as before applying scale or offset.

This PR fixes it. 

- Update scaling and offset logic

@jenshnielsen @astafan8


qcodes/instrument/parameter.py
```diff
@@ -533,28 +533,30 @@ def _from_raw_value_to_value(self, raw_value: ParamRawDataType
...
+            try:
+                value = value - self.offset
+            except TypeError:
+                if isinstance(self.offset, collections.abc.Iterable):
+                    # offset contains multiple elements, one for each value
+                    value = tuple(val - offset for val, offset
+                                  in zip(value, self.offset))
+                elif isinstance(value, collections.abc.Iterable):
+                    # Use single offset for all values
+                    value = tuple(val - self.offset for val in value)
```
astafan8, line 545: what happens on an implicit else here? shall we reraise the TypeError at least
```suggestion
                    value = tuple(val - self.offset for val in value)
                raise
```

qcodes/tests/parameter/test_parameter_scale_offset.py
```diff
@@ -69,11 +69,15 @@ def test_scale_and_offset_raw_value_iterable(values, offsets, scales):
...
     assert p.raw_value == values
+    old_value = p()
 
     # test setting scale and offset does not change anything
     p.scale = scales
     p.offset = offsets
     assert p.raw_value == values
+    new_value = p()
+
+    assert type(new_value) == type(old_value)
```
astafan8, line 80: do we have an explicit test like this for numpy-array-valued parameter?

qcodes/instrument/parameter.py
Akshita07: yes, i'll add it

qcodes/tests/parameter/test_parameter_scale_offset.py
Akshita07: I do not understand your question. Are you suggesting this test condition should be a test in itself?

astafan8: > Are you suggesting this test condition should be a test in itself?

nope. I am asking whether there exists a test case (could be that it's already part of the parametrization of this test) where we assert scale/offset working on a parameter that returns numpy arrays (as opposed to scalars, tuples, lists, etc)

Akshita07: there is a condition tested for when all set values, offset, scale and parameter returned values are a numpy array, and this test checks for the correct data result. But the return data types itself are not checked. This assert condition which I added checks explicitly for that.

astafan8: > values, offset, scale and parameter returned values are a numpy array

is this really true? the way i read the hypothesis strategy, only python lists are tested for, not numpy arrays. perhaps i'm misreading :)

Akshita07: hmm, you are right... I need to understand how hypothesis strategy is working better. What I was referring to in my previous reply is the test on line 87 .. that's just a test for data.

qcodes/instrument/parameter.py
Akshita07: updated

qcodes/tests/parameter/test_parameter_scale_offset.py
```diff
@@ -172,3 +172,20 @@ def test_scale_and_offset_raw_value_iterable_for_set_cache(values,
         event('Scale is array and also offset')
     if isinstance(scales, Iterable) and not isinstance(offsets, Iterable):
         event('Scale is array but not offset')
+
+
+def test_data_return_type_after_setting_scale_and_offset():
```
astafan8, line 177: perhaps a more specific/explicit test name?
```suggestion
def test_numpy_array_valued_parameter_preserves_type_if_scale_and_offset_are_set():
```

qcodes/instrument/parameter.py
```diff
@@ -533,28 +533,34 @@ def _from_raw_value_to_value(self, raw_value: ParamRawDataType
...
-                # offset contains multiple elements, one for each value
-                value = tuple(val - offset for val, offset
-                              in zip(value, self.offset))
-            elif isinstance(value, collections.abc.Iterable):
-                # Use single offset for all values
-                value = tuple(val - self.offset for val in value)
-            else:
-                value -= self.offset
+            try:
+                value = value - self.offset
```
astafan8, line 537: @Akshita07 my main comment is: i think we should also mirror a part of this change into the ``_from_value_to_raw_value`` method, so that setting a numpy-valued parameter with not-None offset/scale works as well. the only change needed for that is to change ``raw_value *= self.scale`` to ``raw_value = raw_value * self.scale`` and ``raw_value += self.offset`` to ``raw_value = raw_value + self.offset``. would you agree @jenshnielsen ?

Akshita07: Yes, I'll update it.

Akshita07:
@astafan8 I have added a test for setting numpy valued param. Please have a look when time allows.

astafan8 (approved):

PR #2982, opened 2021-05-06
add _repr_html_ for qcodes instruments

peendebak:
- Add `_repr_html_` for all qcodes instruments. This gives a nice representation in tools like jupyter notebook.

New representation:

image

Old representation: 

image

@jenshnielsen

### DIFF
--- a/qcodes/instrument/base_representation.py
+++ b/qcodes/instrument/base_representation.py
@@ -93,7 +93,7 @@ def instrument_repr_html(self: "qcodes.instrument.base.Instrument") -> str:
     }
 
     sections = [
-        parameter_section(section_data),  # type: ignore
+        parameter_section(section_data),
         functions_section(functions_section_data),
     ]
 
--- a/qcodes/instrument/base_representation.py
+++ b/qcodes/instrument/base_representation.py
@@ -93,7 +93,7 @@ def instrument_repr_html(self: "qcodes.instrument.base.Instrument") -> str:
     }
 
     sections = [
-        parameter_section(section_data),
+        parameter_section(section_data),  # type: ignore
         functions_section(functions_section_data),
     ]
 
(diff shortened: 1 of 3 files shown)


jenshnielsen:
Thanks @peendebak  Looks very useful. We will have a look

qcodes/instrument/base_representation.py
```diff
@@ -0,0 +1,104 @@
+""" Representation methods for instruments """
+import logging
+import uuid
+from collections import namedtuple
+from functools import partial
+from html import escape
+from typing import Any, Dict, Optional
+
+from xarray.core.formatting_html import _icon, _mapping_section, _obj_repr
```
astafan8, line 9: @peendebak these seem to be private functions from deep inside xarray package, we can't have this as a dependency in qcodes because it's not guaranteed that these will not change (or that xarray will notify the world about changing those), because they are private (the ``_`` prefix in the name).  Hence this must change. I suggest that there functions get re-implemented right here in qcodes (the direct copy-pasting is not an option because xarray and qcodes licenses are different and it may be more trouble to deal with that, than just reimplement those functions), would you be so kind to do that? after that the PR can be merged (perhaps on time for the mid-May release!)

qcodes/instrument/base_representation.py
```diff
@@ -0,0 +1,104 @@
...
+        f"<div class='xr-var-name'><span{cssclass_idx}>{name}</span></div>"
+        f"<div class='xr-var-preview xr-preview'>{preview}</div>"
+        f"<input id='{data_id}' class='xr-var-data-in' type='checkbox' {disabled}>"
+        f"<label for='{data_id}' title='Show/Hide data repr'>"
+        f"{data_icon}</label>"
+        f"<div class='xr-var-data'>{data_repr}</div>"
+    )
+
+
+def create_entries(variables: Dict[str, Entry]) -> str:
```
astafan8, line 41: could you please make all functions/variables apart from ``instrument_repr_html`` private (so that it's clear that they can change without notice)?

PR #2990, opened 2021-05-10
Generalized N-dimensional dond with sweep objects

FarBo:
This PR introduces the generalized dond function.
Actions are still missing to be implemented and they could be added after reviewing the current implementation.
Do we also need flush_columns like do2d?

Following taks will be addressed after completion of dond function.
- [x] Testing
- [x] dond notebook update

@jenshnielsen 
@astafan8


qcodes/utils/dataset/doNd.py
```diff
@@ -483,6 +483,180 @@ def do2d(
...
+    if do_plot is None:
+        do_plot = config.dataset.dond_plot
+    if show_progress is None:
+        show_progress = config.dataset.dond_show_progress
+
+    meas = Measurement(name=measurement_name, exp=exp)
+
+    def _make_nested_setpoints(
+        params: List[_BaseParameter],
+        param_dicts: Dict[_BaseParameter, Dict[str, Union[float, int]]]
```
Akshita07, line 532: Union[float, int] is equivalent to float

FarBo: Yes,
The structure is under change. I think, it is better to review it after refactoring :)

qcodes/utils/dataset/doNd.py
```diff
@@ -483,6 +483,180 @@ def do2d(
...
+        """
+        _params, args_per_parameter = _find_parameters(*params)
+        arg_types = ['start', 'stop', 'num_points', 'delay']
+        params_set: List[_BaseParameter] = []
+        params_meas: List[ParamMeasT] = []
+        params_dicts: Dict[_BaseParameter, Dict[str, Union[float, int]]] = {}
+        for param in _params:
+            arguments = args_per_parameter[param]
+            if not arguments:
+                if isinstance(param, _BaseParameter):
```
Akshita07, line 590: this if statement might be replaced with an assert statement like ` assert isinstance(param, _BaseParameter)` as we are making sure that param is an instance of _BaseParameter as that is expected

qcodes/utils/dataset/doNd.py
```diff
@@ -483,6 +483,180 @@ def do2d(
...
+                    params_meas.append(param)
+            elif len(arguments) == len(arg_types):
+                if isinstance(param, _BaseParameter):
```
Akshita07, line 593: same as above

qcodes/utils/dataset/doNd.py
```diff
@@ -483,6 +483,180 @@ def do2d(
...
+        do_plot = config.dataset.dond_plot
+    if show_progress is None:
+        show_progress = config.dataset.dond_show_progress
+
+    meas = Measurement(name=measurement_name, exp=exp)
+
+    def _make_nested_setpoints(
+        params: List[_BaseParameter],
+        param_dicts: Dict[_BaseParameter, Dict[str, Union[float, int]]]
+    ) -> Union[np.ndarray, List[Tuple[()]]]:
```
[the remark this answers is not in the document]
jenshnielsen: No that is not allowed. We are using disallow_any_generics = True so all Any needs to be written out explicitly. 
This should probably be `List[Tuple[Any, ...]]`

qcodes/utils/dataset/doNd.py
```diff
@@ -483,6 +483,180 @@ def do2d(
...
+                    params_meas.append(param)
+            elif len(arguments) == len(arg_types):
+                if isinstance(param, _BaseParameter):
+                    params_set.append(param)
+                    params_dicts[param] = {}
+                    for arg_type, value in zip(arg_types, arguments):
+                        params_dicts[param][arg_type] = value
+            else:
+                raise ValueError(f'Invalid number of arguments for '
+                                 'parameter {param}.')
```
Akshita07, line 600: f'parameter {param}.') .. `f` is missing

qcodes/utils/dataset/doNd.py
```diff
@@ -483,6 +483,180 @@ def do2d(
...
+        return params_set, params_dicts, params_meas
+
+    params_set, params_dicts, params_meas = _parse_dond_arguments(*params)
+
+    all_setpoint_params = tuple(params_set) + tuple(
+            s for s in additional_setpoints)
+
+    measured_parameters = tuple(param for param in params_meas
+                                if isinstance(param, _BaseParameter))
```
Akshita07, line 609: this if statement is not needed as it is already taken care of in the `_parse_dond_arguments` method

qcodes/utils/dataset/doNd.py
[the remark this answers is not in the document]
FarBo: For do0d, the nested setpoint should be an empty tuple. I did look at the documentation and it is stated like this:
"The type of the empty tuple can be written as Tuple[()]."
https://docs.python.org/3/library/typing.html#typing.Tuple

qcodes/utils/dataset/doNd.py
FarBo: This will go away in the refactor

qcodes/utils/dataset/doNd.py
```diff
@@ -483,6 +483,180 @@ def do2d(
...
+                raise ValueError(f'Invalid number of arguments for '
+                                 'parameter {param}.')
+        return params_set, params_dicts, params_meas
+
+    params_set, params_dicts, params_meas = _parse_dond_arguments(*params)
+
+    all_setpoint_params = tuple(params_set) + tuple(
+            s for s in additional_setpoints)
+
+    measured_parameters = tuple(param for param in params_meas
```
Akshita07, line 608: since `params_meas` is a list it can be converted to tuple as `tuple(params_meas)`

qcodes/utils/dataset/doNd.py
```diff
@@ -483,6 +483,180 @@ def do2d(
...
+                                if isinstance(param, _BaseParameter))
+
+    num_points_params_set: List[Union[float, int]] = []
```
Akshita07, line 611: same as above `Union[float, int]` can be replaced with `float`

qcodes/utils/dataset/doNd.py
```diff
@@ -483,6 +484,242 @@ def do2d(
...
+        delay: Time in second between two consequtive sweep points. The default
+            is 0.
+    """
+    def __init__(self, param: _BaseParameter, start: float, stop: float,
+                 num_points: int, delay: float = 0):
+        self._param = param
+        self._start = start
+        self._stop = stop
+        self._num_points = num_points
+        self._delay = delay
```
jenshnielsen, line 507: Hmn thinking about this perhaps the interface. That way users can define another interface that works differently.

E.g. start, stop and stepsize

qcodes/utils/dataset/doNd.py
```diff
@@ -483,6 +484,242 @@ def do2d(
     return _handle_plotting(dataset, do_plot, interrupted())
 
 
+class _AbstractSweep:
```
jenshnielsen, line 487: I think we should make this an abstract base class https://docs.python.org/3/library/abc.html

qcodes/utils/dataset/doNd.py
```diff
@@ -495,6 +496,242 @@ def do2d(
...
+    Linear sweep class. The class should be instantiated with the signature
+    defined in the _AbstractSweep class.
+    """
+
+    def get_setpoints(self) -> np.ndarray:
+        """
+        Linear numpy array for supplied start, stop and num_points in the
+        class.
+
+        Rturns:
```
jenshnielsen, line 549:
```suggestion
        Returns:
```

qcodes/utils/dataset/doNd.py
```diff
@@ -495,6 +496,242 @@ def do2d(
...
+            elif isinstance(param, _BaseParameter):
+                params_meas.append(param)
+            else:
+                raise ValueError(f'Cannot proceed: {param} is not either a'
+                                 ' sweep instance or measurement parameter.'
+                                 ' Please check the supplied `dond`'
+                                 ' arguments.')
+        return params_set, params_delay, params_meas
+
+    params_set, params_delay, params_meas = _parse_dond_arguments(*params)
```
jenshnielsen, line 682: Lets rewrite _parse so it returns a list of Sweeps and a list of measured parameters. That way the list of sweeps can be passed to _make_nested_setpoints (if that function is updated) and we can simplify the parsing code significantly

qcodes/utils/dataset/doNd.py
```diff
@@ -495,6 +496,242 @@ def do2d(
...
+    all_setpoint_params = tuple(params_set) + tuple(
+            s for s in additional_setpoints)
+
+    measured_parameters = tuple(param for param in params_meas
+                                if isinstance(param, _BaseParameter))
+
+    num_points_params_set: List[int] = []
+    for par in params:
+        if isinstance(par, _AbstractSweep):
+            num_points_params_set.append(int(par._num_points))
```
jenshnielsen, line 694: This indicates that that num_points should have a property on the Sweep interface

qcodes/utils/dataset/doNd.py
FarBo: That was my concern too because I limit people to use this setpoints convension.
I need to get a bit info about intefacing in Python. Then I will change it. Any suggestion to good examples and documents are welcome :)

qcodes/utils/dataset/doNd.py
FarBo: Thanks. I will look into using it.

qcodes/utils/dataset/doNd.py
```diff
@@ -8,6 +8,7 @@
 
 import matplotlib
 import numpy as np
+from numpy.lib.arraysetops import isin
```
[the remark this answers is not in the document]
FarBo: No. Propbably it has been added by running pre-commit locally. I will remove it.

qcodes/utils/dataset/doNd.py
```diff
@@ -495,6 +496,242 @@ def do2d(
...
+    )
+    _set_write_period(meas, write_period)
+
+    with _catch_keyboard_interrupts() as interrupted, meas.run() as datasaver:
+        additional_setpoints_data = _process_params_meas(additional_setpoints)
+        for setpoints in tqdm(nested_setpoints, disable=not show_progress):
+            param_set_list = []
+            param_value_pairs = zip(params_set[::-1], setpoints[::-1])
+            for setpoint_param, setpoint in param_value_pairs:
+                setpoint_param.post_delay = params_delay[setpoint_param]['delay']
```
jenshnielsen, line 722: It should not be needed to set post_delay more than once. E.g. this can be moved outside the for loop. 

We should also ensure that this is reset after the loop. (have a look at the other do1d, do2d functions)

qcodes/utils/dataset/doNd.py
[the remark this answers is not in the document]
jenshnielsen: No our precommit hooks does not do that ;)

qcodes/utils/dataset/doNd.py
FarBo: Exactly

jenshnielsen:
@FarBo I added some suggestions inline. Before we do that I think we should get some tests written and start with that and do some proper tdd

qcodes/utils/dataset/doNd.py
FarBo: Mypy complains if you don't do that. It is the case in the other doNd functions.

qcodes/utils/dataset/doNd.py
```diff
@@ -483,6 +483,180 @@ def do2d(
     return _handle_plotting(dataset, do_plot, interrupted())
 
 
+def dond(
+    *params: Union[_BaseParameter, int, float, ParamMeasT],
+    write_period: Optional[float] = None,
+    measurement_name: str = "",
+    exp: Optional[Experiment] = None,
+    do_plot: Optional[bool] = None,
+    show_progress: Optional[None] = None,
```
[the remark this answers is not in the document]
FarBo: Solved

FarBo:
@jenshnielsen 
Thanks Jens. They are very good suggestions and improve the code a lot. 
I will try to proceed development with the tdd process.

qcodes/utils/dataset/doNd.py
FarBo: It was a very good suggesion. Saved multiple the same redundant parsing in the dond function and gave other related power to work. Thanks 👍

qcodes/utils/dataset/doNd.py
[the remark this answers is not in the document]
FarBo: I couldn't get this last suggestion to work. we need an object to be able to iterate over one time for dond_0d. So somethig like this also works `[[]]` or `[np.array([])]` or `(np. ndarray([]))`, etc.
I took the if statement out of this private function, though and setup the similar logic, when I wanted to get the return of this function. Like this
```
if sweep_instances:
        nested_setpoints = _make_nested_setpoints(sweep_instances)
    else:
        nested_setpoints = [()]  # For zero-dimensional grid (do0d)
```
I don't know if you have any better way to dela with this.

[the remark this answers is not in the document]
astafan8: the loop inside ``with meas.run()`` that calls ``datasaver.add_results`` is implemented in such a way that in order for it to work for 0D cases, nested_setpoints must have a single empty-tuple (or emplty-sequence) value -- it semantically means "do a single 'get' of the measured parameter(s), but there are no specific setpoint values for this 'get'". The reason _make_nested_setpoints can't give that for us automatically is because ``np.vstack`` does not accept an empty list as an input (that's what happens when the list of the given sweeps is empty).

My opinion is that it's OK to let ``_make_nested_setpoints`` return ``np.array([[]])`` (empty list in a list) in case ``flat_setpoint_grids`` list inside the functions is empty. This will remove the need for the ``if sweep_instances`` clause altogether, and keeps the code readable. see here:
image

i tested this locally by calling dond with an arrayparameter, and it works and propagates the shapes correctly.

would you agree @FarBo ?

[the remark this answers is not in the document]
FarBo: Thanks for the explanation @astafan8 
So it explains why I couldn't get it to work with an empty numpy array.

I see you have nicely used numpy array to hide a list of list :). If it works, then makes the the nested_setpoints more consistant and I like it :). Perhaps I only change the if statement to not check the flat_setpoint_grid, but rather setpoint_values as it is also an empty list if there is no set parameters in dond.

I would try your suggestion locally and run my tests on it to see what I get and adapt to that if no problem found :)

astafan8 (commented):
@FarBo see here for my review and suggested fixes for them https://github.com/FarBo/Qcodes/pull/21

jenshnielsen:
@FarBo @astafan8  I have opened a pr here to nicely format all the changed code with darker https://github.com/FarBo/Qcodes/pull/22/files

jenshnielsen (approved):

astafan8 (approved):

PR #3017, opened 2021-05-17
Ensure that plot dataset uses axis in the expected order

jenshnielsen:
As it was the code relied implicitly on the order of axis in the  loaded data dictionary. This used to be stable since the data was alwasy loaded from the database in the order the columns was defined. However now the data may be loaded from the in memory cache the order is no longer guaranteed. This should fix the issue by relying on the order of axis defined by the interdeps object.

As a bonus this simplifies the code a bit by no longer having to pop the dependent data axis out and insert it again


qcodes/dataset/data_export.py
```diff
@@ -86,30 +87,18 @@ def _get_data_from_ds(ds: DataSet) -> List[List[Dict[str, Union[str, np.ndarray]
...
-            my_data_dict: Dict[str, Union[str, np.ndarray]] = {}
-
-            my_data_dict['name'] = param_name
-
-            my_data_dict['data'] = data.flatten()
-
-            ps = ds.paramspecs[param_name]
-            my_data_dict['unit'] = ps.unit
-            my_data_dict['label'] = ps.label
+        dependent = ds.paramspecs[dep_name].base_version()
```
astafan8, line 90: inderdependencies class has the __getitem__ implemented, so this should work
```suggestion
        dependent = ds.description.interdeps[dep_name]
```

astafan8 (approved):

PR #3024, opened 2021-05-18
Improved plotting of pre-shaped 2d data 

### LINKED ISSUE #2995: Plot_dataset for partially completed 2D
In one of the latest updates the plotting functions doesnt handle partially completed 2D dataset. It now falls back to a scatter plot instead of a partially filled map.
Ideally we would like to keep the 0.22.0 version behavior for partially completed plots.

Version 0.25.0
image

Version 0.22.0
image

jenshnielsen:
After the predefined shape was added to the dataset non measured data is filled with nans. This updates the 2d pcolormesh plot to handle this. 

It does this by bypassing the logic that checks for shape since this is no longer needed. However Matplotlib cannot handle a pcolormesh plot where the x and y coordinates are nans so some logic is needed to ensure this is clipped correctly. There are two use cases to consider. If the data is on a rectilinear grid we can plot all rows and columns that contain non nan setpoints. For the other case of non rectilinear grid we currently fall back to flattening the data again and removing all nans. This may not be the optimal solution but is identical to what we had before introducing shapes.

This furthermore avoids using significant resources to find the right plot type when the plot type is already given.
For shaped data we pass the gridded x and y data directly to matplotlib and avoid having to do any reshaping. 

We upgrade the Matplotlib requirements to 3.3 to avoid having to interpolate the edges of the individual patches but rely on Matplotlib to do that. 

@ThorvaldLarsen can you check if this resolves #2995


qcodes/dataset/plotting.py
```diff
@@ -252,17 +256,17 @@ def plot_dataset(dataset: DataSet,
...
-            # we check for that is still unfinished/not optimised
-
-            xpoints = flatten_1D_data_for_plot(data[0]['data'])
-            ypoints = flatten_1D_data_for_plot(data[1]['data'])
-            zpoints = flatten_1D_data_for_plot(data[2]['data'])
-
-            plottype = get_2D_plottype(xpoints, ypoints, zpoints)
-
-            log.debug(f'Determined plottype: {plottype}')
+            if data[2]["shape"] is None:
```
ThorvaldLarsen, line 259: I get a keyerror when trying out on a non-completed dataset. Changing the line to `if "shape" not in data[2].keys():` makes the PR work as intended. Note that this dataset was created by a manual measurement loop created locally. Data is fully grided so everything works but could be some way we set up the measurement context/loop that means shape was not created.

ThorvaldLarsen: Actually I think this is a combination of different bugs. Will have to do some more digging probably tomorrow to fully understand what is happening in the different cases.

It seems there are some issues around complex datatype. Maybe the shape information is not passed through correctly when getting real/imag parts for plotting. See for example `aaaaaaaa-0a00-0001-0000-01795d9a6abb` in DWH - this is a standard do2d-created dataset that was halted midway through. Dataset has two measured parameters - one of them being real and one of them complex. Real parameter plots fine but complex doesnt. But might actually be another bug in how the complex data is being exported.

My suggestion above definitely doesnt work in all cases.

Can probably dig some more tomorrow in this / do more testing if any updates are added in between.

jenshnielsen: @ThorvaldLarsen This definitely was not handling complex plots correctly. I have fixed that now

ThorvaldLarsen:
@jenshnielsen Just had a quick look and everything seems to work with the current version. Updated our plotting system and will continue to test this branch in normal operation over the next couple of days. Many thanks for resolving this issue.

qcodes/dataset/data_export.py
```diff
@@ -10,6 +11,14 @@
 log = logging.getLogger(__name__)
 
 
+class DSPlotData(TypedDict):
```
trevormorgan, line 14: Does the DS stand for data structure?

jenshnielsen: DataSet I think. This is by all means not a great name so if you have some better suggestion happy to change it

trevormorgan: hmm spelling out DataSetPlotData does seem a bit much. Maybe a simple doc string for clarity is god enough?

jenshnielsen:
@ThorvaldLarsen I ended up doing some more cleanup here. It would be great if you have a chance to test it again

jenshnielsen:
@QCoDeS/core ready for review

qcodes/dataset/plotting.py
```diff
@@ -627,6 +637,52 @@ def plot_on_a_plain_grid(x: np.ndarray,
...
+            np.nanmax(np.abs(x_data - x_row)) == 0
+            and np.nanmax(np.abs(y_data - y_row)) == 0
+        )
+
+    if _on_rectilinear_grid_except_nan(x, y):
+        """
+        clip any row or column where there are nans in the first row
+        or column. Since we fill from here we assume that means that no data
+        has been measured for this row
+        """
```
astafan8, line 660: not used to multiline string as a comment :)
```suggestion
        # clip any row or column where there are nans in the first row
        # or column. Since we fill from here we assume that means that no data
        # has been measured for this row
```

qcodes/dataset/plotting.py
```diff
@@ -574,36 +580,40 @@ def plot_on_a_plain_grid(x: np.ndarray,
...
-                              np.array([yrow[-1] + dys[-1]])))
+    if x.ndim == 2 and y.ndim == 2 and z.ndim == 2:
+        if np.logical_or(np.any(np.isnan(x)), np.any(np.isnan(y))):
+            x_to_plot, y_to_plot, z_to_plot = _clip_nan_from_shaped_data(x, y, z)
+            num_points = x_to_plot.size * y_to_plot.size
+        else:
+            # data is on a grid that may or may not be
+            # rectilinear. Rely on matplotlib to plot
+            # this directly
+            x_to_plot, y_to_plot, z_to_plot = x, y, z
```
astafan8, line 591: not a critical sugegstion, but i think it would help the readability to change the order of this if-else, so that the if block gets the easy case, and the else block gets the has-nan-somewherei-in-x-or-y case (which in turn inside a helper function split into rectilinear case that is easier, and non-reclilinear case that gets the same treatment as the else clause here)

astafan8 (approved):

ThorvaldLarsen:
@jenshnielsen @trevormorgan tested across our current database and everything has been working great so far. I would say this is ready for merging.

Only minor thing I ran into was for a completely empty dataset (dataset created but cancelled before any data was added) plotting fails with a "keyword error" which could be done more gracefully. This is a very rare case and for an empty dataset we obviously dont care about the plot. I think this has also been the normal behavior from the beginning and I would say perfectly fine to just leave as is.

trevormorgan:
> Only minor thing I ran into was for a completely empty dataset (dataset created but cancelled before any data was added) plotting fails with a "keyword error" which could be done more gracefully. This is a very rare case and for an empty dataset we obviously dont care about the plot. I think this has also been the normal behavior from the beginning and I would say perfectly fine to just leave as is.

@ThorvaldLarsen thanks for the testing and feedback! This looks good to merge now so we can have it in the upcoming release.  I'll make a note of this issue so we can make improvements for a more graceful failure due to an empty dataset but let's handle that in another PR.

PR #3030, opened 2021-05-19
Add Model_625 to Lakeshore drivers

samantha-ho:
- Adds a driver for the Lakeshore Model 625 magnet driver
- This driver originates from qcodes_contrib_drivers


qcodes/instrument_drivers/Lakeshore/Model_625.py
```diff
@@ -0,0 +1,533 @@
...
+
+from qcodes import VisaInstrument
+from qcodes.utils.validators import Numbers, Enum
+
+
+class Lakeshore625(VisaInstrument):
+    """
+    Driver for the Lakeshore Model 625 superconducting magnet power supply.
+
+    This class uses T/A and A/s as units.
```
astafan8, line 13: i think the field ramp rate unit should also be mentioned for clarity (i was actually surprised that it uses minutes as opposed to seconds, but i guess using minutes for field is more convenient?)
```suggestion
    This class uses T/A and A/s and T/min as units.
```

PR #3032, opened 2021-05-20
Make wait times in blocking AMI430 field ramps adjustible

nikhartman:
Remove 2s wait time and reduce interval for checking `ramping_state` in AMI430 driver. Move this wait loop into a separate function.


qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -340,15 +340,19 @@ def set_field(self,
...
+        exit_state = self.wait_while_ramping()
         self.log.debug(f'Finished blocking ramp')
         # If we are now holding, it was successful
-        if state != 'holding':
+        if exit_state != 'holding':
             msg = '_set_field({}) failed with state: {}'
-            raise AMI430Exception(msg.format(value, state))
+            raise AMI430Exception(msg.format(value, exit_state))
+
+    def wait_while_ramping(self, check_interval=0.05) -> str:
```
astafan8, line 350: i'd suggest to expose this interval as an attribute (or even as a parameter) so that anyone can change that as needed without needing to change the code. @nikhartman would you agree?

nikhartman: That makes sense, this is likely dependent on what interface is being used (TCP/Serial/...)

astafan8 (approved):

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -238,6 +238,10 @@ def __init__(
                                'heating switch': 9,
                                'cooling switch': 10,
                            })
+        self.add_parameter('ramping_state_check_interval',
+                           initial_value=0.05,
+                           unit="s",
+                           vals=Numbers(0, 10))
```
jenshnielsen, line 244:
```suggestion
                           vals=Numbers(0, 10),
                           set_cmd=None)
```
This needs a set_cmd to allow it to be settable

PR #3034, opened 2021-05-20
Keysight VNA P9374A driver

RosenblumLabUser:
Module for the Keysight VNA.
It relies on the classes of a different category of network analyzers but seems to work pretty well.

Fixes #issuenumber.

-
-
-

@mention one core developer


qcodes/instrument_drivers/Keysight/P9374A.py
```diff
@@ -0,0 +1,15 @@
...
+
+class P9374A(N52xx.PNAxBase):
+    def __init__(self, name: str, address: str, **kwargs: Any):
+        super().__init__(name, address,
+                         min_freq=300e6, max_freq=20e9,
+                         min_power=-43, max_power=20,
+                         nports=2,
+                         **kwargs)
+
+        options = self.get_options()
```
astafan8, line 15: is it really necessary to call this? the returned value is unused. or is there some useful side-effect of the get_options method?

astafan8 (approved):
thanks for the contribution @RosenblumLabUser ! I clarified the PR description so that it looks more informative when we make a changelog :)

PR #3073, opened 2021-05-31
add to csv exporter to dataset cache

jenshnielsen:
- [ ] Test export of shaped data to both csv and netcdf with partially completed preshapped datasets


jenshnielsen:
At a second thought I am not sure we really need this

astafan8:
> I am not sure we really need this

probably not. i think this comes from the unclarity of at which place the data is to be accessed from, from ``dataset.`` level or from ``dataset.cache.`` level, hence so far the ``to_*`` methods have been duplicated. But since the work is done already in this PR, i think it's OK to have this feature (perhaps even without tests) (i don't think it adds extra maintenance because the "ParameterData" is at the core of everything now). So, up to you :)

PR #3075, opened 2021-05-31
Dataset protocol

jenshnielsen:
Define a protocol that datasets must implement to be able to act as a storage device for the measurement class:

Note that the protocol has no `__init__` method as the arguments to the init method are currently very class specific


qcodes/dataset/data_set.py
```diff
@@ -322,6 +324,31 @@ def __init__(self, path_to_db: Optional[str] = None,
                 active_datasets=set())
             _WRITERS[self.path_to_db] = ws
 
+    def prepare(
+        self,
+        station: "Optional[Station]",
+        interdeps: InterDependencies_,
+        write_in_background: bool,
+        shapes: Shapes = None,
```
astafan8, line 332: implicit optional? :)
```suggestion
        shapes: Optional[Shapes] = None,
```

qcodes/dataset/data_set_cache.py
```diff
@@ -114,6 +84,15 @@ def data(self) -> 'ParameterData':
 
         return self._data
 
+    def load_data_from_db(self) -> None:
+        """
+        Load the data from a on disk format in case the cache is not live
```
astafan8, line 89:
```suggestion
        Load the data from an on-disk format in case the cache is not live
```

qcodes/dataset/data_set_protocol.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+if TYPE_CHECKING:
+    from qcodes.station import Station
+
+    from .data_set_cache import DataSetCache
+
+
+@runtime_checkable
+class DataSetProtocol(Protocol, Sized):
```
astafan8, line 40: would it be useful to take the opportunity and group the parts of this interface meaningfully? e.g. group metadata-related things together, properties together, data-related methods together?

however, i can also see that the interface/protocol is huge, so there's no silver-bullter grouping that would generally useful for many readers.

qcodes/dataset/data_set_protocol.py
```diff
@@ -0,0 +1,158 @@
...
+    def prepare(
+        self,
+        *,
+        station: Optional[Station],
```
astafan8, line 44: is it necessary for the dataset to know about the station? can't we pass the snapshot (as a str) directly to ``prepare``? this would be consistent with the fact that ``add_snapshot`` takes a string, and ``snapshot_raw`` is the only way to get that snapshot back (no ``snapshot`` property is defined)

qcodes/dataset/data_set_protocol.py
```diff
@@ -0,0 +1,158 @@
...
+    @property
+    def metadata(self) -> Dict[str, Any]:
+        pass
+
+    @property
+    def dependent_parameters(self) -> Tuple[ParamSpecBase, ...]:
+        pass
+
+    @property
+    def cache(self) -> DataSetCache[DataSetProtocol]:
```
astafan8, line 157: get_parameter_data is not part of the protocol, which means that in type-checked code whenever ``datasaver.dataset`` is used to call ``get_parameter_data`` mypy will be confused (right?). what is the way to avoid this? switch to using ``dataset.cache.data()``? and if the specifics of ``get_parameter_data`` are required, then users would use ``cast``?

qcodes/dataset/measurements.py
```diff
@@ -1154,8 +1169,12 @@ def set_shapes(self, shapes: Shapes) -> None:
                                              shapes=shapes)
         self._shapes = shapes
 
-    def run(self, write_in_background: Optional[bool] = None,
-            in_memory_cache: bool = True) -> Runner:
+    def run(
+        self,
+        write_in_background: Optional[bool] = None,
+        in_memory_cache: bool = True,
+        dataset_class: Type[DataSetProtocol] = DataSet,
```
astafan8, line 1176: shall this new argument be added to the docstring?

qcodes/dataset/data_set_protocol.py
jenshnielsen: This is how it was done so far but you are right we should revisit that and see if it still makes sense

qcodes/dataset/data_set_protocol.py
jenshnielsen: If you mean order the members and add some comments in between then yes. We probably cannot split the thing into multiple classes (which I would probably prefer)

qcodes/dataset/data_set_protocol.py
jenshnielsen: None of the accessors are part of the interface and yes I do indeed want to move all of this to the cache

qcodes/dataset/data_set.py
jenshnielsen: Shapes is already `Optional[Dict[str, Tuple[int, ...]]]`

qcodes/dataset/data_set_protocol.py
astafan8: > If you mean order the members and add some comments in between then yes

yes

qcodes/dataset/data_set_protocol.py
astafan8: ok. but the `cast` part will probably be annoying, but let's see

qcodes/dataset/data_set_protocol.py
jenshnielsen: I tried rearranging them and remove a few members that we could easily remove

qcodes/dataset/data_set_protocol.py
jenshnielsen: We now pass the snapshot. Please have a look and see if that is better

qcodes/dataset/data_set_protocol.py
astafan8: yes, this looks good now!

qcodes/dataset/data_set_protocol.py
```diff
@@ -0,0 +1,155 @@
...
+
+    from .data_set_cache import DataSetCache
+
+
+@runtime_checkable
+class DataSetProtocol(Protocol, Sized):
+    def prepare(
+        self,
+        *,
+        snapshot: Mapping[Any, Any],
```
astafan8, line 43: what i'm not sure is the snapshot. so `prepare` takes a dict, while `add_snapshot` takes a string, and the only declared way to get the snapshot is `snapshot_raw` which returns a string, not a dict. I don't feel good about these inconsistencies, and I am not sure what's the best to do. idea 1 is to let datasetprotocol in the first instance only take snapshot as a string (so change prepare to take a string) thus making dataset competely agnostic to the content/structure/serialization of the snapshot. idea 2 is to keep things as they are but also add the `snapshot` property from current DataSet that returns snapshot that's deserialized from string, this forces all datasetprotocol implementers to provide that interface that users actually use (noone cares about `snapshot_raw`). perhaps idea 1 is the simplest for this PR, and idea 2 can be implemented later if needed. what do you think @jenshnielsen ?

jenshnielsen: I think it would be fine to add the snapshot property back to the protocol. I have initially aimed at removing as much as possible. The other question is if we should make some of these methods and properties private. Just because `snapshot_raw` is public on the DataSet class it needs not be public on the interface. The public method would then simply be implemented as a wrapper around the private one

jenshnielsen: OK, I have added snapshot as a property and made snapshot_raw private in the protocol (retaining it in the dataset class for backwards compatibility)

astafan8 (approved):

PR #3078, opened 2021-06-01
Add Complex Voltage Parameter to SR830

trevormorgan:
Update for the SR830 lockin to match the SR86x complex voltage parameter.


qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -442,6 +442,14 @@ def parse_offset_get(s: str) -> Tuple[float, int]:
                            get_parser=float,
                            unit='deg')
 
+        self.add_parameter(
+            "complex_voltage",
+            label="Voltage",
+            get_cmd=self._get_complex_voltage,
+            unit="V",
+            vals=ComplexNumbers(),
```
jenshnielsen, line 450: Should this have a docstring. Especially with respect to x,y or mag + phase

qcodes/instrument_drivers/stanford_research/SR830.py
```diff
@@ -746,6 +754,10 @@ def _set_units(self, unit: str) -> None:
         for param in [self.X, self.Y, self.R, self.sensitivity]:
             param.unit = unit
 
+    def _get_complex_voltage(self) -> complex:
+        x, y = self.get_values("X", "Y")
```
astafan8, line 758: sr830 does not implement the ``get_values`` method (see it's implementation in sr860 https://github.com/QCoDeS/Qcodes/blob/71823308a3a9235580b09b29077eeccb6419e6a2/qcodes/instrument_drivers/stanford_research/SR86x.py#L1015), so the ``get_values`` equivalent needs to be implemented for sr830 (perhaps it's the same visa command as in sr860? would need to check the manual)

trevormorgan: @astafan8 good catch, looks like the 830 does have the same `SNAP?` command that the 860 uses. I pulled it over but we should probably test the hardware before merging.

qcodes/instrument_drivers/stanford_research/SR830.py
trevormorgan: Good point, I'll add in a doc string.

Akshita07:
Hi all, 
I found out that `snap` method has already been implemented which does what `get_values` method was doing. Hence, I have removed the `get_values` method and used `snap` where it was needed. This driver is also tested with actual instrument and notebook has also been updated. Driver works good!

astafan8:
@trevormorgan unfortunately, in the commit https://github.com/QCoDeS/Qcodes/pull/3078/commits/785677df21aafdd8bd1c8c42cec0d7afca78ba52 the whole file gets reformatted, was that intentional? the thing with formatting files completely is that the usefulness of git history is reduced because it's more difficult to see when the lines were changed and for what reason if commits with formatting are in the way. To still keep the formatting effort going, Jens is using "darker" which only formats the diff of the pull-request (but not the changed file completely). For this PR I think the solution is simple - to rebase and get rid of this commit that includes the formatting because get_values is not needed anyway, and further commits use the existing snap method and all is good :)

trevormorgan:
> @trevormorgan unfortunately, in the commit [785677d](https://github.com/QCoDeS/Qcodes/commit/785677df21aafdd8bd1c8c42cec0d7afca78ba52) the whole file gets reformatted, was that intentional? the thing with formatting files completely is that the usefulness of git history is reduced because it's more difficult to see when the lines were changed and for what reason if commits with formatting are in the way. To still keep the formatting effort going, Jens is using "darker" which only formats the diff of the pull-request (but not the changed file completely). For this PR I think the solution is simple - to rebase and get rid of this commit that includes the formatting because get_values is not needed anyway, and further commits use the existing snap method and all is good :)

Sorry this wasn't intentional, I have black set to auto format when saving and forgot to disable. Sounds good about rebasing and using darker to format only the changed sections.

jenshnielsen (approved):

PR #3089, opened 2021-06-04
DelegateInstrument/channel_support

jana-d:
- Add _create_and_add_channels to DelegateInstrument to allow adding instrument channels.
- Make input parameters `parameters` and `channels` optional.
- Add example notebook to docs/examples/driver_examples and the required MockCustomChannel to tests/instrument_mocks.

Fixing typing and adding tests are still on the to-do list.

@jenshnielsen: What are your thoughts on adding a channel instantiated with a custom class? Should such feature be also added to parameters?


qcodes/instrument/delegate/delegate_instrument.py
```diff
@@ -95,24 +99,32 @@ def __init__(
         self,
         name: str,
         station: Station,
-        parameters: Union[Mapping[str, Sequence[str]], Mapping[str, str]],
+        parameters: Optional[Union[Mapping[str, Sequence[str]], Mapping[str, str]]] = None,
+        channels: Optional[Union[Mapping[str, Mapping[str, Any]], Mapping[str, str]]] = None,
```
jenshnielsen, line 103: One concern is that if anyone passes arguments after this one as positional arguments that will break with this change. 
We could consider making all these arguments kwargs only by adding a `*` (probably after station)
@astafan8 what do you think

qcodes/instrument/delegate/delegate_instrument.py
```diff
@@ -95,24 +99,32 @@ def __init__(
...
+        if parameters is not None:
+            self._create_and_add_parameters(
+                station=station,
+                parameters=parameters,
+                setters=setters or {},
+                units=units or {}
+            )
         self._initial_values = initial_values or {}
         if set_initial_values_on_load:
             self.set_initial_values()
```
jenshnielsen, line 120: I guess the order of this code implies that initial_values only work for parameters and not for parameters on the channels? I don't know if that is an issue? Perhaps we should at least document that

jana-d: Or maybe change the order so that it works on channel parameters.

qcodes/instrument/delegate/delegate_instrument.py
```diff
@@ -264,6 +275,99 @@ def _create_and_add_parameter(
...
+                channel is added as is using ``self.add_submodule``.
+
+        """
+
+        channel_wrapper = None
+        if 'type' in channels.keys():
+            module_name = '.'.join(channels['type'].split('.')[:-1])
+            instr_class_name = channels['type'].split('.')[-1]
+            module = importlib.import_module(module_name)
+            channel_wrapper = getattr(module, instr_class_name)
```
jenshnielsen, line 311: I think this can be written a bit cleaner as

```suggestion
        channel_type = channels.pop("type", None)
        if channel_type is not None:
            channel_type_elems = channel_type.split(".")
            module_name = '.'.join(channel_type_elems[:-1])
            instr_class_name = channel_type_elems[-1]
            module = importlib.import_module(module_name)
            channel_wrapper = getattr(module, instr_class_name)
```

Then we can also drop the check for ken != type below

qcodes/instrument/delegate/delegate_instrument.py
```diff
@@ -264,6 +275,99 @@ def _create_and_add_parameter(
...
+                with 'channel' key containing the path of the channel and any
+                other input arguments taken by channel_wrapper.
+            channel_wrapper: Optional class to construct the channel. If none
+                given, the channel is added as is using ``self.add_submodule``.
+
+        """
+
+        if isinstance(input_params, str):
+            try:
+                instrument_name, channel_name = input_params.split('.')
```
jenshnielsen, line 346: Are we sure this is always of the form `inst.channel` it could be `inst.channel.subchannel`

jana-d: Good point. Not sure, it's probably best to use parse_instrument_path in that case. I'll make the change.

qcodes/instrument/delegate/delegate_instrument.py
astafan8: either way it would be a breaking change (with varying degree of breakage), so including the ``*`` will make it more evident which is generally nice. however, there are not so many users of this feature (yet) hence we can probably skip adding the ``*`` at this point. so i'd vote for not adding the star but do note the "breaking change"

qcodes/instrument/delegate/delegate_instrument.py
```diff
@@ -157,7 +186,7 @@ def set_initial_values(self, dry_run: bool = False) -> None:
                         )
                     else:
                         parent, name = self, path
-                    print(parent, name, value)
```
jenshnielsen, line 160: Should we just remove this print ?

qcodes/tests/delegate/test_device.py
```diff
@@ -36,3 +39,27 @@ def test_device_meas(station, chip):
...
             datasaver.dataset.to_pandas_dataframe_dict()["device1_drain"]) == 10
+
+
+def test_device_with_channels(chip, station):
+    device = chip.channel_device
+
+    assert device.gate_1 == station.dac.ch01
+    assert device.gate_1.voltage.post_delay == 0.01
+    assert device.readout.source_parameters == (station.lockin.phase,)
+    assert device.readout() == 1e-5
```
jenshnielsen, line 50: Would it make sense to extend the tests to also ensure that you can set with one parameter and get the sane result from the other one and the other way around

jana-d: Good point. Done!

qcodes/instrument/delegate/delegate_instrument.py
jana-d: It's gone.

astafan8 (commented):
left a few naming/readability comments.

but the biggest question is - it's assumed that all channels would have the same channel wrapper class (if any) - is that really true? i'd expect that each channel can define its wrapper if it so wishes and for convenience all channels can get the same channel_wrapper type specified at "channels:" root level if that's appropriate. see my more specific comment about below

qcodes/tests/delegate/data/chip.yml
```diff
@@ -5,21 +5,42 @@ instruments:
...
+            gate_1: dac.ch01
+
+        channel_device_custom:
+          parameters:
+            readout: lockin.phase
+          channels:
+            type: qcodes.tests.instrument_mocks.MockCustomChannel
+            gate_1:
+              channel: dac.ch01
+              current_valid_range: [-0.5, 0]
```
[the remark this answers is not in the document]
jana-d: Thanks for pointing this out @astafan8. I was thinking about but was not sure about which use cases should be supported. I will add this option.

qcodes/instrument/delegate/delegate_instrument.py
```diff
@@ -329,19 +341,44 @@ def _create_and_add_channel(
...
     def __repr__(self) -> str:
         params = ", ".join(self.parameters.keys())
         return f"DelegateInstrument(name={self.name}, parameters={params})"
+
+
+def _get_channel_wrapper_class(
+    channel_type: str,
+) -> Optional[Type[InstrumentChannel]]:
+    """Get channel class from string specified in yaml."""
+    if channel_type == "None":
```
astafan8, line 377: for the sake of simplicity of this function, can it's input type be Optional[str]? :) it will be same number of lines but cleaner from the typing perspective, meaning that if the argument is a string then it's a "üsable" one

astafan8 (approved):

qcodes/instrument/delegate/delegate_instrument.py
jana-d: I had that first but mypy was complaining. The type of input_params.pop("type", None) is not Optional[str] (but Union[str, Mappingp[str, Any]). So I cast the popped value to str and let _get_channel_wrapper_class take just a string. Is there a better solution to fix the typing issue?

astafan8: (calling ``str`` on an object is not a cast, but a true conversion :) cast would be calling `cast` from `typing` )

aaaah, that's for the channel_type_global case...

hmm, first thing i see is to move the Union in type definition of channels argument of _create_and_add_channels from outside to the key of the outer Mapping: ``Union[Mapping[str, Mapping[str, Any]], Mapping[str, str]]`` -> ``Mapping[str, Union[str, Mapping[str, Any]]]``. have you tried that?

and then in the pop you will have to check that the popped value is not a Mapping (because the "type" key should be a string, not a dict), and then, if mypy still does not understand, then use ``cast`` function to cast the resulting value into optional string. i realize that this will be extra lines of code, but it will be a more readable code because the very flexible "channels" argument is parsed as early as possible into specifics, and then the rest of the code, e.g. _get_channel_wrapper_class becomes much simpler (in that it does not need the ``"None"`` workaround)

jana-d: Thanks @astafan8! Your suggestion works :).

PR #3094, opened 2021-06-08
In Memory only Dataset

jenshnielsen:
Implements an in memory dataset that can be used to measure without writing the raw data to the db.
Furthermore this enables you to reload a dataset from a netcdf file and then write the metadata back to a db file and a plot it / explore it in other ways

- [x] Write a changelog entry


qcodes/dataset/data_set_cache.py
```diff
@@ -194,6 +196,20 @@ def to_xarray_dataset(self) -> xr.Dataset:
...
+        single_file: bool = False,
+        single_file_name: Optional[str] = None,
+    ) -> None:
+        dfdict = self.to_pandas_dataframe_dict()
+        dataframe_to_csv(
+            dfdict=dfdict,
+            path=path,
+            single_file=single_file,
+            single_file_name=single_file_name,
+        )
```
jenshnielsen, line 211: I don't think we need / want this as a public method

jenshnielsen:
Things to figure out.

- [ ] If a dataset is loaded from netcdf should we add it to the runs table in the current db
- [ ] Should there be a function to add a dataset including the results table back to the db
- [ ] Currently changes to metadata of any kind will not be written to the db except when the run is marked started and completed. It will also not be added to exported data after the data has been exported.
- [ ] How do we handle the `__len__` attribute. This is somewhat sqlite specific since it counts the number of sqlite rows. 
- [ ] Upgrade create_runs to somehow mark the runs table for not containing the data but it being external

qcodes/dataset/data_set_cache.py
jenshnielsen: This was written because I confused the different kind of interfaces.

We have `to_xarray ...` and `to_pandas...` so it seemed to make sense to have `to_csv` but actually these are 2 different things. One is an in memory object type and the other one a file format. 

We have for now defined that the file formats should be accessed via export( export_type="csv") so it seems like we should just loose this method. We do have a `write_data_to_text_file` on the DataSet class which should perhaps be deprecated? 

The dataset has `write_data_to_text_file` so perhaps we should implement this. The

qcodes/dataset/data_set.py
```diff
@@ -358,7 +342,7 @@ def captured_run_id(self) -> int:
         return run_id
 
     @property
-    def path_to_db(self) -> str:
+    def path_to_db(self) -> Optional[str]:
```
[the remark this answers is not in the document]
jenshnielsen: no not really since the db file may not exist when loading from netcdf

qcodes/dataset/data_set.py
```diff
@@ -1775,19 +1600,30 @@ def load_by_counter(counter: int, exp_id: int,
...
-    d = DataSet(conn=conn, run_id=run_id)
+    guid = get_guid_from_run_id(conn, run_id)
+
+    if guid is None:
+        raise ValueError(f"Run with run_id {run_id} does not exist in the database")
+    result_table_name = _get_result_table_name_by_guid(conn, guid)
+    if _check_if_table_found(conn, result_table_name):
+        d: DataSetProtocol = DataSet(conn=conn, run_id=run_id)
+    else:
+        d = DataSetInMem._load_from_db(conn=conn, guid=guid)
```
jenshnielsen, line 1626: Check if this and/or the other methods are leaking connections

qcodes/dataset/data_set.py
```diff
@@ -1711,61 +1489,116 @@ def load_by_run_spec(*,
...
+            d = load_by_guid(matched_guids[0], internal_conn)
+        elif len(matched_guids) > 1:
+            print(generate_dataset_table(matched_guids, conn=internal_conn))
+            raise NameError(
+                "More than one matching dataset found. "
+                "Please supply more information to uniquely"
+                "identify a dataset"
+            )
+        else:
+            raise NameError(f"No run matching the supplied information " f"found.")
```
astafan8, line 1517:
```suggestion
            raise NameError(f"No run matching the supplied information found.")
```

qcodes/dataset/data_set_in_memory.py
```diff
@@ -0,0 +1,859 @@
...
+    @classmethod
+    def _load_from_netcdf(
+        cls, path: Union[Path, str], path_to_db: Optional[Union[Path, str]] = None
+    ) -> DataSetInMem:
+        """
+        Create a in memory dataset from a netcdf file.
+        The netcdf file is expected to contain a QCoDeS dataset that
+        has been exported using the QCoDeS netcdf export functions.
+
+
```
astafan8, line 206:
```suggestion
```

qcodes/dataset/data_set_in_memory.py
```diff
@@ -0,0 +1,859 @@
...
+        Args:
+            path: Path to the netcdf file to import.
+            path_to_db: Optional path to a database where this dataset may be
+             exported to. If not supplied the path can be given at export time
+             or the dataset exported to the default db as set in the QCoDeS config.
```
astafan8, line 211:
```suggestion
            path_to_db: Optional path to a database where this dataset may be
                exported to. If not supplied the path can be given at export time
                or the dataset exported to the default db as set in the QCoDeS config.
```

qcodes/dataset/data_set_in_memory.py
```diff
@@ -0,0 +1,859 @@
...
+
+        Returns:
+            The loaded dataset.
+        """
+        # in the code below flots and ints loaded from attributes are explicitly casted
```
astafan8, line 216:
```suggestion
        # in the code below floats and ints loaded from attributes are explicitly casted
```

qcodes/dataset/data_set_in_memory.py
```diff
@@ -0,0 +1,859 @@
...
+
+        metadata_keys = (
+            set(loaded_data.attrs.keys()) - set(RUNS_TABLE_COLUMNS) - non_metadata
+        )
+        metadata = {}
+        for key in metadata_keys:
+            data = loaded_data.attrs[key]
+            if isinstance(data, np.ndarray) and data.size == 1:
+                data = data[0]
+            metadata = {key: data}
```
astafan8, line 266: a typo, right? we want to populate metadata for all keys, right?)
```suggestion
            metadata[key] = data
```

qcodes/dataset/data_set_in_memory.py
```diff
@@ -0,0 +1,859 @@
...
+
+        if self.pristine:
+            self._perform_start_actions()
+
+    @property
+    def pristine(self) -> bool:
+        """Is this :class:`.DataSet` pristine?
+
+        A pristine :class:`.DataSet` has not yet been started,
+        meaning that parameters can still be added and removed, but results
```
astafan8, line 422:
```suggestion
        """Is this :class:`.DataSetInMem` pristine?

        A pristine :class:`.DataSetInMem` has not yet been started,
        meaning that parameters can still be added and removed, but results
```

qcodes/dataset/data_set_in_memory.py
```diff
@@ -0,0 +1,859 @@
...
+        can not be added.
+        """
+        return self._run_timestamp_raw is None and self._completed_timestamp_raw is None
+
+    @property
+    def running(self) -> bool:
+        """
+        Is this :class:`.DataSet` currently running?
+
+        A running :class:`.DataSet` has been started,
```
astafan8, line 432:
```suggestion
        Is this :class:`.DataSetInMem` currently running?

        A running :class:`.DataSetInMem` has been started,
```

qcodes/dataset/data_set_in_memory.py
```diff
@@ -0,0 +1,859 @@
...
+            self._run_timestamp_raw is not None
+            and self._completed_timestamp_raw is None
+        )
+
+    @property
+    def completed(self) -> bool:
+        """
+        Is this :class:`.DataSet` completed?
+
+        A completed :class:`.DataSet` may not be modified in
```
astafan8, line 445:
```suggestion
        Is this :class:`.DataSetInMem` completed?

        A completed :class:`.DataSetInMem` may not be modified in
```

qcodes/dataset/data_set_in_memory.py
```diff
@@ -0,0 +1,859 @@
...
+        any way.
+        """
+        return self._completed_timestamp_raw is not None
+
+    def mark_completed(self) -> None:
+        """
+        Mark :class:`.DataSet` as complete and thus read only and notify the subscribers
```
astafan8, line 452:
```suggestion
        Mark :class:`.DataSetInMem` as complete and thus read only and notify the subscribers
```

qcodes/dataset/data_set_in_memory.py
```diff
@@ -0,0 +1,859 @@
...
+        """
+        if self.completed:
+            return
+        if self.pristine:
+            raise RuntimeError(
+                "Can not mark DataSet as complete before it "
```
astafan8, line 458:
```suggestion
                "Can not mark a dataset as complete before it "
```

qcodes/dataset/data_set_in_memory.py
```diff
@@ -0,0 +1,859 @@
...
+    def cache(self) -> DataSetCacheInMem:
+        return self._cache
+
+    @property
+    def export_info(self) -> ExportInfo:
+        return self._export_info
+
+    def _enqueue_results(self, result_dict: Mapping[ParamSpecBase, np.ndarray]) -> None:
+        """
+        Enqueue the results into self._results
```
astafan8, line 612:
```suggestion
        Enqueue the results, for this dataset directly into cache
```

qcodes/dataset/data_set_in_memory.py
```diff
@@ -0,0 +1,859 @@
...
+            )
+        if self.completed:
+            raise CompletedError(
+                "This DataSet is complete, no further " "results can be added to it."
+            )
+
+    def __len__(self) -> int:
+        """
+        The in memory dataset does not have a concept of sqlite rows
+        so the length is represented by the maximum number of datapoints in any dataset.
```
astafan8, line 759:
```suggestion
        so the length is represented by the number of all datapoints,
        summing across parameter trees.
```

qcodes/dataset/data_set_in_memory.py
```diff
@@ -0,0 +1,859 @@
...
+
+def load_from_netcdf(
+    path: Union[Path, str], path_to_db: Optional[Union[Path, str]] = None
+) -> DataSetInMem:
+    """
+    Create a in memory dataset from a netcdf file.
+    The netcdf file is expected to contain a QCoDeS dataset that
+    has been exported using the QCoDeS netcdf export functions.
+
+
```
astafan8, line 849:
```suggestion
```

qcodes/dataset/data_set_in_memory.py
```diff
@@ -0,0 +1,859 @@
...
+    Args:
+        path: Path to the netcdf file to import.
+        path_to_db: Optional path to a database where this dataset may be
+         exported to. If not supplied the path can be given at export time
+         or the dataset exported to the default db as set in the QCoDeS config.
```
astafan8, line 854:
```suggestion
        path_to_db: Optional path to a database where this dataset may be
            exported to. If not supplied the path can be given at export time
            or the dataset exported to the default db as set in the QCoDeS config.
```

qcodes/dataset/data_set_protocol.py
```diff
@@ -162,10 +218,286 @@ def export_info(self) -> ExportInfo:
...
+        in a human-readable format
+
+        If the run (or the dataset) is not completed, then returns None.
+
+        Consult with ``time.strftime`` for information about the format.
+        """
+        return raw_time_to_str_time(self.completed_timestamp_raw, fmt)
+
+
+class DataSetTypes(str, Enum):
```
astafan8, line 500: shall the name of the class be singular in the same spirit as Color.RED and SmuError.VoltageComplianceHit?
```suggestion
class DataSetType(str, Enum):
```

astafan8 (approved):
Ready to ship! Awesome work! (there's only one important comment, others are less)

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "qcodes/tests/dataset/fixtures/db_files"]
 	path = qcodes/tests/dataset/fixtures/db_files
 	url = https://github.com/QCoDeS/qcodes_db_fixtures.git
+	branch = netcdf_files
```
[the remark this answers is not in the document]
jenshnielsen: Yes

qcodes/dataset/data_set.py
jenshnielsen: thanks really wish darker/black would stop generating this

qcodes/dataset/data_set_in_memory.py
jenshnielsen: looks wrong. Check missing test coverage (at least 2 pieces of metadata in test

qcodes/dataset/data_set_protocol.py
jenshnielsen: yes probably better.

jenshnielsen:
- [ ] should we upgrade exported file names to include guid?

astafan8:
> should we upgrade exported file names to include guid?

yes. but can also be in a separate PR.

PR #3095, opened 2021-06-08
Loop dataset: do not use hardcoded value to look for default parameter name

peendebak:
- Update `default_parameter_name` to not use a hardcoded value for searching. This causus problems for datasets where we have a dependent variable called `qubit_fractions` and a dependent variable such as `amplitude_variation`.


astafan8 (commented):
@peendebak question out of curiosity - would this be an option to use the ``metadata['default_parameter_name']`` to overrule the default `'amplitude'`?

peendebak:
> @peendebak question out of curiosity - would this be an option to use the `metadata['default_parameter_name']` to overrule the default `'amplitude'`?

That is possible, but a lot of work. Setting the `default_parameter_name` in our scan functions is possible, but these datasets get passed, modified, sliced, averaged, etc. in post-processing and most of the time the `metadata` is not fully copied. With hindsight, setting the default value to some value is a mistake I believe.

peendebak:
> @peendebak thanks for explaining. then since it's a breaking change, could you add a note about it to the changelog of the next release (we ask PR authors to do that for breaking changes so that the person who's going to do a release does not have to remember to check for that)?

@astafan8 Where should I add the changelog entry?

astafan8:
> Where should I add the changelog entry?

create a docs/changes/0.27.0.rst file, and add it there (and also mention this new rst file in the docs/changes/index.rst . (see and example here https://github.com/QCoDeS/Qcodes/pull/3089 )

eendebakpt:
@astafan8 I added a changelog

astafan8 (approved):

PR #3096, opened 2021-06-08
Use threaded setting from config in dond functions, and make process_params_meas public

ThorvaldLarsen:
Create a setting for use_threads in the config to be used in `process_params_meas` for doNd functions. Allows control of the setting in the config file rather than a required input at runtime. 
Additionally makes the measure function public for external use of threaded parameter calls. This could be split out to a separate PR but included here in case there are no objections/concerns with this change.

Bug fix: moved dataset reference up in new dond to handle keyboardinterrupts similar to other functions in file.

### DIFF
--- a/qcodes/configuration/qcodesrc_schema.json
+++ b/qcodes/configuration/qcodesrc_schema.json
@@ -306,7 +306,7 @@
                     "default": 5.0,
                     "description": "How often should data be written to disk (s)"
                 },
-                "use_threaded": {
+                "use_threads": {
                         "type": "boolean",
                         "default": false,
                         "description": "Should instruments be addresesed in parallel for process_params_meas() in doNd measurement"
--- a/qcodes/configuration/qcodesrc_schema.json
+++ b/qcodes/configuration/qcodesrc_schema.json
@@ -307,6 +307,11 @@
                     "default": 5.0,
                     "description": "How often should data be written to disk (s)"
                 },
+                "use_threaded": {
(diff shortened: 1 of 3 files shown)


jenshnielsen:
@ThorvaldLarsen I think this is how the cutoff_percentile was intended to be specified https://github.com/QCoDeS/Qcodes/pull/3097

jenshnielsen (approved):

qcodes/configuration/qcodesrc_schema.json
```diff
@@ -214,7 +214,6 @@
                             "description": "Upper and lower percentage of datapoints that may maximally be discarded by the auto color scale. For example for [1,1] the auto color scale will in a set of 10 0000 points not clip more than the 100 points of lowest and highest value.",
                             "type": "array",
                             "items": {
-                                "type": "number",
```
jenshnielsen, line 217: I suggest to revert this and go with the change in https://github.com/QCoDeS/Qcodes/pull/3097

ThorvaldLarsen: Good catch. I was just following pylint gving it as an error. My bad.

jenshnielsen: No worries the original syntax is wrong so this fix is better than the original file but #3097 is even better

astafan8 (approved):
awesome! thank you!

PR #3097, opened 2021-06-08
Fix invalid array syntax in json schema 

jenshnielsen:
The current format is invalid. It should either be just a single item type or like this pr suggest a list of item types which defines exactly the type of each element

Since cutoff_percentile always has two numbers this seems like the correct format.

See https://json-schema.org/understanding-json-schema/reference/array.html for some details 

This also adds a pre-commit hook that check that json is valid which would have caught this. I had to exclude asv.conf.json since it contains comments which is not valid json


astafan8 (approved):
great catch!

Issue #1151, opened 2018-06-17
Extract doc strings from dynanically added paramters


jenshnielsen:
Currently all parameters are dynamically added to instruments. This means that the autodoc documentation of the paramters does not contain the parameters. It would be great if we can find a way to include these docstrings in the autodoc generated documentation

caenrigen:
Hi, @jenshnielsen 

Is there any solution to this problem so far? And perhaps to as well https://github.com/QCoDeS/Qcodes/issues/32 that is somewhat related when, e.g.,  documenting subclasses of `Instrument` s

jenshnielsen:
Not at the moment. There does not seem to be a general solution to this in Sphinx. I think the best option would be to replace the use of add_parameter with an actual class instantiated as a member of the class. We are not actively working on this

With regards to #32 there was a suggestion in https://github.com/QCoDeS/Qcodes/pull/2816 unfortunately that hit some design issues

caenrigen:
@jenshnielsen thank you for the reply and information!

After several hours of prototyping I came up with a solution that solves #32 (and  probably #2816 as well).

At high level it involves:

- Extra (non-breaking) functionality in the base `Instrument` (or `_BaseInstrumet`) class.
- A relatively simple but specific way of adding parameters to instrument classes (older drivers could be upgraded to benefit from this as well).
- A bit of `sphinx` black magic.

## General idea

- Store the information needed to instantiate the parameters at class creation time --> use a dummy private method
- The dummy private method name and docstring will serve the creation of the `Parameter` s
- Use decorator to flag method for "conversion" to parameter.
- Sphinx is able to pick up the docstring as well --> autodoc possible.
- "Hack" sphinx to replace method name with the name of the parameter added at Intrument's instantiation time.
- Glue everything together.

## Sphinx docs result

![Screenshot 2021-05-28 at 14 31 47](https://user-images.githubusercontent.com/31376402/119987128-f9c71080-bfc4-11eb-954e-fccf069ae9c4.png)

## Usage: Instruments instantiation

```python
from quantify.utilities.my_mod import MyInstrumentDriver, SubMyInstrumentDriver

instr = MyInstrumentDriver("instr")
instr.freq(10)
instr.print_readable_snapshot()
print("\ninstr.time.label: ", instr.time.label)
```

```bash
instr:
	parameter value
--------------------------------------------------------------------------------
IDN  :	None 
freq :	10 (Hz)
time :	3 (s)

[...]

PR #3098, opened 2021-06-08
A new way of adding parameters that allows overriding inherited parameters (and sphinx auto documentation)

### LINKED ISSUE #32: Overwriting parameters in inherited instruments
Currently parameters are added to an instrument in the **init**. Once a parameter is added it is not possible to overwrite this by using another add_parameter command. When inheriting an instrument you generally want to overwrite inherited functions as well as some parameters. This is currently not possible. 

The current workaround would be to overwrite the entire **init** and copy paste all the add_parameter commands that were in the parent class. I think it makes sens to run super().**init**(...) and then only overwrite those parameters that are overwritten.

caenrigen:
This is a follow up from the discussion in #1151, marked as Draft for now to gather feedback.

Fixes #32
Also sets the infrastructure for fixing #1151 

- Provides a new way of adding parameters to instrument.
- Allows for `Parameters` inheritance with overriding capabilities (#32).
- Paves the way for sphinx auto documentation of member parameters (with some hacks, see #1151). 

---

Sphinx result + example, please note the `InstanceAttr` in the signature of the added parameters (and source code):

Screenshot 2021-06-09 at 01 42 10

<details><summary>Source code</summary>

```python
# Repository: https://gitlab.com/quantify-os/quantify-core
# Licensed according to the LICENCE file on the master branch
"""test feature"""
from qcodes import ManualParameter, Parameter
from qcodes import validators
import qcodes.instrument.base as bs


class MyInstrumentDriver(bs.Instrument):
    r"""
    MyInstrumentDriver docstring

    .. jupyter-execute::

        from quantify.utilities.my_mod import MyInstrumentDriver

        instr = MyInstrumentDriver(name="instr", special_par=8)
        instr.freq(10)
        instr.print_readable_snapshot()
        print("\ninstr.time.label: ", instr.time.label)

    """

    def __init__(self, special_par, *args, **kwargs):
        self._call_add_params_from_decorated_methods = False
        super().__init__(*args, **kwargs)
        self._freq = special_par

        self._add_params_from_decorated_methods()

    @bs.add_parameter
    def _parameter_time(  # NB the parameter name will be just `time`
        self,
        parameter_class=ManualParameter,
        initial_value=3,
        unit="s",
        label="Time",
        vals=validators.Numbers(),
    ):
        """Docstring of :code:`time` parameter"""

    @bs.add_parameter
    def _parameter_freq(  # NB the parameter name will be just `freq`
        self,
        parameter_class=Parameter,
        unit="Hz",
        label="Frequency",
        vals=validators.Numbers(),
        # `self._freq` will be passed to `Parameter`
        initial_value=bs.InstanceAttr("_freq"),
        # `self._set_freq` will be passed to `Parameter`
        set_cmd=bs.InstanceAttr("_set_freq"),
        # `self._get_freq` will be passed to `Parameter`
        get_cmd=bs.InstanceAttr("_get_freq"),
    ):
        """Docstring of :code:`freq` parameter"""

    def _set_freq(self, value):
        self._freq = value

    def _get_freq(self):
        return self._freq


class SubMyInstrumentDriver(MyInstrumentDriver):
    r"""Same as MyInstrumentDriver but overriding a parameter

    .. jupyter-execute::

        from quantify.utilities.my_mod import SubMyInstrumentDriver

        sub_instr = SubMyInstrumentDriver(name="sub_instr", special_par=99)
        sub_instr.time(sub_instr.time() * 2)
        sub_instr.print_readable_snapshot()
        print("\nsub_instr.time.label: ", sub_instr.time.label)

    """

    @bs.add_parameter
    def _parameter_time(  # NB the parameter name will be just `time`
        self,
        parameter_class=ManualParameter,
        initial_value=7,
        unit="s",
        label="Time long",
        vals=validators.Numbers(),
    ):
        """Docstring of :code:`time` parameter in the subclass"""
```
</details>

---

To do

- [x] Add tests.
- [x] Document this new way of adding parameters.
- [x] Document how to "hack" Sphinx in order to document these parameters automatically.
- [x] Dealing with the inheritance edge cases, e.g., should we still create the parameter in the parent class and `del` it later and replace it with a new one from the child class?

~~After this MRs one with docs on how to "hack" Sphinx in order to document these parameters automatically.~~

---

@astafan8 here is a Draft implementation.

@AdriaanRol @JordyGloudemans your feedback is welcome as well.

### DIFF
--- a/qcodes/sphinx_extensions/add_parameter.py
+++ b/qcodes/sphinx_extensions/add_parameter.py
@@ -167,13 +167,21 @@ def add_parameter_spec_to_docstring(app, what, name, obj, options, lines):
         ]
         for kw_name, par_obj in inspect.signature(obj).parameters.items():
             if kw_name != "self":
+                py_obj = par_obj.default
                 if kw_name == "parameter_class":
                     # create link to the parameter class
-                    mod = par_obj.default.__module__
-                    class_name = par_obj.default.__name__
+                    mod = py_obj.__module__
+                    class_name = py_obj.__name__
                     value = f":class:`{add_links}{mod}.{class_name}`"
                 else:
-                    value = f"*{par_obj.default!r}*"
+                    value = f"*{py_obj!r}*"
+
+                if isinstance(py_obj, type(lambda:0)):
+                    value = inspect.getsource(py_obj)
+                    value = value.replace("\n", "").replace("\r", "")
+                    value = value[:-1] if value[-1] == "," else value # source code might end with `,`
+                    value = value.rsplit("=", 1)[-1].strip()  # source code includes the name of the argument
+                    value = f":code:`{value}`"
 
                 if add_links == "" and kw_name == "vals":
                     kw_name = ":mod:`vals <qcodes.utils.validators>`"


qcodes/instrument/base.py
```diff
@@ -401,6 +410,94 @@ def validate_status(self, verbose: bool = False) -> None:
...
+        # decorated methods are required to have a specific prefix
+        has_parameter_prefix = lambda s: s.startswith(DECORATED_METHOD_PREFIX)
+
+        for obj_name in filter(has_parameter_prefix, dir(self)):
+            meth = getattr(self, obj_name)
+            # the `@add_parameter` decorator adds an attribute we check for here
+            if hasattr(meth, ADD_PARAMETER_ATTR_NAME):
+                kwargs = {
+                    key: special_kwargs_to_meth(self, val.default)
+                    for key, val in sorted(inspect.signature(meth).parameters.items())
```
astafan8, line 433: why is sorting done here?
```suggestion
                    for key, val in inspect.signature(meth).parameters.items()
```

qcodes/instrument/base.py
```diff
@@ -401,6 +410,94 @@ def validate_status(self, verbose: bool = False) -> None:
...
 
+    def _add_params_from_decorated_methods(self):
+
+        def special_kwargs_to_meth(self, kwarg_value):
+            """
+            Returns the attribute that has the name `kwarg_value.attr_name`
+            of this instance if the kwarg_value is of type `InstanceAttr`.
+            """
+            if isinstance(kwarg_value, InstanceAttr):
+                return getattr(self, kwarg_value.attr_name)
```
astafan8, line 421: perhaps it's useful to handle the case of the specified attribute not existing with an informative error message to users?

qcodes/instrument/base.py
```diff
@@ -401,6 +410,94 @@ def validate_status(self, verbose: bool = False) -> None:
...
+    in the :code:`_add_params_from_decorated_methods`.
+
+    Args:
+        method: The method to be wrapped and flagged to be converted to parameter.
+    """
+
+    if DECORATED_METHOD_PREFIX not in method.__name__:
+        raise ValueError(
+            f"Only methods prefixed with '{DECORATED_METHOD_PREFIX}' can be decorated "
+            f"with this decorator."
```
astafan8, line 485:
```suggestion
            f"with add_parameter decorator."
```

qcodes/instrument/base.py
```diff
@@ -401,6 +410,94 @@ def validate_status(self, verbose: bool = False) -> None:
...
+    """
+    A decorator function that wraps a method of an Instrument subclass such that the
+    new method will be converted into the corresponding :code:`param_class`
+    in the :code:`_add_params_from_decorated_methods`.
+
+    Args:
+        method: The method to be wrapped and flagged to be converted to parameter.
+    """
+
+    if DECORATED_METHOD_PREFIX not in method.__name__:
```
astafan8, line 482: shouldn't this be the condition? to be consistent with the _add_params_from_decorated_methods implementation
```suggestion
    if not method.__name__.startswith(DECORATED_METHOD_PREFIX):
```

qcodes/instrument/base.py
```diff
@@ -401,6 +410,94 @@ def validate_status(self, verbose: bool = False) -> None:
...
+        raise ValueError(
+            f"Only methods prefixed with '{DECORATED_METHOD_PREFIX}' can be decorated "
```
astafan8, line 484: ``!r`` repr mode will add quotes around the string ;)
```suggestion
            f"Only methods prefixed with {DECORATED_METHOD_PREFIX!r} can be decorated "
```

qcodes/instrument/base.py
```diff
@@ -401,6 +410,94 @@ def validate_status(self, verbose: bool = False) -> None:
...
+            f"with this decorator."
+        )
+
+    @wraps(method)  # preserves info like `__doc__` and signature
+    def kwargs_and_doc_container(self, *args, **kwargs):
+        raise RuntimeError(
+            f"Method not intended to be called.\n"
+            f"'{method.__name__}' is a special method used as information container "
```
astafan8, line 492:
```suggestion
            f"{method.__name__!r} is a special method used as information container "
```

qcodes/instrument/base.py
```diff
@@ -401,6 +410,94 @@ def validate_status(self, verbose: bool = False) -> None:
...
+"""
+
+ADD_PARAMETER_ATTR_NAME = "_add_parameter"
+"""
+A constant defining the name of the attribute set by
+:obj:`qcodes.instrument.base.add_parameter` decorator to flag a method that will be
+converted to parameter.
+"""
+
+def add_parameter(method):
```
astafan8, line 472: we use typing a lot in qcodes, so the new methods/classes/functions need to be typed. i think that using a Callable with a type variable (``_T = TypeVar("_T")``) should do the trick
```suggestion
def add_parameter(Callable[_T]) -> Callable[_T]:
```

astafan8 (commented):
looks good to me!

> Add tests

I suggest to create a module inside qcodes/instrument_drivers (next to test.py and device.py) with a few example instruments (for example, those from the PR description) that use this new feature. They can also be used later to apply the sphinx hacks and render as part of qcodes docs

> After this MRs one with docs on how to "hack" Sphinx in order to document these parameters automatically.

do you mean that you'd like to land the "new way of adding parameters" first, and then make another PR that would document/implement the sphinx hacks, right?

astafan8:
I have another separate question.

so this feature enables 1) override existing parameters in instrument driver subclasses such that the original parameter won't need even be instantiated, and 2) produce sphinx autodoc documentation for parameters defined in this new way. do you see a lot of widespread use of the capability 1)? i am not so sure about it, because it really depends on how the driver depends on that parameter that one wants to override - for example, if a parameter is used in the init logic of the driver in some way unreplaceable, then overriding may be tricky, and it might be important to keep the original parameter of the parent class so it gets added and used, and the in the subclass one can substitute it with a modified one by doing ``del self.parameters["param_to_override"]`` before ``self.add_parameter("param_to_override", ...)``.

if you agree with me on this, then the capability 2) in my perspective becomes more widespread, hence i'd suggest to add all code and documentation related to in in this same PR :) what do you think?

qcodes/instrument/base.py
caenrigen: Artifact from my previous prototyping. Thanks for spotting :)

caenrigen:
@astafan8 thank you very much for quick review, comments have been addressed.

---

> so this feature enables 1) override existing parameters in instrument driver subclasses such that the original parameter won't need even be instantiated, and 2) produce sphinx autodoc documentation for parameters defined in this new way. 

> do you see a lot of widespread use of the capability 1)?

It was the capability 1) that triggered me to dive into this. Not sure how widespread it will be, but has certainly important applications. The scenario that I believe was most relevant is for virtual instruments that serve as (snapshotable) information container, e.g.,`GenericTransmonQubit` that can serve as a overridable template for a `SpecialTransmonQubit`.
Perhaps @AdriaanRol can elaborate a bit more on this (next week when he is back from vacation).

> i am not so sure about it, because it really depends on how the driver depends on that parameter that one wants to override - for example, if a parameter is used in the init logic of the driver in some way unreplaceable, then overriding may be tricky, and it might be important to keep the original parameter of the parent class so it gets added and used, and the in the subclass one can substitute it with a modified one by doing `del self.parameters["param_to_override"]` before `self.add_parameter("param_to_override", ...)`.

Fair point, there might some complicated edge cases where this solution is not flexible enough, and perhaps it will never be able to cover every possible case, though it would be nice to be able to use this approach and make sphinx work for all or most scenarios. I am happy to make it more versatile to accommodate for a wide range of use cases, though it would be better to have an exemplary case before building a solution for non-yet-existing problem.

Maybe tagging me in future issues/PRs related to this can help mature this approach as people start using it.

---

A related potential problem that pops up in my mind is that the order in which parameters are added might matter as well. And if that is the case then the complexity of the inheritance/overriding scheme might grow.

> if you agree with me on this, then the capability 2) in my perspective becomes more widespread, hence i'd suggest to add all code and documentation related to in in this same PR :) what do you think?

I am up for this. Changed the description and todo list.

---

For now I think I will wait to gather some more feedback on the implementation and the issues you raised. The solution for these potential problems is not very clear for now.

astafan8:
> perhaps it will never be able to cover every possible case, though it would be nice to be able to use this approach and make sphinx work for all or most scenarios

agree. happy to hear more about the ``*Qubit`` virtual instruments use case and their overriding.

> The solution for these potential problems is not very clear for now.

indeed.

> I am up for this. Changed the description and todo list.

great! then i'll review after those items are done, ok?

caenrigen:
> great! then i'll review after those items are done, ok?

Yes, deal! I will tag you when it is ready for a new review :)

AdriaanRol:
> @AdriaanRol @JordyGloudemans your feedback is welcome as well.

Thank you for making this ❤️ , you know how much I have been looking forward to this. 

I'll try to focus on the "user" side of how this change would affect the writing and readability of instrument drivers as well as the resulting docstrings and how easy they are to understand. I'll try to minimize my comments on the implementation details as the sphinx+decorator magic is a bit of well, magic, to me. 

## Example instrument "new-style"

> I suggest to create a module inside qcodes/instrument_drivers (next to test.py and device.py) with a few example instruments (for example, those from the PR description) that use this new feature. They can also be used later to apply the sphinx hacks and render as part of qcodes docs

Issue #1151 contains a nice example of how an instrument driver looks in the new style. I assume there is probably some subtle differences between the initial design and the current implementation. Could you add an example driver to this MR and maybe also include the sphinx hack so that it is rendered in the docs? 

Maybe the best example would be to rewrite the SGS100A microwave source driver using this new style. The reason is that this is the simplest driver, and it would serve as an example on how to write drivers "new-style". However, I see two issues with this. 1. it would delay/block the acceptance of this MR because the new driver needs to be tested, and 2. it enforces a standard before it is tested in working nice. So we should probably postpone the replacing to a later date. 

## Tweaks to implementation of instrument "new-style"
I'll leave some comments here on the example from #1151. 


```python
from qcodes import Instrument, ManualParameter, Parameter
from qcodes import validators

class MyInstrumentDriver(ModifiedInstrumentClass):
    """
    MyInstrumentDriver docstring
    """

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self._freq = None

    @add_parameter
    def _parameter_time(  # NB the parameter name will be just `time`
        self,
        parameter_class=ManualParameter,
        initial_value=3,
        unit="s",
        label="Time",
        vals=validators.Numbers(),
    ):
        """Docstring of :code:`time` parameter"""
```
suggestion, the @add_parameter decorator in combination with the `_parameter_` prefix seems double to me. For a user (someone writing a driver) it would be much simpler to directly use the syntax `@add_parameter \n def time(.... `. For the behind the scenes implementation, this would mean that the `time` attribute would be removed and replaced the moment the add_parameter decorator is invoked. 

Other than that, I really like the syntax. 

I've also left a comment in the `add_parameter` decorator. 


```python

    @add_parameter
    def _parameter_freq(  # NB the parameter name will be just `freq`
        self,
        parameter_class=Parameter,
        unit="Hz",
        label="Frequency",
        vals=validators.Numbers(),
        set_cmd="_set_freq",  # `self._set_freq` will be passed to `Parameter`
        get_cmd="_get_freq",  # `self._get_freq` will be passed to `Parameter`
    ):
        """Docstring of :code:`freq` parameter"""
```

```python
    def _set_freq(self, value):
        self._freq = value

    def _get_freq(self):
        return self._freq
```
Good that you included these in the example. 


**praise** in general, I really like this way of writing drivers from a usability perspective. To me it seems very much in line with the original philosophy of QCoDeS of being easy to write and read for physicists (which does require hiding the magic, but that is fine), but I'll leave that to @astafan8 as the official QCoDeS representative in this MR. 

## Feedback on how docs are rendered 

based on the screenshot above. 

**praiseI** think the fact that parameters now show up in the docs is a vast improvement over not being able to document them at all 👍 

I have some questions as I am not as familiar with the sphinx magic you had to invoke to get this to work. At first sight, the parameter documentation seems a bit off and hard to read. This is because all is crammed into the call signature, and only the docstring shows. 

A big improvement (in my opinion) is that now you don't get the massive docstring of the parameter base class when. This used to bloat the readability of individual parameters for the user. 

I had a look at some of our quantify docs (see screenshot) to see how it works for "normal" objects. 

image

Unless it is trivially easy, I suggest you do not implement these things, I'm mostly trying to assess if this would be possible in the future. 
**question** would it be possible to have "parameter" appear as an italics prefix similar to the "class method" in the screenshot I showed. 
**question** would it be possible to turn the arguments into a list with bold-faced names and then the text after it or is it better to do this using the docstring. **suggestion** if the second is preferred, let's include that in the example docstring (i.e., what the settable is, a link to the set_cmd and get_cmd, the unit, etc.

qcodes/instrument/base.py
```diff
@@ -401,6 +411,143 @@ def validate_status(self, verbose: bool = False) -> None:
...
+converted to parameter.
+"""
+
+_ParamSpec = TypeVar("_ParamSpec")
+"""
+A custom type to annotate the type hints of `add_parameter`.
+"""
+
+def add_parameter(method: Callable[[_ParamSpec], None]) -> Callable[[_ParamSpec], None]:
+    """
```
AdriaanRol, line 517: This docstring could benefit from a usage example. 

Most important is I think that the arguments of a method defined like this are passed to the "add_parameter" method, and that the

astafan8:
> Maybe the best example would be to rewrite the SGS100A microwave source driver using this new style. The reason is that this is the simplest driver, and it would serve as an example on how to write drivers "new-style". However, I see two issues with this. 1. it would delay/block the acceptance of this MR because the new driver needs to be tested, and 2. it enforces a standard before it is tested in working nice. So we should probably postpone the replacing to a later date.

agree with @AdriaanRol (especially happy for the point 2) - let's not add a driver of a real instrument in this PR, just stick to a dummy/test kind of driver.

caenrigen:
Thank you for the review @AdriaanRol 

Will have to process farther some parts of your feedback, here are a few replies.

> suggestion, the @add_parameter decorator in combination with the `_parameter_` prefix seems double to me. For a user (someone writing a driver) it would be much simpler to directly use the syntax `@add_parameter \n def time(.... `. For the behind the scenes implementation, this would mean that the `time` attribute would be removed and replaced the moment the add_parameter decorator is invoked.

definitely agree, it is a technical limitation in python, if a method `def time(...)` is added to the class it cannot be removed... Can't recall anymore if this directly clashes with `instrument.time()` or not, but the point is that it better stay away from such ambiguity (meaning, is `instrument.time` the old dummy method or the added parameter?). To solves this I made the parameter a private method. To farther avoid clashes with the user's private methods, I introduced the `_parameter_` prefix. Not ideal, so the `@add_parameter` decorator will educate driver writers about this mandatory prefix 😉.

> ### Feedback on how docs are rendered

> I have some questions as I am not as familiar with the sphinx magic you had to invoke to get this to work. At first sight, the parameter documentation seems a bit off and hard to read. This is because all is crammed into the call signature, and only the docstring shows.

Agreed, though will likely keep it out of scope for this Pull Request.

> Unless it is trivially easy, I suggest you do not implement these things, I'm mostly trying to assess if this would be possible in the future.

Very likely, sphinx magic and hacking can go pretty far, doing it the "intended" way might be harder.

> **question** would it be possible to have "parameter" appear as an italics prefix similar to the "class method" in the screenshot I showed.

Should be.

> **question** would it be possible to turn the arguments into a list with bold-faced names and then the text after it or is it better to do this using the docstring.

Should be possible to automate this, and I think should be automated, a heads up for making it compatible with both numpy and google style docstring.

**suggestion** if the second is preferred, let's include that in the example docstring (i.e., what the settable is, a link to the set_cmd and get_cmd, the unit, etc.

User will have this freedom anyway, but the idea would be to automate as much as possible.

---

### Note for self

Some Inspiration:
```python
from sphinx.ext import autodoc
autodoc.PropertyDocumenter??
```

```python
Init signature:
autodoc.PropertyDocumenter(
    directive: 'DocumenterBridge',
    name: str,
    indent: str = '',
) -> None
Source:        
class PropertyDocumenter(DocstringStripSignatureMixin, ClassLevelDocumenter):  # type: ignore
    """
    Specialized Documenter subclass for properties.
    """
    objtype = 'property'
    directivetype = 'method'
    member_order = 60

    # before AttributeDocumenter
    priority = AttributeDocumenter.priority + 1

    @classmethod
    def can_document_member(cls, member: Any, membername: str, isattr: bool, parent: Any
                            ) -> bool:
        return inspect.isproperty(member) and isinstance(parent, ClassDocumenter)

    def document_members(self, all_members: bool = False) -> None:
        pass

    def get_real_modname(self) -> str:
        return self.get_attr(self.parent or self.object, '__module__', None) \
            or self.modname

    def add_directive_header(self, sig: str) -> None:
        super().add_directive_header(sig)
        sourcename = self.get_sourcename()
        if inspect.isabstractmethod(self.object):
            self.add_line('   :abstractmethod:', sourcename)
        self.add_line('   :property:', sourcename)
File:           /usr/local/anaconda3/envs/q38/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py
Type:           type
Subclasses:   
```

Note the `self.add_line('   :property:', sourcename)`, will be necessary to define a custom `:parameter:` role in sphinx as well, likely with added benefit of being able to link to specific parameters throughout the docs.

A [custom documenter](https://stackoverflow.com/questions/39565203/how-to-specialize-documenters-in-sphinx-autodoc) as part of qcodes [sphinx extension](https://www.sphinx-doc.org/en/master/extdev/index.html) might be the proper way of doing this.

caenrigen:
Latest prototype, everything auto-generated:

Screenshot 2021-06-16 at 03 30 15

Sphinx hacks are more elegant than I was imagining.

<details><summary>`conf.py` source code</summary>

```python

from sphinx.ext import autodoc
from sphinx.domains import python as sdp

QCODES_ADD_PARAMETER_FLAG = "_add_parameter"
QCODES_METHOD_PREFIX = "_parameter_"


def format_name(self) -> str:
    if hasattr(self.object, QCODES_ADD_PARAMETER_FLAG):
        return ".".join(
            self.objpath[:-1] + [self.objpath[-1][len(QCODES_METHOD_PREFIX) :]]
        )
    return ".".join(self.objpath) or self.modname


original_add_directive_header = autodoc.MethodDocumenter.add_directive_header


def add_directive_header(self, sig: str) -> None:
    original_add_directive_header(self, sig)
    if self.object_name.startswith("_parameter_"):
        sourcename = self.get_sourcename()
        self.add_line("   :parameter:", sourcename)


original_get_signature_prefix = sdp.PyMethod.get_signature_prefix


def get_signature_prefix(self, sig: str) -> str:
    prefix_str = original_get_signature_prefix(self, sig)
    if "parameter" in self.options:
        prefix_str += "parameter "

    return prefix_str


def dont_skip_qcodes_params(app, what, name, obj, skip, options) -> bool:
    if hasattr(obj, QCODES_ADD_PARAMETER_FLAG):
        return False
    return skip


import inspect


def add_parameter_spec_to_docstring(app, what, name, obj, options, lines):
    # name is e.g. `"my_module._parameter_time"
    if QCODES_METHOD_PREFIX in name:
        lines += [
            "",
            ".. rubric:: Arguments passed to "
            ":meth:`~qcodes.instrument.base.InstrumentBase.add_parameter`:",
            "",
        ]
        for kw_name, par_obj in inspect.signature(obj).parameters.items():
            if kw_name != "self":
                if kw_name == "parameter_class":
                    # create link to the parameter class
                    mod = par_obj.default.__module__
                    class_name = par_obj.default.__name__
                    value = f":class:`~{mod}.{class_name}`"
                else:
                    value = f"*{par_obj.default!r}*"

                if kw_name == "vals":
                    kw_name = ":mod:`vals <qcodes.utils.validators>`"
                else:
                    kw_name = f"**{kw_name}**"
                lines.append(f"- {kw_name} = {value}")
        lines += [""]

    return lines


def clear_parameter_method_signature(
    app, what, name, obj, options, signature, return_annotation
):
    if QCODES_METHOD_PREFIX in name:
        signature = "()"
    return (signature, return_annotation)


def setup(app):
    # we can't have a method in the class with the same name as the desired
    # parameter, therefore we patch the method name displayed in the sphinx docs
    # monkey patching `MethodDocumenter`
    # e.g. `_parameter_time` -> `_parameter_time` when displayed in the docs
    autodoc.MethodDocumenter.format_name = format_name
    autodoc.MethodDocumenter.add_directive_header = add_directive_header
    sdp.PyMethod.option_spec.update(
        {"parameter": sdp.PyMethod.option_spec["classmethod"]}
    )
    sdp.PyMethod.get_signature_prefix = get_signature_prefix
    # enforce documenting the decorated private method
    app.connect("autodoc-skip-member", dont_skip_qcodes_params)
    app.connect("autodoc-process-docstring", add_parameter_spec_to_docstring)
    app.connect("autodoc-process-signature", clear_parameter_method_signature)

```
<details>

Good night! 😝

caenrigen:
@astafan8 I have improved the sphinx hacks by creating a sphinx extension and added documentation. Only missing tests.
Could you review these parts? I think the code and docs are in a finished or close to finished state.

qcodes/instrument/base.py
```diff
@@ -401,6 +424,244 @@ def validate_status(self, verbose: bool = False) -> None:
...
+            # the `@add_parameter` decorator adds an attribute we check for here
+            if hasattr(meth, _ADD_PARAMETER_ATTR_NAME):
+                param_name = meth.__name__[len(_DECORATED_METHOD_PREFIX):]
+                kwargs = {
+                    kw_name: kwarg_to_attr(self, kw_name, kw_value.default, param_name)
+                    for kw_name, kw_value in inspect.signature(meth).parameters.items()
+                }
+                self.add_parameter(
+                    name=param_name,
+                    docstring=meth.__doc__,
```
caenrigen, line 488: @peendebak the docstring from the method (decorated with @parameter) will be passed to the constructor of the Parameter classes on this line

let me know if this addresses your question

astafan8:
@caenrigen will review within the coming days! at a quick glance - looks great, especially the documentation depth and the sphinx extension part :)

caenrigen:
> @caenrigen will review within the coming days! at a quick glance - looks great, especially the documentation depth and the sphinx extension part :) in the meantime, there's still some work to do for typing https://github.com/QCoDeS/Qcodes/pull/3098/checks?check_run_id=2877069226

Thanks, @astafan8!

Also created this draft for a driver that uses inheritance and can already use the sphinx extension: https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/98

docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb
```diff
@@ -83,8 +112,34 @@
...
+   },
+   "source": [
+    "QCoDeS provides two ways of adding parameters to intruments:\n",
+    "\n",
+    "- (Conventional style) Via the :meth:`qcodes.instrument.base.Instrument.add_parameter` method.\n",
+    "- (Decorator style) Via the :func:`@qcodes.instrument.base.add_parameter <qcodes.instrument.base.add_parameter>` decorator function.\n",
+    "\n",
+    "The more recent decorator style provides two key advantages:\n",
+    "\n",
+    "- Allows to override parameters when inheriting them from another :class:`qcodes.instrument.base.Instrument` class.\n",
```
astafan8, line 131: strictly speaking the conventional style does allow one to override parameters, it's just that the to-be-overridden parameter needs to be removed first from ``self.parameters``, so i suggest to make it clear in this sentence, and also refer to the "Dynamically adding and removing parameters" that explain that:
```suggestion
    "- Allows to override parameters when inheriting them from another :class:`qcodes.instrument.base.Instrument` class without the need for interacting with the instrument's dictionary of parameters (``self.parameters``), see [Dynamically adding and removing parameters section](Dynamically-adding-and-removing-parameters).\n",
```

docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb
```diff
@@ -98,25 +153,187 @@
...
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "The parameters will be added to the instrument only during its\n",
+    ":code:`__init__`. This means that when using this style of adding parameters we\n",
```
astafan8, line 217:
```suggestion
    ":code:`__init__`. This means that when using this decorator style of adding parameters we\n",
```

docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb
```diff
@@ -98,25 +153,187 @@
...
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "In some cases you might need more control over when exactly the parameters\n",
+    "should be added to the instrument, e.g., when some information needed to create\n",
+    "the parameters is know only when an instance of the Instrument is created:\n",
```
astafan8, line 259:
```suggestion
    "the parameters is known only when an instance of the Instrument is created:\n",
```

docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb
```diff
@@ -98,25 +153,187 @@
...
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "Below you will find an example of the documentation produced by sphinx when making use\n",
+    "of the :mod:`qcodes.sphinx_extensions.add_parameter` and the new\n",
```
astafan8, line 306:
```suggestion
    "of the :mod:`qcodes.sphinx_extensions.add_parameter` and the\n",
```

docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb
```diff
@@ -98,25 +153,187 @@
...
+    ":func:`@add_parameter <qcodes.instrument.base.add_parameter>` decorator style of\n",
+    "adding parameters to an instrument.\n",
+    "\n",
+    "For usage details on building such documentation see :mod:`qcodes.sphinx_extensions.add_parameter`\n",
+    "documentation.\n",
+    "\n",
+    ".. _intrument_drivers_new_style_docs:\n",
+    "\n",
+    ".. NB the `autoclass` would output too many inherited methods so we do some manual tricks here.\n",
```
astafan8, line 315:
```suggestion
    ".. NB the `autoclass` would output too many inherited methods so we use `automethod` to explicitly specify which methods to generate documentation for in this tutorial.\n",
```

docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb
```diff
@@ -307,7 +524,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## DLL-based instruments\n",
+    "## DLL-based instruments (Conventional style)\n",
```
astafan8, line 527: i'd suggest to not mark these 3 sections (VisaInstrument and DLL based isnrtuments) as "conventional style" because these sections are not so much about ``self.add_parameter`` but more about how an instrument driver looks like in general
```suggestion
    "## DLL-based instruments\n",
```

docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb
```diff
@@ -83,8 +112,34 @@
...
     "Golden rule: if a `Parameter` is settable, it must always accept its own output as input.\n",
     "\n",
-    "In most cases you will probably be adding parameters via the `add_parameter` method on the instrument class as shown in the example below. \n",
+    "### Adding parameters to intruments"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
```
astafan8, line 121: a question about these raw cells - i find it not so cool that github's notebook rendering ignores them (see here https://github.com/QCoDeS/Qcodes/blob/ad2ea22cdeeea7be7432b2d8876e84e7f72abcb1/docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb). would you agree to split the content of these cells into parts that can be simple markdown and parts the require the rst syntax? :) this will have additional benefit (not so big but is nice and welcoming) that the non-rst content can be edited/improved by contributors who are less comfortable with rst but are comfortable with markdown. what do you think? :)

docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb
```diff
@@ -98,25 +153,187 @@
...
-    "Every QCoDeS module should have its own logger that is named with the name of the module. So to create a logger put a line at the top of the module like this:\n",
-    "```\n",
-    "log = logging.getLogger(__name__)\n",
-    "```\n",
-    "Use this logger only to log messages that are not originating from an `Instrument` instance. For messages from within an instrument instance use the `log` member of the `Instrument` class, e.g\n",
-    "```\n",
-    "self.log.info(f\"Could not connect at {address}\")\n",
-    "```\n",
-    "This way the instrument name will be prepended to the log message and the log messages can be filtered according to the instrument they originate from. See the example notebook of the logger module for more info ([offline](../logging/logging_example.ipynb),[online](https://nbviewer.jupyter.org/github/QCoDeS/Qcodes/tree/master/docs/examples/logging/logging_example.ipynb)).\n",
+    "## Simple driver examples (Decorator style)"
```
astafan8, line 156: i'd suggest to move the "decorator style" to below the "manual instruments" section. this would be inline with the statement that the "conventional style" is still the default and not going anywhere. also note that this example notebook can't really talk about "details of the ways of adding parameters" before a reader has seen an example of a driver in the "visa isntrument" section where many details of how it all works together are explained (mostly in the comments in the code). would you agree?

qcodes/instrument_drivers/new_style.py
```diff
@@ -0,0 +1,161 @@
...
+Author: Victor Negîrneac, vnegirneac@qblox.com
+"""
+# NB Usage examples are not included in the corresponding docstrings because their
+# code is used throughout the docs with literal include, so we keep them minimal.
+
+# Intentional module level import so that a literalinclude of the objects in this
+# file is as complete as possible
+import qcodes
+from qcodes import ManualParameter, Parameter
+from qcodes import validators
```
astafan8, line 16:
```suggestion
from qcodes.instrument.parameters import ManualParameter, Parameter
from qcodes.utils import validators
```

qcodes/instrument_drivers/new_style.py
```diff
@@ -0,0 +1,161 @@
+"""Driver examples using the new @add_parameter decorator style.
```
astafan8, line 1: i really like the "conventional style"/"method style" and "decorator style" terminology - shall it be used everywhere in the docs, docstring, and module names? the example notebook is already using it, so i'd suggest to rename this module to, say, ``decorator_style``, and also skim thorugh the docstrings/code to replace "new/old-style" with "method/decorator-style" terms

qcodes/instrument/base.py
```diff
@@ -15,8 +15,12 @@
     Type,
     Union,
     cast,
+    TypeVar
 )
```
astafan8, line 20:
```suggestion
```

qcodes/instrument/base.py
```diff
@@ -89,6 +102,16 @@ def add_parameter(self, name: str,
         """
         Bind one Parameter to this instrument.
 
+        .. note::
+
+            QCoDeS provides a new style of adding parameters using the
```
astafan8, line 107:
```suggestion
            QCoDeS provides also a decorator-style for adding parameters using the
```

qcodes/instrument/base.py
```diff
@@ -401,6 +424,259 @@ def validate_status(self, verbose: bool = False) -> None:
...
+        subclass.
+
+        .. seealso::
+
+            - :obj:`qcodes.instrument.base.add_parameter`
+            - :obj:`qcodes.instrument.base.InstanceAttr`
+
+        """
+
+        def kwarg_to_attr(self, kwarg_name, kwarg_value, param_name) -> Any:
```
astafan8, line 443:
```suggestion
        def kwarg_to_attr(self, kwarg_name: str, kwarg_value: Any, param_name: str) -> Any:
```

qcodes/instrument/base.py
```diff
@@ -401,6 +424,259 @@ def validate_status(self, verbose: bool = False) -> None:
...
+"""
+
+_ADD_PARAMETER_ATTR_NAME = "_add_parameter"
+"""
+A constant defining the name of the attribute set by the
+:func:`@add_parameter <qcodes.instrument.base.add_parameter>` decorator to flag a method
+that will be converted to parameter.
+"""
+
+_ParamSpec = TypeVar("_ParamSpec")
```
astafan8, line 538: paramspec has an important other meaning in qcodes (especially in qcodes.dataset.*), hence shall this be renamed to `_ParamArgs` or something similar?

qcodes/instrument/base.py
```diff
@@ -401,6 +424,259 @@ def validate_status(self, verbose: bool = False) -> None:
...
+"""
+A custom type to annotate the type hints of :obj:`qcodes.instrument.base.add_parameter`.
+"""
+
+def add_parameter(method: Callable[[_ParamSpec], Any]) -> Callable[[_ParamSpec], Any]:
+    r"""
+    A decorator function for adding parameters to instruments in the new style.
```
astafan8, line 545: let's avoid old/new style and be more specific
```suggestion
    A decorator function for adding parameters to instruments via method definitions.
```

qcodes/instrument/base.py
```diff
@@ -401,6 +424,259 @@ def validate_status(self, verbose: bool = False) -> None:
...
+
+    The new style has to main advantages (1) Allow to override inherited parameters.;
```
astafan8, line 547: ... let's note again about the ovverriding:
```suggestion
    The decorator style has two main advantages: (1) Allows one to override inherited parameters without interacting with instrument's parameters dictionary.;
```

qcodes/instrument/base.py
```diff
@@ -401,6 +424,259 @@ def validate_status(self, verbose: bool = False) -> None:
...
+    and (2) Allow to document parameters of instruments using the
+    :mod:`qcodes.sphinx_extensions.add_parameter` sphinx extension.
+
+    Intended to be used to decorate a method of an
+    :class:`~qcodes.instrument.base.Instrument` subclass. The information contained in
```
astafan8, line 552:
```suggestion
    :class:`~qcodes.instrument.base.InstrumentBase` subclass. The information contained in
```

qcodes/instrument/base.py
```diff
@@ -401,6 +424,259 @@ def validate_status(self, verbose: bool = False) -> None:
...
+    the definition of the decorated method will be processed by
+    :func:`!qcodes.instrument.base.Instrument._add_params_from_decorated_methods` and
```
astafan8, line 554: could you ensure that _add_params_from_decorated_methods is referenced via InstrumentBase where it is implemented, and not at Instrument?
```suggestion
    :func:`!qcodes.instrument.base.InstrumentBase._add_params_from_decorated_methods` and
```

qcodes/instrument/base.py
```diff
@@ -401,6 +424,259 @@ def validate_status(self, verbose: bool = False) -> None:
...
+
+            sub_instr.time.label:  Time long
+    """
+    if not method.__name__.startswith(_DECORATED_METHOD_PREFIX):
+        raise ValueError(
+            f"Only methods prefixed with {_DECORATED_METHOD_PREFIX!r} can be decorated "
+            f"with `add_parameter` decorator."
+        )
+
+    @wraps(method)  # preserves info like `__doc__` and signature
```
astafan8, line 661: excessive comment :)
```suggestion
    @wraps(method)
```

qcodes/instrument/base.py
```diff
@@ -74,6 +78,15 @@ def __init__(self, name: str, metadata: Optional[Mapping[Any, Any]] = None) -> N
 
         self.log = get_instrument_logger(self, __name__)
 
+        # NB this attribute should not be added here otherwise full control over when
+        # the parameters are instantiated is not possible in sub classes.
+        # self._call_add_params_from_decorated_methods = True
```
astafan8, line 83: i'm a bit confused - so in order to make the decorator actually work and add parameters from decorated methods, one does need to say ``self._call_add_params_from_decorated_methods = True`` before calling the super class' init method (or explicitly call ``_add_params_from_decorated_methods`` in the init), right? i don't see that happening in all the examples in the new_style.py file, and i also think that this is an important behavior that should be documented (maybe it is already, and i just missed it?)
```suggestion
```

qcodes/instrument/base.py
```diff
@@ -401,6 +424,259 @@ def validate_status(self, verbose: bool = False) -> None:
...
+    The new style has to main advantages (1) Allow to override inherited parameters.;
+    and (2) Allow to document parameters of instruments using the
+    :mod:`qcodes.sphinx_extensions.add_parameter` sphinx extension.
+
+    Intended to be used to decorate a method of an
+    :class:`~qcodes.instrument.base.Instrument` subclass. The information contained in
+    the definition of the decorated method will be processed by
+    :func:`!qcodes.instrument.base.Instrument._add_params_from_decorated_methods` and
+    passed to :meth:`qcodes.instrument.base.InstrumentBase.add_parameter` during the
+    instantiation of an instrument.
```
astafan8, line 556: this is the behavior documentation that i am missing (perhaps it's mentioned somewhere else already?)
```suggestion
    passed to :meth:`qcodes.instrument.base.InstrumentBase.add_parameter`.
    :func:`!qcodes.instrument.base.Instrument._add_params_from_decorated_methods` can be called
    directly in the instrument's ``__init__`` or, if ``_call_add_params_from_decorated_methods``
    attribute is set to ``True`` before the call to superclass' ``__init__``, then 
    :func:`!qcodes.instrument.base.Instrument._add_params_from_decorated_methods`
    will be called as part of the superclass' ``__init__``.
```

qcodes/instrument/base.py
```diff
@@ -401,6 +424,259 @@ def validate_status(self, verbose: bool = False) -> None:
...
+            sub_instr.time.label:  Time long
+    """
+    if not method.__name__.startswith(_DECORATED_METHOD_PREFIX):
+        raise ValueError(
+            f"Only methods prefixed with {_DECORATED_METHOD_PREFIX!r} can be decorated "
+            f"with `add_parameter` decorator."
+        )
+
+    @wraps(method)  # preserves info like `__doc__` and signature
+    def kwargs_and_doc_container(self, *args, **kwargs) -> None:
```
astafan8, line 662:
```suggestion
    def kwargs_and_doc_container(self, *args: Any, **kwargs: Any) -> None:
```

astafan8 (commented):
the implementation looks reasonable! most of the comments are about semantics, and clarity in the documentation :) for the rest, looks like the PR will very soon be ready for merge!

qcodes/sphinx_extensions/add_parameter.py
```diff
@@ -0,0 +1,268 @@
...
+    app.connect("autodoc-skip-member", dont_skip_qcodes_params)
+    # convert the signature of the decorated methods to a nicely formatted list in the
+    # docstring
+    app.connect("autodoc-process-docstring", add_parameter_spec_to_docstring)
+    app.connect("autodoc-process-signature", clear_parameter_method_signature)
+
+    return {
+        'version': '0.1',
+        'parallel_read_safe': True,  # Not tested, should not be an issue
+        'parallel_write_safe': True,  # Not tested, should not be an issue
```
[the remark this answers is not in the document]
caenrigen: I do not want to limit anyone that might be using the parallel sphinx build. Let's fix it if it becomes a problem. (I cannot test because on mac parallel build is not supported at all.)

docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb
caenrigen: Good catch, made the changes to avoid using rst as much as possible. I used it because it is so much easier to link to different parts of the docs

docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb
caenrigen: Agreed, "method style" and "decorator style"

qcodes/instrument_drivers/new_style.py
caenrigen: Agreed, "method style" and "decorator style"

qcodes/instrument/base.py
caenrigen: Default behavior: `InstrumentBase` calls `_add_params_from_decorated_methods` during `__init__` without the need for users to set `self._call_add_params_from_decorated_methods` nor explicitly call `_add_params_from_decorated_methods`.

Note the line:
```python
if getattr(self, "_call_add_params_from_decorated_methods", True):
    self._add_params_from_decorated_methods()
```

`InstrumentBase` has no `_call_add_params_from_decorated_methods` by default so that it is more flexible to control when exactly `_add_params_from_decorated_methods` is called (implicitly by parent classes and explicitly by subclasses). I know this explanation is not very specific and that is because I cannot recall the exact (advanced) edge-case it covers.

The behavior is now documented better in `@add_parameter` docstring.

astafan8: ah, of course, the default value of True - thank you pointing it out :)

caenrigen:
@astafan8 I think I have addressed all comment regarding the docs. Will continue later this week (hopefully).

caenrigen:
@astafan8 tests added!

It would be nice to have them for the sphinx extension as well, but I have no idea how that is done nor how difficult or time consuming it is to do it properly, sphinx in general is a bit of a magic black box.

---

Chores:

You need to configure codacy to [ignore either D212 or D213 they are mutually excluding](https://stackoverflow.com/questions/45990301/pep257-d212-and-d213-conflicts) otherwise the pipeline will never pass... Could you take care of these warnings according to the standard that you adopt?

---

Ready to merge? :)

qcodes/instrument/base.py
eendebakpt: @caenrigen Looks good

caenrigen:
Hey folks! Any progress on this PR? let me know if there is anything required from my side :)

jenshnielsen:
@caenrigen Sorry for waiting. Your work on this feature initiated a thorough alignment among maintainers, and now we are coming back with a suggestion. It started with the observation that while what this PR (the private method approach) enables is very neat, the code one the driver developer side is a bit too complex making the drivers somewhat harder to read. We were looking into options of how to preserve the simple/straightforward syntax of defining parameters and still have the benefit of documenting parameters. So, I tried coming up with a suggestion that makes driver implementation simpler but at the cost of moving the logic to enable the documentation into a sphinx extension that has to rely on parsing the code. I pushed the alternative suggestion into https://github.com/QCoDeS/Qcodes/pull/3188. What do you think of it? (we can discuss the tech details of the proposal in that PR)

caenrigen:
Hi! @jenshnielsen  Thank you for the updated, was reading the description of #3188 just before your comment.

> Your work on this feature initiated a thorough alignment among maintainers, and now we are coming back with a suggestion. 

I am glad that you have looked into this more deeply 🙂.

I am taking a few days off and will try to have a look into #3188 somewhere by the end of next week. For now I will leave a few comments.

> It started with the observation that while what this PR (the private method approach) enables is very neat, the code one the driver developer side is a bit too complex making the drivers somewhat harder to read.

I agree, it is a bit more complex, it abuses python in a few ways 😅. We found the benefits worth it since no alternative was available, and I put significant effort into documenting how it is intended to be used. Writing a qcodes driver requires reading docs in most cases anyway, I had to do it myself when writing https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/98.

I would suggest to have a look there to confirm that #3188 can provide the same functionalities, I used the `@add_paramer` style to implement drivers with inheritance.

In short, I am open to alternatives as long as their simplicity still allows for complex functionality. "Simple things should be simple, complex things should be possible."

> So, I tried coming up with a suggestion that makes driver implementation simpler but at the cost of moving the logic to enable the documentation into a sphinx extension that has to rely on parsing the code.

(Disclaimer: I have not look at the implementation) My intuition is that parsing code can be more fragile. Also, I had a hard time creating the sphinx extension, therefore I am not sure if moving a lot of complexity in there is a good idea, since it might be very hard to maintain.

I will try to gather some feedback within Quantify, and play with #3188 when I am back so that I can have an informed opinion on this topic.

One last question: Do you see both approaches possible to co-exist in qcodes? so that developers can choose themselves?

PS thank you for the #3188! From my own experience, I can imagine that takes some significant work.

jenshnielsen:
@caenrigen thanks for your comments

> One last question: Do you see both approaches possible to co-exist in qcodes? so that developers can choose themselves?

Technically I don't think there are any reasons why they cannot. I would say that if there are significant advantages from this approach that the other one does not give we can add this one too. Given the work on abstract parameters in #3125 I see that problem as solved without any more work but perhaps I miss something in this pr? That means that both these prs are largely concerned with improving the documentation. 

> I agree, it is a bit more complex, it abuses python in a few ways 😅. We found the benefits worth it since no alternative was available, and I put significant effort into documenting how it is intended to be used. Writing a qcodes driver requires reading docs in most cases anyway, I had to do it myself when writing QCoDeS/Qcodes_contrib_drivers#98.

This is largely a matter of preferences but I personally tend to prefer idiomatic python style not only for writing code but also for reading (both for humans and for tools). For example if you implement the `Weinschell_8320` as in #3188 the example below will enable a few nice features not possible with the current code or this pr (at least as far as I understand this work).

Screenshot 2021-07-09 170745

* Here vscode can autocomplete attenuation directly. This does not work with the add_paramter interface since the code has to be evaluated before the attribute is created and the ide has to be start enough to go via the delegate attr interface.
* VScode is also able to infer that attenuation is a parameter and therefore detect the methods on the parameter class and autocomplete these (even get and set that are dynamically generated )
* In the last line mypy can detect that `get` is misspelled `bet` since it knows that attenuation is a parameter that does not have an method called `bet`.

> (Disclaimer: I have not look at the implementation) My intuition is that parsing code can be more fragile. Also, I had a hard time creating the sphinx extension, therefore I am not sure if moving a lot of complexity in there is a good idea, since it might be very hard to maintain.

I do agree that is not ideal however I see the trade off a bit differently. To me it is more important that the driver code is as clean as possible and the more complex code is unused at runtime.
However the parser would indeed need significant work before that part of it can be merged. This is largely a proof of concept. 

> PS thank you for the #3188! From my own experience, I can imagine that takes some significant work.

I was able to take significant advantage from your work on how to setup a sphinx extension so thanks for that. It saved me a significant amount of time :)

caenrigen:
Hi guys!

Due to limitations in my bandwidth I won't be able to have a proper look into #3188 until likely somewhere in August.

> Technically I don't think there are any reasons why they cannot.

@jenshnielsen considering that #3188 might take a while to fully develop, and that this PR is ready and in principle not conflicting with #3188, any chance of including it in 0.27.0?

The #3188 is still interesting and I will be glad to help with that later :)

jenshnielsen:
Thanks again for your contribution. I know you have put a lot of time and effort into this thanks to your contribution, the issue was brought to our attention and we were able to think of ways to do it, but unfortunately we have come to the conclusion that this design is too complicated for us to be able to commit to maintaining it unless there is a significant further gain from this method. In addition we are a bit concerned that 3 methods for adding parameters to instruments is too much and will result in confusion among users. The part of #3188 that is concerned with changes to the actual code is very close to landing via #3191 and has basically only been held back due to me wanting to let it bake on master for awhile before releasing. It will therefor target 0.28. Whit that change the existence of parameters can be documented as long as they are documented as any other attribute.

With regards to the parsing of attributes to improve the documentation of them I feel like this is something that we could actually generalise outside QCoDeS. Rather than the code I originally wrote to remove self parameters and instantiate the parameters we can generate a proxy parameter with a repr that contains the full syntax. That will be a lot simpler and easier to reason about and not specific to QCoDeS parameters.

caenrigen:
@jenshnielsen thank you for the reply and consideration

>  but unfortunately we have come to the conclusion that this design is too complicated for us to be able to commit to maintaining it unless there is a significant further gain from this method

fair enough

> With regards to the parsing of attributes to improve the documentation of them I feel like this is something that we could actually generalise outside QCoDeS. Rather than the code I originally wrote to remove self parameters and instantiate the parameters we can generate a proxy parameter with a repr that contains the full syntax. That will be a lot simpler and easier to reason about and not specific to QCoDeS parameters.

Looking forward to this, sounds interesting, keep me updated on the developments of this, feel free to tag me in related PRs. (I also joined the QCoDeS Slack, we can chat there as well when needed.)

Cheers

PR #3105, opened 2021-06-09
Compare shapes in RunDescriber equality method

astafan8:
@jenshnielsen what do you think, should this be happening or not?

for latest version of the class, this seems very fair, but i'm not sure what should happen for the case where RunDescriber from V2 data (without shapes, because there were no notion of shapes back then) is compared with RunDescriber from V3 data (that has the shapes or has None as the shape) and the interdependencies part is equal. or am i overthinking this too much?

### DIFF
--- a/qcodes/dataset/descriptions/rundescriber.py
+++ b/qcodes/dataset/descriptions/rundescriber.py
@@ -125,6 +125,8 @@ class RunDescriber:
             return False
         if self.interdeps != other.interdeps:
             return False
+        if self.shapes != other.shapes:
+            return False
         return True
 
     def __repr__(self) -> str:


jenshnielsen:
One option could be to only to perform this comparison if both are not None? I am not sure how likely it is that someone has a v2 rundescriber that is upgraded and try to compare it with a directly constructed v3 describer

jenshnielsen:
I think #3106 should fix the docs build here

astafan8:
> One option could be to only to perform this comparison if both are not None?

yeah, but that will also rouing the comparison of directly constructed

> I am not sure how likely it is that someone has a v2 rundescriber that is upgraded and try to compare it with a directly constructed v3 describer

indeed. but if this is not likely, then the way the comparison is implemented in this PR is good, right?

jenshnielsen (approved):

PR #3116, opened 2021-06-11
Improve xarray metadata export

jenshnielsen:
- [X] Add units and long_name to match default xarray usage https://xarray.pydata.org/en/stable/getting-started-guide/quick-overview.html#attributes
- [x] Ensure that the per array metadata is also exported to dataarrays (and not just datasets)
- [x] Export parent dataset links
- [x] Tests for all above
- [x] Refactor logic to make the handling of data_vars attrs more clear

### DIFF
--- a/qcodes/dataset/exporters/export_to_xarray.py
+++ b/qcodes/dataset/exporters/export_to_xarray.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import warnings
-from typing import TYPE_CHECKING, Dict, Hashable, Mapping, Union, cast
+from typing import TYPE_CHECKING, Any, Dict, Hashable, Mapping, Union, cast
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


qcodes/dataset/exporters/export_to_xarray.py
```diff
@@ -101,11 +102,26 @@ def load_to_xarray_dataset(dataset: DataSetProtocol, data: ParameterData) -> xr.
...
+    _add_param_spec_to_xarray_coords(dataset, xrdataset)
     _add_metadata_to_xarray(dataset, xrdataset)
 
     return xrdataset
+
+
+def _add_param_spec_to_xarray_coords(
+    dataset: DataSetProtocol, xrdataset: Union[xr.Dataset, xr.DataArray]
+) -> None:
+    for coord in xrdataset.coords:
```
FarBo, line 114: It only cares about coordinates. so data_vars left to be updated here.
Shouldn't we have a similar loop to update those:

```
for var in xrdataset.data_vars:
    ....
```

jenshnielsen: This currently happens inside _load_to_xarray_dataarray_dict_no_metadata thou a refactor of that would perhaps be nice

jenshnielsen:
@FarBo and @astafan8 ready for review

astafan8 (approved):

PR #3118, opened 2021-06-11
Alazar ATS drivers: Fix buffer size for mode='TS' and buffers_per_acquisition > 1

### LINKED ISSUE #3117: AlazarTech ATS mode=TriggeredStreaming fails with > 1 buffer
### Steps to reproduce
1. call ats.aquire(mode='TS', samples_per_record=100e6, buffers_per_acquisition=10)
2. qcodes driver tries to allocate buffer of 1000e6 bytes, which is larger than ATS driver supports.

2nd:
1. call ats.aquire(mode='TS', samples_per_record=1e6, buffers_per_acquisition=10)
2. instead of 10 time handle_buffer with a buffer of 1e6 bytes the qcodes driver returns a buffer with 10e6 bytes.

See fix in pull request.

**qcodes branch**
Master

**qcodes commit**

sldesnoo-Delft:
Fixes #3117

- in mode 'TS' the buffer size = reccord size / nr of buffers

### DIFF
--- a/qcodes/instrument_drivers/AlazarTech/ATS.py
+++ b/qcodes/instrument_drivers/AlazarTech/ATS.py
@@ -386,6 +386,10 @@ class AlazarTech_ATS(Instrument):
         internal_buffer_size_requested = (bits_per_sample * samples_per_record *
                                           records_per_buffer) // 8
 
+        if mode == 'TS':
+            transfer_buffer_size //= buffers_per_acquisition
+            internal_buffer_size_requested //= buffers_per_acquisition
+
         if internal_buffer_size_requested > max_buffer_size:
             raise RuntimeError(f"Requested a buffer of size: "
                                f"{internal_buffer_size_requested / 1024 ** 2}"


astafan8 (commented):
@sldesnoo-Delft thank you very much!

is this a qcodes driver bug or is this something that changed since some Alazar firmware version (or something similar)?

i'm generally surprised that you're the first user to report this bug for the TS mode and ``buffers_per_acquisition != 1`` case :) and i wonder if downstream folks just never worked with this combination of arguments, or have workarounds implemented in their code... (the latter would be aweful)

have you tested the driver with this change for both TS and NPT modes and ``buffers_per_acquisition != 1`` and ``buffers_per_acquisition == 1`` cases just to be sure that the rest of the rest of the driver is working?

sldesnoo-Delft:
@astafan8 I've tested with NPT en TS modes with n_buffers == 1 and > 1.
I don't know whether AlazarTech changed anything in the ATS driver. I've tested with version 7.4.6 from April 2020. I expect the 64 MB limit for the buffer size is related to hardware and not driver related.
So, maybe I'm the first trying to collect more than 64 MB of data.

astafan8 (approved):
thank you  for the thoroguh testing! then let's get this great fix in :)

PR #3119, opened 2021-06-11
Add instrument to delegateparameter

ThorvaldLarsen:
Adding instrument property from source to delegate parameter. This is required to utilize threaded measurements in qcodes doNd function on delegate measurement parameters. With the current implementation the all delegate parameters has `root_instrument=None` which means the part defining number of threads to used based on number of instruments to be pinged comes out to 1 thread.

The issue for number of parallel threads for delegate parameters can alternatively be handled in `_call_threaded_measurements` by specifically accounting for delegate parameter types.

Not sure if we want to add tests to this part if we decide to go with this solution.

@jenshnielsen @astafan8

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -1499,6 +1499,7 @@ class DelegateParameter(Parameter):
 
         initial_cache_value = kwargs.pop("initial_cache_value", None)
         self.source = source
+        self._instrument = source.instrument
         super().__init__(name, *args, **kwargs)
         # explicitly set the source properties as
         # init will overwrite the ones set when assigning source
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -1499,11 +1499,12 @@ class DelegateParameter(Parameter):
 
         initial_cache_value = kwargs.pop("initial_cache_value", None)
         self.source = source
-        self._instrument = source.instrument
         super().__init__(name, *args, **kwargs)
         # explicitly set the source properties as
         # init will overwrite the ones set when assigning source
         self._set_properties_from_source(source)
+        if source is not None:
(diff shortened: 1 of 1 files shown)


astafan8:
@ThorvaldLarsen giving the instrument of the source parameter to the delegate can also be done by passing it as a kwarg (like for any parameter):
```python
p = Parameter(...)
d = DelegateParameter("d", p, instrument=p.instrument, ...)
```
would this suffice your use case? dond functions will pick up the root instrument correctly and do the threading as expected.

-----

I have to mention that this is not the first case where a user wants to use the threading with dond's but parameters are defined/implemented in such a way that the root instrument is not indicating which instrument(s) the parameter actually uses behind the scenes. For one of those cases, a local interface was implemented on the parameter that exposes an "underlying_instrument" property that may return an instrument that is different from the root_instrument, and then the threading subfunction inside dond was reimplemented to make use of this "underlying_instrument" property. We could consider adding this kind of property to the parameter base class, it's default implmenetation would be to return root_instrument, but users can overwrite this (either on instances, or in the subclass implementation) to return an instrument other than the root_instrument.

ThorvaldLarsen:
@astafan8 it is quite easy to work around with the way we use delegate parameters (currently just adding the instrument after instantiating the parameter) but I was thinking that the delegate parameter inheriting the instrument from the source makes sense on a general level - hence this PR.

I dont really have an opinion on a more generalized version for the threaded measurements except that it might be valuable to make the `process_param_meas` function (or the threading subfunction) be part of the measurement module rather than living inside doNd.

ThorvaldLarsen:
So after working with this for a bit I realize that adding the instrument also changes the name on the data in the datasets. So this PR might actually break quite a lot for people that might have data saved through delegate parameters. I think another solution would be better.

Issue #3109, opened 2021-06-10
Git folder weighting more than 99.3% of the repo size


edumur:
Hi all,

I recently wondered why the repo size was so big.
First, let me say that the repo size is 3.6GB which is huge for waht we have in it (mainly text files).

I ran WinDirStat to get some statistics on the repo and I found this:

![Screenshot 2021-06-10 141349](https://user-images.githubusercontent.com/13218882/121523086-1cf6b480-c9f6-11eb-9282-c09ff376affd.png)

Is there any reason to keep the .git folder so big?
After running the following command:
```bash
git repack -a -d --depth=250 --window=250
git prune
git gc --aggressive
```
I got the following stat:
![image](https://user-images.githubusercontent.com/13218882/121540207-94ccdb00-ca06-11eb-9a02-a795ac21fd32.png)

It saved 0.8GB which is better but not perfect...

Well at the end, I guess I just wanted to raise the question of the size of the repo which I find quite big ^^ , and maybe come up with a solution.

### System
**operating system** Windows 10

jenshnielsen:
@edumur it is indeed unfortunate that the repo is as big as it is. I did some stats at some point and it is mainly due to ipy notebooks and the images therein. However, the .git folder contains all the revisions of the code that has ever been created so it is expected to be big compared to the checkout code. It is not really possible to shrink the repository without rewriting the history and chancing every single git commit since the start. In the future we will be more strict about the number of notebooks and their revisions to try to stop the growth

edumur:
Thanks for this reply.
Do you think it is possible to have some files, maybe notebook, which do not exist today that we can erase from the git history to save some space?

jenshnielsen:
Not really, you cannot remove a file from the git history unless you are willing to rewrite the entire history changing every single commit hash and that is a very destructive thing to do

edumur:
I see...
Well, as you said, this is unfortunate ^^" .
You can close the issue or leave it opened, as you wish.

Thanks again for you reply and the work you do on this lib ;) .

jenshnielsen:
After some investigation it turns out that this was largely due to the git history being retained in the gh-pages branch that holds the docs. After deleting that a fresh clone of qcodes is ~140 MB for me.

To get the saving locally you will probably have to:
* Fetch the origin.
* If your own fork or any other fork that you have added as a remove contains a gh-pages branch you may also have to delete the branch from that fork
* Run the above commands to perform gc etc

[...]

PR #3122, opened 2021-06-14
Do not preserve gh pages branch history

jenshnielsen:
This will hopefully go some of the way in fixing https://github.com/QCoDeS/Qcodes/issues/3109 since the
gh-pages branch is fairly heavy and grows with every commit


jenshnielsen:
Note that this does mean that we loose all history in the gh-pages branch but I think that is acceptable

astafan8 (approved):

PR #3125, opened 2021-06-15
Abstract parameter 

sohailc:
Abstract parameters allow us to create abstract instrument types which are guaranteed to have certain parameters present. For instance, this will allow us to create a unified interface for all voltage sources.

Usually, when a parameter is added to an instrument and that instrument is sub classed and a parameter of the same name is added in the subclass as well, this will raise an exception. This issue is circumvented by specifying in the base class that the parameter being added there is of type AbstractParameter. The base class needs to be decorated with `abstract_instrument`.  

A catch is that, if a unit is specified in the base class and a unit is specified in the sub class, these units must match.

@jenshnielsen, @astafan8


qcodes/instrument/abstract.py
```diff
@@ -0,0 +1,101 @@
...
+
+            param = parameter_class(name=name, instrument=self, **kwargs)
+            self.parameters[name] = param
+
+        else:
+            # If it is a parameter other then abstract parameter, call the original
+            # method
+            original_add_parameter(
+                self, name, parameter_class, **kwargs
+            )
```
astafan8, line 96: shall we call the original method anyway, also for the abstract parameter case?
```suggestion
            del self.parameters[name]

        original_add_parameter(
            self, name, parameter_class, **kwargs
        )
```

qcodes/instrument/abstract.py
```diff
@@ -0,0 +1,101 @@
...
+            existing_unit = getattr(existing_parameter, "unit", None)
+            new_unit = kwargs.get("unit", None)
+
+            if existing_unit and existing_unit != new_unit:
+                raise AbstractParameterException(
+                    f"The unit of the parameter '{name}' is '{new_unit}', "
+                    f"which is inconsistent with the unit '{existing_unit}' "
+                    f"specified earlier. This is usually because a driver "
+                    f"is a subclass of a baseclass which defines a parameter "
+                    f"of the same name but with different units"
```
astafan8, line 85: would this be correct? and a more specific error message? :)
```suggestion
                    f"The unit of the parameter '{name}' is '{new_unit}', "
                    f"which is inconsistent with the unit '{existing_unit}' "
                    f"specified in the baseclass {cls.__name__!r} "
                    f"which defines a parameter "
                    f"of the same name but with different units"
```

qcodes/instrument/abstract.py
```diff
@@ -0,0 +1,101 @@
...
+                    f"Class '{cls_name}' has un-implemented Abstract Parameter(s): " +
+                    ", ".join([f"'{name}'" for name in abstract_parameters])
+                )
+
+        sub_cls.__init__ = __init_new__
+
+    original_add_parameter = cls.add_parameter
+
+    def add_parameter(
+            self, name: str, parameter_class: type = Parameter,
```
astafan8, line 68: to indicate that parameter_class is expected to be a subclass of _BaseParameter (as in the original method (i hope) )
```suggestion
            self, name: str, parameter_class: Type[_BaseParameter] = Parameter,
```

qcodes/instrument/abstract.py
```diff
@@ -0,0 +1,101 @@
...
+                raise AbstractParameterException(
+                    f"Class '{cls_name}' has un-implemented Abstract Parameter(s): " +
+                    ", ".join([f"'{name}'" for name in abstract_parameters])
+                )
+
+        sub_cls.__init__ = __init_new__
+
+    original_add_parameter = cls.add_parameter
+
+    def add_parameter(
```
astafan8, line 67: shall we use wraps here to preserve all the docstrings and other aspects of the original method?
```suggestion
    @wraps(original_add_parameter)
    def add_parameter(
```

qcodes/instrument/abstract.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class AbstractParameterException(Exception):
+    """
+    Errors with abstract parameter utilization should
+    raise this exception class.
+    """
+
+
+def abstract_instrument(cls: Type) -> Type:
```
astafan8, line 21:
```suggestion
def abstract_instrument(cls: Type[InstrumentBase]) -> Type[InstrumentBase]:
```

astafan8, line 21: to make it even better, this could be defined as a type variable to underline that the type of the given class will not change (in other words, that the same class will be returned; or, in yet other words, this decorator does not create a completely new class). don't remember the exact way to do it but along the lines of:
```suggestion
InstrumentBaseT = TypeVar("InstrumentBaseT", bound=Type[InstrumentBase])

def abstract_instrument(cls: InstrumentBaseT) -> InstrumentBaseT:
```

astafan8 (commented):
this looks neat and resembles how ``abc.ABC`` and ``abstractmethod`` work together in python! :) and it seems to me that this will still be supported by the new "static method of defining parameters" that folks are cooking in this PR https://github.com/QCoDeS/Qcodes/pull/3098

qcodes/instrument/abstract.py
sohailc: Yeah great point!

qcodes/instrument/abstract.py
sohailc: We can, but not as-is. The original 'add_parameter' will complain that there already is a parameter with the name you are adding. The only way this will work is if in the decorated case we remove the abstract parameter first. It will work either way so I am fine with this solution.

qcodes/instrument/abstract.py
sohailc: Yes, great suggestion, thank you!

qcodes/instrument/abstract.py
astafan8: > The only way this will work is if in the decorated case we remove the abstract parameter first

that's what i suggest in the green snippet indeed :)

qcodes/instrument/abstract.py
sohailc: Is the leading underscore a typo?

qcodes/instrument/abstract.py
jenshnielsen: Before we go down this route I would like to understand what the benefits are of implementing abstract instrument as a class decorator. I would initially think that it is cleaner to implement this logic inside add_parameter in such a way that it directly allows overwriting the parameter iff the overwritten parameter is abstract and the unit match. And we would implement a post init in the base class that performs the check for any remaining non overwritten abstract parameters

qcodes/instrument/abstract.py
```diff
@@ -0,0 +1,96 @@
...
+from qcodes.instrument.base import InstrumentBase
+from qcodes.instrument.parameter import _BaseParameter
+
+
+class AbstractParameter(Parameter):
+    """
+    This is a trivial subclass of 'Parameter' to signal
+    that this parameters *must* be overridden in
+    subclasses of abstract instruments.
+    """
```
jenshnielsen, line 12: I don't think it is a good idea to make express the abstract. How would you mix this with an ArrayParameter, MultiParameter or any other class of parameter. It would be better to let abstract be a property on the BaseParameter class which can be set by passing an argument to add parameter. Furthermore this is consistent with how it is implemented on an Abstract Method

qcodes/instrument/abstract.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def abstract_instrument(klass: Type[InstrumentBase]) -> Type[InstrumentBase]:
+    """
+    A class decorator to create an abstract instrument. Abstract
+    instruments are allowed to have abstract parameters, but
+    their subclasses do not. This works by replacing the
+    'add_parameter' and '__init_subclass__' methods of the class
+    dynamically.
+
```
jenshnielsen, line 29: I do not see the benefit of the existence of abstract instrument. I would simply put all the logic inside the InstrumentBase class  such that all instruments can have abstract parameters. The requirement to use ABC or ABC metaclass for abstract methods is imho only due to historical reasons  so lets not repeat that

qcodes/instrument/abstract.py
```diff
@@ -0,0 +1,96 @@
...
+                    raise AbstractParameterException(
+                        f"The unit of the parameter '{name}' is '{new_unit}', "
+                        f"which is inconsistent with the unit '{existing_unit}' "
+                        f"specified in the baseclass {klass.__name__!r} "
+                    )
+
+                # Remove the original abstract parameter to make room for the
+                # implementation in the subclass.
+                del self.parameters[name]
+
```
jenshnielsen, line 91: This method does not integrate very nicely with the add parameter in the base class. Deleting the parameter here seems ugly. Why not put this code inside the standard add_parameter logic such that this method can be removed completely

sohailc:
@jenshnielsen & @astafan8 Can you have another look at this, please?

qcodes/instrument/base.py
```diff
@@ -106,11 +106,21 @@ def add_parameter(self, name: str,
...
-            raise KeyError(f'Duplicate parameter name {name}')
+            ValueError: If this instrument already has a parameter with this
+                name but its unit are not equal to the unit of the parameter
+                being added.
+        """
+        existing_parameter = self.parameters.get(name, None)
+        if existing_parameter:
+            existing_unit = getattr(existing_parameter, "unit", None)
+            new_unit = kwargs.get("unit", None)
+            if existing_unit != new_unit:
```
jenshnielsen, line 117: I guess we should also raise here if the existing_parameter is not abstract ? Or do we want to allow non abstract parameters to be overwritten?

sohailc: I would vote for raising an exception if the parameter we wish to override is not abstract.

qcodes/instrument/base.py
```diff
@@ -74,6 +76,45 @@ def __init__(self, name: str, metadata: Optional[Mapping[Any, Any]] = None) -> N
 
         self.log = get_instrument_logger(self, __name__)
 
+        if self._call_post_init:
+            self.__post_init__(name, metadata)
+
+    def __init_subclass__(cls, **kwargs: Any) -> None:
```
jenshnielsen, line 82: I think a comment / a docstring explaining how the init subclass logic works would be nice

qcodes/instrument/base.py
```diff
@@ -74,6 +76,45 @@ def __init__(self, name: str, metadata: Optional[Mapping[Any, Any]] = None) -> N
...
+            call_post_init = self._call_post_init
+            # but prevent the init methods of
+            # base classes from also calling post init
+            self._call_post_init = False
+            original_init(self, *args, **sub_class_kwargs)
+
+            if call_post_init:
+                self.__post_init__(*args, **sub_class_kwargs)
+
+        cls.__init__ = __new_init__  # type: ignore
```
jenshnielsen, line 101: Please use ```mypy --show-error-code``` to figure out which error code this is so that we can only skip one error code. 

`# type: ignore[nameoferrorcode] `

qcodes/instrument/base.py
```diff
@@ -106,11 +147,26 @@ def add_parameter(self, name: str,
             **kwargs: Constructor arguments for ``parameter_class``.
 
         Raises:
-            KeyError: If this instrument already has a parameter with this
-                name.
+            ValueError: If this instrument already has a parameter with this
```
jenshnielsen, line 150: This still raises an KeyError if the parameter is not abstract. Changing the error type is an api break so I think we should stick to the same error.

qcodes/instrument/parameter.py
```diff
@@ -261,6 +261,9 @@ class _BaseParameter(Metadatable):
 
         metadata: extra information to include with the
             JSON snapshot of the parameter
+
+        abstract: Specifies if this parameter is abstract or not. Default
+            is False
```
jenshnielsen, line 266: Should we document that the parameter can be overwritten ?

jenshnielsen (approved):

qcodes/instrument/base.py
sohailc: Actually, the docstring is wrong. I am raising the same error when the parameter already exists and its not an abstract parameter: A `KeyError`. I will change the docstring to reflect this.

qcodes/instrument/base.py
sohailc: I have added docstrings here

qcodes/instrument/base.py
sohailc: I have done so now

PR #3126, opened 2021-06-15
Export complex numbers to netcdf

jenshnielsen:
- [X] Basic test that complex numbers can be exported
- [x] Fix the issue see http://xarray.pydata.org/en/stable/howdoi.html

### DIFF
--- a/.github/static/environment_forge_full.yml
+++ b/.github/static/environment_forge_full.yml
@@ -6,15 +6,15 @@ dependencies:
  # recommended to be explicitly specified by conda-incubator/setup-miniconda@v2
  - pip
  # dependencies for setup.cfg
- - numpy>=1.15
+ - numpy>=1.17
  - pyvisa>=1.11.0, <1.12.0
  - h5py>=2.8.0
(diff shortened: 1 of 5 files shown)


jenshnielsen:
One question remains. Should we completely drop the netcdf4 package and go with h5netcdf for all the export? 

You can read about h5netcdf here https://pypi.org/project/h5netcdf/ and why the author anyway thinks its better than the netcdf4 python package

astafan8 (approved):

astafan8:
> Should we completely drop the netcdf4 package

i think we should. i really don't like myself this situation with hdf5 being behind a netcdf interface that does not have full feature parity thus producing invalid netcdf (but valid hdf5), but - if there's one storage format/engine, then it's easier to maintain, right? and we must let users export data with complex numbers, no way around that. so i'm happy to use h5netcdf only.

astafan8 (approved):

jenshnielsen:
This ended up requiring to bump the minimum versions of xarray and in turn numpy and pandas

PR #3127, opened 2021-06-15
Add underlying_instrument property to parameter, use it in call_params_threaded instead of root_instrument, document call_params_threaded

Akshita07:
- Add gettable underlying_instrument property to _BaseParameter
-  Override underlying_instrument property in DelegatePArameter
- Move threading related functions to threading.py
- Make _call_params_threaded a public function
- Use underlying_instrument for threading

Todo:
[x] Add tests for new abstract property
[x] Add all the details to notebook

@astafan8 @jenshnielsen


qcodes/utils/threading.py
```diff
@@ -90,3 +101,87 @@ def thread_map(
...
+    for param in real_parameters:
+        if param.underlying_instrument:
+            output[param.underlying_instrument.full_name] += (param,)
+        else:
+            output[None] += (param,)
+
+    return output
+
+
+def call_params_threaded(param_meas: Sequence[ParamMeasT]) -> OutType:
```
astafan8, line 140: let's adda docstring to this now-public function? :)

Akshita07: added :)

FarBo:
@Akshita07 
It looks good to me. 
How about having a test for process_params_meas?

docs/examples/DataSet/Data acquisition with help of threads .ipynb
```diff
@@ -0,0 +1,602 @@
...
+   "id": "dietary-coordinate",
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2021-06-24T15:43:06.991369Z",
+     "start_time": "2021-06-24T15:43:06.971369Z"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "class SleepyDmmExponentialParameter(Parameter):\n",
```
jenshnielsen, line 108: Perhaps we should explain that we on purpose make this instrument slow to simulate slow communication with the instrument

docs/examples/DataSet/Data acquisition with help of threads .ipynb
```diff
@@ -0,0 +1,602 @@
...
+   "source": [
+    "# Data acquisition with help of threads"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "revised-pontiac",
+   "metadata": {},
+   "source": [
+    "In this notebook, we will explore how threading can be used with measurement context manager or dond functions for faster data acquisition. It is important to note that, QCoDeS provided threading happens per instrument. Meaning, per instrument one thread is created and all parameters from same instrument gets assigned to the same thread for data acquizition. \n",
```
jenshnielsen, line 16:
```suggestion
    "In this notebook, we will explore how threading can be used with measurement context manager or dond functions for faster data acquisition. It is important to note that, the threading QCoDeS provideds happens per instrument. Meaning, per instrument one thread is created and all parameters from same instrument gets assigned to the same thread for data acquizition. It is generally not safe for more than one thread to communicate with the same instrument at the same time.\n",
```

docs/examples/DataSet/Data acquisition with help of threads .ipynb
```diff
@@ -0,0 +1,602 @@
+{
+ "cells": [
```
jenshnielsen, line 2: Remove trailing spaces in file name. I would also call it either `Data acquisition with the help of threads.ipynb` or perhaps easier to read simply  `Threaded Data acquisition.ipynb`

docs/examples/Parameters/Parameters.ipynb
```diff
@@ -141,6 +134,159 @@
...
+   ],
+   "source": [
+    "vp2.underlying_instrument"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now, `vp2` parameter knows that its underlying instrument is `dac`. This is quite useful while measuring these virtual parameters with `doNd` functions with `use_threads` flag set to `True`.   "
```
jenshnielsen, line 287: Should we explain here why that is ?

docs/examples/DataSet/Data acquisition with help of threads .ipynb
Akshita07: Ah shoot! missed that space! Will update the title :)

docs/examples/Parameters/Parameters.ipynb
Akshita07: Explain why param knows about underlying instrument or why is it useful?

Akshita07: Got it, explaining why it is useful.

qcodes/tests/parameter/conftest.py
```diff
@@ -144,6 +145,19 @@ def get_func():
         return get_func
 
 
+class VirtualParameter(Parameter):
+    def __init__(self, name: str, param: Parameter, **kwargs):
+        self._param = param
+        super().__init__(name=name, **kwargs)
+
+    @property
+    def underlying_instrument(self) -> Optional['InstrumentBase']:
```
FarBo, line 154: Why is InstrumentBase a string?

Akshita07: updated

qcodes/instrument/parameter.py
```diff
@@ -969,6 +969,18 @@ def settable(self) -> bool:
...
 
+    @property
+    def underlying_instrument(self) -> Optional['InstrumentBase']:
+        """
+        Property to expose virtual parameters' underlying instrument i.e.
+        to allow the user to set the actual instrument of a parameter to be
+        the underlying instrument for a virtual parameter. A virtual
+        parameter is one which is built over another existing parameter.
+
+        By default it return the root_instrument of the parameter.
```
astafan8, line 980: @Akshita07 i was trying to improve this docstring for clarity, and this is what i came up with, what do you think? is this better?
```suggestion
        Returns an instance of the underlying hardware instrument that this
        parameter communicates with, per this parameter's implementation.

        This is useful in the case where a parameter does not belongs to
        an instrument instance that represents a real hardware instrument
        but actually uses a real hardware instrument in its implementation
        (e.g. via calls to one or more parameters of that real hardware
        instrument). This is also useful when a parameter does belong to
        an instrument instance but that instance does not represent the
        real hardware instrument that the parameter interacts with: hence
        ``root_instrument`` of the parameter cannot be the
        ``hardware_instrument``, however ``underlying_instrument`` can be
        implemented to return the ``hardware_instrument``.

        By default it returns the ``root_instrument`` of the parameter.
```

qcodes/tests/test_threading.py
```diff
@@ -0,0 +1,76 @@
...
+        thread_id2 = params_output[i][1]
+
+        if param1.instrument is param2.instrument:
+            assert thread_id1 == thread_id2
+
+        if param1.instrument is not param2.instrument:
+            assert thread_id1 != thread_id2
+
+        param1 = param2
+        thread_id1 = thread_id2
```
astafan8, line 76: @Akshita07 i am trying to see if this assertion algorithm always works, and i'm having trouble with processing it. it might be just me, but for the sake of providing an alternative i ended up with the following implementation - do you think it's clearer and more strict and explicit? 
```suggestion
    params_per_thread_id = defaultdict(set)
    for param, thread_id in params_output:
        params_per_thread_id[thread_id].add(param)
    assert len(params_per_thread_id) == 2
    expected_params_per_thread = {
        {dummy_1.voltage_1, dummy_1.voltage_2},
        {dummy_2.voltage_1, dummy_2.voltage_2}
    }
    assert set(params_per_thread_id.values()) == expected_params_per_thread
```

qcodes/instrument/parameter.py
Akshita07: @astafan8 this reads much better :). I am comiting it and resolving this comment.

qcodes/tests/test_threading.py
Akshita07: @astafan8 it's an interesting use of defaultdict. I'll note it down for future implementation purposes. And yes, your implementation is much more clearer, strict and explicit. I am comiting this as well :)

astafan8 (approved):

PR #3135, opened 2021-06-18
add export info metadata to dataset

jenshnielsen:
This will eventually make it easier to export and reimport datasets

- [x] Test for meta data being set, written and loaded correctly. 
- [x] A schema (dataclass -> str repr for this data)
- [x] Let export write to this structure

### DIFF
--- a/qcodes/dataset/exporters/export_info.py
+++ b/qcodes/dataset/exporters/export_info.py
@@ -27,5 +27,8 @@ class ExportInfo:
 
     @classmethod
     def from_str(cls, string: str) -> "ExportInfo":
+        if string == "":
+            return cls({})
+
         datadict = json.loads(string)
         return cls(**datadict)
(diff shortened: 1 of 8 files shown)


qcodes/dataset/exporters/export_info.py
```diff
@@ -0,0 +1,34 @@
...
+@dataclass
+class ExportInfo:
+
+    export_paths: Dict[str, str]
+
+    def __post_init__(self) -> None:
+        allowed_keys = tuple(a.value for a in DataExportType)
+        for key in self.export_paths.keys():
+            if key not in allowed_keys:
+                raise TypeError(
```
astafan8, line 20: instead of raising, shall we warn? or warn and skip the unrecognized keys? in case qcodes version A supports export to X, and version qcodes B supports export X and Y, not-raising-an-exception would allow one to open a dataset in qcodes A using method X that was exported with method X and method Y using qcodes B -- because qcodes A would see the method Y that is not known to it but the code will still work because it will actually use the method X that it supports.

jenshnielsen: Hmn that is worth considering. Originally I wanted the enum to be part of the type key but that is hard to json serialize

astafan8 (approved):

PR #3143, opened 2021-06-22
Add customDimensionsFilter to inject custom dim, and inject pythonExecutable

WilliamHPNielsen:
This PR introduces one more piece of information to be collected by the telemetry logging, namely the python executable path. If one has an endpoint for Azure Logs, the messages will now look like this:

image

I could not figure out how to properly type the `LogRecord` instance to account for the injected property, hence the `# type: ignore` comment. Judging from the logging module's [documentation on filters](https://docs.python.org/3/library/logging.html#logging.Filter.filter), I think that this inline modification is correct.

The PR also includes some pre-commit changes.

  @QCoDeS/core


qcodes/logger/logger.py
```diff
@@ -184,6 +184,63 @@ def flush_telemetry_traces() -> None:
...
+            super().__init__()
+            self.custom_dimensions = custom_dimensions
+
+        def filter(self, record: logging.LogRecord) -> bool:
+            """
+            Add the default custom_dimensions into the current log record
+            """
+            cdim = self.custom_dimensions.copy()
+            cdim.update(getattr(record, "custom_dimensions", {}))
+            record.custom_dimensions = cdim  # type: ignore
```
jenshnielsen, line 215: We normally are explicit about which error to ignore. Here I think you should do `# type: ignore[attr-defined]`

WilliamHPNielsen: That's much nicer, thanks.

astafan8 (approved):

PR #3160, opened 2021-06-28
Make measurement current parameter status aware

Akshita07:
- Update current parameter to be aware of if current compliance is hit during measurement.

To do:
[x] Test on actual instrument

@astafan8 @jenshnielsen @trevormorgan


qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -190,6 +191,65 @@ def get_raw(self) -> np.ndarray:  # pylint: disable=E0202
...
 
+class _ParameterWithStatus(Parameter):
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+
+        self._measurement_status: Optional[str] = None
+
+    @property
+    def measurement_status(self) -> Optional[str]:
+        return self._measurement_status
```
jenshnielsen, line 202: Rather than letting the measurement status be a string that one has to match against it would perhaps be better with a string enum?

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -190,6 +191,65 @@ def get_raw(self) -> np.ndarray:  # pylint: disable=E0202
...
+
+    def get_raw(self) -> ParamRawDataType:
+        assert isinstance(self.instrument, KeithleyChannel)
+
+        smu = self.instrument
+        channel = self.instrument.channel
+
+        value = float(smu.ask(f'{channel}.measure.i()'))
+
+        self._meas_status()
```
jenshnielsen, line 248: This means that every time we get a value from the instrument we will have to ask it twice. This seems to likely be a significant slowdown. Is there a way to either get this from a status register or perhaps ask about both at the same time e.g. perhaps one can chain the commands ?

Akshita07: Ok, yes, I'll check the possibility of chaining the commands.

astafan8 (approved):

PR #3170, opened 2021-07-02
Use thread pool when acquiring parameters to save time on recreating threads, new ThreadPoolParamsCaller, use it in do1d, do2d, dond

astafan8:
Use pool of threads (python's ``ThreadPoolExecutor``) to call parameters in different threads and re-use the threads as opposed to ``call_params_threaded`` that creates new threads for each call. Implemented as ``ThreadPoolParamsCaller`` class. ``do1d``, ``do2d`` and ``dond`` functions now use ``ThreadPoolParamsCaller``.

Usage example for a 1D sweep of multiple dependent parameters:
```python
...

meas.register_parameter(param_X)
for dependent_parameter in dependent_parameters:
    meas.register_parameter(dependent_parameter, setpoints=(param_X,))

with (
        meas.run() as datasaver,
        ThreadPoolParamsCaller(*dependent_parameters) as pool_caller
):
    for x_val in x_values:
        param_X.set(x_val)
        datasaver.add_results((param_X, x_val), *pool_caller())

...
```

Note that ``call_params_threaded`` is kept as is for now.


jenshnielsen (approved):

jenshnielsen:
Why is it not used in do0d?

jenshnielsen (approved):

astafan8:
> Why is it not used in do0d?

because i thought it's a readability overkill to create a pool for a single "parameters call", so kept the simpler "process_params_meas" function there. It can be easily changed in the future, if, say, "process_params_meas" is to be removed.

PR #3179, opened 2021-07-05
Update the get started document with installing and using spyder and jupyter

FarBo:
This PR:

clarifies how to install and use jupyter and spyder and Anaconda.


docs/start/index.rst
```diff
@@ -191,24 +191,46 @@ are two widely  used options:
...
-    conda activate qcodes
+    spyder
+
+or for jupyter notebook:
+
+.. code:: bash
+
+    jupyter notebook
+
+or jupyter lab:
```
jenshnielsen, line 219: Nitpick I don't think jupyter lab works unless you install jupyterlab

FarBo: Let me double check that.

astafan8 (approved):

docs/start/index.rst
FarBo: You are right. Mentioned the installation in the last commit.
Also added a line in the Anaconda section to say to be carefull about using it.

PR #3184, opened 2021-07-07
Fix numpy deprecation for ragged arrays

jenshnielsen:
Extract method from where this was already fixed and reuse it.

Perhaps we should think of another location for this method.


astafan8 (approved):
looks good!

> Perhaps we should think of another location for this method.

``qcodes.utils.numpy``? :)

jenshnielsen:
Not a fan of utils but that is probably fine

jenshnielsen:
I don't think having a file called numpy.py is a good idea so I opted for numpy_utils.py

PR #3185, opened 2021-07-08
Decouple measurement name from dataset results table name 

Akshita07:
- Decouple measurement name from dataset results table name to allow special characters in the measurement name
- Update failing tests

@jenshnielsen @astafan8

### DIFF
--- a/qcodes/dataset/measurements.py
+++ b/qcodes/dataset/measurements.py
@@ -511,7 +511,7 @@ class Runner:
         self.station = station
         self._interdependencies = interdeps
         self._shapes: Shapes = shapes
-        self.name = name
+        self.name = name if name else 'results'
         self._parent_datasets = parent_datasets
         self._extra_log_info = extra_log_info
         self._write_in_background = write_in_background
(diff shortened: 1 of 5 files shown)


qcodes/dataset/measurements.py
```diff
@@ -556,7 +556,7 @@ def __enter__(self) -> DataSaver:
         if self._dataset_class is DataSet:
             dataset_class = cast(Type[DataSet], self._dataset_class)
             self.ds = dataset_class(
-                name=self.name,
+                name='results',
```
jenshnielsen, line 559: I don't think this is the level at which we want to change this. We still want to pass the name to dataset so that it can be used in the name column of the runs table. 

What we want to do is to change `_insert_run` so that `formatted_name` does not contain the measurement name but is just of a default format of `results-expid-counter`

jenshnielsen: Screenshot 2021-07-08 121533

As in this example the name column should stay the same but the result_table_name should change

Akshita07: Ok, I'll update it

astafan8 (approved):

astafan8 (approved):
looks good now :)

PR #3188, opened 2021-07-08
Make it possible to set parameters as attribues and document these using a parser

### LINKED ISSUE #1151: Extract doc strings from dynanically added paramters

jenshnielsen:
This is an alternative suggestion to #3098. Since #32 was solved by #3125 I consider that out of scope for both these prs.

This enables parameters to be expressed as attributes on the instrument directly without having to call add_parameter. This has the advantage that the code exactly matches what one would expect a parameter to be namely an attribute on the instrument. 

Sphinx automatically documents attributes if they have a docstring enabling this to fix #1151 However the documentation for instance attributes is less than ideal by default. This therefor also includes a very early proof of concept sphinx extension that instantiates parameter attributes when calling getattr on an instrument. It does this by parsing the code and extracting any argument that contains self (since the instrument class cannot be instantiated) and then instantiating the parameter without these. 

The attributes are documented via their `__repr__` which has been overwritten here to prove the concept. 
This parser would need significant further work. One could either do

* Use autodoc-process-docstring/signature as in #3098 to provide a better formatting than the repr
* Rather than instantiating the object we could create a proxy object that only contains a repr as we want it. 

Compared to the old way of add_parameter this has a few disadvantages.

* One must explicitly pass self as instrument to the parameter. 
* Name is repeated both as the attribute name and as the name passed to the parameter. 

I do however feel that the added clarity of the code more that out ways these issues. 

Note that some tests fail due to the change in repr.
If we go down this route we should probably do the same for qcodes functions and submodules eventually.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -41,12 +41,22 @@ sys.path.insert(0, os.path.abspath('..'))
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ['nbsphinx', 'sphinx.ext.autodoc', 'sphinx.ext.autosummary',
-              'sphinx.ext.napoleon', 'sphinx-jsonschema', 'sphinx.ext.doctest',
-              'sphinx.ext.intersphinx', 'sphinx.ext.todo',
-              'sphinx.ext.coverage', 'sphinx.ext.mathjax',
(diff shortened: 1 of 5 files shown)


jenshnielsen:
replaced by #3224

PR #3191, opened 2021-07-09
Make parameter registration on instrument part of parameter init

jenshnielsen:
Such that parameters can be regular attributes on the instrument. This is lifted from #3188


qcodes/instrument/parameter.py
```diff
@@ -288,8 +288,19 @@ def __init__(
...
+        bind_to_instrument: bool = True,
     ) -> None:
+
+        if instrument is not None and bind_to_instrument:
+            existing_parameter = instrument.parameters.get(name, None)
+
+            if existing_parameter:
+
+                if not existing_parameter.abstract:
+                    raise KeyError(f"Duplicate parameter name {name}")
```
astafan8, line 301:
```suggestion
                    raise KeyError(f"Duplicate parameter name {name} on instrument {instrument}")
```

qcodes/instrument/parameter.py
```diff
@@ -288,8 +288,19 @@ def __init__(
         snapshot_exclude: bool = False,
         max_val_age: Optional[float] = None,
         vals: Optional[Validator[Any]] = None,
-        abstract: Optional[bool] = False
+        abstract: Optional[bool] = False,
+        bind_to_instrument: bool = True,
```
astafan8, line 292: need to add it to docs (also abstract arg is not in the docs), i gues, to both _Base and normal Parameter?

qcodes/instrument/parameter.py
```diff
@@ -1130,20 +1141,49 @@ class Parameter(_BaseParameter):
...
+                if not existing_parameter.abstract:
+                    raise KeyError(f"Duplicate parameter name {name}")
+
+                existing_unit = getattr(existing_parameter, "unit", None)
+                if existing_unit != unit:
+                    raise ValueError(
+                        f"The unit of the parameter '{name}' is '{unit}'. "
+                        f"This is inconsistent with the unit defined in the "
+                        f"base class"
+                    )
```
astafan8, line 1177: if below i explain/undersatnd it correctly, then let's keep the redundancy. OR move the unit check after the super.init?
```suggestion
            if existing_parameter:

                # this check is redundant since its also in the baseclass
                # but if we do not put it here it would be an api break
                # as parameter duplication check won't be done first,
                # hence for parameters that are duplicates ANd have
                # wrong units, users will be getting ValueError where 
                # they used to have KeyError before.
                if not existing_parameter.abstract:
                    raise KeyError(f"Duplicate parameter name {name}")

                existing_unit = getattr(existing_parameter, "unit", None)
                if existing_unit != unit:
                    raise ValueError(
                        f"The unit of the parameter '{name}' is '{unit}'. "
                        f"This is inconsistent with the unit defined in the "
                        f"base class"
                    )
```

astafan8 (commented):
Don't we want to advertize in the notebook that this way ofdeclaring parameters is now possible?) or is that left for further PRs?

jenshnielsen:
> Don't we want to advertize in the notebook that this way ofdeclaring parameters is now possible?) or is that left for further PRs?

Yes it needs some docs work before merging. I broke this out so that we could confirm individually that this is sound (the other pr breaks tests due to non related reasons)

qcodes/instrument/parameter.py
jenshnielsen: I don't think putting after super init is good since that actually does the registration (Which would then have to be undone if unit is wrong)

qcodes/instrument/parameter.py
```diff
@@ -1130,20 +1141,51 @@ class Parameter(_BaseParameter):
...
+    ) -> None:
+        if instrument is not None and bind_to_instrument:
+            existing_parameter = instrument.parameters.get(name, None)
+
+            if existing_parameter:
+
+                # this check is redundant since its also in the baseclass
+                # but if we do not put it here it would be an api break
+                # as parameter duplication check won't be done first,
+                # hence for parameters that are duplicates ANd have
```
jenshnielsen, line 1167:
```suggestion
                # hence for parameters that are duplicates and have
```

qcodes/instrument/parameter.py
```diff
@@ -1130,20 +1141,51 @@ class Parameter(_BaseParameter):
...
+            if existing_parameter:
+
+                # this check is redundant since its also in the baseclass
+                # but if we do not put it here it would be an api break
+                # as parameter duplication check won't be done first,
+                # hence for parameters that are duplicates and have
+                # wrong units, users will be getting ValueError where 
+                # they used to have KeyError before.
+                if not existing_parameter.abstract:
+                    raise KeyError(f"Duplicate parameter name {name}")
```
jenshnielsen, line 1171:
```suggestion
                    raise KeyError(f"Duplicate parameter name {name} on instrument {instrument}")
```

qcodes/instrument/base.py
```diff
@@ -172,24 +172,7 @@ def add_parameter(self, name: str,
...
-            if existing_unit != new_unit:
-                raise ValueError(
-                    f"The unit of the parameter '{name}' is '{new_unit}'. "
-                    f"This is inconsistent with the unit defined in the "
-                    f"base class"
-                )
-
-        param = parameter_class(name=name, instrument=self, **kwargs)
-        self.parameters[name] = param
+        parameter_class(name=name, instrument=self, **kwargs)
```
jenshnielsen, line 175: We should probably be more defensive here and check that the parameter has been registred on the instruement. So that we better handle errors such as the ones fixed by #3187 

If not we could do one of.

* Register it here and warn that this is wrong e.g. the parameter subclass needs fixing
* Error out here and ask the user to fix the parameter class
* do 1 and turn it into 2 after some time

astafan8: > do 1 and turn it into 2 after some time
 :)

jenshnielsen: Done

qcodes/instrument/base.py
```diff
@@ -114,24 +118,30 @@ def add_parameter(self, name: str,
...
+            param = parameter_class(name=name, instrument=self, **kwargs)
+        except TypeError:
+            kwargs.pop("bind_to_instrument")
+            warnings.warn(
+                f"Parameter {name} on instrument {self.name} does "
+                f"not correctly pass kwargs to its baseclass. A "
+                f"Parameter class must take `**kwargs` and forward "
+                f"them to its baseclass."
+            )
+            param = parameter_class(name=name, instrument=self, **kwargs)
```
jenshnielsen, line 134: Need a test for this corner case. And need to verify that all parameter subclasses in qcodes allows forwarding of kwargs,

qcodes/instrument/parameter.py
```diff
@@ -1503,8 +1562,15 @@ def _update_with(self, *,
...
-                 **kwargs: Any):
+    def __init__(
+        self,
+        name: str,
+        source: Optional[Parameter],
+        *args: Any,
+        **kwargs: Any,
+    ):
+        if "bind_to_instrument" not in kwargs.keys():
+            kwargs["bind_to_instrument"] = False
```
jenshnielsen, line 1573: Should be documented in the docstring for delegate parameter

qcodes/tests/instrument_mocks.py
```diff
@@ -128,13 +124,52 @@ def __init__(self, name: str = 'dummy',
...
+class DummyAttrInstrument(Instrument):
+    def __init__(self, name: str = "dummy", **kwargs: Any):
+
+        """
+        Create a dummy instrument that can be used for testing
+
+        Args:
+            name: name for the instrument
+            gates: list of names that is used to create parameters for
+                            the instrument
```
jenshnielsen, line 148: out of date docstring

qcodes/instrument/delegate/delegate_instrument.py
```diff
@@ -237,14 +237,15 @@ def _add_parameter(
...
     ) -> DelegateGroupParameter:
+        param_name = "{group_name}_{name}"
+        if param_name in self.parameters:
+            raise KeyError(f"Duplicate parameter name {param_name} on {self.name}")
         param = self.param_cls(
             name=f"{group_name}_{name}",
             instrument=self,
-            source=source
+            source=source,
+            bind_to_instrument=True,
```
jenshnielsen, line 247: For safety this should probably also be wrapped in a try catch with parameters that do not take bind_to_instrument as a kwarg and handle parameters that do not pass instrument correctly

docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb
```diff
@@ -83,7 +83,15 @@
...
-    "In most cases you will probably be adding parameters via the `add_parameter` method on the instrument class as shown in the example below. \n",
+    "There are two different ways of adding parameters to instruments. They are almost equivalent but comes with some trade-offs. We will show both below.\n",
+    "You may either declare the parameter as an attribute directly on the instrument or add it via the via the `add_parameter` method on the instrument class.\n",
+    "\n",
+    "Declaring a parameter as an attribute directly on the instrument enables Sphinx, IDEs such as VSCode and static tools such as Mypy to work more fluently with \n",
+    "the parameter than if it is created via `add_parameter` however you must take care to remember to pass `instrument=self` to the parameter such that the\n",
+    "parameter will know which instrument it belongs to. \n",
+    "Instrument.add_parameter is better suited for when you want to dynamically or programmatically add a parameter to an instrument. For historical reasons most \n",
+    "instruments currently use `add_parameter`.\n",
+    "\n",
```
jenshnielsen, line 94: Perhaps this notebook should document that parameters must pass kwargs to the baseclass

qcodes/instrument/delegate/delegate_instrument.py
jenshnielsen: This is done now just need to write a test for it

qcodes/instrument/base.py
```diff
@@ -114,24 +118,31 @@ def add_parameter(self, name: str,
                 unit of the new parameter is inconsistent with the existing
                 one.
         """
-        existing_parameter = self.parameters.get(name, None)
-
-        if existing_parameter:
+        if "bind_to_instrument" not in kwargs.keys():
```
astafan8, line 121: ``in`` will look in keys already ;)
```suggestion
        if "bind_to_instrument" not in kwargs:
```

qcodes/instrument/delegate/delegate_instrument.py
jenshnielsen: rewrote this to use add_parameter to avoid code duplication

qcodes/instrument/parameter.py
```diff
@@ -288,8 +291,21 @@ def __init__(
...
+            existing_parameter = instrument.parameters.get(name, None)
+
+            if existing_parameter:
+
+                if not existing_parameter.abstract:
+                    raise KeyError(
+                        f"Duplicate parameter name {name} on instrument {instrument}"
+                    )
+
+            instrument.parameters[name] = self
```
astafan8, line 308: shall we do this after ``if not str(name).isidentifier():`` check? also after the __init__ of the parent class (metadatable)? or would it break something?

qcodes/tests/parameter/test_parameter_registration.py
```diff
@@ -0,0 +1,69 @@
+import pytest
+
+from qcodes.instrument.parameter import Parameter
+from qcodes.tests.instrument_mocks import DummyAttrInstrument
+from qcodes.utils.deprecate import QCoDeSDeprecationWarning
+
+
+class BrokenParameter(Parameter):
+    """A parameter that incorrectly does not forward instrument to the base parameter"""
```
astafan8, line 9:
```suggestion
    """
    A parameter that incorrectly sets instrument to the super-class parameter,
    instead, it should forward it via ``instrument`` argument
    """
```

qcodes/instrument/base.py
jenshnielsen: Imho being explicit makes it more readable but I don't care strongly

qcodes/tests/parameter/test_parameter_registration.py
jenshnielsen: Imho a bit over the top for a test helper class but sure why not ;)

docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb
```diff
@@ -211,46 +223,57 @@
...
+    "            'curr',\n",
+    "            get_cmd='{}.measure.i()'.format(channel),\n",
+    "            get_parser=float,\n",
+    "            set_cmd='{}.source.leveli={}'.format(channel, '{:.12f}'),\n",
+    "            label='Current',\n",
+    "            unit='A',\n",
+    "            instrument=self\n",
+    "        )\n",
+    "\n",
+    "        self.mode(\n",
```
astafan8, line 249:
```suggestion
    "        self.mode = Parameter(\n",
```

qcodes/instrument/base.py
```diff
@@ -114,24 +118,31 @@ def add_parameter(self, name: str,
...
+                QCoDeSDeprecationWarning,
+            )
+            param = parameter_class(name=name, instrument=self, **kwargs)
 
-        param = parameter_class(name=name, instrument=self, **kwargs)
-        self.parameters[name] = param
+        existing_parameter = self.parameters.get(name, None)
+        if not existing_parameter:
+            warnings.warn(
+                f"Parameter {name} does not correctly register on instrument"
```
astafan8, line 140:
```suggestion
                f"Parameter {name} did not correctly register itself on instrument"
```

qcodes/instrument/base.py
```diff
@@ -114,24 +118,31 @@ def add_parameter(self, name: str,
...
+                f" {self.name}. Please check that instrument argument is passed "
```
astafan8, line 141:
```suggestion
                f" {self.name}. Please check that `instrument` argument is passed "
```

qcodes/instrument/base.py
```diff
@@ -114,24 +118,31 @@ def add_parameter(self, name: str,
...
+                f"to '_BaseParameter'. This will be an error in the future.",
```
astafan8, line 142: is the correction correct, or am i misunderstanding something?
```suggestion
                f"from {parameter_class!r} all the way to `_BaseParameter`. "
                "This will be an error in the future.",
```

astafan8 (approved):

docs/examples/writing_drivers/Creating-Instrument-Drivers.ipynb
jenshnielsen: Thanks, applied manually just to be safe

PR #3197, opened 2021-07-12
Revert checking of abstract parameters in post init

jenshnielsen:
It turns out that this approach has unintended side effects when multiple inheritance is used. This will effectively generate init methods for any class that does not have an init method but inherits from something that has an init method. In multiple inheritance scenarios this may change the order in which init methods are called. See https://gist.github.com/jenshnielsen/c3f6457f3510d638722c58e1e163ddee for an example of how this can happen.

### DIFF
--- a/qcodes/tests/test_abstract_instrument.py
+++ b/qcodes/tests/test_abstract_instrument.py
@@ -78,7 +78,7 @@ class VoltageSourceSubSub(VoltageSource):
         super().__init__(name)
 
     def __post_init__(self, *args: Any, **kwargs: Any) -> None:
-        super().__post_init__()
+        super().__post_init__()  # type: ignore[misc]
         self.call_count += 1
 
 
--- a/qcodes/tests/test_abstract_instrument.py
+++ b/qcodes/tests/test_abstract_instrument.py
@@ -120,6 +120,7 @@ def test_sanity(driver):
     assert driver.voltage() == 0.1
 
 
+@pytest.mark.skip("tests unimplemented feature")
 def test_not_implemented_error():
     """
     If not all abstract parameters are implemented, we should see
@@ -138,9 +139,11 @@ def test_unit_value_error():
     """
     with pytest.raises(ValueError, match="This is inconsistent with the unit defined"):
         VoltageSourceBadUnit("driver3")
-    assert not VoltageSourceBadUnit.instances()
+
+    # assert not VoltageSourceBadUnit.instances()
 
 
+@pytest.mark.skip("tests unimplemented feature")
 def test_exception_in_init():
     """
     In previous versions of QCoDeS, if an error occurred in
@@ -162,6 +165,7 @@ def test_exception_in_init():
     instance.close()
 
 
+@pytest.mark.skip("tests unimplemented feature")
 def test_subsub():
     """
     Verify that the post init method is only called once, even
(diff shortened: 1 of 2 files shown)


qcodes/tests/test_abstract_instrument.py
```diff
@@ -78,7 +78,7 @@ def __init__(self, name: str):
         super().__init__(name)
 
     def __post_init__(self, *args: Any, **kwargs: Any) -> None:
-        super().__post_init__()
+        super().__post_init__()  # type: ignore[misc]
         self.call_count += 1
```
astafan8, line 82: since post init does not exist, shall we remove it from here as well?

qcodes/tests/test_abstract_instrument.py
```diff
@@ -138,9 +139,11 @@ def test_unit_value_error():
     """
     with pytest.raises(ValueError, match="This is inconsistent with the unit defined"):
         VoltageSourceBadUnit("driver3")
-    assert not VoltageSourceBadUnit.instances()
 
+    # assert not VoltageSourceBadUnit.instances()
```
astafan8, line 143: can't this be left in as the original intent of the test, now that it's skipped?

qcodes/tests/test_abstract_instrument.py
```diff
@@ -172,6 +176,7 @@ def test_subsub():
     assert instance.call_count == 1
 
 
+@pytest.mark.skip("tests unimplemented feature")
```
astafan8, line 179: shall we do xfail instead? so that these stay runnable during other development work?
```suggestion
@pytest.mark.xfail("tests unimplemented feature")
```

astafan8 (approved):

qcodes/tests/test_abstract_instrument.py
jenshnielsen: This test is not skipped ;) Since the core of the test still works correctly

qcodes/tests/test_abstract_instrument.py
jenshnielsen: This class is only used in the skipped tests. Removing it from here would make the test harder to reestablish

qcodes/tests/test_abstract_instrument.py
jenshnielsen: We could do that but it would require a significant amount of work to make these test run cleanly since they now leak partially created instruments. I do not think it is worth the effort.

astafan8: fair

qcodes/tests/test_abstract_instrument.py
astafan8: assuming that it's a similar thing that will be re-established :)

astafan8: ok, fair

qcodes/tests/test_abstract_instrument.py
jenshnielsen: Why apart from this check the test passes and mostly tests the functionality that is not reverted

astafan8: why - so that this test is also preserved as is, and is executed, but it's failure won't influence the test summary. however, as you see appropriate :)

jenshnielsen: Lets split it into two to make this more clear

PR #3199, opened 2021-07-13
Fix delegate test

jenshnielsen:
* Test needs an experiment to add data to otherwise it cannot run standalone.
* Test was setting v on a device with non zero post delay and a small step size making it much slower than needed


jenshnielsen:
Without this fix the test takes ~ 93 s with it it takes ~ 0.5 s

astafan8 (approved):
amazing!!

PR #3210, opened 2021-07-16
Use qcodes config to have active experiment

FarBo:
This PR :

- Adds a setting in qcodes config file with the active_experiment key to hold the active_experiment exp_id.
- It adds set_active_experiment function to change the config and runs this function inside every create/ load of experiment related functions. There is also a get_active_experiment function to read the config.
- DataSet creation now reads the active_experiment setting first to assign that dataset to the active_experiment, if active_experiment is None, it reads the last experiment as before.
- Resets the active_experiment config setting upon creating/ loading a database

**The PR is on-hold and will remain as draft **

### DIFF
--- a/qcodes/configuration/qcodesrc.json
+++ b/qcodes/configuration/qcodesrc.json
@@ -72,7 +72,8 @@
         "export_automatic": false,
         "export_type": null,
         "export_prefix": "qcodes_",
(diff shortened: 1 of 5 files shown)


jenshnielsen:
closing since #3214 was merged

PR #3214, opened 2021-07-19
Default experiment in qcodes

FarBo:
This PR:

- Introduces functions to set, get and reset the active experiment. Upon loading or creating an experiment, _set_active_experiment_id will be called. get_active_experiment_id is a function to return the active exp_id in the kernel, and reset_active_experiment_id is a function to reset the active experiment to default which is the maximum exp_id in the database.
- The functions are created in qcodes.dataset.experiment_setting.
- dataset creation picks the active experiment exp_id  automatically.
- We make sure to reset active experiment upon load/ create database
- test function to hit the active experiment 
- Experiment notebook updated

### DIFF
--- a/docs/examples/DataSet/The-Experiment-Container.ipynb
+++ b/docs/examples/DataSet/The-Experiment-Container.ipynb
@@ -31,14 +31,14 @@
      "output_type": "stream",
      "text": [
       "Upgrading database; v0 -> v1: : 0it [00:00, ?it/s]\n",
-      "Upgrading database; v1 -> v2: 100%|█████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 250.17it/s]\n",
+      "Upgrading database; v1 -> v2: 100%|█████████████████████████████████████████████████████| 1/1 [00:00<00:00, 333.17it/s]\n",
       "Upgrading database; v2 -> v3: : 0it [00:00, ?it/s]\n",
       "Upgrading database; v3 -> v4: : 0it [00:00, ?it/s]\n",
-      "Upgrading database; v4 -> v5: 100%|█████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 199.91it/s]\n",
+      "Upgrading database; v4 -> v5: 100%|█████████████████████████████████████████████████████| 1/1 [00:00<00:00, 334.18it/s]\n",
       "Upgrading database; v5 -> v6: : 0it [00:00, ?it/s]\n",
-      "Upgrading database; v6 -> v7: 100%|██████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 62.49it/s]\n",
-      "Upgrading database; v7 -> v8: 100%|█████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 200.00it/s]\n",
-      "Upgrading database; v8 -> v9: 100%|█████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 142.75it/s]\n"
+      "Upgrading database; v6 -> v7: 100%|█████████████████████████████████████████████████████| 1/1 [00:00<00:00, 125.02it/s]\n",
+      "Upgrading database; v7 -> v8: 100%|█████████████████████████████████████████████████████| 1/1 [00:00<00:00, 250.00it/s]\n",
+      "Upgrading database; v8 -> v9: 100%|█████████████████████████████████████████████████████| 1/1 [00:00<00:00, 247.69it/s]\n"
      ]
     }
    ],
@@ -68,7 +68,7 @@
    "outputs": [],
    "source": [
     "from qcodes.dataset.experiment_container import experiments\n",
-    "from qcodes.dataset.experiment_setting import get_active_experiment_id"
+    "from qcodes.dataset.experiment_settings import get_active_experiment_id"
    ]
   },
   {
@@ -95,7 +95,11 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Not surprisingly, our new database is empty. If no experiment is created or loaded in the python kernel, the return value of `get_active_experiment_id()` will be None:"
+    "Not surprisingly, our new database is empty.\n",
+    "\n",
+    "We now introduce a term that we call it **active experiment**. In short, the active experiment is the experiment that is used when a QCoDeS `DataSet` is generated, if the user do not explicitly pass an experiment into objects that create that `DataSet`. So, that `DataSet` will belong to the experiment that is active. We do not want to go into the details of `DataSet` here, and refer to the [DataSet notebook](https://qcodes.github.io/Qcodes/examples/DataSet/DataSet-class-walkthrough.html) and [Performing measurements using qcodes parameters and dataset](https://qcodes.github.io/Qcodes/examples/DataSet/Performing-measurements-using-qcodes-parameters-and-dataset.html) for what we mean from `DataSet` and explicitly passing of an experiment.\n",
+    "\n",
+    "By default, `get_active_experiment_id()` will return the maximum **exp_id** in the database, if you load a database already has experiment(s) in it. If the database has no experiment and you have not created or loaded any experiment, the return value will be `None`: "
    ]
   },
   {
@@ -111,7 +115,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Let us add some experiments:"
+    "Let us add some experiments to explore more:"
    ]
   },
   {
@@ -127,25 +131,13 @@
    "cell_type": "code",
    "execution_count": 7,
    "metadata": {},
-   "outputs": [
-    {
-     "data": {
-      "text/plain": [
-       "third_exp#brand_new_sample#3@C:\\Users\\a-fbonabi\\temp_work\\exp_container_tutorial.db\n",
-       "-----------------------------------------------------------------------------------"
-      ]
-     },
-     "execution_count": 7,
-     "metadata": {},
-     "output_type": "execute_result"
-    }
-   ],
+   "outputs": [],
    "source": [
-    "new_experiment('first_exp', sample_name='old_sample')\n",
-    "new_experiment('second_exp', sample_name='slightly_newer_sample')\n",
+    "exp_a = new_experiment('first_exp', sample_name='old_sample')\n",
+    "exp_b = new_experiment('second_exp', sample_name='slightly_newer_sample')\n",
     "# A more convenient function that can load an experiment\n",
     "# OR create a new one if it does not exist:\n",
-    "load_or_create_experiment('third_exp', sample_name='brand_new_sample')"
+    "exp_c = load_or_create_experiment('third_exp', sample_name='brand_new_sample')"
    ]
   },
   {
@@ -177,14 +169,14 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We notice that each experiment is labelled by an integer. This is the **exp_id** that can be used when looking up properties of each experiment."
+    "We notice that each experiment is labelled by an integer number, which is the `exp_id` that we mentioned above. This ID can be used when looking up properties of each experiment."
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Let's check again which experiment is active:"
+    "Let's check again which experiment is active now:"
    ]
   },
   {
@@ -211,14 +203,14 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The latest created or loaded experiment in the kernel becomes the active experiment and the function returns the `exp_id` of the active experiment."
+    "The latest created or loaded experiment in the kernel becomes the active experiment and the function returns the `exp_id` of that experiment."
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Let us add some runs to our experiments. For the sake of clarity, we don't add any data to the runs here. We refer to the [Performing measurements using qcodes parameters and dataset](Performing-measurements-using-qcodes-parameters-and-dataset.ipynb) notebook for detailed information on how to properly create and populate runs. We also refer to the [DataSet-class-walkthrough](DataSet-class-walkthrough.ipynb) notebook for more details on the ``DataSet`` class. Note that the ``new_data_set`` function is used here ONLY for the sake of exercise and should NOT be used in the actual experiment."
+    "Let us add some `DataSet` to our experiments. For the sake of clarity, we don't add any data to the `DataSet` here, and refer to the above-mentioned notebooks for the details. Note that the `new_data_set` function is used here ONLY for the sake of exercise and should NOT be used in the actual experiment."
    ]
   },
   {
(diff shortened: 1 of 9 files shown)


qcodes/dataset/sqlite/database.py
```diff
@@ -271,6 +274,7 @@ def initialised_database_at(db_file_with_abs_path: str) -> Iterator[None]:
             Database file name with absolute path, for example
             ``C:\\mydata\\majorana_experiments.db``
     """
+    _reset_active_exp()
```
jenshnielsen, line 277: No need since initialise_database does not do this

qcodes/dataset/sqlite/database.py
```diff
@@ -257,6 +259,7 @@ def initialise_or_create_database_at(db_file_with_abs_path: str,
             Options are DELETE, TRUNCATE, PERSIST, MEMORY, WAL and OFF. If set to None
             no changes are made.
     """
+    _reset_active_exp()
```
jenshnielsen, line 262: No need since initialise_database does not do this

jenshnielsen (changes requested):
I think we need to spent some effort on reworking the imports. The sqlite modules should not import from the experiment container

qcodes/dataset/experiment_container.py
```diff
@@ -209,9 +226,11 @@ def new_experiment(name: str,
         the new experiment
     """
     conn = conn or connect(get_DB_location())
-    return Experiment(name=name, sample_name=sample_name,
-                      format_string=format_string,
-                      conn=conn)
+    experiment = Experiment(
+        name=name, sample_name=sample_name, format_string=format_string, conn=conn
+    )
+    _set_active_experiment_id(experiment.exp_id)
```
jenshnielsen, line 232: We could also move this into the Experiment constructor

jenshnielsen: @astafan8 what do you think?

qcodes/dataset/experiment_container.py
```diff
@@ -333,4 +356,5 @@ def load_or_create_experiment(experiment_name: str,
                                         conn=conn)
         else:
             raise exception
+    _set_active_experiment_id(experiment.exp_id)
```
jenshnielsen, line 359: As far as I can see this is not needed. Since both load_experiment_by_name and new_experiment does the same

qcodes/tests/dataset/test_experiment_container.py
```diff
@@ -306,3 +313,35 @@ def test_load_last_experiment(empty_temp_db):
     assert last_exp.exp_id == exp2.exp_id
     assert last_exp.exp_id != exp1.exp_id
     assert last_exp.path_to_db == exp2.path_to_db
+
+
+@pytest.mark.usefixtures("empty_temp_db")
+def test_active_experiment():
+    NA = None
```
jenshnielsen, line 320: I would not make this assignment and just compare with is None

qcodes/tests/dataset/test_experiment_container.py
```diff
@@ -306,3 +313,35 @@ def test_load_last_experiment(empty_temp_db):
...
+    assert get_active_experiment_id() == exp_2.exp_id
+    assert get_active_experiment_id() == exp_5.exp_id
+
+    exp_6 = load_last_experiment()
+    assert get_active_experiment_id() == exp_4.exp_id
+    assert get_active_experiment_id() == exp_6.exp_id
+
+    reset_active_experiment_id()
+    exp_id = get_active_experiment_id()
+    assert exp_id == NA
```
jenshnielsen, line 347:
```suggestion

    exp_id = get_active_experiment_id()
    assert exp_id is None

    exp_1 = load_or_create_experiment("test_exp", sample_name="no_sample")
    assert get_active_experiment_id() == exp_1.exp_id

    exp_2 = new_experiment("test_exp_2", sample_name="no_sample")
    assert get_active_experiment_id() == exp_2.exp_id

    exp_3 = load_experiment(1)
    assert get_active_experiment_id() == exp_1.exp_id
    assert get_active_experiment_id() == exp_3.exp_id

    exp_4 = new_experiment("test_exp_3", sample_name="no_sample")

    exp_5 = load_experiment_by_name("test_exp_2", sample="no_sample")
    assert get_active_experiment_id() == exp_2.exp_id
    assert get_active_experiment_id() == exp_5.exp_id

    exp_6 = load_last_experiment()
    assert get_active_experiment_id() == exp_4.exp_id
    assert get_active_experiment_id() == exp_6.exp_id

    reset_active_experiment_id()
    exp_id = get_active_experiment_id()
    assert exp_id is None
```

jenshnielsen (approved):

qcodes/dataset/experiment_container.py
FarBo: I initially wanted to do that, but then decided to put it into individual functions.
This move has a benefit that if someone directly use Experiment class, then this active_experiment also get triggered.

qcodes/tests/dataset/test_experiment_container.py
FarBo: I think I tried that but got error. I will check it again

qcodes/dataset/experiment_container.py
FarBo: AHHHH. Correct :)
One less active_experiment

docs/examples/DataSet/The-Experiment-Container.ipynb
```diff
@@ -159,115 +177,375 @@
...
+   ],
+   "source": [
+    "new_data_set('first_run', exp_id=1)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now, the alert reader will have noticed that `exp_id` is a keyword argument in QCoDeS runs. What is the default then? -- By default, the maximum `exp_id` in the database is the default value, and the experiment with that `exp_id` will be used in QCoDeS runs. However, if the user loads or creates an experiment in the Python session, that experiment will be activated and used in runs."
```
astafan8, line 443: let's use the word DataSet instead of "run"? i think it's inconsistent if all the methods/functions/classes talk about datasets while the notebook text is using "run"

qcodes/dataset/experiment_container.py
astafan8: my perspective is that the global side effect, which is setting the active experiment, should not be attached to the object itself so that there exists a way to create an experiment without activating it immediately. hence I agree with @FarBo .

so i think let's keep setting active experiment outside of the experiment object

qcodes/dataset/data_set.py
```diff
@@ -287,8 +288,10 @@ def __init__(self, path_to_db: Optional[str] = None,
...
             # with no parameters; they are written to disk when the dataset
             # is marked as started
-            if exp_id is None:
-                exp_id = get_last_experiment(self.conn)
+            if exp_id is None:  # First, try to get the active exp_id
+                exp_id = get_active_experiment_id()
+                if exp_id is None:  # If no active exp_id, get the last exp_id
+                    exp_id = get_last_experiment(self.conn)
                 if exp_id is None:  # if it's still None, then...
                     raise ValueError("No experiments found."
```
astafan8, line 296: i think my only real comment is about this place in the code. what happens here is that DataSet's contructor becomes the one owning the logic that "if there's no active experiment, then use the max-exp-id one, and if that's not there, then raise and ask user to create an experiment" - my feeling is that it's not DataSet's responsibility to know that, but rather the responsibility of the "active experiment" feature. If "get_active_experiment" would implement these details then DataSet (and potentially any other objects) can rely on the active experiment as default, and what happens when active experiemtn is not defined is not the business of the object that uses it. so, a bit of responsibility decoupling.

what do you think @FarBo @jenshnielsen ? if the selected code was actually part of get_active_experiment_id?

jenshnielsen: I don't think it should be part of that function necessarily but it would probably make sense to refactor it out next to that one as a higher level function.

FarBo: This is a fair point, and nice to decouple active_experiment from DataSet.
I will try to see what I can do :)

qcodes/dataset/experiment_settings.py
```diff
@@ -0,0 +1,55 @@
...
+    Checks if get_active_experiment_id is an existing exp_id and return it,
+    and if it is None, raise error that no experiment is initialized.
+    """
+    global active_experiment
+    if active_experiment is not None:
+        return active_experiment
+    else:
+        raise ValueError("No experiments found."
+                         " You can create one with:"
+                         " load_or_create_experiment(name, sample_name)")
```
jenshnielsen, line 55: My intention was not that we let new_experiment and initialize_database set the active experiment rather I suggested that we implement a function that would handle the logic implemented in dataset. 
Now circular imports may make it tricky to place this function in the experiment module or the settings module where you really would want it so I suggest that we put it in the dataset module for now.

Something like

```python
det get_default_experiment(conn) -> int:
    exp_id = get_active_experiment_id()
    if exp_id is None:
         exp_id = get_last_experiment(conn)
    if exp_id is None:  # if it's still None, then...
         raise ValueError("No experiments found."
                                     "You can start a new one with:"
                                     " new_experiment(name, sample_name)")
     return exp_id
```

FarBo: What I understood from @astafan8 comment was that we should 1. decouple this logic from DataSet, 2. potential use of this logic somewhere else. Your suggestion  somehow addresses point 1, but point 2 might be a problem due to dependency issue, if we want to use this function somewhere else.

Also, there is a small problem if I revert back setting active exp in the database: If you initialize a database having experiments and run `get_active_experiment_id` after that, it returns `None`, which is not correct, despite not being a major issue.

jenshnielsen: While that may be true your solution creates other unwanted couplings. We explicitly do not want to couple the notion of the active_dataset to the sqlite modules. (We have to live with the call to reset in initialize_database for now but that function should be pushed up one level) 

> Also, there is a small problem if I revert back setting active exp in the database: If you initialize a database having experiments and run get_active_experiment_id after that, it returns None, which is not correct, despite not being a major issue.

This is fully intentional. If you want the default experiment you should call get_default_experiment. Active experiment is the one that is explicitly set. If that is not the case you will use the latest experiment

FarBo: OK,
Then I adapt to what you suggested.

qcodes/dataset/experiment_settings.py
```diff
@@ -1,55 +1,60 @@
...
     """
-    Checks if get_active_experiment_id is an existing exp_id and return it,
-    and if it is None, raise error that no experiment is initialized.
-    """
-    global active_experiment
-    if active_experiment is not None:
-        return active_experiment
-    else:
+    Returns the latest created/ loaded experiment's exp_id as the default
+    experiment. If it is None, maximum exp_id from an initialized database
```
jenshnielsen, line 44:
```suggestion
    experiment. If it is None, maximum exp_id from the currently active initialized database
```

qcodes/dataset/experiment_settings.py
FarBo: @jenshnielsen 

Please look at the commit  of "Redisigned the module"
I will modify the rest of files accordingly, but having this new structure seems working fine.
I don't see any good reason why we should make the final function private. What do you think?

qcodes/dataset/experiment_settings.py
```diff
@@ -1,55 +1,60 @@
...
+    is returned as the default. if no experiment found in the database, then
+    raises an error.
```
jenshnielsen, line 46:
```suggestion
    is returned as the default. if no experiment found in the database, a 
    ValueError is raised.
```

qcodes/tests/dataset/test_experiment_container.py
```diff
@@ -306,3 +312,33 @@ def test_load_last_experiment(empty_temp_db):
...
+    exp_5 = load_experiment_by_name("test_exp_2", sample="no_sample")
+    assert get_default_experiment_id(empty_temp_db_connection) == exp_2.exp_id
+    assert get_default_experiment_id(empty_temp_db_connection) == exp_5.exp_id
+
+    exp_6 = load_last_experiment()
+    assert get_default_experiment_id(empty_temp_db_connection) == exp_4.exp_id
+    assert get_default_experiment_id(empty_temp_db_connection) == exp_6.exp_id
+
+    reset_default_experiment_id()
+    with pytest.raises(ValueError):
```
jenshnielsen, line 343: This should not happen. It should return the last experiment. However as you are using the connection fixture and not passing the connection to new experiment the experiment is not actually created in this db but in the default db of the user. 

Thinking a bit more of this I think we need a per dbpath default experiment for this to be safe

qcodes/dataset/experiment_settings.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    exp_id = _get_latest_default_experiment_id()
+    if exp_id is None:
+        query = "SELECT MAX(exp_id) FROM experiments"
+        c = atomic_transaction(conn, query)
+        exp_id = c.fetchall()[0][0]
+    if exp_id is None:
+        raise ValueError("No experiments found."
+                         " You can create one with:"
+                         " new_experiment(name, sample_name)")
```
jenshnielsen, line 59: This code is duplicated from the queries module where it belongs. Importing that directly will create a circular import since that module imports connect from the database module (same module that defines the initialise_database function which uses a function from this module) The only reason for the import of connect is the function `get_guids_from_multiple_run_ids` which is implemented differently from all other queries. IMHO we should just break this function and put it inline with all the other ones by taking a connection and not a dbpath as input.

FarBo: Good catch. We can break the dependency and import get_last_experiment directly

qcodes/dataset/experiment_settings.py
```diff
@@ -0,0 +1,60 @@
+"""Settings that are indirectly related to experiments."""
+
+from typing import Optional
+from qcodes.dataset.sqlite.connection import ConnectionPlus, atomic_transaction
+
+# The default experiment's exp_id. This changes to the exp_id of a created/
+# loaded experiment.
+default_experiment: Optional[int] = None
```
jenshnielsen, line 8: Perhaps we should make the attribute private. @astafan8 what do you think?

qcodes/dataset/experiment_settings.py
jenshnielsen: This still need to make use of the query. Please do not resolve until this is done

qcodes/tests/dataset/test_experiment_container.py
FarBo: I was initially thinking that the reset function should not be used publically, but in a very certain case, it could be used.
I think this can be fixed by adding a connection argument to the reset function:

```
def reset_default_experiment_id(conn) -> None:
    """
    Resets the default_experiment to the connected database default.
    """
    global default_experiment
    default_experiment = get_last_experiment(conn)
```

And we should change the test to create a temp database

qcodes/dataset/experiment_settings.py
FarBo: It is private by its nature, so we can do that.

qcodes/tests/dataset/test_experiment_container.py
jenshnielsen: I am not convinced that is sufficient. I still think it would be better to index the active experiment by db path

FarBo: No, It is not. you are right

qcodes/dataset/experiment_settings.py
astafan8: agree with making this variable private (adding underscore at the beginning of it's name)

qcodes/dataset/experiment_settings.py
```diff
@@ -1,22 +1,23 @@
...
 # The default experiment's exp_id. This changes to the exp_id of a created/
 # loaded experiment.
-default_experiment: Optional[int] = None
+_default_experiment: Optional[int] = None
 
 
 def _set_default_experiment_id(exp_id: int) -> None:
     """
-    Sets the default_experiment to the exp_id of a created/ loaded experiment.
+    Sets the _default_experiment to the exp_id of a created/ loaded experiment.
```
jenshnielsen, line 14: There is no need to change the docstring to refer to the private variable. The private variable is exactly private since no one on the outside should know about it

jenshnielsen (commented):
@FarBo The code should now be ready please push your changes to the notebook and this should be good to go

FarBo:
@ Thanks Jens for the final work on this. I will finish the notebook today and push it, then we can merge it afterward.

PR #3222, opened 2021-07-22
get_guids_from_multiple_run_ids accepts connection over dbpath in the function [breaking change]

FarBo:
This is for releasing `qcodes.dataset.sqlite.queries` from importing `connect` function in `sqlite.database` that could potentially creates circular dependency. 
The test file also modified accordingly, with a minimal change in the test method.


jenshnielsen:
This breaking change should be documented in the changelog. Will you start a new changelog for 0.28 ?

FarBo:
I will start it after merging this PR

jenshnielsen (approved):

PR #3233, opened 2021-07-29
Simultaneous ramping of magnetic field in all 3 axes

astafan8:
Sweep AMI 3D magnet to (x,y,z) in time T or at vector ramp rate R, ramping all 3 axes together (in parallel), making a linear/vector trajectory from current field to the setpoint in 3D magnetic field space.

For working with ramp duration, see:
```python
i3d.ramp_simultaneously(setpoint=FieldVector(0.5, 1.0, 0.01), duration=2)
# See the docstring of this method for details
```

For working with the new vector ramp rate, see:
```python
i3d.ramp_mode("simultaneous")
i3d.vector_ramp_rate(0.05)  # requires all axes to have same units!!!
i3d.cartesian(x,y,z)  # or any other parameter call that initiates a ramp
```

Note that actually simultaneous ramping is already supported with the current driver, one just needs to set the `block_during_ramp` parameter to `False`, so that the driver does not wait for one axis to finish ramping before starting ramping next one, and then manually/explicitly wait until the ramping is done. However, the benefit of the new simultaneous ramp mode and the `ramp_simultaneously` method is that it automatically calculates and adjusts the ramp rates of the axes.

ToDos:
- [x] ensure that if the requested field change is 0 for one/two/all-three axes, so ramping "simply works" (e.g. the instrument attempts to ramp, find that it's already in the right place, and stops ramping)
- [x] handle the case when the requested field change for a given axis results in a ramp rate that is less than minimal ramp rate - i hope that the instrument will just use whatever the minimum ramp rate is
- [x] example notebook
- [x] test on instruments


dbwz8 (commented):
Did a quick look. Didn't see anything obviously wrong.

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -787,6 +949,13 @@ def _adjust_child_instruments(
                 instrument.set_field(value, perform_safety_check=False,
                                      block=self.block_during_ramp.get())
 
+    def wait_while_all_axes_ramping(self) -> None:
+        axes = (self._instrument_x, self._instrument_y, self._instrument_z)
+        while all(
```
astafan8, line 954:
```suggestion
        while any(
```

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -787,6 +949,13 @@ def _adjust_child_instruments(
...
+            axis_instrument.ramping_state() == "ramping" for axis_instrument in axes
```
astafan8, line 955: expose this so that users can use it in their own while loops

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -724,6 +723,133 @@ def find_ami430_with_name(ami430_name: str) -> AMI430:
...
+        self.log.debug(
+            f"Simultaneous ramp: start {start_field.repr_cartesian()} "
+            f"{common_field_units}"
+        )
+        self.log.debug(
+            f"Simultaneous ramp: delta {(setpoint - start_field).repr_cartesian()} "
+            f"{common_field_units}"
+        )
+
+        new_ramp_rates = self.calculate_ramp_rates_for(
```
astafan8, line 785: add a sumultaneous_ramp_rate parameter for the ramp rate along the vector, so the there's another way of doiojng this ramp which is:
```python
mode('simultaneous')
sumultaneous_ramp_rate(0.050) # T/s
cartesian(x,y,z)  # do the ramp
```

qcodes/instrument_drivers/american_magnetics/AMI430.py
FarBo: This is what we also changed in our test

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -787,6 +949,13 @@ def _adjust_child_instruments(
                 instrument.set_field(value, perform_safety_check=False,
                                      block=self.block_during_ramp.get())
 
+    def wait_while_all_axes_ramping(self) -> None:
+        axes = (self._instrument_x, self._instrument_y, self._instrument_z)
+        while any(
+            axis_instrument.ramping_state() == "ramping" for axis_instrument in axes
+        ):
+            self._sleep(self.ramping_state_check_interval())
```
JouriB, line 957: AMI3D does not have a _sleep method, the individual AMIs do. Raises an error for me now.

astafan8: done

qcodes/instrument_drivers/american_magnetics/AMI430.py
astafan8: done. extracted a ``any_axis_is_ramping`` method

qcodes/instrument_drivers/american_magnetics/AMI430.py
astafan8: @JouriB i've implemented this using:
```python
i3d.ramp_mode("simultaneous")
i3d.vector_ramp_rate(0.05)  # requires all axes to have same units!!!
i3d.cartesian(x,y,z)  # or any other parameter call that initiates a ramp
```

What do you think? does it work as expected?

astafan8: @JouriB tested and confirmed that all works perfectly!

jenshnielsen (approved):

PR #3251, opened 2021-08-09
Make measurement voltage parameter status aware Keithley 2600 smu

trevormorgan:
Update voltage parameter for Keithley2600 SMU to be aware of measurement status (similar to how the current parameter is setup).


qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -236,19 +250,26 @@ class _MeasurementCurrentParameter(_ParameterWithStatus):
...
-            return float(value), MeasurementStatus.NORMAL
+        self._measurement_status = status
+
+        return value
+
+
+class _MeasurementVoltageParameter(_ParameterWithStatus):
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
```
jenshnielsen, line 272: Since this init method just forwards all arguments to the super class I think we can remove it

jenshnielsen: The same is true for the _MeasurementCurrentParameter class but I guess we missed that in the review of that class

jenshnielsen (approved):

astafan8 (approved):

PR #3186, opened 2021-07-08
Enabeling dond function to handle multiple datasets

FarBo:
This PR:

Enables dond to accept grouped measuerment parameters and callables and open a dataset for each group.
This implementation preserves the old way of comma-separated passing meaurements and callables in the function construct. If this is the case, all those arguments will be categorized in one group and one dataset will be created.

- [x] Improve the implementation
- [x] writing tests
- [x] UAT
- [x] Update doNd notebook

PR #3265, opened 2021-08-16
raise if given more than one result for the same parameter name

jenshnielsen:
This would have prevented a bug identified in #3186 where multiple values for the same result was passed to add_result.

It turns out that a number of tests relied on being able to pass in multiple parameters with setpoints that had been pre expanded and therefore multiple values for any shared setpoints. Technically this is an api break and should therefor probably be documented as such. I do not expect that many users would pre expand the setpoints so I am not that concerned about it. 

This also revealed that one of the tests were buggy since it explicitly passes the same parameters multiple times (due to some code being commented out)

@trevormorgan could you take a look


jenshnielsen:
@trevormorgan Ping

qcodes/tests/dataset/test_datasaver.py
```diff
@@ -128,3 +127,29 @@ def test_saving_numeric_values_as_text(numeric_type, bg_writing):
     finally:
         data_saver.dataset.mark_completed()
         data_saver.dataset.conn.close()
+
+
+@pytest.mark.usefixtures("experiment")
+def test_duplicated_parameter_raises():
+    """
+    Test the saving numeric values into 'text' parameter raises an exception
```
astafan8, line 135:
```suggestion
    Test that passing same parameter multiple times raises an exception
```

trevormorgan:
@jenshnielsen changes look good to me.

PR #3283, opened 2021-08-20
Add post sweep actions to doND sweeps

FarBo:
This PR enables dond to take inner actions. The actions are passed in the Sweep instances.
The logic is as follows: actions for a sweep parameter will be called when the set value of that parameter is changed. I didn't put comments in the code, but I think it would be nice to add a few comments.

I couldn't find a a good way to unit test actions, so it is still missing if we want to have the test.

- [x] Update doNd notebook about the usage of actions in dond.

@jenshnielsen

### DIFF
--- a/qcodes/utils/dataset/doNd.py
+++ b/qcodes/utils/dataset/doNd.py
@@ -503,7 +503,7 @@ class AbstractSweep(ABC):
 
     @property
     @abstractmethod
-    def action(self) -> ActionsT:
+    def post_actions(self) -> ActionsT:
         """
         actions to be performed after setting param to its setpoint.
         """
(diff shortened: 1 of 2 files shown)


qcodes/utils/dataset/doNd.py
```diff
@@ -501,6 +501,14 @@ def num_points(self) -> int:
         """
         pass
 
+    @property
+    @abstractmethod
+    def action(self) -> ActionsT:
```
jenshnielsen, line 506: perhaps we could be more descriptive? e.g. something like `post_actions` ?

FarBo: agreed

jenshnielsen (approved):

PR #3289, opened 2021-08-23
Log all installed packages

WilliamHPNielsen:
Towards reproducible experiments! A small step is that we log the versions of installed packages in the environment where QCoDeS runs. 

- Add function to get a list (well, dict) of all installed packages and their versions
- Log this information upon import

I did a quick and dirty comparison of import times before and after these changes, and there is no significant overhead.

@QCoDeS/core

### DIFF
--- a/qcodes/logger/logger.py
+++ b/qcodes/logger/logger.py
@@ -6,6 +6,7 @@ the default configuration.
 """
 
 import io
+import json
 import logging
 
 # logging.handlers is not imported by logging. This extra import is necessary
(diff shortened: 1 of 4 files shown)


qcodes/logger/logger.py
```diff
@@ -332,16 +333,19 @@ def log_qcodes_versions(logger: logging.Logger) -> None:
     """
     Log the version information relevant to QCoDeS. This function logs
     the currently installed qcodes version, whether QCoDeS is installed in
-    editable mode, and the installed versions of QCoDeS' requirements.
+    editable mode, the installed versions of QCoDeS' requirements, and the
+    versions of all installed packages.
```
jenshnielsen, line 337: Do we want to keep logging both qcodes requirements and all installed packages. I guess we can add this now and eventually remove the other?

jenshnielsen (approved):

qcodes/logger/logger.py
WilliamHPNielsen: Yes, that sounds like a good approach. The requirements call adds no new information, although it sometimes conveniently marks a package "not installed". Let's remove it after the installed packages have been logged in one or two releases.

PR #3296, opened 2021-08-24
[WIP] load experiment accepts duplication

FarBo:
This PR:

1. Let load functions that accepting name to avoid raising error if the new bool argument is set to True.
2. new_experiment warns if it creates a duplicate experiment

- [ ] new unit test

### DIFF
--- a/qcodes/dataset/experiment_container.py
+++ b/qcodes/dataset/experiment_container.py
@@ -1,5 +1,6 @@
 import logging
 from collections.abc import Sized
+import sqlite3
 from typing import Any, List, Optional
 
 from qcodes.dataset.data_set import SPECS, DataSet, load_by_id, new_data_set
@@ -218,6 +219,10 @@ def new_experiment(name: str,
         the new experiment
     """
     conn = conn or connect(get_DB_location())
+    rows = _find_exp_rows(name=name, sample=sample_name, conn=conn)
+    if len(rows) > 1:
+        log.warn(f"There is already experiment(s) with the name of {name} "
+                 f"and sample name of {sample_name} in the database.")
     experiment = Experiment(
         name=name, sample_name=sample_name, format_string=format_string, conn=conn
     )
@@ -264,16 +269,22 @@ def load_last_experiment() -> Experiment:
 
 def load_experiment_by_name(name: str,
                             sample: Optional[str] = None,
-                            conn: Optional[ConnectionPlus]=None) -> Experiment:
+                            load_last_duplicate: bool = False,
+                            conn: Optional[ConnectionPlus]=None,
+                            ) -> Experiment:
     """
     Try to load experiment with the specified name.
 
     Nothing stops you from having many experiments with the same name and
-    sample_name. In that case this won't work. And warn you.
+    sample_name. In that case this won't work unless load_last_duplicate
+    is set to True. Then, the last of duplicated experiments will be loaded.
 
     Args:
         name: the name of the experiment
         sample: the name of the sample
+        load_last_duplicate: To prevent raising error for having multiple
+            experiment with the same name and sample name, and to load the
+            last duplicated experiment.
         conn: connection to the database. If not supplied, a new connection
           to the DB file specified in the config is made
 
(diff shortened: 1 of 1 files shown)


qcodes/dataset/experiment_container.py
```diff
@@ -326,28 +322,70 @@ def load_experiment_by_name(name: str,
...
     except ValueError as exception:
         if "Experiment not found" in str(exception):
             experiment = new_experiment(experiment_name, sample_name,
                                         conn=conn)
         else:
             raise exception
     return experiment
+
+
+def _find_exp_rows(
```
jenshnielsen, line 358: I think we should be able to move this function to the queries module where it really belongs

FarBo: hmm. you are right. 
do you have any suggestion for the name or think it is fine?

qcodes/dataset/experiment_container.py
```diff
@@ -218,6 +219,10 @@ def new_experiment(name: str,
         the new experiment
     """
     conn = conn or connect(get_DB_location())
+    rows = _find_exp_rows(name=name, sample=sample_name, conn=conn)
+    if len(rows) > 1:
```
jenshnielsen, line 223: Should this not be >= 1 ?

qcodes/dataset/experiment_container.py
jenshnielsen: I think we should also consider letting the function return a list of exp ids. That way we isolate this code further from sqlite

jenshnielsen: We could call it `get_experiment_id_from_name_sample_name`  if that is not too verbose (assuming that we also change it to return a list of exp ids

FarBo: That would be also useful if it returns list of exp_ids, rather than the rows if we want to move it to queries and make it public. 
Then I need to change the way of deducting in line 307, which is easy.

FarBo: I think we don't need this function at all.
Looking into queries, I found this function `get_matching_exp_ids(conn, **match_conditions)`  which takes any kwarg. I tested it and it returns exp_id list for name and sample_name

qcodes/dataset/experiment_container.py
```diff
@@ -218,6 +219,10 @@ def new_experiment(name: str,
         the new experiment
     """
     conn = conn or connect(get_DB_location())
+    exp_ids = get_matching_exp_ids(conn, name=name, sample_name=sample_name)
```
jenshnielsen, line 222: get_matching_exp_ids works in a different way. If you pass in sample_name=None it will take it to mean experiments that really have no sample name not any sample name

FarBo: Yes. I found out that is the problem
Trying to solve it

FarBo: Then, we need to bring back that function that I replaced

jenshnielsen: Not necessarily. We could do something like ( I also think you should avoid renaming sample to sample_name)
```suggestion
   if sample is not None:
       arguments = {name: name, sample_name: sample}
   else:
       arguments = {name: name} 
    exp_ids = get_matching_exp_ids(conn, **arguments)
```

jenshnielsen: Just discovered that we also have this method _create_exp_if_needed which is very similar https://github.com/QCoDeS/Qcodes/blob/master/qcodes/dataset/database_extract_runs.py#L122

FarBo: Thanks
Look into that

qcodes/dataset/experiment_container.py
```diff
@@ -218,6 +219,10 @@ def new_experiment(name: str,
...
+    if len(exp_ids) >= 1:
+        log.warning(f"There is (are) already experiment(s) with the name of {name} "
+                    f"and sample name of {sample_name} in the database.")
```
jenshnielsen, line 225: If we pass `sample_name=None` to the constructor of experiment it will use `some_sample` as the name so we really should check for that above ?

FarBo: Something like adding this to the function:
`sample_name = sample_name or "some_sample"`?

jenshnielsen: Yes that should do the trick

PR #3327, opened 2021-09-01
Logging: explicitly set encoding for file handler

jenshnielsen:
This prevents issues where the broken pipe character (used for delimiting messages) cannot be written to the log file if the file has a different encoding. Reading the documentation it is not clear to which extend this may also be an issue for the stream handler used for the console logging. This writes to stderr. According to https://docs.python.org/3/library/sys.html#sys.stderr the encoding of stderr should be defined by local on oses other than windows so this probably works without issues on windows (but there are some limitations as noted in the stderr documentation) On most other platforms I would expect local encoding to be utf8 anyway.

### DIFF
--- a/qcodes/logger/logger.py
+++ b/qcodes/logger/logger.py
@@ -283,8 +283,9 @@ def start_logger() -> None:
     filename = get_log_file_name()
     os.makedirs(os.path.dirname(filename), exist_ok=True)
 
-    file_handler = logging.handlers.TimedRotatingFileHandler(filename,
-                                                             when='midnight')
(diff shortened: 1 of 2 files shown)


FarBo:
This seems the way to go to fix the broken pipe encoding.
In a quick search, older Linux server encoding is something else. Newer versions are adapted to utf-8. The rest should be utf-8.

jenshnielsen:
Looks like this fixed the reported issue. @astafan8 could you review

astafan8 (approved):

PR #3339, opened 2021-09-08
Keithley 2600: support returning both voltage and current compliance limit errors

astafan8:
otherwise when in the SMU is used in current sourcing mode and the voltage hits the voltage compliance limit, the measurement_status property of the current/voltage parameters did not reflect that.


jenshnielsen (approved):

qcodes/instrument_drivers/tektronix/Keithley_2600_channels.py
```diff
@@ -204,10 +204,20 @@ class MeasurementStatus(StrEnum):
     """
     Keeps track of measurement status.
     """
-    COMPLIANCE_ERROR = 'Reached compliance limit.'
+    CURRENT_COMPLIANCE_ERROR = 'Reached current compliance limit.'
```
astafan8, line 207: need to leave the old enum item for backwards compatibility

astafan8:
tested on the instrument and confirmed that it works.

PR #3342, opened 2021-09-09
Default devcontainer for github codespaces

astafan8:
todo:
- [ ] make postCreateCommand work/execute, so that all pip requirements are installed automatically
- [ ] figure out why ``pip install -e .`` results in a ``can't create or remove files in install directory`` error from pip


jenshnielsen:
>  figure out why pip install -e . results in a can't create or remove files in install directory error from pip, specically

Why not just move it into the dockerfile ?

astafan8:
> Why not just move it into the dockerfile ?

don't know, honestly. I think because it then becomes "part of the container" so that for every new branch one'd need either a new container or call "pip install -e ." anyway. I know too little about containers to give a good answer))

PR #3362, opened 2021-09-15
Reset measurement status of curr and volt parameters of Keithley 2600

astafan8:
Otherwise, after setting voltage/current to a new value, the statuses won't be reset unless ``.get`` is called.


trevormorgan (approved):
Looks good @astafan8 , I added a small update to test that the parameter status is None when set is called

PR #3390, opened 2021-09-25
Add option for n-based indexing of ChannelList

yoshi74ls181:
Channel numbers of some instruments start with one.
This pull request allows us to be consistent with the labels on the instrument when indexing and slicing a `ChannelList`.

- Add an optional argument `index_origin` to `ChannelList.__init__`
- Modify `ChannelList.__getitem__`, `index`, and `insert` to respect the index origin

 @jenshnielsen


qcodes/instrument/channel.py
```diff
@@ -156,6 +156,9 @@ class ChannelList(Metadatable):
             method of :class:`ChannelList`.
             Should be a subclass of :class:`MultiChannelInstrumentParameter`.
 
+        index_origin: Origin for indexing and slicing.
+            Useful if you want the channel number to start from one.
```
astafan8, line 160:
```suggestion
            Useful if you want the channel indexing to start from a
            number other than ``0`` (e.g. ``1``).
```

qcodes/instrument/channel.py
```diff
@@ -188,6 +192,7 @@ def __init__(self, parent: InstrumentBase,
         self._chan_type = chan_type
         self._snapshotable = snapshotable
         self._paramclass = multichan_paramclass
+        self.index_origin = index_origin
```
astafan8, line 195: shall this attribute be privte so that it's not user-adjustible after init?
```suggestion
        self._index_origin = index_origin
```

astafan8 (commented):
thank you @yoshi74ls181 ! would you be willing to add a few tests for this feature?

qcodes/instrument/channel.py
yoshi74ls181: Right, it doesn't make sense to change the index origin after the initialization.

yoshi74ls181:
@astafan8 Right, I will work on writing a few tests at some point this week.
Thank you for your efforts in maintaining high coding standards!

yoshi74ls181:
I added a few tests of one-based indexing (index_origin = 1) of ChannelList, as suggested by @astafan8.

qcodes/instrument/channel.py
```diff
@@ -208,25 +214,45 @@ def __init__(self, parent: InstrumentBase,
...
+        Args:
+            i: int or None
+
+        Raises:
+            IndexError: If 0 <= i < index_origin
+        """
+        if i is None or i < 0:
+            return i
+        elif 0 <= i < self._index_origin:
+            raise IndexError("Index is smaller than the index_origin")
```
astafan8, line 230:
```suggestion
            raise IndexError(f"Index {i} is smaller than the index_origin {self._index_origin}")
```

qcodes/instrument/channel.py
```diff
@@ -208,25 +214,45 @@ def __init__(self, parent: InstrumentBase,
...
+        Shifts an index by -index_origin unless it is None or negative.
+
+        Args:
+            i: int or None
+
+        Raises:
+            IndexError: If 0 <= i < index_origin
+        """
+        if i is None or i < 0:
+            return i
```
astafan8, line 228: So, per my understanding, the negative indexing (where ``-1`` means last, ``-2`` means 2nd to last, etc) should still "just" work regardless of the index origin, right? That's fine, BUT - why would one ever try to index with ``None``? I'd suggest that it should be an error; in all the methods where the _shift_index is used the argument ``i`` is always an ``int`` and never ``None`` (unless i missed something?). Hence I suggest the following implementation, and let's change the signature from ``_shift_index(self, i: Optional[int]) -> Optional[int]`` to ``_shift_index(self, i: int) -> int``:
```suggestion
        if i is None:
            raise ValueError("Cannot index with None")
        if i < 0:
            return i
```

astafan8 (commented):
Thanks so much for the tests! if they were parametrized against the index origin, that'd be even better, but this is already good.

qcodes/instrument/channel.py
yoshi74ls181: `_shift_index(i)` is called with `i = None` for example at line 246 when a ChannelList is sliced like `channel_list[:3]`, because in this case `channel_list.__getitem__(slice(None, 3, None))` is actually called.
A `None` means `0` when used as the start value and means `len(channel_list)` when used as the end value.
You are right that `i = None` should raise an error when indexing with an int or a tuple.
In the current version, the error is raised when indexing `self._channels`, after `_shift_index(i)` is called.
Do you think it is better to raise the error earlier?

astafan8: > A None means 0 when used as the start value and means len(channel_list) when used as the end value.

aa, i see, thanks for explaining!

> In the current version, the error is raised when indexing self._channels, after _shift_index(i) is called. Do you think it is better to raise the error earlier?

that's probably OK. But it puts more pressure on the implementation of _shift_index to propagate the ``None`` value through and don't try to be smart :) 

Let's do the following - let's keep the current implementation AND add a comment in the code of _shift_index explaining why ``None`` value is just being passed thorugh. Would you agree?

yoshi74ls181: I agree, the intention was not at all obvious. Give me a few minutes to come up with a good comment...

yoshi74ls181:
I added a comment on the pass-through of `None` in `ChannelList._shift_index`, as suggested by @astafan8.

jenshnielsen:
I am not sure how to address this best but `__add__` will currently generate a new channel list by combining two channel lists. Here the offset is not retained.

yoshi74ls181:
You're right.
Perhaps we can ditch this pull request and instead allow passing a string to `ChannelList.__getitem__`?
This way we can name the channels `'1'`, `'2'`, ..., and index them like `channel_list['1']`.
We would lose the slicing though, which we use much more often than adding `ChannelList`s.

jenshnielsen:
I don't think thats the way to go. This is explicitly called a list so it would be surprising if you had to look up by name (which would be a dict or a mapping) 

I think add is lightly used so I would probably initially just play it safe and raise an error if 2 channel lists with different offsets are added and pass the common offset to the new channellist

astafan8:
> __add__ will currently generate a new channel list by combining two channel lists. Here the offset is not retained.

that's good one!

> Perhaps we can ditch this pull request and instead 

i'd say that passing strings does not look pythonic, and is not intutive. if channels have names that are "1", "2", etc, then the ``name`` of the channel can be used to have those convenient numbers and address channels by their names, e.g. ``instrument.ch1`` for ``instrument.channels[0]``. this does not help with slicing, true, but at least it does not bring issues with ``__add__`` and maintaining index_origin, and so forth. If in some cases, the slicing is so useful because it gives easy access to a few channels of interest, then perhaps that convenience should be either part of the driver itself or the code that uses the driver, since that code likely knows why it's useful to "slice those particular channels" - would you agree?

So I'm tempted now to close this PR and leave things as they are now. What do you think?

yoshi74ls181:
> I think add is lightly used so I would probably initially just play it safe and raise an error if 2 channel lists with different offsets are added and pass the common offset to the new channellist

Thank you @jenshnielsen, I like your suggestion.

> If in some cases, the slicing is so useful because it gives easy access to a few channels of interest, then perhaps that convenience should be either part of the driver itself or the code that uses the driver, since that code likely knows why it's useful to "slice those particular channels" - would you agree?

Thank you @astafan8, this is also a very good point.
The discussion here made me realize that maybe all I need is something like
```
cg = instrument.channel_group('ch1', 'ch2', 'ch3')
cg.param1(val1)
cg.param2(val2)
...
```

Let me leave the final decision on whether to close this PR to the maintainers of this project.
I don't want to go against the coding philosophies of this project. (which, by the way, I think is the best among similar projects!)

jenshnielsen:
thanks @yoshi74ls181 to me the challenge is that we want to let this model one of the build in python container types. 
If we try to both model list and dict we are likely to get into inconsistencies 
Currently we are modelling the list/tuple with index. 


It seems to me that perhaps the most clean way to get what you want is a method on the channel list. Something like `get_channels_by_name` would perhaps work? e.g. something like this (not tested)

```
def self.get_chanenls_by_name(*channel_names: str) -> ChannelList:
     selected_channels = [self._channel_mapping[name] for name in channel_names]

      new_channel_list = ChannelList(self._parent, self._name, self._chan_type, selected_channels )
     return new_channel_list 
```

yoshi74ls181:
Thank you @jenshnielsen, I like your idea of `get_channels_by_name`.
I'll close this pull request and open a new one with this new method and a test.

PR #3397, opened 2021-09-28
Build changelog with Towncrier

jenshnielsen:
The idea is that rather than generating a changelog at each release the contributor can write a fragment (as a fine) that automatically
gets included in the changelog. This gives the best of both worlds. Avoid conflicts in the changelog and allows the pr writer to create the changelog entry

- [ ] Change section types to match qcodes sections (drivers, breaking ...)
- [ ] Add some examples and remove outdated example
- [ ] Update contributing and pr guide for this

Optional can be done later in a new pr

- [ ] Use sphinx plugin for generating changlog for prerelease (I think I saw an example of this in the setuptools changelog)


jenshnielsen:
Screenshot 2021-09-28 110956

Is an example of using this from the https://github.com/jenshnielsen/qcodes/tree/towncrier_demo branch

jenshnielsen:
replace `write the changelog` in the wiki with:

1. Generate the changelog

We are generating the changelog using Towncrier from the newsfragment submitted with each pull request.

To generate the changelog for a new release do:

```
* pip install towncrier
* create a new git branch for the changelog.
* towncrier build --version whateverthenewversionis
* modify `docs\changes\index.rst` to include the new changelog.
* Build the docs locally and make sure that everything looks good.
* OPEN a PR against master. Review it, fix, etc.

You may want to check that all important pull requests are included in the changelog. 
The command git log <last release>..HEAD is useful to print all changes since last release. You can add --merges --first-parent to the git log command to only show merge commits. --reverse is helpful too. That is: git log <last release>..HEAD --merges --first-parent --reverse
Another approach is to get a list of merged pull-request since the last release: find the merged pull-request of the previous release, note the date when that happened, and then paste the following into the "Filter" field in the "Pull requests" tab on GitHub: is:pr is:closed merged:>YYYY-MM-DD where YYYY-MM-DD should be replaced by the date of the merge of the pull-request of the previous release.
```

PR #3399, opened 2021-09-28
extract exp_id to attrs function

jenshnielsen:
Lifted from #3094

Since this is a relatively straightforward refactor it makes sense to apply it outside that pr


astafan8 (approved):
sweet new function :)

PR #3410, opened 2021-09-30
move _create_exp_if_needed to exp module where it belongs

jenshnielsen:
This reveals that this function is very similar to the `load_or_create_experiment` function. Perhaps we should unify them?

### DIFF
--- a/qcodes/dataset/database_extract_runs.py
+++ b/qcodes/dataset/database_extract_runs.py
@@ -1,11 +1,11 @@
 import os
-from typing import Optional, Union
 from warnings import warn
 
 import numpy as np
 
 from qcodes.dataset.data_set import DataSet
 from qcodes.dataset.descriptions.versioning.converters import new_to_old
+from qcodes.dataset.experiment_container import _create_exp_if_needed
 from qcodes.dataset.linked_datasets.links import links_to_str
 from qcodes.dataset.sqlite.connection import ConnectionPlus, atomic
 from qcodes.dataset.sqlite.database import (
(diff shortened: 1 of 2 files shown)


qcodes/dataset/database_extract_runs.py
```diff
@@ -116,43 +114,6 @@ def extract_runs_into_db(source_db_path: str,
         target_conn.close()
 
 
-def _create_exp_if_needed(target_conn: ConnectionPlus,
```
astafan8, line 119: hmm, i think the reason this function is here is becasue it's private wrt the logic of extracting runs to another db. But i agree that it belongs to the experiment module, and it can be used to create experiments in the DB **with all their attributes provided** (starttime/endtime/fmtstr/etc), but not by experimentalists/researchers - they should still use the function that only takes exp_name and sample_name.

astafan8 (approved):

qcodes/dataset/database_extract_runs.py
jenshnielsen: Yes no one should use this method outside qcodes which is why it is still private. I wish python had stronger conversions about module/class/submodule private but perhaps we should just agree on our own? The intention with making this private is that is should not be used outside qcodes. The issue with leaving this in database_extract_run is that it very easily results in code duplication e.g. other modules that need to do something similar will implement the same functionality. 

I would prefer to get rid of it completely. I am not even sure comparing against these attributes even always makes sense. E.g. format_string is probably very rarely changed in practice. End_time could easily change between two different db files for the same experiment (if after moving the first dataset one db file has the experiment marked completed exporting a new dataset from and to the same db files will duplicate the experiment in the second one )

PR #3418, opened 2021-10-04
Extract nested helper functions from dond's to module level

FarBo:
This is a small PR to relocate two nested functions from dond to the global scope.
The functions are still private functions, we can discuss if some of the doNd private functions can be in the public API, for example _parse_dond_arguments, then we can think about a better name.

@jenshnielsen 
@astafan8


astafan8 (approved):
this is definitely a good first step. i'd only suggest to give the PR a different title, e.g. "extract nested helper functions from dond's to module level" :) the word "release" has a special meaning, although it's correct, we do "release" these functions from the chains of function scope ;)

PR #3426, opened 2021-10-05
allow open cencus 0.8

jenshnielsen:
Seems like dependabot did not do this correctly in #3423


jenshnielsen:
This seems to fails since the open census azure ext conda package does not correctly depend on the azure core and azure identity package. The open census azure package is here https://github.com/conda-forge/opencensus-ext-azure-feedstock/  and the core package here. The core package is somewhat out of date also

astafan8 (approved):

PR #3434, opened 2021-10-06
add query to get run info from db

jenshnielsen:
Makes it easier to construct a dataset without having to do multiple function calls. To be using in #3094  for loading from the db

Missing:

- [x] Basic tests
- [x] Should there be a version of this that deserializes e.g. snapshot str -> JSON Dict and where should that live?


astafan8:
> Should there be a version of this that deserializes e.g. snapshot str -> JSON Dict and where should that live?

could be useful. i think that it can call the function that returns raw attributes, and the processes them, and returns the deserilized stuff. and it can live on the qcodes.dataset.* module level, that is, not in sqlite.*

PR #3435, opened 2021-10-06
Fix Tektronix AWG70000 naming

aadiyatul:
Adds support for both AWG7000xA and AWG7000xB versions. (The communication protocols for both are [the same](https://www.tek.com/support/faqs/what-are-differences-between-awg70000a-and-awg70000b).)
Tested on a real AWG70002B.
Module and class names not changed to keep the existing API.

@jenshnielsen


jenshnielsen:
Ok, In the future you can continue pushing more commits to the same branch you do not need to open a new pull request

PR #3441, opened 2021-10-08
run test + docs build with python 3.10

jenshnielsen:
Changes

- [x] some more types seems to have been added so assert that extra is not None to fix this typechecking issue
- [x] Output of an error message has changed. Work around a test to fix this
- [ ] Once merged add this to bors config and github required checks

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -18,12 +18,14 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, 3.10]
         exclude:
           - os: windows-latest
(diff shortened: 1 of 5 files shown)


jenshnielsen:
Current status is there are no wheels for h5py. Building h5py from source results in an error. 

It seems that building h5py from source is currently setup to use numpy 1.19.3 for which there are no python 3.10 wheels. It therefore tries to build numpy 1.19.3 from source. This fails with a `numpy/core/src/multiarray/scalartypes.c.src:2967:27: error: incompatible type for argument 1 of ‘_Py_HashDouble’` error. This is likely due to that old version of numpy using a deprecated Python C api. At this stage there does not seem to be much we can do apart from wait for h5py to support 3.10

jenshnielsen:
@astafan8 this should be ready for review

astafan8 (approved):

PR #3534, opened 2021-11-01
Improved dond to not set setpoint parameters with values that are already set

FarBo:
This PR fixes the unnecessary wait for already set parameters in the dond function.

The logic is to store delays of sweep objects in a dictionary and call them when the setpoint value of a parameter changes, otherwise set post_delay of that parameter to zero.
One might argue we should not set a parameter that is already set, then we shouldn't have this unnecessary wait time in the first place. Of course, we can discuss about which way is more appropriate, but the current PR is a straightforward fix that uses the same already implemented logic for post_actions.

@jenshnielsen 
@astafan8


astafan8:
> One might argue we should not set a parameter that is already set, then we wound't have this unnecessary wait time in the first place.

@FarBo i'm actually tending to agree with you on this point -  that we probably should simply not call ``setpoint_param(new_setpoint)`` if ``new_setpoint == old_setpoint``. I think that this is in line with the kind of measurements do*d attempt to support, those that have nested for loops:
```python
for x in x_vals:
    x_param.set(x)
    for y in y_vals:
        y_param.set(y)
        ...
```
where once the value of e.g. x is set, it is not being set more than once due to the structure of the code.

@jenshnielsen do you agree? perhaps I (we) are missing something?

FarBo:
@astafan8 
The loops structure of do2d allows this to happen easily, but not in dond.
We already have that change of setpoint check for post_actions, so I wanted to utilize it for post_delay as well in this PR.
My concerning point in this PR is for the actual instrument parameters that I removed the delay time for them if they are already set. So if parameter p is already set for value x, then we set that again for the same value, technically, we should not be worried having zero post_delay, right?
The second concern is the unnecessary time for setting the parameter again (It requires to communicate again with the instrument for setting it).

If the two concern above are something that we need to be careful of, then going in the direction of not setting the already set parameter with the same value is the way. Of course, it needs a bit of thinking and refactoring, I guess.

If everyone agrees, I can work on that

jenshnielsen:
@FarBo Perhaps we could wrap parameter.set in a small piece of logic that did something like 

```
def conditional_set_for_some_parameter(value)
    if not  value == parameter.cache.get():
        parameter.set(value)
```

That could be generated via functools partial wrapper on a function like


```
def conditional_set(parameter, value):
    if not  value == parameter.cache.get():
        parameter.set()


dmmsetterparameter = partial(conditional_set, dmm.voltage)

```

FarBo:
@jenshnielsen 
Sounds like a good idea. Just to clarify, are we talking to have this wrapper function for reading cache and setting the parameter in the doNd module or somewhere else?

jenshnielsen:
I think we should just put it privately in the dond module for now and have the dond function create the wrappers

qcodes/utils/dataset/doNd.py
```diff
@@ -758,7 +758,8 @@ def dond(
                 param_set_list = []
                 param_value_action = zip(params_set, setpoints, active_actions)
                 for setpoint_param, setpoint, action in param_value_action:
-                    setpoint_param(setpoint)
+                    _conditional_parameter_set(setpoint_param, setpoint)
+                    print(setpoint_param())
```
jenshnielsen, line 762: remove the print?

jenshnielsen:
Looks good. Please remove the print and add a changelog entry

qcodes/utils/dataset/doNd.py
FarBo: Dam...

jenshnielsen (approved):

PR #3541, opened 2021-11-02
add plot to top level api

jenshnielsen:
Mostly trying to debug why https://github.com/QCoDeS/Qcodes/pull/3536 fails


jenshnielsen:
This was mostly experimental

PR #3543, opened 2021-11-03
Dond: Only set parameter if not already at setpoint.

ThorvaldLarsen:
Current implementation of `dond` will ask _every_ instrument to set the setpoint at _every_ measurement point. This leads to an excessive amount of instrument communication as well as adding up all `post_delay`s at every measurement point.
Especially for magnet parameters this becomes a huge issue as setting the parameter even to the current value can take several seconds due to internal PID loops in the magnet power supply.

Change behaviour to only set the parameter if it isn't already at the desired value as known by qcodes. This ensures we inly ping instruments that needs to get the setpoint updated during the run.


ThorvaldLarsen:
closed in favor of #3534 which is functionally the same change.

PR #3555, opened 2021-11-05
Fix power and frequency ranges of Keysight N51x1 and N51x3 signal generators

yoshi74ls181:
This PR improves the drivers for Keysight N51x1 and N51x3 signal generators by:

- Adding explicit support for N5173B
- Fixing the power range of N5183B
- Fixing the frequency range options for N51x1

@jenshnielsen

### DIFF
--- a/qcodes/instrument_drivers/Keysight/Keysight_N5173B.py
+++ b/qcodes/instrument_drivers/Keysight/Keysight_N5173B.py
@@ -1,0 +1,4 @@
+
+from qcodes.instrument_drivers.Keysight.Keysight_N5183B import N5183B as N5173B
+# N5173B has the same interface as N5183B.
+# This file is to allow for improved device labeling.
--- a/qcodes/instrument_drivers/Keysight/Keysight_N5173B.py
+++ b/qcodes/instrument_drivers/Keysight/Keysight_N5173B.py
@@ -1,4 +1,5 @@
-from qcodes.instrument_drivers.Keysight.Keysight_N5183B import N5183B as N5173B
+from qcodes.instrument_drivers.Keysight.Keysight_N5183B import N5183B
 
-# N5173B has the same interface as N5183B.
-# This file is to allow for improved device labeling.
+
+class N5173B(N5183B):
+    pass  # N5173B has the same interface as N5183B
--- a/qcodes/instrument_drivers/Keysight/Keysight_N5183B.py
+++ b/qcodes/instrument_drivers/Keysight/Keysight_N5183B.py
@@ -1,6 +1,8 @@
+from typing import Any
+
 from qcodes.instrument_drivers.Keysight.N51x1 import N51x1
 
 
 class N5183B(N51x1):
-    def __init__(self, name: str, address: str, **kwargs):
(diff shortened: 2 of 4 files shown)


qcodes/instrument_drivers/Keysight/Keysight_N5173B.py
```diff
@@ -0,0 +1,4 @@
+from qcodes.instrument_drivers.Keysight.Keysight_N5183B import N5183B as N5173B
```
astafan8, line 1: this is good, but because it's really a different instrument i'd still suggest to implement it as a class:
```python
class N5173B(N5183B):
    pass
```
otherwise the 173 instances would actually be of class N5183B which i don't think is correct, even though the intention here is that the classes are the same

astafan8 (approved):

qcodes/instrument_drivers/Keysight/Keysight_N5173B.py
yoshi74ls181: That is a very good point, thank you!

qcodes/instrument_drivers/Keysight/Keysight_N5183B.py
```diff
@@ -1,3 +1,6 @@
-from qcodes.instrument_drivers.Keysight.N51x1 import N51x1 as N5183B
-# N5183B has the same interface as N51x1 devices
-# This file is to allow for improved device labeling.
+from qcodes.instrument_drivers.Keysight.N51x1 import N51x1
+
+
+class N5183B(N51x1):
+    def __init__(self, name: str, address: str, **kwargs):
```
jenshnielsen, line 5:
```suggestion
    def __init__(self, name: str, address: str, **kwargs: Any):
```

qcodes/instrument_drivers/Keysight/Keysight_N5183B.py
```diff
@@ -1,3 +1,6 @@
-from qcodes.instrument_drivers.Keysight.N51x1 import N51x1 as N5183B
-# N5183B has the same interface as N51x1 devices
-# This file is to allow for improved device labeling.
+from qcodes.instrument_drivers.Keysight.N51x1 import N51x1
```
jenshnielsen, line 1:
```suggestion
from typing import Any

from qcodes.instrument_drivers.Keysight.N51x1 import N51x1
```

qcodes/instrument_drivers/Keysight/N51x1.py
```diff
@@ -7,10 +7,10 @@
 class N51x1(VisaInstrument):
     """
     This is the qcodes driver for Keysight/Agilent scalar RF sources.
-    It has been tested with N5171B, N5181A, N5171B, N5183B
+    It has been tested with N5171B, N5181A, N5173B, N5183B
     """
 
-    def __init__(self, name: str, address: str, **kwargs: Any):
+    def __init__(self, name: str, address: str, min_power=-144, max_power=19, **kwargs: Any):
```
jenshnielsen, line 13:
```suggestion
    def __init__(self, name: str, address: str, min_power: int = -144, max_power: int = 19, **kwargs: Any):
```

Issue #3564, opened 2021-11-08
Unintentional slowdowns in dond when using parameters with a ramprate set with step and post_delay


JouriB:
When using a Parameter which has a ramp rate set through a combination of step and post_delay (or inter_delay) as a setpoint parameter in a doNd measurement, waiting overheads can increase dramatically. This is the result of the post_delay being set to the delay time provided as an argument to the doNd measurement functions. Slowdowns are most severe when when the stepsize in a doNd measurement loop exceeds the `step` property of the Parameter. In this case the delay time is applied multiple times for each setpoint.

If the step size in a doNd measurement loop is smaller than the `step` property of the Parameter, slowdowns will only occur when sweeping the Parameter over larger ranges, e.g. when the ramped parameter is used in an inner loop and is swept from the final point to the initial point of the inner loop.

Relevant to all doNd functions that (temporarily) overwrite the `post_delay`.

### Steps to reproduce
1. Run and time e.g. do1d for a Parameter without the `step` property set.
2. Adjust the `step` property to a value e.g. five times smaller than the step size in the do1d loop.
3. Run and time do1d again, measurement time should increase ~5x.

### Expected behaviour
The ramped parameter keeps its intended ramping behavior when it's changed and waiting time for delays in the measurement are only applied when a setpoint is reached.

### System
**qcodes version 0.29.1**

FarBo:
Thanks @JouriB for the report

We will improve doNd functions to deal with this issue.

FarBo:
This issue is addressed in the referenced PR and closed.

PR #3576, opened 2021-11-12
Use sleep instead of post_delay in doNd

FarBo:
This PR is created to change the delay time between setpoints in doNd functions. This is done because the current post_delay logic can unintentionally prolong the delay time if a ramped parameter is used, which the ramp time is shorter than the delay time provided in the doNd functions.

- [x] do1d
- [x] do2d
- [x] dond
- [x] test files if needed

The PR is addressing this issue:
#3564


qcodes/utils/dataset/doNd.py
```diff
@@ -286,8 +285,7 @@ def do1d(
             datasaver.add_result(
                 (param_set, set_point), *call_param_meas(), *additional_setpoints_data
             )
-
-    param_set.post_delay = original_delay
+            sleep(delay)
```
jenshnielsen, line 288: One potential optimization that we could do is to get the time before and after calling add_result and only sleep the delay - the time needed to call add_result that way we would save a bit of time if add_result is slow

FarBo: That's a good point. We can substract add_result time from the delay and sleep with the remaining time.

qcodes/utils/dataset/doNd.py
```diff
@@ -283,11 +282,13 @@ def do1d(
 
         for set_point in tqdm(setpoints, disable=not show_progress):
             param_set.set(set_point)
+            t0 = time.time()
             datasaver.add_result(
                 (param_set, set_point), *call_param_meas(), *additional_setpoints_data
             )
-
-    param_set.post_delay = original_delay
+            datasaver_time = abs(time.time() - t0)
```
jenshnielsen, line 289: I suggest using time.perf_counter rather than time.time since that has a better resolution https://docs.python.org/3/library/time.html#time.perf_counter

qcodes/utils/dataset/doNd.py
```diff
@@ -283,11 +282,14 @@ def do1d(
 
         for set_point in tqdm(setpoints, disable=not show_progress):
             param_set.set(set_point)
+            t0 = time.perf_counter()
             datasaver.add_result(
                 (param_set, set_point), *call_param_meas(), *additional_setpoints_data
             )
-
-    param_set.post_delay = original_delay
+            datasaver_time = abs(time.perf_counter() - t0)
```
jenshnielsen, line 289: I think max would be more correct than abs 
```suggestion
            datasaver_time = max((time.perf_counter() - t0), 0)
```

In that way if datasaving takes longer than the sleep time we will not sleep at all. With the current code we would sleep the delta between the two again

FarBo: In the following if statement (please pretend you are not seeing the print below :) ), I only sleep the delta value if the provided delay is greater than datasaver_time, otherwise there is no sleep if datasaver_time is bigger than the delay.

jenshnielsen: It's to a certain extend a matter of preference but to me 

```python
datasaver_time = max((time.perf_counter() - t0), 0)
sleep(datasaver_time)
```
is cleaner than the if statement. I don't think there is any signicifant performance difference

FarBo: Agree. This is cleaner.  I will modify accordingly. Thanks

FarBo:
@jenshnielsen 
As we are replacing post_delay with time.sleep, shouldn't we sleep after setting the setpoint to the value, rather than sleeping after the datasaver?
I think users put delay because they want desired values on instruments to be stabilized before entering to the datasaver part, which our post_delay attribute intended to handle this.

jenshnielsen:
@FarBo Yes the sleep should be between set of setpoints and getting the parameter

jenshnielsen:
@FarBo I think this is close to being ready. Could you add a changelog entry?

astafan8 (approved):

PR #3586, opened 2021-11-15
Make export configurable

jenshnielsen:
This makes it possible to configure the exported name. So far it is possible to include captured_run_id and guid. 

- [x] Should we include other elements now or wait
- [x] Should we change the default to include guid?

This is technically an api break since it now uses captured_run_id and not run_id but I do think that is for the better.

### DIFF
--- a/qcodes/configuration/qcodesrc_schema.json
+++ b/qcodes/configuration/qcodesrc_schema.json
@@ -350,7 +350,7 @@
                         "type": "string",
                         "enum": ["captured_run_id", "guid", "exp_name", "sample_name", "name"]
                     },
-                    "default": ["captured_run_id"]
+                    "default": ["captured_run_id", "guid"]
                 }
             },
             "description": "Settings related to the DataSet and Measurement Context manager",
--- a/qcodes/configuration/qcodesrc_schema.json
+++ b/qcodes/configuration/qcodesrc_schema.json
@@ -344,7 +344,7 @@
                     "description": "Prefix to add to filename for exporting datasets to disk after a measurement finishes. Only activated if export_type is set"
                 },
                 "export_name_elements": {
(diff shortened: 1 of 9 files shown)


astafan8:
>  Should we include other elements now or wait

i'd vote for including :) at least smaple name, exp name, ds name, started ts .

> Should we change the default to include guid?

definitely!

> This is technically an api break since it now uses captured_run_id and not run_id but I do think that is for the better.

agree! we could consider this a bug fix.

qcodes/configuration/qcodesrc_schema.json
```diff
@@ -342,6 +342,15 @@
                     "type": "string",
                     "default": "qcodes_",
                     "description": "Prefix to add to filename for exporting datasets to disk after a measurement finishes. Only activated if export_type is set"
+                },
+                "export_name_elements": {
+                    "description": "elements to use in export name. Names will be formated like '{export_prefix}_{export_name_elements[0]}_{...}.file_extension}",
```
astafan8, line 347: what about allowing folks to provide a format string that use the "elements"? the default would be ``{prefix}_{captured_run_id}_{guid}`` (or so) and the format string give full flexibility for users and reliefs us from maintaining the conversion from "elements" to "post_fix" :) what do you think @jenshnielsen ?

jenshnielsen: I am not really sure. It seems like one more thing that users have to get right and if they don't exporting will throw an exception. I would thing that being able to control the order of the elements should be enough

astafan8: ok, thats fair. then let's go with the elements for now :)

jenshnielsen:
bors merge

jenshnielsen:
need to write a changelog entry

jenshnielsen:
bors merge

jenshnielsen:
bors merge

jenshnielsen:
bors merge

PR #3603, opened 2021-11-19
Fix random test failure

jenshnielsen:
the field may not ramp exactly to 0
since this is implemented as fp math

see an example of this here https://github.com/QCoDeS/Qcodes/runs/4260183190?check_suite_focus=true


astafan8 (approved):

jenshnielsen:
bors merge

PR #3623, opened 2021-11-25
Add two more methods to dataset protocol

jenshnielsen:
Research shows that these are fairly commonly uses so to make it easier to use the protocol I suggest adding them here. The implementation is also fairly straight forward


astafan8 (approved):
agree

PR #3648, opened 2021-12-03
remove incorrect news fragment

jenshnielsen:
This fragment should have been included in 0.30.0 but was not since it was incorrectly formatted (the file name should be underthehood not Underthehood)


jenshnielsen:
I don't think this is particular important so I suggest just removing it rather than retrofitting into 0.30.0 changelog

astafan8 (approved):

jenshnielsen:
bors merge

PR #3650, opened 2021-12-04
AMI430: Fix switch heater parameters that return 0 or 1 as a string

peendebak:
The parameters from the AMI430 switch heater return a string such as `0\n`. This was parsed by stripping and then casting to bool. But `bool('0')` is True, which is not what the parameter should return.


qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -123,7 +124,7 @@ def off(self) -> None:
 
     @_Decorators.check_enabled
     def check_state(self) -> bool:
-        return bool(self.ask("PS?").strip())
+        return bool(ast.literal_eval(self.ask("PS?").strip()) )
```
astafan8, line 127: but if we know that PS command returns an int, why don't we cast it to int directly making the code simpler?
```suggestion
        return bool(int(self.ask("PS?").strip()))
```

astafan8: @peendebak could you respond? :)

peendebak: @astafan8 Good suggestion. Someone else from my team is looking at more places in the driver where this can occur.

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -107,7 +108,7 @@ def enable(self) -> None:
         self._enabled = True
 
     def check_enabled(self) -> bool:
-        return bool(self.ask('PS:INST?').strip())
+        return bool(ast.literal_eval(self.ask('PS:INST?').strip()))
```
peendebak, line 111:
```suggestion
        return bool(int(self.ask('PS:INST?').strip()))
```

qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -1,6 +1,7 @@
 import collections
 import logging
 import numbers
+import ast
```
peendebak, line 4:
```suggestion
```

jenshnielsen:
@peendebak I pushed a basic test and added support for enabling disabling the switch heater in the yaml simulator. I think this should be ready to merge now?

astafan8 (commented):
bors merge

astafan8 (approved):
bors merge

astafan8:
bors merge

PR #3662, opened 2021-12-10
Allow getting an uninitialized DelegateParameter with offset or scale

### LINKED ISSUE #3653: Getting a DelegateParameter with scale fails if it is not initialized
If you create a DelegateParameter and pass a scale argument, it is not possible to get its cache before it is initialized. The scale gets applied to the None that resides in the source parameter and gives a TypeError.

### Steps to reproduce
```python
from qcodes import Parameter, DelegateParameter

p = Parameter("p")
dp = DelegateParameter("dp", p, scale=1.1)

print(dp.source.cache.get(False))  # gives back None
print(dp.cache.get(False))  # fails
```

### Expected behaviour
I expected to get None back, just as you would a plain uninitialized Parameter.

A workaround is to give nan as an initial cache value. This works:
```python
from qcodes import Parameter, DelegateParameter
import numpy as np

p = Parameter("p")
dp = DelegateParameter("dp", p, scale=1.1, initial_cache_value=np.nan)

print(dp.source.cache.get(False))
print(dp.cache.get(False))
```

### Actual behaviour
I got a TypeError:
```python

  File "D:\users\hw_sw_integration\projects\low-level-software\src\lls_input\amber\scripts\example_02.py", line 14, in <module>
    print(dp.cache.get(False))

  File "D:\users\hw_sw_integration\venv-3.9\lib\site-packages\qcodes\instrument\parameter.py", line 1541, in get
    return self._parameter._from_raw_value_to_value(

  File "D:\users\hw_sw_integration\venv-3.9\lib\site-packages\qcodes\instrument\parameter.py", line 601, in _from_raw_value_to_value
    value = value / self.scale

TypeError: unsupported operand type(s) for /: 'NoneType' and 'float'
```

It appears that `_from_raw_value_to_raw_value` applies the scale on the value even if it's None. A simple solution would be to put a big `if value is not None:` in the method, but this may not cover all the use cases (e.g. more intricate ways of handling the scale than dividing/multiplying).

### System
Windows 10
qcodes 0.30.1

haroldmeerwaldt:
- [x] do not apply offset or scale when getting delegate parameter and source is None (in `_from_raw_value_to_value`)
- [x] added two tests

Fixes #3653

### DIFF
--- a/qcodes/tests/parameter/test_parameter_scale_offset.py
+++ b/qcodes/tests/parameter/test_parameter_scale_offset.py
@@ -205,25 +205,3 @@ def test_setting_numpy_array_valued_param_if_scale_and_offset_are_not_none():
     param(values)
 
     assert isinstance(param.raw_value, np.ndarray)
-
-
-def test_getting_uninitialized_delegate_parameter_with_offset_returns_none():
-    param = Parameter(name='test_param',
-                      set_cmd=None,
-                      get_cmd=None)
-
-    d_with_offset = DelegateParameter(name='deleg_param_with_offset',
-                                      source=param, offset=2)
-
-    assert d_with_offset.get() is None
-
-
-def test_getting_uninitialized_delegate_parameter_with_scale_returns_none():
-    param = Parameter(name='test_param',
-                      set_cmd=None,
-                      get_cmd=None)
-
-    d_with_scale = DelegateParameter(name='deleg_param_with_scale',
-                                     source=param, scale=2)
-
-    assert d_with_scale.get() is None
--- a/qcodes/tests/parameter/test_parameter_scale_offset.py
+++ b/qcodes/tests/parameter/test_parameter_scale_offset.py
@@ -4,7 +4,7 @@ import hypothesis.strategies as hst
 from hypothesis import given, event, settings
 import numpy as np
 
-from qcodes.instrument.parameter import Parameter, DelegateParameter
+from qcodes.instrument.parameter import Parameter
 
 
 def test_scale_raw_value():
--- a/qcodes/tests/parameter/test_parameter_scale_offset.py
+++ b/qcodes/tests/parameter/test_parameter_scale_offset.py
@@ -4,7 +4,7 @@ import hypothesis.strategies as hst
 from hypothesis import given, event, settings
 import numpy as np
 
-from qcodes.instrument.parameter import Parameter
+from qcodes.instrument.parameter import Parameter, DelegateParameter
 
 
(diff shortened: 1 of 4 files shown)


haroldmeerwaldt:
@astafan8 I put the tests in test_parameter_scale_offset.py as you asked. A good place would also be test_delegate_parameter.py.

As I was scrolling through it, I found that the previous behaviour may actually be intended:
https://github.com/QCoDeS/Qcodes/blob/c4fa2c8f83854ff989ca381e9bc65cfd9eac067b/qcodes/tests/parameter/test_delegate_parameter.py#L253

What do you think about the other location, and the change of expected behaviour?

astafan8:
> A good place would also be test_delegate_parameter.py.

true, but i don't think that matters much. leave this decision up to your preference :)

> As I was scrolling through it, I found that the previous behaviour may actually be intended:
Qcodes/qcodes/tests/parameter/test_delegate_parameter.py. What do you think about the change of expected behaviour?

hmmm..... that's a very good question! glad that we have these tests so that we think about this. However, i think this test is only capturing the current behavior as opposed to making it "desired". I don't think it makes sense that get() and snapshot() rasie a TypeError with the scale/offset set and ready while the underlying parameter's value is None - that's where the original issue is coming from, right? that this behavior does not bring value. 

we can ask for a third opinion - @jenshnielsen what is your take on this?

jenshnielsen:
I think its fine to change the test to reflect this new behavior. Just make sure that we document  this by adding a changelog fragment

qcodes/tests/parameter/test_parameter_scale_offset.py
```diff
@@ -4,7 +4,7 @@
 from hypothesis import given, event, settings
 import numpy as np
 
-from qcodes.instrument.parameter import Parameter
+from qcodes.instrument.parameter import Parameter, DelegateParameter
```
jenshnielsen, line 7: Could you remove this import again since its not used

haroldmeerwaldt: Good point! Missed that, I'll remove it.

Question: what do you do with the slight decrease in coverage?

jenshnielsen: Looks like the coverage is now back to baseline. The codecov tool can sometimes be a bit flaky. In general I would check on codecov where the drop is coming from and then decide if that is important and how to best write a test for that

jenshnielsen:
bors merge

PR #3673, opened 2021-12-12
Fix ami430 tests

jenshnielsen:
This test was broken in several ways.

1) The test compares the field rate limit in T/S against the current
rate limits in A/s. This means that ix.ramp_rate(target_ramp_rate)
did not actually raise the error tested for.
2) The test only asserted that any Exception was raised not checking
for a specific type of exception.
3) excinfo.value.args inside the context manager raises a AttributeError
which is then in turn caught by the pytest.raises context manager.
This makes the test pass incorrectly.
4) This in turn causes pytest to leave the tests in an inconsistent state
which results in other tests passing asserts that they should not. 
This means the tests pass if running after this one but don't if they are executed before this test (or independently).

For now I have commented out the failing assertions. It is not clear if these should be changed or the code in the driver should be changed. They test the relationship between the ramp rate in [T/S] on one side and the current ramp rate [A/S] and coilconstant on the other side. However the default values are not compatible since the field ramp rate default is set to 25 in the AMI430.yaml file and _DEFAULT_CURRENT_RAMP_LIMIT = 0.06 and the coil constant to 2.0 in the yaml file

### DIFF
--- a/qcodes/instrument/sims/AMI430.yaml
+++ b/qcodes/instrument/sims/AMI430.yaml
@@ -54,7 +54,7 @@ devices:
           q: 'CONF:RAMP:RATE:SEG {}'
 
       ramp rate field first segment:
-        default: '25, 25, 25'
+        default: '0.11, 0.11, 0.11'
         getter:
           q: 'RAMP:RATE:FIELD:1?'
           r: '{}'
--- a/qcodes/tests/drivers/test_ami430.py
+++ b/qcodes/tests/drivers/test_ami430.py
@@ -74,8 +74,8 @@ def current_driver(magnet_axes_instances):
     driver.close()
 
 
-@pytest.fixture(scope='function')
-def ami430():
+@pytest.fixture(scope="function", name="ami430")
+def _make_ami430():
     mag = AMI430_VISA('ami430', address='GPIB::1::INSTR', visalib=visalib,
                       terminator='\n', port=1)
     yield mag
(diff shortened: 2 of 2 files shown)


jenshnielsen:
For now I have commented out the failing assertions. It is not clear if these should be changed or the code in the driver should be changed. They test the relationship between the ramp rate in [T/S] on one side and the current ramp rate [A/S] and coilconstant on the other side. However the default values are not compatible since the field ramp rate default is set to 25 in the AMI430.yaml file and _DEFAULT_CURRENT_RAMP_LIMIT = 0.06 and the coil constant to 2.0 in the yaml file

jenshnielsen:
This was found due to a random error in https://github.com/QCoDeS/Qcodes/pull/3672 probably caused by the tests executing in a different order

astafan8:
> since the field ramp rate default is set to 25 in the AMI430.yaml file and _DEFAULT_CURRENT_RAMP_LIMIT = 0.06 and the coil constant to 2.0 in the yaml file

i'd say that we should change the values in the yaml file to make sense, e.g. the coil constant. and the uncomment the assertions, becausee with them commented we are not sure if the preconditions for all those tests are met :(

jenshnielsen:
> i'd say that we should change the values in the yaml file to make sense, e.g. the coil constant. and the uncomment the assertions, becausee with them commented we are not sure if the preconditions for all those tests are met :(

That is an okish fix but it would be better if the driver was written in a way that these values were always consistent rather than manually making them such.

astafan8:
> it would be better if the driver was written in a way that these values were always consistent

that requires a bit more effort, at least to test on the instrument :)

jenshnielsen:
bors merge

PR #3676, opened 2021-12-13
Better typing of find_or_create_instrument

jenshnielsen:
This should make it clear that the output type is the same as the input type

Also a small cleanup for the same module

### DIFF
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -819,7 +819,7 @@ class Instrument(InstrumentBase, AbstractInstrument):
                 type(self).__name__))
 
 
-T = TypeVar(name="T", bound=Instrument)
+T = TypeVar("T", bound=Instrument)
 
 
 def find_or_create_instrument(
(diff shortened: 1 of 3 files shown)


qcodes/instrument/base.py
```diff
@@ -672,20 +677,21 @@ def find_instrument(cls, name: str,
             TypeError: If a specific class was requested but a different
                 type was found.
         """
+        internal_instrument_class = instrument_class or Instrument
+
         if name not in cls._all_instruments:
             raise KeyError(f"Instrument with name {name} does not exist")
-        ins = cls._all_instruments[name]()
+        ins = cast(T, cls._all_instruments[name]())
```
jenshnielsen, line 684: Not sure if there is a way to get rid of this cast

jenshnielsen: I think its fair that it is here since this method anyway checks for the type of the class

jenshnielsen: but perhaps better at the end of the method ?

astafan8: yes, better at the end, because actually on this line it can still be None because the instrument might have been closed but the  record of its instance is (for some reason) still in the `_all_instruments`.

astafan8 (approved):

jenshnielsen:
bors merge

PR #3681, opened 2021-12-14
add break conditions to dond, do1d, do2d

grothe6:
This PR adds a break_condition argument to dond, do1d and do2d to interrupt a measurement based on the return value.
It uses the existing code for handling KeyboardInterrupts, but modifies it to raise a new BreakConditionInterrupt instead.

While I did not include it in this PR to keep it simple, maybe its beneficial to pass some more data from the current iteration to the break condition function?


jenshnielsen (approved):
bors merge

PR #3679, opened 2021-12-14
fix export to xarray on empty ds

trevormorgan:

PR #3682, opened 2021-12-14
Let get_parameter_data return empty datasets in a more consistent way

jenshnielsen:
This makes it so that get parameter data for an empty dataset returns a dict with empty arrays rather than an empty dict. This makes other things more consistent down the line and makes fixes such as the ones in https://github.com/QCoDeS/Qcodes/pull/3679 not needed 

Todo:

- [x] Handle broken data set cache tests

### DIFF
--- a/qcodes/dataset/sqlite/queries.py
+++ b/qcodes/dataset/sqlite/queries.py
@@ -306,7 +306,7 @@ def get_parameter_data_for_one_paramtree(
     # faster than transposing the data using np.array.transpose
     res_t = map(list, zip(*data))
 
-    for paramspec, column_data in zip_longest(paramspecs, res_t, fillvalue=[]):
+    for paramspec, column_data in zip_longest(paramspecs, res_t, fillvalue=tuple()):
         if paramspec.type == "numeric":
             # there is no reliable way to
             # tell the difference between a float and and int loaded
--- a/qcodes/dataset/sqlite/queries.py
+++ b/qcodes/dataset/sqlite/queries.py
@@ -7,6 +7,7 @@ import sqlite3
 import time
 import unicodedata
 import warnings
+from itertools import zip_longest
 from typing import (
     Any,
     Callable,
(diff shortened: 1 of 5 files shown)


qcodes/dataset/sqlite/queries.py
```diff
@@ -305,7 +306,7 @@ def get_parameter_data_for_one_paramtree(
     # faster than transposing the data using np.array.transpose
     res_t = map(list, zip(*data))
 
-    for paramspec, column_data in zip(paramspecs, res_t):
+    for paramspec, column_data in zip_longest(paramspecs, res_t, fillvalue=[]):
```
astafan8, line 309: probably does not matter, but should the filler be immutable for safety/sanity?
```suggestion
    for paramspec, column_data in zip_longest(paramspecs, res_t, fillvalue=tuple()):
```

qcodes/tests/dataset/test_dataset_export.py
```diff
@@ -319,15 +319,18 @@ def test_same_setpoint_warning_for_df_and_xarray(different_setpoint_dataset):
...
 
 def test_export_to_xarray_dataarray_empty_ds(mock_empty_dataset):
     dad = mock_empty_dataset.to_xarray_dataarray_dict()
     assert len(dad) == 2
-    assert len(dad["y"].coords) == 0
-    assert len(dad["z"].coords) == 0
+    assert len(dad["y"].coords) == 1
+    assert "x" in dad["y"].coords
+    assert len(dad["z"].coords) == 1
+    assert "x" in dad["z"].coords
```
astafan8, line 333: we need a newsfragment for this! :)

astafan8 (approved):
Fantastic!

is the https://github.com/QCoDeS/Qcodes/pull/3679 fix still needed or can it be removed now?

jenshnielsen:
> is the #3679 fix still needed or can it be removed now?

I would say lets keep it. It should not be needed but its unlikely to break anything

qcodes/dataset/sqlite/queries.py
jenshnielsen: Sure, I cannot see how this would matter but there is no harm

jenshnielsen:
bors merge

PR #3694, opened 2021-12-16
N5222B: Support multiple attenuators_options

cjchua:
The previous version only recognized option 419 as an option with attenuation. This proposed change checks for all other possible options that also add source/receiver attenuation.


qcodes/instrument_drivers/Keysight/N5222B.py
```diff
@@ -11,6 +11,7 @@ def __init__(self, name: str, address: str, **kwargs: Any):
                          nports=4,
                          **kwargs)
 
-        options = self.get_options()
-        if "419" in options:
-            self._set_power_limits(min_power=-90, max_power=13)
+        attenuators_options = {'217', '219', '220', '417', '419', '420'}
+        options = set(self.get_options())
+        if attenuators_options.intersection(options):
+            self._set_power_limits(min_power=-95, max_power=13)
```
jenshnielsen, line 17: Is is correct that all these options results in the same power limits ?

cjchua: Hi Jens,

Yes, I think so. Please see the following links for options on the various PNA models:

https://na.support.keysight.com/vna/help/latest/Support/Configurations.htm#N522x
https://na.support.keysight.com/pna/help/latest/Support/Configurations.htm

It states that the attenuation is the same for these options, but varies for every PNA model.

jenshnielsen:
bors merge

jenshnielsen (approved):

jenshnielsen:
bors merge

PR #3696, opened 2021-12-16
Only record instrument instances when successfully created

### LINKED ISSUE #1882: Improve implementation of instrument instance registry
This issue flags the importance of improving how `Instrument._all_instruments` and `Instrument._instances` are implemented. 

There are cases where closing instrument is important not to forget, otherwise some warning like the ones mentioned in this PR https://github.com/QCoDeS/Qcodes/pull/1875 can start showing up for no good reason.

Anyway, `Insrtument._all_instruments` is intended to hold weak references to instrument instances in order to prevent multiple connections to the same hardware instrument. It is now implemented via simple `dict` which requires methods like `Instrument.exists` and `Instrument.find_instance` to account for the fact that some of those weak references can be `None` because instruments may have gotten closed in the meantime. And this seems to be causing weird unnecessary bugs and dirty code.

My suggestion is to consider using the right tools for the job - use `weakref.WeakValueDictionary` for `_all_instruments` which will automatically take care of the case where a weakref is "dead". For a usage example see `DllWrapper` for `Alazar` driver https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/AlazarTech/dll_wrapper.py#L95.

spauka:
This commit fixes the issue of an instrument instance being registered in the instance dictionary even if the instrument creation failed. This will allow users to retry instrument connection when the first attempt fails without having to restart the interpreter or modify private variables in the Instrument class.

It certainly resolves a source of frustration when trying to debug failing instrument connections 😆 

This partially closes #1882, and should also allow us to be smarter about logging successful connections or cleaning up gracefully on failure.

@jenshnielsen @astafan8 What do you think?

### DIFF
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -257,9 +257,8 @@ class InstrumentBase(Metadatable, DelegateAttributes):
             except:
                 # really log this twice. Once verbose for the UI and once
                 # at lower level with more info for file based loggers
-                self.log.warning(f"Snapshot: Could not update "
-                                 f"parameter: {name}")
-                self.log.info(f"Details for Snapshot:", exc_info=True)
+                self.log.warning("Snapshot: Could not update parameter: %s", name)
+                self.log.info("Details for Snapshot:", exc_info=True)
                 snap['parameters'][name] = param.snapshot(update=False)
 
         for attr in set(self._meta_attrs):
@@ -443,9 +442,11 @@ class AbstractInstrumentMeta(ABCMeta):
     instance is returned to the caller.
 
     Instead we use the fact that `__new__` and `__init__` are called inside
-    `type.__call__` (https://github.com/python/cpython/blob/main/Objects/typeobject.c#L1077)
+    `type.__call__`
+    (https://github.com/python/cpython/blob/main/Objects/typeobject.c#L1077)
     which we will overload to insert our own custom code AFTER `__init__` is
-    complete.
+    complete. Note this is part of the spec and will work in alternate python
+    implementations like pypy too.
 
     Note: Because we want AbstractInstrument to subclass ABC, we subclass
     `ABCMeta` instead of `type`.
@@ -609,12 +610,11 @@ class Instrument(InstrumentBase, AbstractInstrument):
         log.info("Closing all registered instruments")
         for inststr in list(cls._all_instruments):
             try:
-                inst: Instrument = cls.find_instrument(inststr)
-                log.info(f"Closing {inststr}")
+                inst: "Instrument" = cls.find_instrument(inststr)
+                log.info("Closing %s", inststr)
                 inst.close()
             except:
-                log.exception(f"Failed to close {inststr}, ignored")
-                pass
+                log.exception("Failed to close %s, ignored", inststr)
 
     @classmethod
     def record_instance(cls, instance: 'Instrument') -> None:
@@ -716,10 +716,12 @@ class Instrument(InstrumentBase, AbstractInstrument):
 
         if not isinstance(ins, internal_instrument_class):
             raise TypeError(
-                f"Instrument {name} is {type(ins)} but {internal_instrument_class} was requested"
+                f"Instrument {name} is {type(ins)} but "
+                f"{internal_instrument_class} was requested"
             )
-        # at this stage we have checked that the instrument is either of type instrument_class
-        # or Instrument if that is None. It is therefor safe to cast here.
+        # at this stage we have checked that the instrument is either of
+        # type instrument_class or Instrument if that is None. It is
+        # therefore safe to cast here.
         ins = cast(T, ins)
         return ins
 
@@ -807,8 +809,8 @@ class Instrument(InstrumentBase, AbstractInstrument):
             cmd: The string to send to the instrument.
         """
         raise NotImplementedError(
-            'Instrument {} has not defined a write method'.format(
-                type(self).__name__))
+            f"Instrument {type(self).__name__} has not defined a write method"
+        )
 
     def ask(self, cmd: str) -> str:
         """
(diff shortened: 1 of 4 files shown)


jenshnielsen:
Looks need. Reading a bit about Class creation it does seem like the correct approach to replace call with a different method. 
As far as I currently understand it what `__call__` roughly does is to call `__new__` and then `__init__` and then return the object. I will need to spent a little bit of time to figure out if this is safe in general or there are unwanted side effects of doing this

spauka:
Hey, yep that's my understanding too. Instance creation is performed by a call to an object that is a subclass of `type`, where the type `type` implements a call that is similar to the following:
```py3
def __call__(cls, *args, **kwargs):
    new_inst = cls.__new__(*args, **kwargs)
    if type(new_inst) != cls:
        # __init__ is only run if we return an instance of the expected type. 
        # This is hit mostly when we want to extend immutable types (int, tuple)
        # where we can't run __init__.
        return new_inst
    new_inst.__init__(*args, **kwargs)
    return new_inst
```

`type.__new__` is responsible for actually creating the new instance in memory, however because of the implementation of `type.__call__` without a hack there is no way to intercept the call to `__init__` and inject code afterwards. The only alternate approaches I can think of that doesn't require modification of user classes would be:
1. Change the return type in `Instrument.__new__` before we return the new instance to `__call__`, then call `__init__` in `Instrument.__new__`. I think this is a less logical solution since we are changing the type of the new instance dynamically and it is far less logical to the end user what is happening.

Since this does touch so much QCoDeS a thorough review is definitely warranted. Let me know if I can answer any questions.

qcodes/instrument/base.py
```diff
@@ -434,7 +433,36 @@ def validate_status(self, verbose: bool = False) -> None:
...
+    """
+    Metaclass used to customize Instrument creation. We want to register the
+    instance iff __init__ successfully runs, however we can only do this if
+    we customize the instance initialization process, otherwise there is no
+    way to run `register_instance` after `__init__` but before the created
+    instance is returned to the caller.
+
+    Instead we use the fact that `__new__` and `__init__` are called inside
+    `type.__call__`
+    (https://github.com/python/cpython/blob/main/Objects/typeobject.c#L1077)
```
jenshnielsen, line 446: could you add a link to https://github.com/python/typeshed/blob/master/stdlib/builtins.pyi#L156
too

That shows how mypy expects call to be typed. Which is consistent with this

qcodes/tests/instrument_mocks.py
```diff
@@ -136,6 +136,24 @@ def __init__(self, name: str = 'dummy',
...
+    def __init__(self, name: str = "dummy", fail: bool = True, **kwargs):
+
+        """
+        Create a dummy instrument that fails on initialization
+        that can be used for testing
+
+        Args:
+            name: name for the instrument
+            gates: list of names that is used to create parameters for
+                            the instrument
```
jenshnielsen, line 149: remove docstring about gates ?

qcodes/tests/test_instrument.py
```diff
@@ -61,6 +66,26 @@ def test_check_instances(testdummy):
     assert testdummy.instances() == [testdummy]
 
 
+def test_instrument_fail(close_before_and_after):
+    with pytest.raises(RuntimeError):
+        instr = DummyFailingInstrument(name="failinginstrument")
```
jenshnielsen, line 71:
```suggestion
        instr = DummyFailingInstrument(name="failinginstrument", match="Failed to create instrument")
```

qcodes/tests/test_instrument.py
```diff
@@ -61,6 +66,26 @@ def test_check_instances(testdummy):
...
+
+    assert Instrument.instances() == []
+    assert DummyFailingInstrument.instances() == []
+    assert Instrument._all_instruments == {}
+
+
+def test_instrument_retry_with_same_name(close_before_and_after):
+    with pytest.raises(RuntimeError):
+        instr = DummyFailingInstrument(name="failinginstrument")
```
jenshnielsen, line 80:
```suggestion
    with pytest.raises(RuntimeError, match="Failed to create instrument"):
        DummyFailingInstrument(name="failinginstrument")
```

qcodes/tests/test_instrument.py
jenshnielsen, line 71:
```suggestion
    with pytest.raises(RuntimeError, match="Failed to create instrument"):
        DummyFailingInstrument(name="failinginstrument")
```

spauka:
Thanks for the review @jenshnielsen - agreed, accepted and implemented all changes :)

jenshnielsen:
@spauka I think this looks good. I will do a bit more manual testing and hold of the merge until the new year in case any issues show up

jenshnielsen:
Once this is merged we should revisit the reverted abstract instrument. https://github.com/QCoDeS/Qcodes/pull/3197 That can be implemented in the same way (Checking that all parameters are non abstract post init) 

Also see https://github.com/QCoDeS/Qcodes/pull/3217

qcodes/instrument/base.py
```diff
@@ -434,7 +433,37 @@ def validate_status(self, verbose: bool = False) -> None:
...
+    https://github.com/python/typeshed/blob/master/stdlib/builtins.pyi#L156)
+    which we will overload to insert our own custom code AFTER `__init__` is
+    complete. Note this is part of the spec and will work in alternate python
+    implementations like pypy too.
+
+    Note: Because we want AbstractInstrument to subclass ABC, we subclass
+    `ABCMeta` instead of `type`.
+    """
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
```
jenshnielsen, line 456:
```suggestion
    def __call__(cls, *args: Any, **kwargs: Any) -> Any:
```

spauka: Good catch :)

jenshnielsen:
bors merge

jenshnielsen (approved):

jenshnielsen:
bors merge

PR #3708, opened 2021-12-20
Drop use of Distutils LooseVersion

jenshnielsen:
All of distutils is deprecated in python 3.10 and the recommended alternative for LooseVersion is packaging.version.parse

This removes all use of distutils outside versioneer

### DIFF
--- a/qcodes/instrument_drivers/Keysight/keysight_e4980a.py
+++ b/qcodes/instrument_drivers/Keysight/keysight_e4980a.py
@@ -1,6 +1,6 @@
-from distutils.version import LooseVersion
 from typing import Any, Sequence, Tuple, Union, cast
 
+from packaging import version
 from pyvisa.errors import VisaIOError
 
 from qcodes import InstrumentChannel, VisaInstrument
@@ -194,9 +194,9 @@ class KeysightE4980A(VisaInstrument):
 
         idn = self.IDN.get()
 
-        self.has_firmware_a_02_10_or_above = (
-                LooseVersion(idn["firmware"]) >= LooseVersion("A.02.10")
-        )
+        self.has_firmware_a_02_10_or_above = version.parse(
+            idn["firmware"]
+        ) >= version.parse("A.02.10")
 
         self.has_option_001 = '001' in self._options()
         self._dc_bias_v_level_range: Union[Numbers, Enum]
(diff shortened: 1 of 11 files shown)


qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -464,8 +465,7 @@ def __init__(self, name: str, address: str, silent: bool = False,
 
         options = self._options()
         self.has_DIG = self.is_34465A_34470A and (
-            'DIG' in options
-            or LooseVersion('A.03') <= LooseVersion(idn['firmware'])
+            "DIG" in options or version.parse("A.03") <= version.parse(idn["firmware"])
```
jenshnielsen, line 468: Looks like versions of this form are considdered legacy and also deprecated using packaging. We should look for an alternative way of doing version comparisons in these cases

qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -194,9 +194,9 @@ def __init__(self,
 
         idn = self.IDN.get()
 
-        self.has_firmware_a_02_10_or_above = (
-                LooseVersion(idn["firmware"]) >= LooseVersion("A.02.10")
-        )
+        self.has_firmware_a_02_10_or_above = version.parse(
+            idn["firmware"]
+        ) >= version.parse("A.02.10")
```
jenshnielsen, line 199: same here

qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
jenshnielsen: Perhaps one could simply convert letters to their ascii mapping before passing to parse

```
In [39]: verstr = "a.1.0"
    ...: b = []
    ...: for v in verstr:
    ...:     if v.isalpha():
    ...:         b.append(str(ord(v.upper())))
    ...:     else:
    ...:         b.append(v)
    ...: c = "".join(b)
```

jenshnielsen:
@astafan8 I think this should be ok. Can you have another look

astafan8 (approved):
looks good!

do we still need the filterwarnings?

jenshnielsen:
> do we still need the filterwarnings?

The filterwarnings are added since xarray and hdf5netcdf raises these warnings too. I have verified both by running the tests and grepping that we do use distutils internally (other than versioneer)

jenshnielsen:
bors merge

PR #3713, opened 2021-12-20
Remove less important docs build from CI

jenshnielsen:
* Drop 3.7 on ubuntu (we are already building with 3.7 on window)
* Drop 3.8 (we both test with 3.7 and 3.9 so this is very unlikely to break)


.github/workflows/docs.yaml
```diff
@@ -18,18 +18,20 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: [3.7, 3.8, 3.9, "3.10"]
+        python-version: [3.7, 3.9, "3.10"]
         exclude:
           - os: windows-latest
             python-version: 3.8
```
astafan8, line 24:
```suggestion
```

astafan8 (approved):

jenshnielsen:
bors merge

jenshnielsen:
bors merge

jenshnielsen:
Bors merge

PR #3714, opened 2021-12-20
Cleanup websockets test errors

jenshnielsen:
test_parameter was not correctly closing the connection before closing the event loop. Using a async context manager makes it simpler to ensure this is the case


astafan8 (approved):
bors merge

PR #3716, opened 2021-12-21
disable timeout on test that sometimes fails in ci

jenshnielsen:
For example here https://github.com/QCoDeS/Qcodes/runs/4591761964?check_suite_focus=true


astafan8 (approved):
bors merge

PR #3718, opened 2021-12-21
Reimplement abstract instrument

jenshnielsen:
This is a start of re implementing the abstract parameter interface reverted in https://github.com/QCoDeS/Qcodes/pull/3197 and https://github.com/QCoDeS/Qcodes/pull/3217 based on top of https://github.com/QCoDeS/Qcodes/pull/3696

Compared to the solution based on init subclass this works a bit differently. The metaclass in use is only implemented for the Instrument class but not for the InstrumentChannel. Changing this would require metaclass inheritance which is a bad idea. 
This meas that it is not an error to create an InstrumentChannel with abstract parameters but it is an error to attach such a thing to an instrument at instrument creation time. I don't think this is a big issue since that is how most instruments are created. 

Still need to update docs again and verify if there are other reverted changes to bring over

### DIFF
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -359,12 +359,10 @@ class InstrumentBase(Metadatable, DelegateAttributes):
 
     def _raise_if_abstract(self) -> None:
         """
-        This method is run after the initialization of a subclass of
-        `InstrumentBase`. If we have a chain of base classes like so:
-        InstrumentBase -> Instrument -> VisaInstrument -> Driver
-        This method is run after the instantiation of the class `Driver`.
-        Please see the docstring of `InstrumentBase.__init_subclass__`
-        for details about the mechanism which makes this possible.
+        This method is run after the initialization of an instrument but
+        before the instrument is registered. It recursively checks that there are
+        no abstract parameters defined on the instrument or any instrument channels.
+        If a non overwritten abstract parameter is found it raises a ``NotImplementedError``
         """
         abstract_parameters = [
             parameter.name
--- a/qcodes/instrument/base.py
+++ b/qcodes/instrument/base.py
@@ -357,13 +357,13 @@ class InstrumentBase(Metadatable, DelegateAttributes):
     def full_name(self) -> str:
         return "_".join(self.name_parts)
 
-    def _raise_if_abstract(self) -> None:
+    def _is_abstract(self) -> bool:
         """
         This method is run after the initialization of an instrument but
         before the instrument is registered. It recursively checks that there are
         no abstract parameters defined on the instrument or any instrument channels.
-        If a non overwritten abstract parameter is found it raises a ``NotImplementedError``
         """
+        is_abstract = False
         abstract_parameters = [
             parameter.name
             for parameter in self.parameters.values()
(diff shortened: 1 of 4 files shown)


astafan8:
> but it is an error to attach such a thing to an instrument at instrument creation time. I don't think this is a big issue since that is how most instruments are created

agree! that's still great and useful!

qcodes/instrument/base.py
```diff
@@ -357,6 +356,35 @@ def name_parts(self) -> List[str]:
...
         return "_".join(self.name_parts)
+
+    def _raise_if_abstract(self) -> None:
+        """
+        This method is run after the initialization of a subclass of
+        `InstrumentBase`. If we have a chain of base classes like so:
+        InstrumentBase -> Instrument -> VisaInstrument -> Driver
+        This method is run after the instantiation of the class `Driver`.
+        Please see the docstring of `InstrumentBase.__init_subclass__`
+        for details about the mechanism which makes this possible.
```
astafan8, line 367:
```suggestion
        Please see `AbstractInstrumentMeta`
        for details about the mechanism which makes this possible.
```

astafan8 (commented):
Looks great! :)

qcodes/instrument/base.py
```diff
@@ -357,6 +356,35 @@ def name_parts(self) -> List[str]:
     @property
     def full_name(self) -> str:
         return "_".join(self.name_parts)
+
+    def _raise_if_abstract(self) -> None:
+        """
+        This method is run after the initialization of a subclass of
+        `InstrumentBase`. If we have a chain of base classes like so:
```
jenshnielsen, line 363: Clarify that this is only subclasses of instrument

qcodes/instrument/base.py
```diff
@@ -357,6 +356,35 @@ def name_parts(self) -> List[str]:
...
+        InstrumentBase -> Instrument -> VisaInstrument -> Driver
+        This method is run after the instantiation of the class `Driver`.
```
spauka, line 365: Although the comment is accurate, I think calling out the chain of base-classes is a bit of a misdirect and makes this a bit confusing. I ended up thinking about whether we were hooking in between __init__ calls, which I think was how the first implementation of this feature worked?

Might be better to remove the chain of base classes and only give the first sentence.

jenshnielsen: No this is copied from the old implementation, my comment above was meant as a reminder to update this

qcodes/instrument/base.py
```diff
@@ -459,6 +486,7 @@ def __call__(cls, *args: Any, **kwargs: Any) -> Any:
         successfully.
         """
         new_inst = super().__call__(*args, **kwargs)
+        new_inst._raise_if_abstract()
```
jenshnielsen, line 489: We should perhaps close the instrument if this throws such that we don't leak visa handles etc

jenshnielsen: Yes @astafan8 and I were brainstorming a bit about this. Looking around it may be a more portable solution to add a weakref finalizer to the visa handle it self so it is always closed if an exception is thrown rather that trying to handle it at this level. In this specific case calling close should be safe thou

astafan8 (approved):

jenshnielsen:
Bors merge

PR #3725, opened 2021-12-23
move pytest related config to pyproject.toml

jenshnielsen:
This was added in pytest 6.0 which is anyway the oldest version that we support. 
Pytest recommends against using setup.cfg for config now https://docs.pytest.org/en/6.2.x/customize.html#setup-cfg


jenshnielsen:
Need to change/remove logic from ci job that copies setup.cfg (has no effect now)

astafan8 (approved):

jenshnielsen:
bors merge

PR #3731, opened 2021-12-24
Replace versioneer with Versioningit

jenshnielsen:
This means that we can remove the need to bundle code from versioneer and rely on pep518 to mange the build time dependency

Unfortunately there is a bit of logic required to ensure that editable installs reports the correct version (the version checked out at time of running not the version at time of install) I suspect some of this logic may be a bit slow and we should benchmark that before merging this.


jenshnielsen:
Remember to remove version files from mypy exclusion etc

qcodes/utils/installation_info.py
```diff
@@ -46,11 +47,40 @@ def is_qcodes_installed_editably() -> Optional[bool]:
...
 def get_qcodes_version() -> str:
     """
     Get the version of the currently installed QCoDeS
     """
-    return qcodes.__version__
+    if is_qcodes_installed_editable_fast():
+        import versioningit
+
+        pyprojectpath = Path(os.path.abspath(__file__)).parent.parent.parent
+        __version__ = versioningit.get_version(project_dir=pyprojectpath)
```
jenshnielsen, line 80: should probably wrap this in a try except log in case the pyproject.toml file cannot be found for whatever reason

jenshnielsen:
replaced by #3803

PR #3732, opened 2021-12-25
cache lookup of editable package

jenshnielsen:
Since this function is rather slow it makes sense to cache the lookup. My local benchmark indicates that this takes ~1.5 s but a cached version should evaluate in ns.

Also refactor the code a bit such that one can lookup editable status by name.

### DIFF
--- a/qcodes/utils/installation_info.py
+++ b/qcodes/utils/installation_info.py
@@ -39,7 +39,7 @@ def is_package_installed_editably(package: str) -> Optional[bool]:
                                  check=True,
                                  stdout=subprocess.PIPE)
         e_pkgs = json.loads(pipproc.stdout.decode('utf-8'))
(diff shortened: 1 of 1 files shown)


jenshnielsen:
Not sure if this is the way to go. Doing this manually without pip as in https://github.com/QCoDeS/Qcodes/pull/3731 is significantly faster and may be fast enough that this is not needed

trevormorgan:
@jenshnielsen is this something we would still like to merge or is it no longer needed due to versioningit?

jenshnielsen:
If we did this, we should cache the output of the pip command e.g. `e_pkgs` rather than pr package since the query is always the same and that is what is slow. That seems useful but not really as something that belongs in qcodes so lets close this for now

PR #3735, opened 2021-12-25
Only import Azure opencensus when needed

jenshnielsen:
This saves around 0.7 s when doing a fresh import without activating logging


Dominik-Vogel (approved):
Can we track this improvement somewhere to also implement it in the rest of the QI stack?

jenshnielsen:
bors merge

PR #3762, opened 2022-01-03
Avoid calculating version twice on startup

jenshnielsen:
this is already statically avaliable as ``qcodes.__version__``

This shaves of 0.3 sec or so from the start up time. 
It may be worthwile deprecating the version module. However https://github.com/QCoDeS/Qcodes/pull/3731 fixes this anyway in a way that ensures version is only calculated once per session

### DIFF
--- a/qcodes/data/hdf5_format.py
+++ b/qcodes/data/hdf5_format.py
@@ -6,7 +6,8 @@ from typing import TYPE_CHECKING
 import h5py
 import numpy as np
 
-from ..version import __version__ as _qcodes_version
+import qcodes as qc
+
 from .data_array import DataArray
 from .format import Formatter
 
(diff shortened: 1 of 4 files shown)


jenshnielsen:
Note that this is technically an api break as qcodes.version used to be a valid module after importing qcodes (since the data array module pulled it in) without explicitly importing it. That is no longer the case

jenshnielsen:
We should perhaps give this a breaking change entry

qcodes/data/hdf5_format.py
```diff
@@ -6,7 +6,8 @@
 import h5py
 import numpy as np
 
-from ..version import __version__ as _qcodes_version
```
astafan8, line 9: > It may be worthwile deprecating the version module

agree :)

astafan8, line 9: > breaking change

yes, let's add the note for the changelog

astafan8 (approved):

jenshnielsen:
bors merge

PR #3763, opened 2022-01-03
Speedup avoid loading pkg_resources at startup

jenshnielsen:
Avoid an import time dependency on pkg_resources. This takes 0.3-4 s to import

To do this 

* Make import lazy in two instances where it is only used for logging
* Dont use this to generate absolute paths to config files. This does in principle mean that these are no longer zip safe. However QCoDeS is currently not marked as zip safe and running from a zip file is not really something we want to support

### DIFF
--- a/qcodes/configuration/config.py
+++ b/qcodes/configuration/config.py
@@ -3,11 +3,11 @@ import copy
 import json
 import logging
 import os
-import pkg_resources as pkgr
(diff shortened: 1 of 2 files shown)


astafan8 (approved):

jenshnielsen:
bors merge

PR #3733, opened 2021-12-25
lazily import xarray in legacy dataset

jenshnielsen:
importing xarray is slow so avoid doing this at qcodes import time

Locally this saves ~ 2 sec on a fresh import of qcodes

PR #3773, opened 2022-01-04
work around issue in pyqt plot with remote process

jenshnielsen:
It seems like the changes in #https://github.com/QCoDeS/Qcodes/pull/3733 made TestQtPlot.test_simple_plot fail for me locally with an error about AttributeError: 'AxisItem' object has no attribute '_qcodes_label' 

This seems to be related to the pyqt remote object used here as the AxisItem clearly has this item assigned. As a work around we use getattr here which does seem to solve the issue


astafan8 (approved):

jenshnielsen:
bors merge

PR #3777, opened 2022-01-05
Support numpy 1.22

jenshnielsen:
Numpy 1.22 adds a fair amount of types so there are a few tweaks to get it correct. 

The most important ones are probably the changes around fromstring / frombuffer. 
* fromstring is effectively deprecated in the format that we use it and it is recommended to use frombuffer as done here
* using from buffer with a ctypes array seems to be not recommended so we use the numpy ctypes api here. This results in a simplifications since we no longer have to manually map between ctypes and numpy types

### DIFF
--- a/docs_requirements.txt
+++ b/docs_requirements.txt
@@ -37,7 +37,8 @@ nbformat==5.1.3
 nbsphinx==0.8.8
 nest-asyncio==1.5.4
 notebook==6.4.6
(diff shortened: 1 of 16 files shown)


astafan8 (approved):

astafan8 (approved):

jenshnielsen:
bors merge

PR #3783, opened 2022-01-06
Improve Keithley SMU 2600 calibration routine

emartinez:
Adding Keithley calibration example.

Some of the instruments/functions used are not defined in the code, and are assumed to already exist in the kernel session.


docs/examples/driver_examples/calibrate_keithley.py
```diff
@@ -0,0 +1,215 @@
...
+)
+
+dmm = Keysight_34470A("dmm", "TCPIP0::10.164.54.211::inst0::INSTR")
+
+#%% Define current experiment
+load_or_create_experiment(
+    "MeasurementSetupDebug", context.devices[-1].name, load_last_duplicate=True
+)
+
+# %% calibrate both channels
```
astafan8, line 166: we'll need to edit the existing example notebook to refer to this

docs/examples/driver_examples/calibrate_keithley.py
```diff
@@ -0,0 +1,215 @@
...
+#%% Load instruments (the SMU is assumed to be imported)
+from qcodes.instrument_drivers.Keysight.Keysight_34470A_submodules import (
+    Keysight_34470A,
+)
+
+dmm = Keysight_34470A("dmm", "TCPIP0::10.164.54.211::inst0::INSTR")
+
+#%% Define current experiment
+load_or_create_experiment(
+    "MeasurementSetupDebug", context.devices[-1].name, load_last_duplicate=True
```
astafan8, line 163:
```suggestion
    "MeasurementSetupDebug", "no sample", load_last_duplicate=True
```

astafan8 (commented):
thanks @emartinez ! i will help with integrating it a bit better into qcodes

docs/examples/driver_examples/calibrate_keithley.py
```diff
@@ -0,0 +1,215 @@
+#%% Function definitions
+import time
+from qcodes.instrument.base import Instrument
+from qcodes.instrument.parameter import Parameter
+
```
astafan8, line 5: need to add a change note

docs/examples/driver_examples/calibrate_keithley.py
```diff
@@ -0,0 +1,215 @@
...
+smu.smua.volt(0)
+smu.smub.volt(0)
+
+#%% calibrate single channel in specific range
+
+dmm.aperture_time(1.0)
+dmm.range(1)
+dmm.autozero("OFF")
+dmm.autorange("OFF")
```
astafan8, line 182: is there are reason for these settings to be outside of the calibration function?

docs/examples/driver_examples/calibrate_keithley.py
```diff
@@ -0,0 +1,215 @@
...
+smun.measurerange_v(200e-3)
+smun.volt(0)
+smun.output("on")
+
+dmm.aperture_time(0.1)
+dmm.range(1)
+dmm.autozero("OFF")
+dmm.autorange("OFF")
+
+do1d(
```
astafan8, line 201: this likely should be wrapped into its own function

docs/examples/driver_examples/calibrate_keithley.py
```diff
@@ -0,0 +1,215 @@
...
+    src_Z: float = 1e-30,
+    time_delay: float = 3.0,
+    save_calibrations: bool = False,
+) -> None:
+    smu_ranges = ["200e-3", "2", "20"]
+    dmm_ranges = [1, 10, 100]
+    for channel in channels:
+        input(f"Please connect channel {channel} to V input on calibrated DMM.")
+        for smu_range, dmm_range in zip(smu_ranges, dmm_ranges):
+            dmm_param.range(dmm_range)
```
astafan8, line 28: what is dmm_param? isn't it the dmm instrument actually? or does dmm have a parameter that has a fancy custom range method on it?

docs/examples/driver_examples/calibrate_keithley.py
```diff
@@ -0,0 +1,215 @@
...
+        time.sleep(time_delay)
+
+        # Debug output
+        print(
+            f"{channel}.source.calibratev({v_range}, {src_Z}, {DMM_Z_rdg:.6e}, {src_FS:.6f}, {DMM_FS_rdg:.6f})"
+        )
+        print(
+            f"{channel}.measure.calibratev({v_range}, Z_rdg, {DMM_Z_rdg:.6e}, FS_rdg, {DMM_FS_rdg:.6f})"
+        )
+
```
astafan8, line 105: these can be removed because qcodes has logging per instrument instance and all commands are logged, so to debug the output it's just a matter of lifting the logging label (see qcodes documentation on logging for details on how to do that)
```suggestion
```

docs/examples/driver_examples/calibrate_keithley.py
emartinez: Fixed, it was supposed to be the DMM instrument.

docs/examples/driver_examples/calibrate_keithley.py
emartinez: Sounds good to me.

docs/examples/driver_examples/calibrate_keithley.py
emartinez: Agree. Fixed

docs/examples/driver_examples/calibrate_keithley.py
astafan8: fixed

astafan8 (approved):

PR #3784, opened 2022-01-06
Dataset: remove imports only needed for typechecking

jenshnielsen:
Will potentially help break some non needed import cycles


qcodes/dataset/data_set_protocol.py
```diff
@@ -41,13 +40,15 @@
...
+
     from .data_set_cache import DataSetCache
 
 array_like_types = (tuple, list, np.ndarray)
 scalar_res_types = Union[str, complex, np.integer, np.floating, np.complexfloating]
 values_type = Union[scalar_res_types, np.ndarray, Sequence[scalar_res_types]]
-res_type = Tuple[Union[_BaseParameter, str], values_type]
-setpoints_type = Sequence[Union[str, _BaseParameter]]
+res_type = Tuple[Union["_BaseParameter", str], values_type]
+setpoints_type = Sequence[Union[str, "_BaseParameter"]]
```
astafan8, line 51: in this module stringifying is not needed due to `from __future__ import annotations` , right?

astafan8 (approved):

qcodes/dataset/data_set_protocol.py
jenshnielsen: Unfortunatly the string repr is still needed in this context https://www.python.org/dev/peps/pep-0563/#forward-references

PR #3801, opened 2022-01-10
Remove constrains on types packages

jenshnielsen:
These are unlikely to fail, releases often and causese conflict between each other so I think it is better to not let dependabot update these


astafan8 (approved):
ok, let's try :)

jenshnielsen:
bors merge

PR #3802, opened 2022-01-10
Update ATS9440.py to fix the 'ABCD' value mapping

gdeboo:
'ABCD' value mapping should be the bitwise or of 0x00000001 | 0x00000002 | 0x00000004 | 0x00000008, which is 15


qcodes/instrument_drivers/AlazarTech/ATS9440.py
```diff
@@ -238,7 +238,7 @@ def __init__(self, name: str,
                                         'AD': 8,
                                         'BD': 9,
                                         'CD': 10,
-                                        'ABCD': 11})
```
jenshnielsen, line 241: It looks like the mapping of D is probably also wrong then 7 should be changed to 8 and therefor AB, AD and CD are also wrong, could you check that?

gdeboo: Yes, D was also wrong. I added another to commit to fix that.

jenshnielsen:
@gdeboo Thanks for the fix. Could you add a changelog entry as described in the pr template ?

- [ ] Create a file in the docs\changes\newsfragments folder with a short description of the change.

This file should be in the format number.categoryofcontribution. Here the number should either be the number
of the pull request. To get the number of the pull request one must
first the pull request and then subsequently update the number. The category of contribution should be
one of ``breaking``, ``new``, ``improved``, ``new_driver`` ``improved_driver``, ``underthehood``.

jenshnielsen:
bors merge

jenshnielsen (approved):

jenshnielsen:
bors merge

PR #3803, opened 2022-01-10
Replace versioneer with Versioningit take 2

jenshnielsen:
This is an alternative version of https://github.com/QCoDeS/Qcodes/pull/3731 which rather than looking up if a package is installed ediatablly simply checks if the top folder is a git repo with a pyproject.toml file. I think this is a good deal simpler and easier to extend to the situation where there is an additional src folder.

I am fairly happy with this as is. If we want to go this way I will need to clean up the git history before merging 

One drawback is that versioningit is a bit slower than versionner

**Versioneer**

image

**versioningit** 

image

Note that this however only affects editable installes. The path taken with non editable install is still fast

image


qcodes/utils/installation_info.py
```diff
@@ -3,19 +3,27 @@
...
     from importlib.metadata import PackageNotFoundError, distribution, version
 else:
     # 3.7 and earlier
-    from importlib_metadata import distribution, version, PackageNotFoundError
+    from importlib_metadata import PackageNotFoundError, distribution, version
+
+if sys.version_info >= (3, 9):
+    from importlib.resources import files
+else:
+    # 3.7 and earlier
```
jenshnielsen, line 24:
```suggestion
    # 3.8 and earlier
```

qcodes/utils/installation_info.py
```diff
@@ -47,7 +53,18 @@ def get_qcodes_version() -> str:
     """
     Get the version of the currently installed QCoDeS
     """
-    return qcodes.__version__
+    package_name = "qcodes"
+
+    qcodes_path = files(package_name)
```
jenshnielsen, line 58: This breaks since this does not give the path to the qcodes install folder but to the source if you run this from the checkout dir.

qcodes/utils/installation_info.py
```diff
@@ -47,7 +46,22 @@ def get_qcodes_version() -> str:
     """
     Get the version of the currently installed QCoDeS
     """
-    return qcodes.__version__
+    import qcodes
+    package_name = "qcodes"
+
+    qcodes_path = Path(qcodes.__file__).parent
+    if _has_pyproject_toml_and_is_git_repo(qcodes_path.parent):
```
astafan8, line 53: ok, so this becomes the only project-specific piece and the rest will be the same when we want to expand this to other packages, right? that's good, i think :)

astafan8 (approved):
so the only difference from https://github.com/QCoDeS/Qcodes/pull/3731 is that instead of trying to be general/generic and implement peps and setuptools-standards, we just using simple git+pyproject.toml check with is good enough for qcodes? if so, great, i approve :)

qcodes/utils/installation_info.py
jenshnielsen: Yes basically if you should check .parent or .parent.parent

jenshnielsen:
bors merge

PR #3808, opened 2022-01-10
Remove manifest

jenshnielsen:
With setuptools 56 the licence is automatically included and there is no need to specify it in the manifest. 
This means that we have one less file to clutter the main source dir


astafan8 (approved):

jenshnielsen:
bors merge

PR #3820, opened 2022-01-12
Channel ordered monitor

ThorvaldLarsen:
Add function argument for monitor creation to allow grouping of parameters according to channels rather than instrument. Specifically, group according to `parameter.instrument` instead of `parameter.root_insturment`. Default behavior is the current behavior with sorting by `root_instrument`.

Code has been tested on qcodes version 0.30.0 on one system.

Potentially it could be nice to be able to pass in a manual sorting by allowing direct input of the parameter dictionary rather than going through `_get_metadata`.


jenshnielsen:
bors merge

PR #3823, opened 2022-01-13
Precommit: Run darker relative to master

jenshnielsen:
As currently configured it would only run darker on the latest commit. That works greak if you have the pre commit installed locally but means that we miss cleanup when the formatter runs in ci.

Also remove an exclude that is no longer needed after we removed versioneer.


jenshnielsen:
it seems that master is not fetched in precommit ci so this does not work

PR #3834, opened 2022-01-18
Register InstrumentModules separately from ChannelLists

jenshnielsen:
- [x] Register channels and channel lists in two separate dicts. This is done on the basis of the ABC to avoid a circular import

Open questions 

- [ ] Deprecate submodules dict (yes but not now)
- [ ] Adapt snapshot etc to use the channels rather than (again not now)
- [x] Better names for the new dicts (channels is not an option since many instruments uses that as the name of the channel list)


astafan8 (approved):

jenshnielsen:
Bors merge

PR #3835, opened 2022-01-18
Let ChannelList implement the collections.abc.Sequence interface

jenshnielsen:
And register this as such so that 
isinstance and issubclass works as expected

Partially taken from #3834

### DIFF
--- a/qcodes/instrument/channel.py
+++ b/qcodes/instrument/channel.py
@@ -1,4 +1,5 @@
 """ Base class for the channel of an instrument """
+import collections.abc
 from typing import (
     Any,
     Callable,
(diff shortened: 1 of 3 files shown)


qcodes/instrument/channel.py
```diff
@@ -543,6 +578,7 @@ def print_readable_snapshot(self, update: bool = False,
                 channel.print_readable_snapshot(update=update,
                                                 max_chars=max_chars)
 
+collections.abc.Sequence.register(ChannelList)
```
astafan8, line 581: so this makes ChannelList a Sequnce subclass without inheriting, right? and why do we not want or cannot inherit? because then "isinstance and issubclass" won't work as expected?

astafan8 (approved):

qcodes/instrument/channel.py
jenshnielsen: Mostly this seems to be recommended for existing classes to avoid introducing any unwanted api changes but in this case it would probably be ok

jenshnielsen: @astafan8 pushed a version that implements this as a subclass. This revealed that index was not implemented fully correctly so I guess its an improvement

astafan8 (approved):

PR #3839, opened 2022-01-19
remove wrongly named news fragment

jenshnielsen:
The format is nnnn.underthehood not nnnn.Underthehood so this was never included (and deleted). Remove now to avoid this being imitated


astafan8 (approved):

jenshnielsen:
bors merge

PR #3840, opened 2022-01-19
Introduce InstrumentModule as a superclass of InstrumentChannel

jenshnielsen:
And clarify (via typing) that submodule in an instrument must be either InstrumentModules or ChannelLists

Open questions:

- [x] Should we retype ChannelList and friends to allow InstrumentModule as a type. Not right now. We may add more channel specific functionality to InstrumentChannel down the line 
- [x] Update docs


astafan8 (commented):
> Should we retype ChannelList and friends to allow InstrumentModule as a type

hmm. so that channellist may contain instrument modules, not just instrument channels? if so, then i'd probably wait with that since i can't think of an instrument were i'd want to have a channel list of non-channels, e.g. of things/modules that don't have a port on the panel of the instrument.

> Update docs

yes! :)

also need to add a newsfragment :)

jenshnielsen:
@astafan8 Thanks docs are updated and newsfragment added

astafan8 (approved):

PR #3841, opened 2022-01-19
Stop logging versions of QCoDeS requirements

jenshnielsen:
We are already logging versions of all installed packages so this seems not needed.
This will eventually allow us to drop the dependency on requirements_parser and is one more step towards dropping the 
run time dependency on setuptools.


astafan8 (approved):

Dominik-Vogel (approved):

jenshnielsen:
bors merge

PR #3852, opened 2022-01-23
Cleanup check for is abstract in instrument.

jenshnielsen:
Now that we can split modules and channellists this can be done
in a cleaner way. This also means that we can now check channels that are not added individually but in a list


jenshnielsen:
Add tests covering

* instruments with abstract channels
* Instruments with abstract channels only in a channel list

jenshnielsen:
> Add tests covering
> 
>     * instruments with abstract channels
> 
>     * Instruments with abstract channels only in a channel list

Done

astafan8 (approved):

jenshnielsen:
bors merge

PR #3856, opened 2022-01-24
ChannelTuple/List correctly handle getting an attribute from an empty list

jenshnielsen:
This will ensure that if you do a instr.channel.parameter you get an attribute error as expected and not an index error.
This also means that getattr etc will work as expected since it can catch the correct error. 

Also refactor the getattr code a bit to split out a method for the multiparameter construction to keep the function at a resonable lenght. 

This is done on top of https://github.com/QCoDeS/Qcodes/pull/3851 which should be merged first.


astafan8 (approved):
bors merge

jenshnielsen:
bors merge

PR #3857, opened 2022-01-24
Drop dependency on requirements parser

jenshnielsen:
The feature we are using from there is liberally one line in a call to pkg_resources. We might as well do that directly


astafan8 (approved):

astafan8:
bors merge

PR #3861, opened 2022-01-25
Stop N52xx intercepting keyboard interrupts

spauka:
Since run_sweep prints a warning message on a keyboard interrupt by catching the exception - it makes it difficult to interrupt a sweep on using the PNA. Let's raise the exception after printing the warning.


jenshnielsen (approved):

jenshnielsen:
Bors merge

PR #3863, opened 2022-01-25
add function to get guids by run spec

jenshnielsen:
TODO
- [ ] Add to public api
- [ ] Tests


Dominik-Vogel:
@jenshnielsen this looks good to me. The only thing that I think is worth considering is what we do about the `get_guids_from_runspec`. It seems quite confusing to have two functions with such similar names and such similar actions. I am aware of the fact that only one of them is public api, but also for devs this is very confusing. 
We could have the code for querying and filtering in withing the more general `by_runspec` and let `from_runspec` call that one, as well as deprecating `from_runspec` with the reference to use `by_runspec`. What do you think?

jenshnielsen:
So we want to leave the code for querying in that module but we could rename the get_guids_from_run_spec in that module to something else `_query_guids_from_run_specs` perhaps  and leave get_guids_from_run_spec around in the queries module (but deprecated)

jenshnielsen:
@astafan8 Can you see if this api makes sense. If so I will finish this up with some tests

astafan8 (commented):
yes, the API looks good!

Dominik-Vogel (approved):
@jenshnielsen so we are basically renaming the function because we have connection as a non-keyword argument, right? I think it does make a lot of sense to me, how you implemented this change.

jenshnielsen:
@Dominik-Vogel yes and also to match load_by_id, load_by_run_spec etc

jenshnielsen:
bors merge

PR #3885, opened 2022-02-01
AMI430: Restore ramp rates of individual magnet axes after simultaneous ramp, for block_during_ramp is True or with call to wait_while_all_axes_ramping

astafan8:
Apparently, it is confusing for users that the `simultaneous` ramp mode does not keep the ramp rate settings of the individual magnet axes intact. User's impression of the `simultaneous` ramp mode with the `vector_ramp_rate` parameter is that it sits on top of the individual axes and after the ramp is over, the individual magnet axes should still have their original individual ramp rates preserved.

In the implementation, ``restore_at_exit`` context managers of the ``ramp_rate`` parameters are used. The get exit-ed aka closed in the ``wait_while_all_axes_ramping``, which means that not only users can run the simultanous ramps with ``block_during_ramp`` is ``True`` and the restoring will happen, but also the restoring will happen when users call ``wait_while_all_axes_ramping`` explicitly, even when ``block_during_ramp`` is set to ``False``.

- [x] test on real instrument
- [x] test if after initiating a ramp, it is possible to change ramp rate on the fly; does instrument allow it or errors out, does it change the ramp rate or ignores it, etc

Credits to @jourib for the idea and the help with testing, and @jenshnielsen for refining the implementation!


qcodes/instrument_drivers/american_magnetics/AMI430.py
```diff
@@ -951,33 +956,58 @@ def _update_individual_axes_ramp_rates(
...
-                f"target field to {value} {axis_instrument.field.unit}"
-            )
-            axis_instrument.set_field(value, perform_safety_check=False, block=False)
 
-        if self.block_during_ramp() is True:
-            self.log.debug(f"Simultaneous ramp: blocking until ramp is finished")
-            self.wait_while_all_axes_ramping()
+            if self.block_during_ramp() is True:
+                self.log.debug(f"Simultaneous ramp: blocking until ramp is finished")
+                self.wait_while_all_axes_ramping()
```
jenshnielsen, line 1001: I know this code is just moved but I am surprised to see this log before waiting

docs/examples/driver_examples/Qcodes example with AMI430.ipynb
```diff
@@ -584,6 +584,56 @@
...
+    "\n",
+    "with (\n",
+    "        ix.ramp_rate.restore_at_exit(),\n",
+    "        iy.ramp_rate.restore_at_exit()\n",
+    "        iz.ramp_rate.restore_at_exit()\n",
+    "):\n",
+    "    # Refer to qcodes documentation for more details about\n",
+    "    # the ``restore_at_exit`` context manager of ``Parameter``\n",
+    "\n",
+    "    i3d.ramp_simultaneously(setpoint=FieldVector(0.5, 1.0, 0.01), duration=2)\n",
```
jenshnielsen, line 614: Would it be less error prone to split this into two functions `ramp_simultaneously_blocking / not_blocking` that way its perhaps more clear that one restores and the other does not

qcodes/instrument_drivers/american_magnetics/AMI430.py
astafan8: hmm, that log message was there before, the "blocking until ramp is finished" one. so what is surprising and why?

docs/examples/driver_examples/Qcodes example with AMI430.ipynb
astafan8: for the ramp_simultanously method, probably yes, but the ramp can also be initiated suing cartersian/spherical/etc parameters, so i'm not convinced about making blocking and nonblocking versions of those (certainly not in this PR at least)

qcodes/instrument_drivers/american_magnetics/AMI430.py
jenshnielsen: nevermind misread

JouriB:
I tested the PR, it works as expected.

I also tested what happens when the ramp rate is changed during a ramp started with blocking. The ramp rate is in fact changed on the fly for the active ramp. So restoring the ramp rates for unblocked ramps does not seem to be possible.

image

JouriB (approved):

docs/examples/driver_examples/Qcodes example with AMI430.ipynb
jenshnielsen: Well I still think having the restore of the ramp rate depend on an unrelated setting is a really bad design. Can't we come up with something better

astafan8: let's think :)

astafan8: after an offline discussion, the "something better" is now that restoration happens in the wait_while_all_axes_ramping call.

docs/examples/driver_examples/Qcodes example with AMI430.ipynb
```diff
@@ -638,23 +638,75 @@
...
...`\"simultaneous\"`` changes the ramp rates of the individual magnet axes. The implementation is such that, given the current and the setpoint magnetic field vectors together with either vector ramp rate (``vector_ramp_rate`` parameter) or the desired ramp duration (``ramp_simultaneously`` method), the required ramp rates of the individual axes are calculated and then set via the ``ramp_rate`` parameters of the individual magnet axes. Once the individual ramp rates are set, the ramping is initiated on all axes (almost) at the same time. However, the values of the individual ramp rates can be **restored** by calling ``wait_while_all_axes_ramping``. Note that if ``block_during_ramp`` parameter is set to ``True``, ``wait_while_all_axes_ramping`` method is called automatically by the ramping logic, thus the ramp rates of the individual magnet axes are **restored** after the ramp is finished. If you'd like to use the simultanous ramp with ``block_during_ramp`` parameter set to ``False`` **AND** you'd like the ramp rates of the individual magnet axes to be restored, just call ``wait_while_all_axes_ramping`` at the appropriate moment in your measurement code, see the following example:\n",
```
jenshnielsen, line 641: I would try to refrase this to always encurage users to call ``wait_while_all_axes_ramping`` after doing a non blocking ramp. Something like `this will ensure that the magnet is in a known good state before doing the next operation``

qcodes/tests/drivers/test_ami430.py
```diff
@@ -494,6 +495,170 @@ def test_ramp_rate_exception(current_driver):
         ix.ramp_rate(target_ramp_rate)
 
 
+def test_simultaneous_ramp_mode_does_not_reset_individual_axis_ramp_rates_if_nonblocking_ramp(
+    current_driver, caplog, request
```
jenshnielsen, line 499: Could do the usual rename fixture and set name dance to avoid this warning (for current_driver)

qcodes/tests/drivers/test_ami430_visa.py
```diff
@@ -486,6 +487,170 @@ def test_ramp_rate_exception(current_driver):
...
+        ami3d._instrument_y.ramp_rate.restore_at_exit()
+    )
+    restore_parameters_stack.enter_context(
+        ami3d._instrument_z.ramp_rate.restore_at_exit()
+    )
+
+    restore_parameters_stack.enter_context(ami3d.ramp_mode.set_to("simultaneous"))
+
+    restore_parameters_stack.enter_context(ami3d.block_during_ramp.set_to(False))
+
```
jenshnielsen, line 515: All of the above is to reset the instrument to a known good state after the test right ?

Would it make sense to refactor that out into a context manager

jenshnielsen (approved):

docs/examples/driver_examples/Qcodes example with AMI430.ipynb
astafan8: done

qcodes/tests/drivers/test_ami430.py
astafan8: yes

qcodes/tests/drivers/test_ami430_visa.py
astafan8: most of the above is indeed to reset the instrument, but some of the statements are part of "setting up the instrument into a state required for the test". the ask is fair, but not super critical. (ideally, we'd restore all parameters of the instrument at the teardown of the fixture)

PR #3889, opened 2022-02-02
Drop python 37 support

jenshnielsen:
Most of the scientific python echo system is dropping support for 3.7 so it makes sense for us to do so too.

This is not to be merged until after the 0.35.0 release

todo

- [x] Scipy version specifier


sldesnoo-Delft:
The Keysight PXI systems do not support fully Python > 3.7 when the firmware or drivers are older than December 2021.
In many cases the update of firmware and drivers will require the purchase of new licenses.
I've not yet had time to properly test the December 2021 firmware and drivers.

Some people might have to stay on Python 3.7 for quite some time.
A well defined last release of qcodes supporting 3.7 would be highly appreciated.

jenshnielsen:
@sldesnoo-Delft Thinking a bit about this @astafan8 and I suggest that we push out the drop of python 3.7 support to 0.37 (in June) does that work for you?

sldesnoo-Delft:
That would be fine for me.  So, version 0.37 will not support Python 3.7 anymore. The version number is remarkable. :-)

jenshnielsen:
@sldesnoo-Delft @astafan8 We have missed a few monthly releases so I am going to suggest that the upcoming release that I will make today will be the last to support 3.7

We now have even more upstream packages that do not support 3.7 plus issues like the one that prevents us from merging #4546

.github/workflows/upload_to_pypi.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v4.2.0
       with:
-        python-version: '3.7'
+        python-version: '3.10'
```
astafan8, line 18: not 3.8 ?

qcodes/validators/validators.py
```diff
@@ -12,12 +12,11 @@
 from typing import Dict as TDict
 from typing import Generic, Hashable
 from typing import List as TList
-from typing import Optional
+from typing import Literal, Optional
 from typing import Sequence as TSequence
```
astafan8, line 16: interesting that black/isort dont group all these imports into `from typing import (all the stuff here ....)`

astafan8 (approved):

.github/workflows/upload_to_pypi.yaml
jenshnielsen: No, I tried to stream line this such that the latest is the default and we focus the testing of 3.8 on the old dependencies (then this does not have to change again when we drop 3.8)

qcodes/validators/validators.py
jenshnielsen: I think it's due to the import as ... use

We should probably rework those imports to just use typing.List rather than from typing import typing.List as TList but that is not for here

.github/workflows/upload_to_pypi.yaml
astafan8: ah, cool!

PR #3908, opened 2022-02-09
add ci job that runs darker

jenshnielsen:
pre-commit ci is not that great for running darker since it cannot see the git history. This runs darker as a separate ci job

### DIFF
--- a/.github/workflows/darker.yaml
+++ b/.github/workflows/darker.yaml
@@ -1,0 +1,16 @@
+name: Lint with Darker
+
+on: [push, pull_request]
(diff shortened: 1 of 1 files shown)


astafan8 (approved):
great!

but this does not commit the changes to the branch, right? that's a sad different with precommit ci

jenshnielsen:
No it does not seem to have that feature but the pre-commit ci never triggers on darker so I think this is better than nothing. I suggest that we merge this now but don't make it required until we have seen how it works

PR #3913, opened 2022-02-10
fix pytest7 deprecation warnings

jenshnielsen:
Passing None to pytest.warns is deprecated. Followed https://docs.pytest.org/en/latest/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests to update the code

Also sneaked in some cleanup of imports and fixtures in the b1500 tests since I was there


astafan8 (approved):

jenshnielsen:
bors merge

PR #3914, opened 2022-02-10
Strip attributes from channels when closing an instrument.

jenshnielsen:
Contains the changes from https://github.com/QCoDeS/Qcodes/pull/3912 which should be merged before this.

instrument.close stripes the attributes of an instrument but not of the modules within the instrument. It is unclear to me if it really makes sense to strip the atters but if we do that we should also do it here


astafan8 (approved):

jenshnielsen:
bors merge

jenshnielsen:
bors merge

PR #3916, opened 2022-02-11
Feat: Improved Keysight Infiniium Driver

### LINKED ISSUE #1147: Problem with acquiring long sequences on Infiniium scope
It has been reported that when a score trace with more than 16400 points is acquired, the data array has phantom spikes at every ~16400 points apart. We should investigate this and fix

### LINKED ISSUE #1335: Driver / Infiniium oscilloscope : feature requests
As requested by our friends in Sydney.

They would love:

1. To have a way to use the streaming method for large number of points as it is possible to hit the 1GB limit.
2. For the get trace to return the current trace not to trigger and wait. (don't use DIGITIZE)
3. The sin(x)/x interpolation to be disabled so number of points requested = the number of points returned
4. To have functions to control the state of the scope ie Run, Single, Stop
5. A function to block until the scope has been triggered and acquired its data/number of averages. using ESR/STB polling or equivalent instead of *OPC? and a long timeout as the scope becomes unresponsive if we decide to interrupt the program.
6. For the `trace` command not to change the acquisition mode to real-time. It shouldn't be reset every time `get_raw()` is called, since the scope may be in equivalent-time or high-res mode

@StevenWaddy @Dominik-Vogel

spauka:
This PR adds a new driver for the Keysight Infiniium series of oscilloscopes. It solves several outstanding issues with the previous driver and adds a number of new useful features.

Closes #1147 (uses streaming mode) and closes #1335

~Depends on #3889 since we use the := operator.~ removed this since py37 might be sticking around :D

## Changes and Features
 - Trace parameters are updated to use `ParameterWithSetpoints`.
 - The driver now works "out of the box", removing the need to `prepare_curvedata` before measurement. For example, the following code will now work seamlessly:
```py3
dso = scfg.load_instrument("dso")
qcodes.utils.dataset.doNd.do0d(dso.ch1.trace, dso.ch2.trace)
```
 - Added the option to disable digitize per measurement (`dso.auto_digitize`), similar to how the N52xx driver works. This is useful in two situations:
   - If we just want to acquire exactly what is on screen (e.g. after averaging for a long time or after setting up a complex trigger) without taking a new measurement.
   - If we want to measure with the scope in free-running mode.
   - If we want to acquire from multiple channels at once.
```py3
# Acquire from 4 channels in a single trace
dso = scfg.load_instrument("dso")
dso.auto_digitize(False)
qcodes.utils.dataset.doNd.do0d(dso.digitize, dso.ch1.trace, dso.ch2.trace, dso.ch3.trace, dso.ch4.trace)
```
 - Added the option to cache setpoints/offsets for when measurements are taken in a loop. For example, we can do the following:
```py3
# Acquire in a loop but cache oscilloscope parameters to minimize measurement time
dso = scfg.load_instrument("dso")
dac = scfg.load_instrument("dac")
dso.auto_digitize(False)
dso.cache_setpoints(True)
qcodes.utils.dataset.doNd.do1d(
    dac.ch01.voltage, 0, 1, 101, 0.01, 
    dso.digitize, dso.ch1.trace, dso.ch2.trace,
    enter_actions=(dso.update_all_setpoints,)
)
```
 - There is now a measurement subsystem tied to each channel so it's possible to read from multiple channels in a single measurement:
```py3
dso = scfg.load_instrument("dso")
qcodes.utils.dataset.doNd.do0d(dso.ch1.measure.vrms, dso.ch2.measure.vrms)
```
 - The `digitize` method is now able to handle long acquisitions without timing out. This is useful when sampling a particularly long process, one that may be longer than the timeout set on the device.
 - Oscilloscope capabilities (bandwidth, memory depth, sample rate) are queried and stored in the snapshot, rather than assumed. These values can now also be set using parameters. (see `dso._query_capabilities`)
 - Added support for retrieving functions from scope (although not for configuring them in code). Both FFT and normal functions are supported. It is now possible to do something like:
```py3
dso = scfg.load_instrument("dso")
qcodes.utils.dataset.doNd.do0d(dso.func1.trace)
```

## Non-backwards compatible changes
 - ~`prepare_curvedata` is removed. This could potentially be replaced with a stub that does nothing to maintain backwards compatibility?~ Added stub
 - Parameter `trigger_enabled` is removed. This parameter did not do what it said in the docstring and it's presence is misleading. (It actually enabled trigger sequence mode!)
 - Parameter `data_source` removed. This was causing a warning on startup and is not useful for users anyway.
 - `acquire_timespan` removed. It is a derived parameter that won't be correct for math mode.

## Testing
The driver has been tested on a MSOS254A.

@jenshnielsen @astafan8 What do you think? If it is too large a change we could also split these changes into a new file as was done with the 34xxx DMMs?

## Outstanding items

- [x] Update documentation for this scope
- [x] Add math function support.
- [x] Add support for pulling FFT data off the scope.


qcodes/instrument_drivers/Keysight/Infiniium.py
```diff
@@ -1,646 +1,1102 @@
...
+            self.points,
+            endpoint=False,
+        )
 
 
-class RawTrace(ArrayParameter):
+class DSOFrequencyAxisParam(Parameter):
     """
-    raw_trace will return a trace from OSCIL
+    Time axis parameter for the Infiniium series DSO.
```
jenshnielsen, line 46:
```suggestion
    Frequency axis parameter for the Infiniium series DSO.
```

qcodes/instrument_drivers/Keysight/Infiniium.py
```diff
@@ -1,646 +1,1102 @@
...
-                         setpoint_names=('Time',),
-                         setpoint_labels=(
-                             f'Channel {channel} time series',),
-                         setpoint_units=('s',),
-                         docstring='raw trace from the scope',
-                         instrument=instrument
-                         )
+    def __init__(self, xorigin: float, xincrement: float, points: int, **kwargs: Any):
+        """
+        Initialize time axis. If values are unknown, they can be initialized to zero and
```
jenshnielsen, line 51:
```suggestion
        Initialize frequency axis. If values are unknown, they can be initialized to zero and
```

qcodes/instrument_drivers/Keysight/Infiniium.py
```diff
@@ -1,646 +1,1103 @@
...
+        ###################################
+        # Voltage Parameters
+        self.add_parameter(
+            name="amplitude",
+            label="Voltage amplitude",
+            get_cmd=self._create_query("VAMP"),
+            get_parser=float,
+            unit="V",
+            snapshot_value=False,
+        )
```
jenshnielsen, line 240: Since you are rewriting the driver you may want to consider using the new form of parameter declaration

As in this example

https://qcodes.github.io/Qcodes/examples/writing_drivers/Creating-Instrument-Drivers.html#VisaInstrument:-Simple-example

This has the benefits that tap completion works correctly for parameters and the parameters can be documented as attributes. See https://qcodes.github.io/Qcodes/api/generated/qcodes.instrument_drivers.weinschel.html#qcodes.instrument_drivers.weinschel.Weinschel_8320.Weinschel_8320.attenuation for an example

spauka: Hey @jenshnielsen - implemented this in a8f49ad

Unfortunately this does cause a couple of new mypy errors to appear, since the results of a parameter get (which are used in a couple of places) are not defined...

For example, Infiniium.py#115 now fails the mypy check, since we query the parameter `dso.funcX.function` to figure out whether we are measuring an FFT.

Any suggestions on how to fix/correctly type annotate parameters?

jenshnielsen: I think https://github.com/QCoDeS/Qcodes/pull/3935 should resolve that by clarifying the difference of get and set. The optional part is for set.

jenshnielsen: There is a remaining issue of the form
```
qcodes\instrument_drivers\Keysight\Infiniium.py:109:31: error: Cannot determine type of "cache_setpoints"  [has-type]
qcodes\instrument_drivers\Keysight\Infiniium.py:190:16: error: Cannot determine type of "cache_setpoints"  [has-type]
qcodes\instrument_drivers\Keysight\Infiniium.py:200:12: error: Cannot determine type of "auto_digitize"  [has-type]
```

It looks like mypy cannot always infer the type of the attribute correctly
This can be mitigated by doing:

image

jenshnielsen: We could also ignore the error on those lines. Hopefully this bug in mypy will eventually be resolved

jenshnielsen: @astafan8 Yes but that does not fix the issue. An alternative could be to just ignore the mypy warning

jenshnielsen: @astafan8 `Cannot determine type of` does not mean mypys need a type annotation here but something more like mypy cannot figure out what has happened here and is giving up. Adding a type annotation does not help

astafan8: ah, that's indeed weird and unexpected.

jenshnielsen: @astafan8 Actually it does seem like you are right 

Doing

self.cache_setpoints: Parameter = Parameter(...)

does fix the issue.

jenshnielsen: I have created a pr here https://github.com/spauka/Qcodes/pull/18 that makes it type check against master. Note that you need to also rebase or merge with master for that to work

spauka: Success ♥️

Thanks for the suggestions for how to solve. I'm surprised that the type annotation of `self.cache_setpoints` worked given that it's literally during the definition of the parameter as a `Parameter` 🤣

jenshnielsen (approved):
bors merge

PR #3918, opened 2022-02-11
Add types to utils.command

jenshnielsen:
Input types are hard to type until we have general support for variadic types or paramspec the input types are hard to type better since there can be any number of input arguments. 

I don't currently see a good way to get the call command to type cleanly. 

Locally this fails for me with 

```
❯ mypy .\qcodes
qcodes\utils\command.py:153:16: error: Incompatible return value type (got "ParsedOutput", expected "ParsedOutput")  [return-value]
qcodes\utils\command.py:161:16: error: Incompatible return value type (got "ParsedOutput", expected "ParsedOutput")  [return-value]
qcodes\utils\command.py:170:16: error: Incompatible return value type (got "ParsedOutput", expected "ParsedOutput")  [return-value]
qcodes\utils\command.py:177:16: error: Incompatible return value type (got "ParsedOutput", expected "ParsedOutput")  [return-value]
qcodes\utils\command.py:185:16: error: Incompatible return value type (got "ParsedOutput", expected "ParsedOutput")  [return-value]
qcodes\utils\command.py:193:16: error: Incompatible return value type (got "ParsedOutput", expected "ParsedOutput")  [return-value]
Found 6 errors in 1 file (checked 474 source files)
```
No clue what that is supposed to mean


jenshnielsen:
Fumbling around with reveal_type 

I see the following 

```
qcodes\utils\command.py:153:21: note: Revealed type is "def (Output`-1) -> ParsedOutput`-2"
qcodes\utils\command.py:155:21: note: Revealed type is "ParsedOutput`-2"
qcodes\utils\command.py:157:21: note: Revealed type is "ParsedOutput`-1"
```
comparing the type in the called function its output and the type of a random variable assigned in the method. 

Not sure what those modifiers mean

qcodes/utils/command.py
```diff
@@ -45,26 +51,35 @@ class Command:
...
+        if no_cmd_function is not None and not is_function(no_cmd_function, arg_count):
+            raise TypeError(
+                f"no_cmd_function must be none or a function "
+                f"taking the same args as the command, not "
+                f"{no_cmd_function}"
+            )
 
         if input_parser is None:
-            parse_input = False
+            parse_input: Union[bool, str] = False
```
astafan8, line 74:
```suggestion
            parse_input: Union[bool, Literal["multi"]] = False
```

jenshnielsen:
The issue can be reproduces with a minimum example like the following 

```python
from typing import Any, Callable, TypeVar

Output = TypeVar("Output")
ParsedOutput = TypeVar("ParsedOutput")


class Command:

    def __init__(
        self,
        cmd: Callable[..., Output],
        output_parser: Callable[[Output], ParsedOutput],
    ):

        self._cmd = cmd
        self.output_parser = output_parser

    def call_cmd_parsed_out(self, *args: Any) -> ParsedOutput:
        """Execute a function with output parsing."""
        return self.output_parser(self._cmd(*args))

```

jenshnielsen:
Aha this is just a missing generic

This typechecks 

```
from typing import Any, Callable, TypeVar, Generic

Output = TypeVar("Output")
ParsedOutput = TypeVar("ParsedOutput")


class Command(Generic[Output, ParsedOutput]):

    def __init__(
        self,
        cmd: Callable[..., Output],
        output_parser: Callable[[Output], ParsedOutput],
    ):

        self._cmd = cmd
        self.output_parser = output_parser

    def call_cmd_parsed_out(self, *args: Any) -> ParsedOutput:
        """Execute a function with output parsing."""
        return self.output_parser(self._cmd(*args))
```

qcodes/utils/command.py
```diff
@@ -73,11 +90,15 @@ def __init__(self, arg_count, cmd=None, exec_str=None, input_parser=None,
             self.output_parser = output_parser
         else:
             raise TypeError(
-                'output_parser must be a function with one arg,'
-                ' not {}'.format(repr(output_parser)))
+                f"output_parser must be a function with one arg,"
```
jenshnielsen, line 93: or None

Dominik-Vogel:
@jenshnielsen allow me to interpret:
The logic behind all this was:
- the MyPy error said: here are two types that are not equivalent, but they are both called `ParsedOutput`
- the reason for that was that they were simply type variables
- the type variables were not linked to anything so that `call_cmd_parsed_out -> ParsedOutput` was actually equivalent to `call_cmd_parsed_out -> Any`
- by making it a generic class the type vars of `call_cmd_parsed_out` and `__init__` got linked from a type perspective and did not only have the same name.

jenshnielsen:
Thanks for the explanation you should contribute that to the mypy docs :)

Dominik-Vogel (approved):
Beautiful!

jenshnielsen:
bors merge

PR #3919, opened 2022-02-11
make channeltuple/list generic in the channeltype

jenshnielsen:
This meas that in an instrument you will get a type such as ChannelTuple[MyInstrumentModuleSubclass] rather than just ChannelTuple. That in turn allows you to type safe call methods that are only defined on that type after taking a channel out of the channelTuple.

### DIFF
--- a/qcodes/instrument/channel.py
+++ b/qcodes/instrument/channel.py
@@ -4,12 +4,14 @@ from typing import (
     Any,
     Callable,
     Dict,
+    Generic,
     Iterable,
     List,
     MutableSequence,
     Optional,
     Sequence,
     Tuple,
+    Type,
     TypeVar,
     Union,
     cast,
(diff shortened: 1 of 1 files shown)


qcodes/instrument/channel.py
```diff
@@ -146,10 +156,9 @@ def full_names(self) -> Tuple[str, ...]:
         return self.names
 
 
-T = TypeVar("T", bound="ChannelTuple")
-
-
-class ChannelTuple(Metadatable, Sequence[InstrumentChannel]):
+class ChannelTuple(
+    Metadatable, Sequence[InstrumentModuleType], Generic[InstrumentModuleType]
+):
```
jenshnielsen, line 161: Not sure this Generic is needed since the Sequence is already generic

astafan8 (approved):

PR #3920, opened 2022-02-12
explicitly import collections.abc

jenshnielsen:
This silences a warning from pyright/pycharm about there not being a module abc in collections.

I suspect the automatic import of acb only works because of the deprecated import of abc members into the main module so its probably better this way


Dominik-Vogel (approved):

jenshnielsen:
bors merge

jenshnielsen:
bors merge

PR #3929, opened 2022-02-15
Promote unit of parameter to be a property

jenshnielsen:
Inspired by the typing hacks needed in https://github.com/QCoDeS/Qcodes/pull/3916 to make unit a property 

This eliminates the need for those hacks See https://github.com/QCoDeS/Qcodes/pull/3916/files#diff-5478d126afe3357045220d8be1ef9c6859ee8427f28984ff8d169b9ab1b75902R126

- [x] Should we do this for other _BaseParameter subclasses. No need

### DIFF
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -1304,6 +1304,14 @@ class Parameter(_BaseParameter):
                 '',
                 self.__doc__))
 
+    @property
+    def unit(self) -> Optional[str]:
+        return self._unitval
+
+    @unit.setter
+    def unit(self, unit: Optional[str]) -> None:
+        self._unitval = unit
+
     def __getitem__(self, keys: Any) -> 'SweepFixedValues':
         """
         Slice a Parameter to get a SweepValues object
--- a/qcodes/instrument/parameter.py
+++ b/qcodes/instrument/parameter.py
@@ -1275,8 +1275,9 @@ class Parameter(_BaseParameter):
 
         #: Label of the data used for plots etc.
         self.label: str = name if label is None else label
-        #: The unit of measure. Use ``''`` for unitless.
+
         self.unit = unit if unit is not None else ''
+        self._unitval: str
 
         if initial_value is not None and initial_cache_value is not None:
             raise SyntaxError('It is not possible to specify both of the '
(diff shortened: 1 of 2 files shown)


qcodes/instrument/parameter.py
```diff
@@ -1295,6 +1295,14 @@ def __init__(
                 '',
                 self.__doc__))
 
+    @property
+    def unit(self) -> Optional[str]:
```
astafan8, line 1299: actually unit cannot be None, see line 1270
```suggestion
    def unit(self) -> str:
```

astafan8 (commented):
making a property is nice, but if it's only about typing, would annotating the ``unit`` attribute in `__init__` where it's defined suffice?

about adding unit property to _BaseParameter - i don't think it's particularly useful since most of the "useful" parmaeters inherit from Parameter, and the only two that inherit from _BaseParameter are Array and Multi parameter , where the first one is sort of deprecated and the latter one has concept of units with s as opposed to unit.

jenshnielsen:
> making a property is nice, but if it's only about typing, would annotating the unit attribute in __init__ where it's defined suffice?

The reason for doing this it to allow a subclass to do the same without having to ignore type checking errors. See https://github.com/QCoDeS/Qcodes/pull/3916/files#diff-5478d126afe3357045220d8be1ef9c6859ee8427f28984ff8d169b9ab1b75902R126

astafan8 (approved):

qcodes/instrument/parameter.py
```diff
@@ -1295,6 +1296,18 @@ def __init__(
                 '',
                 self.__doc__))
 
+    @property
+    def unit(self) -> str:
```
astafan8, line 1300: needs a news fragment

qcodes/instrument/parameter.py
```diff
@@ -1266,8 +1266,9 @@ def __init__(
 
         #: Label of the data used for plots etc.
         self.label: str = name if label is None else label
-        #: The unit of measure. Use ``''`` for unitless.
+
         self.unit = unit if unit is not None else ''
+        self._unitval: str
```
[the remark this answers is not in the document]
jenshnielsen: Why?

qcodes/instrument/parameter.py
jenshnielsen: I honestly dont think so. From the end users point of view there is zero difference. But sure we can add it

astafan8: ah, you're right. well, there is a difference for those implementing subclasses of Parameter :)

qcodes/instrument/parameter.py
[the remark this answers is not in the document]
astafan8: readability bike shedding, rightfully ignore :)

qcodes/instrument/parameter.py
jenshnielsen: added one anyway

astafan8 (approved):

PR #3937, opened 2022-02-17
_SetParamContext: explicitly annotate Parameter to work around mypy issue

jenshnielsen:
This is the same issue as showed up in #3916 Here we work around it in the same way by explicitly annotating the member attribute


astafan8 (approved):

jenshnielsen:
bors merge

PR #3938, opened 2022-02-17
Config: Cleanup typing of dotdict

jenshnielsen:
* Remove ignore by allowing any object to be passed to contains (and return False if not a str)
* Make explicit in type annotation that keys must be str
* Take a generic mapping as input to constructor


astafan8 (approved):

jenshnielsen:
bors merge

PR #3944, opened 2022-02-18
Correct calculation of time in interactive widget

jenshnielsen:
Total time is end - start not the other way around
Also rewrite code to avoid type ignores


astafan8 (approved):
bors merge

PR #3945, opened 2022-02-18
Fix loading of >1 dim netcdf datasets from netcdf to DataSetinmem

jenshnielsen:
The wrong from of meshgrid was being used

- [x] Newsfragment
- [x] Test for 3d?

### DIFF
--- a/qcodes/dataset/data_set_cache.py
+++ b/qcodes/dataset/data_set_cache.py
@@ -315,8 +315,8 @@ def _merge_data(existing_data: Mapping[str, np.ndarray],
 
     if not set(new_data.keys()).issubset(set(existing_data.keys())):
         raise RuntimeError(
-            "Trying to add unxpected key to cache."
-            "The following keys were unexpected."
+            "Trying to add unexpected key to cache."
+            "The following keys were unexpected: "
             f"{set(new_data.keys() - existing_data.keys())}"
         )
 
--- a/qcodes/dataset/data_set_cache.py
+++ b/qcodes/dataset/data_set_cache.py
@@ -300,6 +300,9 @@ def _merge_data(existing_data: Mapping[str, np.ndarray],
 
     subtree_merged_data = {}
     subtree_parameters = existing_data.keys()
+
+    assert set(existing_data.keys()) == set(new_data.keys())
+
     new_write_status: Optional[int]
     single_param_merged_data, new_write_status = _merge_data_single_param(
         existing_data.get(meas_parameter),
--- a/qcodes/dataset/data_set_cache.py
+++ b/qcodes/dataset/data_set_cache.py
@@ -465,6 +465,7 @@ class DataSetCacheWithDBBackend(DataSetCache["DataSet"]):
         if self._dataset.completed:
             self._loaded_from_completed_ds = True
 
+        self._data = self.rundescriber.interdeps._empty_data_dict()
         (
             self._write_status,
             self._read_status,
--- a/qcodes/dataset/data_set_cache.py
+++ b/qcodes/dataset/data_set_cache.py
@@ -313,7 +313,12 @@ def _merge_data(existing_data: Mapping[str, np.ndarray],
     subtree_merged_data = {}
     subtree_parameters = existing_data.keys()
 
-    assert set(new_data.keys()).issubset(set(existing_data.keys()))
+    if not set(new_data.keys()).issubset(set(existing_data.keys())):
+        raise RuntimeError(
+            "Trying to add unxpected key to cache."
+            "The following keys were unexpected."
+            f"{set(new_data.keys() - existing_data.keys())}"
+        )
 
     new_write_status: Optional[int]
     single_param_merged_data, new_write_status = _merge_data_single_param(
--- a/qcodes/dataset/data_set_cache.py
+++ b/qcodes/dataset/data_set_cache.py
@@ -85,6 +85,12 @@ class DataSetCache(Generic[DatasetType]):
         return self._data
 
     def prepare(self) -> None:
+        """
+        Set up the internal datastructure of the cache.
+        Must be called after the dataset has been setup with
+        interdependencies but before data is added to the dataset.
+        """
+
         if self._data == {}:
             self._data = self.rundescriber.interdeps._empty_data_dict()
         else:
--- a/qcodes/dataset/data_set_cache.py
+++ b/qcodes/dataset/data_set_cache.py
@@ -84,6 +84,12 @@ class DataSetCache(Generic[DatasetType]):
 
         return self._data
 
+    def prepare(self) -> None:
+        if self._data == {}:
+            self._data = self.rundescriber.interdeps._empty_data_dict()
+        else:
+            raise RuntimeError("Cannot prepare a cache that is not empty")
+
     def load_data_from_db(self) -> None:
         """
         Load the data from an on-disk format in case the cache is not live
@@ -468,7 +474,7 @@ class DataSetCacheWithDBBackend(DataSetCache["DataSet"]):
         if self._dataset.completed:
             self._loaded_from_completed_ds = True
         if self._data == {}:
-            self._data = self.rundescriber.interdeps._empty_data_dict()
+            self.prepare()
         (
             self._write_status,
             self._read_status,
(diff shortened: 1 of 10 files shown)


qcodes/tests/dataset/test_dataset_in_memory.py
```diff
@@ -43,6 +46,53 @@ def test_dataset_in_memory_reload_from_db(
...
+    ) as datasaver:
+        for set_v in np.linspace(0, 25, shape1):
+            for set_v2 in np.linspace(0, 100, shape2):
+                DAC.ch1.set(set_v)
+                DAC.ch2.set(set_v2)
+                get_v = DMM.v1()
+                datasaver.add_result(
+                    (DAC.ch1, set_v), (DAC.ch2, set_v2), (DMM.v1, float(i))
+                )
+                i = i + 1
```
astafan8, line 78: unused, right? 
```suggestion
```

astafan8 (commented):
> Test for 3d?

yes, would be nice :)

qcodes/tests/dataset/test_dataset_in_memory.py
jenshnielsen: No we are inserting a counter into the data rather then the output from the instrument to make sure that the data is written and read in the expected order

qcodes/dataset/descriptions/dependencies.py
```diff
@@ -257,6 +269,20 @@ def _to_dict(self) -> InterDependencies_Dict:
...
+        for key, value in self.dependencies.items():
+            ps_id = self._paramspec_to_id[key]
+            output[ps_id] = {ps_id: np.array([])}
+            dep_ids = [self._paramspec_to_id[ps] for ps in value]
+            for dep_id in dep_ids:
+                output[ps_id][dep_id] = np.array([])
+        standalones = [self._paramspec_to_id[ps] for ps in self.standalones]
+        for standalone in standalones:
+            output[standalone] = {}
+            output[standalone][standalone] = np.array([])
```
jenshnielsen, line 283: Should add a test that the cache keys are the same before and after inserting data into the cache. Especially the standalone keys should be covered. Also check for different datatypes (e.g. the empty arrays here have a default type which should not matter)

qcodes/dataset/descriptions/dependencies.py
```diff
@@ -257,6 +269,20 @@ def _to_dict(self) -> InterDependencies_Dict:
                                           "standalones": standalones}
         return output
 
+    def _empty_data_dict(self) -> Dict[str, Dict[str, np.ndarray]]:
```
jenshnielsen, line 272: Docstring

qcodes/tests/dataset/test_dataset_in_memory.py
```diff
@@ -43,6 +46,53 @@ def test_dataset_in_memory_reload_from_db(
...
+    )
+    i = 0
+    with meas_with_registered_param_2d.run(
+        dataset_class=DataSetType.DataSetInMem
+    ) as datasaver:
+        for set_v in np.linspace(0, 25, shape1):
+            for set_v2 in np.linspace(0, 100, shape2):
+                DAC.ch1.set(set_v)
+                DAC.ch2.set(set_v2)
+                get_v = DMM.v1()
```
jenshnielsen, line 74: remove

qcodes/dataset/descriptions/dependencies.py
jenshnielsen: Considder refactoring the logic that preallocates the cache into prepare too so we can simplify the conditions that cache update must run under

qcodes/dataset/descriptions/dependencies.py
jenshnielsen: Added tests of cache of standalones and an assert to trigger if non expected parameters 
are added to the cache

qcodes/dataset/descriptions/dependencies.py
jenshnielsen: Added a prepare method to the cache. Will leave the work of refactoring allocation into this method to a different pr

jenshnielsen:
@astafan8 ready for a first review I think. Will add a newsfragment before merging

qcodes/dataset/descriptions/dependencies.py
```diff
@@ -257,6 +269,26 @@ def _to_dict(self) -> InterDependencies_Dict:
...
+
+        output: Dict[str, Dict[str, np.ndarray]] = {}
+        for key, value in self.dependencies.items():
+            ps_id = self._paramspec_to_id[key]
+            output[ps_id] = {ps_id: np.array([])}
+            dep_ids = [self._paramspec_to_id[ps] for ps in value]
+            for dep_id in dep_ids:
+                output[ps_id][dep_id] = np.array([])
+        standalones = [self._paramspec_to_id[ps] for ps in self.standalones]
+        for standalone in standalones:
```
astafan8, line 287: suggest 1) using paramspecs directly since ``.name`` does the job, while the notion of the "id" is less readable (for some historical reason that i already forgot), 2) more explicit variable names for readability of the logic
```suggestion
        for dependent, independents in self.dependencies.items():
            dependent_name = dependent.name
            output[dependent_name] = {dependent_name: np.array([])}
            for independent in independents:
                output[dependent_name][independent.name] = np.array([])
        for standalone in (ps.name for ps in self.standalones):
```

qcodes/tests/dataset/test_dataset_in_memory.py
```diff
@@ -43,6 +46,124 @@ def test_dataset_in_memory_reload_from_db(
...
+
+    paramspecs = ds.get_parameters()
+    assert len(paramspecs) == 3
+    assert paramspecs[0].name == "dummy_dac_ch1"
+    assert paramspecs[1].name == "dummy_dac_ch2"
+    assert paramspecs[2].name == "dummy_dmm_v1"
+
+    # if the indexes are not correct here it will break the exported
+    # xarray and therefor netcdf order below and therefor the loaded data
+    # will have the coordinates inverted.
```
astafan8, line 89:
```suggestion
    # if the indexes (their order) are not correct here, the exported xarray, and thus
    # the exported netcdf will have a wrong order of axes in the data, so that
    # the loaded data will have the coordinates inverted. Hence we assert that
    # the order is exactly the same as declared via Measurement.register_parameter
    # calls above
```

qcodes/tests/dataset/test_dataset_in_memory.py
```diff
@@ -43,6 +46,124 @@ def test_dataset_in_memory_reload_from_db(
...
+    paramspecs = ds.get_parameters()
+    assert len(paramspecs) == 4
+    assert paramspecs[0].name == "dummy_dac_ch1"
+    assert paramspecs[1].name == "dummy_dac_ch2"
+    assert paramspecs[2].name == "dummy_dac_ch3"
+    assert paramspecs[3].name == "dummy_dmm_v1"
+
+    # if the indexes are not correct here it will break the exported
+    # xarray and therefor netcdf order below and therefor the loaded data
+    # will have the coordinates inverted.
```
astafan8, line 151:
```suggestion
    # if the indexes (their order) are not correct here, the exported xarray, and thus
    # the exported netcdf will have a wrong order of axes in the data, so that
    # the loaded data will have the coordinates inverted. Hence we assert that
    # the order is exactly the same as declared via Measurement.register_parameter
    # calls above
```

astafan8 (approved):
awesome!

qcodes/dataset/data_set_cache.py
```diff
@@ -299,7 +311,10 @@ def _merge_data(existing_data: Mapping[str, np.ndarray],
                 ) -> Tuple[Dict[str, np.ndarray], Optional[int]]:
 
     subtree_merged_data = {}
-    subtree_parameters = set(existing_data.keys()) | set(new_data.keys())
+    subtree_parameters = existing_data.keys()
+
+    assert set(new_data.keys()).issubset(set(existing_data.keys()))
```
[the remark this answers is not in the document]
jenshnielsen: Should probably just convert this to a proper error.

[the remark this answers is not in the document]
jenshnielsen: To prevent this from ever happening we should probably require that prepare is always called on the dataset. (see the last tests that I just had to fix since they did not call prepare but the lower lying api)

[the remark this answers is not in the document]
jenshnielsen: Rewrote this so cache is prepared at the same step that marks the dataset non pristine so we avoid additional logic

PR #3961, opened 2022-02-25
Add initialised_database_at to public API

Dominik-Vogel:
I think it would make sense to add the initialised_database_at-context manager to the public api. A typical usage pattern would be to use it with `load_by_guid` or similar.


jenshnielsen (approved):
agreed

PR #3971, opened 2022-03-01
Dataset/add initialised database at to dataset api

jenshnielsen:
To match the context manager being added to qcodes top level api.

Also declare __all__ rather than disabling warning since that will correctly mark these apis as reexported by various tools.
@Dominik-Vogel


jenshnielsen:
bors merge

PR #3981, opened 2022-03-02
Instrument/agilent fix resolution

jenshnielsen:
Fixes #3979 

The new form should have the instrument passed explicitly. That was missing here. Add a better error message along with some tests

PR #3982, opened 2022-03-02
Instrument/make parameter a required arg

jenshnielsen:
This means that issues such as the ones spotted by https://github.com/QCoDeS/Qcodes/pull/3981 will be caught by mypy.
I expect the impact on end users to be fairly limited since most parameters are constructed by add_parameter But obviously this is a breaking changes so we need to thing if we want to do this

### DIFF
--- a/docs/changes/newsfragments/3981.improved_driver
+++ b/docs/changes/newsfragments/3981.improved_driver
@@ -1,0 +1,1 @@
+Fixed a typo that prevented initialization of Agilent 34400A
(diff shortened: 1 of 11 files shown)


jenshnielsen:
Rather than doing this we could also considder having add_parameter return the instance of the parameter  so that one 
could do

```
self.myparameter = self.add_parameter("myparameter", ....) 
"""Docstring of my totally cool parameter"""
```

This would make porting from the old form easier and less error prone,
To get the same docs rendering we would need to extend the sphinx extension to also understand this form

jenshnielsen:
Closed in favor of #4412

PR #3985, opened 2022-03-03
Top level notebooks: Import from modules

jenshnielsen:
This is part of my attempt to make modules more consistent and truly modular. 
The end goal is to ensure that end users can import every thing from separate modules (dataset, instrument ...) 
This is small step in that direction


astafan8 (approved):

jenshnielsen:
bors merge

PR #3995, opened 2022-03-04
Disable deadlines for ami tests

jenshnielsen:
These tests fails occasionally in CI due to timeouts so lets drop these

### DIFF
--- a/qcodes/tests/drivers/test_ami430.py
+++ b/qcodes/tests/drivers/test_ami430.py
@@ -193,7 +193,11 @@ def test_instantiation_from_badly_typed_argument(
 
 
 @given(set_target=random_coordinates["cartesian"])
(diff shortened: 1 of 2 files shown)


qcodes/tests/drivers/test_ami430.py
```diff
@@ -193,7 +193,11 @@ def test_instantiation_from_badly_typed_argument(
 
 
 @given(set_target=random_coordinates["cartesian"])
-@settings(max_examples=10, suppress_health_check=(HealthCheck.function_scoped_fixture,))
+@settings(
+    max_examples=10,
+    suppress_health_check=(HealthCheck.function_scoped_fixture,),
+    deadline=None,
```
astafan8, line 199: do we need to also drop the deadlines in the ami430 VISA tests? or are those "fine"?

astafan8 (approved):

jenshnielsen:
Bors merge

PR #4008, opened 2022-03-09
Remove deprecated ZI driver

jenshnielsen:
With the release of ZI-Qcodes ~ 0.3 all instruments are supported so this deprecated driver can safely be removed ?


qcodes/instrument_drivers/ZI/__init__.py
```diff
@@ -1 +0,0 @@
-# empty __init__ file
```
astafan8, line 1: add a newsfragment?

astafan8 (approved):

astafan8 (approved):

PR #4018, opened 2022-03-14
Make use of version in git 1.1.0

jenshnielsen:
This enables us to use a cmdclass to overwrite the definition of ``__version__`` when building an s or b dist similar to how versioneer does it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = [
     "setuptools >= 56.0",
     "wheel >= 0.29.0",
-    "versioningit ~= 0.3.0",
+    "versioningit ~= 1.1.0",
 ]
 build-backend = 'setuptools.build_meta'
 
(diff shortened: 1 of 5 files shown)


astafan8 (approved):
so now at s and b build time, the new _version.py file will be "build"/"evaluated" so that in the built package it contains just the __version__ variable with the value, right?

sweet :) feature upstreaming in action ;)

jenshnielsen:
> 

> so now at s and b build time, the new _version.py file will be "build"/"evaluated" so that in the built package it contains just the `__version__` variable with the value which makes the import of an installed qcodes faster, right?
> 
> sweet :) feature upstreaming in action ;)

Yes almost the function will still be left in place but the __version__ = _get_version() line will be overwritten with ``__version__ = "x.y.z"`` when generating an sdist. 

That is why I wrapped all the imports in a local function so its not called unless needed.

PR #4020, opened 2022-03-14
Avoid setup.py when using cmd_classes

jenshnielsen:
This is wip and requires an experimental branch of versioningit but seems to be working ok. Not to be merged before that has been approved upstream


jenshnielsen:
After the latest changes to versioningit this needs a rething

PR #4039, opened 2022-03-23
remove the need for InstrumentABC by using a protocol

jenshnielsen:
As recommended by mypy. This avoids having to subclass from abcmeta when defining the metaclass of the instrument and 
overall makes the inheritance a bit simpler. See https://mypy.readthedocs.io/en/latest/more_types.html#mixin-classes


astafan8 (approved):
neat!

PR #4049, opened 2022-03-28
[WIP] migrate setuptools config to pyproject.toml

jenshnielsen:
Setuptools 61 added experimental support for storring config in pyproject.toml

- [ ] Wait for it to be marked non experimental
- [ ] Verifty how this works with older versions of pip and make sure we handle that reasonably
- [ ] Check that editable/normal install from source works as expected
- [ ] Verify that package content is idential

This is generated using ini2toml 
```
ini2toml.exe .\setup.cfg  -o test.toml
```
and then coping to pyproject.toml. 

Only manual change was a need to change the format of comments inside the list of dependencies


jenshnielsen:
Support for project secton in pyproject.toml has been declared as stable so we could imho move that part from setup.cfg now. 

https://setuptools.pypa.io/en/latest/history.html

pyproject.toml
```diff
@@ -160,6 +160,28 @@ filterwarnings = [
     "ignore:SelectableGroups dict interface is deprecated:DeprecationWarning"
 ]
 
+[tool.setuptools]
+zip-safe = false
```
jenshnielsen, line 164: https://setuptools.pypa.io/en/latest/deprecated/zip_safe.html

This line can just be removed

pyproject.toml
```diff
@@ -160,6 +160,28 @@ filterwarnings = [
...
+    "monitor/dist/css/*",
+    "configuration/*.json",
+    "instrument/sims/*.yaml",
+    "tests/dataset/fixtures/2018-01-17/*/*",
+    "tests/delegate/data/*.yml",
+    "tests/drivers/auxiliary_files/*",
+    "py.typed",
+    "dist/schemas/*",
+    "dist/tests/station/*",
+]
```
jenshnielsen, line 183: If we move this to MANIFEST.in we can set include-package-data to True (The default) and remove that line too

astafan8 (approved):

PR #4058, opened 2022-03-30
use next_version for prerelease version number

jenshnielsen:
This mens that a prerelease will correctly be identifed as > current relase


astafan8 (approved):

jenshnielsen:
bors merge

PR #4074, opened 2022-04-04
Feature/Measurement loop

nulinspiratie:
This pull request introduces a new way of doing measurements, called the `MeasurementLoop`, that sits somewhere in between the current `Measurement` class and the `do1d/do2d/dond` functions. The key idea of this measurement scheme is to offer flexibility in measurements while having as much as possible be implicit.

The `MeasurementLoop` introduced here is ported from the UNSW fork `Measurement` class.

This PR is still very much a work in progress, but I thought of starting the PR now. Is this something that could fit into QCoDeS? 
Tagging @jenshnielsen @WilliamHPNielsen @RasmusBC59 

**Features**
- Implicit setpoints using `Sweep` class
- Implicit registering of measurements the first time a parameter is measured
- Allows execution of arbitrary python code in measurements
- Masking of parameters/attributes within a loop/measurement
- Sweep/measure values without needing it to be a parameter
- Allows nested measurements.  
  Nested measurements recognize that they are nested and attach themselves to the parent measurement.  
  This enables functions that perform a measurement. If the function is called within a measurement, it becomes part of the larger measurement

**Limitations**
- Currently incompatible with measurements without a predefined array shape (e.g. keep measuring until X).  
  This can be implemented.
- Creating the dataset only happens once the measurement is finished.  
  The problem is that once the dataset is created no new parameters can be included. I'm looking into ways of circumventing this, if anyone has a suggestion please tell!

**Illustrative example**
```python
from qcodes import ManualParameter
from qcodes.utils.dataset.doNd import LinSweep
from qcodes.dataset.measurement_loop import MeasurementLoop, Sweep

# Initialize parameters
# Note that we do not need to register them in the measurement
p1_set = ManualParameter('outer_sweep_parameter')
p2_set = ManualParameter('inner_sweep_parameter')
p1_get = ManualParameter('parameter_get')
p_sum_above_ten = ManualParameter('sum_above_ten')

with MeasurementLoop('test') as msmt:
   # We encapsulate the LinSweep in a Sweep
   # This registers the sweep as being a setpoint for everything measured inside
   # This sweep also automatically sets the parameter
   for val1 in Sweep(LinSweep(p1_set, 0, 1, 11)):
      for val2 in Sweep(LinSweep(p2_set, 0, 1, 11)):
         p1_get(val1+val2)
         
         # Measuring this parameter automatically recognizes that it is inside
         # a 2D loop and automatically attaches those setpoints
         # Note that we also here don't need to preregister p1_get
         msmt.measure(p1_get)
      
      # Also include measurement outside inner loop
      p_sum_above_ten(p1_get() > 10)
      msmt.measure(p_sum_above_ten)
```

### DIFF
--- a/qcodes/tests/dataset/measurement_loop/test_measurement_loop_basics.py
+++ b/qcodes/tests/dataset/measurement_loop/test_measurement_loop_basics.py
@@ -187,9 +187,15 @@ def test_2D_measurement_initialization(create_dummy_database):
       p2_set = ManualParameter('p2_set')
 
       with MeasurementLoop('test') as msmt:
-         for k, val in enumerate(Sweep(LinSweep(p1_set, 0, 1, 11))):
+         outer_sweep = Sweep(LinSweep(p1_set, 0, 1, 11))
+         for k, val in enumerate(outer_sweep):
             assert p1_set() == val
-            for val2 in Sweep(LinSweep(p2_set, 0, 1, 11)):
+            assert outer_sweep.is_first_sweep
+
+            inner_sweep = Sweep(LinSweep(p2_set, 0, 1, 11))
+            assert not inner_sweep.is_first_sweep
+            
+            for val2 in inner_sweep:
                assert p2_set() == val2
                p1_get(val+1)
                msmt.measure(p1_get)
--- a/qcodes/tests/dataset/measurement_loop/test_measurement_loop_basics.py
+++ b/qcodes/tests/dataset/measurement_loop/test_measurement_loop_basics.py
@@ -1,13 +1,71 @@
-import math
+import shutil
 import pytest
+import contextlib
+import tempfile
+from pathlib import Path
 
-def test_sqrt():
-   num = 25
-   assert math.sqrt(num) == 5
+from qcodes import Parameter, ManualParameter
+from qcodes.utils.dataset.doNd import LinSweep
+from qcodes.dataset.measurement_loop import MeasurementLoop, Sweep
+from qcodes.dataset import (
+    initialise_or_create_database_at,
+    load_by_run_spec,
+    load_or_create_experiment,
+)
 
-def testsquare():
-   num = 7
-   assert 7*7 == 40
+# def get_data_array(dataset, label):
 
-def testequality():
-   assert 10 == 11
+
+
+@pytest.fixture
+def create_dummy_database():
+
+   @contextlib.contextmanager
+   def func_context_manager():
+      with tempfile.TemporaryDirectory() as temporary_folder:
+         temporary_folder = tempfile.TemporaryDirectory()
+         print(f'Created temporary folder for database: {temporary_folder}')
+
+         assert Path(temporary_folder.name).exists()
+         db_path = Path(temporary_folder.name) / 'test_database.db'
+         initialise_or_create_database_at(str(db_path))
+
+         yield load_or_create_experiment("test_experiment")
+   return func_context_manager
+
+
+def test_create_measurement(create_dummy_database):
+   with create_dummy_database():
+      measurement_loop = MeasurementLoop('test')
+
+
+def test_basic_1D_measurement(create_dummy_database):
+   with create_dummy_database():
+      # Initialize parameters
+      p1_get = ManualParameter('p1_get')
+      p1_set = ManualParameter('p1_set')
+
+      with MeasurementLoop('test') as msmt:
+         for val in Sweep(LinSweep(p1_set, 0, 1, 11)):
+            assert p1_set() == val
+            p1_get(val+1)
+            msmt.measure(p1_get)
+
+   
+
+
+def test_basic_2D_measurement(create_dummy_database):
+   with create_dummy_database():
+      # Initialize parameters
+      p1_get = ManualParameter('p1_get')
+      p1_set = ManualParameter('p1_set')
+      p2_set = ManualParameter('p2_set')
+
+      with MeasurementLoop('test') as msmt:
+         for val in Sweep(LinSweep(p1_set, 0, 1, 11)):
+            assert p1_set() == val
+            for val2 in Sweep(LinSweep(p2_set, 0, 1, 11)):
+               assert p2_set() == val2
+               p1_get(val+1)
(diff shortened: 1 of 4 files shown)


jenshnielsen:
@nulinspiratie Thanks a lot, I am super busy this week but I promise to get back to this ASAP

nulinspiratie:
@jenshnielsen sounds good. I think at this point it doesn't make sense yet to do a full review. Of there's the sense that this could/should be part of Qcodes I can go ahead and write a tutorial on it's features. That'll make it easier to understand it's use

qcodes/utils/helpers.py
```diff
@@ -36,6 +39,8 @@
 
 import numpy as np
 
+from qcodes.configuration.config import DotDict
```
jenshnielsen, line 42: Avoid circular import

nulinspiratie:
So far I've been using it on our measurement setup. It's been working quite well, and I'm now updating the Sweep functionality so it's easier to create.

I'll probably have some more time over the upcoming weeks to work on this. @jenshnielsen can you give me an idea of what needs to happen before this PR can be accepted? I suspect that all the tests need to be converted to pytest, correct? What other things need to be taken care of?

jenshnielsen:
@nulinspiratie I will have a quick look tomorrow

nulinspiratie:
> @nulinspiratie I will have a quick look tomorrow

Great! I've begun on a tutorial at [docs/examples/DataSet/MeasurementLoop.ipynb](https://github.com/QCoDeS/Qcodes/pull/4074/files#diff-d699ec763d84302a2fc6142a88447f26cfc86aefac7bf22d776a850cfb01f9c9), it explains most of the features and also restrictions. I still need to write about some of the more advanced features, but the important parts are there.

jenshnielsen:
Thanks @nulinspiratie 

The notebook looks great. A few points on that:

I have been working on a more consistent API docs for qcodes. The aim is that everything public should be imported from the second level trying to make qcodes more try to the promise of being modular.

For the notebook that means:

Please update all imports to be from the second level. E.g. `from qcodes.parameters import ManualParameter` 
The api is reflected here https://qcodes.github.io/Qcodes/api/index.html#qcodes-api

For the new code please add the classes and functions that you think should be public to `qcodes/dataset/__init__.py` and `__all__` therein and they will be part of the api docs. 

Note these changes are not reflected in the current notebooks. I am waiting til after the next release with starting an update of those but for new notebooks I would like to stick to this.

qcodes/tests/dataset/measurement_loop/test_measurement_loop_basics.py
```diff
@@ -0,0 +1,256 @@
...
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+from qcodes import ManualParameter, Parameter
+from qcodes.dataset import initialise_or_create_database_at, load_or_create_experiment
+from qcodes.dataset.data_set import load_by_id
+from qcodes.dataset.measurement_loop import MeasurementLoop, Sweep
+from qcodes.utils.dataset.doNd import LinSweep
```
jenshnielsen, line 13: Please update imports to new public api as mentioned below

jenshnielsen:
I had a quick look at the code. Github seems to have some issue rendering the diff at the moment so no inline comments. 

Overall looks good. 

* We generally add types to all code within qcodes. I am happy to help with this if its a blocker
* I would like to think / understand more explicitly how much of this we consider public api
* I am a little bit worried that we have many different sweep classes. E.g in dond, here and the sweep thats already bound to the parameter (for use with the loop). I would like to think about how that can eventually be unified if possible. But perhaps the new sweeps here need not be public?

nulinspiratie:
Hey @jenshnielsen thanks for looking over it. Your comments all seem quite manageable, I can go ahead and start adding type hints and proper imports.

When you refer to "how much we should consider public api", do you mean which classes / functions should be placed in the `DataSet.__init__.py'?

The two important classes are `MeasurementLoop` and `Sweep`, I think it's worth making those part of the public API, though I agree that the number of sweeps is getting overcrowded. We could make the new `Sweep` a subclass of `AbstractSweep`, in which case I think it can be used in `dond` as well. In this case, since `LinSweep`, `LogSweep`, and `ArraySweep` can all be implemented straightforwardly by `Sweep`, one option could be to make `Sweep` the default. However, this would be quite a major change so I'm not sure if it's a good idea.  
  Another option would indeed be to not make `Sweep` part of the public API. However, it is a necessary object for the `MeasurementLoop`. One could consider attaching it to the `MeasurementLoop` like `msmt.sweep()`, though it can be useful to create a `Sweep` outside of a `MeasurementLoop` as it allows one to pass it as an argument in measurement functions.
As for `Parameter.sweep` that is tied to the original `Loop`, I see no easy way of unifying it with the rest. But is it still being actively used?

jenshnielsen:
> When you refer to "how much we should consider public api", do you mean which classes / functions should be placed in the `DataSet.init.py'?

Yes the goal is that users should only import stuff that is available from the second level e.g. users should do from `qcodes.parameters import Parameter` but not `from qcodes.parameters.parameter import parameter`

I would like to eventually get rid of the top level imports (`from qcodes import Parameter` since that is very confusing) 
But I don't expect to be able to remove that anytime soon if ever but plan to stop using it in examples.

jenshnielsen:
> The two important classes are MeasurementLoop and Sweep, I think it's worth making those part of the public API, though I agree that the number of sweeps is getting overcrowded. We could make the new Sweep a subclass of AbstractSweep, in which case I think it can be used in dond as well. In this case, since LinSweep, LogSweep, and ArraySweep can all be implemented straightforwardly by Sweep, one option could be to make Sweep the default. However, this would be quite a major change so I'm not sure if it's a good idea.


I think its worthwile investigating letting dond take your sweeps. Initially it can take those as a supplement to the existing ones and we can eventually migrate away from them

nulinspiratie:
I think the basic functionality of the MeasurementLoop has now been implemented. The tests I've written are working locally.

I see that MyPy is raising typing errors, I can fix that. But other than that, @jenshnielsen what needs to happen before accepting this PR?

PR #4086, opened 2022-04-08
Remove runtime dependency on setuptools

jenshnielsen:
With these changes setuptools is only used as a build time dependency

* Replace the use of `pkg_resources` (which is part of setuptools) with `importlib_metadata.distributions` part of std library
* Drop dependency on setuptools
* Remove 2 deprecated methods that uses setuptools

- [x] remove tests for removed methods 
- [x] Benchmark that this is not significantly slower.

Before 

```
In [1]: %time from qcodes.utils.installation_info import get_all_installed_package_versions
CPU times: total: 391 ms
Wall time: 712 ms

In [2]: %time get_all_installed_package_versions()
CPU times: total: 406 ms
Wall time: 402 ms
```

now

```

In [1]: %time from qcodes.utils.installation_info import get_all_installed_package_versions
CPU times: total: 391 ms
Wall time: 697 ms

In [2]: %time get_all_installed_package_versions()
CPU times: total: 172 ms
Wall time: 294 ms
Out[2]:

```

Old results with versioningit 1.1.0 (in a different env)

With this
```
In [1]: %time from qcodes.utils.installation_info import get_all_installed_package_versions
CPU times: total: 2.8 s
Wall time: 4.83 s

In [2]: %time get_all_installed_package_versions()
CPU times: total: 625 ms
Wall time: 920 ms
Out[2]:
```

Without

```
In [1]: %time from qcodes.utils.installation_info import get_all_installed_package_versions
CPU times: total: 2.34 s
Wall time: 4.28 s

In [2]: %time get_all_installed_package_versions()
CPU times: total: 0 ns
Wall time: 1 ms
```

So somewhat slower. pkg_resources is faster than importlib_metadata.distributions here since it does all the work at import time. The question is if this time then is spent somewhere else? 

Fresh terminal on this branch
```
In [1]: %time import pkg_resources
CPU times: total: 938 ms
Wall time: 1.08 s
```
Importing qcodes first

```
In [1]: %time import qcodes
CPU times: total: 3.17 s
Wall time: 6.1 s

In [2]: %time import pkg_resources
CPU times: total: 0 ns
Wall time: 0 ns
```
So clearly even with this change we are still importing pkg_resources indirectly. If
we get rid of that we should be able to regain all the time lost on this change


qcodes/utils/installation_info.py
```diff
@@ -8,15 +8,15 @@
...
-from qcodes.utils.deprecate import deprecate
-
-if sys.version_info >= (3, 8):
-    from importlib.metadata import PackageNotFoundError, distribution, version
+if sys.version_info >= (3, 10):
+    # distribution.name used below became part of the
+    # official api in 3.10
+    from importlib.metadata import distributions
 else:
     # 3.7 and earlier
```
jenshnielsen, line 18:
```suggestion
    # 3.9 and earlier
```

astafan8 (approved):
do the benchmarks in the description tell us that by removing setuptools as dependency, which is a good thing on it's own, we did not get a slowdown for sure, but also did not get any speed-ups? am i reading it right?

jenshnielsen:
> do the benchmarks in the description tell us that by removing setuptools as dependency, which is a good thing on it's own, we did not get a slowdown for sure, but also did not get any speed-ups? am i reading it right?

Yes with the fixed version of versioningit (1.1.1) we are not getting the slowdown. Its not a surprise that it performs roughly the same since both versions must inspect the same set of packages. The only difference is that importlib-metadata does this at runtime but setuptools (or rather pgk_resources) does it at import time.

jenshnielsen:
bors merge

PR #4153, opened 2022-05-03
Fix/partial with docstring args and return

jenshnielsen:
Allows partially applied functions to take non kwargs and return values just as a regular functools.partial function would

Note that it is only possible to pass args if you have partially applied later kwargs. As in the test you can pass a positionally if you have partially applied b but not the other way around. This matched how functools.partial works


jenshnielsen:
@sohailc I think you wrote this originally. Can you have a look?

qcodes/utils/helpers.py
```diff
@@ -666,16 +666,16 @@ def partial_with_docstring(func: Callable[..., Any],
     >>> ... pass
     >>> g = partial(f)
     >>> g.__doc__ = "bla"
-    >>> help(g) # this will print an unhelpful message
+    >>> help(g) # this will print the docstring of partial and not the docstring set above
 
     Args:
-        func: A function that its docstring will be accessed.
+        func: A function that izts docstring will be accessed.
```
jenshnielsen, line 672:
```suggestion
        func: A function that its docstring will be accessed.
```

qcodes/tests/helpers/test_partial_with_docstring.py
```diff
@@ -8,3 +8,32 @@ def f():
     docstring = "some docstring"
     g = partial_with_docstring(f, docstring)
     assert g.__doc__ == docstring
+
+
+def test_partial_with_docstring_returns_value():
```
jenshnielsen, line 13:
```suggestion
def test_partial_with_docstring_pass_args():
```

qcodes/tests/helpers/test_partial_with_docstring.py
```diff
@@ -8,3 +8,32 @@ def f():
...
+        return a + b
+
+    docstring = "some docstring"
+    g = partial_with_docstring(f, docstring, b=1)
+
+    assert g.__doc__ == docstring
+    assert g(2) == 3
+
+
+def test_partial_with_docstring_returns_value_2():
```
jenshnielsen, line 31:
```suggestion
def test_partial_with_docstring_returns_value():
```

trevormorgan (approved):

jenshnielsen:
bors merge

jenshnielsen:
bors retry

PR #4178, opened 2022-05-17
feature: Add check whether instrument name is valid identifier

### LINKED ISSUE #4128: Check that instrument name is a valid identifier
See https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/127 for background. 

For naming parameters etc it makes sense to enforce that instrument names are always valid identifiers. e.g. check `isidentifier(name)` and raise a suitable error if not

maximilianluc:
Minor feature addition. 
Checks whether instrument name is valid identifier. Whether name consists only of alphanumeric characters and underscores.

closes #4128 

- Added check in InstrumentBase to verify if name of instrument/channel/module is valid identifier.
- Adjusted instrument name "AMI430-3D" -> "AMI430_3D" in relevant test files to adhere to valid identifier naming.


jenshnielsen:
@maximilianluc Thanks for your pr.

A few technical details. We are using a formatter called darker which will basically run another formatter called black but only on the lines changed. To get the formatting job to pass you can do `pip install darker[isort]` and then run `darker -r master qcodes` 

We have also set the repo up with a [pre-commit hook](https://pre-commit.com/) this ensures that if you have it installed every time you commit you will run darker and other hooks automatically. You may want to set this up

qcodes/instrument/base.py
```diff
@@ -596,6 +596,8 @@ def __init__(
 
         self._t0 = time.time()
 
+        if not name.isidentifier():
+            raise ValueError(f"{name} invalid instrument identifier")
```
jenshnielsen, line 600: I think it might make sense to put this in the InstrumentBase class since not only istruments but actually also the channels and modules should have names that are valid identifiers

jenshnielsen:
To document changes we create a little file in docs/changes/newsfragments of the form pr number.type
In this case I think we should document this as a breaking change since some users may be creating instruments with invalid names. So in this case the file should be called  `4178.breaking` this file will later be merged with other changes to create a changelog

maximilianluc:
Thanks @jenshnielsen, that'll help with the commits! 

Perfect, will look into implementing it into the InstrumentBase class. Was thinking of it but wasn't sure if it was necessary for the others. I'll add a document to record the changes as well. Thanks!

maximilianluc:
Moved the check to InstrumentBase class and added the docs changelog file @jenshnielsen!

Only having a bit of trouble with the pre-commit.ci - pr check which doesn't pass here (locally it does). Any clues?
Also, the pre-commit package didn't seem to work with miniconda because of a missing "_ctypes" source file. With the virtualenv package the pre-commit package does work smoothly.

astafan8 (approved):

jenshnielsen:
Looking at examples in qcodes contrib drivers I found a number of drivers that have a `-` in their names https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/134

I think it would be a good idea to rather than error on those replace the `-` with `_` and print a warning if that is the only thing that makes the name invalid.

jenshnielsen:
@maximilianluc We are hoping to do a new qcodes release within a week or so. Would you have time to implement the thing suggested above before that?

maximilianluc:
Hi @jenshnielsen, yes for sure. Will try tomorrow, should be done before end of the week!

qcodes/instrument/base.py
```diff
@@ -409,6 +408,19 @@ def short_name(self) -> str:
...
+    @staticmethod
+    def _is_valid_identifier(name: str) -> str:
+        """Check whether given name is a valid instrument identifier."""
+        new_name = name.replace("-", "_")
+        if not new_name.isidentifier():
+            raise ValueError(f"{name} invalid instrument identifier")
+        if name != new_name:
+            warnings.warn(
+                f"Changed {name} to {new_name} for valid instrument identifier"
+            )
```
astafan8, line 420: shall the order be changed so that if the "new name" is still not an identifier then the user sees both the original name and the "new name" and knows what happened? also, ValueError should likely include "new name", not "name" (or perhaps both)?
```suggestion
        if name != new_name:
            warnings.warn(
                f"Changed {name} to {new_name} for valid instrument identifier"
            )
        if not new_name.isidentifier():
            raise ValueError(f"{new_name} invalid instrument identifier")
```

maximilianluc: I agree, this makes more sense. If order is changed then "new_name" in ValueError. Having both can be confusing since "new_name" can still be the same as "name". 
I'll change the warning string to: "Changed {name} to {new_name} for instrument identifier" since it is not automatically valid, an error can still be thrown.
Thanks! I'll update it now.

qcodes/tests/test_instrument.py
```diff
@@ -84,6 +84,23 @@ def test_instrument_fail(close_before_and_after):
...
+def test_instrument_on_invalid_identifier(close_before_and_after):
+    # Check if error raised when invalid identifer name given
+    with pytest.raises(ValueError, match="!-name invalid instrument identifier"):
+        DummyInstrument(name="!-name")
+    assert Instrument.instances() == []
+    assert DummyInstrument.instances() == []
+    assert Instrument._all_instruments == {}
+
+    # Check if name is valid identifier when dashes '-'
+    # are converted to underscores '_'
```
jenshnielsen, line 96: When testing code that triggers a warning its a good idea to check that the warning is actually raised. Both because it ensures that the code works correctly and because it will avoid the warning being shown when running the test suite. 
https://docs.pytest.org/en/6.2.x/warnings.html#warns shows how this can be done using pytest

jenshnielsen:
bors merge

PR #4182, opened 2022-05-18
Add polar parameter to PNA

guenp:
Add a parameter to the PNA driver to fetch the data using the polar format.
This reduces the measurement time; instead of pulling magnitude and phase separately, we now get all complex data points in a single trace.


guenp:
CC @ThorvaldLarsen

astafan8 (approved):
great, thank you!

guenp:
@astafan8, thanks for your review! It looks like we can't merge until code coverage is increased. I don't see any unit tests for the Keysight N52xx PNA; whereabouts would you recommend me to add test coverage for this feature?

astafan8:
bors merge

astafan8:
@guenp no worries about the coverage, it's an optional requirement. The more important requirement is to add a newsfragment file, which i just added, so no worries, it should get merged within an hour or so since this message :)

astafan8:
bors merge

PR #4206, opened 2022-05-31
Drop testing with conda python

jenshnielsen:
And replace windows job with a regular pip job

Conda should be readded to https://github.com/conda-forge/qcodes-feedstock


.github/workflows/pytest.yaml
```diff
@@ -17,8 +17,15 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest]
-        python-version: [3.7, 3.8, 3.9, "3.10"]
+        os: [ubuntu-latest, windows-latest]
```
Dominik-Vogel, line 20: @jenshnielsen, please ignore my ignorance: Does this simply mean we want to run the tests on the last four python versions on linux and only on the latest on Windows?

jenshnielsen: Yes, this is rather random and I think we should do a different pr to sub select the most relevant elements from the maxtrix. For now I just ported the one version on windows from conda to here. 

Exclude only seems to take since values (no arrays) so each job has to be excluded separately. There is a trick where you first exclude all jobs and then reinclude the one that you want but this seems more readable (if more verbose)

Dominik-Vogel (approved):

PR #4216, opened 2022-06-01
Do not import qcodes inside of _version.py

basnijholt:
This is not needed AFAIU.

### DIFF
--- a/qcodes/_version.py
+++ b/qcodes/_version.py
@@ -3,10 +3,7 @@ def _get_version() -> str:
 
     import versioningit
 
-    import qcodes
-
-    qcodes_path = Path(qcodes.__file__).parent
-    return versioningit.get_version(project_dir=qcodes_path.parent)
+    return versioningit.get_version(project_dir=Path(__file__).parent.parent)
 
 
 __version__ = _get_version()


qcodes/_version.py
```diff
@@ -3,10 +3,7 @@ def _get_version() -> str:
 
     import versioningit
 
-    import qcodes
-
-    qcodes_path = Path(qcodes.__file__).parent
-    return versioningit.get_version(project_dir=qcodes_path.parent)
+    return versioningit.get_version(project_dir=Path(__file__).parent.parent)
```
astafan8, line 6: this is likely correct, but perhaps the original code makes it more explicit what's going on? or an alternative below: ?
```suggestion
    qcodes_path = Path(__file__).parent
    return versioningit.get_version(project_dir=qcodes_path.parent)
```

jenshnielsen: I think the intension of doing it this way was to decouple this from the location of this specific file but either way is fine. In any case I don't think there is a need for a variable assigned

jenshnielsen:
superseded by https://github.com/QCoDeS/Qcodes/pull/4797 which got rid of any imports

PR #4219, opened 2022-06-02
Small improvements to visa backend

jenshnielsen:
* Slightly better docs and logging
* Correct the backend name to reflect current pyvisa
* Remove deprecated fallback for visa address with library included
* Let pyvisa figure out the default termination char

### DIFF
--- a/docs/changes/newsfragments/4219.breaking.1
+++ b/docs/changes/newsfragments/4219.breaking.1
@@ -1,0 +1,1 @@
+The deprecated fallback for visa addresses containing an ``@`` and the backend has been removed.
--- a/docs/changes/newsfragments/4219.breaking.2
+++ b/docs/changes/newsfragments/4219.breaking.2
@@ -1,0 +1,3 @@
+QCoDeS no longer explicitly sets the read and write terminators to `""` (The empty string) by default.
+The default terminators are now set by PyVISA. This should only have implications for RS232 instruments where
+the default read and write terminators now are ``\r``.
--- a/qcodes/instrument/visa.py
+++ b/qcodes/instrument/visa.py
@@ -103,7 +103,7 @@ class VisaInstrument(Instrument):
             self.visa_log.info('Opening PyVISA Resource Manager with default'
                           ' backend.')
             resource_manager = visa.ResourceManager()
-            self.visabackend = 'ni'
+            self.visabackend = "ivi"
 
         self.visa_log.info(f'Opening PyVISA resource at address: {address}')
         resource = resource_manager.open_resource(address)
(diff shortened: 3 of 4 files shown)


astafan8 (approved):

jenshnielsen:
@astafan8 Could you have another look. I made a change to not set the visa termination char unless explicitly given to match pyvisa here https://pyvisa.readthedocs.io/en/1.8/resources.html#termination-characters

astafan8:
>  I made a change to not set the visa termination char unless explicitly given to match pyvisa here https://pyvisa.readthedocs.io/en/1.8/resources.html#termination-characters

seems like the right thing to do.

astafan8 (approved):
breaking changes notes are added, so i think it's good to go

gztony1227:
@jenshnielsen @astafan8 Thank you both for incorporating this fix.
How do you pass down terminator when defining the tek 2012B?

I made terminator as a default argument in my local 2012 driver. kwargs won't be able to pass it down to visa.py right? I don't see visa.py looking for kwargs to determine terminator.

So far I like qcodes being installed and updated on pip. I haven't used it like a repository. So it will take a while for the new version to be pip?

    ```
                 timeout: float = 20, terminator: str = '', **kwargs: Any):
        """
        Initialises the TPS2012.

        Args:
            name: Name of the instrument used by QCoDeS
            address: Instrument address as used by VISA
            timeout: visa timeout, in secs. long default (180)
              to accommodate large waveforms
        """

        super().__init__(name, address, timeout=timeout, terminator = terminator, **kwargs) 
```

jenshnielsen:
It would be nice if you can test the changes before we release a new version so we are sure it works.
You can install from the master branch by doing `pip install git+https://github.com/QCoDeS/Qcodes.git` 
Once you are done testing you can then do `pip uninstall qcodes` and `pip install qcodes` to get back to the release version.
But if you have the time I would recommend lerning a bit of git to be able to install from a cloned repo

The value will be passed to the instrument with the current `__init__` class

If you do `TPS2012("name", "address", terminator="\n")` terminator will be part of the kwargs dict to the base vis aclass.
Here it will be avilable via the terminator argument whic is passed to set_terminator.

What I would like you to do is.

* install qcodes from source as above
* test if `scope = TPS2012('tek', 'ASRL5::INSTR')` works
* If that does not work test that `scope = TPS2012('tek', 'ASRL5::INSTR', terminator='\n')` works

gztony1227:
Dear Jens @jenshnielsen and Mikhail @astafan8 ,
The version worked well. I used a new environment on a separate PC to clone tttps://github.com/QCoDeS/Qcodes.git with pycharm and tps2012B is connected well.
scope = TPS2012('tek', 'ASRL5::INSTR') works

TPS2012('tek', 'ASRL5::INSTR', terminator='\n') also works.

The line "terminator = None" as default is good.
Please go ahead and release the version when you can. Thank you both very much.

jenshnielsen:
Thanks @gztony1227 i will go ahead and close the issue then

PR #4224, opened 2022-06-03
Public API of logging module

jenshnielsen:
This defines a public api for the logging module at qcodes.logger and updates examples to use this. Depends on 
https://github.com/QCoDeS/Qcodes/pull/4223


jenshnielsen:
This will generate one-page docs for the logging module like this:

image

The summary is generated by autodocsumm

astafan8 (approved):
bors merge

PR #4230, opened 2022-06-06
Build docs with Sphinx 4.5.0

jenshnielsen:
This tries to fix issue with multiple references in sphinx 4.4.0 and newer by importing fully qualified names. 

In https://github.com/QCoDeS/Qcodes/pull/4224 we use autodocsumm which officially does not support Sphinx 5 so we limit this to Sphinx 4.5 for now

### DIFF
--- a/qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
+++ b/qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
@@ -10,7 +10,7 @@ from . import constants
 from .constants import ModuleKind, SlotNr, MeasurementStatus, ChannelName, ChNr
 
 if TYPE_CHECKING:
-    from .KeysightB1500_base import KeysightB1500
+    import qcodes.instrument_drivers.Keysight.keysightb1500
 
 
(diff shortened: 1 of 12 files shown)


qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
```diff
@@ -10,7 +10,7 @@
 from .constants import ModuleKind, SlotNr, MeasurementStatus, ChannelName, ChNr
 
 if TYPE_CHECKING:
-    from .KeysightB1500_base import KeysightB1500
+    import qcodes.instrument_drivers.Keysight.keysightb1500
```
astafan8, line 13: so is there no way for sphinx to understand type importd with relative imports? :(

astafan8 (approved):

qcodes/instrument_drivers/Keysight/keysightb1500/KeysightB1500_module.py
jenshnielsen: Yes that does seem to be a bug/limitation in sphinx

PR #4248, opened 2022-06-09
Remove lower-case keysight folder

### LINKED ISSUE #1709: Filename collision due to case-sensitivity in Keysight folder

swernli:
Closes #1709

This fixes the case sensitivity issue that duplicate the Keysight folder with a lower case letter. While the net changes appear to just be the removal of the folder, there are five commits that effectively perform the removal through a sequence of renames that should work for both case sensitive file systems (Linux) and case insensitive systems (Windows). Note that newly cloned repos on Windows will no longer have the warning about a filename collision, but existing repos that update after this change will show an unstaged removal of qcodes/instrument_drivers/Keysight/__init__.py. Just reverting the change in VSCode or running `git restore qcodes/instrument_drivers/Keysight/__init__.py` from the cmdline will be enough to put the repo back into a clean state. Existing Linux and MacOS repos will be unaffected.


swernli:
@spauka Hope this works to resolve the issue we discussed!

jenshnielsen:
> but existing repos that update after this change will show an unstaged removal of qcodes/instrument_drivers/Keysight/init.py

We reached the conclusion when this was first discovered that this is unacceptable which is why this has not been fixed see for example https://github.com/QCoDeS/Qcodes/pull/1710

jenshnielsen:
While that is fairly trivial to undo for experienced git users its very confusing for someone not familiar with git

swernli:
> While that is fairly trivial to undo for experienced git users its very confusing for someone not familiar with git

Ah, I see. Taking that into account, I can see how this approach would be undesirable.

jenshnielsen:
That being said QCoDeS now has much more stable releases than last time we considered this so there will probably be fewer gic clones around so perhaps we should just do this

swernli:
@jenshnielsen Turns out that if I include content in the Keysight/__init__.py it will show up as a changed file and prevent the extraneous unstaged deletion on Windows. But if anyone ever removes the comment in there and changes it back into a blank file, then the unstaged deletion will come back. Probably too fragile to be useful in terms of an option, but I figured I'd reopen the PR just to show the possibility.

jenshnielsen:
@swernli Thanks for experimenting. I will have a chat with the other maintainers and try to decide if we should move forward with this. It would be really nice to get id of this extra folder

jenshnielsen:
@swernli I was intending to get back to this after the next release so took the liberty of reopening

jenshnielsen (approved):

jenshnielsen:
Landing this now since I discovered that any atttempt of editing `__init__.py` in either Keysight folder on windows will trigger this so this is wors that I thought

PR #4264, opened 2022-06-15
use short parameter annotation in docs

jenshnielsen:
This will become the default in sphinx 5 and makes the docs more readable

E.g. changes 

image

to

image


astafan8 (approved):
lovely!

PR #4266, opened 2022-06-15
[WIP] Document calibration module

jenshnielsen:
Not sure if this really should be documented. If so we should probably add a docstring to the function


astafan8 (approved):
added a docstring :) let's get it in :)

jenshnielsen:
bors merge

PR #4273, opened 2022-06-16
Docs: Enable autosummary for instrument module

jenshnielsen:
This currently requires a branch of sphinx to fix documenting inherited attributes. That branch is open as https://github.com/sphinx-doc/sphinx/pull/10691

To install sphinx 5.x we need the master branch of autodocsumm. That does not otherwise carry any fixes

This is done as a branch that carries my changes directly on top of 5.0.x since 5.1.0 broke the build. See https://github.com/sphinx-doc/sphinx/issues/10701

@astafan8 What do you think is it acceptable to build with a patched version of sphinx until that is merged.

This will also enable us to activate autodocsumm for https://github.com/QCoDeS/Qcodes/pull/4371 which will make that way more readable

### DIFF
--- a/docs/api/instrument/index.rst
+++ b/docs/api/instrument/index.rst
@@ -4,3 +4,4 @@ qcodes.instrument
 =================
 
 .. automodule:: qcodes.instrument
+    :autosummary:
--- a/docs_requirements.txt
+++ b/docs_requirements.txt
@@ -2,7 +2,7 @@ alabaster==0.7.12
 argon2-cffi==21.3.0
 async-generator==1.10
 attrs==21.4.0
-autodocsumm==0.2.8
+git+https://github.com/Chilipp/autodocsumm.git
 Babel==2.10.3
 backcall==0.2.0
 bleach==5.0.1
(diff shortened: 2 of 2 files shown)


astafan8 (commented):
> What do you think is it acceptable to build with a patched version of sphinx until that is merged

to keep things safe, i'd wait. i guess it all can easily take a month, right? the fix in sphinx is one thing, but also installing autodocsumm from master branch does not look super good.

what is the backup plan? we can just revert to a released version of sphinx and autodocsumm and remove the `:autosummary:` directive, right? if it's indeed that easy, then we can try to this risk, but perhaps pin autodocsumm to a particular commit?

jenshnielsen:
> to keep things safe, i'd wait. i guess it all can easily take a month, right? the fix in sphinx is one thing, but also installing autodocsumm from master branch does not look super good.

I would be happy to install it from a specific commit. The only reason that I did not bother is that its a project that moves fairly slow. 

> what is the backup plan? we can just revert to a released version of sphinx and autodocsumm and remove the :autosummary: directive, right? if it's indeed that easy, then we can try to this risk, but perhaps pin autodocsumm to a particular commit?

Yes, this can be undone by either removing the autosummary directive from modules that makes use of inherited attributes 
or by allowing building the docs with warnings as errors in which case the inherited attributes will simply be missing from the tabels in the docs.

astafan8 (approved):
ok, with the backup plan laid out, let's go :)

PR #4275, opened 2022-06-16
Refactor internal parameters api

jenshnielsen:
Now that parameters are in their own modules it makes sense to split this up into smaller modules.


qcodes/parameters/array_parameter.py
```diff
@@ -0,0 +1,214 @@
+import collections.abc
+import os
+from typing import TYPE_CHECKING, Any, Mapping, Optional, Sequence, Tuple, Type, Union
```
jenshnielsen, line 3:
```suggestion
from typing import TYPE_CHECKING, Any, Mapping, Optional, Sequence, Type, Union
```

qcodes/parameters/array_parameter.py
```diff
@@ -0,0 +1,214 @@
+import collections.abc
+import os
+from typing import TYPE_CHECKING, Any, Mapping, Optional, Sequence, Type, Union
```
jenshnielsen, line 3:
```suggestion
from typing import TYPE_CHECKING, Any, Mapping, Optional, Sequence, Type
```

trevormorgan (approved):

jenshnielsen:
bors merge

PR #4286, opened 2022-06-20
API: Move command module into parametes and undocument

jenshnielsen:
To make this possible also move Function into Parameters where it more naturally belongs


astafan8 (commented):
> move Function into Parameters where it more naturally belongs

could you share the reasoning behind this? i'm happy to share mine: it seems to me that Function belongs to the Instruments because it's an instrument-specific feature, and does not relate to parameters at all. Parameters don't use it and Functions dont use parameters, however Instruments use Functions and Function is means to live on an instrument (even as per it's docstring). Command on the other hand is indeed a thing that only Parameters happen to use, so Command belonging to Parameters makes a lot of sense (and it's not a package-scope utility, which is the original motivation of the PR, right?).

jenshnielsen:
> could you share the reasoning behind this? i'm happy to share mine: it seems to me that Function belongs to the Instruments because it's an instrument-specific feature, and does not relate to parameters at all. Parameters don't use it and Functions dont use parameters, however Instruments use Functions and Function is means to live on an instrument (even as per it's docstring). Command on the other hand is indeed a thing that only Parameters happen to use, so Command belonging to Parameters makes a lot of sense (and it's not a package-scope utility, which is the original motivation of the PR, right?).

Both parameters and functions are things that the instrument uses at exactly the same level. 1Both make use of Command so it makes sense to group all of these things together. In reality functions is a limited version of a parameter. I agree that the nameing of the module paraemters is not optimal from this way but since Functions is a minor thing that we really want to get rid of and deprecate (or at least strongly discourage the use of) I am not to worried

astafan8:
> Both make use of Command so it makes sense to group all of these things together

arggh, missed that Function uses Command. ok, no things make sense, thanks!

astafan8 (approved):

jenshnielsen:
Also see https://github.com/QCoDeS/Qcodes/pull/4292 which should fully eliminate the use of Functions

jenshnielsen:
bors merge

PR #4294, opened 2022-06-21
Docs: add api docs for interactive_widget

jenshnielsen:
And fix formatting of the same

### DIFF
--- a/qcodes/interactive_widget.py
+++ b/qcodes/interactive_widget.py
@@ -505,7 +505,7 @@ def experiments_widget(
     as plotting or the ability to easily browse
     the `~qcodes.dataset.data_set.DataSet`\s snapshot.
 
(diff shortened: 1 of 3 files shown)


qcodes/interactive_widget.py
```diff
@@ -532,3 +532,6 @@ def experiments_widget(
     tab = create_tab(do_display=False)
     grid = _experiment_widget(data_sets, tab)
     return VBox([title, tab, grid])
+
+
+__all__ = ["experiments_widget"]
```
astafan8, line 537: let's add it since it's neat? :)
```suggestion
__all__ = ["experiments_widget", "nested_dict_browser"]
```

astafan8 (approved):
🎉

PR #4295, opened 2022-06-21
Api: add get_instrument_logger to public api

jenshnielsen:
This is used outside the logging module so should be public IMHO @Dominik-Vogel


Dominik-Vogel (approved):
@jenshnielsen aggreed, this should be part of the public API.

PR #4306, opened 2022-06-23
API: Make numpyJSONencoder public

jenshnielsen:


jenshnielsen:
We could also create a json sub module?

astafan8:
> We could also create a json sub module?

yes, that sounds more appropriate. but up to you to decide :)

PR #4309, opened 2022-06-24
API: Add DelatedKeyboardInterrupt and depreacate to api

jenshnielsen:
A few questions

* The deprecate module contains 2 methods to assert deprecation messages. One of those were used in test_station. I have replaced that with the warnings testing logic from pytest. I think that using the pytest code is cleaner and more portable so I suggest that we only use that going forward and don't document these methods. 
* test_station still makes use of the function to generate a test message ``deprecation_message`` I don't however think this should be part of the public api. We could rewrite the tests to not make use of this function?

### DIFF
--- a/qcodes/tests/test_station.py
+++ b/qcodes/tests/test_station.py
@@ -17,8 +17,8 @@ from qcodes.parameters import DelegateParameter, Parameter
 from qcodes.station import SCHEMA_PATH, Station, ValidationWarning, update_config_schema
 from qcodes.tests.common import default_config
 from qcodes.tests.instrument_mocks import DummyInstrument
-from qcodes.utils import NumpyJSONEncoder
-from qcodes.utils.deprecate import assert_deprecated, deprecation_message
+from qcodes.utils import NumpyJSONEncoder, QCoDeSDeprecationWarning
+from qcodes.utils.deprecate import deprecation_message
 from qcodes.utils.helpers import YAML, get_qcodes_path
 
 from .common import DumyPar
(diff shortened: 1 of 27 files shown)


qcodes/tests/test_station.py
```diff
@@ -755,11 +755,11 @@ def test_deprecated_driver_keyword():
...
     type: DummyChannelInstrument
     """)
-    with assert_deprecated(
-        deprecation_message(
-            'use of the "driver"-keyword in the station configuration file',
-            alternative='the "type"-keyword instead, prepending the driver value'
-                        ' to it')):
+    message = deprecation_message(
+        'use of the "driver"-keyword in the station configuration file',
+        alternative='the "type"-keyword instead, prepending the driver value' " to it",
```
astafan8, line 760:
```suggestion
        alternative='the "type"-keyword instead, prepending the driver value to it',
```

astafan8 (approved):
> The deprecate module contains 2 methods to assert deprecation messages. One of those were used in test_station. I have replaced that with the warnings testing logic from pytest. I think that using the pytest code is cleaner and more portable so I suggest that we only use that going forward and don't document these methods.

agree with using pytest's syntax and don't expose those assertion methods as public api.

> test_station still makes use of the function to generate a test message deprecation_message I don't however think this should be part of the public api. We could rewrite the tests to not make use of this function?

deprecation_message indeed should not be part of the qcodes api for use outside of qcodes. our users should only want to silence/catch/etc the specific qcodes warning, and maybe use the convenient issue warning and decorate functions (but that I could also doubt about). However, i don't see any harm in using this function INSIDE qcodes, even in the tests (where actually matching on the "message" and "alternative" parts of the warning would be enough).

jenshnielsen:
bors merge

PR #4316, opened 2022-06-27
Add public api for installation info functions

jenshnielsen:
Stop using get_qcodes_version and deprecate. It does not seem to add any value over using qcodes.__version__ which is a convention


astafan8 (approved):

jenshnielsen:
bors merge

PR #4317, opened 2022-06-27
replace chardet with  charset-normalizer

jenshnielsen:
This is actually what we depend upon now since requests switched from one to the other


astafan8 (approved):

jenshnielsen:
bors merge

PR #4330, opened 2022-06-30
Move all members of utils.helpers to more descriptive locations and either make public or make part of individual modules

jenshnielsen:
* Move create_on_off_val ... to its own submodule and make public
* Move path helpers to their own module and rework to use importlib_resources rathat than importing qcodes
* Move abstract_method to its own submodule and rename so its more obvious that this is  not the std lib abstract decorator


qcodes/utils/path_helpers.py
```diff
@@ -0,0 +1,30 @@
...
+QCODES_USER_PATH_ENV = "QCODES_USER_PATH"
+
+
+def get_qcodes_path(*subfolder: str) -> str:
+    """
+    Return full file path of the QCoDeS module. Additional arguments will be
+    appended as subfolder.
+
+    """
+    import qcodes
```
jenshnielsen, line 13: This does not belong in utils due to the import of qcodes.

Perhaps the best is to replace with using importlib-resoucrces 

https://importlib-resources.readthedocs.io/en/latest/#:~:text=importlib_resources%20is%20a%20library%20which%20provides%20for%20access,that%20is%20located%20within%20an%20importable%20Python%20package.

jenshnielsen, line 13: Replace this with  https://importlib-resources.readthedocs.io/en/latest/ to avoid depending on qcodes here

jenshnielsen:
```
str(importlib_resources.files('qcodes').joinpath(*subfolders)
```

qcodes/utils/path_helpers.py
```diff
@@ -0,0 +1,36 @@
...
+QCODES_USER_PATH_ENV = "QCODES_USER_PATH"
+
+
+def get_qcodes_path(*subfolder: str) -> str:
+    """
+    Return full file path of the QCoDeS module. Additional arguments will be
+    appended as subfolder.
+
+    """
+    return str(files("qcodes").joinpath(*subfolder))
```
jenshnielsen, line 22:
```suggestion
    return str(files("qcodes").joinpath(*subfolder)) + os.sep
```

qcodes/utils/val_mapping.py
```diff
@@ -0,0 +1,30 @@
+from collections import OrderedDict
+from typing import Any, Dict, Tuple, Union
+
+
+def create_on_off_val_mapping(
```
astafan8, line 5: we could move this util into parameters module because it's used only there and is relevant for paramters only really. that would reduce the size of utils a bit :)

astafan8 (approved):

qcodes/tests/common.py
```diff
@@ -1,17 +1,18 @@
...
 import tempfile
-from typing import Any, Callable, Type, TYPE_CHECKING, Optional
 from contextlib import contextmanager
 from functools import wraps
 from time import sleep
-import cProfile
-import copy
+from typing import TYPE_CHECKING, Any, Callable, Dict, Hashable, Optional, Tuple, Type
+
+from _pytest._code.code import ExceptionChainRepr
```
astafan8, line 10: wow, amazing that this hasn't broke yet :)

jenshnielsen: Yes, this is only used in an assert which I guess is only there for typing? We could replace that assert with a cast or perhaps one can replace it with assert_type https://docs.python.org/3.11/library/typing.html#typing.assert_type, `__from__ future import annotations` and wrap the import in `if type_checking`?

qcodes/tests/helpers/test_permissive_range.py
```diff
@@ -1,5 +1,5 @@
 import pytest
```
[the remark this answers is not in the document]
jenshnielsen: Yes I agree but I think either once this refactor is completed or just after this pr will make it easier to manage.

qcodes/tests/common.py
jenshnielsen: should be fixed by https://github.com/QCoDeS/Qcodes/pull/4333

qcodes/plots/pyqtgraph.py
```diff
@@ -14,7 +14,7 @@
 from pyqtgraph.multiprocess.remoteproxy import ClosedError, ObjectProxy
 
 import qcodes
-import qcodes.utils.helpers
```
astafan8, line 17: a quick look at this code makes me thing that qcodes.utils are not used in this module, and then also the `self.qc_helpers`/`cls.qc_helpers` is not used at all, so perhaps we can just remove this import? or maybe i am wrong?

astafan8 (approved):
amazing work! looks great now :)

qcodes/plots/pyqtgraph.py
jenshnielsen: Lets keep it here but not make it part of the public api

jenshnielsen:
bors merge

PR #4338, opened 2022-07-01
Implement get_idn for Dummy instruments.

jenshnielsen:
Fixes log warnings showing up in examples such as [here](https://qcodes.github.io/Qcodes/examples/DataSet/Performing-measurements-using-qcodes-parameters-and-dataset.html#Implementing-a-measurement)

Also switch logging of this error to warning level (but keep traceback) since this is not fatal


jenshnielsen:
bors merge

PR #4339, opened 2022-07-01
API: move sphinx extension to extensions

jenshnielsen:
IMHO no need for this to be a top level module


qcodes/sphinx_extensions/parse_parameter_attr.py
```diff
@@ -1,215 +1,18 @@
...
+    find_init_func,
+    parse_init_function_from_str,
+    qcodes_parameter_attr_getter,
+    setup,
+)
+from qcodes.utils import issue_deprecation_warning
+
+issue_deprecation_warning(
+    "QCoDeS Sphinx extension: qcodes_parameter_attr_getter has been removed "
+    "from the public API. Please raise an issue at github.com/qcodes/qcodes "
```
astafan8, line 16: @caenrigen do you use the sphinx extension in your projects that depends on qcodes for better docs generation for parameters?

astafan8, line 16: @caenrigen do you use the sphinx extension in your projects that depends on qcodes for better docs generation for parameters?

astafan8 (approved):

qcodes/sphinx_extensions/parse_parameter_attr.py
caenrigen: Hi Mikhail! Thanks for the consideration :)

Yes, see for example https://quantify-quantify-core.readthedocs-hosted.com/en/v0.6.2/api_reference.html#quantify_core.measurement.control.MeasurementControl.instr_plotmon

And its usage at https://gitlab.com/quantify-os/quantify-core/-/blob/v0.6.2/docs/conf.py#L73

My initial intention was to contribute with something that more people and repos could use in their docs. Documenting qcodes parameters manually specially when having a lot of them is painful.

PS I have switched jobs in the meantime so better to ask folks from Qblox/Orange Quantum Systems, CC @AdriaanRol

astafan8: @jenshnielsen  then perhaps let's remove the warning for now?

jenshnielsen: @astafan8 Probably we should just drop this and not move it. Sphinx extensions need to have a setup function in their module so it does not really make sense to let this share a module with any thing else

astafan8: ok!

jenshnielsen: Thanks for the quick feedback @caenrigen

PR #4342, opened 2022-07-01
fix option detection on Keysight N51x1 driver

simonzihlmann:
make frequency option detection compatible with multiple options installed on device. The old version of the driver resulted in a `KeyError` if multiple options were installed on the device.

I was unable to figure out from the manual / spec sheet if the frequency option will always be the first one in the string returned by the instrument. For the moment, all other options are ignored.


jenshnielsen:
@simonzihlmann Thanks, could you add a news fragment as described in the pr template?

jenshnielsen (approved):

jenshnielsen:
bors merge

qcodes/instrument_drivers/Keysight/N51x1.py
```diff
@@ -13,6 +13,27 @@ class N51x1(VisaInstrument):
     def __init__(self, name: str, address: str, min_power: int = -144, max_power: int = 19, **kwargs: Any):
         super().__init__(name, address, terminator='\n', **kwargs)
 
+        self._options = self.ask("*OPT?")
```
astafan8, line 16: the comment in the old code makes me think that the instrument response is a comma-separated list of numbers, right? if so, then it would've been better to split the returned string into an actualy list of strings with numbers. the ``if f_option in self._options`` check would still work, but would be more robust, so that e.g. option `50101` does not get mistakenly treated as option `501` (but of course likely option `50101` does not exist, so this all is likely not a problem in real life)
```python
self._options = self.ask("*OPT?").split(",")
```

PR #4346, opened 2022-07-03
Update N51x1.py

bguttel:
The added `.split(',')[0]` helps with model N5183B (where the returned string is, e.g.  '520,UNY,UNW') and does no harm when the returned value is a simple string (e.g. '520').


jenshnielsen:
@bguttel Thanks for your contributuion. Unfortunatly the same fix was also contributed in https://github.com/QCoDeS/Qcodes/pull/4342 just slightly before this one

PR #4349, opened 2022-07-04
API: Make numpy utils public

jenshnielsen:


Dominik-Vogel (approved):
I am surprised by what is all public API (like deprecate). I am really happy that you are driving this process to clean things up.

astafan8 (approved):

jenshnielsen:
> I am surprised by what is all public API (like deprecate). I am really happy that you are driving this process to clean things up.

My guiding lines what that if QCoDeS should be truely modular then each module should only use the public API of other modules. We are not fully there yet

PR #4350, opened 2022-07-04
Api: plotting

jenshnielsen:
Move plotting module to its own toplevel module and split code in mpl and generic files. This module is imho too big to be part of utils.


astafan8 (approved):

jenshnielsen:
bors merge

PR #4352, opened 2022-07-04
Enable pep 585 and pep 604 using from __future__ import annotations

jenshnielsen:
This enables the new style type annotations in pep851 and pep604. Since these things are not supported for python <3.10
this only works in files that uses `from __future__ import annotations` I imagine that if we do this we would gradually add that to all files after this is merged. The actual transforms are done by pyupgrade by removing `--keep-runtime-typing` from the options

 One drawback is that the new syntax still cannot be used in Type aliases. 

Another issue is that this does not build docs without warnings for python <3.10 
I have therefor changed the build to upload from 3.10 build and allow sphinx warnings for python < 3.10

See [pep 585](https://peps.python.org/pep-0585/) and [pep 604](https://peps.python.org/pep-0604/) for more context.


astafan8 (approved):
stepping further into the future :)

jenshnielsen:
bors merge

PR #4354, opened 2022-07-04
Avoid using utils.types in dataset.

jenshnielsen:
Use base classes which all the concrete types are instances of. This simplifies the code and avoids dependencies on the types module outside tests which can therefor stay as is and not be part of the api


astafan8 (dismissed):
looks good! as long as sqlite3.register_adapter indeed understands the subtyping (e.g. we don't need to expliciutly mention every type regardless of numpy version or something else)

jenshnielsen:
We have round trip tests for all integer and float types I think we should do the same for complex

PR #4382, opened 2022-07-11
migrate setuptools pep621 config to pyproject.toml

jenshnielsen:
This moves the supported and non experimental
parts of setuptools config to pyproject.toml
These are the parts standardized by pep621 and
supported in setuptools >=65 

These are the supported parts of https://github.com/QCoDeS/Qcodes/pull/4049

- [x] Verify how this works with older versions of pip and make sure we handle that reasonably
- [x] Check that editable/normal install from source works as expected
- [x] Verify that package content is identical
- [x] requirements-builder does not yet support this so out min requirements build fails: Pr open here https://github.com/inveniosoftware/requirements-builder/pull/56

### DIFF
--- a/.github/workflows/pytest-min-versions.yaml
+++ b/.github/workflows/pytest-min-versions.yaml
@@ -45,6 +45,8 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-${{ matrix.python-version }}-pip-
     - name: generate min_requirements.txt
+#     Use fork of requirements-builder until official version
+#     has support for pyproject.toml (pep621)
       run: |
         pip install git+https://github.com/jenshnielsen/requirements-builder.git
         pip install versioningit
--- a/.github/workflows/pytest-min-versions.yaml
+++ b/.github/workflows/pytest-min-versions.yaml
@@ -46,7 +46,7 @@ jobs:
           ${{ runner.os }}-${{ matrix.python-version }}-pip-
     - name: generate min_requirements.txt
       run: |
-        pip install requirements-builder
+        pip install git+https://github.com/jenshnielsen/requirements-builder.git
         pip install versioningit
         requirements-builder -l min -e test setup.py -o min_requirements.txt
     - name: install min_requirements.txt
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,63 +1,6 @@
-[metadata]
-name = qcodes
-maintainer = QCoDeS Core Developers
-maintainer_email = qcodes-support@microsoft.com
-description = Python-based data acquisition framework developed by the Copenhagen / Delft / Sydney / Microsoft quantum computing consortium
-long_description = file: README.rst
-long_description_content_type = text/x-rst
-url = https://github.com/QCoDeS/Qcodes
-classifiers =
-    Development Status :: 3 - Alpha
-    Intended Audience :: Science/Research
-    License :: OSI Approved :: MIT License
-    Programming Language :: Python :: 3 :: Only
-    Programming Language :: Python :: 3.7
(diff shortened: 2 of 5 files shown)


setup.cfg
```diff
@@ -1,63 +1,6 @@
...
-    Programming Language :: Python :: 3.10
-    Topic :: Scientific/Engineering
-license = MIT
-project_urls =
-    Documentation = https://qcodes.github.io/Qcodes/
-    Source = https://github.com/qcodes/qcodes
-    Tracker = https://github.com/QCoDeS/Qcodes/issues
-    Changelog = https://qcodes.github.io/Qcodes/changes/index.html
-
 [options]
```
astafan8, line 1: do you have a feeling on when the options section will be added to pyproject.toml?

.github/workflows/pytest-min-versions.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
           ${{ runner.os }}-${{ matrix.python-version }}-pip-
     - name: generate min_requirements.txt
       run: |
-        pip install requirements-builder
+        pip install git+https://github.com/jenshnielsen/requirements-builder.git
```
astafan8, line 49: since this is used only for OUR CI it's ok to use the fork. hope the upstream will be updated soon :)

setup.py
```diff
@@ -1,6 +1,21 @@
+import setuptools
 from setuptools import setup
 from versioningit import get_cmdclasses
 
+if int(setuptools.__version__.split(".")[0]) < 61:
+    raise RuntimeError(
+        "At least setuptools 61 is required to install qcodes from source"
```
astafan8, line 7: do i understand correctly that we need these guards for cases where `pip install ...` is not used, which is not a commonplace but nice to keep being supported?

setup.py
```diff
@@ -1,6 +1,21 @@
+import setuptools
 from setuptools import setup
 from versioningit import get_cmdclasses
 
+if int(setuptools.__version__.split(".")[0]) < 61:
```
astafan8, line 5: are we sure that `__version__` attribute for setuptools and pip exist on all the versions that realistically can be used? i understand that this is soooooo unlikely :)

setup.cfg
jenshnielsen: They are there but the syntax is not officially supported yet. No eta as far as I know

setup.py
jenshnielsen: I think there are 2 ways we could get here with a version of setuptools that is too old. 

* Installing from source using either `python setup.py install` or `pip install . --no-build-isolation` from an env that has an older version of setuptools or using some other tool that does not respect pep 517 build isolation. 
* Using a version of pip that is too old to support pep517 (this was added in pip 19)

In any case this will only happen if you install from source not from a wheel)

setup.py
jenshnielsen: For setuptools this has been there at least since 34.0.0 which was the earliest version that I can actually install.

For pip at least since 10.0 which was the earliest version I could get to work

setup.py
astafan8: then we are quite safe :)

setup.py
astafan8: great!

setup.cfg
```diff
@@ -1,63 +1,6 @@
-[metadata]
-name = qcodes
```
jenshnielsen, line 2: Should add a note here and to setup.py of where config should go

setup.py
jenshnielsen: verified manually that editable install using pip 19.0.0 works as expected

Issue #3774, opened 2022-01-04
Use weakref.finalize to safely close instruments?


jenshnielsen:
Currently we are relying on either `__del__` to ensure that an instruments connections are closed when it is garbage collected.

This does not work well and is very hard to get right since ``__del__` may be called after the object has been destryed and close is no longer possible to call. 

Therefor we are using atexit.finalize to close all instruments. This seems to work fairly well however it may make sence to use [weakref.finalize](https://docs.python.org/3/library/weakref.html#weakref.finalize) to close each instrument. 

Lifted from https://github.com/QCoDeS/Qcodes/issues/1882

jenshnielsen:
This has been implemented

PR #4386, opened 2022-07-12
feature: replace __del__ with weakref.finalize()

maximilianluc:
Feature addition: replacing __del__ with weakref.finalize() as called method during object garbage collection.

Issue: #3774 

- renamed `__del__` as `__finalize`, logic stays the same. 
- added weakref.finalize() in `__init__` in instrument.py calling upon '__finalize()' during gc.

### DIFF
--- a/qcodes/instrument/instrument.py
+++ b/qcodes/instrument/instrument.py
@@ -69,6 +69,7 @@ class Instrument(InstrumentBase, metaclass=InstrumentMeta):
         super().__init__(name=name, metadata=metadata, label=label)
 
         self.add_parameter("IDN", get_cmd=self.get_idn, vals=Anything())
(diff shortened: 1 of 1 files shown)


jenshnielsen:
@maximilianluc Thanks for this. It looks great but I will need a little bit of time to test this manually to see how it works in practice

jenshnielsen:
@maximilianluc Sorry for the late response. I have spent a bit of time investigating this and unfortunately its a bit more complicated as the finalize method cannot contain a reference to self since that would increase the ref count to the instrument preventing the finalizer to be called. https://docs.python.org/3/library/weakref.html#weakref.finalize

In https://github.com/QCoDeS/Qcodes/pull/5341 I have implemented a partial solution for visa instruments that do not have this issue

PR #4391, opened 2022-07-13
use xvfb action to run in xvfb

jenshnielsen:
Trying to debug failure of ci build


jenshnielsen:
Doen't work and further more break multiline runs

PR #4393, opened 2022-07-14
Try fixing xvfb

jenshnielsen:
As suggested in
https://github.com/actions/virtual-environments/issues/5894


astafan8 (approved):
@jenshnielsen thank you sooooooooooo much, looks like the workaround is working!

PR #4395, opened 2022-07-14
Docs: use furo theme

astafan8:
https://github.com/pradyunsg/furo

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -216,11 +216,11 @@ html_theme = "furo"
 # The name for this set of Sphinx documents.
 # "<project> v<release> documentation" by default.
 #
-# html_title = 'QCoDeS v1'
+html_title = f"{project} v{release}"
 
 # A shorter title for the navigation bar.  Default is the same as html_title.
 #
-# html_short_title = None
+html_short_title = f"{project}"
 
 # The name of an image file (relative to this directory) to place at the top
 # of the sidebar.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -203,7 +203,7 @@ todo_include_todos = True
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = "sphinx_rtd_theme"
+html_theme = "furo"
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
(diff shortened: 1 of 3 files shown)


astafan8:
looks good , but we need to do some customizations so that more than one level of sub titles and sub pages is properly rendered, ie not as a flat list, see here:
image

trevormorgan (approved):

trevormorgan:
Looks really clean, definitely an upgrade.

jenshnielsen:
@astafan8 I took the liberty of rebasing this on top of master with #4434

jenshnielsen:
Hmn not sure what happens here?

```
ERROR: Cannot install qcodes[docs]==0.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    qcodes[docs] 0.0 depends on furo>=2022.06.21; extra == "docs"
    The user requested (constraint) furo~=2022.06.21

To fix this you could try to:
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

jenshnielsen:
Seems like this is caused by using `~=` with the non conformant version number `2020.06.21`

trevormorgan:
@astafan8 I think it would be nice to land this PR, would you be okay if I took it over and tried to finish it up?

jenshnielsen:
@trevormorgan I think that would be great. @astafan8 and I talked about landing https://github.com/QCoDeS/Qcodes/issues/4457 first since that would significantly reduce the number of nested tabels in the toc. If you want to research if there is a way to get multiple layers in the toc to render correctly that would IMHO be great

astafan8:
@trevormorgan sure, any help is appreciated :)

and indeed, as Jens said, it does make sense to finish the instrument driver api part first, which will help reduce the levels , which in turn should make our docs structure work better with furo.

However, another thing we could do is to find another more modern theme that does not prescribe limits on the nesting levels in the docs, e.g. matplotlib/numpy/pandas perhaps are worthy examples to look at (i havent' done that yet).

trevormorgan:
Sounds good @jenshnielsen and @astafan8. I'll help out on #4457 first before jumping on this PR.

astafan8:
closing as furo might not be the best theme to use, see above in the comments

PR #4413, opened 2022-07-24
Document correct signature for Instruments

### LINKED ISSUE #4370: Instruments are documented with unhelpful init signature
Considder for example the docs of the [ HP8133A ](https://qcodes.github.io/Qcodes/api/generated/qcodes.instrument_drivers.HP.html#module-qcodes.instrument_drivers.HP.HP8133A)

image

Note how only `args: Any, kwargs: Any` is shown

This can in turn be found for all subclasses until [Instrument](https://qcodes.github.io/Qcodes/api/instrument/index.html?highlight=instrument#qcodes.instrument.Instrument)

This appears to be due to Sphinx picking the signature of `__call__` defined in the `InstrumentMeta` metaclass rather than the `__init__` signature

jenshnielsen:
Fixes #4370 by not using custom metaclass when documenting to work around sphinx documenting __call__ rather than __init__

This feels like a bit of a hack so not sure if its ok. Inspired by https://stackoverflow.com/questions/20843737/check-if-sphinx-doc-called-the-script

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -33,9 +33,13 @@ from packaging.version import parse
 # rather than `__init__` that is unhelpful
 # for instruments. When building the docs
 # we patch it back to ABCMeta
+# this should happen as early as possible
 import qcodes.instrument.instrument_meta
 
 qcodes.instrument.instrument_meta.InstrumentMeta = ABCMeta
+# we need to reload any module that has been imported and
+# makes use of this metaclass. The modules below are all imported
+# by importing qcodes.instrument so we need to reload them
 reload(qcodes.instrument.instrument)
 reload(qcodes.instrument.ip)
 reload(qcodes.instrument.visa)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,6 +18,8 @@
 #
 import os
 import sys
+from abc import ABCMeta
+from importlib import reload
 
 # Import matplotlib and set the backend
 # before qcodes imports pyplot and automatically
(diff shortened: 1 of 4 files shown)


qcodes/instrument/instrument.py
```diff
@@ -28,46 +28,15 @@
...
-                f"and cannot be initialized"
-            )
 
-        new_inst.record_instance(new_inst)
-        return new_inst
+# setting the metaclass will cause sphinx
+# to document the signature of `__call__`
+# rather than `__init__` that is unhelpful
+# for instruments.
+if "sphinx" in sys.modules:
```
astafan8, line 36: can't we do this as part of conf.py, so that it's exclusively for building docs? perhaps using patching from the mock module?

qcodes/instrument/instrument_meta.py
```diff
@@ -0,0 +1,44 @@
+from abc import ABCMeta
+from typing import Any
+
+
+class InstrumentMeta(ABCMeta):
```
astafan8, line 5: moving this class into its own module makes sense on its own :)

qcodes/instrument/instrument.py
jenshnielsen: I agree, Actually I did that because I could not get assigning the metaclass to a variable to typecheck but no need to tell anyone ;)

qcodes/instrument/instrument_meta.py
jenshnielsen: charset-normalizer

qcodes/instrument/instrument.py
jenshnielsen: Yes I agree the current way is not great mocking sounds like a interesting solution

jenshnielsen: I have implemented an alternative solution using monkeypatching in conf.py. I think that is cleaner

docs/conf.py
```diff
@@ -26,6 +28,19 @@
...
+# to document the signature of `__call__`
+# rather than `__init__` that is unhelpful
+# for instruments. When building the docs
+# we patch it back to ABCMeta
+import qcodes.instrument.instrument_meta
+
+qcodes.instrument.instrument_meta.InstrumentMeta = ABCMeta
+reload(qcodes.instrument.instrument)
+reload(qcodes.instrument.ip)
+reload(qcodes.instrument.visa)
```
astafan8, line 41: would a little comment saying "we need to reload all the modules that import InstrumentMeta" here be useful for anyone reading/developing/improving?

astafan8 (approved):

jenshnielsen:
bors merge

PR #4419, opened 2022-07-25
Upgrade versioningit to 2.0.0

jenshnielsen:
I manually verified that
* both sdist and bdist have correct version both in metadata and _version.py
* pip install gives the correct version (including ``qcodes.__verison__`` from editable)


astafan8 (approved):
awesome! thank you for explicitly verifying those things

PR #4430, opened 2022-07-27
Add missing init files

jenshnielsen:
There were not committed in #4408 since the whole dist dir was gitignored. I don't think that dir should be ignored so this unignores dist within qcodes. The toplevel dist dir remains ignored as it should


astafan8 (approved):
good catch!

jenshnielsen:
> good catch! 

You can thank build and setuptools for some very verbose warnings ;)

PR #4436, opened 2022-07-29
Update Stanford drivers to conform to docs standard

jenshnielsen:
I have left the drivers in the format SR860. I guess one could argue that they should be named SRSSR860 since SR is part of the model number but that feels too much.


astafan8 (approved):
> I have left the drivers in the format SR860. I guess one could argue that they should be named SRSSR860 since SR is part of the model number but that feels too much.

agree!

jenshnielsen:
bors merge

PR #4443, opened 2022-07-31
Update yokogawa drivers to conform to standard

jenshnielsen:


qcodes/instrument_drivers/yokogawa/Yokogawa_GS200.py
```diff
@@ -0,0 +1,698 @@
+from functools import partial
```
astafan8, line 1: is it worth it copying the whole driver? wouldn't  a subclass with old name work as well (same with the `_float_round`)?

astafan8 (approved):

qcodes/instrument_drivers/yokogawa/Yokogawa_GS200.py
jenshnielsen: Yes it could be done but being very defensive someone could depend on that the name of the class of the measurement module inside the driver so I would need to customize it such that the classes of the module attributes stayed the same. That did not feel worth the risk of getting it wrong

jenshnielsen: @astafan8 Does that tradeoff seem ok to you?

astafan8: (sorry, forgot to respond) yes, it's good!

PR #4445, opened 2022-08-01
VisaInstrument replace kwargs with metadata

jenshnielsen:
VisaInstrument used to take kwargs but actually the only
non explicit argument is metadata so we might as well write that
out.

### DIFF
--- a/qcodes/instrument/visa.py
+++ b/qcodes/instrument/visa.py
@@ -1,6 +1,6 @@
 """Visa instrument driver based on pyvisa."""
 import logging
-from typing import Any, Dict, Optional, Sequence, Tuple
+from typing import Any, Dict, Mapping, Optional, Sequence, Tuple
 
 import pyvisa
 import pyvisa.constants as vi_const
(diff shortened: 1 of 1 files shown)


astafan8 (approved):
do we need to add changelog item because it might be a breaking change for some external code?

jenshnielsen:
I dont think so. Instrument did not take and more kwargs so that would just throw. I guess it could throw in multiple inheritance?

astafan8:
> I guess it could throw in multiple inheritance?

oh, that's even a better catch. need to write a test to check....

jenshnielsen:
@astafan8 I mean the following will certainly break.

```
class Foo

   def __init__(fooness=None, **kwargs):
       self._fooness = fooness
       super().__init__(**kwargs)


class Bar(VisaInstrument, Foo)

    def __init__(name, **kwargs)


bar = Bar(name="bar", address="someaddr", fooness="max")

```
so if we care about that we should not make the change

astafan8:
> so if we care about that we should not make the change

i do hope that we don't care :) we could try to release with this change and revert it if someone reports a breakage. or reserve this change for the v1.0 times :)

jenshnielsen:
@astafan8 We could also be more defensive and add metadata as a keyword argument and trigger a warning if len(kwargs) > 0

astafan8:
> We could also be more defensive and add metadata as a keyword argument and trigger a warning if len(kwargs) > 0

ah! great idea, i support this :)

PR #4473, opened 2022-08-04
Update lakeshore drivers to conform with standard

jenshnielsen:
I could be convinced that model should not be part of the class names. What do you think?

I have copied the drivers rather than making the old ones aliases since there was a fair bit of renaming involved but imported 
the channel/modul classes from the new locations under aliases to avoid too much duplication.


astafan8 (commented):
> I could be convinced that model should not be part of the class names. What do you think?

yeah... it's annoying, however it is part of the official name of the instrument, so if we want to follow our own convention, then we should. or we have to relax the conveniention in a clear way :)

> I have copied the drivers rather than making the old ones aliases since there was a fair bit of renaming involved

sad, so much diplication :( and the tests then also need to be copied (or at least use the new drivers)

astafan8 (approved):
lovely!

PR #4485, opened 2022-08-08
Run mypy against installed qcodes

jenshnielsen:
This should make it more likely that mypy will spot files missing from the install


Dominik-Vogel (approved):
Great! I remember the long debugging session we have had that could have been avoided with this flag.

PR #4499, opened 2022-08-12
Add `connect` and `ConnectionPlus` to public API

Dominik-Vogel:
Many functions of the public API, like the `load_by_*`s,  are using a `ConnectionPlus` to avoid creating a new connection on every call. This PR exposes `connect` in order to provide the user a way of creating such a connection. `ConnectionPlus` is exposed to enable typing.


jenshnielsen (approved):
Thanks

PR #4500, opened 2022-08-12
Configure strict options for mypy and enable some of them

jenshnielsen:
This explicitly configures all of mypys strict modes and enables some of them.

Along the way this fixes what looks like 2 bugs in drivers

### DIFF
--- a/qcodes/instrument_drivers/QDevil/QDevil_QDAC.py
+++ b/qcodes/instrument_drivers/QDevil/QDevil_QDAC.py
@@ -383,7 +383,7 @@ class QDac(VisaInstrument):
 
                 self._assigned_fgs[chan] = Generator(fg)
                 self._assigned_fgs[chan].t_end = time_end
-                if trigger != 0:
+                if int(trigger) != 0:
                     self._assigned_triggers[fg] = int(trigger)
                 for syn in range(1, self._num_syns+1):
                     self.write(f'syn {syn}')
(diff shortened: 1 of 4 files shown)


qcodes/instrument_drivers/QDevil/QDevil_QDAC.py
```diff
@@ -383,7 +383,7 @@ def _load_state(self) -> None:
 
                 self._assigned_fgs[chan] = Generator(fg)
                 self._assigned_fgs[chan].t_end = time_end
-                if trigger != 0:
```
jenshnielsen, line 386: @jpsecher This looks to me like a bug in the (Qdevil QDac1) driver. On this line trigger is a string but its compared against 0 so that comparison will always fail. Does this check look ok to you

WilliamHPNielsen (approved):

qcodes/instrument_drivers/QDevil/QDevil_QDAC.py
astafan8: yes, indeed! i think it's a very good catch!

astafan8 (approved):

jenshnielsen:
bors merge

qcodes/instrument_drivers/QDevil/QDevil_QDAC.py
jpsecher: Yes, it seems like a bug

PR #4536, opened 2022-08-21
Upgrade AMI drivers to conform to our standard

jenshnielsen:


jenshnielsen:
The issue seems to be caused by the AMI430.py file next to the visa fill. Need to debug why the stops sphinx looking for classes in the mro

jenshnielsen:
Thinking a bit more about this. Its not strange that this trips up. Both the old Module and the class are called AMI430 so sphinx ends up trying to import the old AMI430 module rather than the class. This also has a AMI430 class but which inherits from ipinstrument so its lacking the expected attributes. 

Not sure what the best is. 

1) Rename the old module to AMI_430.py perhaps. This would be the cleanest but is also a breaking change. 
2) Rename the class as exported to something else. Can be done without breaking changes but is less ideal

qcodes/instrument_drivers/american_magnetics/AMI430_visa.py
```diff
@@ -561,31 +552,35 @@ def ask_raw(self, cmd: str) -> str:
         return result
 
 
-class AMI430_3D(Instrument):
+class AMI430(AMIModel430):
+    pass
+
+
+class AMIModel4303D(Instrument):
```
astafan8, line 559: perhaps we can allow an underscore before 3D for readability?
```suggestion
class AMIModel430_3D(Instrument):
```

astafan8 (approved):

qcodes/instrument_drivers/american_magnetics/AMI430_visa.py
jenshnielsen: Nah, I think we want to stick to conformant names and pep8 does not endorse  `_` in class names

astafan8: ok, fair!

jenshnielsen:
bors merge

PR #4538, opened 2022-08-21
Upgrade Rigol drivers to conform to our standard

jenshnielsen:


qcodes/instrument_drivers/rigol/Rigol_DS4000.py
```diff
@@ -0,0 +1,323 @@
...
+
+from qcodes import validators as vals
+from qcodes.instrument import ChannelList, InstrumentChannel, VisaInstrument
+from qcodes.parameters import ArrayParameter, ParamRawDataType
+
+log = logging.getLogger(__name__)
+
+
+class RigolDS4000TraceNotReady(Exception):
+    pass
```
astafan8, line 19: wouldn't this help with NOT making the exception class be in the list of the breaking changes? but it add an unnecessary extra class to the "new well-named" driver module which is not super nice. Perhaps the same could work for the channels? Anyways, noting here but leaving the decision to you.
```suggestion
class RigolDS4000TraceNotReady(Exception):
    pass


class TraceNotReady(RigolDS4000TraceNotReady):
    pass
```

astafan8 (approved):

qcodes/instrument_drivers/rigol/Rigol_DS4000.py
jenshnielsen: Would that work. If you catch a subclass of the exception being raises it will not do what you expect.

```
In [8]: class MyException(Exception):
   ...:
   ...:     pass
   ...:

In [9]: class MyExceptionAlias(MyException):
   ...:
   ...:     pass
   ...:

In [10]: try:
    ...:     raise MyException
    ...: except MyExceptionAlias:
    ...:     print("Caught an exception")
    ...:
MyException                               Traceback (most recent call last)
Cell In [10], line 2
      1 try:
----> 2     raise MyException
      3 except MyExceptionAlias:
      4     print("Caught an exception")

MyException:
```

jenshnielsen: Thou this would work

```
In [11]: MyExceptionAlias = MyException

In [12]: try:
    ...:     raise MyException
    ...: except MyExceptionAlias:
    ...:     print("Caught an exception")
    ...:
```
so lets just do that

astafan8: you're right that the example before wouldn't work, however i thought that our use case is this:
```
In [10]: try:
    ...:     raise MyExceptionAlias
    ...: except MyException:
    ...:     print("Caught an exception")
    ...:
```
so, the "new" driver raises MyExceptionAlias, and old code is still catching MyException, not the other way around. if that's the case, then with the new driver, both old MyException and the new MyExceptionAlias can be except-ed.

-----

but `MyExceptionAlias = MyException` this is even simpler :)

jenshnielsen: We could do that but then RigolDS4000TraceNotReady should be a subclass of TraceNotReady since RigolDS4000TraceNotReady is the one that we raise

astafan8: a, yes, you're right. confusing subclassing))

PR #4546, opened 2022-08-24
Mypy: Enable no implicit re export

jenshnielsen:
* Avoid imports from pyvisa/parso that are imported from modules where they are not defined/added to `__all__`
* Update internal updated to either use public api or original definition

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -164,10 +164,7 @@ warn_unused_ignores = true
 # this triggers failures due to returning
 # values from untyped 3th party libs
 warn_return_any = false
-# this happens both internally
-# and in external libs that dont use
-# __all__ in their public reexported api
-no_implicit_reexport = false
+no_implicit_reexport = true
 strict_equality = true
 strict_concatenate = true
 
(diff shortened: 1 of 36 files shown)


jenshnielsen:
This seems to trigger a type issue with python 3.7 / old numpy versions

astafan8 (approved):

jenshnielsen:
This is waiting for us to drop 3.7 support #3889

qcodes/tests/drivers/test_keithley_26xx.py
```diff
@@ -4,19 +4,15 @@
 import pytest
 
 import qcodes.instrument.sims as sims
-from qcodes.instrument_drivers.tektronix.Keithley_2600_channels import (
-    Keithley_2600,
-    MeasurementStatus,
-)
+from qcodes.instrument_drivers.Keithley import Keithley2614B
+from qcodes.instrument_drivers.Keithley._Keithley_2600 import MeasurementStatus
```
jenshnielsen, line 8: This should be added to the public api

jenshnielsen: https://github.com/QCoDeS/Qcodes/pull/4687

PR #4569, opened 2022-09-02
Don't show progressbars when upgrading an empty db

### LINKED ISSUE #4566: Allow silencing the 'Upgrading database` calls
I am processing and setting up >100.000 QCoDeS databases concurrently (in 256 threads). Every single time a new database is created I am presented with ≈10 lines like
> Upgrading database; v6 -> v7:   0%|       | 0/1 [00:00<?, ?it/s]

Eventually the output grew so large it crashed my browser 😅

I see that the progress bar just prints to stdout

https://github.com/QCoDeS/Qcodes/blob/fe24a9ac07659b4d8f9f13b2481ecdd76ea0d0fa/qcodes/dataset/sqlite/db_upgrades/__init__.py#L151

Ideally the verboseness of these upgrade should be configurable.

In the meantime, what is the recommended way to capture and trash this output? Just [`contextlib.redirect_stdout`](https://docs.python.org/3.10/library/contextlib.html#contextlib.redirect_stdout)?

I have tried it with:
```python
import os
import sys
import contextlib
import qcodes as qc

import tempfile

names = tempfile._get_candidate_names()

with open(os.devnull, 'w') as f, contextlib.redirect_stdout(f):
    with qc.initialised_database_at(f"/tmp/{next(names)}"):
        "Do the things here"
```

but this gives some tqdm errors :(

jenshnielsen:
This should help mitigate the issue in #4566 by not showing the progress bar when creating a new db.

Fixes #4566

### DIFF
--- a/docs/changes/newsfragments/4969.improved
+++ b/docs/changes/newsfragments/4969.improved
@@ -1,0 +1,2 @@
+The Dataset no longer prints upgrade progressbars when upgrading an empty database. This means that
+the progressbar will no longer be shown when creating a new database.
--- a/qcodes/dataset/sqlite/db_upgrades/__init__.py
+++ b/qcodes/dataset/sqlite/db_upgrades/__init__.py
@@ -15,7 +15,7 @@ principle, the upgrade functions should not have dependecies from
 import logging
 import sys
 from functools import wraps
-from typing import Any, Callable, Dict
+from typing import Callable, Dict
 
 import numpy as np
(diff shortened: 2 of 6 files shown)


Dominik-Vogel (approved):

jenshnielsen:
bors merge

basnijholt:
Thanks @jenshnielsen, this is much better! 😄

PR #4580, opened 2022-09-06
Use Towncriers new config format

jenshnielsen:
The old one is deprecated and this is less verbose and nicer.

Note that the format documented in the readme is actually wrong https://github.com/twisted/towncrier/pull/425


Dominik-Vogel (approved):

jenshnielsen:
bors merge

PR #4601, opened 2022-09-12
add precommit hook to ensure no case conflicts

jenshnielsen:
With the landing of #4248 we no longer have case conflicts in the QCoDeS codebase. Lets ensure we don't get them again


astafan8 (approved):
yes!

astafan8 (approved):
yes!

PR #4325, opened 2022-06-29
Extend do_nd with option to perform "Parallel sweeps"

jenshnielsen:
In the process refactor Sweeps and Measurements to be a bit more modular

closes #4366 by refactoring the setpoint logic to generate the input type. 

- [x] More robust parsing of Parameters (see todos in code)
- [x] Document
- [x] More complete tests (there should be a test that this can work around a parameterwithsetpoints (and perhaps multi/arrayparameter)

PR #4605, opened 2022-09-12
split dond out from do0d ...

jenshnielsen:
This module has grown rather large so its easier to work with if we split it up. Taken from https://github.com/QCoDeS/Qcodes/pull/4325


astafan8 (approved):
much better!

agreed offline to put all do_*d modules into qcodes.dataset.dond submodule/folder, and also extract sweep classes into their own submodule inside qcodes.dataset.dond

astafan8 (approved):

PR #4608, opened 2022-09-13
Update Keysight drivers to conform to new standard

jenshnielsen:


qcodes/instrument_drivers/Keysight/Keysight_N5173B.py
```diff
@@ -1,5 +1,11 @@
 from qcodes.instrument_drivers.Keysight.Keysight_N5183B import N5183B
```
astafan8, line 1: this should use the new name, right?
```suggestion
from qcodes.instrument_drivers.Keysight.Keysight_N5183B import KeysightN5183B
```

qcodes/instrument_drivers/Keysight/N52xx.py
```diff
@@ -353,6 +357,11 @@ def _set_Sparam(self, val: str) -> None:
         self.write(f"CALC:PAR:MOD:EXT \"{val}\"")
 
 
+
+# PNATrace =KeysightPNATrace
```
astafan8, line 361:
```suggestion
PNATrace = KeysightPNATrace
```

astafan8 (approved):

qcodes/instrument_drivers/Keysight/Keysight_N5173B.py
jenshnielsen: That would indeed be more forward looking

PR #4623, opened 2022-09-15
Minicircuits 4SPDT driver fix for .NET 4.5+ versions

ThorvaldLarsen:
There are two different dll files depending on the version of .NET on the system:
.Net 4.5 and above: mcl_RF_Switch_Controller_NET45.dll
Earlier .Net versions: mcl_RF_Switch_Controller64.dll

Had a quick look but couldn't find an easy way to check the .NET version from python although would expect it should be possible. The most rudimentary solution is to just try importing both and renaming the package in python as shown suggested here but would be nice with a cleaner implementation.

### DIFF
--- a/qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
+++ b/qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
@@ -70,7 +70,7 @@ class USB_SPDT(SPDT_Base):
             )
         try:
             import mcl_RF_Switch_Controller64 as mw_driver  # pyright: ignore[reportMissingImports]
-        except:
+        except ImportError:
             import mcl_RF_Switch_Controller_NET45 as mw_driver  # pyright: ignore[reportMissingImports]
 
         self.switch = mw_driver.USB_RF_SwitchBox()
--- a/qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
+++ b/qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
@@ -68,8 +68,12 @@ class USB_SPDT(SPDT_Base):
                 the dll to open properties and check the 'unblock' checkmark
                 in the bottom. Check that your python installation is 64bit."""
             )
-        import mcl_RF_Switch_Controller64  # pyright: ignore[reportMissingImports]
-        self.switch = mcl_RF_Switch_Controller64.USB_RF_SwitchBox()
+        try:
+            import mcl_RF_Switch_Controller64 as mw_driver  # pyright: ignore[reportMissingImports]
+        except:
+            import mcl_RF_Switch_Controller_NET45 as mw_driver  # pyright: ignore[reportMissingImports]
+
+        self.switch = mw_driver.USB_RF_SwitchBox()
 
         if not self.switch.Connect(serial_number):
             raise RuntimeError('Could not connect to device')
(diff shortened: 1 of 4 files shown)


qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
```diff
@@ -66,8 +66,12 @@ def __init__(self, name: str, driver_path: Optional[str] = None,
                 the dll to open properties and check the 'unblock' checkmark
                 in the bottom. Check that your python installation is 64bit."""
             )
-        import mcl_RF_Switch_Controller64
-        self.switch = mcl_RF_Switch_Controller64.USB_RF_SwitchBox()
+        try:
+            import mcl_RF_Switch_Controller64 as mw_driver
+        except:
```
guenp, line 71: Could we add the specific type of exception that we want to catch here for clarity?

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
```diff
@@ -66,8 +66,12 @@ def __init__(self, name: str, driver_path: Optional[str] = None,
                 the dll to open properties and check the 'unblock' checkmark
                 in the bottom. Check that your python installation is 64bit."""
             )
-        import mcl_RF_Switch_Controller64
-        self.switch = mcl_RF_Switch_Controller64.USB_RF_SwitchBox()
+        try:
+            import mcl_RF_Switch_Controller64 as mw_driver
```
astafan8, line 70: @ThorvaldLarsen when using the net45 one, don't you also need to pass the driver_path explicitly, so that it takes the passed path to DLL? what i have in mind is that we could take the suggested simple solution, and first try importing the net45 driver from the net45 dll and, if that fails, fall back to the "64" one.

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
```diff
@@ -66,8 +66,12 @@ def __init__(self, name: str, driver_path: Optional[str] = None,
...
+        except:
+            import mcl_RF_Switch_Controller_NET45 as mw_driver
+
+        self.switch = mw_driver.USB_RF_SwitchBox()
```
astafan8, line 74: actually, since the name of the import changes depending on the dll, we should probably add an argument to the driver for that name as well? such that if there's yet another new dll with yet a new name to import, that changes to the driver code are not necessarily needed.

but perhaps that's too much

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
guenp: I think the try-except statement is Pythonic here so would be fine to go with that, it's just good practice to also catch the actual exception. It could be that this import throws a different exception, even on a system that has the older .NET installed and that will then fail in an unclear way because we didn't catch the exception.

qcodes/instrument_drivers/Minicircuits/USB_SPDT.py
ThorvaldLarsen: I think it is fine to just check for the two names that are in use. Seems unlikely that they are going to make yet another file. Would assume it is pretty rare that .NET binaries needs to be updated.

jenshnielsen: ```mcl_RF_Switch_Controller_NET45``` needs to be added to the list of modules that we don't have types for in `pyproject.toml``` next to `mcl_RF_Switch_Controller64`

guenp:
@jenshnielsen and I just discussed, it looks like @ThorvaldLarsen won't have time to finish it up so Jens will take care of it. Thanks all!

jenshnielsen:
@astafan8 as agreed with @guenp I have updated this pr with the agreed changes. Could you have a quick look

astafan8 (approved):

PR #4627, opened 2022-09-16
Type annotate sweeps with dtype

jenshnielsen:
This makes it clear that

* Lin and Log sweep will always generate floats
* Array sweeps input type maps to output type

This requires numpy 1.21 so bump the version


astafan8 (approved):
that's cool!

and so in arraysweep the dtype of the given array is propagated now? :)

jenshnielsen:
> and so in arraysweep the dtype of the given array is propagated now? :)

Yes if constructed from a numpy array. I could not find a good way to make it propagate from a sequence. The typevar in `Sequence[T]` is a basic python type but the typevar in `npt.NDArray[K]` is a numpy dtype (subclass of np.generic)  

Note that numpy as also not always that great at figureing out dtypes. E.g. if you do

```
a: npt.NDArray[np.int32] = np.array([1,2,3])
ArraySweep(a)
```
then it will work 
but with 
```
a = np.array([1,2,3])
ArraySweep(a)
```
the dtype is unknown. This will hopefully get better as numpy types are improved

jenshnielsen:
bors merge

PR #4640, opened 2022-09-21
typecheck tqdm against stubs

jenshnielsen:
Removes one more package from the list of not typed packages.

Inspired by the list in https://github.com/python/mypy/pull/13698


jenshnielsen:
bors merge

PR #4641, opened 2022-09-21
Alazar generic output must be input

jenshnielsen:
With the upcoming mypy 0.980 mypy will enforce that a generic Typevar output is matched by a generic input such that it can infer the type of the output. This reveals that `acquire` would return an unbound typevar. This fixes that by matching it to the type of the ```acquisition_controller ```

Note that the code is typed to accept acquisition_controller being None but that will actually unconditionally raise. 

```python
       if acquisition_controller is None:
            raise RuntimeError("Cannot call acquire without an "
                               "acquisition_controller")
```

Also add `from __future__ import annotations` to make types cleaner

### DIFF
--- a/qcodes/instrument_drivers/AlazarTech/ATS.py
+++ b/qcodes/instrument_drivers/AlazarTech/ATS.py
@@ -1,22 +1,12 @@
+from __future__ import annotations
+
 import ctypes
 import logging
 import sys
 import time
 import warnings
 from contextlib import contextmanager
-from typing import (
-    Any,
-    Dict,
-    Generic,
-    Iterator,
-    List,
-    Optional,
-    Sequence,
-    Type,
-    TypeVar,
-    Union,
-    cast,
-)
+from typing import Any, Generic, Iterator, Sequence, Type, TypeVar, Union, cast
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -584,14 +580,14 @@ def acquire(
         # return result
         return acquisition_controller.post_acquire()
 
-    def _set_if_present(self,
-                        param_name: str,
-                        value: Union[int, str, float, None]) -> None:
+    def _set_if_present(self, param_name: str, value: int | str | float | None) -> None:
```
astafan8, line 583:
```suggestion
    def _set_if_present(self, param_name: str, value: str | float | None) -> None:
```

qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -584,14 +580,14 @@ def acquire(
...
         if value is not None:
             parameter = self.parameters[param_name]
             parameter.set(value)
 
-    def _set_list_if_present(self, param_base: str, value: Sequence[Union[int, str, float]]) -> None:
+    def _set_list_if_present(
+        self, param_base: str, value: Sequence[int | str | float]
```
astafan8, line 589:
```suggestion
        self, param_base: str, value: Sequence[str | float]
```

astafan8 (approved):

jenshnielsen:
bors merge

PR #4658, opened 2022-09-27
DataSet.get_metadata to return value only if the casing of the tag matches the column name in the database

astafan8:
``DataSet.get_metadata`` method for retrieving a metadata value for a
given tag is now case-sensitive with respect to the tag.
For example, if metadata was added with ``dataset.add_metadata('something', 1)``,
it can only be retrieved by using its exact casing of the tag,
``dataset.get_metadata('something')``, and not e.g.
``dataset.get_metadata('SomeThinG')``. In the previous versions of QCoDeS,
any casing of the tag in ``DataSet.get_metadata`` would work and return the
value stored under that tag. Note that this change brings consistency
with how getting metadata via the ``dataset.metadata`` works:
``dataset.metadata['something']`` does return the value, and
``dataset.metadata['SomeThinG']`` does not.

Also adds extra testing for ``one()`` helper function and clarifies docstrings of a few helper functions.

### DIFF
--- a/docs/changes/newsfragments/4658.breaking
+++ b/docs/changes/newsfragments/4658.breaking
@@ -1,0 +1,11 @@
+``DataSet.get_metadata`` method for retrieving a metadata value for a
+given tag is now case-sensitive with respect to the tag.
+For example, if metadata was added with ``dataset.add_metadata('something', 1)``,
+it can only be retrieved by using its exact casing of the tag,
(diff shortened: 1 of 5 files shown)


jenshnielsen (approved):

jenshnielsen (approved):

astafan8:
bors merge

PR #4664, opened 2022-09-29
Mark two clases with test in the name as not tests

jenshnielsen:
This fixes a warning from pytest that these cannot be collected as tests


astafan8 (approved):

jenshnielsen:
bors merge

PR #4688, opened 2022-10-04
Add a callback to get_parameter_data to follow data loading

edumur:
Hi all,

In the effort to use qcodes as the data loader of the [pyplotter](https://github.com/pyplotter/pyplotter) I added a callback to `get_parameter`. This is usefull to track the progress of the data download.
Since sqlite3 does not allow to keep track of the data loading progress, we compute how many sqlite requests correspond to a certain percentage of progress which is dictated by a config parameter. Then we perform `x` sql request instead of one, running the callback everytime. 

The whole process add an overhead of ~13% on my laptop when used which is not negligible.
However, I would argue that keeping track of the download progress for a GUI is precious and is worth the overhead.

To test the download time
```python
import qcodes as qc
from time import time

def callback(progress):
    # print(progress)
    pass

qc.initialise_or_create_database_at('test.db')

start = time()
qc.load_by_id(2).get_parameter_data('phase')
stop = time()
dt1 = stop-start
print(dt1)
start = time()
qc.load_by_id(2).get_parameter_data('phase', callback=callback)
stop = time()
dt2 = stop-start
print(dt2)

print(dt2/dt1)
```

I kept the change minimal hoping other people will have comments.


edumur:
Hi all,

I am writing to keep the thread and see if anyone has comments on it.

See you.

qcodes/dataset/sqlite/queries.py
```diff
@@ -456,18 +467,62 @@ def get_parameter_tree_values(
...
+        # a progress of Config.callback_percent
+
+        # First, we get the number of dependent parameters
+        rd = get_rundescriber_from_result_table_name(conn, result_table_name)._to_dict()
+        # New qcodes
+        if 'interdependencies_' in rd.keys():
+            nbParamDependent = len(rd['interdependencies_']['dependencies'])
+        # Old qcodes
+        else:
+            nbParamDependent = len([i for i in rd['interdependencies']['paramspecs'] if i['name']==toplevel_param_name][0]['depends_on'])
```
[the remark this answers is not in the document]
edumur: Yep, my bad.

qcodes/configuration/config.py
```diff
@@ -80,6 +80,10 @@ class Config:
     defaults_schema: DotDict
     """The default schema"""
 
+    callback_percent: int = 5
+    """If user wants to callback a function while loading data, the callback is
+    done every callback_percent"""
```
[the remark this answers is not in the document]
edumur: My bad, it is done now.

qcodes/dataset/data_set_protocol.py
```diff
@@ -221,6 +221,7 @@ def get_parameter_data(
         *params: str | ParamSpec | ParameterBase,
         start: int | None = None,
         end: int | None = None,
+        callback: Callable | None = None,
```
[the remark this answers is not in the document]
edumur: Yep, done.

qcodes/dataset/sqlite/query_helpers.py
```diff
@@ -44,7 +44,8 @@ def one(curr: sqlite3.Cursor, column: int | str) -> Any:
         return res[0][column]
 
 
-def many(curr: sqlite3.Cursor, *columns: str) -> list[Any]:
+def many(curr: sqlite3.Cursor,
+         columns: str | List[str] | Tuple[str]) -> list[Any]:
```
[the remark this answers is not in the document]
edumur: Thanks, I will change the typing.
Also it looks like:
```python
    if isinstance(columns, str):
        columns = [columns]
```

Should be added after `res = curr.fetchall()`

qcodes/dataset/sqlite/query_helpers.py
```diff
@@ -60,19 +61,29 @@ def many(curr: sqlite3.Cursor, *columns: str) -> list[Any]:
         return [res[0][c] for c in columns]
 
 
-def many_many(curr: sqlite3.Cursor, *columns: str) -> list[list[Any]]:
+def many_many(curr: sqlite3.Cursor,
+              columns: str | List[str] | Tuple[str],
```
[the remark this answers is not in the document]
edumur: Thanks, I will change the typing.

For the other question of the API, I tried to used the old code as much as possible, hence the modification of `many` and `many_many`.
I guess it can be done another way.

edumur:
[test.zip](https://github.com/QCoDeS/Qcodes/files/9870933/test.zip)
Please find the test database and the test file so that you can see the difference with and without callback.

qcodes/dataset/sqlite/queries.py
```diff
@@ -482,15 +537,31 @@ def get_parameter_tree_values(
...
+        progress = 0.
+        callback(progress)
+        for i in range(len(offset)-1):
+            sql = f"""
+                  SELECT {columns_for_select}
+                  FROM {sql_subquery}
+                  LIMIT {limit} OFFSET {offset[i]}
+                  """
+            cursor.execute(sql)
+            res = many_many(cursor, columns, res)
```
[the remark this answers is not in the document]
edumur: Sorry, I do no understand your suggestion.
Could you explain it a little bit more?
Thanks!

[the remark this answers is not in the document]
edumur: I think I understood your question.
I benchmark the code of the PR and compare it to this one:
```python
res = []
progress = 0.
callback(progress)
for i in range(len(offset)-1):
    sql = f"""
            SELECT {columns_for_select}
            FROM {sql_subquery}
            LIMIT {limit} OFFSET {offset[i]}
            """
    cursor.execute(sql)
    res_temp = many_many(cursor, columns)
    for i in res_temp:
        res.append(i)
    progress += iteration
    callback(progress)
```
where the appending is done outside `many_many`.

the result is:
image

I would conclude it takes the same time.
So if you wish, I can make `many` and `many_many` as before and do the appending in `get_parameter_tree_values`.

qcodes/dataset/sqlite/queries.py
```diff
@@ -223,6 +228,7 @@ def get_shaped_parameter_data_for_one_paramtree(
     output_param: str,
     start: int | None,
     end: int | None,
+    callback: Callable | None,
```
jenshnielsen, line 231: Please update remaining callbacks to the full signature

qcodes/dataset/sqlite/queries.py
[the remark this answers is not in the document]
jenshnielsen: Yes please change them back, my main concern is that users outside of qcodes may be using these functions using the old signature

edumur:
All should be done.
I had trouble incorporating Florian Blanchet changes from commit b476b7935c636937a3b8671bdda16cee0133edde, I hope everything is alright.

When I benchmarked the merged code I saw that:
1. SQL request where 2 to 3 times faster than before the merging (maybe an effect of Florian's work).
2.  The overhead increased from 20% to 50% while using the callback.

image

jenshnielsen:
@edumur Looks like some mypy issues have creapt in. will you be able to have a look?

edumur:
I had to add `callback: Callable[[float], None] | None = None,` in `get_parameter_data` in `data_set_in_memory`.
But I do not know what I did here so please someone check.

docs/changes/newsfragments/4688.new
```diff
@@ -0,0 +1,6 @@
+Add a callback to 1get_parameter1.
```
jenshnielsen, line 1:
```suggestion
Add a callback to ``dataset.get_parameter_data``.
```

qcodes/dataset/sqlite/queries.py
```diff
@@ -461,27 +472,97 @@ def get_parameter_tree_values(
...
+
     offset = max((start - 1), 0) if start is not None else 0
     limit = max((end - offset), 0) if end is not None else -1
 
     if start is not None and end is not None and start > end:
         limit = 0
 
+    # start and end currently not working with callback
+    if start is None and end is None and callback is not None:
+
```
jenshnielsen, line 495: Looking at this again I am a bit confused about the calculations. Max(ID) will return the number of rows in the table and each row is fetched as one entry so why is the need to incorporate the number of dependent parameters in the calculation?

jenshnielsen: I think it would also make the code somewhat clear if we break this logic to calculate offset out into one or more functions

qcodes/dataset/sqlite/query_helpers.py
```diff
@@ -6,7 +6,7 @@
 
 import itertools
 import sqlite3
-from typing import Any, List, Mapping, Sequence, Union
+from typing import Any, List, Mapping, Sequence, Tuple, Union
```
jenshnielsen, line 9: Please remove this unused import

jenshnielsen:
The types looks good. Left a few comments inline. There is now a test.db file which should not be committed

qcodes/dataset/sqlite/query_helpers.py
edumur: Done

docs/changes/newsfragments/4688.new
edumur: Done

qcodes/dataset/sqlite/queries.py
edumur: My understanding is that when a user downloads a certain parameter, a lot of rows are useless to him since containing `NULL` values:
image
Thus the number of data points is the total number of row divided by the number of dependent parameters, the  number of column.
That ratio `int(max_id/nb_param_dependent)` gives then the true number of row being downloaded.

The main idea here is that we are divided one big SQL request into many smaller SQL requests.
By doing so, we need to keep track of how many request are being done and where we are in that process.
By using `LIMIT` and `OFFSET`, we can do so but need the true number of downloaded rows.

jenshnielsen: That sounds correct if each parameter has been measured the same number of times but need not be case. Ideally we want to know the number of rows where dependent_parameter is not Null. One could perhaps use count to get the number of rows where dependent_parameter is not Null. See https://www.w3schools.com/sql/sql_count_avg_sum.asp

edumur: Yes, I think you're right.
I have break down the logic into smaller functions and change the way I count the number of row to be downloaded.

edumur:
Hi all,

All things seem correct (I think).
I do not understand the "Lint with Darker" error and I do not know what it is checking ^^'.

jenshnielsen:
Darker runs the black formatter on only the part of the code that you have changed. We do this because we would like to move to a more consistent formatting but not rewrite all out code in one go losing history. You can run these lints automatically by installing the pre-commit hooks defined in https://github.com/QCoDeS/Qcodes/blob/master/.pre-commit-config.yaml using the pre-commit tool. https://pre-commit.com/ For now don't worry about it and I will run it when merging

qcodes/tests/dataset/test_dataset_loading.py
```diff
@@ -357,3 +361,17 @@ def create_ds_with_exp_id(exp_id):
 
     empty_guid_list = get_guids_by_run_spec(conn=conn, experiment_name="nosuchexp")
     assert empty_guid_list == []
+
+
+def test_callback(scalar_dataset: DataSet) -> None:
+    called_progress = []
```
jenshnielsen, line 367:
```suggestion
    called_progress: list[float] = []
```

qcodes/tests/dataset/test_dataset_loading.py
```diff
@@ -358,3 +361,17 @@ def create_ds_with_exp_id(exp_id):
...
+
+    def callback_closure(called_progress: list[float]):
+        def callback(progress: float) -> None:
+            called_progress.append(progress)
+
+        return callback
+
+    scalar_dataset.get_parameter_data(callback=callback_closure(called_progress))
+    # TODO this should only be called upto 100.0 not 200.0
+    assert called_progress == list(np.arange(0.0, 201.0, 5.0))
```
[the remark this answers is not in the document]
edumur: Yep,

Thanks to have caught that...
I left a hard coded 2 from my local test.
I replace that by the proper operation (hopefully!)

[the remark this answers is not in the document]
jenshnielsen: I have extended the test to cover multiple numbers of parameters as well as rows less than 100. This has uncovered another bug as the code that generates the callbacks do not take into account the progress when there are less than 100 rows

qcodes/dataset/sqlite/queries.py
```diff
@@ -461,27 +552,65 @@ def get_parameter_tree_values(
...
+        assert isinstance(offset, np.ndarray)
+
+        # 0
+        progress = 0.0
+        callback(progress)
+
+        # 1
+        cursor.execute(sql, (limit, offset[0]))
+        res = many_many(cursor, *columns)
+        progress += config.dataset.callback_percent
```
jenshnielsen, line 603: This does not correctly take into account what happens when there are less than 100 rows

edumur:
I really appreciate you spending time checking this.
I updated the code so that it works now.

I tested it with various database and various percentage and so far it worked 🤞.

qcodes/tests/dataset/test_dataset_loading.py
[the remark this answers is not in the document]
edumur: Yep, should be fixed

qcodes/dataset/sqlite/queries.py
edumur: Yep, should be fixed

qcodes/tests/dataset/test_dataset_loading.py
```diff
@@ -357,3 +360,22 @@ def create_ds_with_exp_id(exp_id):
...
+
+        return callback
+
+    scalar_datasets_parameterized.get_parameter_data(
+        callback=callback_closure(called_progress)
+    )
+    if len(scalar_datasets_parameterized) > 100:
+        assert called_progress == list(np.arange(0.0, 101.0, 5.0))
+    else:
+        assert called_progress == [0.0, 5.0]
```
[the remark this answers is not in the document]
edumur: Yep, should have been correctly updated

jenshnielsen (approved):

Issue #4712, opened 2022-10-10
Random test failure in monitor tests in ci (windows)


jenshnielsen:
Seen fairly frequently on windows.

```
=========================== short test summary info ===========================
FAILED qcodes/tests/test_monitor.py::test_setup_teardown - RuntimeError: Fail...
FAILED qcodes/tests/test_monitor.py::test_monitor_replace - RuntimeError: Fai...
FAILED qcodes/tests/test_monitor.py::test_double_join - RuntimeError: Failed ...
FAILED qcodes/tests/test_monitor.py::test_connection - RuntimeError: Failed t...
FAILED qcodes/tests/test_monitor.py::test_use_root_instrument[True] - Runtime...
FAILED qcodes/tests/test_monitor.py::test_use_root_instrument[False] - Runtim...
ERROR qcodes/tests/test_monitor.py::test_parameter - RuntimeError: Failed to ...
```

```
=================================== ERRORS ====================================
______________________ ERROR at setup of test_parameter _______________________
[gw1] win32 -- Python 3.10.7 C:\hostedtoolcache\windows\Python\3.10.7\x64\python.exe

    @pytest.fixture(name="inst_and_monitor")
    def _make_inst_and_monitor():
        instr = DummyInstrument("MonitorDummy")
        param = Parameter("DummyParam",
                          unit="V",
                          get_cmd=None,
                          set_cmd=None)
        param(1)
        monitor_parameters = tuple(instr.parameters.values())[1:]
>       my_monitor = monitor.Monitor(*monitor_parameters, param, interval=0.1)

qcodes\tests\test_monitor.py:27: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <Monitor(Thread-10, stopped 4404)>, interval = 0.1
use_root_instrument = True
parameters = (<qcodes.parameters.parameter.Parameter: dac1 at 2660446451168>, <qcodes.parameters.parameter.Parameter: dac2 at 26604...ters.parameter.Parameter: dac3 at 2660446451264>, <qcodes.parameters.parameter.Parameter: DummyParam at 2660446442384>)
parameter = <qcodes.parameters.parameter.Parameter: DummyParam at 2660446442384>

    def __init__(
        self,
        *parameters: Parameter,
        interval: float = 1,
        use_root_instrument: bool = True,
    ):
        """
[...]

PR #4723, opened 2022-10-13
Monitor simplify eventloop handling

jenshnielsen:
This makes use of `asyncio.run` and async context managers to manage the state of the context manager. 

We use a future + call_soon_threasafe to signal that the loop should exit as recommended by the websockets docs

Note that this also fixes a deprecation warning from asyncio

- [x]  Verify that this still works with the min supported version
- [x] Test manually against gui

May close https://github.com/QCoDeS/Qcodes/issues/4712

### DIFF
--- a/qcodes/monitor/monitor.py
+++ b/qcodes/monitor/monitor.py
@@ -93,7 +93,10 @@ def _handler(
     """
     Return the websockets server handler.
     """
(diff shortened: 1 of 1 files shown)


jenshnielsen:
@astafan8 Manually verified working both against earliest supported version and newest version. I think this is ready for review

astafan8 (approved):
looks clean! :)

jenshnielsen:
bors merge

PR #4736, opened 2022-10-18
Add Exception to except

nulinspiratie:
This PR changes an except statement in `InstrumentBase.snapshot_base` from `except:` to `except Exception`.
This allows other errors, e.g. a `KeyboardInterrupt`, to halt the snapshot.
Without this modification, a snapshot can take a long time if there is a communication error.


jenshnielsen (approved):
thanks, Indeed this would capture a KeyboardInterupt which we probably don't want here

jenshnielsen:
bors merge

jenshnielsen:
For reference this is since a bare except will capture any subclass of BaseException and importantly KeyboardInterrupt is a BaseException but not an Exception

jenshnielsen:
bors merge

Issue #4695, opened 2022-10-05
dond to support time-dependence measurements (perhaps with ElapsedTimeParameter)


astafan8:
first of all, ElapsedTimeParameter does not have a set method, so it can't be used in dond as an independent parameter. Secondly, the ask is to run a measurement until elapsed time is more than a specified value (e.g. until elapsed time is more than 10 minutes).

Originally requested by @guenp @samantha-ho

guenp:
Thanks @astafan8 for creating this issue!
The reason why I initially thought it would be great for `dond` to support time as a parameter is because in the users' mind, time is just another dimension along which we want to measure a set of parameters or rasters. I'm not fixated on using `dond` for this if it is not within the package's design principles, but the convenience of being able to set up a measurement with `dond`, then decide "What if I monitor this thing over time"? and only having to make a small change will be great to be able to cover somehow.

astafan8:
Another request for this just came in: for every value of an outer parameter, measure some other parameter as a function of time by acquiring a value every 10 seconds for 5 minutes. I suggested the following solution (i'll come back with updates on how this worked):

```python
elapsed_time_param = ElapsedTimeParameter(...)

some_outer_sweep = LinSweep(
    ...,
    # reset the elapsed time parameter
    # for every new value of this sweep
    post_actions = (elapsed_time.reset,),
)

time_sweep = LinSweep(
    elapsed_time_param,
    start=0,  # value does not matter
    stop=5,  # 5 minutes, value does not matter
    num_points=5*60/10,  # 5 mins / 10 seconds, number of points matter
    delay=10,  # seconds; this is what "sweeps" the time
    get_after_set=True,  # thanks to #4759 
)

dond(some_outer_sweep, time_sweep, measure_something_param, ...)
```

guenp:
@astafan8, this issue just came up again. This is the workaround that @ThorvaldLarsen came up with:
```python
time_param = ElapsedTimeParameter('time')
time_param._settable = True
time_param.set = lambda x: None
time_sweep = LinSweep(time_param, 1, 1000, 1000, 0, get_after_set=True)

time_param.reset_clock()
ds, _, _ = dond(time_sweep, *params_to_measure)
```
This solution is not ideal because we have to pick the number of repetitions based on how long we want the measurement to run, instead of just specifying "please measure for 10 minutes".

We were wondering if it might make sense to implement something like this that behaves more like a `while` loop:
```python
time_param = ElapsedTimeParameter('time')
[...]

PR #4759, opened 2022-10-25
Add ability to get parameter after set to dond

jenshnielsen:
For some instruments it may be a good idea to get the value after setting it and store that in the dataset rather than relying on the instrument being able to set the exact value. 

This implements that by extending the Sweeps classes with a boolean. This could be extended to read a flag from the parameter giving the parameter a trait to indicate if it has been set

This is also a small step towards fixing #4695 since it will become possible to get the value of a parameter that one sets

To make this easier to demonstrate ManualParameters have been improved to log using the instrument logger

TODO 
- [x] Documentation (examples + docstrings)
- [x] Tests
- [x] release notes


astafan8 (commented):
very very nice idea! and also quite a small change as enabled by the earlier refactor :)

jenshnielsen:
@astafan8 ready for review

astafan8 (approved):
looks great :)

jenshnielsen:
bors merge

jenshnielsen:
bors merge

jenshnielsen:
bors merge

jenshnielsen:
bors cancel

Issue #4504, opened 2022-08-12
Fix issue with guid time resolution on windows


jenshnielsen:
And revert related xfailed tests in #4501 #4503

jenshnielsen:
fixed by #4763

PR #4763, opened 2022-10-25
Add random sample to guid

jenshnielsen:
The current time resolution, especially on windows, is insufficient to ensure that we always avoid collisions in guids. Since sample names are available within the dataset metadata and the sample part of the guid is cumbersome to use I suggest that we replace that with a random string.

The broad plan is 
* Make guid type configurable
* Warn if the sample name guid is set to something other than the default value. 
* Update tests that have problems with guid collisions to use the new guid type
* Eventually change the default to the new type

For now we ami to completely remove the support of setting a sample_id. If actual use shows up we are open to allow this to be enabled as an explicit option

### DIFF
--- a/qcodes/tests/dataset/test_guids.py
+++ b/qcodes/tests/dataset/test_guids.py
@@ -220,7 +220,7 @@ def test_random_sample_guid():
     cfg = qc.config
     cfg["dataset"]["GUID_type"] = "random_sample"
 
-    expected_guid_prefixes = ["d82c07cd", "6baa9455", "82e2e662"]
+    expected_guid_prefixes = ["d82c07ce", "629f6fbf", "c2094cad"]
     for expected_guid_prefix in expected_guid_prefixes:
         guid = generate_guid()
         assert guid.split("-")[0] == expected_guid_prefix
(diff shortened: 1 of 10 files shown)


jenshnielsen:
resolves https://github.com/QCoDeS/Qcodes/issues/4504

astafan8 (approved):
to the open question: i agree, it might not be worth the effort to have 3 options, my hope is that due to lack of advertizing not many users used the guid customization.

jenshnielsen:
Todo

- [ ] Add suitable warnings to loading by sample
- [ ] Check that docs are up to date

jenshnielsen:
@astafan8 should be ready for a real review

qcodes/tests/dataset/test_guids.py
```diff
@@ -17,18 +19,39 @@
...
+)
 def test_generate_guid(loc, stat, smpl):
     # update config to generate a particular guid. Read it back to verify
     cfg = qc.config
     cfg["GUID_components"]["location"] = loc
     cfg["GUID_components"]["work_station"] = stat
     cfg["GUID_components"]["sample"] = smpl
 
-    guid = generate_guid()
+    if smpl == 0 or smpl == 2_863_311_530:
```
jenshnielsen, line 46: should probably also clean this up as suggested by codacy

PR #4768, opened 2022-10-26
adding test duration comparison in CI

trevormorgan:
Adding action to record test duration comparison on serial run unit tests. Let's see if this adds any insight to our test flow.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -58,7 +58,7 @@ jobs:
 # a subset of the tests fails when run in parallel on Windows so run those in serial here
     - name: Run serial tests
       run: |
-        pytest -m "serial" -n 0 --dist no --cov=qcodes --cov-report xml --cov-append --hypothesis-profile ci --csv test_durations.csv --csv-columns name,duration qcodes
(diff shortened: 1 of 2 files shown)


.github/workflows/pytest.yaml
```diff
@@ -58,7 +58,12 @@ jobs:
 # a subset of the tests fails when run in parallel on Windows so run those in serial here
     - name: Run serial tests
       run: |
-        pytest -m "serial" -n 0 --dist no --cov=qcodes --cov-report xml --cov-append --hypothesis-profile ci qcodes
+        pytest -m "serial" -n 0 --dist no --cov=qcodes --cov-report xml --cov-append --hypothesis-profile ci qcodes --csv test_durations.csv --csv-columns name,duration
+    - name: Compare test duration time
+      uses: actions/upload-artifact@v3
+      with:
+        name: blueracer
+        path: test_durations.csv
```
jenshnielsen, line 66: @trevormorgan there are only 3 serial tests so I think it makes a lot more sense to do this on the parallel tests

trevormorgan: I'll switch to parallel test! Thanks!

jenshnielsen (approved):

PR #4770, opened 2022-10-27
Add text-wrapping for extra-long plot titles

samantha-ho:
This PR text-wraps long plot titles.

### DIFF
--- a/docs/changes/newsfragments/4770.improved
+++ b/docs/changes/newsfragments/4770.improved
@@ -1,0 +1,1 @@
+`plot_datset` will now correctly wrap long titles into multiple lines.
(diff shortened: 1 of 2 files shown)


trevormorgan:
@samantha-ho change looks good. You might consider undoing black on the file though and only running with darker. Qcodes CI is setup to check that formatting is only correct on lines that were modified. Unfortunately qcodes didn't initially enforce formatting, using darker is a nice middle ground and  keeps prs from being mostly formatting changes.

trevormorgan (approved):

jenshnielsen:
@samantha-ho @trevormorgan I think its ok to go either way here. For future prs we would like to stick to darker and not black the whole file. If we merge it as is I think we should squash merge it to one commit

jenshnielsen:
bors merge

PR #4772, opened 2022-10-28
Jupyter notebooks and RST as Markdown in MyST format

basnijholt:
Because Markdown is just easier to read and write, and Markdown notebooks are better with version control.


jenshnielsen:
Hi @basnijholt we have actually considered these things earlier but evaluated them to be less favorable than you do.

> Because Markdown is just easier to read and write, and Markdown notebooks are better with version control.

If we go this route the pr needs to be split into at least two different things. 

With respect to using MYST for non notebook docs rather than rst my initial impression of MYST last time I looked at it was that it's not so much markdown as strange dialect of markdown with a lot of non-standard syntax that I found worse than writing rst but I could be convinced otherwise. e.g. all of this embedding rst into the md files seems to defy the purpose. 

With respect to notebooks, I don't disagree with this statement, but this conversion does increase the barrier to entry for new users as they have to deal with how to visualize these notebooks which is why we have explicitly chosen to not go this route. 
Another issue is that it looks like this discards all the output from the notebooks. While this may be fine in most cases QCoDeS is a bit unusual in that we have a lot of notebooks that can only run with specific instrument setups and we do not want to loose the output of those 

IMHO the merge conflicts are rarely an actual problem but the two other things are potential problems.

basnijholt:
> Another issue is that it looks like this discards all the output from the notebooks. While this may be fine in most cases QCoDeS is a bit unusual in that we have a lot of notebooks that can only run with specific instrument setups and we do not want to loose the output of those

Very good point. 

> all of this embedding rst into the md files seems to defy the purpose.

That is because I used an automated to to make the conversions. There is a more efficient syntax for most things.

PR #4783, opened 2022-11-01
Add basic read only mode to connect

jenshnielsen:
Note we probably want to make use of this in lots of places. However, I think its fine to merge this as is and then usage can be improved later. However, at a minimum the following should be done.

- [x] load_by ... should expose a read_only flag so its easy for data analysis + plotting to load as read only.
- [x] ~~Figure out how read only is exposed to the dataset. Ensure that adding metadata via call to `ds.add_metadata` on a read only dataset gives some kind of reasonable error.~~ Won't implement since the error from sqlite is informative enough.

### DIFF
--- a/src/qcodes/dataset/data_set.py
+++ b/src/qcodes/dataset/data_set.py
@@ -260,7 +260,7 @@ class DataSet(BaseDataSet):
             # raise valueerror here because if no run id, a new dataset will be created
             # and it will be written to the database
             raise ValueError(
-                "Cannot open database in read-only mode without a run_id provided"
+                "Cannot instantiate a dataset in read-only mode without a run_id provided"
                 " since a new dataset will be created."
             )
         self.conn = conn_from_dbpath_or_conn(conn, path_to_db, read_only=read_only)
@@ -272,6 +272,7 @@ class DataSet(BaseDataSet):
         self._cache: DataSetCacheWithDBBackend = DataSetCacheWithDBBackend(self)
         self._results: list[dict[str, VALUE]] = []
         self._in_memory_cache = in_memory_cache
+        self._read_only = read_only
 
         if run_id is not None:
             if not run_exists(self.conn, run_id):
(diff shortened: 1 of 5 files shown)


qcodes/dataset/sqlite/database.py
```diff
@@ -137,8 +140,14 @@ def connect(name: str | Path, debug: bool = False, version: int = -1) -> Connect
     # register binary(TEXT) -> numpy converter
     sqlite3.register_converter("array", _convert_array)
 
-    sqlite3_conn = sqlite3.connect(name, detect_types=sqlite3.PARSE_DECLTYPES,
-                                   check_same_thread=True)
+    path = f"file:{str(name)}"
+
+    if read_only:
+        path = path + "?ro"
```
astafan8: isn't it ``?mode=ro``? https://docs.python.org/3/library/sqlite3.html#how-to-work-with-sqlite-uris

astafan8: apparently for WAL-mode there are some extra concerns, see here https://www.sqlite.org/wal.html#read_only_databases - this feeds into the testing todo item

src/qcodes/dataset/data_set.py
```diff
@@ -583,9 +594,10 @@ def add_metadata(self, tag: str, metadata: Any) -> None:
             metadata: actual metadata
 
         """
+        if self._read_only:
+            raise RuntimeError("Cannot add metadata to a read-only dataset")
```
jenshnielsen: Could we add a test that checks that this error and the one below is raised when the add methods are called on a read only dataset.

jenshnielsen:
I think we should add a newsfragment in docs/changes/newsfragments

src/qcodes/dataset/data_set.py
einsmein: This code is removed.

src/qcodes/dataset/data_set.py
```diff
@@ -251,9 +252,18 @@ def __init__(
...
                 and available as part of the `dataset.cache` object.
+            read_only: whether to open the connection in read-only mode.
+                Only takes effect if `conn` is not given.
 
         """
-        self.conn = conn_from_dbpath_or_conn(conn, path_to_db)
+        if run_id is None and read_only:
+            # raise valueerror here because if no run id, a new dataset will be created
+            # and it will be written to the database
+            raise ValueError(
```
astafan8, line 262: Could you add a test for this error? And also a test of where read_only flag does nothing in case connection instance is passed explicitly.

astafan8 (approved):
Looks good! Some of the open comments are out-dated so can be closed. And a newsfragment is missing.

Issue #4543, opened 2022-08-23
qcodes import takes a long time


eendebakpt:
Importing qcodes with `import qcodes` takes a long time (> 2 seconds on the system tested). The import time can be analysed with tools like https://github.com/asottile/importtime-waterfall

Several components could perhaps not be loaded by default, or with a lazy loader (e.g. https://scientific-python.org/specs/spec-0001/, but there are more tools)

An example is the import `from opencensus.ext.azure.log_exporter import AzureLogHandler` from `qcodes.logger.logger`. By default `qc.config.telemetry.enabled` is False, so the import is executed at startup, but not used.

(more examples can be found by analysing the output of importtime-waterfall)

jenshnielsen:
@eendebakpt Which version of qcodes have you tested. I spend a fair amount of time improving this before 0.32 for example. For example the imports of opencensus azure are lazy and has been so since 0.32. See https://github.com/QCoDeS/Qcodes/commit/9f90a67607558689786e347665a2d2e9b1dad1bf

eendebakpt:
@jenshnielsen With qcodes updated to master the import of azure is indeed lazy. There is still room for improvement though.

Running:
```
python -c "import time; import xarray; import pandas; import scipy; t0=time.perf_counter(); import qcodes; dt=time.perf_counter()-t0; print(f'{1e3*dt:.2f} ms')"
```
results in a > 1 second import time of qcodes on my system. 

Some more candidates to reduce the import time:

* From the qcodes init:
```
from qcodes.instrument_drivers.test import test_instrument, test_instruments
```
Can be removed, or made lazy (for example with a mechanism like https://github.com/sympy/sympy/pull/23832). Tests should not impact runtime import performances.

* `qcodes.data.hdf5_format`. This is part of the legacy dataset, but still imported via

https://github.com/QCoDeS/Qcodes/blob/master/qcodes/parameters/array_parameter.py#L10


* The whole legacy dataset module is imported via

https://github.com/QCoDeS/Qcodes/blob/master/qcodes/dataset/__init__.py#L27

(this triggers also imports of xarray and scipy, maybe they would have been imported anyway)

* The qcodes.monitor triggers websockets. Perhaps import of `qcodes.monitor` can be removed from the `qcodes.__init__`

[...]

PR #4788, opened 2022-11-02
Make h5py and ruamel lazy loading

peendebak:
This PR uses the [lazy_loader](https://github.com/scientific-python/lazy_loader) package to make the qcodes import time lower. The lazy_loader package has been adapoted by several scientific python packages, for details see https://scientific-python.org/specs/spec-0001/

The h5py and ruamal components have been selected because they take time in the import and are only used in components that not all users will use (e.g. the legacy dataset and the Monitor). The PR is intended to be fully backwards compatible.

More gains can be made with non-backwards compatible changes. For example in the main qcodes `__init__.py` the number of direct imports which could be reduced.

Also see #4543

@jenshnielsen

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,7 @@ dependencies = [
     "importlib-resources>=2.0.0; python_version < '3.9'",
     "ipywidgets>=7.5.0,<9.0.0",
     "jsonschema>=3.0.0",
+    "lazy_loader>=0.1",
     "matplotlib>=3.3.0",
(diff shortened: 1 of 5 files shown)


jenshnielsen:
@peendebak Thanks, happy to try this out. I agree that when we make a breaking change it would be good to avoid the top level imports of corners of qcodes. I have been hoping to do that in an official 1.0 release at some point down the line

jenshnielsen (approved):

PR #4789, opened 2022-11-04
Allow pyvisa-sim instrs to be loaded from yaml file

jenshnielsen:
This makes it simpler to write examples of station yaml files that makes use of pyvisa-sim instruments 

The visalib argument to the driver needs to have the full path to the sim file so it is not practical to put in the file in a portable way. As a workaround, I have added a different argument that allows the users to supply the name of the file which is then looked up in the default bundled location. Happy for other suggestions to how this could be done

In the process I have added logic to allow the loading of a pyvisa-sim yaml file using importlib-resources in the recommended portable way. 

TODO

- [ ] Tests of the station loading
- [ ] Docstrings
- [ ] Update other driver tests to make use of the new loaded rather than the `__file__` hack
- [ ] Examples of this in docs?

### DIFF
--- a/qcodes/dist/schemas/station-template.schema.json
+++ b/qcodes/dist/schemas/station-template.schema.json
@@ -51,6 +51,9 @@
                         "init": {
                             "type": "object"
                         },
+                        "virtual_instrument_yaml_file": {
+                            "type": "string"
+                        },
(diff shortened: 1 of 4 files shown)


jenshnielsen:
closed in favour of #4790

PR #4797, opened 2022-11-09
Stop using __file__ outside tests

jenshnielsen:
This replaces the remaining use of `__file__` with `files` from `imporlib.resources` in qcodes outside of the tests.

Note that there is a trade off here as we now have hardcoded the path to config and monitor dist files in the python file rather than the relative path to that file. I think its cleaner to hardcode the full path but I could see the argument the other way around too.

### DIFF
--- a/qcodes/_version.py
+++ b/qcodes/_version.py
@@ -2,9 +2,9 @@ def _get_version() -> str:
     import sys
 
     if sys.version_info >= (3, 9):
-        from importlib.resources import as_file, files
+        from importlib.resources import files
     else:
-        from importlib_resources import as_file, files
+        from importlib_resources import files
 
     import versioningit
 
--- a/qcodes/_version.py
+++ b/qcodes/_version.py
@@ -1,11 +1,14 @@
 def _get_version() -> str:
-    from pathlib import Path
+    import sys
+
+    if sys.version_info >= (3, 9):
+        from importlib.resources import as_file, files
+    else:
+        from importlib_resources import as_file, files
 
     import versioningit
 
-    import qcodes
-
-    qcodes_path = Path(qcodes.__file__).parent
+    qcodes_path = files("qcodes")
     return versioningit.get_version(project_dir=qcodes_path.parent)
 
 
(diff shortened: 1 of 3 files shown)


qcodes/_version.py
```diff
@@ -1,11 +1,14 @@
 def _get_version() -> str:
-    from pathlib import Path
+    import sys
 
-    import versioningit
+    if sys.version_info >= (3, 9):
+        from importlib.resources import as_file, files
+    else:
+        from importlib_resources import as_file, files
```
astafan8, line 7: no need for as_file here
```suggestion
        from importlib.resources import files
    else:
        from importlib_resources import files
```

qcodes/configuration/config.py
```diff
@@ -40,14 +47,14 @@ class Config:
     schema_file_name = "qcodesrc_schema.json"
     """Name of schema file"""
     # get abs path of packge config file
-    default_file_name = str(Path(__file__).parent / config_file_name)
+    default_file_name = str(files("qcodes.configuration") / config_file_name)
```
astafan8, line 50: > Note that there is a trade off here as we now have hardcoded the path to config and monitor dist files in the python file rather than the relative path to that file. I think its cleaner to hardcode the full path but I could see the argument the other way around too

would it be possible to use relative imports to address this? e.g. would this work?

```suggestion
    default_file_name = str(files(".") / config_file_name)
```

or perhaps instead of the "dot" there's a builtin variable for "this module where you are now"?

astafan8 (approved):

qcodes/configuration/config.py
jenshnielsen: It's not documented to work but we give it a try

jenshnielsen: According to stack overflow getting the current module is not something that has a nice api

jenshnielsen: Found out that you can use `__loader__.name` to get the name of the current module. 
This works but is perhaps not too clean. Also ran into https://github.com/python/mypy/issues/4182

astafan8: if it's so not-easy/not-clean to find the path to the current module, then i'm fine with files("qcodes.configuration") for the sake of consistency

PR #4802, opened 2022-11-14
Load coordinates from netcdf in the correct order.

jenshnielsen:
data_array.coords is not guaranteed to return
in the same order as its written. However, this is guaranteed for data_array.dims so use that to order coords when loading

TODO

- [x] Changelog
- [x] See if we can find a way to test this

### DIFF
--- a/qcodes/tests/dataset/test_dataset_export.py
+++ b/qcodes/tests/dataset/test_dataset_export.py
@@ -5,7 +5,7 @@ import pytest
 import xarray as xr
 
 import qcodes
-from qcodes.dataset import new_data_set
+from qcodes.dataset import load_from_netcdf, new_data_set
 from qcodes.dataset.descriptions.dependencies import InterDependencies_
 from qcodes.dataset.descriptions.param_spec import ParamSpecBase
 from qcodes.dataset.descriptions.versioning import serialization as serial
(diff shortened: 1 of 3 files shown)


guenp (approved):
Tested and this fixes my problem. Thank you @jenshnielsen!

qcodes/tests/dataset/test_dataset_export.py
```diff
@@ -464,6 +492,31 @@ def test_to_xarray_da_dict_paramspec_metadata_is_preserved(mock_dataset_label_un
...
+    expected_path = f"qcodes_{mock_dataset_inverted_coords.captured_run_id}_{mock_dataset_inverted_coords.guid}.nc"
+    assert os.listdir(path) == [expected_path]
+    file_path = os.path.join(path, expected_path)
+    ds = qcodes.dataset.load_from_netcdf(file_path)
+
+    with pytest.warns(UserWarning):
+        xr_ds_reimported = ds.to_xarray_dataset()
+    assert xr_ds_reimported["z1"].dims == ("x", "y")
+    assert xr_ds_reimported["z2"].dims == ("y", "x")
+    assert xr_ds == xr_ds_reimported
```
astafan8, line 517: perhaps use identical for strictness https://docs.xarray.dev/en/stable/user-guide/combining.html?highlight=equality#equals-and-identical ? or even https://docs.xarray.dev/en/stable/generated/[xarray.testing.assert_identical](https://docs.xarray.dev/en/stable/generated/xarray.testing.assert_identical.html#xarray-testing-assert-identical).html#xarray-testing-assert-identical
```suggestion
    assert xr_ds.identical(xr_ds_imported)
```

astafan8 (approved):
great! (the only comment can be ignored)

qcodes/tests/dataset/test_dataset_export.py
jenshnielsen: done, required reordering so that conversion is done after export (so both have the the same export info attr)

jenshnielsen:
bors merge

PR #4810, opened 2022-11-18
Use StrEnum from std lib when avilable

jenshnielsen:
Python 3.11 added StrEnum to the std library so lets use that when available rather than rolling out our own


astafan8 (approved):
totally!

PR #4829, opened 2022-11-29
Pin all < 1.0.0 requirements exactly

jenshnielsen:
Since these effectively use the patch version as a minor version


requirements.txt
```diff
@@ -1,130 +1,136 @@
...
-broadbean~=0.11.0
+broadbean==0.11.0
 cachetools~=5.2.0
 certifi~=2022.9.24
 cffi~=1.15.1
 charset-normalizer~=2.1.1
 click~=8.1.3
 click-default-group~=1.2.2
-colorama~=0.4.5
+colorama-=0.4.5
```
jenshnielsen, line 26:
```suggestion
colorama==0.4.5
```

astafan8: typo? or is minus also allowed?
```suggestion
colorama==0.4.5
```

astafan8 (approved):

requirements.txt
jenshnielsen: This was a typo (fixed elsewhere)

PR #4840, opened 2022-11-29
Add the option to set pulse modulation in the Keysight `N51x1` signal generators

bguttel:
Notice that there are two parameters there, one that sets the general state of the pulse modulation (boolean) and one that sets whether it will be 'internal' or 'external'.

Not all the signal generators come with this feature so I added an `if` condition to test for it.


jenshnielsen:
@bguttel Thanks, looks good. Could you create a file at /docs/changes/news_fragments/4840.improved_driver with a short description of the fix so it makes it into the changelog

bguttel:
> @bguttel Thanks, looks good. Could you create a file at /docs/changes/news_fragments/4840.improved_driver with a short description of the fix so it makes it into the changelog

Added the file.
@jenshnielsen let me know if there is anything more that you think I should add! :)

jenshnielsen (approved):

PR #4842, opened 2022-11-30
remove non-working blueracer, replace with manual print of durations

jenshnielsen:
It seems like blueracer changed their integration but the documentation is unclear. This has never really worked correctly for us. We can come back to this when/if blueracer is more mature.

To get a little bit more feedback I have added ``--durations=20`` to print the run time of the slowest 20 tests to std out when running in CI


astafan8 (approved):

jenshnielsen:
bors merge

PR #4857, opened 2022-12-06
Add dataset export entry point

jenshnielsen:
This is useful for kicking of postprocessing, backup, upload etc.

- [x] Document
- [x] Test using a entry point that logs the export event

### DIFF
--- a/docs/examples/DataSet/Exporting-data-to-other-file-formats.ipynb
+++ b/docs/examples/DataSet/Exporting-data-to-other-file-formats.ipynb
@@ -310,6 +310,20 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "## Automatically post process exported datasets."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "QCoDeD will attempt to call any [EntryPoint](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#entry-points-for-plugins) registered for the group \"qcodes.dataset.on_export\". This will allow a user to setup a function that can trigger post processing such as backup to cloud, external drive, plotting or post process analysis. Functions registered for this entry point group are expected to take a Path to the file as input and return None. Please consult the [Setuptools docs](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#entry-points-for-plugins) for more information on the use of EntryPoints."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "## Importing exported datasets into a new database"
    ]
   },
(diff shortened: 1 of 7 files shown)


qcodes/dataset/data_set_protocol.py
```diff
@@ -1,5 +1,16 @@
 from __future__ import annotations
 
+import sys
+
+if sys.version_info >= (3, 10):
+    # distribution.name used below became part of the
```
jenshnielsen, line 6:
```suggestion
    # entrypoint selectable became part of the
```

qcodes/dataset/data_set_protocol.py
```diff
@@ -414,16 +426,30 @@ def _export_data(
...
+        # twice here
+        export_callbacks = set(entry_points(group="qcodes.dataset.on_export"))
+
+        if len(export_callbacks) == 0:
+            LOG.info("No export callbacks defined")
+
+        for export_callback in export_callbacks:
+            export_callback_function = export_callback.load()
+            LOG.info("Executing on_export callback {export_callback.name}")
+            export_callback_function(export_path)
```
jenshnielsen, line 450: This should probably take a Path object and not a str. Also update to reflect that this may be None

Dominik-Vogel: Do you think it could make sense to wrap this call with a try-except, so that the hook cannot crash the entire app, or does the hook mechanism already do this?

qcodes/extensions/_print_export_info.py
```diff
@@ -0,0 +1,2 @@
+def print_dataset_export_info(path: str) -> None:
```
Dominik-Vogel, line 1: For testing this at a pilot station, we could consider to replace this print with a log message.

qcodes/dataset/data_set_protocol.py
jenshnielsen: Yes thats probably a good idea

jenshnielsen: Its probably also a good idea to wrap the logic into a function that returns a collection of functions to execute. That way we can mock that in a test so that we can test that functions are exectuted regardless of the entrypoint logic.

Dominik-Vogel (approved):
This is beautiful!
I have tested it and it works ✔️

docs/examples/DataSet/Exporting-data-to-other-file-formats.ipynb
```diff
@@ -306,6 +306,20 @@
     "For more information about how to configure QCoDeS datasets see [the page about configuration](https://qcodes.github.io/Qcodes/user/configuration.html)  in the QCoDeS docs."
```
jenshnielsen, line 306: Fix this to point to the api docs page for configuration

PR #4870, opened 2022-12-15
Fix: DelegateParameter.validate also validates the source parameter

Junliang-Wang:
Prior to this change, DelegateParameter.validate only uses its own validator, ignoring the source parameter.

This change adds an extra layer of validation by using the source parameter's validator in DelegateParameter.validate.

These changes are related to the discussion [#4819](https://github.com/QCoDeS/Qcodes/discussions/4819)


qcodes/tests/parameter/test_delegate_parameter.py
```diff
@@ -512,3 +513,27 @@ def test_underlying_instrument_property_for_delegate_parameter():
...
+    with pytest.raises(ValueError):
+        delegate_param.validate(6)
+
+    delegate_param.vals = vals.Numbers(-10, 10)
+    source_param.vals = vals.Numbers(-5, 5)
+    delegate_param.validate(1)
+    with pytest.raises(ValueError):
+        delegate_param.validate(6)
+    with pytest.raises(ValueError):
+        delegate_param.validate(11)
```
jenshnielsen, line 539: I am not sure how this would interact with having a scale and or offset on the delegate parameter. It would probably be nice to add tests to ensure that works correctly

Junliang-Wang: I have just checked that .validate only checks on the value and ignores the raw_value (where offset and scale are considered).
I will add a couple more tests in the next commit.

On the other hand, while checking the effect of offset and scale, I found out that DelegateParameter does not inherit the offset and scale from the source parameter (see the snippet below). Is this intentional?

```python
param = Parameter('param', set_cmd=None, get_cmd=None, offset=10)
delegated = DelegateParameter('delegated', source=param) # no offset defined

print(f'{param.offset=}') # returns 10
print(f'{delegated.offset=}') # returns None

param(0) 
print(f'{param.get()=}') # returns 0
print(f'{param.get_raw()=}') # returns 10 as expected

delegated(0)
print(f'{delegated.get()=}') # returns 0
print(f'{delegated.get_raw()=}') # returns 0 --> it ignores source offset.
```

jenshnielsen: Yes, I belive the idea is that you can have a delegate parameter in a different unit. 
So you could do something like having a temperate in Kelvin with a delegate parameter in Celsius or Fahrenheit

Junliang-Wang: ah ok, that makes sense, thanks!

jenshnielsen:
@Junliang-Wang Thanks, Could you add a small description of the change in docs/changes/newsfragments/4870.improved

Junliang-Wang:
> @Junliang-Wang Thanks, Could you add a small description of the change in docs/changes/newsfragments/4870.improved

Sure, although I have a technical question: my local branch is now behind the master, I suppose that I should pull for the updates, and then commit the file in docs/changes/newsfragments/4870.improved, right? If I do so, I see that there is no 4870.improved, but rather 4876.improved. Should I create 4870.improved and write my changes?

astafan8:
> Sure, although I have a technical question: my local branch is now behind the master, I suppose that I should pull for the updates, and then commit the file in docs/changes/newsfragments/4870.improved, right? If I do so, I see that there is no 4870.improved, but rather 4876.improved. Should I create 4870.improved and write my changes?

@Junliang-Wang "Should I create 4870.improved and write my changes" - yes :) these change-files are written PER pull-request, so don't add information to the existing ones :) So, the 4876.improved file that you see was written for PR number 4876, and the one you're about to write is for 4870 because the PR you have open has number 4870.

qcodes/parameters/delegate_parameter.py
```diff
@@ -256,3 +256,21 @@ def snapshot_base(
...
+        Args:
+            value: value to validate
+
+        Raises:
+            TypeError: If the value is of the wrong type.
+            ValueError: If the value is outside the bounds specified by the
+               validator.
+        """
+        if self.vals is not None:
+            super().validate(value)
```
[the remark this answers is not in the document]
Junliang-Wang: You are right, it is not necessary. I confirm that removing it all tests pass.

Junliang-Wang:
@astafan8 Thanks for the answer! I have pushed your suggestion, along with 4870.improved file.

qcodes/tests/parameter/test_delegate_parameter.py
```diff
@@ -512,3 +513,56 @@ def test_underlying_instrument_property_for_delegate_parameter():
...
+    source_param.validate(1)  # raw_value = 100
+    delegate_param.validate(1)  # raw_value = 1
+
+    source_param.offset = None
+    source_param.scale = None
+
+    delegate_param.offset = 100
+    delegate_param.scale = None
+    source_param.validate(0)  # raw_value = 0
+    delegate_param.validate(0)  # raw_value = 100
```
astafan8, line 563: sorry, looking at this a bit more i have now a question: shouldn't this validation actually fail? so the delegate parameter has offset of 100, so this means that when the delegate gets set to 0, then the source parameter gets set to 100 (because the delegate parameter passes its raw value to the source parameter, not its value), and this would trigger an error on set, but it does not trigger an error on validate. to confirm my logic or not, could you add `delegate_param.set(0)` to this test and see if it still passes (my guess is that it will fail)?

i think what needs to be done in the implementaion is that validate's call to source.validate should mirror what happens in set_raw where source.set is called or in the delegate cache of DelegateParameter - and that is the call to ``_from_value_to_raw_value(value)``, so the implementation of the validate should actully be:
```python
        super().validate(value)
        if self.source is not None:
            self.source.validate(self._from_value_to_raw_value(value))
```
so if my guess is true, that the ``delegate_param.set(0)`` now fails, then after this change is added to the implementation of validate the ``delegate_param.validate(0)`` would also raise an exception.

Junliang-Wang: Yes, I realized this, but I wasn't sure if vals shoud check the raw value or the input value.
From the user point of view, imagine I have a delegate_param to control a DC voltage with 1/1000 voltage divider, i.e. scale=1/1000. If I want to apply 1mV, I can simply use delegate_param(1). Now, to set a safety limit of +/-5mV, should vals be (-5,+5),? or (-5/1000, +5/1000) even though we are using scale=1/1000?

If one looks inside ```DelegateParameter.set```:
```python
def set(self, value: ParamDataType) -> None:
    if self._parameter.source is None:
        raise TypeError(
            "Cannot set the cache of a DelegateParameter "
            "that delegates to None"
        )
    self._parameter.validate(value)
    self._parameter.source.cache.set(
        self._parameter._from_value_to_raw_value(value)
    )
```
it validates the input value, but set the raw_value.
I agree that .validate should somehow mimic .set, but it would change the assignment of vals.

P.S: adding ```delegate_param.set(0)```, the test fails as expected.

Junliang-Wang: @astafan8 after thinking about it, I think I understood the problem.
What DelegateParameter.validate should check is the input value using its own validator, and check the raw value with the source's validator.
I am going to implement this in the next commit.

astafan8: > What DelegateParameter.validate should check is the input value using its own validator, and check the raw value with the source's validator.

@Junliang-Wang yes, thank you! now things look correct :)

jenshnielsen (approved):

PR #4876, opened 2022-12-20
make it a no op to mark a run completed twice

jenshnielsen:
Otherwise one may very easily trigger an unintentional change of the time stamp

### DIFF
--- a/qcodes/dataset/sqlite/queries.py
+++ b/qcodes/dataset/sqlite/queries.py
@@ -976,7 +976,10 @@ def new_experiment(
 # TODO(WilliamHPNielsen): we should remove the redundant
 # is_completed
 def mark_run_complete(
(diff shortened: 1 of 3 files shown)


qcodes/dataset/sqlite/queries.py
```diff
@@ -985,7 +988,16 @@ def mark_run_complete(
...
         timestamp: time stamp for completion. If None the function will
             automatically get the current time.
+        override: Mark already completed run complted again and override completed
+            timestamp with new value. If False marking an already completed run completed
+            is a noop.
+
     """
+    if override is False:
+        if completed(conn=conn, run_id=run_id):
+            log.warning("Trying to mark a run completed that was already completed.")
```
jenshnielsen, line 998: Not sure if this really should be a warning. Could just be at info

WilliamHPNielsen (approved):

PR #4894, opened 2023-01-02
[WIP] Refactor types to check with pyright

jenshnielsen:
Over Christmas I spent a bit of time experimenting with getting types to check with pyright in addition to mypy
This is mainly motivated by having typechecking inline in vscode 
There are a number of differences.

* Pyright is somewhat less flexible if a parameter defined within a branch and later used within another branch with the same condition.
* Pyright does not allow inline operations such as - (minus) on numpy arrays unless the dtype is known (could be string etc)
* pyright is less flexible with type narrowing when unknown types are used
* pyright checks code even if the function is not typed at all, so exlude a bunch of stuff that is not typed
* pyright has a way of generation stubs automatically so I have generated stubs to check agains for some of our packages


qcodes/dataset/data_set.py
```diff
@@ -1353,7 +1354,7 @@ def _finalize_res_dict_numeric_text_or_complex(
                     flat_results[dep.name] = result_dict[dep].ravel()
             for inff in inff_params:
                 if numpy.shape(result_dict[inff]) == ():
-                    flat_results[inff.name] = numpy.repeat(result_dict[dep], N)
```
jenshnielsen, line 1356: This looks like a bug in the original code but also looks like we have no tests for this

qcodes/instrument/instrument_base.py
```diff
@@ -390,7 +390,7 @@ def parent(self) -> InstrumentBase | None:
         return None
 
     @property
-    def ancestors(self) -> list[InstrumentBase]:
```
jenshnielsen, line 393: This should be documented as a breaking change

qcodes/station.py
```diff
@@ -748,10 +754,12 @@ def instrument_names_from_module(module: ModuleType) -> Tuple[str, ...]:
         submodules = list(pkgutil.walk_packages(module.__path__, module.__name__ + "."))
         res = set()
         for s in submodules:
-            with suppress(Exception):
```
astafan8, line 751: what does pyright not like here? i'd argue that in this case with-suppress is more readable

qcodes/validators/validators.py
```diff
@@ -455,7 +455,9 @@ def validate(
         Raises:
             TypeError: If not a complex number.
         """
-        if not isinstance(value, self.validtypes):
+        # for some reason pyright does not thing numpy complex
```
astafan8, line 458:
```suggestion
        # for some reason pyright does not think numpy complex
```

astafan8 (commented):
looks nice! some of the found things are indeed well-found bugs!

qcodes/station.py
jenshnielsen: If the exception is raised ms might be undefined in the original code and the for loop below would raise. Could probably also fix this by letting ms be defined to [] before this block

jenshnielsen:
Ipython became typed in https://github.com/QCoDeS/Qcodes/pull/4897 so those stubs should be removed

qcodes/dataset/dond/do_nd_utils.py
```diff
@@ -38,13 +38,13 @@
...
-    List[Optional["matplotlib.colorbar.Colorbar"]],
+    Tuple[Optional["matplotlib.axes.Axes"], ...],
+    Tuple[Optional["matplotlib.colorbar.Colorbar"], ...],
 ]
 MultiAxesTupleListWithDataSet = Tuple[
     Tuple[DataSetProtocol, ...],
-    Tuple[List["matplotlib.axes.Axes"], ...],
-    Tuple[List[Optional["matplotlib.colorbar.Colorbar"]], ...],
+    Tuple[Tuple["matplotlib.axes.Axes", ...], ...],
+    Tuple[Tuple[Optional["matplotlib.colorbar.Colorbar"], ...], ...],
```
jenshnielsen, line 47: Document this as a breaking change

docs/changes/newsfragments/4894.breaking
```diff
@@ -0,0 +1,5 @@
+``InstrumentBase.ancestors`` now returns a tuple of ancestors
+to the current Instrument, Instrumentchannel or Instrumentmodule and not a list.
```
astafan8, line 2:
```suggestion
to the current Instrument, InstrumentChannel or InstrumentModule and not a list.
```

astafan8 (approved):

PR #4899, opened 2023-01-04
Add export location config

jenshnielsen:
Makes it possible to define the export path relative to 

@Dominik-Vogel 

- [x] Tests both with and without this option in use
- [x] Verify that this option is only used via the accessor which does the correct expansion
- [x] Document
- [x] This currently requires this dir to already exists. Its probably more user friendly to create it if not already there
- [x] Add a public api for getting the expanded path


qcodes/dataset/data_set_protocol.py
```diff
@@ -427,6 +430,7 @@ def _export_data(
         # Set defaults to values in config if the value was not set
         # (defaults to None)
         path = path if path is not None else get_data_export_path()
+        os.makedirs(path, exist_ok=True)
```
Dominik-Vogel, line 433: as you switched to pathlib here you could consider using Path.mkdirs, and we could consider also creating the parents.

qcodes/dataset/data_set_protocol.py
```diff
@@ -377,6 +377,9 @@ def export(
             ValueError: If the export data type is not specified or unknown,
                 raise an error
         """
+        if isinstance(path, str):
```
Dominik-Vogel, line 380: Personally, I find path = Path(path) more readable, but I am sure you have a good motivation for the explicit isinstance.

Dominik-Vogel (approved):

qcodes/dataset/data_set_protocol.py
jenshnielsen: The reason is that it could be noneIt could be None

PR #4906, opened 2023-01-08
Rework config fixtures

jenshnielsen:
Replace existing config fixtures with 2 auto use fixtures.

* A session fixture that makes sure that we load the default config and do not read / write config from any user folder. This makes sure that we dont fail tests due to config changes of the user or modify user configs. Also let this handle setting changes that we want to be different from the default. Disable telemetry, disable subscribers, set db path to a non existing file so test can't write to the default db if not using dataset / experiment fixtures.
* A function fixture that resets any modification that a test has made to the config, closes any instruments and resets the default station

Since these are auto use we can remove all explicit use of config fixtures.

In the process fix some issues.

* The old default_config fixture would replace qc.config with a new config object. This however will not work if the test had already imported the config object as `from qcodes import config` so instead patch the existing config object.
* some dond tests were not correctly using the experiment fixture. This is also fixed.
* remove deprecated function
* config tests no longer needs to be skipped when running from user home as we patch the location of user home automatically before the test runs
* Fix some logging tests that were leaking state or depending on leaked state

### DIFF
--- a/qcodes/tests/conftest.py
+++ b/qcodes/tests/conftest.py
@@ -185,11 +185,3 @@ def _make_standalone_parameters_dataset(
     )
     dataset.mark_completed()
     yield dataset
-
-
-@pytest.fixture(name="set_default_station_to_none")
-def _make_set_default_station_to_none():
-    """Makes sure that after startup and teardown there is no default station"""
-    Station.default = None
-    yield
-    Station.default = None
--- a/qcodes/tests/conftest.py
+++ b/qcodes/tests/conftest.py
@@ -18,6 +18,7 @@ from qcodes.dataset.data_set import DataSet
 from qcodes.dataset.descriptions.dependencies import InterDependencies_
 from qcodes.dataset.descriptions.param_spec import ParamSpecBase
 from qcodes.dataset.experiment_container import Experiment, new_experiment
+from qcodes.instrument import Instrument
 from qcodes.station import Station
 
 settings.register_profile("ci", deadline=1000)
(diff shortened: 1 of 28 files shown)


qcodes/tests/conftest.py
```diff
@@ -51,7 +51,7 @@ def disable_telemetry():
         qc.config.telemetry.enabled = original_state
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="function", autouse=True)
```
Dominik-Vogel, line 54: @jenshnielsen maybe the default_config fixture adds a lot of overhead if applied to every test. The reason it is so complicated is that it takes into account that the config might be reloaded somewhere. Maybe this is the time to go through the use of the config and make sure it is not dynamically reloaded, or make the reloading easier. In that case we could reduce this to:

```python
   current_config, current_schema = qc_config.current_config, qc_config.current_schema
    try:
        qc_config.current_config, qc_config.current_schema = qc_config.load_default()
        yield
    finally:
        qc_config.current_config, qc_config.current_schema = current_config, current_schema

```

jenshnielsen: @Dominik-Vogel reworked this please have a look at the new version

qcodes/tests/dataset/dond/test_do0d.py
```diff
@@ -20,7 +20,7 @@
 @pytest.mark.usefixtures("plot_close", "experiment")
 @pytest.mark.parametrize("period", [None, 1])
 @pytest.mark.parametrize("plot", [None, True, False])
-@pytest.mark.parametrize("plot_config", [None, True, False])
+@pytest.mark.parametrize("plot_config", [True, None, False])
```
jenshnielsen, line 23: This test was leaking config but if done like this it will leak into the next variant here which causes a test failure

astafan8 (approved):

Dominik-Vogel (approved):

PR #4912, opened 2023-01-09
Fix use of experiment fixture

jenshnielsen:
One test was missing the fixture causing it to write to the default db
The other tests were incorrectly using mark on fixtures which is not supported. This was partially fixed
by applying the experiment fixture again to the test but that would result in the wrong order

Discovered as part of https://github.com/QCoDeS/Qcodes/pull/4899 and https://github.com/QCoDeS/Qcodes/pull/4906


Dominik-Vogel (approved):
Thank you for teaching us abut this 'bug' with usefixture!

jenshnielsen:
For reference https://github.com/pytest-dev/pytest/issues/3664

PR #4919, opened 2023-01-11
Driver Keithley add union type

jenshnielsen:
I wanted to update the types in the calibration module to use the official api but noticed that there were no good way of saying give me any of the supported 26xx drivers. This fixes that by adding a Union type


astafan8 (approved):

qcodes/instrument_drivers/Keithley/__init__.py
```diff
@@ -35,6 +37,22 @@
     KeithleyS46RelayLock,
 )
 
+Keithley26xxB = Union[
```
jenshnielsen, line 40: The type should not contain B as we also have at least one A driver

PR #4923, opened 2023-01-12
Move VoltageDivider into parameters and deprecate

jenshnielsen:
This is not an instrument but a Parameter so it makes more sense for it to live there. 
Since DelegateParamerer does the same more general and better deprecate this one


jenshnielsen:
bors merge

jenshnielsen:
bors retry

PR #4941, opened 2023-01-23
Update Minicircuits and zhist drivers to conform with new public api, clear docs build scripts

jenshnielsen:
This completes the refactoring of instruments to be importable from the vendor module


qcodes/instrument_drivers/Minicircuits/__init__.py
```diff
@@ -0,0 +1,8 @@
+from .RUDAT_13G_90 import MiniCircuitsRudat13G90Usb
+from .USB_SPDT import MiniCircuitsSwitchChannelUsb, MiniCircuitsUsbSPDT
```
jenshnielsen, line 2: Perhaps these names could be updated to make it more clear that one is a channel of the other

`MiniCircuitsSPDTUsb` and `MiniCircuitsSPDTChannelUsb` may be better names?

astafan8 (approved):

PR #4956, opened 2023-01-27
Attach default experiment to a measurement

einsmein:
This PR finds a default experiment and attach it to a measurement when it is created.
This behavior matches the document of [`Measurement` class ](https://github.com/QCoDeS/Qcodes/blob/master/qcodes/dataset/measurements.py#L687), which states 
```
...

    Args:
        exp: Specify the experiment to use. If not given
            the default one is used. The default experiment
            is the latest one created.
...
```

### DIFF
--- a/qcodes/dataset/measurements.py
+++ b/qcodes/dataset/measurements.py
@@ -48,10 +48,11 @@ from qcodes.dataset.descriptions.dependencies import (
     InterDependencies_,
 )
 from qcodes.dataset.descriptions.param_spec import ParamSpec, ParamSpecBase
-from qcodes.dataset.descriptions.rundescriber import RunDescriber
 from qcodes.dataset.descriptions.versioning.rundescribertypes import Shapes
-from qcodes.dataset.experiment_container import Experiment
+from qcodes.dataset.experiment_container import Experiment, load_experiment
+from qcodes.dataset.experiment_settings import get_default_experiment_id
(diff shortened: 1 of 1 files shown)


jenshnielsen:
@einsmein QCoDeS is using darker https://pypi.org/project/darker/ to only format changes in order to not obscure the history so please don't format the entire file when making small changes

jenshnielsen:
```
    Args:
        exp: Specify the experiment to use. If not given
            the default one is used. The default experiment
            is the latest one created.
```

This logic is already implemented but it lives in the dataset class. See line 272 in https://github.com/QCoDeS/Qcodes/blob/master/qcodes/dataset/data_set.py

einsmein:
@jenshnielsen I understand a default experiment is attached to a dataset. However, `measurement.experiment` is empty even though the underlying dataset is attached to a default experiment, which can be misleading at runtime imo.

jenshnielsen:
> @jenshnielsen I understand a default experiment is attached to a dataset. However, measurement.experiment is empty even though the underlying dataset is attached to a default experiment, which can be misleading at runtime imo.

I agree, experiment really should have been a private property on the measurement class. There is no reason why the users should need to interact with it there. But since its not I am happy with this change provided:

* It gets a test
* The other tests are fixed to pass
* The unrelated formatting is removed

PR #4968, opened 2023-02-02
Github actions cancel old jobs

jenshnielsen:
See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

This also means that we can get rid of the cancel old jobs work flow


astafan8 (approved):
Awesome!

PR #4975, opened 2023-02-06
Add types to part of test module 

jenshnielsen:
Primarily validator and parameter tests. 

TODO 

- [x] Replace all with Parameter(Raw)DataTypes when possible


qcodes/validators/validators.py
```diff
@@ -2,31 +2,37 @@
...
 """
-import collections.abc
+from __future__ import annotations
+
 import math
+import typing
+from collections import abc
 
 # rename on import since this file implements its own classes
 # with these names.
```
astafan8, line 12:
```suggestion
```

qcodes/validators/validators.py
```diff
@@ -1155,12 +1161,12 @@ def __repr__(self) -> str:
         return "<Callable>"
 
 
-class Dict(Validator[TDict[Hashable, Any]]):
+class Dict(Validator[typing.Dict[Hashable, Any]]):
```
astafan8, line 1164: not dict?
```suggestion
class Dict(Validator[dict[Hashable, Any]]):
```

astafan8 (approved):

qcodes/validators/validators.py
jenshnielsen: Pyupgrade says no. I don't think dict is indexable in this context until we drop 3.8 support. Since this needs to work at runtime

PR #4983, opened 2023-02-08
Unify license into one file

jenshnielsen:
For some reason we have a rst version of the license file. However, this file is not even used in the docs so there seems to be no point to have that so remove and just use the regular license file.

Update copyright year as well


astafan8 (approved):

jenshnielsen:
Bors merge

einsmein (approved):

PR #4986, opened 2023-02-08
Try building docs with warnings as errors on 3.11, 3.12

jenshnielsen:
Looks like the original issue is fixed

Thou I am seeing a warning from building Lakeshore documentation. It looks like this is coming from the std lib documentation of 
``LakeshoreModel325Status`` Probably from the std lib docs of IntFlag

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -39,11 +39,8 @@ jobs:
             python-version: 3.11
     env:
       OS: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


astafan8 (approved):

jenshnielsen:
Submitted https://github.com/python/cpython/pull/117847 for fix the formatting upstream.

jenshnielsen:
We have worked around the issue in a different way

PR #4990, opened 2023-02-09
Towards checking code with ruff

jenshnielsen:
This pr contains a bunch of cleanups found by ruff using the default settings (apart from the inline and pyproject.toml ) exclusions. This is not yet complete but I think its worth merging as is to reduce the chance of merge conflicts


astafan8 (approved):
this IS lovely!

PR #5000, opened 2023-02-10
Configure basic ruff linter as precommit hook

jenshnielsen:
Only checks for issues like https://github.com/QCoDeS/Qcodes/pull/4999 for now. That pr should be merged first and then rebased away from this pr


jenshnielsen:
Folded into #5006

PR #5006, opened 2023-02-13
Enable basic checks with ruff

jenshnielsen:
Enables default E and F categories + the changes from #5000


qcodes/tests/test_interactive_widget.py
```diff
@@ -5,9 +5,10 @@
 import pytest
 from ipywidgets import HTML, Button, GridspecLayout, Tab, Textarea
 
+from qcodes import interactive_widget
+
 # set matplotlib backend before importing pyplot
 matplotlib.use("Agg")
```
astafan8, line 11: dont konow why this is needed but the import seems to need to happen after the matplotlib line :) so "# noqa E402"?

astafan8 (approved):

qcodes/tests/test_interactive_widget.py
jenshnielsen: Did you test this manually. I would not expect that to be the case for `interactive_widget` since that now lacily imports matplotlib.pyplot within _plot_ds

astafan8: no i didnt test this locally. i only suggested from the point of view of "keeping the existing code/behavior".

but what you say makes sense, and the ci is green(?), so then let's keep this change

jenshnielsen: ok good. I don't think the test has been updated since the import of matplotlib was made lazy (for speed)

PR #5011, opened 2023-02-14
Add types for math utils module

jenshnielsen:


astafan8 (approved):
AWESOME!!!! So glad to this finally getting done (remember making an attempt long time ago in #1825)

jenshnielsen:
Yes, I think I went for a simple solution than #1825 Could probably do something better but at least this gets the code to typecheck

PR #5012, opened 2023-02-14
Add zhinst-qcodes as a test dependency

jenshnielsen:
Such that we can type check against it and remove it from excluded modules (both for pyright and mypy)

Also remove exclusions no various other modules that no longer require it.
Ipython has types
The others are no longer (direct) dependencies of qcodes


jenshnielsen:
This fails in CI because the version of qcodes is unknow when doing a shallow clone and zhinst-qcodes depends on qcodes.
We could

* Set a default version >= 0.30 This should be safe since it would only apply to new clones which are anyway that.
* Update CI to always do a non shallow clone
* Don't do this

astafan8 (approved):

jenshnielsen:
@astafan8 any opinon on the 2 suggestions of how to fix this in CI?

astafan8:
> Set a default version >= 0.30 This should be safe since it would only apply to new clones which are anyway that.

how would that be done?

> Update CI to always do a non shallow clone

i've refreshed my undersatnding of this by reading https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/# and this sounds option seems very reasonable. I realized that because qcodes uses versioningit for indentifiying its version, it actually puts additional requirements on how we should clone the repo for use in CI, that is, it should have the tags and it should know how far it is from latest tag. But this does not need to hold for every CI job though. So, with the shallow clone we get qcodes version is 0.0, and so far for the pytest jobs it was fine, we did not need qcodes version to be correct. However e.g. for docs, and package publishing, we do need a correct version, hence we pass `fetch-depth: '0'` to the checkout github action for the docs build, which, i guess, makes a "treeless" clone (from the article) downloading the commit info and tags, but no actual content of the commits (hopefully it is not much slower?). So, the question is if it is worth it to have correct qcodes version for tests - i'd say yes, if we have tests that depend on it, and if we have dependencies that need it for some reason. Seemingly our tests dont care about the version, so good. And our dependencies so far also didn't, except zhinst-qcodes as a test dependency.... But then i'm getting included towards the point below:

> Don't do this

i guess you mean "don't do the zhinst-qcodes part"? if so, then i agree, that could also be reasonable since we'd like to remove those aliases to zhinst instruments eventually.

jenshnielsen:
> how would that be done?

In pyproject.toml we set

```
[tool.versioningit]
default-version = "0.0"
```

Since we are now past 0.37.0 it would be safe to set that to `0.38.0.dev0+couldnotfindgithistory` or something like that

astafan8:
> it would be safe to set that to 0.38.0.dev0+couldnotfindgithistory or something like that

ah! ok, understand. but looks like a hack))) on the other hand, maybe that's a much more useful default that 0.0 (if only it would be updating itself after every release ;) )

jenshnielsen:
> ah! ok, understand. but looks like a hack))) on the other hand, maybe that's a much more useful default that 0.0 (if only it would be updating itself after every release ;) )

Yes one could have a bot that did that automatically on tag but seems a bit overkill


The time to do a shallow clone and a full clone in CI is 3 s and 8 s respectively so I would say that this is not a big issue

PR #5013, opened 2023-02-14
fix(Keithley3706A): Fix interlock status dictionary

brandonaltieri1:
When instantiating an object of the class `Keithley3706A`, the interlock status is checked for each card that is installed in the unit. When querying the unit for the interlock status (for a specific slot), the unit can return any of the following values:

image

See page 742 of https://download.tek.com/manual/3700AS-901-01D_Jul_2018_Ref_0.pdf. The interlock state is mapped using a dictionary (`interlock_status` in the `get_interlock_state` method), where the returned value is used as the key to find the appropriate status string. However, this dictionary was initially defined using only two keys, `0` and `1`. When querying the interlock status, the possible return values are `nil`, `0`, `1`, `2`, and `3`. As some of these return values are not included as keys in the `interlock_status` dictionary, when the unit does not return `0` or `1`, a `KeyError` is raised, and the driver cannot be instantiated. This PR updates this `interlock_status` dictionary to include all possible return values as keys, and adds the associated status strings (taken directly from the reference manual). The unit test/`sim` file are both updated to reflect these changes.

### DIFF
--- a/docs/changes/newsfragments/5013.improved_driver
+++ b/docs/changes/newsfragments/5013.improved_driver
@@ -1,6 +1,6 @@
 Fixed a bug in interlock status querying for Keithley 3706A. Originally, not all
-potential responses from the system were accounted for when querying for interlock 
-status. A dictionary is used to map the response from the system to a string describing 
+potential responses from the system were accounted for when querying for interlock
+status. A dictionary is used to map the response from the system to a string describing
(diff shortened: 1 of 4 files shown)


jenshnielsen:
Thanks @brandonaltieri1 could you add a short message to a file called `5013.improved_driver` in `docs\changes\newsfragments` explaining the fix

astafan8 (approved):

PR #5020, opened 2023-02-17
Remove bors

jenshnielsen:
Github merge queues is working better for us already than bors has ever had so it makes sense to remove this


astafan8 (approved):
~~bros merge~~ 🙋‍♀️🙋‍♂️

PR #5023, opened 2023-02-19
Reduce pyright issues in tests

jenshnielsen:
From 192 issues to 99


qcodes/dataset/sqlite/db_upgrades/__init__.py
```diff
@@ -36,9 +36,17 @@
 
 
 # INFRASTRUCTURE FOR UPGRADE FUNCTIONS
+# use a Protocol to typo optional arguments
```
astafan8, line 39:
```suggestion
# use a Protocol to type optional arguments
```

astafan8 (approved):

PR #5030, opened 2023-02-21
Run pyright on tests

jenshnielsen:
And fix remaining issues in the tests.

Only important change should be that 
the_same_dataset as has been promoted to be part of the dataset protocol

### DIFF
--- a/qcodes/tests/dataset/dond/test_do2d.py
+++ b/qcodes/tests/dataset/dond/test_do2d.py
@@ -154,9 +154,12 @@ def test_do2d_output_data(_param, _param_complex, _param_set, _param_set_2):
     )
     data = exp[0]
 
-    assert data.parameters == (
-        f"{_param_set.name},{_param_set_2.name}," f"{_param.name},{_param_complex.name}"
-    )
+    assert set(data.description.interdeps.names) == {
+        _param.name,
+        _param_complex.name,
+        _param_set.name,
+        _param_set_2.name,
+    }
     loaded_data = data.get_parameter_data()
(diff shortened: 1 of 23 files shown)


qcodes/tests/dataset/dond/test_do2d.py
```diff
@@ -154,9 +154,12 @@ def test_do2d_output_data(_param, _param_complex, _param_set, _param_set_2):
     )
     data = exp[0]
 
-    assert data.parameters == (
```
astafan8, line 157: as agreed offline, let's keep at least some coverage for (the discouraged) `.parameters` to ensure the order is preserved

qcodes/tests/dataset/dond/test_do2d.py
```diff
@@ -154,9 +154,12 @@ def test_do2d_output_data(_param, _param_complex, _param_set, _param_set_2):
...
-        f"{_param_set.name},{_param_set_2.name}," f"{_param.name},{_param_complex.name}"
-    )
+    assert set(data.description.interdeps.names) == {
```
astafan8, line 157: as agreed offline, let's keep testing for the order here so that we have more chanced to know when it changed if it changed

astafan8 (approved):

qcodes/tests/dataset/dond/test_do2d.py
jenshnielsen: added to test_dataset_basic

qcodes/tests/dataset/dond/test_do2d.py
jenshnielsen: done

PR #5031, opened 2023-02-21
Dataset: change default guid type to use a random str

jenshnielsen:
As earlier suggested this switches the default to replace sample by a random string eliminating duplicate guids. That was especially likely if generating lots of guids on the same system rapidly after each other on a system with low time resolution

### DIFF
--- a/qcodes/dataset/sqlite/queries.py
+++ b/qcodes/dataset/sqlite/queries.py
@@ -25,7 +25,7 @@ from qcodes.dataset.descriptions.rundescriber import RunDescriber
 from qcodes.dataset.descriptions.versioning import serialization as serial
 from qcodes.dataset.descriptions.versioning import v0
 from qcodes.dataset.descriptions.versioning.converters import new_to_old, old_to_new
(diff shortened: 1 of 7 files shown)


jenshnielsen:
test_update_existing_guids is failing since its reading back a sample_id and then trying to generate a new sample with that id. And since its now an error to do that it will raise

qcodes/dataset/sqlite/queries.py
```diff
@@ -25,7 +25,7 @@
 from qcodes.dataset.descriptions.versioning import serialization as serial
 from qcodes.dataset.descriptions.versioning import v0
 from qcodes.dataset.descriptions.versioning.converters import new_to_old, old_to_new
-from qcodes.dataset.guids import generate_guid, parse_guid
```
trevormorgan, line 28: Do we plan to deprecate the generate_guid function or is there still a need to keep it?

jenshnielsen: Not at the moment this is still used to generate guids when a new dataset is created

PR #5036, opened 2023-02-26
Deprecate `parameter.__getitem__`

jenshnielsen:
And therefor the ability to sweep a parameter using the syntax
param[0,1.0,10]

* This syntax is confusing since its not really indexing the parameter
* The implementation requires the input to be floats which is not
pytonic
* It does nothing that cannot be done with param.sweep

### DIFF
--- a/qcodes/parameters/parameter.py
+++ b/qcodes/parameters/parameter.py
@@ -7,6 +7,8 @@ import logging
 import os
 from typing import TYPE_CHECKING, Any, Callable, Literal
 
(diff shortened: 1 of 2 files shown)


astafan8 (approved):

jenshnielsen:
This is not a realistic approach since qcodes_loop makes to heavy use of __getitem__ in all its forms.

Suggest to add a class to qcodes loop that takes a parameter as input and exposes __getitem__ and sweep to use in that package

PR #5040, opened 2023-02-28
get_component: In InstrumentBase fall back to a longer name  when not matching a last part of a name

jenshnielsen:
This makes it possible to get `voltage_range` on the Yokogawa gs200 as shown in the relevant test.

It does perhaps make the error message a little less clear in the context that you try to get `foo_bar` where `foo` is a component but `bar` is not but I think we can live with that


jenshnielsen:
This is not a realistic approach since qcodes_loop makes to heavy use of __getitem__ in all its forms.

Suggest to add a class to qcodes loop that takes a parameter as input and exposes __getitem__ and sweep to use in that package

astafan8 (approved):

PR #5043, opened 2023-03-01
add driver for Lakeshore Model 335

jbeckerwsi:
- Add driver for Lakeshore Model 335
- Increase range on setpoint property of Lakeshore Model 335,336,372 to support degree Celsius


jbeckerwsi:
Pending test on hardware

qcodes/instrument_drivers/Lakeshore/Lakeshore_model_335.py
```diff
@@ -0,0 +1,63 @@
+from typing import Dict
+
+import pyvisa as visa
+
+from .lakeshore_base import LakeshoreBase
+from .Lakeshore_model_336 import (
+    LakeshoreModel336Channel,
+    LakeshoreModel336CurrentSource,
```
jenshnielsen, line 8: We have been slowly converting to exposing the drivers as a public api at a level above this. So could you add the driver to the `__init__` file in this folder (following the pattern of the rest) and make this file private (start the name with an underscore_

qcodes/instrument_drivers/Lakeshore/Lakeshore_model_335.py
```diff
@@ -0,0 +1,63 @@
+from typing import Dict
+
+import pyvisa as visa
```
jenshnielsen, line 3: To not depend on implementation details of imports in the pyvisa module you should also explicitly import pyvisa.constants

I would prefer to get rid of the pyvisa as visa alias and just import pyvisa as is

qcodes/instrument/sims/lakeshore_model335.yaml
```diff
@@ -0,0 +1,102 @@
...
+devices:
+  device 1:
+    eom:
+      GPIB INSTR:
+        q: "\r\n"
+        r: "\r\n"
+    error: ERROR
+    dialogues:
+      - q: "*IDN?"
+        r: "QCoDeS, m0d3l, 336, 0.0.01"
```
jenshnielsen, line 11: Does the 335 really return this on idn?
```suggestion
        r: "QCoDeS, m0d3l, 336, 0.0.01"
```

jenshnielsen (approved):

PR #5079, opened 2023-03-24
Keysight PNA driver from the N52xx base

jpdehollain:
Simple addition of a PNA N52xx driver that has slightly different frequency and port characteristics than the other ones available.


jenshnielsen:
Thanks @jpdehollain We are trying to move to a format where the drivers are exposed one level above so this can be imported directly from the Keysight folder. Would you add this driver there next to the other ones and perhaps rename this file to start with a `_` to indicate that it is not to be imported from directly

jpdehollain:
Thanks @jenshnielsen, I think I created some confusion with my initial PR description. This is the driver for the N5232B model, which uses the N52xx base that was already available. The file I created (`Keysight_N5232B.py`) is in the Keysight folder alongside the other similar files that use the N52xx base.
Is that correct or did you mean something different?
One thing I could do is add the _ prefix to the existing `N52xx.py` file, and modify the imports of all the existing driver files (as well as this new one). Would you like me to do that?

jenshnielsen:
Sorry I was probably being unclear. I meant add it here https://github.com/QCoDeS/Qcodes/blob/master/qcodes/instrument_drivers/Keysight/__init__.py

This will make sure this version is documented as part of the page at https://qcodes.github.io/Qcodes/drivers_api/Keysight.html
and make it such that the driver is importable like this `from qcodes.instrument_drivers.Keysight import KeysightN5232B`

The current files in this folder and not marked private and I am reluctant to change that due to backwards compatibility issues but in hindsight I wish they were all private so it was clear to import from  qcodes.instrument_drivers.Keysight only.

jpdehollain:
Ah got it, makes sense now... I think that should be done now!

PR #5090, opened 2023-03-28
Fix auto_sweep in Keysight N52xx Driver

### LINKED ISSUE #5089: Disabling auto_sweep in Keysight N52xx driver broken
The `auto_sweep` function in the Keysight N52xx driver is broken. This feature allows data to be taken directly off screen without running a scan when disabled, which is necessary when measuring multiple parameters at once.

It looks like the check for auto_sweep was accidentally disabled in commit 78a989f

## Expected Behaviour
When running the below snippet the PNA should only take a single measurement
```python3
pna.auto_sweep(False)
qc.dataset.do0d(pna.magnitude, pna.phase)
```

## Actual
The pna will run the trace twice, once for magnitude and once for phase.

@jenshnielsen

spauka:
This fixes #5089 and reenables auto_sweep.

I also fix the units for polar which should be a ratio rather than a voltage.

### DIFF
--- a/docs/changes/newsfragments/5090.improved_driver
+++ b/docs/changes/newsfragments/5090.improved_driver
@@ -1,0 +1,2 @@
+Fix auto_sweep in Keysight PNA N52xx driver
+Fix units of polar sweeps in Keysight PNA N52xx driver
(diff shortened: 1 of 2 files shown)


jenshnielsen:
Thanks @spauka would you mind adding a changelog note for this?

spauka:
Done :)

jenshnielsen (approved):

PR #5144, opened 2023-05-03
Drop python 3 8 support

jenshnielsen:
IPython and pandas has dropped support for python 3.8 so this seems to be a good time to start considering doing the same.

* Change the metadata for the minimum supported version
* Upgrade precommit hook
* Rerun precommit hook
* Drop 3.8 from CI
* Remove python 3.8 specific requirements


astafan8 (approved):
i reviewer the non-py files, so looks good))

PR #5200, opened 2023-06-09
Make in memory cache configurable

jenshnielsen:
so it may be disabled for huge datasets

### DIFF
--- a/qcodes/dataset/dond/do_nd.py
+++ b/qcodes/dataset/dond/do_nd.py
@@ -571,6 +571,7 @@ def dond(
     log_info: str | None = None,
     break_condition: BreakConditionT | None = None,
     dataset_dependencies: Mapping[str, Sequence[ParamMeasT]] | None = None,
+    in_memory_cache: bool = True,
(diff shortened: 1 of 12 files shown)


jenshnielsen:
Maybe there should be a way to disable this in the config file?

qcodes/dataset/dond/do_nd.py
```diff
@@ -636,6 +637,10 @@ def dond(
             measurement names to Sequence of Parameters. Note that a dataset must
             depend on at least one parameter from each dimension but can depend
             on one or more parameters from a dimension sweeped with a TogetherSweep.
+        in_memory_cache:
+            Should be keep a cache of the data available in memory for faster
+            plotting and exporting. Useful to disable if the data is very large.
+            If None value will be read from ``qcodesrc.json`` config file.
```
astafan8, line 643:
```suggestion
            Should a cache of the data be kept available in memory for faster
            plotting and exporting. Useful to disable if the data is very large
            in order to save on memory consumption.
            If ``None``, the value for this will be read from ``qcodesrc.json`` config file.
```

PR #5201, opened 2023-06-09
Change R&S ZNB timeout calculation

einsmein:
R&S ZNB timeout was set to sweep time, plus a private attribute called `_additional_wait` .
This PR increases timeout to 1.5 * sweep time, and expose `additional_wait` attribute as public (no leading underscore).

### DIFF
--- a/qcodes/instrument_drivers/rohde_schwarz/ZNB.py
+++ b/qcodes/instrument_drivers/rohde_schwarz/ZNB.py
@@ -363,11 +363,9 @@ class RohdeSchwarzZNBChannel(InstrumentChannel):
             existing_trace_to_bind_to: Name of an existing trace on the VNA.
                 If supplied try to bind to an existing trace with this name
                 rather than creating a new trace.
-            additional_time: Additional wait before instrument timeout.
         """
         n = channel
         self._instrument_channel = channel
-        self.additional_wait = additional_wait
 
         if vna_parameter is None:
             vna_parameter = name
@@ -830,8 +828,8 @@ class RohdeSchwarzZNBChannel(InstrumentChannel):
                     data_format_command = "SDAT"
                 else:
                     data_format_command = "FDAT"
-                timeout = self.sweep_time() * 1.5 + self._additional_wait
-                with self.root_instrument.timeout.set_to(timeout):
+
+                with self.root_instrument.timeout.set_to(self._get_timeout()):
                     # instrument averages over its last 'avg' number of sweeps
                     # need to ensure averaged result is returned
                     for _ in range(self.avg()):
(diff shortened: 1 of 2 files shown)


qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -830,7 +830,7 @@ def _get_sweep_data(self, force_polar: bool = False) -> np.ndarray:
                     data_format_command = "SDAT"
                 else:
                     data_format_command = "FDAT"
-                timeout = self.sweep_time() + self._additional_wait
+                timeout = self.sweep_time() * 1.5 + self._additional_wait
```
jenshnielsen, line 833: I would suggest that rather than exposing additional_wait that we simply allow the user to overwrite timeout.

This could be done as (handling the sentinel value of timeout of None meaning no timeout)
```
old_timeout = self.root_instrument.timeout()
if old_timeout is None:
    old_timeout = float("+inf")
timeout = max(self.sweep_time() * 1.5,  old_timeout)
```

That way a user can simply overwrite timeout if the calculated time is too short.

einsmein: I would expect that if the timeout is not set, then the default `1.5 * self.sweep_time()` should be used, and zero or negative timeout is `+inf`.

jenshnielsen: Yes but that is not how it is implemented in pyvisa. We follow pyvisa except that the qcodes timeout is in sec not ms.

See https://pyvisa.readthedocs.io/en/latest/introduction/resources.html#timeout

```
Here, my_device may be a device, an interface or whatever, and its timeout is set to 25 seconds. To set an infinite timeout, set it to None or float('+inf') or:

del my_device.timeout
Now every operation of the resource takes as long as it takes, even indefinitely if necessary.

To set it to immediate, set it to 0 or a negative value. (Actually, any value smaller than 1 is considered immediate)

Attributes of MessageBase resources[](https://pyvisa.readthedocs.io/en/latest/introduction/resources.html#attributes-of-messagebase-resources)
```

jenshnielsen: thanks @einsmein Could I get you to add a changelog item for this and then I think we are ready to merge

qcodes/instrument_drivers/rohde_schwarz/ZNB.py
```diff
@@ -350,6 +350,7 @@ def __init__(
         channel: int,
         vna_parameter: Optional[str] = None,
         existing_trace_to_bind_to: Optional[str] = None,
+        additional_wait: int = 0,
```
jenshnielsen, line 353: @einsmein This should be removed since no longer used right?

jenshnielsen (changes requested):
Looks good, Lets add a changelog entry and we can remove the argument from the method

jenshnielsen (approved):

PR #5209, opened 2023-06-14
Update run_id when adding ds in mem to db

Dominik-Vogel:
This should fix the issue of run_ids not getting incremented when importing netcdf datasets or datasets in memory in general.
Addressing the issue we have talked about @jenshnielsen .

I have replaced the table name return value, as I could not find it used anywhere and it is not a part of a public API.


jenshnielsen:
@Dominik-Vogel Thanks. Looks like it is used in database_extract_runs.py see typechecking error. 

I think we should see if we can write a test for this too

jenshnielsen:
@Dominik-Vogel I have added tests and implemented the same fix for counter and exp_id. Could you have another look?

astafan8 (approved):

PR #5289, opened 2023-07-27
Instrument Measurement and dond with opentelemetry

jenshnielsen:
This will make it possible to better track the duration and errors within experiments.

TODO

- [x] Changelog
- [x] Clarify the intensions to move away from opencensus
- [x] Some minimal documentation
- [x] Instrument DataSaver Builder

### DIFF
--- a/qcodes/dataset/measurements.py
+++ b/qcodes/dataset/measurements.py
@@ -623,11 +623,10 @@ class Runner:
                 self.ds.subscribe(callble, min_wait=0, min_count=1, state=state)
         self._span.set_attributes(
             {
-                "guid": self.ds.guid,
+                "qcodes_guid": self.ds.guid,
                 "run_id": self.ds.run_id,
-                "captured_run_id": self.ds.captured_run_id,
                 "exp_name": self.ds.exp_name,
-                "sample_name": self.ds.sample_name,
+                "SN": self.ds.sample_name,
                 "ds_name": self.ds.name,
                 "write_in_background": self._write_in_background,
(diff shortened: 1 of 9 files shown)


qcodes/dataset/measurements.py
```diff
@@ -82,7 +84,9 @@ def __init__(
         dataset: DataSetProtocol,
         write_period: float,
         interdeps: InterDependencies_,
+        span: trace.Span,
```
astafan8, line 87: not sure how relevant it is, but perhaps it should be possible to create a datasaver without a span (e.g. with None) for e.g. unit tests? otherwise a span needs to be created always if one wants to create a datasaver?

jenshnielsen: but yes there is no reason not to allow this to be none too

qcodes/dataset/measurements.py
```diff
@@ -605,13 +630,23 @@ def __enter__(self) -> DataSaver:
...
-
+        self._span.set_attributes(
+            {
+                "qcodes_guid": self.ds.guid,
+                "run_id": self.ds.run_id,
+                "exp_name": self.ds.exp_name,
+                "SN": self.ds.sample_name,
+                "ds_name": self.ds.name,
+                "write_in_background": self._write_in_background,
+                "extra_log_info": self._extra_log_info,
```
astafan8, line 647:
```suggestion
                "extra_log_info": self._extra_log_info,
                "dataset_class": self._dataset_class.name,
```

astafan8 (approved):

PR #5291, opened 2023-07-31
Small test cleanups

jenshnielsen:
* Port logging tests to use visa ami driver rather than adaptor
* Remove unused import from example
* Bring ip to visa adaptor in sync with base class (private field has been renamed)
* Add missing close to instrument "name" in test

### DIFF
--- a/qcodes/instrument/ip_to_visa.py
+++ b/qcodes/instrument/ip_to_visa.py
@@ -94,7 +94,7 @@ class IPToVisa(VisaInstrument, IPInstrument):  # type: ignore[misc]
         if hasattr(self, 'connection') and hasattr(self.connection, 'close'):
             self.connection.close()
 
-        strip_attrs(self, whitelist=['_short_name'])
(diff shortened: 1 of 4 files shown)


qcodes/instrument/ip_to_visa.py
```diff
@@ -94,7 +94,7 @@ def close(self) -> None:
         if hasattr(self, 'connection') and hasattr(self.connection, 'close'):
             self.connection.close()
 
-        strip_attrs(self, whitelist=['name'])
```
astafan8, line 97: strip attrs is not gonna strip the "name" property, right? hence thanks to not stripping _short_name the name property will still work after close()?

astafan8 (approved):

qcodes/instrument/ip_to_visa.py
jenshnielsen: Yes exactly. Before if you tried getting the name property it would complain that there is no property name which is confusing. Now you can get name of a closed instrument (but nothing else)

PR #5316, opened 2023-08-14
Oxford Triton: Add control of pumps and temperature channels

FredWolff:
This pull request adds the following parameters to the driver: pid_p, pid_i, pid_d, turb1_speed, turb1_state, comp_state and Tx_state where x represents the temperature channel id and runs from 1 to 17. The pid-parameters set and get the pid-values for the temperature controller. turb1_speed gets the speed of the turbo in Hz. The _state parameters get and set whether the pumps and temperature channels are on or off.


qcodes/instrument_drivers/oxford/triton.py
```diff
@@ -7,148 +7,222 @@
...
 
     Args:
-        tmpfile: Optional: an exported windows registry file from the registry
+        name (str): name of the cryostat.
+        address (str, optional): IP-address of the fridge computer. Defaults to None.
+        port (int, optional): port of the oxford program running on the fridge computer.
+            The relevant port can be found in the manual for the fridge
+            or looked up on the fridge computer. Defaults to None.
+        terminator (str): Defaults to '\r\n'
+        tmpfile (str, optional): an exported windows registry file from the registry
```
jenshnielsen, line 24: Lets remove the types from the docstrings. They are already in the signature and these are likely to fall out of sync

jenshnielsen (approved):

qcodes/instrument_drivers/oxford/triton.py
FredWolff: The types are now removed from the docstring

PR #5327, opened 2023-08-23
Do more validation with ruff.

jenshnielsen:
Replace the usage of pyupgrade and validate pyproject with ruff. 
This is faster and fewer precommit hooks to have installed locally

### DIFF
--- a/qcodes/instrument_drivers/AlazarTech/ATS.py
+++ b/qcodes/instrument_drivers/AlazarTech/ATS.py
@@ -137,7 +137,7 @@ class AlazarTech_ATS(Instrument):
 
         if not self._handle:
             raise Exception(
-                "AlazarTech_ATS not found at " f"system {system_id}, board {board_id}"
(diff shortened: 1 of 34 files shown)


qcodes/instrument_drivers/AlazarTech/ATS.py
```diff
@@ -136,8 +136,9 @@ def __init__(
         self._handle = self.api.get_board_by_system_id(system_id, board_id)
 
         if not self._handle:
-            raise Exception('AlazarTech_ATS not found at '
-                            'system {}, board {}'.format(system_id, board_id))
+            raise Exception(
+                "AlazarTech_ATS not found at " f"system {system_id}, board {board_id}"
```
jenshnielsen, line 140: Will need to check for this kind of thing manually

astafan8 (approved):
yay!

jenshnielsen:
Note that nbqa pyupgrade could also be done with ruff but that requires a bit more work since the notebooks should pass all things

PR #5341, opened 2023-09-03
Use weakref finalize to close visa connections

jenshnielsen:
Partial fix for #3774 it's tricky to do this generally since the finalizer cannot take a reference to the instrument it self and therefore also not a method on this such as `self.close`. 

For visa instruments this is possible to implement since we only need a reference to the visa_handle to close the connection. This could also be implemented more generally if we created a handle/closer class on all instruments. 

This also adds tests that confirms that visa instruments are closed once the instrument is out of scope and gc'ed. 

Closes https://github.com/QCoDeS/Qcodes/issues/5356 by closing the resource_manager in close if the backend is sim. Since this only applies to simulated instruments we don't add it to the finalizer but assume as always that a test must cleanup after it self.

### DIFF
--- a/qcodes/instrument/visa.py
+++ b/qcodes/instrument/visa.py
@@ -140,7 +140,6 @@ class VisaInstrument(Instrument):
         self.set_terminator(terminator)
         self.timeout.set(timeout)
 
-
     def _connect_and_handle_error(
         self, address: str, visalib: str | None
     ) -> tuple[pyvisa.resources.MessageBasedResource, str, pyvisa.ResourceManager]:
--- a/qcodes/instrument/visa.py
+++ b/qcodes/instrument/visa.py
@@ -27,7 +27,9 @@ def _close_visa_handle(
     handle: pyvisa.resources.MessageBasedResource, name: str
 ) -> None:
     log.info(
-        f"Closing VISA handle to {name} as there are no non weak references to the instrument."
+        "Closing VISA handle to %s as there are no non weak "
+        "references to the instrument.",
+        name,
     )
     handle.close()
 
@@ -114,6 +116,7 @@ class VisaInstrument(Instrument):
                 )
         else:
             visa_handle, visabackend = self._connect_and_handle_error(address, visalib)
+        finalize(self, _close_visa_handle, visa_handle, str(self.name))
 
         self.visabackend: str = visabackend
         self.visa_handle: pyvisa.resources.MessageBasedResource = visa_handle
@@ -128,7 +131,7 @@ class VisaInstrument(Instrument):
 
         self.set_terminator(terminator)
         self.timeout.set(timeout)
-        finalize(self, _close_visa_handle, self.visa_handle, str(self.name))
+
 
     def _connect_and_handle_error(
         self, address: str, visalib: str | None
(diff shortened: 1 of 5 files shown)


qcodes/tests/test_visa.py
```diff
@@ -109,6 +112,33 @@ def _make_mock_visa():
...
+        assert list(Instrument._all_instruments.keys()) == ["x"]
+
+    # ensure that any unused instruments that have not been gced are gced before running
+    gc.collect()
+    caplog.clear()
+    with caplog.at_level(logging.INFO, logger="qcodes.instrument.visa"):
+        use_magnet()
+        gc.collect()
+    # at this stage the instrument created in use_magnet has gone out of scope
+    # and we have triggered an explicit gc so the finalize method has been triggered/
```
astafan8, line 132:
```suggestion
    # and we have triggered an explicit gc so the weakref.finalize function has been triggered
```

qcodes/instrument/visa.py
```diff
@@ -22,6 +23,14 @@
 log = logging.getLogger(__name__)
 
 
+def _close_visa_handle(
+    handle: pyvisa.resources.MessageBasedResource, name: str
+) -> None:
+    log.info(
+        f"Closing VISA handle to {name} as there are no non weak references to the instrument."
+    )
+    handle.close()
```
astafan8, line 32: do we need to implement anything here that would work without issues if the handle has already been closed or even gc-collected? Or is weakref.finalize gracefully taking care of that? I guess, because of this concern the log message goes first so that we are sure to get a message about this event even if handle.close() raises?

qcodes/instrument/visa.py
```diff
@@ -119,6 +128,7 @@ def __init__(
 
         self.set_terminator(terminator)
         self.timeout.set(timeout)
+        finalize(self, _close_visa_handle, self.visa_handle, str(self.name))
```
astafan8, line 131: shall we move this line right below self.visa_handle is assigned? so that possible exceptions in the rest of the init can't prevent the weakref.finalize to be "armed"?

qcodes/tests/test_visa.py
```diff
@@ -109,6 +112,33 @@ def _make_mock_visa():
...
+    gc.collect()
+    caplog.clear()
+    with caplog.at_level(logging.INFO, logger="qcodes.instrument.visa"):
+        use_magnet()
+        gc.collect()
+    # at this stage the instrument created in use_magnet has gone out of scope
+    # and we have triggered an explicit gc so the weakref.finalize function has been triggered
+    # We test this
+    # and the instrument should no longer be in the instrument registry
+    assert list(Instrument._all_instruments.keys()) == []
```
astafan8, line 135: should we have an additional assert here on pyvisa.resources.?? (ResourceManager?) to ensure that there are no open handles?

astafan8 (approved):
and a newsfragment? :)

qcodes/tests/test_visa.py
jenshnielsen: Why not, that seems like a good idea

qcodes/instrument/visa.py
jenshnielsen: I don't think so. This has always been called from `__del__` after the instrument was closed but I will check

qcodes/tests/test_visa.py
jenshnielsen: @astafan8  I could not find a way to get from the visa resource to the resource handler. I have implemented this by exposing the resource handler on the instrument which I think is probably a good idea anyway.

astafan8: sounds good!

astafan8 (approved):

PR #5353, opened 2023-09-07
Typecheck with Pyright 1.1.326

jenshnielsen:


jenshnielsen:
Versions in between had some issues related to numpy which has been fixed

astafan8 (approved):

PR #5355, opened 2023-09-11
integrate min requirements job in pytest job

jenshnielsen:
Reduces the code duplication in CI jobs


astafan8 (approved):
Not sure if this is more readable as the min-versions job is a bit too special for my taste, but happy to agree.

jenshnielsen:
> Not sure if this is more readable as the min-versions job is a bit too special for my taste, but happy to agree. 

I am happy with any suggestions for a cleaner way to do that. I have pinged to ask for a new release of requirements-builder. 

Things that I can think of that would make it cleaner would be.

1. Pip had build in support for a min-version strategy.
2. requirements builder would produce a requirement file that could be used as constraint. Then the only difference between the jobs is that you do `pip install .[test] -c requirements.txt` or `pip install .[test] -c min-requirements.txt` That currently does not work since min_requirements.txt contains requirements such as coverage[toml] which is not supported in a constraint file. 


It would also be nice if github matrix jobs rendered the variable name in addition to the value.

Issue #5356, opened 2023-09-11
Pyvisa sim instruments are not cleanup on close.


jenshnielsen:
This can be exposed by reordering tests for example with https://pypi.org/project/pytest-randomly/

It turns out that this is due to a bug in pyvisa-sim requireing the resource manager to also be shutdown.  https://github.com/pyvisa/pyvisa-sim/issues/82

We could perhaps extend close to also kill the resource manager if we are running on a simulated instrument.

jenshnielsen:
Reopened since the solution did not work and is reverted in #5363

A draft solution is in https://github.com/QCoDeS/Qcodes/pull/5360

PR #5360, opened 2023-09-13
Instrument tests: Avoid leaking state between tests

jenshnielsen:
When all pyvisa-sim instruments defined by a yaml file is closed reset the instruments. This in practice will ensure that every test will reset the state of pyvisa-sim instruments on exit since the test must close instruments anyway.

closes https://github.com/QCoDeS/Qcodes/issues/5356

See https://github.com/pyvisa/pyvisa-sim/issues/83 for details

### DIFF
--- a/qcodes/instrument/ip_to_visa.py
+++ b/qcodes/instrument/ip_to_visa.py
@@ -93,7 +93,9 @@ class IPToVisa(VisaInstrument, IPInstrument):  # type: ignore[misc]
         if getattr(self, "visabackend", None) == "sim" and getattr(
             self, "resource_manager", None
         ):
-            # work around for https://github.com/QCoDeS/Qcodes/issues/5356 and
+            # work around for https://github.com/pyvisa/pyvisa-sim/issues/83
+            # see other issues for more context
+            # https://github.com/QCoDeS/Qcodes/issues/5356 and
             # https://github.com/pyvisa/pyvisa-sim/issues/82
             self.resource_manager.visalib._init()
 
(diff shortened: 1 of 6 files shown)


qcodes/instrument/ip_to_visa.py
```diff
@@ -90,6 +90,15 @@ def close(self) -> None:
...
             self.visa_handle.close()
 
+        if getattr(self, "visabackend", None) == "sim" and getattr(
+            self, "resource_manager", None
+        ):
+            # work around for https://github.com/pyvisa/pyvisa-sim/issues/83
+            # see other issues for more context
+            # https://github.com/QCoDeS/Qcodes/issues/5356 and
+            # https://github.com/pyvisa/pyvisa-sim/issues/82
+            self.resource_manager.visalib._init()
```
jenshnielsen, line 100: Perhaps it would be safer if this is the last active session.

qcodes/instrument/ip_to_visa.py
```diff
@@ -90,6 +90,23 @@ def close(self) -> None:
...
+            session_found = self.resource_manager.session in known_sessions
+
+            n_sessions = len(known_sessions)
+            # if this instrument is the last one or there are no connected instruments its safe to reset the device
+            if (session_found and n_sessions == 1) or n_sessions == 0:
+                # work around for https://github.com/pyvisa/pyvisa-sim/issues/83
+                # see other issues for more context
+                # https://github.com/QCoDeS/Qcodes/issues/5356 and
+                # https://github.com/pyvisa/pyvisa-sim/issues/82
+                self.resource_manager.visalib._init()
```
astafan8, line 108: do we want to guard this with try/except for "_init" attribute defined in cacse pyvisa sim changes the implementation and/or name of this fucntion in the future? or is the damage very unlikely and only applicable to "sim" which is only used for tests and hence not in production, hence it's good as is?

astafan8 (approved):

qcodes/instrument/ip_to_visa.py
jenshnielsen: My original thought was that this is only used in testing but there is no harm in downgrading this to a warning if `_init` is not there.

astafan8: > downgrading this to a warning if _init is not there.

i like this!

PR #5381, opened 2023-09-21
Add register_name to register parameters by in datasets

samantha-ho:
This PR adds a `register_name` property and kwarg to the `ParameterBase` class. This property defaults to returning the Parameter `full_name`.

Additionally, this PR modifies the `register_parameter` and `add_result` methods in the Measurement context to use the `register_name` of the parameters rather than the `full_name`. 

These changes are made to split how parameters are organized within an Instrument or Station from how they show up in datasets. This allows greater flexibility on both ends. Attaching a Parameter to an instrument, for example, does not have to change the name of the parameter in the dataset.

These changes have been implemented for ParameterBase and registration has been updated for `ParameterWithSetpoints`. `MultiParameter` and `ArrayParameter` setpoints registration has not been updated in this PR.

To do:
- [x] Add docstrings and an example notebook
- [x] Expand tests
- [x] Add changelog file


qcodes/dataset/measurements.py
```diff
@@ -257,11 +262,13 @@ def _unpack_partial_result(
...
-            parameter = self._interdeps._id_to_paramspec[str(param)]
+            parameter = self._interdeps._id_to_paramspec[str_or_register_name(param)]
         except KeyError:
-            raise ValueError('Can not add result for parameter '
-                             f'{param}, no such parameter registered '
-                             'with this measurement.')
+            raise ValueError(
+                "Can not add result for parameter "
+                f"{param}, no such parameter registered "
+                "with this measurement."
```
astafan8, line 270: perhaps worth adding str_or_register_name(param) to the exception so that in case of this happens users are reminded that the register_name may be part of why this error happened.

qcodes/dataset/measurements.py
```diff
@@ -404,40 +424,50 @@ def _validate_result_deps(
...
+    ) -> None:
         """
         Validate that all sizes of the ``results_dict`` are consistent.
         This means that array-values of parameters and their setpoints are
         of the same size, whereas parameters with no setpoint relation to
         each other can have different sizes.
         """
-        toplevel_params = (set(self._interdeps.dependencies)
-                           .intersection(set(results_dict)))
+        toplevel_params = set(self._interdeps.dependencies).intersection(
```
astafan8, line 440: looks like the formatting of the whole file got changed - in this repo darker is used to make the formatting improvements gradual (by doing them only on the diffs and not the whole files) so that the formatting changes don't distract from the changes in the code. So, see if you can run darker (via pre commit hooks) instead of running black on the whole file.

astafan8 (commented):
looks good!

qcodes/dataset/measurements.py
samantha-ho: Yes. I have my IDE set up to format-on-save using Black. It's not clear to me how to roll back those formatting changes without doing it manually at this point. The other option that comes to mind is completely starting over. Both of these feel equally unappealing.

samantha-ho: If it helps, I believe this is a complete list of the methods which were changed beyond formatting:
DataSaver.add_result
DataSaver._conditionally_expand_parameter_with_setpoints
DataSaver._unpack_partial_result
Measurement.register_parameter
Measurement._register_parameter
Measurement._register_parameter_with_setpoints
str_or_register_name

astafan8: > It's not clear to me how to roll back those formatting changes without doing it manually at this point.

ok, then perhaps indeed too late. Perhaps going forward you could remove the automatic formatting via IDE and rely on pre-commit (which uses darker with black)?

jenshnielsen: I also have my editor setup for auto formatting with black but in QCoDeS I have the following in my `.vscode\settings.json` file to avoid this taking effect on qcodes

```
"[python]": {
    "editor.defaultFormatter": "ms-python.black-formatter",
    "editor.formatOnType": true,
    "editor.formatOnSave": false
    },
...
```

qcodes/dataset/__init__.py
```diff
@@ -41,7 +41,7 @@
     datasaver_builder,
     dond_into,
 )
-from .measurements import Measurement
+from .measurements import Measurement, str_or_register_name
```
astafan8, line 44: what is the purpose of this function being public? I'd think that it's very specific to the implementation of the Measurement class, and does not need to be used outside at all.

docs/changes/newsfragments/5381.new
```diff
@@ -0,0 +1,2 @@
+Add a register_name keyword to the ParameterBase constructor. If supplied, this value
+will be used to register parameters in datasets instead of the default full_name.
```
astafan8, line 2:
```suggestion
will be used to register parameters in datasets instead of the default full_name.
Note that the new register_name keyword argument does not work for MultiParameter and ArrayParameter.
```

qcodes/parameters/parameter_base.py
```diff
@@ -173,6 +173,9 @@ class ParameterBase(MetadatableWithName):
 
         bind_to_instrument: Should the parameter be registered as a delegate attribute
             on the instrument passed via the instrument argument.
+
+        register_name: Specifies if the parameter should be registered in datasets
+            using a different name than the parameter full_name
```
astafan8, line 178:
```suggestion
            using a different name than the parameter's full_name
```

qcodes/parameters/parameter_base.py
```diff
@@ -830,6 +834,15 @@ def full_name(self) -> str:
         """
         return "_".join(self.name_parts)
 
+    @property
+    def register_name(self) -> str:
+        """
+        Name that will be used to register this parameter in a dataset
+        By default, this returns the full_name, but may be overriden if custom
+        behavior is desired.
```
astafan8, line 842:
```suggestion
        By default, this returns the ``full_name``, or the value of
        ``register_name`` argument if it was passed at inititalization of the parameter.
        This property may be overriden if custom behavior is desired.
```

astafan8 (approved):

qcodes/parameters/parameter_base.py
samantha-ho: On second thought, there's no need to include the last sentence, as all python properties can be overriden to produce custom behavior.

PR #5390, opened 2023-09-27
Update index.rst (Add instructions about fetching tags)

DCEM:
Added a section to the installation guide advising users to fetch tags from the upstream QCoDeS repository when installing from a forked repository.


jenshnielsen (approved):
thanks

PR #5391, opened 2023-09-28
Use Dask delayed to export large datasets to NetCDF

jenshnielsen:
Avoid loading a potentially huge dataset into memory just to be able to rewrite it to disk. 

This writes each row of a dataset to disk individually and then uses dask/xarray to lazily merge the dataset into one file

### DIFF
--- a/docs/changes/newsfragments/5391.new
+++ b/docs/changes/newsfragments/5391.new
@@ -1,2 +1,3 @@
 Large datasets are now exported to NetCDF4 using Dask delayed writer.
 This avoids allocating a large amount of memory to process the whole dataset at the same time.
+Size threshold at the moment is set to approximately 1 GB.
--- a/docs/changes/newsfragments/5391.new
+++ b/docs/changes/newsfragments/5391.new
@@ -1,0 +1,2 @@
+Large datasets are now exported to NetCDF4 using Dask delayed writer.
+This avoids allocating a large amount of memory to process the whole dataset at the same time.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ dependencies = [
     "xarray>=2022.06.0",
     "cf_xarray>=0.8.4",
     "opentelemetry-api>=1.15.0",
-    "dask>=2022.1.0",
+    "dask>=2022.1.0", # we are making use of xarray features that requires dask implicitly
     # transitive dependencies. We list these explicitly to",
     # ensure that we always use versions that do not have",
     # known security vulnerabilities",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
     "ruamel.yaml>=0.16.0,!=0.16.6",
     "tabulate>=0.8.0",
     "typing_extensions>=4.1.1",
-    "tqdm>=4.32.2",
+    "tqdm>=4.59.0",
     "uncertainties>=3.1.4",
     "versioningit>=2.0.1",
     "websockets>=9.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,7 @@ dependencies = [
     "xarray>=2022.06.0",
     "cf_xarray>=0.8.4",
     "opentelemetry-api>=1.15.0",
+    "dask>=2022.1.0",
     # transitive dependencies. We list these explicitly to",
     # ensure that we always use versions that do not have",
     # known security vulnerabilities",
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -16,7 +16,6 @@ from typing import TYPE_CHECKING, Any
 
 import numpy
 from tqdm.auto import trange
-from tqdm.dask import TqdmCallback
 
 import qcodes
 from qcodes.dataset.data_set_protocol import (
@@ -97,6 +96,7 @@ from .exporters.export_to_pandas import (
 from .exporters.export_to_xarray import (
     load_to_xarray_dataarray_dict,
     load_to_xarray_dataset,
+    xarray_to_h5netcdf_with_complex_numbers,
 )
 from .subscriber import _Subscriber
 
@@ -1473,17 +1473,14 @@ class DataSet(BaseDataSet):
             with tempfile.TemporaryDirectory() as temp_dir:
                 temp_path = Path(temp_dir)
                 for i in trange(len(self), desc="Writing individual files"):
-                    self.to_xarray_dataset(start=i + 1, end=i + 1).to_netcdf(
-                        temp_path / f"ds_{i:03d}.nc", engine="h5netcdf"
+                    xarray_to_h5netcdf_with_complex_numbers(
+                        self.to_xarray_dataset(start=i + 1, end=i + 1),
+                        temp_path / f"ds_{i:03d}.nc",
                     )
                 files = tuple(temp_path.glob("*.nc"))
                 data = xr.open_mfdataset(files)
                 try:
-                    write_job = data.to_netcdf(
-                        file_path, compute=False, engine="h5netcdf"
-                    )
-                    with TqdmCallback(desc="Combining files"):
-                        write_job.compute()
+                    xarray_to_h5netcdf_with_complex_numbers(data, file_path)
                 finally:
                     data.close()
         else:
--- a/qcodes/dataset/data_set.py
+++ b/qcodes/dataset/data_set.py
@@ -4,6 +4,7 @@ import importlib
 import json
 import logging
 import sys
+import tempfile
 import time
 import uuid
 from collections.abc import Callable, Mapping, Sequence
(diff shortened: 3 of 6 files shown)


qcodes/dataset/data_set_protocol.py
```diff
@@ -2,6 +2,9 @@
 
 import sys
 
+from dask.diagnostics.progress import ProgressBar
```
jenshnielsen, line 5: Add dask requirement. Should this be optional?

qcodes/dataset/data_set_protocol.py
```diff
@@ -242,6 +245,22 @@ def get_parameter_data(
...
 
+    def _estimate_ds_size(self) -> float:
+        """
+        Give an estimated size of the dataset as the size of a single row
+        times the len of the dataset. Result is returned in Mega Bytes.
+
+        Note that this does not take overhead into account so it works best
+        if the row size is "large"
+        """
+        sample_data = self.get_parameter_data(start=1, end=1)
```
jenshnielsen, line 256: How does this work with dataset in memory?

qcodes/dataset/data_set_protocol.py
```diff
@@ -465,9 +484,25 @@ def _export_file_name(self, prefix: str, export_type: DataExportType) -> str:
...
     def _export_as_netcdf(self, path: Path, file_name: str) -> Path:
         """Export data as netcdf to a given path with file prefix"""
+        import xarray as xr
         file_path = path / file_name
-        xarr_dataset = self.to_xarray_dataset()
-        xarray_to_h5netcdf_with_complex_numbers(xarr_dataset, file_path)
+        if self._estimate_ds_size() > 1000:
+            print("large dataset export.")
+            temp_dir = Path(f"./temp_export_{self.guid}/")
+            temp_dir.mkdir(exist_ok=True)
```
jenshnielsen, line 492: This dir should probably be cleaned up automatically

qcodes/dataset/data_set_protocol.py
```diff
@@ -465,9 +484,25 @@ def _export_file_name(self, prefix: str, export_type: DataExportType) -> str:
...
+            for i in tqdm(range(len(self))):
+                self.to_xarray_dataset(start=i + 1, end=i + 1).to_netcdf(
+                    str(temp_dir / f"ds_{i:03d}.nc"), engine="h5netcdf"
+                )
+            files = [f for f in temp_dir.glob("*.nc")]
+            data = xr.open_mfdataset(files)
+            write_job = data.to_netcdf(file_path, compute=False, engine="h5netcdf")
+            with ProgressBar():
+                print(f"Writing to {file_path}")
```
jenshnielsen, line 501: Replace all prints with logging

qcodes/dataset/data_set_protocol.py
```diff
@@ -465,9 +484,25 @@ def _export_file_name(self, prefix: str, export_type: DataExportType) -> str:
 
     def _export_as_netcdf(self, path: Path, file_name: str) -> Path:
         """Export data as netcdf to a given path with file prefix"""
+        import xarray as xr
         file_path = path / file_name
-        xarr_dataset = self.to_xarray_dataset()
-        xarray_to_h5netcdf_with_complex_numbers(xarr_dataset, file_path)
+        if self._estimate_ds_size() > 1000:
```
jenshnielsen, line 489: Should the limit be configurable. Should we be able to disable this

astafan8: yes. maybe.

qcodes/dataset/data_set_protocol.py
```diff
@@ -465,9 +484,25 @@ def _export_file_name(self, prefix: str, export_type: DataExportType) -> str:
...
+            print("large dataset export.")
+            temp_dir = Path(f"./temp_export_{self.guid}/")
+            temp_dir.mkdir(exist_ok=True)
+            for i in tqdm(range(len(self))):
+                self.to_xarray_dataset(start=i + 1, end=i + 1).to_netcdf(
+                    str(temp_dir / f"ds_{i:03d}.nc"), engine="h5netcdf"
+                )
+            files = [f for f in temp_dir.glob("*.nc")]
+            data = xr.open_mfdataset(files)
```
astafan8, line 498: check that attributes are still present in the final dataset

qcodes/dataset/data_set_protocol.py
jenshnielsen: Yes that seems ok. Note that could could also depend on xarray[parallel] to pull in dask but that installs dask with the complete subtarget which has even more dependencies

qcodes/dataset/data_set_protocol.py
jenshnielsen: It does not. Moved to the sqlite dataset

pyproject.toml
```diff
@@ -48,6 +48,7 @@ dependencies = [
     "xarray>=2022.06.0",
     "cf_xarray>=0.8.4",
     "opentelemetry-api>=1.15.0",
+    "dask>=2022.1.0",
```
jenshnielsen, line 51: Document why this is a dependency with a comment

qcodes/dataset/data_set.py
```diff
@@ -1457,19 +1458,76 @@ def _set_export_info(self, export_info: ExportInfo) -> None:
...
+                        "qcodes_guid": self.guid,
+                        "ds_name": self.name,
+                        "exp_name": self.exp_name,
+                        "temp_dir": temp_dir,
+                    },
+                )
+                for i in trange(len(self), desc="Writing individual files"):
+                    xarray_to_h5netcdf_with_complex_numbers(
+                        self.to_xarray_dataset(start=i + 1, end=i + 1),
+                        temp_path / f"ds_{i:03d}.nc",
```
jenshnielsen, line 1494: Don't hardcode

qcodes/dataset/data_set.py
```diff
@@ -1457,19 +1458,81 @@ def _set_export_info(self, export_info: ExportInfo) -> None:
...
+                "Dataset is expected to be larger that threshold. Using distributed export.",
+                extra={
+                    "file_name": file_path,
+                    "qcodes_guid": self.guid,
+                    "ds_name": self.name,
+                    "exp_name": self.exp_name,
+                },
+            )
+            print(
+                "Large dataset detected. Will write to individual files and combine to reduce memory overhead."
```
astafan8, line 1477:
```suggestion
                "Large dataset detected. Will write to multiple files first and combine after, to reduce memory overhead."
```

qcodes/dataset/data_set.py
```diff
@@ -1457,19 +1458,81 @@ def _set_export_info(self, export_info: ExportInfo) -> None:
...
+
+        if self._estimate_ds_size() > self._export_limit:
+            file_path = path / file_name
+            log.info(
+                "Dataset is expected to be larger that threshold. Using distributed export.",
+                extra={
+                    "file_name": file_path,
+                    "qcodes_guid": self.guid,
+                    "ds_name": self.name,
+                    "exp_name": self.exp_name,
```
astafan8, line 1473:
```suggestion
                    "exp_name": self.exp_name,
                    "_export_limit": self._export_limit,
                    "_estimated_ds_size": self._estimate_ds_size(),
```

qcodes/dataset/data_set.py
```diff
@@ -1457,19 +1458,81 @@ def _set_export_info(self, export_info: ExportInfo) -> None:
...
+        else:
+            file_path = super()._export_as_netcdf(path=path, file_name=file_name)
+        return file_path
+
+    def _estimate_ds_size(self) -> float:
+        """
+        Give an estimated size of the dataset as the size of a single row
+        times the len of the dataset. Result is returned in Mega Bytes.
+
+        Note that this does not take overhead into account so it works best
```
astafan8, line 1526:
```suggestion
        Note that this does not take overhead into account so it is more accurate
```

docs/changes/newsfragments/5391.new
```diff
@@ -0,0 +1,2 @@
+Large datasets are now exported to NetCDF4 using Dask delayed writer.
+This avoids allocating a large amount of memory to process the whole dataset at the same time.
```
astafan8, line 2:
```suggestion
This avoids allocating a large amount of memory to process the whole dataset at the same time.
Size threshold at the moment is set to approximately 1 GB.
```

qcodes/tests/dataset/test_dataset_export.py
```diff
@@ -675,6 +714,75 @@ def test_export_2d_dataset(
...
+    assert_allclose(loaded_ds.z, expected_z)
+
+    _assert_xarray_metadata_is_as_expected(loaded_ds, mock_dataset_numpy)
+
+
+def test_export_numpy_dataset_complex(
+    tmp_path_factory: TempPathFactory, mock_dataset_numpy_complex: DataSet, caplog
+) -> None:
+    tmp_path = tmp_path_factory.mktemp("export_netcdf")
+    mock_dataset_numpy_complex._export_limit = 0
```
astafan8, line 754: What do you think about having 1 test that asserts that with the currently-set export limit we are NOT using multi-file/delayed-dask export? So that we know that simple export is also working.

astafan8 (approved):

qcodes/tests/dataset/test_dataset_export.py
jenshnielsen: sure no harm in that

PR #5399, opened 2023-09-30
Update qcodes to support numpy 2.0

peendebak:
Numpy 2.0 is scheduled to be released this year (https://github.com/numpy/numpy/issues/24300). In this PR we adapt qcodes to work with numpy 2.0.

The `numpy_non_concrete_floats_instantiable` and `numpy_non_concrete_complex_instantiable` are now empty, perhaps it is better to just remove them. (they are public, but not used in the qcodes codebase)


jenshnielsen:
> The numpy_non_concrete_floats_instantiable and numpy_non_concrete_complex_instantiable are now empty, perhaps it is better to just remove them. (they are public, but not used in the qcodes codebase)

They are added to `numpy_floats` and `numpy_non_concrete_complex_instantiable` I think its better to remove them and add a breaking change notice.

PR #5420, opened 2023-10-04
Run pytest using worksteal

jenshnielsen:
This safes nearly 50% of runtime when running the tests locally for me on a 6 core machine.

This is possible now since most(all?) leaking state between the tests has been fixed.

https://pytest-xdist.readthedocs.io/en/stable/distribution.html#running-tests-across-multiple-cpus

Also set the deadline higher on a test that often fail for me locally


jenshnielsen:
Unfortunatly this seems to be significantly slower on windows

PR #5421, opened 2023-10-04
Update quantify doc links

ereehuis:
Quick MR to update Quantify documentation links as we've moved to new domain


jenshnielsen (approved):
Thansk for keeping this up to date.

PR #5426, opened 2023-10-06
Allow additional validators on parameter at runtime

jenshnielsen:
This allows the user to add additional validators to a parameter at runtime either permanently (until explicitly removed) or temporarily via a context manager.

TODO
- [ ] Docstrings
- [ ] Examples


qcodes/tests/parameter/test_validators.py
```diff
@@ -0,0 +1,137 @@
...
+from hypothesis import given
+
+from qcodes.parameters import Parameter
+from qcodes.validators import Ints
+
+
+def test_add_ints_validator_to_parameter():
+    p = Parameter("test_param", set_cmd=None, get_cmd=None)
+    p.add_validator(Ints(min_value=0, max_value=10))
+    assert len(p.validators) == 1
```
astafan8, line 12:
```suggestion
    ints_val = Ints(min_value=0, max_value=10)
    p.add_validator(ints_val)
    assert len(p.validators) == 1
    assert p.validators[0] is ints_val
    assert p.vals is ints_val
```

qcodes/tests/parameter/test_validators.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def test_remove_ints_validator_from_parameter():
+    p = Parameter("test_param", set_cmd=None, get_cmd=None)
+    p.add_validator(Ints(min_value=0, max_value=10))
+    p.remove_validator()
+    assert len(p.validators) == 0
```
astafan8, line 19:
```suggestion
    assert len(p.validators) == 0
    assert p.vals is None
```

qcodes/tests/parameter/test_validators.py
```diff
@@ -0,0 +1,137 @@
...
+
+        for validator in validators[1:]:
+            p.add_validator(validator)
+    else:
+        p = Parameter("test_param", set_cmd=None, get_cmd=None)
+
+        for validator in validators:
+            p.add_validator(validator)
+
+    assert len(p.validators) == n_validators
```
astafan8, line 56:
```suggestion
    assert len(p.validators) == n_validators
    if len(validators) > 0:
        assert p.vals is validators[0]
```

qcodes/tests/parameter/test_validators.py
```diff
@@ -0,0 +1,137 @@
...
+    if value_to_validate >= max_min_value and value_to_validate <= min_max_value:
+        p.validate(value_to_validate)
+    else:
+        with pytest.raises(ValueError):
+            p.validate(value_to_validate)
+
+    while len(p.validators) > 0:
+        p.remove_validator()
+
+    assert len(p.validators) == 0
```
astafan8, line 67:
```suggestion
    assert len(p.validators) == 0
    assert p.vals is None
```

qcodes/tests/parameter/test_validators.py
```diff
@@ -0,0 +1,137 @@
...
+    assert len(p.validators) == n_validators
+
+    if value_to_validate >= max_min_value and value_to_validate <= min_max_value:
+        p.validate(value_to_validate)
+    else:
+        with pytest.raises(ValueError):
+            p.validate(value_to_validate)
+
+    while len(p.validators) > 0:
+        p.remove_validator()
```
astafan8, line 65:
```suggestion
        p.remove_validator()
        if len(p.validators) > 0:
            assert p.vals is p.validators[0]
```

astafan8 (commented):
Looks great!

qcodes/parameters/parameter_base.py
```diff
@@ -307,6 +312,55 @@ def __init__(
...
+            self._vals = [vals]
+        elif len(self._vals) == 1:
+            self._vals = []
+        elif len(self._vals) > 1:
+            raise RuntimeError(
+                "Cannot remove default validator from parameter with additional validators."
+            )
+        else:
+            # setting the validator to None but the parameter already doesn't have a validator
+            pass
```
astafan8, line 339: maybe it's a taste thing, but i found the below more readable as it more epxlicitly separates the None/not-None cases
```suggestion
        if vals is not None:
            if len(self._vals) > 0:
                self._vals[0] = vals
            else:
                self._vals = [vals]
        else:  # vals is None
            if len(self._vals) == 1:
                self._vals = []
            elif len(self._vals) > 1:
                raise RuntimeError(
                    "Cannot remove default validator from parameter with additional validators."
                )
            else:
                # setting the validator to None but the parameter already doesn't have a validator
                pass
```

qcodes/parameters/parameter_base.py
```diff
@@ -307,6 +312,55 @@ def __init__(
...
+    def vals(self) -> Validator | None:
+        if len(self._vals):
+            return self._vals[0]
+        else:
+            return None
+
+    @vals.setter
+    def vals(self, vals: Validator | None) -> None:
+        if vals is not None and len(self._vals) > 0:
+            self._vals[0] = vals
```
astafan8, line 328: i think this case still needs a test, right? a non-trivial one))

qcodes/parameters/parameter_base.py
jenshnielsen: I don't really care either way but many other people find nested if statements a huge code smell

qcodes/parameters/parameter_base.py
jenshnielsen: This case is not new but existing functionality that should already be tested but sure why not add a new one

astafan8: well, the exsting tests don't assert for the case where there are more than 2 validators - so the assertion would be that the 2nd validator stays instact after `.vals` is set to something new

astafan8 (approved):

astafan8:
oh! a news fragment?

PR #5427, opened 2023-10-06
Switch qcodes to src layout

jenshnielsen:
This is mainly due to the fact that editable mode does not work well for flat layout https://github.com/pypa/setuptools/issues/3518 which does not seem likely to be fixed anytime soon. This also has the benefit of making it cleaner that tests are executed against an installed version of the code since the src is not on path by default.

It is expected that darker will fail on this branch since all the files counts as new when moved but I have not linted them


astafan8 (approved):
looks good.

and in qcodes we don't use __file__ of a module for anything, that might break, right?

PR #5467, opened 2023-10-29
Implement 2 more ruff rules

jenshnielsen:
Implement

* RUF010: avoid str and repr in f strings
* RUF012: Use ClassVar for mutable classvariables to prevent mutation on an instance level (which would affect all instances)

### DIFF
--- a/src/qcodes/validators/validators.py
+++ b/src/qcodes/validators/validators.py
@@ -171,7 +171,7 @@ class Bool(Validator[Union[bool, np.bool_]]):
             TypeError: IF not a boolean.
         """
         if not isinstance(value, bool) and not isinstance(value, np.bool_):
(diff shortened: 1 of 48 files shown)


src/qcodes/validators/validators.py
```diff
@@ -411,7 +410,7 @@ def validate(self, value: numbertypes, context: str = "") -> None:
                 castvalue = intrepr
             else:
                 raise TypeError(
-                    f"{repr(value)} is not an int or close to an int; {context}"
+                    f"{value:r} is not an int or close to an int; {context}"
```
jenshnielsen, line 413:
```suggestion
                    f"{value!r} is not an int or close to an int; {context}"
```

astafan8 (approved):

PR #5485, opened 2023-11-06
Visa allow setting custom resource args.

jenshnielsen:
Add visa_kwargs argument to visa instrument init to allow settings to be passed to pyvisas `open_resource`.

TODO
- [ ] News fragment
- [ ] Can this be tested with pyvisa-sim


jenshnielsen:
This could also be done by allowing the user to pass an open visa resource to the instrument and using that one rather than creating a new one. Perhaps that is cleaner (it's at least more type safe)

jenshnielsen:
replaced by #7978

PR #5499, opened 2023-11-09
upgrade pyright to 1.1.334

jenshnielsen:
And remove two no longer required ignores


jenshnielsen:
Not upgrading to 1.1.335 due to https://github.com/microsoft/pyright/issues/6387

astafan8 (approved):

PR #5507, opened 2023-11-10
Make warnings hard errors in the test suite

jenshnielsen:
Except for the min versions job where we ignore them.


jenshnielsen:
This does not work since the min ver job still triggers lots of deprecation warnings

astafan8 (approved):

PR #5517, opened 2023-11-14
Fix expansion of multi index when loading im memory cache from a netcdf file

jenshnielsen:
- [x] News fragment


jpdehollain (approved):
I tested it and it solves my issue

astafan8 (approved):

PR #5520, opened 2023-11-14
Lakeshore 335 Output: Add parameter mapping for input=None

jbeckerwsi:
According to the manual (https://www.lakeshore.com/docs/default-source/product-downloads/335_manual.pdf?sfvrsn=9f5c5b7f_7) , page 121:

> `input` Specifies which input to use for control: 0 = None, 1 = A, 2 = B

Here, we add the parameter mapping for the value 0.

### DIFF
--- a/docs/changes/newsfragments/5520.improved
+++ b/docs/changes/newsfragments/5520.improved
@@ -1,0 +1,1 @@
+Lakeshore 335 Output: Add parameter mapping for input=None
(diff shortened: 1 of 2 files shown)


jenshnielsen:
@jbeckerwsi thanks. Would you mind adding a news fragment for the change?

jenshnielsen (approved):

PR #5563, opened 2023-11-29
Add pyvisa sim yaml file for RohdeSchwarz SGS100A

jasonl33:
Adds a pyvisa sim yaml file for the RohdeSchwarz SGS100A driver.

Basic unit tests to validate each supported parameter has been included.


tests/drivers/test_RS_SGS100A.py
```diff
@@ -0,0 +1,80 @@
...
+from typing import Any
+
+import numpy as np
+import pytest
+
+from qcodes.instrument_drivers.rohde_schwarz.SGS100A import RohdeSchwarz_SGS100A
+
+
+@pytest.fixture(scope='function')
+def sg():
```
jenshnielsen, line 11:
```suggestion
@pytest.fixture(scope='function', name="sg")
def _make_sg():
```

If you follow this pattern you can avoid the pylint warning about name shadowing name from outer scope

jenshnielsen (approved):

tests/drivers/test_RS_SGS100A.py
jasonl33: I was wondering how to address that.  Thanks for the suggestion!

PR #5568, opened 2023-12-02
Fix typo in keithley 2450 driver

aaronsharpe:
Fix typo in keithley 2450. Parameter block_during_ramp is not callable

### DIFF
--- a/src/qcodes/instrument_drivers/Keithley/Keithley_2450.py
+++ b/src/qcodes/instrument_drivers/Keithley/Keithley_2450.py
@@ -607,7 +607,6 @@ class Keithley2450(VisaInstrument):
 
         self.add_parameter(
             "output_enabled",
-            initial_value="0",
             set_cmd=":OUTP {}",
             get_cmd=":OUTP?",
             val_mapping=create_on_off_val_mapping(on_val="1", off_val="0"),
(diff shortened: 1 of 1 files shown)


src/qcodes/instrument_drivers/Keithley/Keithley_2450.py
```diff
@@ -485,7 +485,7 @@ def __init__(self, parent: "Keithley2450", name: str, proper_function: str) -> N
 
     def _set_proper_function(self, value: float) -> None:
         self.write(f"SOUR:{self._proper_function} {value}")
-        if self.block_during_ramp():
+        if self.block_during_ramp:
```
jenshnielsen, line 488: @aaronsharpe I am a bit confused by this. `block_during_ramp` is a QCoDeS parameter. This means that it is callable. Calling it without arguments is equivalent to calling `.get()` on the parameter. With your change this would test an instance of the parameter class as a bool which will always evaluate to True.


```

In [1]: from qcodes.parameters import Parameter
In [2]: a = Parameter("a")
In [3]: bool(a)
Out[3]: True
```

aaronsharpe: Not sure what happened. On one of our setups, it was throwing an error but not on my test rig. Will delete pull request pending what I find

PR #5572, opened 2023-12-04
Remove initial value for output of Keithley 2450

aaronsharpe:
Setting an initial value for the keithley output is quite dangerous, especially when recovering from a kernel crash. Initializing the instrument should leave the output state unchanged


aaronsharpe:
Sorry about the mix up on the last pull request. Let me know if this change warrants newsfragment

jenshnielsen:
@aaronsharpe thanks, I think it would be good with a neswfragment since this changes the behaviour even if it’s for the better

aaronsharpe:
Added. Thanks for the feedback

jenshnielsen (approved):

PR #5589, opened 2023-12-11
Smarter selection of multi index export to xarray

jenshnielsen:
* Only automatically select multi index export if shapes are unset
* Allow the user to select which export type to use

- [x] Tests?
- [x] Changelog

### DIFF
--- a/docs/changes/newsfragments/5589.improved
+++ b/docs/changes/newsfragments/5589.improved
@@ -1,0 +1,5 @@
+The use of multi index when exporting to xarray (and netcdf files) has been made smarter
+such that any dataset with a known shape (such as those measured by doNd etc) will
+never be exported using multi index even in the case of incomplete datasets (i.e. due to an interrupted measurement).
+Furthermore `to_xarray_dataset` and `to_xarray_dataarray_dict` have gained a key word argument `use_multi_index` to allow the user
+to control the use of multi indexes.
--- a/docs/changes/newsfragments/5589.improved
+++ b/docs/changes/newsfragments/5589.improved
@@ -1,5 +1,5 @@
 The use of multi index when exporting to xarray (and netcdf files) has been made smarter
 such that any dataset with a known shape (such as those measured by doNd etc) will
-never be exported using multi index even in the case of incomplete datasets (i.e. due to an interrupted measurement).
+never be automatically exported using multi index even in the case of incomplete datasets (i.e. due to an interrupted measurement).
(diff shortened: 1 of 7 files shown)


src/qcodes/dataset/data_set.py
```diff
@@ -1052,7 +1059,11 @@ def to_xarray_dataset(
                 if None
             end: end value of selection range (by results count); ignored if
                 None
-
+            use_multi_index: Should the data be exported using a multi index.
+                If the data is not on a grid this will create a xarray dataset
+                without missing datapoints. If set to "auto" multi index will be
```
astafan8, line 1064: "without missing datapoints" - could this be said differently such that the word "missing" does not trigger thinking about data points that were missed during a measurement or alike? perhaps "... only with datapoints that are existing in the dataset, that is without trying to fit the data onto a grid thus creating a very sparsely-popuated grid" although it's lengthy. Is there any vocabulary we can take from xarray?))

astafan8 (commented):
Looks good! I think a test would be good idea - same dataset with auto/never values covered, so that also if anything changes on xarray/pandas side we can catch it.

jenshnielsen:
Testing plan

- [ ] For a complete dataset on a grid without shapes set:  Test that multiindex is used iff always always selected
- [ ] For a complete dataset on grid with shape: The same
- [ ] For a incomplete dataset on grid with shape: The same
- [ ] For a dataset not on a grid: muitiindex: test multi index triggered iff always or auto
- [ ] For an incomplte dataset on a grid: test multi index triggered iff always or auto

astafan8 (approved):
Beautiful clear tests!

docs/changes/newsfragments/5589.improved
```diff
@@ -0,0 +1,5 @@
+The use of multi index when exporting to xarray (and netcdf files) has been made smarter
+such that any dataset with a known shape (such as those measured by doNd etc) will
+never be exported using multi index even in the case of incomplete datasets (i.e. due to an interrupted measurement).
```
astafan8, line 3: if the use_multi_index is always, the export of a dataset with shape will still use multi index, while here it says "never" - so perhaps add "automatically" or rephrase in a different way, so that it's clear that the flag can be used to force whatever behavior the user desires?
```suggestion
never be automatically exported using multi index even in the case of incomplete datasets (i.e. due to an interrupted measurement).
```

PR #5600, opened 2023-12-20
Fix XArray FutureWarning

jenshnielsen:
fixes 
> FutureWarning: The return type of `Dataset.dims` will be changed to return a set of dimension names in future, in order to be
> more consistent with `DataArray.dims`. To access a mapping from dimension names to lengths, please use `Dataset.sizes`.

This fixes all places where we make use of .dims as a dict There are a number of places where we use it to get the names/number of the dimensions. Here we continue using .dims


jenshnielsen:
For reference dims only causes a warning when used as a dict due to the implementation here https://github.com/TomNicholas/xarray/blob/ddbab7214622e382c2a4dd7a51721a5394500f3b/xarray/core/utils.py#L479

astafan8 (approved):

PR #5618, opened 2024-01-04
Make delayed/chunked export configurable and off by default

jenshnielsen:
There are unfortunately a few too many corner cases where this does not work well so leaving it off by default.
There may still be cases where this is useful so it can be enabled in those cases by temporarily setting the config. For now I have not added this as an option to the export method since it does not really seem to fit well in the api

TODO

- [x] News fragment
- [x] Are the config option names what we want

### DIFF
--- a/src/qcodes/configuration/qcodesrc.json
+++ b/src/qcodes/configuration/qcodesrc.json
@@ -76,6 +76,8 @@
         "export_prefix": "qcodes_",
         "export_path": "{db_location}",
         "export_name_elements": ["captured_run_id", "guid"],
+        "export_chunked_export_of_large_files": false,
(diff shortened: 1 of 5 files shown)


src/qcodes/configuration/qcodesrc.json
```diff
@@ -76,6 +76,8 @@
         "export_prefix": "qcodes_",
         "export_path": "{db_location}",
         "export_name_elements": ["captured_run_id", "guid"],
+        "export_chunked_export_of_large_files": false,
```
astafan8, line 79: Perhaps add _enabled to the end?

astafan8 (approved):

PR #5620, opened 2024-01-04
RFC: Drop python 39

jenshnielsen:
Most of the scientific python stack has adopted [SPEC 0](https://scientific-python.org/specs/spec-0000/) which means that python 3.9 support should be dropped from late 2023. As an example we can see in https://github.com/QCoDeS/Qcodes/pull/5602 that Ipython has dropped support for 3.9.

This is to propose that QCoDeS also drops support for 3.9 

- [x] This should also update the readme
- [x] Remove 3.9 ci jobs

### DIFF
--- a/src/qcodes/dataset/data_set_protocol.py
+++ b/src/qcodes/dataset/data_set_protocol.py
@@ -2,10 +2,10 @@ from __future__ import annotations
 
 import logging
 import os
-import sys
 import warnings
 from collections.abc import Callable, Mapping, Sequence
 from enum import Enum
+from importlib.metadata import entry_points
 from pathlib import Path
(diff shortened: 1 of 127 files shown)


astafan8 (approved):
exciting :) and removing 3.9 ci and updating readme is still todo

src/qcodes/dataset/data_set_protocol.py
```diff
@@ -61,17 +54,17 @@
 # even with from __future__ import annotations
 # type aliases must use the old format until we drop 3.8/3.9
```
jenshnielsen, line 55:
```suggestion
```
No longer relevant

PR #5627, opened 2024-01-08
Avoid roundtrip to qcodes format when loading as xarray from netcdf

jenshnielsen:
Todo

- [x] add tests that calling to_xarray_dataset does not trigger the conversion code 
- [x] Newsfragment

### DIFF
--- a/docs/changes/newsfragments/5627.improved
+++ b/docs/changes/newsfragments/5627.improved
@@ -1,3 +1,3 @@
 When loading a QCoDeS dataset from a netcdf file using load_from_netcdf and converted to a Xarray dataset using
 ``to_xarray_dataset`` or ``cache.to_xarray_dataset`` we avoid converting the data to QCoDeS format and back to Xarray format.
-This should safe time and avoid any potential corner cases when roundtripping the data.
+This should save time and avoid any potential corner cases when roundtripping the data.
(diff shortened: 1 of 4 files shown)


jenshnielsen:
Done on top of https://github.com/QCoDeS/Qcodes/pull/5320 which should first be reviewed

src/qcodes/dataset/data_set_in_memory.py
```diff
@@ -298,8 +298,7 @@ def _load_from_netcdf(
             export_info=export_info,
```
jenshnielsen, line 298: Above we should use open_dataset to lazily open the dataset and pass a reference to the dataset path down to the laze cache to avoid loading any real data before its required

docs/changes/newsfragments/5627.improved
```diff
@@ -0,0 +1,3 @@
+When loading a QCoDeS dataset from a netcdf file using load_from_netcdf and converted to a Xarray dataset using
+``to_xarray_dataset`` or ``cache.to_xarray_dataset`` we avoid converting the data to QCoDeS format and back to Xarray format.
+This should safe time and avoid any potential corner cases when roundtripping the data.
```
astafan8, line 3:
```suggestion
This should save time and avoid any potential corner cases when roundtripping the data.
```

astafan8 (approved):

PR #5631, opened 2024-01-09
shut down monitor thread on test exit

jenshnielsen:
The monitor tests are somewhat flaky on windows failing with a not able to bind to socket error. One reason could be that the monitor is shutdown properly in other tests. This adds a auto fixture  to stop the monitor thread

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,6 +4,7 @@ import copy
 import gc
 import os
 import sys
+import threading
 from collections.abc import Generator
 from pathlib import Path
 from typing import TYPE_CHECKING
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -98,14 +100,17 @@ def reset_state_on_exit() -> Generator[None, None, None]:
     Fixture to clean any shared state on exit
 
     Currently this resets the config to the default config,
-    closes the default station and closes all instruments.
+    closes the default station shutdown the monitor and closes all instruments.
```
jenshnielsen, line 103:
```suggestion
    closes the default station, shutdown the monitor, and closes all instruments.
```

astafan8 (approved):

PR #5655, opened 2024-01-17
Remove deprecated instrument drivers and replace remaining use of deprecate with typing_extensions.deprecated

jenshnielsen:
- [ ] News fragment


src/qcodes/__init__.py
```diff
@@ -91,9 +93,9 @@
     )
 
 
-@deprecate(
-    reason="tests are no longer shipped as part of QCoDeS",
-    alternative="Clone git repo to matching tag and run `pytest tests` from the root of the repo.",
+@deprecated(
+    "tests are no longer shipped as part of QCoDeS Clone git repo to matching tag and run `pytest tests` from the root of the repo.",
```
astafan8, line 97:
```suggestion
    "tests are no longer shipped as part of QCoDeS. Clone git repo to matching tag and run `pytest tests` from the root of the repo.",
```

astafan8 (approved):

PR #5690, opened 2024-01-24
remove unused py package from requirements

jenshnielsen:
This fixes the one remaining vulnerability in https://github.com/microsoft/Qcodes/security/code-scanning/52


jenshnielsen:
This was a transitive dependency of pytest but is no longer in any version that we support

trevormorgan (approved):

PR #5721, opened 2024-02-03
Run pyright with standard checks

jenshnielsen:
This basically means checking that overrides are compatible with the base class

- [x] Merge https://github.com/microsoft/Qcodes/pull/5659 before this one
- [x] Handle TODOs (remaining todos handled via https://github.com/microsoft/Qcodes/pull/5724)
- [x] Are we ok with the breaking changes
- [x] Document the breaking changes

Breaking changes

* The first argument to NumpyJSONEncoder.default is not o not obj. I don't expect this to be called directly. Only public so it can be passed to a json encoder. 
* Unused arguments to QDac connect_messsage have changed. Not an issue since they are unused.
* Keysight N52XX and Infiniium Unused arguments to stub setpoints and unit methods have changed from val to setpoints / unit. Not an issue since these are unused stubs. 
* Alazar DemodulationAcquisitionController handle_data argument changed from data to buffer. Probably unlikely to be called as a kwarg.
* ~InstrumentModule/Channel obj -> value in Sequence/MutableSequence methods. IMHO unlikely to be called by arg since these are simple single argument methods.~ Left this one out for now and ignored the warnings 
* Private: but could still be used in other drivers parameter._wrap_set and _wrap_get no longer takes an argument.

### DIFF
--- a/src/qcodes/instrument/channel.py
+++ b/src/qcodes/instrument/channel.py
@@ -301,8 +301,13 @@ class ChannelTuple(MetadatableWithName, Sequence[InstrumentModuleType]):
         name_parts.append(self.short_name)
         return name_parts
 
-    def index(
-        self, obj: InstrumentModuleType, start: int = 0, stop: int = sys.maxsize
+    # the parameter obj should be called value but that would
+    # be an incompatible change
+    def index(  #  pyright: ignore[reportIncompatibleMethodOverride]
+        self,
+        obj: InstrumentModuleType,
+        start: int = 0,
+        stop: int = sys.maxsize,
     ) -> int:
         """
         Return the index of the given object
@@ -314,7 +319,9 @@ class ChannelTuple(MetadatableWithName, Sequence[InstrumentModuleType]):
         """
         return self._channels.index(obj, start, stop)
 
-    def count(self, obj: InstrumentModuleType) -> int:
+    def count(  #  pyright: ignore[reportIncompatibleMethodOverride]
+        self, obj: InstrumentModuleType
+    ) -> int:
         """Returns number of instances of the given object in the list
 
         Args:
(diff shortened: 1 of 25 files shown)


src/qcodes/instrument/channel.py
```diff
@@ -312,15 +312,15 @@ def index(
...
         """
-        return self._channels.index(obj, start, stop)
+        return self._channels.index(value, start, stop)
 
-    def count(self, obj: InstrumentModuleType) -> int:
+    def count(self, value: InstrumentModuleType) -> int:
         """Returns number of instances of the given object in the list
 
         Args:
             obj: The object to find in the ChannelTuple.
```
jenshnielsen, line 321: If we do this remember to check docstrings

src/qcodes/parameters/parameter.py
```diff
@@ -273,14 +278,15 @@ def _set_manual_parameter(x: ParamRawDataType) -> ParamRawDataType:
                     )
 
                 exec_str_ask = getattr(instrument, "ask", None) if instrument else None
-
-                self.get_raw = Command(
+                # TODO this should be a method like above
```
astafan8, line 281:
```suggestion
                # TODO this should be a MethodType like for _get_manual_parameter
```

src/qcodes/parameters/parameter.py
```diff
@@ -302,11 +308,12 @@ def _set_manual_parameter(x: ParamRawDataType) -> ParamRawDataType:
                 exec_str_write = (
                     getattr(instrument, "write", None) if instrument else None
                 )
-                self.set_raw = Command(
+                # TODO this should also be a method
```
astafan8, line 311:
```suggestion
                # TODO this should also be a MethodType like for _set_manual_parameter
```

astafan8 (approved):
I approve.

assuming the "Private: but could still be used in other drivers parameter._wrap_set and _wrap_get no longer takes an argument." part is not breaking anything at least in qcodes_contrib_drivers.

jenshnielsen:
> assuming the "Private: but could still be used in other drivers parameter._wrap_set and _wrap_get no longer takes an argument." part is not breaking anything at least in qcodes_contrib_drivers.

It does not. The only example I found where this was used incorrectly was in the mercury driver

trevormorgan (approved):

PR #5754, opened 2024-02-21
use uv to generate min requirements

jenshnielsen:
This removes the dependency on requirements builder

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -71,12 +71,9 @@ jobs:
       run: python -m pip install --upgrade pip setuptools wheel
       shell: bash
     - name: install qcodes with minimum requirements
-#     Use fork of requirements-builder until official version
-#     has support for pyproject.toml (pep621)
       run: |
-        pip install git+https://github.com/jenshnielsen/requirements-builder.git
-        pip install versioningit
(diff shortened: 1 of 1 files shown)


.github/workflows/pytest.yaml
```diff
@@ -71,12 +71,9 @@ jobs:
       run: python -m pip install --upgrade pip setuptools wheel
       shell: bash
     - name: install qcodes with minimum requirements
-#     Use fork of requirements-builder until official version
-#     has support for pyproject.toml (pep621)
       run: |
-        pip install git+https://github.com/jenshnielsen/requirements-builder.git
-        pip install versioningit
-        requirements-builder -l min -e test setup.py -o min-requirements.txt
```
jenshnielsen, line 79: Does not seem to work correct at the moment

trevormorgan (approved):

.github/workflows/pytest.yaml
trevormorgan: hmm looks exactly how it's described in the docs. could it be the reference to test dependencies in requirments.in?

jenshnielsen: @trevormorgan This should now be working by compiling the dependencies directly from pyproject.toml

trevormorgan (approved):

PR #5757, opened 2024-02-22
regenerate requirements with uv as an exact lockfile

jenshnielsen:
This means replacing ~= with == 

This will resolve in many more dependabot pull requests since all patch versions will now require a bump. 
On the other hand issues such as https://github.com/microsoft/Qcodes/security/code-scanning/52 will be
fixed automatically as we bump the dependencies


astafan8 (approved):
awesome! and uv is really coool!

PR #5873, opened 2024-03-20
Improvements by Equal1 (station indexed param names; Rohde&Schwarz, Stahl, Stanford Research drivers)

hp-peti:
* Extended JSON schema and parser for Station files to allow indexing in instrument parameter name references
  * rationale: some drivers (ahem Zurich Instruments) only expose parameters through indexed access
* Additional parameters for the Rohde & Schwarz SGS100A to support some expansion modules (pulse mod, trigger)
* Fixed some issues with and improved the Stahl BS power source driver
  * regular expression did not cover all possible floating point values coming from the instrument 
  * feature to access it over TCP/IP connection (e.g the serial port can be socat'd over a port -- solution not included)
* a completely new Stanford Research SR570 driver based on SR560, but this one actually talks to the instrument

### DIFF
--- a/src/qcodes/dist/schemas/station-template.schema.json
+++ b/src/qcodes/dist/schemas/station-template.schema.json
@@ -59,7 +59,7 @@
                                     "properties": {
                                         "source": {
                                             "type": "string",
-                                            "pattern": "^(\\w+)(\\.\\w+)*$"
+                                            "pattern": "^(\\w+(\\[\\d+\\])*)(\\.\\w+(\\[\\d+\\])*)*$"
                                         },
                                         "label": {
                                             "type": "string"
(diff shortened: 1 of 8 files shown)


jenshnielsen:
@hp-peti Thanks for the contributions. I would like to request that you submit the new driver into the qcodes_contrib_drivers package here https://github.com/QCoDeS/Qcodes_contrib_drivers  As none of the qcodes maintainers have acces to this instrument we prefer to add the instrument to that package as we cannot test and verify the driver

For the other contributions we would prefer if you can split them up into a pr for each of the drivers and one for the JSON/attribute fixes. This makes it easier to review and the changelog easier to follow both for users and maintainers. Let me know if this is possible for you.

hp-peti:
PR split in the following:
1. #5887
2. #5888 
3. #5889 
4. #5890
Closing PR.

PR #5887, opened 2024-03-25
Simple indexed attribute access in station files

hp-peti:
Some drivers (see zurich-instruments) only expose parameters through an interface which needs indexing in addition to attribute access. This patch enables access to those parameters in the station yaml file.

### DIFF
--- a/src/qcodes/station.py
+++ b/src/qcodes/station.py
@@ -48,6 +48,8 @@ from qcodes.parameters import (
 from qcodes.utils import (
     DelegateAttributes,
     checked_getattr,
+    checked_getattr_indexed,
+    getattr_indexed,
     get_qcodes_path,
     get_qcodes_user_path,
 )
@@ -595,7 +597,7 @@ class Station(Metadatable, DelegateAttributes):
             level = levels[0]
             try:
                 for level in levels:
-                    instrument = checked_getattr(
+                    instrument = checked_getattr_indexed(
                         instrument, level, (InstrumentBase, ChannelTuple)
                     )
             except TypeError:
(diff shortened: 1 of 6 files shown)


src/qcodes/station.py
```diff
@@ -595,7 +597,7 @@ def resolve_instrument_identifier(
             level = levels[0]
             try:
                 for level in levels:
-                    instrument = checked_getattr(
+                    instrument = checked_getattr_indexed(
```
jenshnielsen, line 600: Could we extend the docstring above with an example of what this adds

I guess something like ?

dac.channels[0]

jenshnielsen:
Would be nice with an example of this in the station notebook here https://github.com/microsoft/Qcodes/blob/main/docs/examples/Station.ipynb  but I think that is optional

jenshnielsen (approved):

src/qcodes/station.py
hp-peti: on it

jenshnielsen:
@hp-peti BTW unless there are conflicts you do not need to worry about merging main into the feature branch. Github actions will take care of making sure that it is tested against main before merging

hp-peti:
need workflow approval

> Would be nice with an example of this in the station notebook here https://github.com/microsoft/Qcodes/blob/main/docs/examples/Station.ipynb but I think that is optional.
I don't have the courage to modify that notebook; the example station.yaml is an image file i don't want to mess with.
> 
> Other than that the only remaining issue is IMHO the precommit hooks
auto organized utils/__init__.py to fix issue

hp-peti:
need another workflow approval. 

not sure what's happening on the merge queue. 
@jenshnielsen let me know there is anything i need to do
thanks in adv

hp-peti:
pinging again, i need workflow approval and please put back in merge queue. thanks in advance

hp-peti:
thanks everyone

PR #5889, opened 2024-03-25
Rohde & Schwarz SGS100A Trigger and Pulse Modulator parameters

hp-peti:
Additional parameters for the Pulse Modulation module of the R&S SG100A Signal Generator

Based on the manual.

### DIFF
--- a/docs/changes/newsfragments/5888.improved_driver
+++ b/docs/changes/newsfragments/5888.improved_driver
@@ -1,0 +1,2 @@
+The driver for ``Rohde&Schwarz SG100A`` has been extendend with parameters
+controlling the optional pulse modulation and trigger modules.


src/qcodes/instrument_drivers/rohde_schwarz/SGS100A.py
```diff
@@ -139,7 +138,90 @@ def __init__(self, name: str, address: str, **kwargs: Any) -> None:
...
+        self.add_parameter('trigger_connector_mode',
+                           label='Trigger Connector Mode',
+                           get_cmd='CONN:TRIG:OMOD?',
+                           set_cmd='CONN:TRIG:OMOD {}',
+                           vals=vals.Enum('SVAL',   # SVALid - Signal valid
+                                          'SNVAL',  # SNValid - Signal not valid
+                                          'PVO',    # PVOut - Pulse video out (K22 Only)
+                                          'PET',    # PETrigger - Pulse mod ext trigger - PETrigger (K22 Only)
+                                          'PEMS',   # PEMSource - Pulse mode ext source (K22 Only)
+                                          'sval', 'snval', 'pvo', 'pet', 'pems'))
```
jenshnielsen, line 151: Is the reason to have lower case here to allow the user to both input lower and upper case? If so I think it would be cleaner to add a `set_parser=lambda x: x.upper` to the parameter. That way the user can alway use the lower case but there is only one format internally

hp-peti: i didn't know that would work with vals.Enum. also i wanted to be able to roundtrip correctly. i don't have access to the isntrument at this point

hp-peti: the instrument accepts both lowercase and uppercase and I didn't want to impose either on the user. set_parser is applied after the vals.Enum so it's not good for what I wanted.

hp-peti: it was also unclear whether the instrument would always return the uppercase version but yes i could add a get_parser to ensure that? i dont' know

hp-peti: the set_parser is applied after the validator so it won't work for what i wanted; also i don't want all crazy combinations of lowercase and uppercase to work

jenshnielsen:
@hp-peti Could you please add a newsfragment descriping this change and documented in the pull request template text

hp-peti:
added missing news fragment. pleas approve 6 workflows
thanks in advance

jenshnielsen (approved):

PR #5917, opened 2024-04-02
Add tch ruff lints.

jenshnielsen:
This will hopefully significantly reduce the import time since we avoid imports that are not required.

Draft since this likely will require some manual inspection

### DIFF
--- a/src/qcodes/instrument_drivers/Keithley/Keithley_2450.py
+++ b/src/qcodes/instrument_drivers/Keithley/Keithley_2450.py
@@ -1,5 +1,4 @@
-from types import TracebackType
-from typing import Any, ClassVar, Optional, Union, cast
+from typing import TYPE_CHECKING, Any, ClassVar, Optional, Union, cast
 
 import numpy as np
 from typing_extensions import TypedDict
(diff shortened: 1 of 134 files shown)


astafan8 (approved):
went thorugh the files, not suuuper deep, but i think the changes are ok.

src/qcodes/instrument_drivers/Keithley/Keithley_2450.py
```diff
@@ -132,7 +134,7 @@ def __exit__(
         self,
         exception_type: Optional[type[BaseException]],
         value: Optional[BaseException],
-        traceback: Optional[TracebackType],
+        traceback: Optional["TracebackType"],
```
astafan8, line 137: for the quotes usage - with move to python 3.10+ those will be converted to not-strings, right?

jenshnielsen: @astafan8 No, once we move to 3.10+ we can use `|` for union types without ```from __future__ import annotations` but the forward references has not stabilized yet in any version and its not clear that it will and when that might be. We could add a `from __future__ import annotations` to all these files instead if you prefer?

jenshnielsen: Does look like [pep649](https://peps.python.org/pep-0649/) has been accepted for 3.13 so this will probably eventually become redundant

PR #5957, opened 2024-04-14
Keep pyright up to date as a pip package

jenshnielsen:
This enables keeping pyright up to date as a pip package making use of the new PATH feature in the pyright action.

This uses the unofficial pip package. As an alternative this can also be done directly using the official npm package https://github.com/microsoft/Qcodes/pull/5956


jenshnielsen:
Abandoned since we merged #5958

PR #5958, opened 2024-04-14
Keep Pyright up to date as a npm package.

jenshnielsen:
Rather than installing the package this parses the version out of a packages.json file making the setup simpler

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -81,9 +81,19 @@ jobs:
       run: |
         pip install .[test] -c requirements.txt
       if: ${{ !matrix.min-version }}
+    - name: Install node.js
+      uses: actions/setup-node@v4
+      with:
+        node-version: '20.x'
+    - name: install pyright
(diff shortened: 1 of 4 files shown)


.github/workflows/pytest.yaml
```diff
@@ -81,9 +81,19 @@ jobs:
...
         pip install .[test] -c requirements.txt
       if: ${{ !matrix.min-version }}
+    - name: Install node.js
+      uses: actions/setup-node@v4
+      with:
+        node-version: '20.x'
+    - name: install pyright
+      run: |
+        npm install
+        echo "$PWD/node_modules/.bin" >> $GITHUB_PATH
```
jenshnielsen, line 91: Document that npm does not support installing packages globally from a package.json file which is why we do this path modification

jenshnielsen: This also needs to be updated to support windows

astafan8 (approved):

PR #5967, opened 2024-04-16
Add missing type annotations

jenshnielsen:
that mypy requires when run as a deamon via dmypy

### DIFF
--- a/src/qcodes/utils/delaykeyboardinterrupt.py
+++ b/src/qcodes/utils/delaykeyboardinterrupt.py
@@ -2,7 +2,7 @@ import logging
 import signal
 import threading
 from types import FrameType, TracebackType
-from typing import Optional, cast
+from typing import Any, Callable, Optional, Union, cast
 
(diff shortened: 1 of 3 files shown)


src/qcodes/utils/delaykeyboardinterrupt.py
```diff
@@ -16,8 +16,9 @@ class DelayedKeyboardInterrupt:
 
     Inspired by https://stackoverflow.com/questions/842557/how-to-prevent-a-block-of-code-from-being-interrupted-by-keyboardinterrupt-in-py
     """
-    signal_received = None
-    old_handler = None
+
+    signal_received: Optional[tuple[int, FrameType | None]] = None
+    old_handler: Optional[signal._HANDLER] = None
```
jenshnielsen, line 21: I don't really like using a private type here but I don't know what the alternative is (use Any?)

astafan8 (approved):

PR #5991, opened 2024-04-22
Deprecate opencensus integration

jenshnielsen:
And remove default install of opencensus and opencensus azure integration.

Note that this does not yet remove opencensus from requirements.txt. This will happen once a release of qcodes with these changes is out. This is due to the "circular" dependencies between qcodes[test], qcodes-loop, and qcodes. 

TODO

- [ ] Breaking change notice

### DIFF
--- a/src/qcodes/logger/logger.py
+++ b/src/qcodes/logger/logger.py
@@ -33,6 +33,7 @@ from qcodes.utils import (
 )
 
 AzureLogHandler = Any
+Envelope = Any
 
 log: logging.Logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


src/qcodes/logger/logger.py
```diff
@@ -201,11 +198,17 @@ def flush_telemetry_traces() -> None:
         telemetry_handler.flush()
 
 
+@deprecated(
+    "OpenCensus integration is deprecated. Please use your own telemetry integration as needed",
```
astafan8, line 202:
```suggestion
    "OpenCensus integration is deprecated. Please use your own telemetry integration as needed, we recommend OpenTelemetry",
```

astafan8 (approved):

PR #5998, opened 2024-04-23
DelegateParameter infer methods to find root parameters, instruments, and channels

samantha-ho:
This PR adds methods to recursively search a chain of DelegateParameters with the aim of finding the root parameter. This also allows users to infer the InstrumentModule or Instrument that a chain of DelegateParameters stems from. 

These methods also allow the user to customize which attributes of a Parameter link to other parameters.

### DIFF
--- a/src/qcodes/parameters/infer.py
+++ b/src/qcodes/parameters/infer.py
@@ -1,0 +1,109 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, ClassVar
+
+from qcodes.instrument import Instrument, InstrumentBase, InstrumentChannel
+from qcodes.instrument.parameter import DelegateParameter, Parameter
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable, Sequence
+
+
+class InferError(AttributeError): ...
+
+
+class InferAttrs:
+    """Holds a global set of attribute name that will be inferred"""
+
+    _known_attrs: ClassVar[set[str]] = set()
+
+    @classmethod
+    def add_attr(cls, attr: str) -> None:
+        cls._known_attrs.add(attr)
+
+    @classmethod
+    def known_attrs(cls) -> tuple[str, ...]:
+        return tuple(cls._known_attrs)
+
+    @classmethod
+    def discard_attr(cls, attr: str) -> None:
+        cls._known_attrs.discard(attr)
+
+    @classmethod
+    def clear_attrs(cls) -> None:
+        cls._known_attrs = set()
+
+
+def get_root_param(
+    param: Parameter | DelegateParameter | None,
+    parent_param: Parameter | None = None,
+    alt_source_attrs: Sequence[str] | None = None,
+) -> Parameter:
+    """Return the root parameter in a chain of DelegateParameters or other linking Parameters
+
+    This method recursively searches on the initial parameter.
+    - If the parameter is a DelegateParameter, it returns the .source.
+    - If the parameter is not a DelegateParameter, but has an attribute in
+    either alt_source_attrs or the InferAttrs class which is a parameter,
+    then it returns that parameter
+    - If the parameter is None, because the previous DelegateParameter did not have a source
+    it raises an InferError
+
+
+    """
+    parent_param = param if parent_param is None else parent_param
+    if alt_source_attrs is None:
+        alt_source_attrs_set: Iterable[str] = InferAttrs.known_attrs()
+    else:
+        alt_source_attrs_set = set.union(
+            set(alt_source_attrs), set(InferAttrs.known_attrs())
+        )
+
+    if param is None:
+        raise InferError(f"Parameter {parent_param} is not attached to a source")
+    if isinstance(param, DelegateParameter):
+        return get_root_param(param.source, parent_param)
+    for alt_source_attr in alt_source_attrs_set:
+        alt_source = getattr(param, alt_source_attr, None)
+        if alt_source is not None and isinstance(alt_source, Parameter):
+            return get_root_param(
+                alt_source, parent_param=parent_param, alt_source_attrs=alt_source_attrs
+            )
+    return param
+
+
+def infer_instrument(
+    param: Parameter,
+    alt_source_attrs: Sequence[str] | None = None,
+) -> InstrumentBase:
+    """Find the instrument that owns a parameter or delegate parameter."""
+    root_param = get_root_param(param, alt_source_attrs=alt_source_attrs)
+    instrument = get_instrument_from_param(root_param)
+    if isinstance(instrument, InstrumentChannel):
+        return instrument.root_instrument
+    elif isinstance(instrument, Instrument):
+        return instrument
+
+    raise InferError(f"Could not determine source instrument for parameter {param}")
+
+
+def infer_channel(
+    param: Parameter,
+    alt_source_attrs: Sequence[str] | None = None,
+) -> InstrumentChannel:
+    """Find the instrument module that owns a parameter."""
+    root_param = get_root_param(param, alt_source_attrs=alt_source_attrs)
+    channel = get_instrument_from_param(root_param)
+    if isinstance(channel, InstrumentChannel):
+        return channel
+    raise InferError(
+        f"Could not determine a root instrument channel for parameter {param}"
+    )
+
+
+def get_instrument_from_param(
+    param: Parameter,
+) -> InstrumentBase:
+    if param.instrument is not None:
+        return param.instrument
+    raise InferError(f"Parameter {param} has no instrument")
--- a/src/qcodes/parameters/infer.py
+++ b/src/qcodes/parameters/infer.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, ClassVar
 
-from qcodes.instrument import Instrument, InstrumentBase, InstrumentChannel
+from qcodes.instrument import Instrument, InstrumentBase, InstrumentModule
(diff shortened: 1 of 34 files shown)


samantha-ho:
While writing tests, I realized there is a possibility for a DelegateParameter "loop" which would cause both `get_root_parameter` and `get_parameter_chain` to recurse infinitely.

There are certainly ways I could stop this, but I'm not sure the case is likely enough to bother with at this time. I invite other perspectives on the matter.

astafan8:
> There are certainly ways I could stop this, but I'm not sure the case is likely enough to bother with at this time. I invite other perspectives on the matter.

I think it’d be useful to prevent the loops or at least error out when loops are discovered because that will be a welcoming and helpful user experience, so I’d encourage to do it

src/qcodes/parameters/infer.py
```diff
@@ -0,0 +1,144 @@
...
+    def known_attrs(cls) -> tuple[str, ...]:
+        return tuple(cls._known_attrs)
+
+    @classmethod
+    def discard_attr(cls, attr: str) -> None:
+        cls._known_attrs.discard(attr)
+
+    @classmethod
+    def clear_attrs(cls) -> None:
+        cls._known_attrs = set()
```
astafan8, line 39: Suggesting less verbose names because these methods are called on a class that’s called attrs so it’s clear what they add/discard/etc
```suggestion
class InferAttrs:
    """Holds a global set of attribute name that will be inferred"""

    _known_attrs: ClassVar[set[str]] = set()

    @classmethod
    def add(cls, attrs: str | Iterable[str]) -> None:
        if isinstance(attrs, str):
            attrs = (attrs,)
        cls._known_attrs.update(set(attrs))

    @classmethod
    def known_attrs(cls) -> tuple[str, ...]:
        return tuple(cls._known_attrs)

    @classmethod
    def discard(cls, attr: str) -> None:
        cls._known_attrs.discard(attr)

    @classmethod
    def clear(cls) -> None:
        cls._known_attrs = set()
```

src/qcodes/parameters/infer.py
```diff
@@ -0,0 +1,144 @@
...
+    alt_source_attrs_set = _merge_user_and_class_attrs(alt_source_attrs)
+
+    if isinstance(param, DelegateParameter):
+        if param.source is None:
+            raise InferError(f"Parameter {param} is not attached to a source")
+        return get_root_parameter(param.source)
+
+    for alt_source_attr in alt_source_attrs_set:
+        alt_source = getattr(param, alt_source_attr, DOES_NOT_EXIST)
+        if alt_source is None:
```
astafan8, line 56: Do I understand correctly that the DOES_NOT_EXIST case is simply ignored? E.g. we don’t tell the user that al_source_attr doesn’t exist or something else?

src/qcodes/parameters/infer.py
```diff
@@ -0,0 +1,144 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+DOES_NOT_EXIST = "Does not exist"
+
+
+class InferError(AttributeError): ...
+
+
+class InferAttrs:
```
astafan8, line 18: Should we have any concerns for the fact that this will hold global state within a given python session? Is it really convenient and helpful to have this state or would it be opaque to users? And another concern about this state is - do we want to encourage a pattern where some attrs are added in the first place in the code, and then some further place in the code relies on the fact that the previous attrs were already added, rather than making it explicit which attrs it (this second place in the code) relies on?

src/qcodes/parameters/infer.py
```diff
@@ -0,0 +1,144 @@
+from __future__ import annotations
+
+from collections.abc import Sequence
+from typing import TYPE_CHECKING, ClassVar
+
+from qcodes.instrument import Instrument, InstrumentBase, InstrumentModule
```
jenshnielsen, line 6: The qcodes.instrument module depends on the qcodes on the qcodes.parameters module. 

Thinking about this more generally it feels like these convenience functions really should be methods on the parameter and or instrument classes but I realize that it may take some time to iron that out. Therefor I think I would suggest to add this to the qcodes.extensions module for now.

src/qcodes/parameters/infer.py
```diff
@@ -0,0 +1,144 @@
...
+from qcodes.instrument.parameter import DelegateParameter, Parameter
```
jenshnielsen, line 7: Parameters should be imported from qcodes.parameters. This module is a deprecated alias.

src/qcodes/parameters/infer.py
```diff
@@ -0,0 +1,144 @@
...
+    instrument = get_instrument_from_param(root_param)
+    if isinstance(instrument, InstrumentModule):
+        return instrument.root_instrument
+    elif isinstance(instrument, Instrument):
+        return instrument
+
+    raise InferError(f"Could not determine source instrument for parameter {param}")
+
+
+def infer_channel(
```
jenshnielsen, line 80:
```suggestion
def infer_instrument_module(
```

src/qcodes/parameters/infer.py
jenshnielsen: It feels to me like all state like this should live in the station and this should be a private attribute of the station that the functions can dive into

src/qcodes/parameters/infer.py
```diff
@@ -0,0 +1,144 @@
...
+        f"Could not determine a root instrument channel for parameter {param}"
+    )
+
+
+def get_instrument_from_param(
+    param: Parameter,
+) -> InstrumentBase:
+    if param.instrument is not None:
+        return param.instrument
+    raise InferError(f"Parameter {param} has no instrument")
```
jenshnielsen, line 99: What is the purpose of this thing. It seems to me that this just converts an Optional into an exception that you then don't get any help from the type-system to remember to handle.

src/qcodes/parameters/infer.py
```diff
@@ -0,0 +1,144 @@
...
+        elif isinstance(alt_source, Parameter):
+            return get_root_parameter(alt_source, alt_source_attrs=alt_source_attrs)
+    return param
+
+
+def infer_instrument(
+    param: Parameter,
+    alt_source_attrs: Sequence[str] | None = None,
+) -> InstrumentBase:
+    """Find the instrument that owns a parameter or delegate parameter."""
```
jenshnielsen, line 69: For all these methods it would be good to document what they raise and when

jenshnielsen (commented):
Left some comments inline

src/qcodes/parameters/infer.py
samantha-ho: Correct. DOES_NOT_EXIST distinguishes between two cases
1. The linking attribute is there, but no linking parameter has been added. This should return None, resulting in stopping the chain or an error
2. The linking attribute didn't exist in the first place. Which means this *isn't* a linking parameter of a type recognized by the inference.

src/qcodes/parameters/infer.py
samantha-ho: What I was thinking here is that the number of non-Delegate linking parameters is likely to be small, but may be widely used. In particular, dyno Testpoints use `external_parameter` as the linking attribute. (for reasons I now regret), and the qcodes ScaledParameter uses `wrapped_parameter`. 

I do not want the experimenters to have to think about "Oh, does this use a dyno Testpoint or just a DelegateParameter or something else." I want inference to be completely transparent.

If we make the entries in the InferAttrs class Instance variables instead of class variables, then I need to start carrying around references to the specific instance in the Station, and I thought that felt clunky too. This seemed like a fairly harmless balance.

src/qcodes/parameters/infer.py
samantha-ho: Ooops. That's what I get for using Ruff's auto-import.

src/qcodes/parameters/infer.py
samantha-ho: I'm ok with this, but then we should also alias `infer_channel`, as that's what they're named now in `dyno` and `raster`

src/qcodes/parameters/infer.py
samantha-ho: Yes. This exists because I got very tired of handling the None case every time I wanted to use infer methods. In almost all cases where I have used infer or similar methods, I *know* that the parameter is properly attached to an instrument, so handling it at that level seems unnecessarily burdensome.

samantha-ho:
> > There are certainly ways I could stop this, but I'm not sure the case is likely enough to bother with at this time. I invite other perspectives on the matter.
> 
> I think it’d be useful to prevent the loops or at least error out when loops are discovered because that will be a welcoming and helpful user experience, so I’d encourage to do it

src/qcodes/parameters/infer.py
```diff
@@ -0,0 +1,144 @@
...
+    root_param = get_root_parameter(param, alt_source_attrs=alt_source_attrs)
+    instrument = get_instrument_from_param(root_param)
+    if isinstance(instrument, InstrumentModule):
+        return instrument.root_instrument
+    elif isinstance(instrument, Instrument):
+        return instrument
+
+    raise InferError(f"Could not determine source instrument for parameter {param}")
```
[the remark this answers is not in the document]
samantha-ho: Ah. Yes. This returns `root_instrument` by default, which I considered using, but I think I ran into an unhandled issue with it. I will try to remember what the issue was.

src/qcodes/parameters/infer.py
astafan8: ok, i see. Let's then live withthis for now and address this "attrbiutes state" and hopefully the "linking attributes" in the mid-/long-term.

PR #6012, opened 2024-04-27
Use typeddicts for better kwargs typing in Instrument classes

jenshnielsen:
This means that the **kwargs of a Instrument class will exactly match the super class while transparently keeping the ability to add another kwarg to the instrument class when required.

The one downside of this is that it is not possible to override the default value of a kwarg in a subclass since that results in an error about overlapping kwargs. We use this pattern fairly frequently in visa instruments to change the default value of the terminator. In this code we therefor do not add the terminator kwarg to the TypedDict

- [x] Update docs on writing drivers to reflect this
- [x] Update docs to reflect use of attribute for terminator and timeout
- [x] Add docstrings to typed dict classes
- [x] Migrate drivers until and including instruments vendors starting with K
- [x] Changelog. Document that all these arguments are now kw only
- [x] Verify that the change to Instrument is acceptable. (label and metadata is kw only)

In a separate pr
- [ ] Migrate remaining visa drivers in qcodes to the new format


src/qcodes/instrument/visa.py
```diff
@@ -45,6 +47,18 @@ def _close_visa_handle(
         # the resource is already closed
         pass
 
+
+class VisaInstrumentNoTerminatorKWArgs(InstrumentBaseKWArgs):
+    timeout: float
+    device_clear: bool
+    visalib: str | None
+    pyvisa_sim_file: str | None
```
jenshnielsen, line 55: These would need to use not required

src/qcodes/instrument/visa.py
```diff
@@ -45,6 +47,18 @@ def _close_visa_handle(
...
+
+
+class VisaInstrumentKWArgs(VisaInstrumentNoTerminatorKWArgs):
+    terminator: str | None
```
jenshnielsen, line 59: same

src/qcodes/instrument/instrument.py
```diff
@@ -56,16 +55,11 @@ class Instrument(InstrumentBase, metaclass=InstrumentMeta):
     _type = None
     _instances: weakref.WeakSet[Instrument] = weakref.WeakSet()
 
-    def __init__(
-        self,
-        name: str,
-        metadata: Mapping[Any, Any] | None = None,
-        label: str | None = None,
-    ) -> None:
+    def __init__(self, name: str, **kwargs: Unpack[InstrumentBaseKWArgs]) -> None:
```
jenshnielsen, line 58: This is technically an api break since name and label now cannot be passed as positional arguments. Perhaps we should revert thsi

astafan8: but do you see label/metadata passed positionally inside qcodes or qcodes contrib drivers? if not, then i'd say it's fine (we are still not v1.0, even though, yes, it's good to never break public api)

src/qcodes/instrument/visa.py
```diff
@@ -45,6 +47,51 @@ def _close_visa_handle(
         # the resource is already closed
         pass
 
+
+class VisaInstrumentKWArgs(TypedDict):
+    """
```
astafan8, line 52: and we can't or we should'nt inherit from InstrumentBaseKWArgs?

jenshnielsen: Minimal example of this can be seen here https://github.com/jenshnielsen/debugproject/tree/error_on_typed_dict_inherit

src/qcodes/instrument/instrument.py
jenshnielsen: Found one example in contib drivers. I think its fair to fix that and do this https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/330

jenshnielsen: Will fix, merge and do a new contrib driver release and close this

jenshnielsen: tagged qcodes_contrib_drivers v0.22.0 with the fix

PR #6032, opened 2024-05-02
[Wip] experiment with model for metadata

jenshnielsen:
This is an experiment to allow the user to enforce better types for a metadatables metadata and snapshot. In the first instance this is applied to parameter.

There are a quite a few things left to do but mainly validation on addition of data is missing.


einsmein:
#6119 continues the work started in this PR

PR #6073, opened 2024-05-15
Accept sequences of values for setting MultiChannelInstrumentParameter

thangleiter:
As of now, the `MultiChannelInstrumentParameter` is inconsistent; its setter does not accept the values returned by its getter as the first accepts a scalar `ParamRawDataType`, but the last returns a tuple thereof. This is particularly problematic as it breaks the `set_to` context manager.

This proposed change modifies the setter to allow both the current scalar behavior as well as a sequence, in which case it is zipped together with `self._channels`.

### DIFF
--- a/src/qcodes/parameters/multi_channel_instrument_parameter.py
+++ b/src/qcodes/parameters/multi_channel_instrument_parameter.py
@@ -61,7 +61,11 @@ class MultiChannelInstrumentParameter(MultiParameter, Generic[InstrumentModuleTy
                 getattr(chan, self._param_name).set(value)
         except Exception as err:
             try:
-                for chan, val in zip(self._channels, value, strict=True):
+                # Catch wrong length of value before any setting is done
+                value_list = list(value)
+                if len(value_list) != len(self._channels):
+                    raise ValueError
+                for chan, val in zip(self._channels, value_list):
                     getattr(chan, self._param_name).set(val)
             except (TypeError, ValueError):
                 note = (
@@ -72,7 +76,7 @@ class MultiChannelInstrumentParameter(MultiParameter, Generic[InstrumentModuleTy
                     err.add_note(note)
                 else:
                     _LOG.error(note)
-                raise
+                raise err from None
 
     @property
     def full_names(self) -> tuple[str, ...]:
--- a/src/qcodes/parameters/multi_channel_instrument_parameter.py
+++ b/src/qcodes/parameters/multi_channel_instrument_parameter.py
@@ -49,10 +49,10 @@ class MultiChannelInstrumentParameter(MultiParameter, Generic[InstrumentModuleTy
 
     def set_raw(self, value: ParamRawDataType | Sequence[ParamRawDataType]) -> None:
         """
-        Set all parameters to this value.
+        Set all parameters to this/these value(s).
 
         Args:
-            value: The value to set to. The type is given by the
+            value: The value(s) to set to. The type is given by the
                 underlying parameter.
         """
         try:
--- a/src/qcodes/parameters/multi_channel_instrument_parameter.py
+++ b/src/qcodes/parameters/multi_channel_instrument_parameter.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import sys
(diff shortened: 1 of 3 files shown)


thangleiter:
@microsoft-github-policy-service agree company="RWTH Aachen University"

src/qcodes/parameters/multi_channel_instrument_parameter.py
```diff
@@ -45,16 +47,31 @@ def get_raw(self) -> tuple[ParamRawDataType, ...]:
...
         """
-        for chan in self._channels:
-            getattr(chan, self._param_name).set(value)
+        try:
+            for chan in self._channels:
+                getattr(chan, self._param_name).set(value)
+        except Exception as err:
+            try:
+                for chan, val in zip(self._channels, value, strict=True):
+                    getattr(chan, self._param_name).set(val)
```
thangleiter, line 64: Might actually make more sense to check if value complies with zip(..., strict=True) before setting the parameters.

src/qcodes/parameters/multi_channel_instrument_parameter.py
```diff
@@ -12,6 +13,7 @@
     from .parameter_base import ParamRawDataType
 
 InstrumentModuleType = TypeVar("InstrumentModuleType", bound="InstrumentModule")
+LOG = logging.getLogger(__name__)
```
jenshnielsen, line 16: Since writing many of the QCODES modules we have learned that this probably should be private

```suggestion
_LOG = logging.getLogger(__name__)
```

src/qcodes/parameters/multi_channel_instrument_parameter.py
```diff
@@ -45,16 +47,33 @@ def get_raw(self) -> tuple[ParamRawDataType, ...]:
...
+            try:
+                for chan, val in zip(self._channels, value, strict=True):
+                    getattr(chan, self._param_name).set(val)
+            except (TypeError, ValueError):
+                note = (
+                    "Value should either be valid for a single parameter of the channel list "
+                    "or a sequence of valid values of the same length as the list."
+                )
+                try:
+                    err.add_note(note)
```
thangleiter, line 71: @jenshnielsen I don't know why pyright gets offended by this line. It is in a try-except block handling exactly that error pyright tries to warn about.

src/qcodes/parameters/multi_channel_instrument_parameter.py
thangleiter: I just realized that the strict kwarg has only been added in 3.10 (which gives rise to that weirdly misleading pyright error [here](https://github.com/microsoft/Qcodes/actions/runs/9147210941/job/25193293073#step:10:23). On the other hand, 3.9 support is due to be dropped in #5620.

src/qcodes/parameters/multi_channel_instrument_parameter.py
jenshnielsen: The way type checkers work they assume that the code must be valid for that python version. The way to fix that is usually to explicitly check for the version of python. See example here https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks (From the mypy docs but it works the same in pyright)

This also has be benefit that ruff/pyupgrade can automatically remove the work around code block when we drop support that the specific version

src/qcodes/parameters/multi_channel_instrument_parameter.py
```diff
@@ -45,16 +48,35 @@ def get_raw(self) -> tuple[ParamRawDataType, ...]:
...
+            for chan in self._channels:
+                getattr(chan, self._param_name).set(value)
+        except Exception as err:
+            try:
+                # Catch wrong length of value before any setting is done
+                value_list = list(value)
+                if len(value_list) != len(self._channels):
+                    raise ValueError
+                for chan, val in zip(self._channels, value_list):
+                    getattr(chan, self._param_name).set(val)
```
jenshnielsen, line 69: Would it be possible to add a test that covers this behaviour ?

thangleiter: Done.

jenshnielsen (approved):

PR #6074, opened 2024-05-15
Allow parameter contexts for uninitialized parameters

thangleiter:
Currently, using the `set_to` context fails on exit for parameters whose value does not pass the set parser or validator. The only -- but common -- case where this happens are uninitialized parameters, in which case the value is `None`.

This change catches exceptions when setting the original value and then sets the cache value instead. The exception catch is very broad, but as far as I can tell, validators can raise any kind of exception they like. Only doing this if `original_value` is `None` might be a bit safer -- are there other cases?

### DIFF
--- a/src/qcodes/parameters/parameter_base.py
+++ b/src/qcodes/parameters/parameter_base.py
@@ -81,13 +81,12 @@ class _SetParamContext:
                 self._parameter.set(self._original_value)
             except Exception:
                 # Likely an uninitialized Parameter
-                vals = self._parameter._vals
-                set_parser = self._parameter.set_parser
-                self._parameter._vals = []
-                self._parameter.set_parser = None
-                self._parameter.cache.set(self._original_value)
-                self._parameter.set_parser = set_parser
-                self._parameter._vals = vals
+                LOG.info(
+                    "Encountered an exception setting the original value "
+                    "when exiting set_to context of "
+                    f"{self._parameter.full_name}",
+                    exc_info=True,
+                )
 
 
 def invert_val_mapping(val_mapping: Mapping[Any, Any]) -> dict[Any, Any]:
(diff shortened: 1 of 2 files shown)


src/qcodes/parameters/parameter_base.py
```diff
@@ -77,7 +77,17 @@ def __exit__(
...
+                self._parameter.set(self._original_value)
+            except Exception:
+                # Likely an uninitialized Parameter
+                vals = self._parameter._vals
+                set_parser = self._parameter.set_parser
+                self._parameter._vals = []
+                self._parameter.set_parser = None
+                self._parameter.cache.set(self._original_value)
+                self._parameter.set_parser = set_parser
+                self._parameter._vals = vals
```
jenshnielsen, line 90: I actually think that it might be better to just leave the parameter as is if we see a warning. 

I suggest that we log the Exception here with level info 

LOG.info("some message" exc_info=True)

tests/parameter/test_parameter_context_manager.py
```diff
@@ -158,6 +166,13 @@ def test_context(instrument: DummyTrackingInstrument) -> None:
         assert instrument.a.get() == 3
     assert instrument.a.get() == 2
 
+    # The cached value does not pass the validator (since it was not
+    # initialized). Make sure no exception is raised when __exit__ing the
+    # context
+    with instrument.uninitialized_param.set_to(2):
+        assert instrument.uninitialized_param.get() == 2
+    assert instrument.uninitialized_param.cache.get() is None
```
jenshnielsen, line 174: The test now needs to be updated

thangleiter: Whoops, my bad.

jenshnielsen (approved):

PR #6075, opened 2024-05-15
Add update_snapshot arg to load_instrument

thangleiter:
`Station.add_component()` allows the keyword argument `update_snapshot`. This change modifies the `load_instrument()` method to also accept this keyword and pass it on to `add_component()` when calling it.


jenshnielsen:
Thanks @thangleiter looks good. Could you add a newsfragment describing the change as detailed in the pull request template

jenshnielsen (approved):

jenshnielsen:
@thangleiter I would recommend you installing the precommit hooks that qcodes uses. This should safe you some time on these error from darker and the precommit hooks. See https://pre-commit.com/

thangleiter:
They are installed... no complaints locally. Anyhow, I made the change manually.

PR #6076, opened 2024-05-15
Driver bugfixes [Needs testing Keysight 34465 and  Triton 450]

thangleiter:
These are three small bugfixes on different drivers.

- The DecaDAC range did not include the highest bit
- Our Keysight 34465 model returns a firmware version string incompatible with `convert_legacy_version_to_supported_version`, so it's chopped.
- The current section parsing did not work with our Triton 450. Replaced by a regex.

It would be great if someone could verify the last two changes work on someone else's instruments.


src/qcodes/instrument_drivers/oxford/triton.py
```diff
@@ -461,7 +461,7 @@ def _get_temp_channel_names(self, file: str) -> None:
             options = config.options(section)
             namestr = '"m_lpszname"'
             if namestr in options:
-                chan_number = int(section.split("\\")[-1].split("[")[-1]) + 1
+                chan_number = int(re.findall('\d+', section)[-1]) + 1
```
jenshnielsen, line 464:
```suggestion
                chan_number = int(re.findall(r'\d+', section)[-1]) + 1
```
`\` that is not for explicit control seqs like `\n` needs to either be escaped or be in a raw string. IHMO raw strings are probably easier to read so thats what I propose above

jenshnielsen:
@thangleiter Thanks for the contribution. I don't have access to test these but they seems sensible to me. Could you add a newsfragment as descripted in the pull request template. You will also need to run the precommit hooks to fix small formatting issues.

jenshnielsen (approved):

jenshnielsen:
@thangleiter FYI you don't need to worry about merging main into the branch. We will automatically test against latest main when merging

jenshnielsen:
@thangleiter I changed the title to try to ask for testing a bit more clear. I will leave it for a few days before merging

jenshnielsen:
@thangleiter Lets go ahead and merge this. Can you fix the darker issue or give contributors access to push to your branch so I can fix it

Issue #6080, opened 2024-05-16
Parameter.__getitem__ behavior


thangleiter:
`Parameter`s can be indexed to return `SweepFixedValues`. This is a handy shortcut if you know it exists and use it correctly. However, it can lead to extremely unpredictable behavior when parameters are used in the wrong context, for example if a `Parameter` is passed to a function that expects an Iterable.

The function might try to convert the Iterable to a list and calls `list(param)`, which runs forever if the parameter's validator passes all positive integers. In my opinion, this is very unhandy behavior and not worth the syntactic sugar of writing `param[:10]` instead of having a method that achieves the same thing without the unintended consequences of making `Parameter` iterable.

An example, where the unassuming user passes a Parameter instead of a sequence of parameters as setpoints to `Measurement.register_parameter`:

```python
from qcodes.parameters import ManualParameter, ParameterWithSetpoints
from qcodes.validators import Arrays
from qcodes.dataset import Measurement

getme = ManualParameter('foo')
setme = ManualParameter('bar')

meas = Measurement()
meas.register_parameter(setme)
meas.register_parameter(getme, setpoints=setme)
# should error, since setpoints should be Sequence[Union[str, "ParameterBase"]]
# instead, raises obscure error
AttributeError: 'SweepFixedValues' object has no attribute 'register_name'
```

For `ParameterWithSetpoints`, `register_parameter` ends up trying to convert the argument to a list:
```python
getme_setpoints = ParameterWithSetpoints('baz', setpoints=(setme,), vals=Arrays(shape=(10,)))

meas = Measurement()
meas.register_parameter(setme)
meas.register_parameter(getme_setpoints, setpoints=setme)
# should error, instead runs forever
```
In this case, it would even be cumbersome to check in `_register_parameter_with_setpoints()` if `setpoints` is the correct type, i.e., , because `iter(setpoints)` works even if setpoints is a `Parameter`! (*Well, `Parameter` is not an instance of `Sequence`, so not too cumbersome.)

Also runs forever:
```python
for _ in setme:
    pass
```

My proposal is therefore to deprecate `Parameter.__getitem__` and instead introduce a method like `Parameter.sweep(keys: int | slice)` or even mirror `xarray` and call it `Parameter.iloc` or something.

jenshnielsen:
It should indeed be deprecated see https://github.com/microsoft/Qcodes/pull/5036 

[...]

PR #6086, opened 2024-05-17
Deprecate convenience methods on Instrument classes

jenshnielsen:
Following the discussion in #6080 I think we should consider getting rid of these methods on the instrument classes.

All of this functionality can be done in ways that is more obvious, direct and more frequently used 

E.g. parameter and functions can be looked up from the `parameters` and `functions` delegate attr dicts if you need to look them up dynamically. 

- [x] Fix any use of this in qcodes
- [x] Merge https://github.com/QCoDeS/Qcodes_contrib_drivers/pull/362 first
- [x] Newsfragment

### DIFF
--- a/docs/changes/newsfragments/6086.breaking
+++ b/docs/changes/newsfragments/6086.breaking
@@ -1,0 +1,3 @@
+The methods `get`, `set`, `call` and `__getitem__` on the `InstrumentBase` class have been deprecated.
+Parameters can be looked up by name using the `Instrument.parameters` dict and functions using `instrument.functions`
+which is cleaner and fully equivalent.
--- a/docs/examples/DataSet/Offline Plotting Tutorial.ipynb
+++ b/docs/examples/DataSet/Offline Plotting Tutorial.ipynb
@@ -977,7 +977,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.12.3"
+   "version": "3.12.4"
   },
   "toc": {
    "base_numbering": 1,
--- a/src/qcodes/instrument/instrument.py
+++ b/src/qcodes/instrument/instrument.py
@@ -129,7 +129,7 @@ class Instrument(InstrumentBase, metaclass=instrument_meta_class):
         # start with an empty dict, just in case an instrument doesn't
         # heed our request to return all 4 fields.
         idn = {"vendor": None, "model": None, "serial": None, "firmware": None}
-        idn.update(self.get(idn_param))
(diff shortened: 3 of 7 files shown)


astafan8 (approved):

sldesnoo-Delft:
This deprecation results in quite some warnings in our code. I agree that the code is cleaner, but I don't like it that code used in the labs will break in many places when this functionality is removed.

jenshnielsen:
@sldesnoo-Delft Thanks for your input. In the code that I had access to I saw very few uses of these patterns so good to know they are being used elsewhere. I would suggest that we leave this code deprecated but there is no reason that we have to remove it any time soon if there is real usage of it. That way users will be guided in the right direction. 

BTW since we are using the new 3.13 deprecated decorator (backported from typing extensions) you can use a type checker for catching most usage of deprecated code. I have been using pyright with a config like this (pyright.json)

```
{
    "typeCheckingMode": "off",
    "reportDeprecated": true
}
```

as a quick way to find all calls to deprecated methods etc. 

I believe that mypy is also in the process of adding support for this in the next release or so.

grahamnorris:
I just want to chime in here as someone from a lab that has built a [software package](https://github.com/QudevETH/PycQED_py3.) where many objects (qubits, AWGs, *etc*.) are instances of qcodes InstrumentBase.
This change results in a *lot* of deprecation warnings at once.
Many measurements may now emit more than 10 of these.

While we will refactor the several hundred instances of get/set, I would like to understand the reasoning here a bit better.
Are there any performance or other technical reasons for this change, or is it rather one of code organization?
#6080 doesn't seem to explicitly discuss get and set, but rather focuses on indexing.

I also briefly note that these methods are not shown as deprecated in the [API documentation](https://microsoft.github.io/Qcodes/api/instrument/index.html#qcodes.instrument.InstrumentBase)

jenshnielsen:
@grahamnorris 

Thanks for your message #6657 will add the deprecation to the docstrings. 

We made this deprecation since these methods do not add any value to qcodes. The main aim is to make it easier for users to 

Some of the issues

* Looking up a parameter by name is not type safe unlike using the attribute. As you cannot know the type of instrument["myparam"]
* They are surprising and not very pytonic. E.g. In python you don't look up attributes by name on classes using `__getitem__/[]` That is something that is reserved for dicts/mappings. You dont call an object by passing the name of that object to a function etc.
* They violate the zen of python. `There should be one-- and preferably only one --obvious way to do it.` Since they add an additional way to do things and they are not obvious. 


Note that as with all other python warnings you have the option to silence them until you have the time to perform a transition. 

https://docs.python.org/3/library/warnings.html#the-warnings-filter

PR #6105, opened 2024-05-22
Refactor AimTTI drivers 

jenshnielsen:
* Convert attributes to be assigned using automatic tool
* Document baseclass and disable documentation of inherited attributes. This is such that the duplication is reduced, and the docs becomes more readable.

### DIFF
--- a/src/qcodes/instrument_drivers/AimTTi/_AimTTi_PL_P.py
+++ b/src/qcodes/instrument_drivers/AimTTi/_AimTTi_PL_P.py
@@ -8,7 +8,7 @@ from qcodes.instrument import (
     VisaInstrument,
     VisaInstrumentKWArgs,
 )
-from qcodes.parameters import create_on_off_val_mapping
+from qcodes.parameters import Parameter, create_on_off_val_mapping
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


src/qcodes/instrument_drivers/AimTTi/_AimTTi_PL_P.py
```diff
@@ -85,23 +93,26 @@ def __init__(
             "Here, the integer 1 is for the Low range, "
             "and integer 2 is for the High range.",
         )
+        """Set the current range of the output.Here, the integer 1 is for the Low range, and integer 2 is for the High range."""
```
astafan8, line 96: Missing space
```suggestion
        """Set the current range of the output. Here, the integer 1 is for the Low range, and integer 2 is for the High range."""
```

astafan8 (approved):

src/qcodes/instrument_drivers/AimTTi/_AimTTi_PL_P.py
jenshnielsen: Yes. this is a typo in the original docstring too so should also be fixed there

PR #6139, opened 2024-06-02
Rewrite refactor tools using python 3.10 match

jenshnielsen:
The match syntax is imho cleaner and more natural that the libcst matcher

Draft as this will require us to drop python 3.9 support first


src/qcodes/extensions/_refactor.py
```diff
@@ -53,69 +52,35 @@ def visit_Arg(self, node: cst.Arg) -> None:
...
+        match node:
+            case cst.Arg(
+                value=cst.SimpleString(e_value), keyword=None
+            ) if self._arg_num[self._call_stack[-1]]:
+                self.annotations.name = e_value.strip("\"'")
+            case cst.Arg(
+                keyword=cst.Name(value="name"), value=cst.SimpleString(e_value)
+            ):
+                self.annotations.name = e_value.strip("\"'")
+            case cst.Arg(
```
astafan8, line 64: this looks indeed more readable except the variables like "arg_is_docstring_dedent" were really helping in the previous version, and now they are not there :( could they be added to the case statements? or perhaps at least as comments?

jenshnielsen: The case statement is more than a boolean so it cannot be extracted into a variable but we can at least add comments

astafan8 (approved):

PR #6192, opened 2024-06-21
Trmorgan/keyboard interrupt

trevormorgan:
# Improved Interrupt Handling in dond Function

## Description
This PR enhances the interrupt handling capabilities of the `dond` function in the QCoDeS dataset module. It addresses issues with the current implementation of `catch_interrupts` and ensures that keyboard interrupts are properly propagated and handled during measurement sweeps.

## New
- Ensured that `KeyboardInterrupt` is properly raised and propagated through the `dond` function

## Testing
The updated test now correctly simulates a `KeyboardInterrupt` within the `dond` function and verifies that it's propagated as expected. This change improves our ability to test and ensure robust interrupt handling in complex measurement scenarios.

## Impact
These changes improve the reliability of measurements that use the `dond` function, particularly in situations where users might need to interrupt long-running sweeps. It ensures that the system responds appropriately to keyboard interrupts, possible saving time due to prior option being complete abort

## Additional Notes
- This PR does not introduce any changes to the core `dond` function itself, only to its testing infrastructure
- Users should not experience any changes in normal operation, but will benefit from improved interrupt handling


tests/dataset/dond/test_do1d.py
```diff
@@ -313,19 +313,15 @@ def test_do1d_break_condition(caplog: LogCaptureFixture, _param_set, _param) ->
...
+            _param_set,
+            start,
+            stop,
+            num_points,
+            delay,
+            _param,
+            break_condition=break_condition,
+        )
+
+    assert "Measurement has been interrupted, data may be incomplete" in caplog.text
```
jenshnielsen, line 327: That you no longer get a partial dataset back if you interrupt the measurement feels like a breaking change that we should be clear about and at least document

src/qcodes/dataset/dond/do_nd_utils.py
```diff
@@ -120,14 +120,22 @@ def _register_actions(
...
-    def get_interrupt_exception() -> MeasInterruptT:
+    def get_interrupt_exception() -> MeasInterruptT | None:
         nonlocal interrupt_exception
         return interrupt_exception
 
     try:
         yield get_interrupt_exception
     except (KeyboardInterrupt, BreakConditionInterrupt) as e:
         interrupt_exception = e
+        interrupt_raised = True
```
jenshnielsen, line 135: I don't think the `BreakConditionInterrupt` should bubble up. That one is intended to cancel a specific dond when a given condition is meet.

tests/dataset/dond/test_do1d.py
trevormorgan: thats a good point, is this a breaking change we are okay with or should we preserve the data that was captured before the interrupt?

trevormorgan: I believe my latest push makes only keyboard interrupts that result in no partial datasets created. Does this sound fair? I'll update the docs and notes for upcoming change logs accordingly if that sounds good.

jenshnielsen: @trevormorgan I think this is nearly ready to go. Could you add a changelog fragment. There also seems to be some test errors on windows that needs to be sorted

trevormorgan: @jenshnielsen I'll get the fragement added and resolve the test errors today. Thanks for the review!

jenshnielsen (approved):

PR #6293, opened 2024-07-30
Keysight Infiniium: feat: add input (coupling & impedance) for channels

konosubakonoakua:
also fix channel numbers for trigger


konosubakonoakua:

jenshnielsen (approved):

PR #6305, opened 2024-08-02
Fix: Check exp_id in experiment list

GOTYGuyOfTheYear:
Instead of checking with just the length of the experiment list.
Check that the exp_id is in the experiment list.
Useful when working with an offset or missing entries in the database.


jenshnielsen:
@GOTYGuyOfTheYear Thanks

jenshnielsen (approved):

PR #6341, opened 2024-08-12
Docs: use furo theme

jenshnielsen:
This is a replacement for #4395 We originally closed that since the table of content did not work as expected, see https://github.com/microsoft/Qcodes/pull/4395#issuecomment-1184377533

However, this seems to mostly be due to the way we organized our examples. This reorganizes the examples so it looks like this:

image

Unlike the rtd theme this theme supports the latest versions of sphinx and docutils.


jenshnielsen:
Some other examples

Furo without the reorg:

image

pydata-sphinx-theme without the reorg

image

Reorganized examples with pydata-sphinx-theme

image

astafan8 (approved):

PR #6394, opened 2024-08-28
Gracefully handle that a parameter could shadow a class attr

jenshnielsen:
When removing the parameter

### DIFF
--- a/tests/parameter/test_parameter_override.py
+++ b/tests/parameter/test_parameter_override.py
@@ -1,3 +1,4 @@
+import logging
 from typing import TYPE_CHECKING
 
 import pytest
@@ -120,7 +121,7 @@ def test_removed_parameter_from_prop_instrument_works(request):
     a.voltage.set(1)
 
 
-def test_remove_parameter_from_class_attr_works(request):
+def test_remove_parameter_from_class_attr_works(request, caplog):
     request.addfinalizer(DummyClassAttrInstr.close_all)
     a = DummyClassAttrInstr("my_instr")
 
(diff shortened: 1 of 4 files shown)


tests/parameter/test_parameter_override.py
```diff
@@ -96,3 +119,41 @@ def test_removed_parameter_from_prop_instrument_works(request):
...
+    a = DummyClassAttrInstr("my_instr")
+
+    # removing a parameter that is assigned as an attribute
+    # with a class level type works
+    assert hasattr(a, "current")
+    a.remove_parameter("current")
+    assert not hasattr(a, "current")
+
+    # when we remove a parameter with an attr that shadows a class attr
+    # we get the class attr after the removal
```
einsmein, line 135: This behavior might be unexpected. If we keep it this way, we should document it.

jenshnielsen: @einsmein This is exactly the same way as instance attributes and class attributes already work in python so I don't think this is surprising as such. And if it is surprising then this is a thing that should be documented in python the language not here. 

Furthermore, we already raise a warning when a parameter overrides an attribute so this is not really a situation that anyone should get in unless their code is already generating warnings. 

```python
In [11]: class Foo():
    ...:     a: str = "classattr"
    ...:     def __init__(self):
    ...:         self.a = 1
    ...:

In [12]: foo = Foo()

In [13]: foo.a
Out[13]: 1

In [14]: delattr(foo, 'a')

In [15]: foo.a
Out[15]: 'classattr'
```

einsmein (approved):

PR #6439, opened 2024-09-12
Remove opencensus telemetry

jenshnielsen:
This has been deprecated for a while so lets get rid of it.

- [x] Document breaking change

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,10 +113,6 @@ docs = [
     "qcodes_loop>=0.1.1", # legacy dataset import examples
     "jinja2>=3.1.3", # transitive dependency pin due to cve in earlier version
 ]
-opencensus = [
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -113,10 +113,6 @@ docs = [
     "qcodes_loop>=0.1.1", # legacy dataset import examples
     "jinja2>=3.1.3", # transitive dependency pin due to cve in earlier version
 ]
-opencensus = [
```
astafan8, line 116: add newsfragment?

astafan8 (approved):

pyproject.toml
jenshnielsen: @astafan8 Rather remember to push the fragment that I had already written ;)

PR #6487, opened 2024-10-04
Samanthaho/registered parameter snapshot

samantha-ho:
This PR adds a new category to the dataset snapshot. Previously, when a qcodes Measurement ran, only the station snapshot was saved. The parameters directly involved in the measurement were not explicitly snapshotted, and if those parameters were not part of the station, metadata about them could be lost.

The dataset snapshot dictionary now contains a new top-level key, "parameters" in addition to the original "station" key. Under parameters, there are snapshots of all parameters registered in the Measurement.


src/qcodes/dataset/measurements.py
```diff
@@ -1278,23 +1288,37 @@ def unregister_parameter(self, parameter: setpoints_type) -> None:
...
-            paramspec: ParamSpecBase = self._interdeps[param]
+            paramspec: ParamSpecBase = self._interdeps[param_name]
         except KeyError:
             return
 
         self._interdeps = self._interdeps.remove(paramspec)
 
-        log.info(f"Removed {param} from Measurement.")
+        # Must follow interdeps removal, because interdeps removal may error
+        if isinstance(parameter, ParameterBase):
```
trevormorgan, line 1308: This looks good, do you think it's worth adding a test?

trevormorgan (approved):
Nice addition!

src/qcodes/dataset/measurements.py
samantha-ho: The existing tests cover this pretty well, it turns out. I had to fix a bunch of stuff to get them to pass.

PR #6494, opened 2024-10-06
Mark python 3.13 as supported and test / type check with it

jenshnielsen:
Filter new warnings.

- [x] Add 3.13 to CI
- [x] Add markers for 3.13
- [x] Can we find a clean way to ensure that connections to db are always closed? Tracked as #6584

Requires pr to bump pywin32 to be merged

Still missing release with support for 3.13 for 
(could be disabled)

- [x] Pyarrow https://github.com/microsoft/Qcodes/pull/6569
- [x] zhinst-core https://github.com/microsoft/Qcodes/pull/6582

Other issues
- [x] Docs don't build. Tracked in #6585

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -29,14 +29,14 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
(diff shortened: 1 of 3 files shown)


jenshnielsen:
disabling the extensions one by one seems to indicate that the docs build error is related to autodocsumm

einsmein (approved):

PR #6544, opened 2024-10-18
Fix add_to_spyder_UMR_excludelist for Spyder 6

sldesnoo-Delft:
The import of qcodes fails with Spyder 6, because add_to_spyder_UMR_excludelist raises an exception.

This pull request fixes the code to work with Spyder 6.0.
I've removed a bit of old code that was only useful for Spyder 3.x

I don't think this is easy to test with an automatic test, because a test would require Spyder 6.

### DIFF
--- a/src/qcodes/utils/spyder_utils.py
+++ b/src/qcodes/utils/spyder_utils.py
@@ -1,5 +1,7 @@
 import logging
 import os
+from packaging.version import Version
+
 
 _LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


src/qcodes/utils/spyder_utils.py
```diff
@@ -11,39 +13,46 @@ def add_to_spyder_UMR_excludelist(modulename: str) -> None:
...
+                # In Spyder 4 and 5 UMR is a variable in module spydercustomize
+                try:
+                    from spyder_kernels.customize import spydercustomize  # pyright: ignore
+                except ImportError:
+                    return
+                else:
+                    umr = spydercustomize.__umr__
             else:
-                sitecustomize_found = True
+                # In Spyder 6 UMR is an attribute of the SpyderCodeRunner object.
```
eendebakpt, line 34: I also encountered qcodes not running with the new Spyder 6. The `UserModuleReloader` is also available via a normal import:
```
import spyder_kernels.customize.umr; spyder_kernels.customize.umr.UserModuleReloader
```
Using that instead of `sitecustomize.UserModuleReloader` works for me so far.

@ccordoba12 Do you have an opinion on how to approach this?

sldesnoo-Delft: I tried that as well. 
Unfortunately Spyder6 does not use the global __umr__. So assignment of the changed UMR has no effect.

That's why I searched for the UMR instance that is being used and modify that.

jenshnielsen: @eendebakpt @sldesnoo-Delft  I am not using spyder my self at the moment so unlikely to spot this. 

At some point a long time ago I did look into addressing this issue on the spyder side https://github.com/spyder-ide/spyder/issues/2451 If anyone have bandwidth to do that it would be great as this is a much cleaner option. 

Otherwise happy to merge this once you have converted on the best solution

ccordoba12: > @ccordoba12 Do you have an opinion on how to approach this?

I wasn't aware people was adding modules to our UMR programmatically (no one has reported it to our issue tracker).

Since the code to do that for Spyder 6/Spyder-kernels 3 is too convoluted (from what I can see below), we can add a public API at the shell level (i.e. the object you get with `get_python()`) to make the life of other projects easier.

@jenshnielsen, what do you think?

sldesnoo-Delft: qcodes adds itselft to the list of packages to exclude. 

I had a closer look at the UMR. It checks if the module is in the library path. (`path_is_library` in utils.py). When qcodes is normally installed, i.e. not in "develop mode", then it will not be reloaded by UMR.  

So, I believe the add_to_spyder_UMR_excludelist function is not really needed. It might be a remainder from a distant past.

@jenshnielsen , @eendebakpt do you agree?

jenshnielsen: @sldesnoo-Delft I think this has ever only been a problem with editable installs. As I recall the problem is that when spyder does reload all singletons are destroyed so instruments are lost. I think that when we originally added this qcodes did not see regular releases, so a lot of users were installing from source. This is much less of an issue now so I think it would be fine to deprecate/remove this.

eendebakpt: I have no strong opinion on this, we usually disable the spyder UMR (other packages have issues with automatic reloading as well). We only noticed because the new spyder 6 would not import qcodes.

If we do remove the functionality, I do think it warrants some attention in the release notes (and maybe the documentation?).

jenshnielsen: @sldesnoo-Delft I opted for removing this from automatically running and deprecating the function in #6647

Are you ok closing this

PR #6565, opened 2024-10-25
[wip] Try to remove invalid netcdf feature

jenshnielsen:
This should no longer be required with latest xarray/h5netcdf

To merge this we either need to bump min versions of xarray and h5netcdf or probably more realistically keep the feature for older versions


jenshnielsen:
Replaced by https://github.com/microsoft/Qcodes/pull/6566

PR #6573, opened 2024-10-29
Hotfix data_set.py: flatten 2D arrays

GOTYGuyOfTheYear:
When adding a 2D array as a standalone parameter in a dataset, the whole numpy array would be save in bit form. Flatten is needed to make a proper array to be saved alongside the data.

### DIFF
--- a/src/qcodes/dataset/data_set.py
+++ b/src/qcodes/dataset/data_set.py
@@ -1395,13 +1395,13 @@ class DataSet(BaseDataSet):
                     res_list += new_res
             elif param.type == "numeric":
                 if value.shape:
-                    res_list += [{param.name: number} for number in value]
+                    res_list += [{param.name: number} for number in value.flatten()]
                 else:
                     new_res = [{param.name: float(value)}]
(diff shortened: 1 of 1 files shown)


jenshnielsen:
> When adding a 2D array as a standalone parameter in a dataset, the whole numpy array would be save in bit form. Flatten is needed to make a proper array to be saved alongside the data.

I am not sure that I follow what the issue is. When a user stores an array of a given shape I would imagine that they would also expect this to be read back.

Could you perhaps supply an example or a test that shows the issue that you are facing?

GOTYGuyOfTheYear:
Hello @jenshnielsen,

Thank you for your interest. 
The issue appears when value is a 2D numpy array. 
I guess the python line : [{param.name: number} for number in value] interates through the columns of the array.
When we try to load the array and convert it to xarray again we get this error:
image

In the sqlite db, the fields for the independent parameter are filled with a blob type:
image

There might be a better fix or a better way of working, this is just how I have found as the simplest option.

PR #6599, opened 2024-11-07
Add methods to find parameter chain links by type

samantha-ho:
This PR adds two new methods to find links in a parameter chain of a particular type. These methods are most useful with user-defined linking parameters.


src/qcodes/extensions/infer.py
```diff
@@ -1,16 +1,18 @@
...
 from qcodes.instrument import Instrument, InstrumentBase, InstrumentModule
-from qcodes.parameters import DelegateParameter, Parameter
+from qcodes.parameters import DelegateParameter, Parameter, ParameterBase
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
 
 DOES_NOT_EXIST = "Does not exist"
 
+C = TypeVar("C", bound=ParameterBase)
```
picarro-yren, line 14: Hi Samantha, methods you added here are quite similar to the one that you suggested me to add about one month ago. The difference is that I need a method to find specific type of instrument from delegation chain. Could you make `C` bounded to `Union[ParameterBase, InstrumentBase]` so new methods can also be used to find instruments?

samantha-ho: If we do that, then I'd like to add the instrument lookup versions too. 

Alternatively, we could add the instrument lookup methods in a separate PR and update the TypeVar then. For expediency, I prefer this second option.

picarro-yren: Ok, I'll create a PR to update TypeVar.

picarro-yren (approved):

PR #6648, opened 2024-11-20
DelegateParameters inherit their source's settable/gettable/snapshot_value/unit/label

samantha-ho:
This PR fixes a bug observed when chaining multiple DelegateParameters, and changing the source at the bottom of the chain.

In the previous implementation, a DelegateParameter only updated it's internal `_settable` and `_gettable` attributes when the `source` was changed. But it is possible to set up a chain of DelegateParameters and then change the 'root' source Parameter. This updates the bottom-most DelegateParameter, but all of the other DelegateParameters in the chain will not update.

This PR changes the implementation to override the `ParameterBase` implementation of the `gettable` and `settable` properties, so that they reflect the underlying `source` parameter, no matter what it may be.

### DIFF
--- a/docs/changes/newsfragments/6648.improved
+++ b/docs/changes/newsfragments/6648.improved
@@ -1,0 +1,2 @@
+DelegateParameter now reads delegated properties from source rather than setting them at source change.
+This resolves an issue when delegating from another delegate parameter may result in incorrect attributes.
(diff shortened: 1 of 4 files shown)


samantha-ho:
@jenshnielsen 
One possible issue... there is a seemingly rarely used `_SetParamContext` manager in the ParameterBase module that directly sets the `parameter._settable` attribute. This implementation will not properly handle that case, as the internal `_settable` and `_gettable` attributes on a DelegateParameter do not get used.

Not sure if this is a big enough issue to rethink the implementation. What are your thoughts?

einsmein (approved):

PR #6657, opened 2024-11-22
Add note about deprecated methods to docstrings

jenshnielsen:
As noted in #6648 this really should be documented.

### DIFF
--- a/src/qcodes/instrument/instrument_base.py
+++ b/src/qcodes/instrument/instrument_base.py
@@ -701,7 +701,7 @@ class InstrumentBase(MetadatableWithName, DelegateAttributes):
 
 
         Note:
-            This is deprecated call set directly on the parameter.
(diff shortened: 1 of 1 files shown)


src/qcodes/instrument/instrument_base.py
```diff
@@ -708,6 +719,9 @@ def get(self, param_name: str) -> Any:
         Returns:
             The current value of the parameter.
 
+        Note:
+            This is deprecated call get directly on the parameter.
```
einsmein, line 723: "This is deprecated. Call get directly on the parameter."
Also apply for the change below.

einsmein (changes requested):
One minor comment and LGTM

jenshnielsen:
@einsmein Fixed, ready for another review.

einsmein (approved):

PR #6666, opened 2024-11-25
Fix deprecation warning from codecov action

jenshnielsen:
file has been replaced by files (called the same way since we only upload one file)


jenshnielsen:
image

samantha-ho (approved):

PR #6671, opened 2024-11-27
Fix bug: delegate parameter doesn't call its source to get value

picarro-yren:
**Bug description:**
When a dyno test point is read, it calls external_parameter  which should redirect to its source for the value but instead <function Parameter.__init__.<locals>._get_manual_parameter> is called and 0.0 is returned.
 
**Cause**
The reason why DelegateParameter doesn't call its source for the value is because of the following lines in Parameter.__init__:

```
elif not self.gettable and get_cmd is not False:
      if get_cmd is None:
                self.get_raw = MethodType(_get_manual_parameter, self)
```
 
When external_parameter is created, its source is None so it's not gettable and its get_raw is thus bound to _get_manual_parameter.

**Solution**
It turns out that even though gettable property of DelegateParameter is False when source is None, _gettable attribute is True because `get_raw` is defined in parameter
 
So the solution we propose is to use _gettable attribute instead of gettable property in Parameter.__init__.

### DIFF
--- a/src/qcodes/parameters/parameter.py
+++ b/src/qcodes/parameters/parameter.py
@@ -245,7 +245,9 @@ class Parameter(ParameterBase):
             **kwargs,
         )
 
-        no_instrument_get = not self.gettable and (get_cmd is None or get_cmd is False)
+        # Note: use _gettable/_settable attributes in __init__ function.
+        # Avoid using gettable/settable properties here.
+        no_instrument_get = not self._gettable and (get_cmd is None or get_cmd is False)
         # TODO: a matching check should be in ParameterBase but
         #   due to the current limited design the ParameterBase cannot
         #   know if this subclass will supply a get_cmd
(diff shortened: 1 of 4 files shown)


src/qcodes/parameters/parameter.py
```diff
@@ -261,13 +263,13 @@ def _set_manual_parameter(
         # in the scope of this class.
         # (previous call to `super().__init__` wraps existing
         # get_raw/set_raw into get/set methods)
-        if self.gettable and get_cmd not in (None, False):
+        if self._gettable and get_cmd not in (None, False):
```
jenshnielsen, line 266: Lets expose a property on ParameterBase (_has_get_raw,_has_set_raw) that we can check here. That will make the code more readable and expose what we are actually checking

einsmein (approved):

PR #6700, opened 2024-12-04
Add `on_value_change` callback to ParameterBase

DCEM:
Introduce an optional `on_value_change` callable to the `ParameterBase` class. This callback is invoked whenever the parameter's value changes, providing: old value, new value, and the source of the change (`'get'` or `'set'`).


DCEM:
I think there is a better way to do this. I hopefully will do another PR regarding this

PR #6723, opened 2024-12-16
Use correct sense function for sense parameters in Keysight 344xxA driver

janekfleper:
The Keysight344xxA class implements the sense parameters without checking the sense function.

The parameters in the SENSE subsystem have to be handled based on the current sense function. By default the VISA path is always `"SENSe:VOLTage:DC"` and the parameters will only get/set the values for the DC voltage measurement. The actual sense function that is currently used by the instrument completely ignored.

I was not entirely sure how to implement this. I had the following other ideas how to solve this:
1. Create a `SenseParameter` class that overrides `get_raw()` and `set_raw()` to read the sense function just before calling either `get()` or `set()`, but according to the documentation you cannot use both `get_raw()` and `get_cmd` at the same time. And it would also be overkill since I actually just need a decorator for the methods `get_raw()` and `set_raw()`, I don't need to completely modify them. Is there a way to add such a decorator?
2. Create an instrument module `Keysight344xxASense` for all the sense parameters to move them from the root instrument to `self.sense`. I could then override the methods `ask()` and `write()` in the module to "inject" the correct sense function path just before actually sending the command to the VISA handle.
3. Pass callable methods to the parameters `get_cmd` and `set_cmd` and manually call `self.ask()` and `self.write()` with the correct commands.

The merge request currently implements solution 3 for the parameters `NPLC`, `autorange`, `autozero`, `aperture_mode` and `aperture_time`. The parameters `range` and `resolution` are still missing since their validators require some extra work.

Are you happy with the current approach or would you prefer one of the other two solutions? Or is there something that I am missing completely that would make the implementation a lot more straight forward?

I am looking forward to your feedback!


jenshnielsen:
@einsmein Could you review this since I think you looked at the 344xx not that long ago

einsmein:
@janekfleper The second option would be ideal. It reflects the instrument interface where we specify measurement types with every sense function.
That being said, this approach makes sense given the interface we have now. (The non-ideal part is that by setting `sense_function` parameter, we implicitly modify the behavior of other parameters. But we can address that in the future.)

Have you tested this code on the instrument? In that case LGTM

src/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -1212,14 +1212,32 @@ def read(self) -> np.ndarray:
         raw_vals: str = self.ask("READ?")
         return _raw_vals_to_array(raw_vals)
 
+    def _ask_with_sense_function(self, cmd: str) -> str:
+        function = self.sense_function.get_latest.get_raw().strip("\"")
```
astafan8, line 1216: this is a more correct way of getting raw value from cache
```suggestion
        function = self.sense_function.cache.raw_value.strip("\"")
```

src/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -1212,14 +1212,32 @@ def read(self) -> np.ndarray:
...
+        return self.ask(f"SENSe:{function}:{cmd}?")
+
+    def _write_with_sense_function(self, cmd: str, value: str) -> None:
+        function = self.sense_function.get_latest.get_raw().strip("\"")
```
astafan8, line 1220:
```suggestion
        function = self.sense_function.cache.raw_value.strip("\"")
```

astafan8 (approved):
just a tiny correction :)

janekfleper:
I have tested the changes on the model 34465A and I would also have the models 34470 and 34411A available for further testing. But since these models are all equally affected by the changes, I don't think it is necessary here.

src/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
janekfleper: The unit tests are failing here since `cache.raw_value` is None. (I don't know if this was an issue before because the CI required approval to run.) Is my understanding correct that the cache will only be populated after `self.sense_function()` is executed at least once?

jenshnielsen: I agree this is a bug in the cache implementation that there is no way to do this. I am also pusled why the code did not error before. For now, I have pushed code to call get on the cache before using the raw value. That should ensure its always up to date

src/qcodes/instrument_drivers/Keysight/private/Keysight_344xxA_submodules.py
```diff
@@ -748,7 +748,7 @@ def __init__(
 
         self.NPLC: Parameter = self.add_parameter(
             "NPLC",
-            get_cmd="SENSe:VOLTage:DC:NPLC?",
+            get_cmd=self._get_with_sense_function("NPLC"),
```
jenshnielsen, line 751: @janekfleper The tests are now failing because this changes the command from `SENSe:VOLTage:DC:NPLC?` to `SENSe:VOLTage:NPLC?` Is that intentional? If so the Keysight_34465A yaml file would need to be updatd with the new command

janekfleper: The :DC suffix is the default for the voltage (and the current) and therefore optional. Both versions are correct, the instrument won't mind. In the manual the return values for the sense function are listed as:
```
...
RESistance
TEMPerature
VOLTage:AC
VOLTage[:DC]
...
```
If you read the sense function from the instrument the raw value is just `"VOLT"`, and the resulting get command here would be `SENSE:VOLT:NPLC?"`. This is just an unfortunate consequence of the SCPI lanuage.
Replacing `VOLTage:DC` by `VOLT` in the yaml file for the properties `NPLC`, `autorange`, `autozero`, `aperture_mode` (called `aperture_enabled` in the yaml file) and `aperture_time` should therefore make the tests pass again. But this would just test the old case of the sense function being hard-coded to a DC voltage. Do you think there is a way to make the tests/simulation of the instrument aware of the actual sense function? Or is this not necessary?

jenshnielsen: @janekfleper The pyvisa yaml language is rather stateless so I don't think we can do much else than ensure that the yaml file knows about the valid commands without trying to model that some of them only works when the instrument is in a specific state. 

As in we can add `NPLC`, `autorange`, `autozero`, `aperture_mode` commands for all valid sense modes and test that they all work but that is the limit of what we can do.

janekfleper: Do you just want me to add all the commands then? I would name the properties `voltage_dc_NPLC` etc. and use the raw values for the sense functions in the actual commands to make sure that the tests pass.

jenshnielsen: Yes that would be great

PR #6792, opened 2025-01-14
Oxford Triton driver: only add magnet parameters if one is present

thangleiter:
This patch tests if the Triton can communicate with the magnet. If not, parameters pertaining to the magnet are not added to the instrument. This avoids complications when snapshotting the instrument for instance.

Reasons for communication failure are 
1) no magnet is equipped to begin with.
2) the magnet power supply (Mercury iPS) is set up for a different communication protocol (TCP/IP for example).

### DIFF
--- a/docs/changes/newsfragments/6792.improved
+++ b/docs/changes/newsfragments/6792.improved
@@ -1,0 +1,1 @@
+Only add magnet parameters if a magnet is detected in the Oxford Triton driver.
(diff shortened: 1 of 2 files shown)


jenshnielsen:
thanks @thangleiter Looks fine, Note that this driver is not particularly well maintained. There is also a driver in https://github.com/QCoDeS/Qcodes_contrib_drivers/blob/main/src/qcodes_contrib_drivers/drivers/OxfordInstruments/Triton.py which may be of interest.

thangleiter:
> thanks @thangleiter Looks fine, Note that this driver is not particularly well maintained. There is also a driver in https://github.com/QCoDeS/Qcodes_contrib_drivers/blob/main/src/qcodes_contrib_drivers/drivers/OxfordInstruments/Triton.py which may be of interest.

Thanks, unfortunately that driver is basically just an offline log file parser and does not interact with the fridge at all.

PR #6802, opened 2025-01-17
Check for numeric type using numbers.Real

thangleiter:
In several places code tests variables for numeric type using `isinstance(value, (int, float))`. This creates problems for native numpy types (among others I would presume). For example, `isinstance(np.int32(0), int)` evaluates to false.

Checking instead for `numbers.Real` should resolve this.

### DIFF
--- a/src/qcodes/instrument_drivers/american_magnetics/AMI430.py
+++ b/src/qcodes/instrument_drivers/american_magnetics/AMI430.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import collections.abc
 import logging
-import numbers
 import time
 import warnings
 from collections import defaultdict
@@ -20,6 +19,7 @@ from qcodes.instrument import Instrument, InstrumentChannel, IPInstrument
 from qcodes.math_utils import FieldVector
 from qcodes.parameters import Parameter
 from qcodes.utils import QCoDeSDeprecationWarning
+from qcodes.utils.types import NumberType
 from qcodes.validators import Anything, Bool, Enum, Ints, Numbers
 
 if TYPE_CHECKING:
@@ -635,7 +635,7 @@ class AMI430_3D(Instrument):
         self._field_limit: float | Iterable[CartesianFieldLimitFunction]
         if isinstance(field_limit, collections.abc.Iterable):
             self._field_limit = field_limit
-        elif isinstance(field_limit, numbers.Real):
+        elif isinstance(field_limit, NumberType):
             # Conversion to float makes related driver logic simpler
             self._field_limit = float(field_limit)
         else:
@@ -978,7 +978,7 @@ class AMI430_3D(Instrument):
     def _verify_safe_setpoint(
         self, setpoint_values: tuple[float, float, float]
     ) -> bool:
-        if isinstance(self._field_limit, numbers.Real):
+        if isinstance(self._field_limit, NumberType):
             return bool(np.linalg.norm(setpoint_values) < self._field_limit)
 
         answer = any(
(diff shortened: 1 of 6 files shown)


jenshnielsen:
Thanks @thangleiter 

I am a bit torn on this. On one hand I fully agree with you.

On the other hand the numbers classes don't work well with type checkers. 
See https://github.com/python/mypy/issues/3186#issuecomment-1571512649
This is also the reason you ase seeing some failures in your test suite. 
If possible I think my preference would be to try to get rid of these isinstance checks and transform input into the correct type when passed to a function or method.

Timing wise this is about 10 times slower than checking just for the build in types 
but not that much slower than checking for both numpy and build in types.
I guess in most of these use cases that probably does not matter too much. 
``
In [14]: %timeit isinstance(1, numbers.Real)
273 ns ± 23.3 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [15]: %timeit isinstance(1, int)
36 ns ± 1.7 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [16]: %timeit isinstance(1, float)
56 ns ± 2.84 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [17]: %timeit isinstance(1, (float, np.float64))
152 ns ± 9.11 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [18]: %timeit isinstance(1, (float, np.floating))
219 ns ± 41.1 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [19]: %timeit isinstance(1, np.float64)
90.1 ns ± 3.46 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [20]: %timeit isinstance(1, (int,float))
59.7 ns ± 4.21 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

**Comment by thangleiter** [comment]
I see, I wasn't aware of the unclear state of the `numbers` module.

> If possible I think my preference would be to try to get rid of these isinstance checks and transform input into the correct type when passed to a function or method.

I think this is fine in drivers where you know which type you want to pass on. But the most important point where this comes up is `get_ramp_values()`, and there you'd need to carefully check validators etc to decide whether to cast to int or float (like in the `step` setter), which would definitely defeat the performance argument. 

Since likely the numpy dtypes are the most relevant, how about introducing a `NumbersType: TypeAlias = int | float | np.integer | np.floating` and checking for that? That, by some magic, appears to also be very fast!

```python
>>> %timeit isinstance(1, (int, float, np.integer, np.floating))
114 ns ± 2.95 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
>>> %timeit isinstance(1, NumbersType)
31.3 ns ± 0.763 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
>>> %timeit isinstance(1, int)
28.3 ns ± 0.556 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

jenshnielsen:
@thangleiter That sounds like a good solution until hopefully a better solution appears upstream.

thangleiter:
Looks like this is still a problem for mypy. The `FieldVector` bit should be easily fixed by casting to float, but the `permissive_range` bit... There is a typing inconsistency in the code as is. 

The `step` getter pretends to return a float, but the setter accepts integers as well. `get_ramp_values()` and `permissive_range()` 
assume floats, however, so parameters with integer validators are not covered by the type annotations while they should work perfectly fine.

jenshnielsen:
thanks @thangleiter I think this looks good

jenshnielsen (approved):

PR #6817, opened 2025-01-23
add vals property and setter to delegate_parameter

esbenborkhansen:
Add `vals` property and  `@vals.setter` to qcodes.parameters.delegate_parameter.DelegateParameter to propagate the Validator from source to DelegateParameter.

PR #6832, opened 2025-01-27
[WIP] Add convenience methods to DelegateParameter

jenshnielsen:
``DelegateParameter`` now includes validators of its source Parameter into its validators. This ensures that a ``DelegateParameter``
with a non numeric source parameter is registered correctly in a measurement when the ``DelegateParameter`` it self does not
set a validator. Furthermore `DelegateParameter`` has gained a ``root_source`` attribute that makes it easier to get the
root source of a ``DelegateParameter`` that delegates to another ``DelegateParameter``.

 Alternative to #6817

### DIFF
--- a/src/qcodes/parameters/delegate_parameter.py
+++ b/src/qcodes/parameters/delegate_parameter.py
@@ -212,7 +212,7 @@ class DelegateParameter(Parameter):
         this behaves the same as ``self.source``
 
         :getter: Returns the current source.
-        :setter: Sets the source of the first parameter in the tree that has a None source or non-DelegateParameter source
+        :setter: Sets the source of the first parameter in the tree that has a non-DelegateParameter source.
         """
         if isinstance(self.source, DelegateParameter):
             return self.source.root_source
@@ -228,6 +228,7 @@ class DelegateParameter(Parameter):
         """
         If this parameter is part of a chain of DelegateParameters return
         the first Parameter in the chain that has a non DelegateParameter source
+        else return self.
         """
 
         if not isinstance(self.source, DelegateParameter):
--- a/src/qcodes/parameters/delegate_parameter.py
+++ b/src/qcodes/parameters/delegate_parameter.py
@@ -203,6 +203,23 @@ class DelegateParameter(Parameter):
         self._source: Parameter | None = source
 
     @property
+    def root_source(self) -> Parameter | None:
+        """
+        The root source parameter that this :class:`DelegateParameter` is bound to
+        or ``None`` if this  :class:`DelegateParameter` is unbound. If
+        the source is it self a DelegateParameter it will recursively return that Parameter's
+        source until a non DelegateParameter is found. For a non DelegateParameter source
+        this behaves the same as ``self.source``
+
+        :getter: Returns the current source.
+        :setter: Sets the source.
+        """
+        if isinstance(self.source, DelegateParameter):
+            return self.source.root_source
+        else:
+            return self.source
+
+    @property
     def snapshot_value(self) -> bool:
         if self.source is None:
             return False
--- a/src/qcodes/parameters/delegate_parameter.py
+++ b/src/qcodes/parameters/delegate_parameter.py
@@ -8,6 +8,8 @@ if TYPE_CHECKING:
     from collections.abc import Sequence
     from datetime import datetime
 
+    from qcodes.validators.validators import Validator
+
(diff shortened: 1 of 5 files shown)


src/qcodes/parameters/delegate_parameter.py
```diff
@@ -314,3 +333,18 @@ def validate(self, value: ParamDataType) -> None:
...
+        includes validators of the source parameter if source parameter is set
+        and has any validators.
+
+        :getter: All validators associated with the parameter.
+        """
+        source_validators: tuple[Validator, ...] = (
+            self.source.validators if self.source is not None else ()
+        )
+
+        return source_validators + tuple(self._vals)
```
jenshnielsen, line 350: Is this the right order?

jenshnielsen: Seems most natural that the delegateparameters own validator comes first and is used as the default

tests/dataset/measurement/test_measurement_context_manager.py
```diff
@@ -201,6 +207,23 @@ def test_register_custom_parameter(DAC) -> None:
         )
 
 
+def test_register_delegate_parameters():
+    x_param = Parameter("x", set_cmd=None, get_cmd=None)
+
+    complex_param = Parameter(
+        "complex_param", get_cmd=None, set_cmd=None, vals=ComplexNumbers()
+    )
+    delegate_param = DelegateParameter("delegate", source=complex_param)
```
esbenborkhansen, line 216: The case that failed previously was when the delegate parameter was first created with no source and then appointed a source afterwards, so I suggest to test that case. Unless it is covered by the other new tests and does not need to be part of this test.

```suggestion
    delegate_param = DelegateParameter("delegate", source=None)
    delegate_param.source = complex_param
```

esbenh4 (approved):

tests/dataset/measurement/test_measurement_context_manager.py
jenshnielsen: Extended with another test that covers this

tests/dataset/measurement/test_measurement_context_manager.py
```diff
@@ -201,6 +207,43 @@ def test_register_custom_parameter(DAC) -> None:
...
+    meas = Measurement()
+
+    meas.register_parameter(x_param)
+
+    delegate_param.source = complex_param
+
+    meas.register_parameter(delegate_param, setpoints=(x_param,))
+    assert len(meas.parameters) == 2
+    assert meas.parameters["delegate"].type == "complex"
+    assert meas.parameters["x"].type == "numeric"
```
jenshnielsen, line 244: Lets also add a test for delegate of delegate of complex

jenshnielsen: A -> B -> C -> None

Add source to C

verify that A has the correct validator.

tests/parameter/test_delegate_parameter.py
```diff
@@ -594,6 +599,66 @@ def test_value_validation() -> None:
...
+    with pytest.raises(ValueError):
+        delegate_param.validate(6)
+    assert delegate_param.validators == (some_other_validator, some_validator)
+    assert delegate_param.vals == some_other_validator
+
+    assert delegate_param.source is not None
+    delegate_param.source.vals = None
+
+    assert delegate_param.validators == (some_other_validator,)
+    assert delegate_param.vals == some_other_validator
```
jenshnielsen, line 660: Extend to 2 delegate parameters too

jenshnielsen:
Moved the validators to https://github.com/microsoft/Qcodes/pull/6865 will figure out how to integrate the test with the infer module here in a more consistent way

src/qcodes/parameters/delegate_parameter.py
```diff
@@ -314,3 +350,18 @@ def validate(self, value: ParamDataType) -> None:
...
+        includes validators of the source parameter if source parameter is set
+        and has any validators.
+
+        :getter: All validators associated with the parameter.
+        """
+        source_validators: tuple[Validator, ...] = (
+            self.source.validators if self.source is not None else ()
+        )
+
+        return tuple(self._vals) + source_validators
```
astafan8, line 367: just wonder, if it's cleaner (and even possible) to use super here?
```suggestion
        return super().validators + source_validators
```

astafan8 (approved):

PR #6931, opened 2025-03-03
Feature/control parameters plotted in plot dataset 

### LINKED ISSUE #6898: [feature request] option to control which parameters are plotted in plot_dataset
When using the `plot_dataset` function, all plots for a given dataset are generated i.e. a new axis is created for each dependent parameter in the dataset. 

For example, in the [Offline Plotting Tutorial - Two interleaved 1D sweeps](https://microsoft.github.io/Qcodes/examples/DataSet/Offline%20Plotting%20Tutorial.html#Two-interleaved-1D-sweeps) two parameters `y` and `y2` are measured as a function of the same parameter `x`. Using `plot_dataset` then results in: 

Image

where two plots are generated for the two dependent parameters. I understand there is a way to plot these two on the same graph but is there a way to only plot one of these parameters from the `plot_dataset` function? 

Removing the setpoint `x` from parameter `y` at the point of registering the parameter would achieve this since `y` is no longer dependent on `x` so won't get plotted. However, in the case of using a `dond` sweep function I believe the setpoint dependencies are automatically assigned to any of the `param_meas` parameters and so removing the setpoint is not possible. For example, in a 2D sweep you may be recording time data (along with various other dependent parameters) at each step but it's not necessarily relevant to then see a 2D plot with time when you plot the dataset. Besides, there may be cases where a parameter is legitimately dependent on another but you don't want to generate the plot for it. Therefore, my question is would there be any interest in adding an optional argument to the `plot_dataset` function such that a user could control which parameters are plotted? 

I have tried to implement this such that using e.g. `plot_dataset(dataset, parameters=['y2'])` will only generate the plot for `y2` (default behaviour of `plot_dataset` remains unchanged).

harrietoxinst:
This PR introduces a `parameters` argument to the `plot_dataset` function such that not all dependent parameters have to be plotted if this behaviour is not desired. This closes #6898 . Motivation for this change is further outlined there.

Default behaviour of `plot_dataset` remains unchanged. I have also added the `parameters` argument to the `plot_by_id` function preserving the equivalent behaviour shared by `plot_by_id` and `plot_dataset`. 

To provide an example of when/how this might be used I have added the following under 1.2 Two interleaved 1D sweeps in the Offline Plotting Tutorial notebook: 

image

### DIFF
--- a/tests/dataset/test_plotting.py
+++ b/tests/dataset/test_plotting.py
@@ -5,6 +5,8 @@ import pytest
 from hypothesis import HealthCheck, assume, example, given, settings
 from hypothesis.strategies import data, floats, just, lists, one_of, sampled_from, text
 from matplotlib.collections import QuadMesh
+from matplotlib.lines import Line2D
+from numpy.testing import assert_allclose
 from pytest import FixtureRequest
 
 import qcodes as qc
@@ -395,7 +397,7 @@ def test_plot_dataset_parameters(experiment, request: FixtureRequest, params) ->
 
     dataset = datasaver.dataset
 
-    axes, cbaxes = plot_dataset(dataset=dataset, parameters=params)
+    axes, _ = plot_dataset(dataset=dataset, parameters=params)
 
     # None is default for parameters
     # so if parameters is None all dependent parameters should be plotted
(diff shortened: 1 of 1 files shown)


harrietoxinst:
Thanks for reviewing. I have committed the suggested change and fixed the errors for pre-commit check.

harrietoxinst:
I have looked at the pytest fail which seems unrelated to the changes I've made, consistent with previous iterations of this PR passing that test. Is this an intermittently failing test perhaps? I have tried to re-run this test locally and it passes. Would someone be able to re-run this test please?

jenshnielsen:
@harrietoxinst The failure is indeed unrelated. Would you have the time to write a few tests for this code. tests would probably go to tests/dataset/test_plotting.py I would expect you to be able to write tests by copying and modifying some of the existing tests

jenshnielsen (approved):

PR #6934, opened 2025-03-04
Feat/parameter value change callback

mahmut-oqs:
## Description
This PR introduces the basic infrastructure needed to use QCoDeS in our systems to transition from memory-intensive full snapshots to an efficient parameter-level tracking system. It adds a global callback mechanism to `ParameterBase` that allows individual parameter changes to be logged directly to a time series database (or any other custom logging system).

## Motivation
Our current systems store full snapshots of experiment states, which has led to several issues:
- Memory exhaustion on our production systems
- Performance degradation with large instrument configurations
- Lack of real-time parameter tracking
- Inefficient storage due to redundant copying of unchanged values

This PR is the first step towards a more efficient, database-backed parameter tracking system that will:
- Track only parameter changes rather than full snapshots
- Provide real-time parameter monitoring
- Reduce memory usage and storage requirements
- Serve as a foundation for integrating with time-series databases (e.g., InfluxDB)

## Implementation details
- Added a class variable `global_value_changed_callback: Callable[[ParameterBase, Any], None] | None = None` in `ParameterBase` to support parameter-level tracking.
- Introduced a class method `set_global_value_changed_callback(callback: Callable[[ParameterBase, Any], None] | None)` to configure (or clear) the global callback.
- Modified the `_wrap_set` method to invoke the global callback after each parameter change.
- Included error handling with logging to ensure that failures in the callback do not affect normal operation.
- Designed the changes to have zero impact on existing functionality if no callback is configured.
- Added comprehensive tests to verify the correct behavior of the callback mechanism, including scenarios for basic functionality, validation, error handling, thread safety, and delayed parameter setting.

## Breaking changes
None. This is a backwards-compatible addition that is only activated when the callback is explicitly set.

## Example usage
```python
import time
from qcodes.parameters import Parameter, ParameterBase

def monitor_callback(param: ParameterBase, value: any) -> None:
    """
    Example callback that logs all parameter changes with timestamps.
    Replace the print statement with database-writing logic (e.g., write_to_database)
    to store parameter changes persistently.
    """
    timestamp = time.strftime("%Y-%m-%d %H:%M:%S")
    print(f"{timestamp}: {param.full_name} changed to {value}")

# Set up the global callback
ParameterBase.set_global_value_changed_callback(monitor_callback)

# Create some test parameters
voltage = Parameter("voltage", set_cmd=None, get_cmd=None)
current = Parameter("current", set_cmd=None, get_cmd=None)

# Any changes to these parameters will now trigger the callback
voltage(1.0)  # For example: "2025-03-19 11:29:08: voltage changed to 1.0"
current(0.1)  # For example: "2025-03-19 11:29:08: current changed to 0.1"

# The callback can be disabled by setting it to None
ParameterBase.set_global_value_changed_callback(None)

# Changes no longer trigger the callback
voltage(2.0)  # No output
current(0.2)  # No output
```

*Note:* You can easily swap out `monitor_callback` with a function like `write_to_database` that inserts the parameter change record into a database. Just replace the print statement with your database logic.


mahmut-oqs:

src/qcodes/parameters/parameter_base.py
```diff
@@ -778,6 +780,14 @@ def set_wrapper(value: ParamDataType, **kwargs: Any) -> None:
 
                     self.cache._update_with(value=val_step, raw_value=raw_val_step)
 
+                    if ParameterBase._value_changed_callback is not None:
```
jenshnielsen, line 783: Is there a reason for not doing 
```suggestion
                    if self._value_changed_callback is not None:
```

which would allow a custom calllback in subclasses?

mahmut-oqs: I made my comments in my previous PR (which I emailed you about accidentally closing it). (including naming suggestion/convention) Link: https://github.com/microsoft/Qcodes/pull/6881

So my comment there is as follows:
We had to change our approach to tracking parameter changes. Instead of monitoring each individual instance, we now track changes on a global level. This change came after many discussions with our physicists and engineers. Our systems rely heavily on third-party instruments and drivers, and the old method wasn't practical. It would have required updating thousands of parameter instances across different instruments and submodules - many of which we don't control because we use their drivers.

With global tracking, we can easily keep track of all parameter changes in our system, including third-party instruments, without having to change each one individually.

mahmut-oqs: I have just made the changes based on your suggestions, may I ask how it is now?

Example usage:

```python
def my_callback(param: ParameterBase, value: Any) -> None:
    print(f"Parameter {param.name} changed to {value}")

# Set global callback
ParameterBase.set_global_value_changed_callback(my_callback)

# Clear global callback
ParameterBase.set_global_value_changed_callback(None)
```

jenshnielsen:
@microsoft/qcodes-maintainers Does anyone else want add an opinion on this feature before merge. I think it looks like a generally useful feature that we should be fine with adding.

src/qcodes/parameters/parameter_base.py
jenshnielsen: Thanks that looks good I think

jenshnielsen:
@mahmut-oqs Just waiting to see if there is any other input. Other than that I think this is good but it would be nice with some kind of example + a changelog entry (see the template that shows up when you create the pr) to let others know about this feature

mahmut-oqs:
> @mahmut-oqs Just waiting to see if there is any other input. Other than that I think this is good but it would be nice with some kind of example + a changelog entry (see the template that shows up when you create the pr) to let others know about this feature

I have done both. Please let me know if any changes are needed.

astafan8 (approved):
Thank you for a thorough contribution! Great description, tests, newsfragment -- thank you!

jenshnielsen:
@mahmut-oqs We realized that an instance level callback is also required for some of our use cases. Could you have a look at #6934 which extends your pr with this feature while keeping the feature as you wanted it too.

jenshnielsen:
@mahmut-oqs I merged https://github.com/microsoft/Qcodes/pull/6934 ok to close this one?

mahmut-oqs:
> @mahmut-oqs I merged #6934 ok to close this one?

@jenshnielsen Yes, I am closing the PR right now. Thank you for the feature and the cooperation.

PR #6942, opened 2025-03-06
Replace ConnectionPlus with a subclass of sqlite Connection

jenshnielsen:
Newer versions of python makes it possible to subclass sqlite connection.
This will make it possible to drop wrapt as a direct dependency of qcodes

- [x] Find a better name than CPP
- [x] Fix all type annotations to allow this type
- [x] Deprecated CP and make_connection_plus_from...
- [x] Update any examples

### DIFF
--- a/src/qcodes/dataset/sqlite/connection.py
+++ b/src/qcodes/dataset/sqlite/connection.py
@@ -28,7 +28,7 @@ log = logging.getLogger(__name__)
 )
 class ConnectionPlus(wrapt.ObjectProxy):  # pyright: ignore[reportUntypedBaseClass]
     """
-    Note this is a legacy class. Please refer to ``AtomicConnection``
+    Note this is a legacy class. Please refer to :class:`AtomicConnection`
 
     A class to extend the sqlite3.Connection object. Since sqlite3.Connection
     has no __dict__, we can not directly add attributes to its instance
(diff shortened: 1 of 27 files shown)


src/qcodes/dataset/sqlite/connection.py
```diff
@@ -7,22 +7,29 @@
...
 log = logging.getLogger(__name__)
 
 
+@deprecated(
+    "ConnectionPlus is deprecated. Please use connect to create an AtomicConnection.",
+    category=QCoDeSDeprecationWarning,
+)
 class ConnectionPlus(wrapt.ObjectProxy):  # pyright: ignore[reportUntypedBaseClass]
     """
+    Note this is a legacy class. Please refer to ``AtomicConnection``
```
astafan8, line 31: for fancier docs with clickable links? :)
```suggestion
    Note this is a legacy class. Please refer to :class:`AtomicConnection`
```

src/qcodes/dataset/sqlite/database.py
```diff
@@ -133,18 +135,20 @@ def connect(name: str | Path, debug: bool = False, version: int = -1) -> Connect
 
     Returns:
         connection object to the database (note, it is
-        :class:`ConnectionPlus`, not :class:`sqlite3.Connection`)
+        :class:`AtomicConnection`, not :class:`sqlite3.Connection`)
```
astafan8, line 138:
```suggestion
        :class:`AtomicConnection`, which is a subclass of :class:`sqlite3.Connection`)
```

astafan8 (approved):

PR #7014, opened 2025-03-28
Feat/parameter value change callback

einsmein:
This PR modify a feature implemented in an existing PR https://github.com/microsoft/Qcodes/pull/6934
Specifically, it adds a capability to set an `on_set_callback` on parameter instances, which overwrites the `global_on_set_callback` set on a class.


src/qcodes/parameters/parameter_base.py
```diff
@@ -373,6 +377,7 @@ def _build__doc__(self) -> str | None:
     @property
     def vals(self) -> Validator | None:
         """
+        N
```
jenshnielsen, line 380: Typo?

src/qcodes/parameters/parameter_base.py
```diff
@@ -778,12 +782,27 @@ def set_wrapper(value: ParamDataType, **kwargs: Any) -> None:
...
 
+                    self._call_on_set_callback(val_step)
+
             except Exception as e:
                 e.args = (*e.args, f"setting {self} to {value}")
                 raise e
 
         return set_wrapper
 
+    def _call_on_set_callback(self, value: NumberType | Sized) -> None:
```
mahmut-oqs, line 793: We often use [uncertainties](https://uncertainties.readthedocs.io) package in our experiments/measurements as well as complex numbers. So may I kindly ask if this type hint can be changed to 'Any' here?

jenshnielsen (approved):

src/qcodes/parameters/parameter_base.py
mahmut-oqs: Thank you

PR #7016, opened 2025-03-30
Fix SR830 lock-in amplifier unit error

liuyichao82:
Fixed the issue that the units of the parameters "R", "X", "Y" of the SR830 lock-in amplifier were incorrect.
(should be current "A" but were written as voltage "V")


liuyichao82:

jenshnielsen:
@liuyichao82 Could you provide some documentation that the output unit is A (current)
In the manual I see 

image

Which explicitly says that the output unit of this command is either V or degree. 

Page 5-15 in this pdf https://www.thinksrs.com/downloads/pdfs/manuals/SR830m.pdf

liuyichao82:
Sorry for the late reply. The confusion comes from how the SR830 handles current input: it converts current to voltage internally using a transimpedance gain (e.g., 1 MΩ = 1 V/µA). So even in current mode, OUTP? returns volts, not amps. When I saw that the value I read using the current mode was in voltage, I thought there was something wrong with the program.Maybe we need to add ISRC? to determine the current input mode. Thanks for catching this!
屏幕截图 2025-04-04 225243

jenshnielsen:
Thanks @liuyichao82 I suspect that the qcodes driver does not capture that usecase. If you can find a way to make the driver work better in current mode (while still retaining the functionality in voltage mode) We are happy to merge that.

liuyichao82:
I noticed that there is a `_set_units` function in the driver to set the units correctly. I will find an SR830 machine in a few days to continue testing why this function didn't work when I used it last time.

liuyichao82:
First of all, I need to admit that I made a mistake earlier: when using the SR830 to read, you don't need to consider the transimpedance gain. I believe that the manual's statement that the units returned are always volts is a typo. I tested it with my SR830, and the unit error mentioned earlier will only appear after connecting the instrument and changing input configuration. When connecting the instrument, `self.input_config()` will be executed to ensure that the units are correct at the time. This error can actually be fixed - just ask ISRC? every time you ask OUTP?, but doing so means asking twice for each reading, which will cause additional resource overhead, and the benefit is only to correct the unit error that sometimes occurs. Is this cost-effective?

liuyichao82:
I should add that what I just said about "changing input configuration" refers to manual adjustment through the SR830 front panel rather than using qcodes adjustment. If use the `input_config` parameter to change, the unit will automatically adjust correctly, so I think there is nothing wrong with the program if use qcodes completely, and no changes are needed.

PR #7065, opened 2025-04-11
Replace do0d and do1 with a wrapper around dond

### LINKED ISSUE #7061: Refactor do0d, do1d, do2d to make use of dond internally
As we make it possible to capture more complex data into qcodes datasets it makes sense to minimize the amount of code defining datasets. do0d, do1d and do2d does nothing that dond cannot do but are for historic reasons implemented independently. It makes sense to refactor this code to call dond directly so less code needs to be maintained

jenshnielsen:
Will close #7061 together with #7066

### DIFF
--- a/src/qcodes/dataset/dond/do_0d.py
+++ b/src/qcodes/dataset/dond/do_0d.py
@@ -1,92 +1,71 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, Literal, overload
 
 from opentelemetry import trace
 
-from qcodes import config
-from qcodes.parameters import ParameterBase
+from .do_nd import DondKWargs, dond
 
-from ..descriptions.detect_shapes import detect_shape_of_measurement
-from ..measurements import Measurement
-from ..threading import process_params_meas
-from .do_nd_utils import _handle_plotting, _register_parameters, _set_write_period
+if TYPE_CHECKING:
+    from typing_extensions import Unpack
+
+    from .do_nd_utils import (
+        AxesTupleListWithDataSet,
+        MultiAxesTupleListWithDataSet,
+        ParamMeasT,
+    )
 
 LOG = logging.getLogger(__name__)
 TRACER = trace.get_tracer(__name__)
 
-if TYPE_CHECKING:
-    from ..descriptions.versioning.rundescribertypes import Shapes
-    from ..experiment_container import Experiment
-    from .do_nd_utils import AxesTupleListWithDataSet, ParamMeasT
+
+@overload
+def do0d(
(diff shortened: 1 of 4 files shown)


src/qcodes/dataset/dond/do_2d.py
```diff
@@ -57,21 +34,7 @@ def do2d(
...
     *param_meas: ParamMeasT,
-    set_before_sweep: bool | None = True,
-    enter_actions: ActionsT = (),
-    exit_actions: ActionsT = (),
-    before_inner_actions: ActionsT = (),
-    after_inner_actions: ActionsT = (),
-    write_period: float | None = None,
-    measurement_name: str = "",
-    exp: Experiment | None = None,
-    flush_columns: bool = False,
```
jenshnielsen, line 68: flush_columns and set_before_sweep are not supported by dond yet. Need to figure out how to deal with that

src/qcodes/dataset/dond/do_0d.py
```diff
@@ -5,88 +5,43 @@
...
         The QCoDeS dataset.
 
     """
-    if do_plot is None:
-        do_plot = cast("bool", config.dataset.dond_plot)
-    meas = Measurement(name=measurement_name, exp=exp)
-    if log_info is not None:
-        meas._extra_log_info = log_info
-    else:
-        meas._extra_log_info = "Using 'qcodes.dataset.do0d'"
```
astafan8, line 68: shall we keep track of whether do0d was called epxlicitly by passing do0d's default log_info to the underlying dond? If someone in the commiunity still relies on this, it could be a 'breaking-in-sense' change. Or do we only rely on the better `@TRACER.start_as_current_span("qcodes.dataset.do0d")`?

src/qcodes/dataset/dond/do_1d.py
```diff
@@ -74,106 +42,28 @@ def do1d(
...
-        use_threads: If True measurements from each instrument will be done on
-            separate threads. If you are measuring from several instruments
-            this may give a significant speedup.
-        show_progress: should a progress bar be displayed during the
-            measurement. If None the setting will be read from ``qcodesrc.json``
-        log_info: Message that is logged during the measurement. If None a default
-            message is used.
-        break_condition: Callable that takes no arguments. If returned True,
-            measurement is interrupted.
+        **kwargs: kwargs are the same as for dond and forwarded directly to dond.
```
astafan8, line 49: perhaps use fancy sphinx references so that users can get to documentation of these arguemtns in hte docs quickly via a link ?
```suggestion
        **kwargs: kwargs are the same as for :func:`dond` and forwarded directly to :func:`dond`.
```

astafan8 (approved):
So nice a clean now!

src/qcodes/dataset/dond/do_0d.py
jenshnielsen: Agreed. Done both for do1d and do0d

PR #7109, opened 2025-04-28
Add units and docstring to SGS100 driver

eendebakpt:
Add units and docstrings for a few parameters of the SGS100 driver.

### DIFF
--- a/src/qcodes/instrument_drivers/rohde_schwarz/SGS100A.py
+++ b/src/qcodes/instrument_drivers/rohde_schwarz/SGS100A.py
@@ -164,8 +164,8 @@ class RohdeSchwarzSGS100A(VisaInstrument):
             set_cmd="SOUR:IQ:IMP:LEAK:I {:.2f}",
             get_parser=float,
             vals=vals.Numbers(-10, 10),
-            unit='%',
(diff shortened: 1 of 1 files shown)


src/qcodes/instrument_drivers/rohde_schwarz/SGS100A.py
```diff
@@ -164,6 +164,8 @@ def __init__(
             set_cmd="SOUR:IQ:IMP:LEAK:I {:.2f}",
             get_parser=float,
             vals=vals.Numbers(-10, 10),
+            unit='%',
+            docstring='Carrier offset for I channel in percentage of the peak envelope power.'
```
jenshnielsen, line 168: Could you add the docstring to the line below, so it also shows up in the sphinx docs? Same for the 2 below

eendebakpt: For reference: documentation is rendered here: https://microsoft.github.io/Qcodes/drivers_api/RohdeSchwarz.html#qcodes.instrument_drivers.rohde_schwarz.RohdeSchwarzSGS100A

PR #7114, opened 2025-04-30
fix cryomagnetics4g driver bug.

liuyichao82:
I am trying to fix several bugs in cryomagnetics4g:
1. In the write_raw function of the code, when VisaIOError occurs in the first communication, the code will use `self._RETRY_WRITE_ASK` to determine whether to re-communicate, but I did not see the definition of `self._RETRY_WRITE_ASK` in the code, so I added 
```
self._RETRY_WRITE_ASK = True
self._RETRY_TIME = 1 
```
2. This magnet power supply has three states: Standby, Holding, and Ramping. In the original code, each time the `set_field` function is executed, it will determine whether the magnetic field setting is normal by judging whether the final state is the Holding state. However, he forgot one situation——when the magnetic field is set to 0T, the magnet power supply will enter the Standby state instead of Holding, so I added a judgment to fix this problem.
3. This program waits for the excitation to end by checking the status code `status_byte = int(self.ask("*STB?"))`, when the magnet state is not ramping, it means that the excitation is complete. However, there is a problem with this magnet power supply——even if the magnetic field has been added to the specified value, the state is still ramping and will not change! So I changed to ask the current magnetic field value `self._get_field()` to determine whether the excitation is complete


docs/changes/newsfragments/7114.improved_driver
```diff
@@ -0,0 +1,3 @@
+1. Fixed the problem that there is no specific initialization variable in self._RETRY_WRITE_ASK.
+2. Fix the instrument status recognition problem in self.set_field.
+3. Fixed the ramping state recognition issue in wait_while_ramping.
```
jenshnielsen, line 3: Could you reword this so it's clear which driver this refers to

src/qcodes/instrument_drivers/cryomagnetics/_cryomagnetics4g.py
```diff
@@ -298,17 +301,18 @@ def set_field(self, field_setpoint: float, block: bool = True) -> None:
             self.log.debug("Finished blocking ramp")
             # If we are now holding, it was successful
 
-            if not exit_state.holding:
-                msg = "_set_field({}) failed with state: {}"
-                raise Cryomagnetics4GException(msg.format(field_setpoint, exit_state))
+            if not exit_state.standby:
```
jenshnielsen, line 304:
```suggestion
            if not exit_state.standby and not exit_state.holding:
```
Writing this on one line will make the logic clearer.

liuyichao82: yes! I've updated the code accordingly.

docs/changes/newsfragments/7114.improved_driver
liuyichao82: Sorry about that! I've updated the code in a new commit following your suggestion.

src/qcodes/instrument_drivers/cryomagnetics/_cryomagnetics4g.py
```diff
@@ -307,8 +310,8 @@ def wait_while_ramping(
     ) -> CryomagneticsOperatingState:
         """Waits while the magnet is ramping, checking the status byte instead of field value."""
```
samantha-ho, line 311: This docstring is no longer accurate, given the changes below.

src/qcodes/instrument_drivers/cryomagnetics/_cryomagnetics4g.py
```diff
@@ -355,7 +358,7 @@ def _get_field(self) -> float:
             )
 
         # Return value in Tesla, only converting if necessary
-        if self.units() == "T":
+        if self.units().strip() == "T":
```
samantha-ho, line 361: Nitpick:
If you need to call `.strip()` to make this equality work, I'd prefer to make that part of the `units` parameter's get_parser via something like:
`get_parser = lambda x: str(x).strip()`

But this is a nitpick, as this and line 355 appear to be the only places we use the units parameter internally.

jenshnielsen:
@liuyichao82 @trlemon this currently fails because the test test_set_field_blocking mocks `STB?` when testing blocking ramp. Since this is no longer used the mock driver ends up hanging when this happens and the test eventually times out

PR #7163, opened 2025-05-21
Change DelayedKeyboardInterrupt to work with Spyder 6.0

sldesnoo-Delft:
DelayedKeyboardInterrupt does not work with Spyder 6.0, because Spyder already installs its own handler for SIGINT. Other environment might do the same. 

The modified version checks if the installed interrupt handler is from the DelayedKeyboardInterrupt. If not it installs a new handler.

Automatic test generating KeyboardInterrupt is not straightforward possible.

### DIFF
--- a/src/qcodes/utils/delaykeyboardinterrupt.py
+++ b/src/qcodes/utils/delaykeyboardinterrupt.py
@@ -46,15 +46,19 @@ class DelayedKeyboardInterrupt:
 
     def __enter__(self) -> None:
         is_main_thread = threading.current_thread() is threading.main_thread()
-        is_default_sig_handler = (
-            signal.getsignal(signal.SIGINT) is signal.default_int_handler
-        )
-        if is_default_sig_handler and is_main_thread:
-            self.old_handler = signal.signal(signal.SIGINT, self.handler)
-        elif is_default_sig_handler:
+        if not is_main_thread:
             log.debug(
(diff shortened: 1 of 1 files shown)


jenshnielsen:
@sldesnoo-Delft Sorry for the slow response. If I understand this correctly the idea of this pr is to make sure that the signal handler is always replaced? Originally I wrote the code such that it would only replace the handler if the default one was in place. This was an attempt to not break other libraries handlers but perhaps that is not the best approach?

sldesnoo-Delft:
Yes, indeed. The problem is that Spyder 6 already replaces the interrupt handler.
In our lab we use the DelayedKeyboardInterrupt to protect communication with hardware. It prevents that the request/response process gets interrupted resulting in failures on the next request.
With Spyder 6 we suddenly saw strange communication failures after keyboard interrupts.
Other environments might also replace the interrupt handler.

My current solution is to use our own simple version of the DelayedKeyboardInterrupt to protect the communication. But I think it would be nice if the QCoDeS implementation would also work in these situations.

jenshnielsen:
@sldesnoo-Delft Makes sense. What do you think about implementing your pr as it currently is (fixing the small type checking issues and precommit) but then adding a config option to the qcodes config to allow users to disable this in case they really want the other handler?

sldesnoo-Delft:
That sounds like a good addition. It's good to have an option to get the old behaviour.

PR #7233, opened 2025-06-13
Remove deprecated modules

jenshnielsen:
This removes most deprecated modules, classes and functions from QCoDeS that have been deprecated for at least 2 releases with exception of the convenience methods on Parameter since issues have been raised about these.


tests/dataset/test_sqlite_connection.py
```diff
@@ -242,102 +183,11 @@ def test_two_nested_atomics() -> None:
     assert atomic_in_progress == atomic_conn_2.atomic_in_progress
 
 
-@pytest.mark.parametrize(
-    argnames="create_conn_plus",
-    argvalues=(make_connection_plus_from, ConnectionPlus),  # pyright: ignore[reportDeprecated]
-    ids=("make_connection_plus_from", "ConnectionPlus"),
-)
-def test_that_use_of_atomic_commits_only_at_outermost_context(
```
astafan8, line 250: do we have a similar test for atomic connection?

astafan8 (approved):
beautiful cleanup!

tests/dataset/test_sqlite_connection.py
jenshnielsen: Yes when AtomicConnection was added this test was mirrored for that as `test_that_use_of_atomic_commits_only_at_outermost_context_atomic_connection` 

We could consider renaming the duplicated tests back to the original name now that it's no longer ambiguous

jenshnielsen: renamed `test_that_use_of_atomic_commits_only_at_outermost_context_atomic_connection` to `test_that_use_of_atomic_commits_only_at_outermost_context`

PR #7242, opened 2025-06-16
Run all test steps in CI and fail at the end.

jenshnielsen:


jenshnielsen:
This has been manually tested by inserting a type checking error manually which I will now revert
image

PR #7256, opened 2025-06-26
InterDependencies_ using NetworkX DiGraph

samantha-ho:

### DIFF
--- a/src/qcodes/dataset/descriptions/dependencies.py
+++ b/src/qcodes/dataset/descriptions/dependencies.py
@@ -6,13 +6,15 @@ which parameters depend on each other is handled here.
 
 from __future__ import annotations
 
-from copy import deepcopy
-from typing import TYPE_CHECKING, Any
+from collections import defaultdict
+from itertools import chain, product
+from typing import TYPE_CHECKING, Any, Self, cast
 
+import networkx as nx
 import numpy as np
 import numpy.typing as npt
 
-from .param_spec import ParamSpec, ParamSpecBase
+from .param_spec import ParamSpecBase
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Sequence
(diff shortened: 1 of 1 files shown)


src/qcodes/dataset/descriptions/dependencies.py
```diff
@@ -50,262 +52,176 @@ def __str__(self) -> str:
...
-        standalones = [self._paramspec_to_id[ps] for ps in self.standalones]
-        output: InterDependencies_Dict = {
-            "parameters": parameters,
-            "dependencies": dependencies,
-            "inferences": inferences,
-            "standalones": standalones,
-        }
-        return output
-
-    def _empty_data_dict(self) -> dict[str, dict[str, npt.NDArray]]:
```
samantha-ho, line 275: This method has not been converted. I can simply copy it as is into the new implementation, but my instinct is that it doesn't belong here in the first place. 

It is only used in the DataSetCache's `prepare` method, and it seems like it would be better to create the empty arrays there than here.

pyproject.toml
```diff
@@ -28,6 +28,7 @@ dependencies = [
     "ipykernel>=6.12.0", # implicitly required by ipywidgets >=8.0.5
     "jsonschema>=4.9.0",
     "matplotlib>=3.6.0",
+    "networkx>=3.1",
```
jenshnielsen, line 31: We should also add types-networkx to the test dependencies

jenshnielsen:
@samantha-ho From a highlevel this looks good and I support the idea. Unless you think otherwise I don't think a detailed review makes sense until its out of preview?

samantha-ho:
> @samantha-ho From a highlevel this looks good and I support the idea. Unless you think otherwise I don't think a detailed review makes sense until its out of preview?
Agreed. A detailed review can wait a bit longer

src/qcodes/dataset/descriptions/dependencies.py
samantha-ho: Decided to copy/paste as it was to keep the scope of this PR limited.

src/qcodes/dataset/descriptions/dependencies.py
```diff
@@ -63,163 +53,153 @@ def __init__(
...
-        if link_error is not None:
-            error, mssg = link_error
-            raise error(mssg)
-
+        self._graph: nx.DiGraph[str] = nx.DiGraph()
+        self.add_dependencies(dependencies)
+        self.add_inferences(inferences)
+        self.add_standalones(standalones)
+
+    def add_paramspecs(self, paramspecs: list[ParamSpecBase]) -> None:
```
jenshnielsen:
```suggestion
    def add_paramspecs(self, paramspecs: Sequence[ParamSpecBase]) -> None:
```
Since this doesn't modify the input the type should be immutable

src/qcodes/dataset/descriptions/dependencies.py
```diff
@@ -6,46 +6,36 @@
...
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable, Sequence
+    from collections.abc import Sequence
 
     from .versioning.rundescribertypes import InterDependencies_Dict
 
 ParamSpecTree = dict[ParamSpecBase, tuple[ParamSpecBase, ...]]
-ParamNameTree = dict[str, tuple[str, ...]]
+ParamNameTree = dict[str, list[str]]
```
jenshnielsen, line 26: Is there a good reason this needs to be a mutable type.

src/qcodes/dataset/descriptions/dependencies.py
```diff
@@ -63,163 +53,153 @@ def __init__(
...
+                and self.graph.nodes[paramspec.name]["value"] != paramspec
+            ):
+                raise ValueError(
+                    f"A ParamSpecBase with name {paramspec.name} already exists in the graph and\n"
+                    f"{paramspec} != {self.graph.nodes[paramspec.name]['value']} "
+                )
+            self._graph.add_node(paramspec.name, value=paramspec)
+
+    def _add_interdeps_by_type(
+        self, links: list[tuple[ParamSpecBase, ParamSpecBase]], type: str
```
jenshnielsen:
```suggestion
        self, links: Sequence[tuple[ParamSpecBase, ParamSpecBase]], type: str
```

jenshnielsen:
@samantha-ho Looks good. Left a few inline comments for types that could be a bit stricter. Copilot poininted out what looks like some minor typos that are probably worth fixing

src/qcodes/dataset/descriptions/dependencies.py
samantha-ho: The `InterDependencies_Dict` class from dataset.descriptions.versioning.rundescribertypes.py types these as `list[str]`. Rather than do a cast internally or modify that definition, I chose to keep everything consistent.

PR #7262, opened 2025-06-30
Deprecate legacy alias modules

jenshnielsen:
These are aliases that were added for backwards compatibility. Now is the time to deprecate them.

### DIFF
--- a/docs/changes/newsfragments/7262.breaking
+++ b/docs/changes/newsfragments/7262.breaking
@@ -1,0 +1,19 @@
+The following modules have been deprecated in this release and will be removed in a future version:
+
+- ``qcodes.instrument.delegate.grouped_parameter``
+- ``qcodes.instrument.function``
+- ``qcodes.instrument.group_parameter``
+- ``qcodes.instrument.specialized_parameters``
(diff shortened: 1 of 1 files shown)


src/qcodes/utils/dataset/doNd.py
```diff
@@ -25,42 +25,6 @@
 from qcodes.dataset.dond.sweeps import AbstractSweep, ArraySweep, LinSweep, LogSweep
 from qcodes.dataset.plotting import plot_and_save_image as plot
```
astafan8, line 27: no deprecation warning for this module?

astafan8 (approved):

src/qcodes/utils/dataset/doNd.py
jenshnielsen: The module above is deprecated (`utils.dataset`) and therefor also this one.

jenshnielsen:
Seems to have hit a race condition in github. This is merged but was somehow still in the merge queue.

jenshnielsen:
While it looks like this is not merged it has actually been merged so closing

PR #7305, opened 2025-07-16
Add ipycytoscape plotting for InterDependencies_ objects

samantha-ho:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,6 +114,9 @@ docs = [
 refactor = [
     "libcst>=1.2.0"
 ]
+live_plotting = [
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -114,6 +115,9 @@ docs = [
 refactor = [
     "libcst>=1.2.0"
 ]
+live_plotting = [
+    "ipycytoscape>=1.2.2"
+]
```
jenshnielsen, line 120: I think we should call this something like `jupyer` and then move the dependency on ipykernel and ipywidgets down here. The latter should probably not be done in this pr. That way we can give users the option to pull in the juypyter deps if they want to

PR #7307, opened 2025-07-17
dond enhancement: get after set only when the parameter is set

jpdehollain:
When running a dond measurement with a sweep having the `get_after_set` flag, the loop will get the parameter value after every iteration of the loop, including when the sweep parameter is not being set because the new event is from an inner parameter. 
This is inefficient and can cause inconsistencies in the dataset structure is the parameter value changes slightly while the inner loop is running.

This PR improves the way that `get_after_set` is handled, by only getting the parameter value if the value was set in the same event.


jpdehollain:

jpdehollain:
I modified `test_dond_get_after_set` to show the issue that is being addressed, by turning it into a 2D sweep. Previously, the outer parameter `a` would have `set_counter == n_points` and `get_counter == n_points ** 2`, with this PR they're the same.

Since this modified test function tests both cases with and without `get_after_set`, I removed `test_dond_no_get_after_set`.

jenshnielsen:
Thanks @jpdehollain I think this makes sense and probably better reflects the name of that argument. Could you have a look at the failing precomit hook and then I think this should be ready to merge

jpdehollain:
Thanks @jenshnielsen, I have fixed an error in the test I modified and ran pre-commit, hopefully all the checks will pass now

PR #7335, opened 2025-07-25
Bump minimum versions of dependencies

jenshnielsen:
Two reasons.

* CI is 3 times slower with the minimum dependencies. Debug if this resolves it.
* Trying to follow spec0000 https://scientific-python.org/specs/spec-0000/


jenshnielsen:
It seems the issue was coverage https://github.com/microsoft/Qcodes/pull/7382 fixes that and this pr is probably not required

PR #7346, opened 2025-07-29
Allow Parameter subclasses to define and unpack their own implicit dependencies and inferences

samantha-ho:
In the `Measurement.register_parameter` and `DataSaver.add_result` methods, the explicit Parameter type is used to register and unpack implicit dependencies for the ParameterWithSetpoints. This pattern requires that new sub-classes with different internal dependencies or inferences must exist within Qcodes and these method must be modified to handle them on a case-by-case basis.

In order to support more complicated parameter interdependencies at higher-level abstractions, this PR modifies `register_parameter` and `add_result` to use new ParameterBase interfaces to perform registration and unpacking. New subclasses will then be able to use these new interfaces to register and unpack more complicated patterns.

In detail:
ParameterBase now defines new properties and methods:
- `depends_on` is an order-preserving set, ParameterSet, of other ParameterBase objects it depends_on. These will be automatically added to the InterDependencies dependency tree upon registration. (Similar to the behavior of setpoints with ParameterWithSetpoints)
- `is_controlled_by` is a set of parameters which use the `inferred_from` or `basis` relationship
- `has_control_of` is a set with the reverse direction of the `is_controlled_by` relationship, used to find related parameters during registration
- `param_spec` is a new property that returns the `ParamSpecBase` of a parameter. It infers the `paramtype` from the associated Validators and defaults to "numeric"
- `paramtype` is used by `param_spec` and its setter ensures that a Parameter's validators match its data representation in SQLite

`Measurement.register_parameter` has been refactored to add a new `_self_register_parameter` method:
- It automatically adds `dependency` and `inference` links in the InterDependencies based on the associated ParameterBase properties
- Recursively calls itself on parameter objects defined as dependencies, inferences, and reverse-inferences to ensure that all related parameters are registered
- Adds parameters to the private `_registered_parameters` attribute. This is then passed to the DataSaver via the Runner
- Setting `paramtype` kwarg to `register_parameter` now modifies the Validator of the Parameter to match

`DataSaver.add_result` has also been refactored:
- Now allows results added via ParameterBase objects to "unpack" themselves (ie, to include related values)
- Duplicate results are compared for approximate equality. An error is raised if multiple values are added for a parameter that do not match (implicit vs. explicit argument to add_result).
- BREAKING CHANGE: Previously, a ParameterWithSetpoints whose setpoints values were explicitly added in `add_result` would use the explicit version. Now, an error is raised if the explicit values are not within some tolerance of the internal values (as with other duplication).


src/qcodes/parameters/parameter.py
```diff
@@ -15,6 +15,7 @@
 if TYPE_CHECKING:
     from collections.abc import Callable
 
+    from qcodes.dataset.descriptions.param_spec import ParamSpecBase
```
jenshnielsen: I think we have previously avoided this since we did not want the parameters module to depend on the dataset. Perhaps better solution would be to move ParamSpecBase to the parameters module (the main issue with that is that the ParamSpecBase contains a type that maps to sqlite data types)

src/qcodes/parameters/parameter_base.py
```diff
@@ -1144,6 +1160,82 @@ def underlying_instrument(self) -> InstrumentBase | None:
...
+            case Strings():
+                paramtype = "text"
+            case ComplexNumbers():
+                paramtype = "complex"
+            case _:
+                paramtype = "numeric"
+
+        return ParamSpecBase(
+            name=self.register_name,
+            paramtype=paramtype,
```
jenshnielsen: In the long run I think we should change the name of this property to something like sqlite_storage_class ... but probably not in this pr

src/qcodes/parameters/parameter_base.py
```diff
@@ -1144,6 +1160,82 @@ def underlying_instrument(self) -> InstrumentBase | None:
...
+            label=None,
+            unit=None,
+        )
+
+    @property
+    def paramtype(self) -> str:
+        return self.param_spec.type
+
+    @paramtype.setter
+    def paramtype(self, paramtype: str) -> None:
```
jenshnielsen, line 1190: Lest make the type a literal Literal["array", "text", "complex", "numeric"]

src/qcodes/parameters/parameter_base.py
```diff
@@ -1144,6 +1160,82 @@ def underlying_instrument(self) -> InstrumentBase | None:
...
+        return ParamSpecBase(
+            name=self.register_name,
+            paramtype=paramtype,
+            label=None,
+            unit=None,
+        )
+
+    @property
+    def paramtype(self) -> str:
+        return self.param_spec.type
```
jenshnielsen, line 1187: Since this type is sqlite specific I think I would rather avoid exposing it here

src/qcodes/dataset/measurements.py
```diff
@@ -941,61 +1112,43 @@ def register_parameter(
...
-                basis,
-                paramtype,
-            )
-        else:
-            raise RuntimeError(
-                f"Does not know how to register a parameter of type {type(parameter)}"
-            )
-        self._registered_parameters.append(parameter)
+        match parameter:
+            case Parameter() | ParameterWithSetpoints():
```
jenshnielsen: Can we push this further such that the self registration path is used for all Subclasses of ParameterBase that are not ArrayParameter, MultiParameter, GroupedParameter ?

src/qcodes/parameters/parameter_base.py
samantha-ho: The reason it's exposed is because `Measurement.register_parameter` has an optional kwarg to assign a specific paramtype at registration. This functionality is used extensively in testing, and can conflict with the self-reported ParamSpecBase's paramtype.

Thus, this is a compromise. The property exists so that the kwarg can be compared, and the setter exists so that it can be modified at registration (provided it does not conflict with the Validator).

src/qcodes/parameters/parameter_base.py
samantha-ho: That was my first instinct as well, but I ran into lower level typing incompatibilities where `paramtype` is typed as `str`. Rather than resolve those conflicts, I relaxed this typing.

I think I'd rather convert to a Literal in a follow-up PR than in this one.

src/qcodes/parameters/parameter_base.py
samantha-ho: I agree it would be nice to have a more descriptive name, and also that it is beyond the scope of this PR.

src/qcodes/parameters/parameter.py
samantha-ho: I can understand the reasoning, but this PR ties `ParamSpecBase` much more closely to `ParameterBase`. 

If we really want to maintain a separation, I think we'd need another layer at the `register_parameter`/`InterDependencies_` level to store the `ParameterBase` versions of the dependencies before converting everything to `ParamSpecBase`. That seems like a much bigger change than I'm willing to pursue right now.

Shall we agree to leave it as is for the present, and revisit the structure later?

jenshnielsen: No importing the dataset module into the parameters module is unacceptable since it creates a circular import. Moving ParamSpecBase to the parameters module as is should be fine thou

src/qcodes/parameters/parameter_base.py
jenshnielsen: Ok, i dont follow why that cannot be done via self.param_spec.type thou?

src/qcodes/parameters/parameter.py
jenshnielsen: This import is not a big issue since its behind a if TYPE_CHECKING but the one in parameter_base is problematic

src/qcodes/parameters/parameter_base.py
```diff
@@ -1211,3 +1303,104 @@ def __call__(self) -> ParamDataType:
         It is recommended to use ``parameter.cache()`` instead.
         """
         return self.cache()
+
+
+# Does not implement __hash__, not clear it needs to
+class ParameterSet:  # noqa: PLW1641
```
jenshnielsen: I think it would be good to inherit from the `collections.abc.MutableSet` that will help ensure that this implementes the api consistently with the official python set api. The channeltuple/list has several inconsistencies due to not doing this. I would like to avoid this in the here if possible.

src/qcodes/parameters/parameter_base.py
samantha-ho: The way this is currently implemented, the `self.param_spec.type` depends entirely on the Parameter's Validator. There is no internal ParamSpecBase attribute that holds information about the type -- it is constructed on read.

I suppose I could change this so that a default ParamSpecBase is created if one does not already exist, and then allow it to be further modified, but without making this change, modifying `self.param_spec.type` directly does not ensure that the param_spec type stays modified.

The other alternative would be to move the paramtype/validator check to `register_parameter`, but that decouples the values of two intimately related properties. This is how things are currently structured, but I think it'd be an error to allow (for example), a Parameter with `ComplexValues()` validation to be written to the database as `Numeric`.

src/qcodes/parameters/parameter_base.py
samantha-ho: Done.

src/qcodes/parameters/parameter.py
samantha-ho: Merged your move_paramspec branch into this one.

src/qcodes/parameters/parameter_base.py
samantha-ho: After some re-thinking, I decided to implement the second plan above. ParameterBase generates a ParamSpecBase default if none has already been created, but in order to keep the warning on changing to a paramtype that doesn't match a validator, I needed to keep the paramtype attribute.

src/qcodes/parameters/parameter_with_setpoints.py
```diff
@@ -174,24 +200,27 @@ def expand_setpoints_helper(
...
+    #     setpoint_params.append(setpointparam)
+    #     setpoint_data.append(these_setpoints)
+    # output_grids = np.meshgrid(*setpoint_data, indexing="ij")
+    # for param, grid in zip(setpoint_params, output_grids):
+    #     res.append((param, grid))
+    # if results is None:
+    #     data = parameter.get()
+    # else:
+    #     data = results
+    # res.append((parameter, data))
```
jenshnielsen: I guess we can remove this commented out code?

src/qcodes/parameters/parameter_with_setpoints.py
```diff
@@ -154,6 +155,31 @@ def validate(self, value: ParamDataType) -> None:
             self.validate_consistent_shape()
         super().validate(value)
 
+    @property
+    def depends_on(self) -> ParameterSet:
+        return ParameterSet(self.setpoints)
+
+    def unpack_self(self, value: ValuesType) -> list[tuple[ParameterBase, ValuesType]]:
```
jenshnielsen, line 162:
```suggestion
    def _unpack_self(self, value: ValuesType) -> list[tuple[ParameterBase, ValuesType]]:
```

Could this be private?

src/qcodes/dataset/measurements.py
```diff
@@ -122,11 +129,52 @@ def __init__(
...
+        else:  # param_or_str is a str
+            candidate_params = [
+                param
+                for param in self._registered_parameters
+                if param.register_name == result_tuple[0]
+            ]
+            if len(candidate_params) > 1:
+                raise ValueError(
+                    "More than one parameter matched this name"
+                )  # TODO: Expand this
```
jenshnielsen: Should anything be done about this todo before merge?

src/qcodes/dataset/measurements.py
```diff
@@ -122,11 +129,52 @@ def __init__(
...
+            elif len(candidate_params) < 1:
+                raise ValueError("No matching parameters")
+            return (candidate_params[0], result_tuple[1])
+
+    def add_result(self, *result_tuples: ResType) -> None:
```
jenshnielsen, line 178: Changing the name here is technically an api break but I guess its very unlikely that anyone pass this arg with an explicit name

src/qcodes/parameters/parameter_with_setpoints.py
samantha-ho: It gets called by `datasaver.add_result`, so it's not private? Also, this is an override of the `ParameterBase` method by this name.

src/qcodes/dataset/measurements.py
samantha-ho: Fair. I found the original naming scheme confusing, and this was part of my attempt to clean it up.

I am willing to move forward with the api break.

src/qcodes/parameters/parameter_with_setpoints.py
samantha-ho: On further thought, I suppose you mean that it's not "public API", and the underscore indicates that more clearly. I could maybe be persuaded to think of it as akin to a Friend scope from other languages.

jenshnielsen: Yes in a open source package we want to clearly communicate to users outside of the package that they should not call this function. That is IMHO way more important than any class private concern since that leaves us free to make changes to the method while only changing things in QCoDeS it self.

src/qcodes/dataset/measurements.py
jenshnielsen: I am ok with that but lets make sure we document it in the changelog

jenshnielsen:
```suggestion
    def add_result(self, *result_tuples: ResType, /) -> None:
```
It could also make sense to make this positional only so the user cannot pass the argument name

docs/changes/newsfragments/7346.breaking
```diff
@@ -0,0 +1,10 @@
+Registration and Unpacking interfaces created in ParameterBse
```
jenshnielsen, line 1: Since this contains braking changes I think it should be .breaking (or at least that part of it)

docs/examples/Parameters/Parameter_defined_InterDependencies.ipynb
```diff
@@ -0,0 +1,281 @@
...
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.11.8"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 5
+}
```
jenshnielsen, line 281: Nice clear example :)

src/qcodes/parameters/parameter_base.py
```diff
@@ -1144,6 +1161,90 @@ def underlying_instrument(self) -> InstrumentBase | None:
...
+            logging.warning(
+                f"Tried to set a new paramtype {paramtype}, but this parameter already has paramtype {self.paramtype} which does not match"
+            )
+        self.param_spec.type = paramtype
+
+    @property
+    def depends_on(self) -> ParameterSet:
+        return self._depends_on
+
+    # TODO: Decide if this should return a frozenset to make it somewhat harder to mutate accidentally
```
jenshnielsen: Is this todo something we want to address in this pr?

src/qcodes/parameters/parameter_with_setpoints.py
jenshnielsen: On the other hand we probably expect downstream implementers to override this method so it should probably be public. I am ok with leaving it as is

jenshnielsen (approved):
Great work! This is a very nice improvement. Only left a few very minor suggestions

src/qcodes/parameters/parameter_base.py
samantha-ho: I think I'll remove this. We may eventually want to revisit, but it's pretty useful to be able to do 
`parameter.depends_on.add(dependency)`. Whether that is too dangerous to allow isn't obvious to me.

src/qcodes/dataset/measurements.py
samantha-ho: That suggestion creates a syntax error: '/' parameter must appear before '*' parameter. So I guess we can't do that.

PR #7360, opened 2025-08-04
Added Keysight P5002B driver and corrected power range for Keysight P5004B

HolmPhilip:


astafan8 (approved):
Thank you so much for the contribution!

PR #7398, opened 2025-08-18
Add Driver for Cryomagnetics TM-620 Cryogenic Temperature Monitor

trlemon:
### Summary
- Added driver for Cryomagnetics TM-620
- Added simulator YAML configuration
- Added tests for TM-620 driver
- Added notebook containing usage examples for the TM-620

### DIFF
--- a/src/qcodes/instrument_drivers/cryomagnetics/_TM620.py
+++ b/src/qcodes/instrument_drivers/cryomagnetics/_TM620.py
@@ -28,7 +28,7 @@ class CryomagneticsModelTM620(VisaInstrument):
         self.shield: Parameter = self.add_parameter(
             name="shield",
             unit="K",
-            get_cmd=self.get_A,
+            get_cmd=self._get_A,
             get_parser=float,
             docstring="55K Shield Temp",
         )
@@ -37,7 +37,7 @@ class CryomagneticsModelTM620(VisaInstrument):
         self.magnet: Parameter = self.add_parameter(
             name="magnet",
             unit="K",
-            get_cmd=self.get_B,
+            get_cmd=self._get_B,
             get_parser=float,
             docstring="4K Magnet Temp",
         )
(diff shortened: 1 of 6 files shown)


trlemon:

src/qcodes/instrument_drivers/cryomagnetics/_TM620.py
```diff
@@ -0,0 +1,116 @@
...
+        self.magnet: Parameter = self.add_parameter(
+            name="magnet",
+            unit="K",
+            get_cmd=self._get_B,
+            get_parser=float,
+            docstring="4K Magnet Temp",
+        )
+        """4K magnet temperature"""
+
+        self.add_function("remote", call_cmd="REMOTE")
```
jenshnielsen: We are trying to remove the use of function since they add little value over a method and are not typed so I would prefer adding this as a method

src/qcodes/instrument_drivers/cryomagnetics/_TM620.py
```diff
@@ -0,0 +1,116 @@
...
+        """Extract floating point number from the instrument output string.
+
+        Args:
+            output: the string returned from the instrument.
+
+        Returns:
+            parsed string containing extracted floating point number.
+
+        """
+        pattern = r"[0-9]+\.[0-9]+"
```
jenshnielsen: Since this is used on every read I think it would be nice to compile the regex for a small performance improvement https://docs.python.org/3/library/re.html#re.compile

jenshnielsen (approved):

src/qcodes/instrument_drivers/cryomagnetics/_TM620.py
trlemon: Converted this to a method as requested.

src/qcodes/instrument_drivers/cryomagnetics/_TM620.py
trlemon: I made this change as requested.

PR #7405, opened 2025-08-19
Fix Cryomagnetics 4G Driver Bugs

trlemon:
### Background

This PR extends an initial PR (closes #7114) that was opened by @liuyichao82. Both his changes and mine are reflected in this PR. 

### Summary of Issues Resolved

As noted in the `newsfragments`, this PR makes the following changes:

1. Added missing initialization of `self._RETRY_WRITE_ASK` and `self._RETRY_TIME` in the `write_raw` method to handle communication retries on VisaIOError.
2. Updated `set_field` to correctly recognize Standby as a valid final state when setting the magnetic field to 0 T.
3. Modified `wait_while_ramping` to check the actual magnetic field value via `self._get_field()` instead of relying on the status byte, since the instrument may stay in the Ramping state even after reaching the target field.
4. Fixed unit parsing to handle unexpected trailing characters in the instrument's response. (e.g., "\r").
5. Fixed get_parser for `Vmag`, `Iout`, and `Vout` to remove units and return float.
6. Added timeout condition to `wait_while_ramping` to prevent infinite loop. Will trigger Exception if timeout occurs.
7. Fixed conversion calculations in `_set_rate` and `_get_rate`.
8. Added function `get_rates` to return rates for all ranges in Tesla/min as a dictionary.
9. Modified `_set_rate` to update all ranges (0-4) on the instrument (picks min value between set rate and maximum current limit for each range)
10. Added small sleep (0.1 seconds) between write calls during `max_current_limit` initialization.
11. Fixed/added unit tests to accurately reflect code changes.
12. Modified `set_field()` to allow users to control the strictness of `wait_while_ramping()` timeout condition (1e-4 was too precise and was triggering false timeouts).


docs/changes/newsfragments/7405.improved_driver
```diff
@@ -0,0 +1,12 @@
+1. Added missing initialization of `self._RETRY_WRITE_ASK` and `self._RETRY_TIME` in the `write_raw` method to handle communication retries on VisaIOError.
```
jenshnielsen, line 3: Can we add a header/title clarifying that this is about the Cryomagnetics driver. Otherwise I don't think that is obvious in the changelog

trlemon: Done.

Issue #7062, opened 2025-04-11
Inferred data is not exported by get_parameter_data, to_pandas and to_xarray


jenshnielsen:
> There is also a further issue, that `get_parameter_data` will not display parameter data for `inferred_from` parameters by default. Since this function is used in `to_xarray_dataset` this also means that `inferred_from` data will not appear in the associated xarray dataset even if it is captured in the Qcodes dataset
> 
> This happens because `get_parameter_data` uses `Interdependencies_.non_dependencies` to identify `top_level` params, which will not include anything from the `inferences` dict. 

 _Originally posted by @samantha-ho in [#7051](https://github.com/microsoft/Qcodes/issues/7051#issuecomment-2790295443)_

jenshnielsen:
Export via get_parameter_data was fixed via the fix for #7051

jenshnielsen:
see #7522 for remaining work when the shape is not known

PR #7413, opened 2025-08-21
Export data directly to xarray when shape is known

jenshnielsen:
When the shape is known up front there is no need to go via pandas multi index

TODO
- [x] Cleanup nested if statements
- [x] Cover this in tests
- [x] Some form of validation (Only export if the shape of data matches the metadata shape)


jenshnielsen:
This is the first step towards resolving #7062

astafan8 (approved):

PR #7432, opened 2025-08-27
Export inferred parameters related to dependencies

jenshnielsen:
Export inferred to and from parameters related to setpoints with the same shape as the setpoints when exporting to xarray and the shape is known. Export inferred related to data parameters as data parameters.

TODO
- [x] Add docs and changelog

### DIFF
--- a/src/qcodes/dataset/data_set.py
+++ b/src/qcodes/dataset/data_set.py
@@ -97,6 +97,7 @@ from .exporters.export_to_pandas import (
 from .exporters.export_to_xarray import (
     load_to_xarray_dataarray_dict,
     load_to_xarray_dataset,
+    load_to_xarray_dataset_dict,
     xarray_to_h5netcdf_with_complex_numbers,
 )
 from .subscriber import _Subscriber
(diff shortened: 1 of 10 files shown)


src/qcodes/dataset/exporters/export_to_xarray.py
```diff
@@ -162,20 +162,63 @@ def _xarray_data_array_direct(
...
+        name: (tuple(dep_axis.keys()), subdict[name])
+    }
+    data_vars.update(extra_data_vars)
+
+    ds = xr.Dataset(data_vars, coords=coords)
+    da = ds[name]
+    if len(extra_data_vars) > 0:
+        # stash extra data vars to be added at dataset assembly time
+        # mapping: var_name -> (dims_tuple, numpy array)
+        da.attrs["_qcodes_extra_data_vars"] = extra_data_vars
```
astafan8: Opting for using a private temporary datarray attrbiute as opposed to returning those extras explicitly from the method? It's a matter of taste, but i'd suggest a more explicit way of returning these extras-to-be-added-to-dataset-later, i'd be a few more lines of code, but hopefully more readable due to explicit-ness?

astafan8 (approved):
pinging @samantha-ho for review as well :)

src/qcodes/dataset/exporters/export_to_xarray.py
jenshnielsen: Agreed, Actually we should investigate if we can return a Dataset and merge a collection of datasets rather than dataarrays.

src/qcodes/dataset/exporters/export_to_xarray.py
```diff
@@ -148,42 +142,85 @@ def _xarray_data_array_from_pandas_multi_index(
...
+    ds = xr.Dataset(data_vars, coords=coords)
+    # da = ds[name]
+    # if len(extra_data_vars) > 0:
+    #     # stash extra data vars to be added at dataset assembly time
+    #     # mapping: var_name -> (dims_tuple, numpy array)
+    #     da.attrs["_qcodes_extra_data_vars"] = extra_data_vars
+    return ds
 
 
 def load_to_xarray_dataarray_dict(
```
jenshnielsen, line 232: Once we deprecate the methods using this we should also deprecate this (and replace with load_to_xarray_dataset_dict

src/qcodes/dataset/data_set.py
```diff
@@ -1020,7 +1020,7 @@ def to_xarray_dataarray_dict(
             self, data, use_multi_index=use_multi_index
         )
 
-        return datadict
+        return {key: val[key] for key, val in datadict.items()}
```
jenshnielsen: I think we should deprecate these to_xarray_dataarray_dict and replace with to_xarray_dataset_dict, so that the exported datasets for each measured parameter can contain inferred coordinates.

PR #7442, opened 2025-08-28
Remove warning about partially overlapping coordinates

### LINKED ISSUE #7419: Should we remove warning about setpoints not being equal from xarray export?
Currently we warning with `Independent parameter setpoints are not equal. Check concatenated output carefully.` ... 
when exporting to xarray but consider what actually happens (taken from test_same_setpoint_warning_for_df_and_xarray)

Image

The only think that happens is that the 2 coordinated gets a _1 and _2 prefix

jenshnielsen:
Resolves #7419


astafan8 (approved):
newsfragment needed?

PR #7454, opened 2025-09-01
Improved test coverage for exporting dataset with multiple setpoints

jenshnielsen:
This tries to ensure that we cover in test

* Known shape vs not known shape
* Pandas export vs xarray export
* Exports where 2 different setpoints are scanned in parallel vs using the same setpoints but with different values for different measured parameters.

### DIFF
--- a/docs/changes/newsfragments/7442.improved
+++ b/docs/changes/newsfragments/7442.improved
@@ -1,2 +1,4 @@
-Exporting datasets to XArray no longer warns if two or more variables have setpoints of the same name but with different values.
-XArray automatically takes care of creating coordinates with names postfixed by `_n` to disambiguate these.
+Exporting datasets to XArray no longer warns if two or more variables are exported with different setpoints.
+No functionality of the export has changed
+If the setpoints have the same name missing values will be replaced by NaN and if different setpoints
+are used the exporter will automatically handle merging coordinates so each data variable is assigned its own coordinates.
--- a/tests/dataset/conftest.py
+++ b/tests/dataset/conftest.py
@@ -271,12 +271,15 @@ def multi_dataset(experiment, request: FixtureRequest):
         datasaver.dataset.conn.close()
 
 
-@pytest.fixture(scope="function", params=["array"])
+@pytest.fixture(scope="function", params=[True, False])
 def different_setpoint_dataset(experiment, request: FixtureRequest):
     meas = Measurement()
     param = Multi2DSetPointParam2Sizes()
 
-    meas.register_parameter(param, paramtype=request.param)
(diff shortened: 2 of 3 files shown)


docs/changes/newsfragments/7442.improved
```diff
@@ -1,2 +1,4 @@
-Exporting datasets to XArray no longer warns if two or more variables have setpoints of the same name but with different values.
-XArray automatically takes care of creating coordinates with names postfixed by `_n` to disambiguate these.
+Exporting datasets to XArray no longer warns if two or more variables are exported with different setpoints.
+No functionality of the export has changed
+If the setpoints have the same name missing values will be replaced by NaN and if different setpoints
+are used the exporter will automatically handle merging coordinates so each data variable is assigned its own coordinates.
```
astafan8: I think it's be useful to add a bit more explanation here. I suggest mentioning "values" of setpoints, not just setpoints, and I think "missing values" needs a bit of explanation, e.g. missing from where, from the "expected" grid, right? Also added punctuation for clarity, what do you think?
```suggestion
Exporting datasets to XArray no longer warns if two or more variables are exported with different setpoint values.
No functionality of the export has changed.
If the setpoints for different measured parameters have the same name, missing values will be replaced by NaN, and if different setpoint values
are used, the exporter will automatically handle merging coordinates so each data variable is assigned its own coordinates.
```

tests/dataset/conftest.py
```diff
@@ -267,12 +271,15 @@ def multi_dataset(experiment, request: FixtureRequest):
         datasaver.dataset.conn.close()
 
 
-@pytest.fixture(scope="function", params=["array"])
+@pytest.fixture(scope="function", params=[True, False])
```
astafan8: does this reflect cleanly in the name of the test when the summary of the execution is shown? I'd think that adding pytest.parameterize with a clear parameter name, e.g. "is_shape_set_explicitly", would improve the readability of the test summary and the test itself

astafan8 (approved):
More coverage - more confidence!

tests/dataset/conftest.py
jenshnielsen: Well this is a parameterized fixture and not a test so that decorator cannot be used. We could probably do something like this https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#using-markers-to-pass-data-to-fixtures to pass args to the fixture but that does not imho improve readability that much so I will rather change the arg from a Boolean to a strenum that is self-describing.

docs/changes/newsfragments/7442.improved
jenshnielsen: Applied and reworded a bit further

PR #7497, opened 2025-09-15
Add New Copper Mountain M5065 and M5180 Drivers

trlemon:
### Summary of Changes
- Added base class for Copper Mountain Technologies M5065 VNA driver.
- Added subclass for M5065 driver (subclasses CMT M5xx base).
- Added subclass for M5180
- Added pyvisa-sim YAML for M5065
- Added tests for M5065
- Added example notebook for M5065

### Notes
- pyvisa-sim doesn't appear to be cooperating outside of driver instantiation and getting IDN, so tests are limited until those issues are resolved.
- This driver was adapted from `qcodes_contrib_drivers` for the [CMT M5180](https://github.com/QCoDeS/Qcodes_contrib_drivers/blob/main/src/qcodes_contrib_drivers/drivers/CopperMountain/M5180.py)

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -39,12 +39,12 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@192325c86100d080feab897ff886c34abd4c83a3 # v3.29.5
+      uses: github/codeql-action/init@303c0aef88fc2fe5ff6d63d3b1596bfd83dfa1f9 # v3.29.5
       with:
         languages: ${{ matrix.language }}
 
     - name: Autobuild
-      uses: github/codeql-action/autobuild@192325c86100d080feab897ff886c34abd4c83a3 # v3.29.5
+      uses: github/codeql-action/autobuild@303c0aef88fc2fe5ff6d63d3b1596bfd83dfa1f9 # v3.29.5
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@192325c86100d080feab897ff886c34abd4c83a3 # v3.29.5
+      uses: github/codeql-action/analyze@303c0aef88fc2fe5ff6d63d3b1596bfd83dfa1f9 # v3.29.5
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -71,6 +71,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@192325c86100d080feab897ff886c34abd4c83a3 # v3.29.5
+        uses: github/codeql-action/upload-sarif@303c0aef88fc2fe5ff6d63d3b1596bfd83dfa1f9 # v3.29.5
         with:
           sarif_file: results.sarif
--- a/requirements.txt
+++ b/requirements.txt
@@ -87,7 +87,7 @@ fonttools==4.60.0
     # via matplotlib
 fsspec==2025.9.0
     # via dask
-furo==2025.7.19
+furo==2025.9.25
     # via qcodes (pyproject.toml)
 h5netcdf==1.6.4
     # via
@@ -335,7 +335,7 @@ pyvisa-sim==0.7.1
     # via qcodes (pyproject.toml)
 pywin32==311; sys_platform == 'win32'
     # via jupyter-core
-pyyaml==6.0.2
+pyyaml==6.0.3
     # via
     #   dask
     #   pyvisa-sim


src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,853 @@
...
+            label="Trigger source",
+            get_parser=str,
+            get_cmd=self._get_trigger,
+            set_cmd=self._set_trigger,
+            docstring="Selects the trigger source.",
+            vals=Enum("bus", "external", "internal", "manual"),
+        )
+        """Selects the trigger source"""
+
+        self.date_transfer_format: Parameter = self.add_parameter(
```
astafan8: typo
```suggestion
        self.data_transfer_format: Parameter = self.add_parameter(
```

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,853 @@
...
+            initial_value=True,
+            vals=Bool(),
+            docstring="Parameter that enables a few commands, which are called"
+            "before each get of a point_sxx parameter checking whether the vna"
+            "is setup correctly. Is recommended to be True, but can be turned"
+            "off if one wants to minimize overhead.",
+        )
+
+        # set the unit of the electrical distance to meter
+        self.write("CALC1:CORR:EDEL:DIST:UNIT MET")
```
astafan8: turn this into a parameter?

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,853 @@
...
+        self.write("CALC1:PAR1:DEF S11")  # Choose S11 for trace 1
+        self.write("CALC1:PAR2:DEF S12")  # Choose S12 for trace 2
+        self.write("CALC1:PAR3:DEF S21")  # Choose S21 for trace 3
+        self.write("CALC1:PAR4:DEF S22")  # Choose S22 for trace 4
+        self.write("CALC1:TRAC1:FORM SMITH")  # Trace format
+        self.write("CALC1:TRAC2:FORM SMITH")  # Trace format
+        self.write("CALC1:TRAC3:FORM SMITH")  # Trace format
+        self.write("CALC1:TRAC4:FORM SMITH")  # Trace format
+        self.write("TRIG:SEQ:SING")  # Trigger a single sweep
+        self.ask("*OPC?")  # Wait for measurement to complete
```
astafan8: this will raise visa timeout error is the duration of the measurement is longer than the visa timeout (that's set via .timeout() parameter that every visa instrument class has), so likely this function should be wrapped around `with self.timeout.set_to(max(self.timeout(), expected_duration_of_the_measurement)):`

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,853 @@
...
+
+        # Get data as string
+        freq_raw = self.ask("SENS1:FREQ:DATA?")
+        s11_raw = self.ask("CALC1:TRAC1:DATA:FDAT?")
+        s12_raw = self.ask("CALC1:TRAC2:DATA:FDAT?")
+        s21_raw = self.ask("CALC1:TRAC3:DATA:FDAT?")
+        s22_raw = self.ask("CALC1:TRAC4:DATA:FDAT?")
+
+        # Get data as numpy array
+        freq = np.fromstring(freq_raw, dtype=float, sep=",")
```
astafan8, line 566: doesn't this assume that format parameter is set to a certain value or does this work for all the formats?

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,853 @@
...
+        self.write("CALC1:PAR:COUN 4")  # 4 trace
+        self.write("CALC1:PAR1:DEF S11")  # Choose S11 for trace 1
+        self.write("CALC1:PAR2:DEF S12")  # Choose S12 for trace 2
+        self.write("CALC1:PAR3:DEF S21")  # Choose S21 for trace 3
+        self.write("CALC1:PAR4:DEF S22")  # Choose S22 for trace 4
+        self.write("CALC1:TRAC1:FORM SMITH")  # Trace format
+        self.write("CALC1:TRAC2:FORM SMITH")  # Trace format
+        self.write("CALC1:TRAC3:FORM SMITH")  # Trace format
+        self.write("CALC1:TRAC4:FORM SMITH")  # Trace format
+        self.write("TRIG:SEQ:SING")  # Trigger a single sweep
```
astafan8: perhaps define a method for this for clarity/readability?

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,853 @@
...
+        Returns:
+            Tuple[ParamRawDataType, ...]: magnitude, phase
+
+        """
+        assert isinstance(self.instrument, CopperMountainM5xxx)
+        self.instrument.write("CALC1:PAR:COUN 1")  # 1 trace
+        self.instrument.write(f"CALC1:PAR1:DEF {self.name}")
+        self.instrument.trigger_source("bus")  # set the trigger to bus
+        self.instrument.write("TRIG:SEQ:SING")  # Trigger a single sweep
+        self.instrument.ask("*OPC?")  # Wait for measurement to complete
```
astafan8: same comment about timeout here as well

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,853 @@
...
+
+        # get data from instrument
+        self.instrument.write("CALC1:TRAC1:FORM SMITH")  # ensure correct format
```
astafan8: does this ignore the format parameter defined in instrument's init? that'd be confusing for users, i beleive.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,853 @@
...
+        sxx_raw = self.instrument.ask("CALC1:TRAC1:DATA:FDAT?")
+        self.instrument.write("CALC1:TRAC1:FORM MLOG")
```
astafan8: why is the format set to something and then data is not asked anymore?

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,853 @@
...
+        """Gets data from instrument
+
+        Returns:
+            Tuple[ParamRawDataType, ...]: magnitude, phase
+
+        """
+        assert isinstance(self.instrument, CopperMountainM5xxx)
+        self.instrument.write("CALC1:PAR:COUN 1")  # 1 trace
+        self.instrument.write(f"CALC1:PAR1:DEF {self.name}")
+        self.instrument.trigger_source("bus")  # set the trigger to bus
```
astafan8: Does it makes sense that this S parameter is expected to be used only with bus triggering? Wouldn't it be better to use the trigger source as specified by users outside of this parameter's get(), as opposed to hardcoding the bus trigger source here?

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: Done

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: Agreed and done.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: The [driver documentation](https://coppermountaintech.com/help-s2/trigsing.html) states that an error will be returned when writing "TRIG:SEQ:SING" unless the trigger source is set to bus. 

> Generates a trigger signal and initiates a sweep under the following conditions.
>  •  Trigger source is set to the BUS (set by the command [TRIG:SOUR](https://coppermountaintech.com/help-s2/trigsour.html) BUS), otherwise an error occurs and the command is ignored.
>  •  Analyzer must be in the trigger waiting state, otherwise (the Analyzer is in the measurement state or in the hold state) an error occurs, and the command is ignored.

>
>As opposed to the [TRIG](https://coppermountaintech.com/help-s2/trig.html) command this command is pending till the end of the sweep. The end of the sweep initiated by the [TRIG:SING](https://coppermountaintech.com/help-s2/trigsing.html) command can be waited using the [*OPC?](https://coppermountaintech.com/help-s2/opc_question.html) query.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: This assumes that the data transfer format is ascii (which is the default). It will throw an error if the data transfer format is set to "real" or "real32".

src/qcodes/instrument_drivers/CopperMountain/_M5065.py
```diff
@@ -0,0 +1,20 @@
+from typing import TYPE_CHECKING
+
+from ._M5xxx import CopperMountainM5xxx
```
jenshnielsen, line 3: I think we should also keep these 2 modules private so the only public entry is the top level module. I wish we did this for other drivers but we did not. I however, still think that makes sense doing going forward

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: I restricted this data format parameter to be ASCII only.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: The `data_transfer_format` parameter controls the format of the response (ASCII, REAL, REAL32), not the format of the returned array.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: Added context manager for timeout here as well.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: It must have been a remnant from the original authors... I removed it since it appears to do nothing (after some testing, of course).

src/qcodes/instrument_drivers/CopperMountain/_M5065.py
trlemon: I made both modules private.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,898 @@
...
+            sxx_raw = self.instrument.ask("CALC1:TRAC1:DATA:FDAT?")
+
+        # Get data as numpy array
+        sxx = np.fromstring(sxx_raw, dtype=float, sep=",")
+        sxx = sxx[0::2] + 1j * sxx[1::2]
+
+        # Return the average of the trace, which will have "start" as
+        # its setpoint
+        sxx_mean = np.mean(sxx)
+        return 20 * math.log10(abs(sxx_mean)), (cmath.phase(sxx_mean))
```
samantha-ho: Nitpick: 
We import the entire cmath library just to calculate the phase? I would prefer np.angle, which is widely used elsewhere instead. Same for math.log10 -> np.log10

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,898 @@
...
+        """
+        stop = self.stop()
+        if val >= stop:
+            raise ValueError("Stop frequency must be larger than start frequency.")
+        self.write(f"SENS1:FREQ:STAR {val}")
+        # we get start as the vna may not be able to set it to the
+        # exact value provided.
+        start = self.start()
+        if abs(val - start) >= 1:
+            log.warning(f"Could not set start to {val} setting it to {start}")
```
samantha-ho: Is a warning appropriate here? I would categorize this more as INFO, than warning, since I would not necessarily expect the user to react to this. Same for line 436.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,898 @@
...
+        if r.lower() == "int":
+            return "internal"
+        elif r.lower() == "ext":
+            return "external"
+        elif r.lower() == "man":
+            return "manual"
+        else:
+            return "bus"
+
+    def _set_trigger(self, trigger: str) -> None:
```
samantha-ho: Nitpick:
Since the allowed strs come from a specific set, consider typing `trigger` as `Literal["external", "internal", "manual", "bus"]` This isn't required, as the Parameter Validator handles it, but it's nice.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,898 @@
...
+
+        Returns:
+            None
+
+        """
+
+        for trace in traces:
+            self.write(f"CALC1:TRAC{trace}:FORM SMITH")
+
+    def get_s(
```
samantha-ho: Perhaps rename this `get_s_parameters` for greater readability.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,898 @@
...
+
+        Returns:
+            data: data transformed in dB.
+
+        """
+
+        return 20.0 * np.log10(np.abs(data))
+
+
+class FrequencySweepMagPhase(MultiParameter):
```
samantha-ho, line 625: For info only:
I consider the MultiParameter class to be Legacy, and many of the newer structures do not necessarily handle it in a new way. Instead, I would recommend returning a complex value or returning the Magnitude and Phase sweeps individually.

I see that this is how the Rohde-Schwarz ZNB-class VNA's implement their mag/phase sweep, so I'm not  asking for this to be changed.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,898 @@
...
+        sxx = np.fromstring(sxx_raw, dtype=float, sep=",")
+        sxx = sxx[0::2] + 1j * sxx[1::2]
+
+        # Return the average of the trace, which will have "start" as
+        # its setpoint
+        sxx_mean = np.mean(sxx)
+        return 20 * math.log10(abs(sxx_mean)), (cmath.phase(sxx_mean))
+
+
+class PointIQ(MultiParameter):
```
samantha-ho, line 828: No FrequencySweepIQ? 

May be worth double-checking with the experimentalists. I am more used to thinking about FrequencySweep measurements in IQ first (since this can be easily encoded in a single complex value), and converting to Mag/Phase later.

samantha-ho (approved):
Nice work, Thomas. I have left a few comments, most of which are nitpicky or suggestions on different approaches. 

I don't think any of my comments necessitate holding up this PR, though, so feel free to move forward.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: Updated log warnings to info.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: Done.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: Done

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: Done

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: Discussed with experimentalists and this driver seems to have all necessary/desired features.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: Noted for future reference. I may address this in a future PR.

docs/examples/driver_examples/QCoDeS example with CopperMountain_M5065.ipynb
```diff
@@ -0,0 +1,537 @@
...
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "placed-maldives",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "NoTagError: `git describe --long --dirty --always --tags '--match=v*'` could not find a tag\n"
```
jenshnielsen: Could we remove this output from the file. I think its because its generated from an editble install where tags are missing

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,898 @@
...
+        """
+        stop = self.stop()
+        if val >= stop:
+            raise ValueError("Stop frequency must be larger than start frequency.")
+        self.write(f"SENS1:FREQ:STAR {val}")
+        # we get start as the vna may not be able to set it to the
+        # exact value provided.
+        start = self.start()
+        if abs(val - start) >= 1:
+            log.info(f"Could not set start to {val} setting it to {start}")
```
jenshnielsen: There is a logger attached to instruments by default which we should use inside the instruments as it makes filtering simpler, https://github.com/microsoft/Qcodes/blob/main/src/qcodes/instrument/instrument_base.py#L116

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
```diff
@@ -0,0 +1,898 @@
...
+
+        """
+
+        for trace in traces:
+            self.write(f"CALC1:TRAC{trace}:FORM SMITH")
+
+    def get_s_parameters(
+        self, expected_measurement_duration: float = 600
+    ) -> tuple[
+        np.ndarray,
```
jenshnielsen: [NIT] we prefer the NDArray type from numpy.typing since it has better defaults for the shape and dtype variables. If we know that these are floats of a specific size e.g. 64bit we can also do NDArray[np.float64] to make this clear

docs/examples/driver_examples/QCoDeS example with CopperMountain_M5065.ipynb
trlemon: I removed it.

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: Done

src/qcodes/instrument_drivers/CopperMountain/_M5xxx.py
trlemon: Done

PR #7521, opened 2025-09-24
Keysight33000types

jenshnielsen:


src/qcodes/instrument_drivers/Keysight/KeysightAgilent_33XXX.py
```diff
@@ -411,16 +415,81 @@ def __init__(
 
         self.num_channels = no_of_channels[self.model]
 
-        for i in range(1, self.num_channels + 1):
-            channel = Keysight33xxxOutputChannel(self, f"ch{i}", i)
-            self.add_submodule(f"ch{i}", channel)
+        if dynamic_channels:
```
jenshnielsen, line 425: We should have some logic here to verify that the class type matches what get_idn type.

astafan8 (approved):
Good approach for these drivers. Perhaps SingleChannel instead of 1Channel reads better? I’m not sure, just wondering here.

PR #7534, opened 2025-09-29
Add Blocking to DynaCool PPMS During Temperature Ramp

trlemon:
### Summary of Changes
- Add blocking function to DynaCool PPMS driver that waits for temperature state to be stable before allowing other functions to be called when setting the temperature.

### DIFF
--- a/src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
+++ b/src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
@@ -83,17 +83,16 @@ class DynaCool(VisaInstrument):
         )
         """Parameter temperature_setpoint"""
 
-        self.blocking_t: Parameter = self.add_parameter(
-            "blocking_t",
+        self.block_while_ramping_temperature: Parameter = self.add_parameter(
+            "block_while_ramping_temperature",
             label="Block instrument while ramping temperature",
-            unit="K",
-            vals=vals.Numbers(1.6, 400),
+            initial_value=False,
+            vals=vals.Bool(),
             get_cmd=False,
-            set_cmd=partial(
-                self._temp_setter, "temperature_setpoint", block_while_ramping=True
-            ),
+            set_cmd=False,
         )
-        """Parameter blocking_t will block instrument interaction while temperature is ramping to setpoint."""
+        """Parameter block_while_ramping_temperature, when set to True,
+        will block further interaction while temperature is ramping to setpoint."""
 
         self.blocking_t_state_check_interval: Parameter = self.add_parameter(
             name="blocking_t_state_check_interval",
@@ -104,7 +103,8 @@ class DynaCool(VisaInstrument):
             set_cmd=None,
             get_cmd=None,
         )
-        """Parameter blocking_t_state_check_interval sets how often blocking_t checks for temperature stability."""
+        """Parameter blocking_t_state_check_interval sets how often
+        temperature_setpoint checks for temperature stability when block_while_ramping_temperature is True."""
 
         self.temperature_rate: Parameter = self.add_parameter(
             "temperature_rate",
@@ -423,7 +423,6 @@ class DynaCool(VisaInstrument):
             "temperature_setpoint", "temperature_rate", "temperature_settling"
         ],
         value: float,
-        block_while_ramping: bool = False,
     ) -> None:
         """
         The setter function for the temperature parameters. All three are set
@@ -435,12 +434,11 @@ class DynaCool(VisaInstrument):
 
         self.write(f"TEMP {values[0]}, {values[1]}, {values[2]}")
 
-        if block_while_ramping:
+        if self.block_while_ramping_temperature():
             while self.temperature_state() != "stable":
                 sleep(self.blocking_t_state_check_interval())
 
         self.setpoint.cache._set_from_raw_value(values[0])
-        self.blocking_t.cache._set_from_raw_value(values[0])
 
     def write(self, cmd: str) -> None:
         """
--- a/docs/examples/driver_examples/Qcodes example with DynaCool PPMS.ipynb
+++ b/docs/examples/driver_examples/Qcodes example with DynaCool PPMS.ipynb
@@ -200,9 +200,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Alternatively, you can use the `blocking_t` parameter to prevent users from taking any action until the setpoint is reached. This may be useful for measurements where you are sweeping temperature and want to ensure the temperature is stable before taking a measurement.\n",
+    "Alternatively, you can use the `block_while_ramping_temperature` parameter to prevent users from taking any action until the setpoint is reached. This may be useful for measurements where you are sweeping temperature and want to ensure the temperature is stable before taking a measurement.\n",
     "\n",
-    "The interval at which `blocking_t` checks for temperature stability can be modified through the parameter `blocking_t_state_check_interval`."
+    "The interval at which `temperature_setpoint` checks for temperature stability when `block_while_ramping_temperature` is True can be modified through the parameter `blocking_t_state_check_interval`."
    ]
   },
   {
@@ -212,7 +212,8 @@
    "outputs": [],
    "source": [
     "dynacool.blocking_t_state_check_interval(0.5)\n",
-    "dynacool.blocking_t(dynacool.temperature() + 1.3)"
+    "dynacool.block_while_ramping_temperature(True)\n",
+    "dynacool.temperature_setpoint(dynacool.temperature() + 1.3)"
    ]
   },
   {


src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
```diff
@@ -410,6 +410,12 @@ def _temp_setter(
         values[self.temp_params.index(param)] = value
 
         self.write(f"TEMP {values[0]}, {values[1]}, {values[2]}")
+        self._block_on_temperature_ramp()
+
+    def _block_on_temperature_ramp(self) -> None:
+        """Block all instrument interactions until temperature stabilizes."""
+        while self.temperature_state() != "stable":
+            sleep(1)
```
astafan8: let's make this sleep time configurable via an attrbiute on the class?

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
```diff
@@ -410,6 +410,12 @@ def _temp_setter(
         values[self.temp_params.index(param)] = value
 
         self.write(f"TEMP {values[0]}, {values[1]}, {values[2]}")
+        self._block_on_temperature_ramp()
```
astafan8: i'd suggest to leave the current T parameter's implementation as-is in order not to break any existing code in terms of the fact that now setting the temperature blocks; and instead implement a new parameter that includes the "block on ramp". What do you think?

trlemon: I created a new parameter to address this but also added a default bool to _temp_setter to control blocking functionality without breaking current implementation of the original temperature setting parameters.

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
trlemon: Added `_blocking_t_sleep` attribute.

SherwanMicrosoft:
## Testing Results - Bug Found

**Environment:**
- PPMS: Quantum Design DynaCool
- QCodes version: 0.55.0.dev194 (this PR branch)
- Testing date: December 19, 2025

### Issue: `blocking_t()` causes PPMS server crash

**Steps to Reproduce:**
1. Fresh PPMS connection - reads work normally (`ppms.temperature()` ✓)
2. Execute: `ppms.blocking_t(300)`
3. Result: `VI_ERROR_IO` on the TEMP write command
4. Subsequent reads fail with `VI_ERROR_CONN_LOST`
5. PPMS server connection is lost

**Error Details:**
VisaIOError: VI_ERROR_IO (-1073807298): Could not perform operation because of I/O error
at DynaCool.py:434 in write(): self._error_code = int(self.visa_handle.read())

Then:
VisaIOError: VI_ERROR_CONN_LOST (-1073807194): The connection for the given session has been lost


**Observations:**
- ✅ `ppms.temperature_setpoint()` (non-blocking) works fine
- ❌ `ppms.blocking_t()` crashes the connection
- ✅ `blocking_t` parameter exists and is configured correctly
- Issue occurs immediately when TEMP command is sent with blocking enabled

**Root Cause Analysis:**
The failure occurs at line 434 when trying to read the error code response after sending the TEMP command. The PPMS receives the command but fails to respond, then the connection is lost.

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
jenshnielsen: I am not a big fan of having two parameters that control the temperature. Which one is the source of truth? I would have made a manual boolean parameter (block_during_temperature_ramp) and then make that one control the behaviour of the existing temperature parameter

trlemon:
@SherwanMicrosoft My last commit resolves this issue. Thank you for the detailed report!

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
```diff
@@ -83,6 +84,17 @@ def __init__(
         )
         """Parameter temperature_setpoint"""
 
+        self.blocking_t: Parameter = self.add_parameter(
+            "blocking_t",
+            label="Block instrument while ramping temp",
```
astafan8:
```suggestion
            label="Block instrument while ramping temperature",
```

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
```diff
@@ -83,6 +84,17 @@ def __init__(
         )
         """Parameter temperature_setpoint"""
 
+        self.blocking_t: Parameter = self.add_parameter(
```
astafan8: could you mention this parameter in the example notebook https://microsoft.github.io/Qcodes/examples/driver_examples/Qcodes%20example%20with%20DynaCool%20PPMS.html ?

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
astafan8: This is better, however it's a private attribute, so as if it's not supposed to be touched by the users, which is not the case, i think it's totally fair that users can adjust that sleep time. So with that logic, i suggest to make it a memory/in-memory parameter (`get_cmd=None, set_cmd=None`), similar to how it's done with this parameter https://github.com/microsoft/Qcodes/blob/d3378235466091104068012601edb08a5d769728/src/qcodes/instrument_drivers/american_magnetics/AMI430_visa.py#L856 that controls sleep interval for a while loop that monitors magnetic field ramp.

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
astafan8: My guess is that we're trying to follow the pattern that was established in lakeshore temperature controller driver https://microsoft.github.io/Qcodes/examples/driver_examples/Qcodes%20example%20with%20Lakeshore%20336%20or%20372%20-%20Bluefors%20T%20control.html#Sweeping/blocking-paramameter where there's `setpoint` parameter that's not blocking, and a `blocking_t` parameter that does the same as `setpoint` but also blocks until it's reached.

I'd be in favor of keeping consistency across the drivers (even if it's likely opinionated), so having both setpoint and blocking_t parameters is ok as long as their caches are in sync, i left a separate comment about it.

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
```diff
@@ -83,6 +84,17 @@ def __init__(
...
+            "blocking_t",
+            label="Block instrument while ramping temp",
+            unit="K",
+            vals=vals.Numbers(1.6, 400),
+            set_cmd=partial(
```
astafan8: let's make this parameter not gettable (explicitly :) )
```suggestion
            get_cmd=False,
            set_cmd=partial(
```

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
```diff
@@ -411,6 +424,10 @@ def _temp_setter(
 
         self.write(f"TEMP {values[0]}, {values[1]}, {values[2]}")
 
+        if block_while_ramping:
+            while self.temperature_state() != "stable":
+                sleep(self._blocking_t_sleep)
+
```
astafan8, line 440: let's make sure that the caches of the two temperature parameters are in sync when either of them gets set. Note that private method _set_from_raw_value is used since we operate inside this method with raw values (not values, which could have offset/scale), so for correctness reasons we need to use the private method, which is OK to use inside qcodes and inside driver code, just not in user facing measurement code.
```suggestion
        self.setpoint.cache._set_from_raw_value(values[0])
        self.blocking_t.cache._set_from_raw_value(values[0])
```

trlemon: Thank you for pointing this out. I have committed the suggested changes.

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
trlemon: Noted. I have committed this change.

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
trlemon: I ended up adding a default bool argument to the `_temp_setter` function which is passed from a call to blocking_t so that we only have one function that is really setting the temperature.

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
trlemon: I committed this change :)

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
trlemon: Done. I added a short description and added a basic usage example.

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
trlemon: @astafan8 I followed your suggestion and added this as a parameter. I also updated the PPMS notebook example to make it known to future users.

src/qcodes/instrument_drivers/QuantumDesign/DynaCoolPPMS/DynaCool.py
jenshnielsen: @trlemon I have a conversation with @astafan8 and I think I ended up convincing him that a single temperature parameter would be better plus another parameter that controls the value of the boolean. Would that still be possible to implement?

Alternatively, I think we should add a comment inline that we are using this pattern with 2 temperature parameters to maintain compatibility with existing drivers but that we do not recommend copying this to unrelated implementations.

trlemon: @jenshnielsen I debated a bit back and forth with myself, and I ultimately agree with this approach. I was initially hesitant to do this because maintaining compatibility with the Lakeshore336 `blocking_t` seemed like the "easy" thing to do for measurements that use both temperature controllers. However, the setup is different enough between the temperature controllers that it makes sense to just set the blocking functionality with a Bool on the PPMS. I've gone ahead and made the change.

PR #7606, opened 2025-10-30
Update LakeshoreModel3xx pyvisa-sim YAMLs and tests

trlemon:
### Summary of Changes 
1. Added missing commands to `lakeshore_model335.yaml`, `lakeshore_model336.yaml`, `lakeshore_model372.yaml` and rename some properties for clarity.
2. Added logic to `lakeshore_base.py` to bypass "blocking" functionality in `blocking_t` while in simulated mode.
3. Updated tests for Lakeshore Models 325, 335, 336, and 372 to use pyvisa-sim backend instead of mocked classes.

### DIFF
--- a/tests/drivers/test_lakeshore_372.py
+++ b/tests/drivers/test_lakeshore_372.py
@@ -207,6 +207,13 @@ class LakeshoreModel372Mock(MockVisaInstrument, LakeshoreModel372):
         # start at 7K.
         return max(4, 7 - delta)
 
+    @query("KRDG?")
+    def temperature(self, output):
+        chan = self.channel_mock[output]
+        if self.simulate_heating:
+            return self.get_t_when_heating()
+        return f"{chan.T}"
+
 
 def instrument_fixture(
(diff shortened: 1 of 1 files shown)


tests/drivers/test_lakeshore_336.py
```diff
@@ -1,208 +1,21 @@
...
-        "lakeshore_336_fixture",
+    """Create a Lakeshore 336 instance using PyVISA-sim backend."""
+    inst = LakeshoreModel336(
+        "lakeshore_336",
         "GPIB::2::INSTR",
         pyvisa_sim_file="lakeshore_model336.yaml",
         device_clear=False,
     )
+    try:
+        yield inst
```
picarro-yren: Why is the `try` statement needed here? I don't see it's used for lakeshore 335 and 372.

trlemon: It is used in the Lakeshore372 tests. I just updated the two other Lakeshore tests (335 and 336 legacy) to use `yield` with `try` and `finally`.

PR #7641, opened 2025-11-13
Add more explicit channel atttributes

jenshnielsen:
Work towards more strongly typed instruments

### DIFF
--- a/src/qcodes/instrument/channel.py
+++ b/src/qcodes/instrument/channel.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import sys
 import warnings
 from collections.abc import Callable, Iterable, Iterator, MutableSequence, Sequence
-from typing import TYPE_CHECKING, Any, Generic, cast, overload
+from typing import TYPE_CHECKING, Any, Generic, Self, cast, overload
 
 from typing_extensions import TypeVar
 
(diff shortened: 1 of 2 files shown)


src/qcodes/instrument_drivers/rohde_schwarz/ZNB.py
jenshnielsen: This line now fails to type check since an attribute on a ChannelList can be a multichannelparameter, a channel etc. I think we should change the type of `__getattr__` to just return Any and provide a type safe interface for parameters and channels respectively

src/qcodes/instrument/channel.py
```diff
@@ -441,15 +441,88 @@ def snapshot_base(
             }
         return snap
 
-    def __getattr__(
-        self, name: str
-    ) -> MultiChannelInstrumentParameter | Callable[..., None] | InstrumentModuleType:
+    def multi_parameter(
+        self: Self, name: str
+    ) -> MultiChannelInstrumentParameter[InstrumentModuleType]:
```
jenshnielsen, line 446: Need to add test coverage for this code + changelog

jenshnielsen:
Moved the pr here https://github.com/microsoft/Qcodes/pull/7686 to experiment with copilot

PR #7655, opened 2025-11-18
[wip] Generic instrument in parameter

jenshnielsen:


src/qcodes/parameters/parameter_base.py
```diff
@@ -109,7 +116,7 @@ def invert_val_mapping(val_mapping: Mapping[Any, Any]) -> dict[Any, Any]:
     return {v: k for k, v in val_mapping.items()}
 
 
-class ParameterBase(MetadatableWithName):
+class ParameterBase(MetadatableWithName, Generic[_InstrumentType_co]):
```
jenshnielsen: Before merging this we need to decide which generic parameters the parameter class should take and which order they should be in. 3.13+ allows some generic arguments to be omitted but since there are no support for kwargs they must be included in order so the first argument should be the one that must be costumized the most often 

Candidates are ParamDataType, ParamRawDataType and Instrument. 

It probably makes sense that ParamDataType is the first since each Parameter instance may have a different datatype while the Instrument type typically only needs to be customized in a subclass. 

Should we include ParamRawData type or just leave it as any (if we include it should it come before or after Instrument)

src/qcodes/parameters/parameter_base.py
```diff
@@ -109,7 +117,9 @@ def invert_val_mapping(val_mapping: Mapping[Any, Any]) -> dict[Any, Any]:
     return {v: k for k, v in val_mapping.items()}
 
 
-class ParameterBase(MetadatableWithName):
+class ParameterBase(
+    MetadatableWithName, Generic[_ParameterDataTypeVar, _InstrumentType_co]
+):
```
jenshnielsen, line 122: We need to agree on this order of generic arguments before merging since changing this would be a breaking change for all Parameter usage and implementations.

Roughly there are three relevant generic parameters for an instrument. 

1. ParameterData type (input/output from get/set)
2. InstrumentType (what self.instrument returns)
3. RawParameterData. What is passed to self.write/ask etc

Here I propose that we add 1 and 2 but omit 3 for the time being. 
This means that we cannot add 3 before

jenshnielsen:
Closing since everything has landed or is open elsewhere

PR #7659, opened 2025-11-19
Make InstrumentModule/Channel take a generic instrument argument

jenshnielsen:
This makes it possible to write drivers that use features from the parent without explicit cast

- [x] Changelog

### DIFF
--- a/docs/changes/newsfragments/7659.improved
+++ b/docs/changes/newsfragments/7659.improved
@@ -1,0 +1,1 @@
+The ``InstrumentModule`` and its alias ``InstrumentChannel`` now take an optional generic argument allowing you to specify the type of the parent instrument for type checking.
--- a/docs/changes/newsfragments/7659.improved_driver
+++ b/docs/changes/newsfragments/7659.improved_driver
@@ -1,0 +1,2 @@
+The ``KeysightE4980A`` driver now names the ``correction`` submodule correctly as ``correction`` reflecting the public attribute to access the module.
+This also means that in the snapshot ``correction`` is used as the module name rather than ``_correction``


src/qcodes/instrument_drivers/Keysight/keysight_e4980a.py
```diff
@@ -371,14 +371,16 @@ def __init__(
...
+            "_correction", KeysightE4980ACorrection(self, "correction")
+        )
         self._set_signal_mode_on_driver_initialization()
         self.connect_message()
 
     @property
     def correction(self) -> KeysightE4980ACorrection:
-        submodule = self.submodules["_correction"]
-        return cast("KeysightE4980ACorrection", submodule)
+        submodule = self._correction
```
jenshnielsen: Perhaps just drop this property and rename the module. That would change the snapshot but for the better

src/qcodes/instrument_drivers/rigol/Rigol_DG1062.py
```diff
@@ -415,13 +414,14 @@ def __init__(
...
-            self.add_submodule(ch_name, channel)
+        self.ch1 = self.add_submodule("ch1", RigolDG1062Channel(self, "ch1", 1))
+        """Channel 1 submodule"""
+        self.ch2 = self.add_submodule("ch2", RigolDG1062Channel(self, "ch2", 2))
+        """Channel 2 submodule"""
 
-        self.add_submodule("channels", channels.to_channel_tuple())
+        self.channels: ChannelTuple[RigolDG1062Channel] = self.add_submodule(
+            "channels",
+            ChannelTuple(self, "channel", RigolDG1062Channel, (self.ch1, self.ch2)),
```
jenshnielsen: snapshotable=False

astafan8 (approved):

PR #7669, opened 2025-11-28
Improve performance of is_function

eendebakpt:
For instruments with many parameters (100 - 1000 range) the overhead of qcodes parameters becomes a bottleneck in startup time. In this PR we improve performance of `is_function`:
```
main: 6.65 μs ± 185 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
PR: 668 ns ± 31.3 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
Testing code:
```
def f(x):
    return None
arg_count=1
%timeit is_function0(f, arg_count) 
```

(there are more bottlenecks, but `is_fuction` is used in creation of every Parameter and is quite heavy)

### DIFF
--- a/src/qcodes/utils/function_helpers.py
+++ b/src/qcodes/utils/function_helpers.py
@@ -22,7 +22,7 @@ def is_function(f: object, arg_count: int, coroutine: bool | None = False) -> bo
     if not callable(f):
         return False
     if coroutine is not None:
-        if  bool(coroutine) is not iscoroutinefunction(f):
+        if bool(coroutine) is not iscoroutinefunction(f):
             return False
 
     if isinstance(f, type):


jenshnielsen:
Thanks @eendebakpt That seems very useful. In the long term I think we should consider completely removing and simplifying the command class from parameter but until then this seems like a nice improvement. I did not look into why the tests doesn't like this as is

eendebakpt:
@microsoft-github-policy-service agree [company="TNO"]

eendebakpt:
> Thanks @eendebakpt That seems very useful. In the long term I think we should consider completely removing and simplifying the command class from parameter but until then this seems like a nice improvement. I did not look into why the tests doesn't like this as is


Tests are passing now. Simplifying the command class would indeed be great! 

One more performance improvement would be to skip the `iscoroutinefunction` check. It is used for anything?

@jenshnielsen

eendebakpt:
@microsoft-github-policy-service agree company=TNO

jenshnielsen:
@eendebakpt I don't think it's used much. Perhaps we could change the value for `coroutine` argument to `coroutine: bool | None = None` and skip the check if coroutine=None ? Note that I changed the import in https://github.com/microsoft/Qcodes/pull/7753 since the function in asyncio was deprecated in python 3.14

eendebakpt:
> @eendebakpt I don't think it's used much. Perhaps we could change the value for `coroutine` argument to `coroutine: bool | None = None` and skip the check if coroutine=None ? Note that I changed the import in #7753 since the function in asyncio was deprecated in python 3.14

Yes, I had a similar idea. Added to the PR!

jenshnielsen (approved):
@eendebakpt Thanks, looks good. Could you run the precommit hooks and this should be good to merge. The other ci issue is resolved on main so should be fixed if you rebase or merge main into this

eendebakpt:
@jenshnielsen I had to run the linting once more

src/qcodes/utils/function_helpers.py
```diff
@@ -19,15 +19,38 @@ def is_function(f: object, arg_count: int, coroutine: bool = False) -> bool:
     if not isinstance(arg_count, int) or arg_count < 0:
         raise TypeError("arg_count must be a non-negative integer")
 
-    if not (callable(f) and bool(coroutine) is iscoroutinefunction(f)):
+    if not callable(f):
         return False
+    if coroutine is not None:
+        if  bool(coroutine) is not iscoroutinefunction(f):
```
jenshnielsen:
```suggestion
        if bool(coroutine) is not iscoroutinefunction(f):
```

jenshnielsen: There seems to be one extra whitespace here breaking formatting in ci

PR #7705, opened 2025-12-11
Track time spend in add_result

jenshnielsen:
Track time spend in add_result of Datasaver and record as span attribute for each measurement run to allow comparison to total run time.

### DIFF
--- a/src/qcodes/dataset/measurements.py
+++ b/src/qcodes/dataset/measurements.py
@@ -17,7 +17,7 @@ from copy import deepcopy
 from inspect import signature
 from itertools import chain
 from numbers import Number
-from time import perf_counter
+from time import perf_counter, perf_counter_ns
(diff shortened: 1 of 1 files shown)


src/qcodes/dataset/measurements.py
```diff
@@ -207,6 +209,7 @@ def add_result(self, *result_tuples: ResType) -> None:
                 its type.
 
         """
+        start_time = perf_counter()
```
jenshnielsen: Perhaps better to use the perf_counter_ns to make sure this is always an int and convert to ms before logging to match otel that measures durations in ms

jenshnielsen: Done. Note this is safe since python integers are variable length so there should never be an issue expressing the duration of a measurement in int ns and any relevant duration in ms can be expressed as a 64 float

PR #7713, opened 2025-12-12
Implement unpack_self() for MultiParameter

thangleiter:
This PR implements the `unpack_self()` method for `MultiParameter`, extending #7346. This makes the special treatment of this type of parameter in `DataSaver.add_result()` redundant and should allow users to make use of the new `has_control_of` and `is_controlled_by` features in custom `MultiParameter` classes.

To Do
------
- [x] modify `DataSet._register_multiparameter()` to account for `has_control_of` and `is_controlled_by` attributes.

Considerations
-----------
If `MultiParameter` is ever going to be refactored, I think it makes sense for each result (name) that it returns to be its own parameter with validators etc. This would allow assigning correct `paramtype`s (currently, paramtype is basically always numeric, and hence inefficiently stored if not overridden. 

Similar to `unpack_self`, registering a parameter in a measurement should also be delegated to the parameter itself so that `Measurement` does not need to be aware of the parameter's internals and make case distinctions. So something like adding a `register_self` method to `ParameterBase`.


thangleiter:
I've implemented the self-registering, which works as I'd expect it to work as far as I can tell. However, it required some changes to the way `ParamSpecBase` is handled in `Parameter` and how `ParameterWithSetpoints` deals with dependencies, which leads to some failing tests. Before addressing those, I'd like some feedback if those changes are acceptable.

PR #7725, opened 2025-12-15
Fix 7717, Improve handling of data variables with inferred parameters + tests

### LINKED ISSUE #7717: Dataset converts to ungridded data if two ParameterWithSetpoints share inference and setpoints
Take the following situation, where two `ParameterWithSetpoints`s share the same `setpoints`, but one has control of the other. This could be the case if one is a physically measured parameter with time points as setpoints and the other is a meta-parameter that describes some instrument ramp sampled at the time points. Since only the first is actually measured, having it take control of the second parameter simplifies defining the measurement.

```python
import tempfile
import numpy as np

from qcodes.validators import Arrays
from qcodes.parameters import ManualParameter, ParameterWithSetpoints
from qcodes.dataset import Measurement, new_experiment, initialise_or_create_database_at

class MySp(ParameterWithSetpoints):
    def unpack_self(self, vals):
        res = super().unpack_self(vals)
        res.append((p1, p1()))
        return res

mp = ManualParameter('mp', vals=Arrays(shape=(10,)), initial_value=np.arange(10))
p1 = ParameterWithSetpoints('p1', vals=Arrays(shape=(10,)), setpoints=(mp,), set_cmd=None)
p2 = MySp('p2', vals=Arrays(shape=(10,)), setpoints=(mp,), set_cmd=None)
p2.has_control_of.add(p1)

p1(np.linspace(-1, 1, 10))
p2(np.random.randn(10))

initialise_or_create_database_at(tempfile.mkdtemp() + '/baz.db')
new_experiment('foo', 'bar')

meas = Measurement()
meas.register_parameter(p2)
with meas.run() as ds:
    ds.add_result((p2, p2()))

xds = ds.dataset.to_xarray_dataset()  # does not unravel to grid
```

Since both `p1` and `p2` share the same setpoints, I'd expect the exported data to be a `DataSet` with `p1` and `p2` as data variables and `mp` as coordinates. But the data is actually not gridded and instead returned as a ravelled array.

jenshnielsen:
* Add failing test based on #7717
* Closes #7717

### DIFF
--- a/src/qcodes/dataset/exporters/export_to_xarray.py
+++ b/src/qcodes/dataset/exporters/export_to_xarray.py
@@ -113,6 +113,18 @@ def _add_inferred_data_vars(
                 dims,
                 flat.reshape(tuple(xr_dataset.sizes[d] for d in dims)),
             )
+        else:
+            _LOG.warning(
+                "Cannot add inferred parameter '%s' to xarray dataset for '%s' "
+                "(run_id=%s): data size %d does not match its parent parameter "
+                "'%s' size %d. This is likely a user error in the measurement setup.",
+                inf.name,
+                name,
+                dataset.run_id,
+                flat.shape[0],
+                parent_name,
+                expected_size,
+            )
 
     return xr_dataset
 
--- a/src/qcodes/dataset/exporters/export_to_xarray.py
+++ b/src/qcodes/dataset/exporters/export_to_xarray.py
@@ -6,6 +6,7 @@ from importlib.metadata import version
 from math import prod
 from typing import TYPE_CHECKING, Literal
 
+import numpy as np
 from packaging import version as p_version
 
 from qcodes.dataset.linked_datasets.links import links_to_str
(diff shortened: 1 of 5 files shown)


tests/dataset/test_parameter_with_setpoints_has_control.py
```diff
@@ -0,0 +1,39 @@
...
+
+from qcodes.dataset import Measurement
+from qcodes.parameters import ManualParameter, ParameterWithSetpoints
+from qcodes.validators import Arrays
+
+if TYPE_CHECKING:
+    from qcodes.dataset.experiment_container import Experiment
+
+
+def test_parameter_with_setpoints_has_control(experiment: "Experiment"):
```
astafan8: add xfail and merge?

astafan8 (approved):

jenshnielsen:
@astafan8 converted to draft until it actually fixes the bug

tests/dataset/test_parameter_with_setpoints_has_control.py
jenshnielsen: @astafan8 Copilot has been working on fixing the issue. Could you rereview

tests/dataset/test_parameter_with_setpoints_has_control.py
```diff
@@ -0,0 +1,68 @@
...
+    # mp should be the only dimension (not a generic 'index')
+    assert list(xds.sizes.keys()) == ["mp"]
+    assert xds.sizes["mp"] == 10
+
+    # mp values used as coordinate axis
+    npt.assert_array_equal(xds.coords["mp"].values, mp_data)
+
+    # p2 is the primary data variable with correct values
+    assert "p2" in xds.data_vars
+    npt.assert_array_almost_equal(xds["p2"].values, p2_data)
```
astafan8, line 83: shall an assrtion be added for p1 in the xarray dataset as well?

astafan8 (approved):
great to see this fixed! And "add a news fragment"?

src/qcodes/dataset/exporters/export_to_xarray.py
```diff
@@ -61,6 +62,56 @@ def _calculate_index_shape(idx: pd.Index | pd.MultiIndex) -> dict[Hashable, int]
...
+
+        # Only add if the data length matches the existing dataset size
+        expected_size = 1
+        for d in dims:
+            expected_size *= xr_dataset.sizes[d]
+        if flat.shape[0] == expected_size:
+            xr_dataset[inf.name] = (
+                dims,
+                flat.reshape(tuple(xr_dataset.sizes[d] for d in dims)),
+            )
```
jenshnielsen, line 122: We should probably warn if this is not the case

tests/dataset/test_parameter_with_setpoints_has_control.py
jenshnielsen: Code updated to correctly handle this case and add a warning

src/qcodes/dataset/exporters/export_to_xarray.py
jenshnielsen: Done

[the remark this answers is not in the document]
jenshnielsen: Rewrote logic to export when shape matches the parameter inferred from and ensured warning is logged if shapes does not match. All with tests.

astafan8 (approved):

src/qcodes/dataset/exporters/export_to_xarray.py
```diff
@@ -61,6 +62,73 @@ def _calculate_index_shape(idx: pd.Index | pd.MultiIndex) -> dict[Hashable, int]
...
+        # Only add if the data has the same size as the parameter
+        # it is inferred from
+        inferred_from_params = interdeps.inferences.get(inf, ())
+        if len(inferred_from_params) == 0:
+            continue
+        parent_name = inferred_from_params[0].name
+        if parent_name not in sub_dict:
+            continue
+        expected_size = sub_dict[parent_name].ravel().shape[0]
+        if flat.shape[0] == expected_size:
```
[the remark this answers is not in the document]
jenshnielsen: The logic was updated to check all inferred from parameters and export if one of them matches. Perhaps it would be more correct if all of them match? @astafan8 what do you think?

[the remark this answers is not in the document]
astafan8: > Perhaps it would be more correct if all of them match?

that depends on whether we expect the data shapes for the inferred from parameters to be the same as of the parent. do we? that's not obvious to me.

PR #7952, opened 2026-03-20
Drop Python 3.11 from CI matrix

astafan8:
## Changes
QCoDeS already requires Python >=3.12. This PR updates CI to match:

- Updated pytest.yaml: removed 3.11 from matrix, min-version test now uses 3.12
- Updated docs.yaml: removed 3.11, gh-pages deploy uses 3.12
- Updated upload_to_pypi.yaml: build uses 3.12
- Removed obsolete pyvisa-sim deprecation warning filter (was only needed for 3.11)


src/qcodes/parameters/delegate_parameter.py
```diff
@@ -36,7 +36,7 @@
 
 class DelegateParameter(
     Parameter[ParameterDataTypeVar, InstrumentTypeVar_co],
-    Generic[ParameterDataTypeVar, InstrumentTypeVar_co],
+    Generic[ParameterDataTypeVar, InstrumentTypeVar_co],  # noqa: UP046 -- Generic kept because multi-inheritance with type params is not yet supported
```
jenshnielsen: This comment is incorrect. The reason this cannot be done is that the typevar is imported from typing_extensions but ruff cannot see this since the typevar is defined in another module. IMHO it is cleaner to copy the typevar definition here (perhaps make it private) then it can be deleted when 3.12 is dropped

jenshnielsen (approved):

PR #7993, opened 2026-03-31
Start adding support for new ruff defaults

jenshnielsen:


jenshnielsen:
replaced by #8368  and others

PR #8033, opened 2026-04-10
USB Direct Programming Interface for MinicircuitsUSBSPDT

samantha-ho:


samantha-ho:
@jenshnielsen 
Here's my draft PR for the change to the MinicircuitsUSBSPDT driver that removes the need for the DLL and Pythonnet/CLR packages. 

I have tested this locally, and it works as far as it goes. There are a few errors, and I do not know enough about the libusb1 interface to address them yet.

PR #8037, opened 2026-04-13
feat(dataset): expose ParamSpecBase from qcodes.dataset

### LINKED ISSUE #7369: qcodes.dataset exports ParamSpec and not ParamSpecBase
The qcodes.dataset __init__.py exports the `ParamSpec` class and not the `ParamSpecBase` class, even though the `ParamSpec` class is not used anywhere.

officialasishkumar:
## Summary

Closes #7369.

``ParamSpecBase`` is the base class used for building ``InterDependencies_`` graphs on the public API surface, but until this change it was only importable from ``qcodes.parameters``. ``qcodes.dataset.__init__`` only re-exported the legacy ``ParamSpec`` class, which meant users following the documented dataset-centric entry points had to reach into ``qcodes.parameters`` to obtain it.

This PR re-exports ``ParamSpecBase`` from ``qcodes.dataset`` and adds it to ``__all__`` so the public namespace matches the recommended usage.

- Add ``ParamSpecBase`` to the re-exports in ``src/qcodes/dataset/__init__.py`` and ``__all__``.
- Add a regression test in ``tests/dataset/test_dependencies.py`` that pins the identity of the re-export and its membership in ``__all__``.
- Add a newsfragment.

## Test plan

- [x] ``pytest tests/dataset/test_dependencies.py`` — 19 passed locally.
- [x] ``python -c "from qcodes.dataset import ParamSpecBase; from qcodes.parameters import ParamSpecBase as P; assert ParamSpecBase is P"`` succeeds.

### DIFF
--- a/docs/changes/newsfragments/7369.improved
+++ b/docs/changes/newsfragments/7369.improved
@@ -1,0 +1,3 @@
+``ParamSpecBase`` is now re-exported from :mod:`qcodes.dataset` so that users
+defining custom ``InterDependencies_`` graphs via the public ``qcodes.dataset``
+namespace do not need to reach into :mod:`qcodes.parameters` to obtain it.
--- a/src/qcodes/dataset/__init__.py
+++ b/src/qcodes/dataset/__init__.py
@@ -17,6 +17,8 @@ from .database_extract_runs import (
     export_datasets_and_create_metadata_db,
     extract_runs_into_db,
 )
+from qcodes.parameters import ParamSpecBase
+
 from .descriptions.dependencies import InterDependencies_, ParamSpecTree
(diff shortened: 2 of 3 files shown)


jenshnielsen:
@officialasishkumar
Thansk for the contribution. However,  ParamSpecBase is exported from qcodes.parameters where it is currently defined. This is because the dataset module depends on the parameters module which means that the parameters module cannot depend on the dataset module and ParamSpecBase is required in the parameters module. See https://microsoft.github.io/Qcodes/api/parameters/index.html#qcodes.parameters.ParamSpecBase

We normally only export the same symbol once.

PR #8038, opened 2026-04-13
fix(instrument): silence get_idn warning on virtual instruments

### LINKED ISSUE #4611: get_idn should not throw for virtual instruments
Virtual instruments should either override get_idn to not fail and respond with something useful, or we should rework this so get_idn is not implemented like this for the generic class.

the reason for the failure is that get_idn calls `.ask` which is not implemented for virtual instruments.

An example of this issue is in #4590

officialasishkumar:
## Summary

Closes #4611.

``Instrument.get_idn`` calls ``self.ask("*IDN?")`` which delegates to ``ask_raw``. The base ``Instrument`` class leaves ``ask_raw`` as a stub that raises ``NotImplementedError``, so any virtual instrument that does not override it — e.g. a purely in-memory model — always produced a warning on every call to ``get_idn``. That regularly appears when a virtual instrument is added to a ``Station`` and the station queries its IDN during snapshotting.

This PR narrows the ``except`` block so that ``NotImplementedError`` is treated as the expected signal of a virtual instrument and returns the default ``None``-valued dict silently. All other exceptions still go through ``self.log.warning`` with ``exc_info=True``, so genuine communication failures remain visible.

- ``src/qcodes/instrument/instrument.py``: split the broad ``except Exception`` so ``NotImplementedError`` is handled without a warning.
- ``tests/test_instrument.py``: two regression tests — one confirms no warning is emitted for a virtual instrument without ``ask_raw``, the other confirms that unexpected errors (e.g. ``RuntimeError`` from a broken ``ask_raw``) still produce a warning.
- Added a newsfragment.

## Test plan

- [x] ``pytest tests/test_instrument.py`` — 34 passed locally, including the two new tests.
- [x] ``pytest tests/test_instrument.py -k get_idn`` — 3 passed.

### DIFF
--- a/src/qcodes/instrument/instrument.py
+++ b/src/qcodes/instrument/instrument.py
@@ -91,10 +91,6 @@ class Instrument(InstrumentBase, metaclass=instrument_meta_class):
         semicolon and colon are also common separators so we accept them here
         as well.
 
-        Virtual instruments that do not define ``ask_raw`` will produce a dict
-        with ``None`` entries (and ``name`` as the model) without emitting a
-        warning, since the absence of real hardware communication is expected.
-
         Returns:
             A dict containing vendor, model, serial, and firmware.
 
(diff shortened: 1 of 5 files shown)


src/qcodes/instrument/instrument.py
```diff
@@ -109,6 +113,11 @@ def get_idn(self) -> dict[str, str | None]:
             # in case parts at the end are missing, fill in None
             if len(idparts) < 4:
                 idparts += [None] * (4 - len(idparts))
+        except NotImplementedError:
```
jenshnielsen: Rather than doing this I think my preference would be to add a VirtualInstrument base class. This should probably be a sibling class to Instrument e.g. inherit from InstrumentBase but not from instrument. This method can supply an empty idn dict

PR #8080, opened 2026-04-29
fix(dataset): handle plot for dataset with incorrect parameter data

einsmein:
This PR modifies `plot_dataset` to handle dataset with incorrect parameter data.

When dataset is somehow exported incorrectly, such that `get_parameter_data` shows setpoints and dependent params as
```python
{
  "dep": {"dep": [...], "index": [...]}, 
  "setpoint": {"setpoint": [...], "index": [...]}, 
}
```
instead of
```python
{"dep" {"dep": [...], "setpoint": [...]}}
```
`plot_dataset` function previously throws a `KeyError`.
(Specifically, we see this when a dataset with unsorted setpoint was exported to netcdf with flag `use_multi_index="always"`.)


jenshnielsen (approved):
Looks good. Perhaps we should add a changelog entry

PR #8096, opened 2026-05-04
Convert to PEP 695 type parameters and deprecate removed TypeVars + type aliases 

jenshnielsen:
- Convert 22 classes/functions to PEP 695 type parameter syntax (UP046/UP047)
- Add noqa: UP046 with comments for 14 cases that require PEP 696 (TypeVar defaults), which needs Python 3.13+
- Fix ruff autofix bug in Base_SPDT.py (TINSTR vs _TINSTR mismatch)
- Clean up unused TypeVar definitions and imports across ~13 files
- Add _make_deprecated_typevars_getattr helper in deprecate.py for module-level __getattr__ deprecation warnings
- Add deprecation warnings (QCoDeSDeprecationWarning) for 9 modules with removed public TypeVars, guarded by if not TYPE_CHECKING so type checkers still flag removed imports as errors
- Add tests for the deprecation helper

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -78,7 +78,7 @@ def retry_until_does_not_throw(
     return retry_until_passes_decorator
 
 
-def profile(func: Callable[P, T]) -> Callable[P, T]:
+def profile[**P, T](func: Callable[P, T]) -> Callable[P, T]:
     """
     Decorator that profiles the wrapped function with cProfile.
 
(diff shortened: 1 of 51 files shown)


tests/common.py
jenshnielsen, line 56: This is missing an update

astafan8 (approved):

PR #8195, opened 2026-06-06
Fix lower bound of the span parameter validator of the KeysightPNABase

hbeukers:
The lower bound of the span in the KeysightPNABase should be 0 Hz, it was currently set to min_freq, which incorrectly limits setting the span for small values.


jenshnielsen (approved):
Thanks

PR #8219, opened 2026-06-12
feat: split raw data storage into per-dataset SQLite files

astafan8:
## Summary

This PR implements **split raw data storage** for QCoDeS: an opt-in feature that writes raw measurement data (results table rows) into individual per-dataset SQLite files while keeping all metadata in the main database. The goal is to prevent the main DB file from growing excessively large as datasets accumulate, so that a user can keep a single lightweight metadata database on their machine instead of many ever-growing ones.

It also includes **dataset management helpers** for maintaining the main database and raw data files over time.

## Motivation

The main QCoDeS SQLite database stores both metadata (experiments, runs, parameter layouts, dependencies) and raw measurement data (results tables) in a single file. Over time this file can grow to many gigabytes, slowing down operations that only need metadata. By splitting the raw data into per-dataset files, the main DB stays lightweight while data integrity is preserved.

### Transparent routing via `_data_conn`
- A single `_data_conn` property on `DataSet` is the routing point for all results-data read/write operations. It returns the per-dataset raw data connection when split storage is enabled, otherwise the main DB connection.
- All write paths (`add_results`, `_BackgroundWriter`) and read paths (`get_parameter_data`, `DataSetCacheWithDBBackend`, `number_of_results`, `__len__`) go through it.
- **Zero changes to the public DataSet API** — every existing method behaves identically whether the feature is on or off.

### Config — follows the existing export-path pattern
Two new options in the `dataset` section of `qcodesrc.json`:
- `raw_data_to_separate_db` (bool, default `false`) — enables split storage.
- `raw_data_path` (string, default `"{db_location}"`) — folder for per-dataset files. `{db_location}` expands to a folder derived from the main DB path (reusing `_expand_export_path`).

### Per-dataset files — lightweight, GUID-named
- Each file is named `<guid>.db` and contains only the results table + numpy type adapters — no QCoDeS metadata schema.
- Created via the shared `_create_run_table()` / `_connect_to_sqlite_file()` helpers so the schema and connection settings stay identical to the main DB.

### No empty results table in the main DB
- When the feature is enabled, **no results table is created in the main database at all** — it stores only run metadata. This mirrors how `DataSetInMem` (the netcdf-backed dataset) records runs (`create_run_table=False`).
- Split-storage runs are identified by a dedicated `raw_data_db_path` column in the `runs` table, recorded at **dataset creation**. This positively distinguishes them from `DataSetInMem` runs (which also have no results table) in `_get_datasetprotocol_from_guid`, and is future-proof for non-SQLite raw-data backends.
- `raw_data_db_path` is an internal storage detail: it is stored in the `runs` table but **excluded from the user-facing `metadata` dict**, and read/written through `get_raw_data_db_path_for_run` / `set_raw_data_db_path_for_run`. This keeps it out of metadata comparisons (`the_same_dataset_as`) and out of extracted/exported databases.

### Features that depend on the results table — handled explicitly
- **Counting** (`number_of_results` / `__len__`) return `0` when the results table does not exist (e.g. a pristine, not-yet-started split dataset).
- **Subscribers**: a subscription requested before the dataset is started (before the raw table exists) is **deferred** and materialised at start time. Factored into `_queue_pending_subscriber` / `_create_and_start_subscriber` / `_start_pending_subscribers`.
- **Subscriber triggers**, `unsubscribe`, `unsubscribe_all` operate on `_data_conn` so they target the DB that actually holds the results table.

### Extract / export compatibility
- `extract_runs_into_db` reads results via `_data_conn` (from the per-dataset file), and the copied run does not carry over the source's `raw_data_db_path`.
- `export_datasets_and_create_metadata_db` works: the NetCDF path goes through `get_parameter_data` (already routed), and the copy-as-is fallback uses the extract fix above.

### Shared connection helpers (deduplication)
- Extracted `_connect_to_sqlite_file()` and `_register_numpy_sqlite_adapters_and_converters()` in `sqlite/database.py`; both `connect()` and `connect_to_raw_data_db()` reuse them, so the numpy adapters and connection settings are defined once.

## Dataset management helpers

Exposed via `qcodes.dataset`:

- **`update_raw_data_paths(db_path, new_raw_data_folder)`** — update stored paths after per-dataset files have been moved to a new folder.
- **`purge_orphaned_datasets(db_path, *, dry_run=True)`** — remove dataset records from the main DB whose raw data files no longer exist on disk.
- **`cleanup_datasets(db_path, *, older_than_days=None, sample_name=None, larger_than_mb=None, dry_run=True)`** — remove datasets (DB records AND raw data files) matching criteria (AND logic).

Both purge and cleanup default to `dry_run=True` and return dataclass results with full details. Dataset removal reuses the existing `remove_dataset_from_db` / `_validate_table_name` machinery.

## Files changed

| File | Type | Description |
|------|------|-------------|
| `src/qcodes/dataset/_raw_data_storage.py` | **New** | Storage config helpers, `create_raw_data_db()`, `connect_to_raw_data_db()`, `update_raw_data_paths()`, `purge_orphaned_datasets()`, `cleanup_datasets()`, result dataclasses |
| `src/qcodes/dataset/data_set.py` | Modified | `_data_conn`/`_results_table_exists` properties, `_raw_data_db_path`, routing in `__init__`, `_perform_start_actions`, `add_results`, `get_parameter_data`, `number_of_results`, `__len__`, `_BackgroundWriter`, deferred subscribers, `unsubscribe`/`unsubscribe_all`, split-vs-inmem disambiguation |
| `src/qcodes/dataset/sqlite/queries.py` | Modified | `get_raw_data_db_path_for_run`/`set_raw_data_db_path_for_run`, `RawDataDatasetRecord` dataclass + `get_datasets_with_raw_data_path`, `remove_dataset_from_db`, `RAW_DATA_DB_PATH_COLUMN` excluded from metadata |
| `src/qcodes/dataset/sqlite/database.py` | Modified | Shared `_connect_to_sqlite_file()` + `_register_numpy_sqlite_adapters_and_converters()` reused by `connect()` |
| `src/qcodes/dataset/database_extract_runs.py` | Modified | `_extract_single_dataset_into_db` reads via `_data_conn` |
| `src/qcodes/dataset/data_set_cache.py` | Modified | `load_data_from_db()` uses `_data_conn` |
| `src/qcodes/dataset/subscriber.py` | Modified | Trigger creation uses `_data_conn` |
| `src/qcodes/dataset/__init__.py` | Modified | Exports the new public helpers |
| `src/qcodes/configuration/qcodesrc.json` + `qcodesrc_schema.json` | Modified | New config defaults + schema |
| `docs/dataset/introduction.rst`, `docs/dataset/dataset_design.rst`, `docs/examples/DataSet/Database.ipynb` | Modified | Feature + management docs, design notes |
| `docs/changes/newsfragments/8219.new` | **New** | Towncrier news fragment |
| `tests/dataset/test_raw_data_storage.py` | **New** | Unit + integration + management + extract/export + no-empty-table tests |

## Verification

- `tests/dataset/test_raw_data_storage.py`: all pass (unit, integration, management, extract/export, subscribe-before-start, count-before-start, no-main-table).
- Full `tests/dataset` suite in default mode: **no regressions**.
- Validated with split storage globally enabled (temporary local harness, not part of the PR): remaining failures are all expected/inherent (tests that assert disabled-mode behaviour, low-level tests that call `queries.get_parameter_data(ds.conn, ...)` directly bypassing the DataSet API, and db-overview/guid-scan tests that inherently assume data in the main DB).
- Pyright: 0 errors. Ruff check + format: pass. pre-commit: pass.

### DIFF
--- a/docs/changes/newsfragments/8219.new
+++ b/docs/changes/newsfragments/8219.new
@@ -1,0 +1,14 @@
+Added **split raw data storage** for the DataSet: an opt-in feature
+(``dataset.raw_data_to_separate_db``) that writes raw measurement data
+into individual per-dataset SQLite files while keeping all metadata in
+the main database.  This prevents the main DB file from growing
+excessively large.
+
+Management helpers included:
+
+- ``update_raw_data_paths()``: update stored paths after raw data files
+  have been moved to a new folder.
+- ``purge_orphaned_datasets()``: find and remove dataset records whose
+  raw data files no longer exist on disk (``dry_run=True`` by default).
+- ``cleanup_datasets()``: remove datasets (DB records and raw data files)
+  by age, sample name, or file size (``dry_run=True`` by default).
--- a/docs/changes/newsfragments/8219.new
+++ b/docs/changes/newsfragments/8219.new
@@ -1,8 +1,10 @@
 Added **split raw data storage** for the DataSet: an opt-in feature
-(``dataset.raw_data_to_separate_db``) that writes raw measurement data
-into individual per-dataset SQLite files while keeping all metadata in
-the main database.  This prevents the main DB file from growing
-excessively large.
+enabled via the ``dataset.raw_data_to_separate_db`` config option that
+writes raw measurement data into individual per-dataset SQLite files
+while keeping all metadata in the main database.  This prevents the
+main DB file from growing excessively large and allows users to have a
+single main SQLite QCoDeS database on their PC as opposed to many that
+grow infinitely.
 
 Management helpers included:
 
--- a/docs/changes/newsfragments/8219.new
+++ b/docs/changes/newsfragments/8219.new
@@ -4,7 +4,9 @@ writes raw measurement data into individual per-dataset SQLite files
 while keeping all metadata in the main database.  This prevents the
 main DB file from growing excessively large and allows users to have a
 single main SQLite QCoDeS database on their PC as opposed to many that
-grow infinitely.
+grow infinitely.  When enabled, no results table is created in the main
+database at all - it holds only metadata - and all public ``DataSet``
+APIs continue to work unchanged.
 
 Management helpers included:
 
--- a/docs/dataset/dataset_design.rst
+++ b/docs/dataset/dataset_design.rst
@@ -75,3 +75,33 @@ We note that the dataset currently exclusively supports storing data in an
 SQLite database. This is not an intrinsic limitation of the dataset and
 measurement layer. It is possible that at a future state support for writing
 to a different backend will be added.
+
+.. _sec:design_split_storage:
+
+Split Raw Data Storage
+======================
+
+As the main SQLite database grows with many datasets, browsing experiments and
+loading metadata can become slower due to the file size. To address this,
+QCoDeS supports an optional **split raw data storage** mode (see
+:ref:`sec:intro_split_raw_data` for user-facing details).
+
+From a design perspective, this feature adds a thin routing layer inside the
+``DataSet`` class without changing any public interfaces:
+
+- A ``_data_conn`` property transparently returns either the main database
+  connection or a per-dataset raw data connection, depending on the
+  configuration.
+- Write paths (``add_results``, ``_BackgroundWriter``) and read paths
+  (``get_parameter_data``, ``DataSetCacheWithDBBackend``, ``number_of_results``,
+  ``__len__``) all go through this single routing point.
+- The per-dataset SQLite file is a lightweight database containing only the
+  results table and numpy type adapters -- no QCoDeS metadata schema.
+- Subscriber triggers (used for real-time data callbacks) are created on the
+  data connection so that they fire regardless of which database holds the
+  results table.
+
+The implementation is contained in ``qcodes.dataset.raw_data_storage`` (helper
+functions) and a handful of additions to ``qcodes.dataset.data_set`` (routing
+logic). The ``Measurement`` context manager, ``DataSaver``, and all export
+functions work without modification.
--- a/docs/dataset/dataset_design.rst
+++ b/docs/dataset/dataset_design.rst
@@ -81,8 +81,8 @@ to a different backend will be added.
 Split Raw Data Storage
 ======================
 
-As the main SQLite database grows with many datasets, browsing experiments and
-managing the database file can become inconvenient due to the file size. To address this,
+As the main SQLite database grows with many datasets, managing the database
+file can become inconvenient due to the file size. To address this,
 QCoDeS supports an optional **split raw data storage** mode (see
 :ref:`sec:intro_split_raw_data` for user-facing details).
 
@@ -101,7 +101,7 @@ From a design perspective, this feature adds a thin routing layer inside the
   data connection so that they fire regardless of which database holds the
   results table.
 
-The implementation is contained in ``qcodes.dataset.raw_data_storage`` (helper
+The implementation is contained in ``qcodes.dataset._raw_data_storage`` (helper
 functions) and a handful of additions to ``qcodes.dataset.data_set`` (routing
 logic). The ``Measurement`` context manager, ``DataSaver``, and all export
 functions work without modification.
--- a/docs/dataset/dataset_design.rst
+++ b/docs/dataset/dataset_design.rst
@@ -82,7 +82,7 @@ Split Raw Data Storage
 ======================
 
 As the main SQLite database grows with many datasets, browsing experiments and
-loading metadata can become slower due to the file size. To address this,
+managing the database file can become inconvenient due to the file size. To address this,
 QCoDeS supports an optional **split raw data storage** mode (see
 :ref:`sec:intro_split_raw_data` for user-facing details).
 
--- a/docs/dataset/dataset_design.rst
+++ b/docs/dataset/dataset_design.rst
@@ -97,9 +97,18 @@ From a design perspective, this feature adds a thin routing layer inside the
   ``__len__``) all go through this single routing point.
 - The per-dataset SQLite file is a lightweight database containing only the
   results table and numpy type adapters -- no QCoDeS metadata schema.
+- When raw data storage is enabled, **no results table is created in the main
+  database** at all -- only the run metadata is stored there. This mirrors how
+  ``DataSetInMem`` records runs (with ``create_run_table=False``). A run is
+  identified as a split-storage dataset by the ``raw_data_db_path`` column in
+  the ``runs`` table (recorded at dataset creation), which is used both to
+  reconnect to the raw data file and to distinguish such runs from
+  ``DataSetInMem`` runs (which also have no results table). This column is an
+  internal storage detail and is kept out of the user-facing metadata.
 - Subscriber triggers (used for real-time data callbacks) are created on the
-  data connection so that they fire regardless of which database holds the
-  results table.
+  data connection. Because the results table only exists once the dataset is
+  started (in the per-dataset file), subscriptions requested before the dataset
+  is started are deferred and materialised at start time.
 
 The implementation is contained in ``qcodes.dataset._raw_data_storage`` (helper
 functions) and a handful of additions to ``qcodes.dataset.data_set`` (routing
--- a/docs/dataset/introduction.rst
+++ b/docs/dataset/introduction.rst
@@ -81,7 +81,7 @@ For dataset operations, QCoDeS provides functions for:
 Split Raw Data Storage
 ======================
 
-By default, all measurement data (the results table rows) is stored in the same SQLite database alongside metadata such as experiments, runs, parameter layouts, and dependencies. Over time, the main database file can grow very large, which can slow down operations like browsing experiments and loading metadata.
+By default, all measurement data (the results table rows) is stored in the same SQLite database alongside metadata such as experiments, runs, parameters and their dependencies. Over time, the main database file can grow very large, which can slow down operations like browsing experiments and make managing a large database file inconvenient.
 
 QCoDeS supports an optional **split raw data storage** mode in which the actual measurement data for each ``DataSet`` is written to an individual, per-dataset SQLite file while all metadata remains in the main database. Each per-dataset file is named after the dataset's GUID (e.g. ``<guid>.db``) and is stored in a configurable folder.
 
@@ -123,6 +123,4 @@ QCoDeS provides helpers for managing per-dataset raw data files over time:
 - :func:`~qcodes.dataset.purge_orphaned_datasets` — remove dataset records whose raw data files no longer exist on disk.
 - :func:`~qcodes.dataset.cleanup_datasets` — remove datasets (DB records and raw data files) by age, sample name, or file size.
 
-For usage examples, see the :doc:`Database notebook <../examples/DataSet/Database>`.
-
-For more details on database management, see the :doc:`Database notebook <../examples/DataSet/Database>`.
+For usage examples and more details on database management, see the :doc:`Database notebook <../examples/DataSet/Database>`.
--- a/docs/dataset/introduction.rst
+++ b/docs/dataset/introduction.rst
@@ -104,4 +104,15 @@ Example runtime configuration::
     qc.config.dataset.raw_data_to_separate_db = True
     qc.config.dataset.raw_data_path = "/data/raw_measurements/"
 
+If the per-dataset raw data files are moved to a different folder (e.g. during data migration or archival), the stored paths in the main database will become stale. Use the :func:`~qcodes.dataset.update_raw_data_paths` helper to update them::
+
+    from qcodes.dataset import update_raw_data_paths
+
+    update_raw_data_paths(
+        db_path="/path/to/main_database.db",
+        new_raw_data_folder="/new/location/of/raw_files/"
+    )
+
+This scans all datasets with a ``raw_data_db_path`` metadata entry, checks whether the corresponding ``.db`` file exists in the new folder, and updates the stored path accordingly.
+
 For more details on database management, see the :doc:`Database notebook <../examples/DataSet/Database>`.
--- a/docs/dataset/introduction.rst
+++ b/docs/dataset/introduction.rst
@@ -75,3 +75,33 @@ For dataset operations, QCoDeS provides functions for:
 - **Exporting datasets**: :doc:`Exporting data to other file formats <../examples/DataSet/Exporting-data-to-other-file-formats>`
 - **Extracting runs between databases**: :doc:`Extracting runs from one DB file to another <../examples/DataSet/Extracting-runs-from-one-DB-file-to-another>` and :func:`qcodes.dataset.extract_runs_into_db`
 - **Bulk export and metadata-only databases**: :func:`qcodes.dataset.export_datasets_and_create_metadata_db` for creating lightweight metadata-only databases while exporting all data to NetCDF files
+
+.. _sec:intro_split_raw_data:
+
+Split Raw Data Storage
+======================
+
+By default, all measurement data (the results table rows) is stored in the same SQLite database alongside metadata such as experiments, runs, parameter layouts, and dependencies. Over time, the main database file can grow very large, which can slow down operations like browsing experiments and loading metadata.
+
+QCoDeS supports an optional **split raw data storage** mode in which the actual measurement data for each ``DataSet`` is written to an individual, per-dataset SQLite file while all metadata remains in the main database. Each per-dataset file is named after the dataset's GUID (e.g. ``<guid>.db``) and is stored in a configurable folder.
+
+This feature is controlled by two configuration options in ``qcodesrc.json``:
+
+- ``dataset.raw_data_to_separate_db`` (bool, default ``false``): enables or disables split storage.
+- ``dataset.raw_data_path`` (string, default ``"{db_location}"``): the folder where per-dataset files are created. The ``{db_location}`` placeholder is expanded to a folder derived from the main database path (e.g. ``~/experiments.db`` becomes ``~/experiments_db/``).
+
+When enabled:
+
+- The main database retains the full results table schema (column definitions) but no data rows are written to it, keeping it lightweight.
+- All ``INSERT`` and ``SELECT`` operations on results data are transparently routed to the per-dataset file.
+- The path to the per-dataset file is persisted in the run's metadata (``raw_data_db_path``), so ``load_by_id`` and related loading functions automatically reconnect to the correct file.
+- All public ``DataSet`` APIs (``get_parameter_data``, ``to_pandas_dataframe``, ``to_xarray_dataset``, ``cache``, ``export``, etc.) work identically whether split storage is enabled or not.
+
+Example runtime configuration::
+
+    import qcodes as qc
+
+    qc.config.dataset.raw_data_to_separate_db = True
+    qc.config.dataset.raw_data_path = "/data/raw_measurements/"
+
+For more details on database management, see the :doc:`Database notebook <../examples/DataSet/Database>`.
--- a/docs/dataset/introduction.rst
+++ b/docs/dataset/introduction.rst
@@ -115,4 +115,37 @@ If the per-dataset raw data files are moved to a different folder (e.g. during d
 
 This scans all datasets with a ``raw_data_db_path`` metadata entry, checks whether the corresponding ``.db`` file exists in the new folder, and updates the stored path accordingly.
 
+Purging Orphaned Datasets
+-------------------------
+
+After archiving or deleting selected raw data files, some dataset records in the main database may reference files that no longer exist on disk. Use :func:`~qcodes.dataset.purge_orphaned_datasets` to identify and remove those orphaned records::
+
+    from qcodes.dataset import purge_orphaned_datasets
+
+    # First, see what would be removed (dry run — the default)
+    result = purge_orphaned_datasets("/path/to/main_database.db")
+    print(f"{len(result.orphaned_datasets)} orphaned datasets found")
+
+    # Then actually remove them
+    result = purge_orphaned_datasets("/path/to/main_database.db", dry_run=False)
+
+Cleaning Up Datasets by Criteria
+---------------------------------
+
+To free disk space by removing datasets (both the raw data files **and** their records in the main database) based on age, sample name, or file size, use :func:`~qcodes.dataset.cleanup_datasets`::
+
+    from qcodes.dataset import cleanup_datasets
+
+    # Remove datasets older than 30 days (dry run first)
+    result = cleanup_datasets("/path/to/db.db", older_than_days=30)
+    print(f"{len(result.matching_datasets)} datasets would be removed")
+
+    # Remove datasets for a specific sample
+    result = cleanup_datasets("/path/to/db.db", sample_name="old-sample", dry_run=False)
+
+    # Remove datasets with raw data larger than 500 MB
+    result = cleanup_datasets("/path/to/db.db", larger_than_mb=500, dry_run=False)
+
+Criteria are combined with AND logic — a dataset must match **all** specified criteria to be selected for removal.
+
 For more details on database management, see the :doc:`Database notebook <../examples/DataSet/Database>`.
--- a/docs/dataset/introduction.rst
+++ b/docs/dataset/introduction.rst
@@ -115,37 +115,14 @@ If the per-dataset raw data files are moved to a different folder (e.g. during d
 
 This scans all datasets with a ``raw_data_db_path`` metadata entry, checks whether the corresponding ``.db`` file exists in the new folder, and updates the stored path accordingly.
 
-Purging Orphaned Datasets
--------------------------
+Managing Datasets
+-----------------
 
-After archiving or deleting selected raw data files, some dataset records in the main database may reference files that no longer exist on disk. Use :func:`~qcodes.dataset.purge_orphaned_datasets` to identify and remove those orphaned records::
+QCoDeS provides helpers for managing per-dataset raw data files over time:
 
-    from qcodes.dataset import purge_orphaned_datasets
+- :func:`~qcodes.dataset.purge_orphaned_datasets` — remove dataset records whose raw data files no longer exist on disk.
+- :func:`~qcodes.dataset.cleanup_datasets` — remove datasets (DB records and raw data files) by age, sample name, or file size.
 
-    # First, see what would be removed (dry run — the default)
-    result = purge_orphaned_datasets("/path/to/main_database.db")
-    print(f"{len(result.orphaned_datasets)} orphaned datasets found")
-
-    # Then actually remove them
-    result = purge_orphaned_datasets("/path/to/main_database.db", dry_run=False)
-
-Cleaning Up Datasets by Criteria
----------------------------------
-
-To free disk space by removing datasets (both the raw data files **and** their records in the main database) based on age, sample name, or file size, use :func:`~qcodes.dataset.cleanup_datasets`::
-
-    from qcodes.dataset import cleanup_datasets
-
-    # Remove datasets older than 30 days (dry run first)
-    result = cleanup_datasets("/path/to/db.db", older_than_days=30)
-    print(f"{len(result.matching_datasets)} datasets would be removed")
-
-    # Remove datasets for a specific sample
-    result = cleanup_datasets("/path/to/db.db", sample_name="old-sample", dry_run=False)
-
-    # Remove datasets with raw data larger than 500 MB
-    result = cleanup_datasets("/path/to/db.db", larger_than_mb=500, dry_run=False)
-
-Criteria are combined with AND logic — a dataset must match **all** specified criteria to be selected for removal.
+For usage examples, see the :doc:`Database notebook <../examples/DataSet/Database>`.
 
 For more details on database management, see the :doc:`Database notebook <../examples/DataSet/Database>`.
--- a/docs/dataset/introduction.rst
+++ b/docs/dataset/introduction.rst
@@ -92,9 +92,9 @@ This feature is controlled by two configuration options in ``qcodesrc.json``:
 
 When enabled:
 
-- The main database retains the full results table schema (column definitions) but no data rows are written to it, keeping it lightweight.
+- No results table is created in the main database; it stores only the run metadata, keeping it lightweight.
 - All ``INSERT`` and ``SELECT`` operations on results data are transparently routed to the per-dataset file.
-- The path to the per-dataset file is persisted in the run's metadata (``raw_data_db_path``), so ``load_by_id`` and related loading functions automatically reconnect to the correct file.
+- The path to the per-dataset file is recorded in the ``runs`` table, so ``load_by_id`` and related loading functions automatically reconnect to the correct file.
 - All public ``DataSet`` APIs (``get_parameter_data``, ``to_pandas_dataframe``, ``to_xarray_dataset``, ``cache``, ``export``, etc.) work identically whether split storage is enabled or not.
 
 Example runtime configuration::
--- a/src/qcodes/dataset/_raw_data_storage.py
+++ b/src/qcodes/dataset/_raw_data_storage.py
@@ -209,9 +209,7 @@ def update_raw_data_paths(
     if not db_path.is_file():
         raise FileNotFoundError(f"Database file not found: {db_path}")
     if not new_raw_data_folder.is_dir():
-        raise FileNotFoundError(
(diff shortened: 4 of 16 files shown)


src/qcodes/dataset/data_set.py
```diff
@@ -290,6 +316,13 @@ def __init__(
             self._export_info = ExportInfo.from_str(
                 self.metadata.get("export_info", "")
             )
+            # If this dataset was saved with raw data in a separate db,
+            # re-open that connection for reads.
+            raw_db_path = self._metadata.get("raw_data_db_path")
+            if raw_db_path is not None and Path(raw_db_path).is_file():
```
jenshnielsen: What is the error mode here if the file has been deleted?

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -0,0 +1,174 @@
+"""
+Module for managing per-dataset raw data SQLite files.
```
jenshnielsen, line 2: I think this module should be private `_raw_data_storage.py`

src/qcodes/dataset/data_set.py
astafan8: Fixed in f9a3ea4. Now raises FileNotFoundError with a descriptive message when the per-dataset raw data file is missing on load, instead of silently falling back to the empty main DB table. Also added test_missing_raw_data_file_raises to verify this behavior.

src/qcodes/dataset/_raw_data_storage.py
astafan8: Done in f9a3ea4. Renamed to _raw_data_storage.py and updated all imports.

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -0,0 +1,174 @@
...
+    conn = connect_to_raw_data_db(path)
+
+    if paramspecs:
+        columns = ",".join(p.sql_repr() for p in paramspecs)
+        sql = f"""
+        CREATE TABLE IF NOT EXISTS "{table_name}" (
+            id INTEGER PRIMARY KEY,
+            {columns}
+        );
+        """
```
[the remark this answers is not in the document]
astafan8: Good catch. Fixed in f9a3ea4 - now uses quoted identifiers instead of sql_repr() when building the CREATE TABLE statement in create_raw_data_db.

tests/dataset/test_raw_data_storage.py
```diff
@@ -0,0 +1,315 @@
...
+        """When split is enabled, DataSet should have a raw data connection."""
+        ds = new_data_set("test-split")
+        x = ParamSpecBase("x", "numeric")
+        y = ParamSpecBase("y", "numeric")
+        idps = InterDependencies_(dependencies={y: (x,)})
+        ds.set_interdependencies(idps)
+        ds.mark_started()
+        assert ds._raw_data_conn is not None
+        ds.conn.close()
+
```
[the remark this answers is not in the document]
astafan8: Fixed in f9a3ea4. Added a _close_ds() helper that closes both _raw_data_conn and conn, and updated all test methods to use it.

src/qcodes/dataset/data_set.py
```diff
@@ -290,6 +316,13 @@ def __init__(
...
+                self._raw_data_conn = connect_to_raw_data_db(
+                    raw_db_path, read_only=read_only
+                )
```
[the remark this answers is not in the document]
astafan8: Fixed in f9a3ea4. Now raises FileNotFoundError with a clear message instead of silently falling back. Added test_missing_raw_data_file_raises to verify.

astafan8:
## Added: update_raw_data_paths helper

Added a utility function for users who move their per-dataset raw data files to a new location. This mirrors the pattern used for exported netCDF files.

**Usage:**

`python
from qcodes.dataset import update_raw_data_paths

update_raw_data_paths(
    db_path="/path/to/main_database.db",
    new_raw_data_folder="/new/location/of/raw_files/"
)
`

The function:
- Scans all datasets in the main DB that have raw_data_db_path metadata
- For each, checks if the corresponding .db file exists in the new folder
- Updates the stored path to point to the new location
- Logs warnings for any files not found in the new folder

4 tests added, documentation updated in introduction.rst and Database.ipynb.

docs/dataset/introduction.rst
```diff
@@ -115,4 +115,37 @@ If the per-dataset raw data files are moved to a different folder (e.g. during d
...
+    from qcodes.dataset import purge_orphaned_datasets
+
+    # First, see what would be removed (dry run — the default)
+    result = purge_orphaned_datasets("/path/to/main_database.db")
+    print(f"{len(result.orphaned_datasets)} orphaned datasets found")
+
+    # Then actually remove them
+    result = purge_orphaned_datasets("/path/to/main_database.db", dry_run=False)
+
+Cleaning Up Datasets by Criteria
```
astafan8: please don't copy this information here, just make a reference to this same info in the database ipy notebook. as it's done above for exporting datasets and such.

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -258,3 +261,357 @@ def update_raw_data_paths(
...
+
+def _get_datasets_with_raw_data(
+    conn: AtomicConnection,
+) -> list[DatasetInfo]:
+    """Query all datasets that have a raw_data_db_path metadata entry."""
+    if not is_column_in_table(conn, "runs", "raw_data_db_path"):
+        return []
+
+    sql = """
+    SELECT r.run_id, r.guid, e.name, e.sample_name,
```
astafan8: move these query kind of stuff into qcodes.dataset.sqlite.queries module so that they can be re-used across qcodes as needed.

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -258,3 +261,357 @@ def update_raw_data_paths(
...
+            table_name,
+            raw_path,
+        ) = row
+
+        raw_size: int | None = None
+        if raw_path and Path(raw_path).is_file():
+            raw_size = Path(raw_path).stat().st_size
+
+        datasets.append(
+            DatasetInfo(
```
astafan8, line 285: instead of doing this, i think there's a way to give the sqlite python module this dataclass as the row instantiator or something so that when you iterate rows on a cursor it actually already returns these objects. this can save amount of lines of code.

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -258,3 +261,357 @@ def update_raw_data_paths(
...
+        raise FileNotFoundError(f"Database file not found: {db_path}")
+
+    conn = connect(str(db_path))
+    all_datasets = _get_datasets_with_raw_data(conn)
+
+    # Find orphaned datasets: raw_data_db_path is set but the file is missing
+    orphaned = [
+        ds
+        for ds in all_datasets
+        if ds.raw_data_db_path and not Path(ds.raw_data_db_path).is_file()
```
astafan8: would you like to rely on the fact that _get_datasets_with_raw_data returns raw_szie as None if it was not able to find the file? or would you like to do the is-file check here "again"? see pros/cons and make a decsion.

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -258,3 +261,357 @@ def update_raw_data_paths(
...
+    ]
+
+    log.info(
+        "Found %d datasets with raw data references, %d orphaned (file missing).",
+        len(all_datasets),
```
astafan8: also add DB file path to this logging

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -258,3 +261,357 @@ def update_raw_data_paths(
...
+                log.info(
+                    "Removed orphaned dataset run_id=%d (guid=%s) from database.",
+                    ds_info.run_id,
+                    ds_info.guid,
+                )
+            except Exception as exc:
+                log.error("Failed to remove run_id=%d: %s", ds_info.run_id, exc)
+                errors.append((ds_info.run_id, exc))
+
+    conn.close()
```
astafan8: instead of this use, the "with closing(conn):" context manager

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -258,3 +261,357 @@ def update_raw_data_paths(
...
+        orphaned_datasets=orphaned,
+        removed_datasets=removed,
+        dry_run=dry_run,
+        errors=errors,
+    )
+
+    if dry_run:
+        log.info("Dry run: %d orphaned datasets would be removed.", len(orphaned))
+    else:
+        log.info("Removed %d orphaned datasets.", len(removed))
```
astafan8: for these two please also add a print to console.

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -258,3 +261,357 @@ def update_raw_data_paths(
...
+    sample_name: str | None = None,
+    larger_than_mb: float | None = None,
+    dry_run: bool = True,
+) -> CleanupResult:
+    """Remove datasets (DB records and raw data files) matching given criteria.
+
+    This function helps manage disk space by removing datasets that match
+    one or more of the specified criteria. It removes both the raw data
+    SQLite file on disk and the corresponding records in the main database
+    (run metadata, layouts, dependencies, results table schema).
```
astafan8: no need to mention "    (run metadata, layouts, dependencies, results table schema)."

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -258,3 +261,357 @@ def update_raw_data_paths(
...
+            if (
+                ds.raw_data_size_bytes is None
+                or ds.raw_data_size_bytes <= size_threshold_bytes
+            ):
+                continue
+
+        matching.append(ds)
+
+    log.info(
+        "Found %d datasets with raw data, %d match cleanup criteria.",
```
astafan8: also add a print statememt for this

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -258,3 +261,357 @@ def update_raw_data_paths(
...
+                log.info(
+                    "Removed dataset run_id=%d (guid=%s) from database.",
+                    ds_info.run_id,
+                    ds_info.guid,
+                )
+            except Exception as exc:
+                log.error("Failed to remove run_id=%d: %s", ds_info.run_id, exc)
+                errors.append((ds_info.run_id, exc))
+
+    conn.close()
```
astafan8: use with closing, as mentioned above

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -258,3 +261,357 @@ def update_raw_data_paths(
...
+        total_size = sum(
+            ds.raw_data_size_bytes for ds in matching if ds.raw_data_size_bytes
+        )
+        log.info(
+            "Dry run: %d datasets (%d bytes) would be removed.",
+            len(matching),
+            total_size,
+        )
+    else:
+        log.info("Removed %d datasets, freed %d bytes.", len(removed), total_freed)
```
astafan8: also add a print for these messages as they are useful for the users to see

docs/changes/newsfragments/8219.new
```diff
@@ -0,0 +1,14 @@
+Added **split raw data storage** for the DataSet: an opt-in feature
+(``dataset.raw_data_to_separate_db``) that writes raw measurement data
```
astafan8: mention the name of hte qcodes config option rather here.

docs/changes/newsfragments/8219.new
```diff
@@ -0,0 +1,14 @@
...
+into individual per-dataset SQLite files while keeping all metadata in
+the main database.  This prevents the main DB file from growing
+excessively large.
```
astafan8: ... and allows users to have a single main sqlite qcodes on their PC as opposed to many that grow infinitely.

docs/dataset/introduction.rst
astafan8: Addressed in c76baadf7.

src/qcodes/dataset/_raw_data_storage.py
astafan8: Addressed in c76baadf7.

src/qcodes/dataset/_raw_data_storage.py
astafan8: Addressed in c76baadf7.

src/qcodes/dataset/_raw_data_storage.py
astafan8: Addressed in c76baadf7.

src/qcodes/dataset/_raw_data_storage.py
astafan8: Addressed in c76baadf7.

src/qcodes/dataset/_raw_data_storage.py
astafan8: Addressed in c76baadf7.

src/qcodes/dataset/_raw_data_storage.py
astafan8: Addressed in c76baadf7.

src/qcodes/dataset/_raw_data_storage.py
astafan8: Addressed in c76baadf7.

src/qcodes/dataset/_raw_data_storage.py
astafan8: Addressed in c76baadf7.

src/qcodes/dataset/_raw_data_storage.py
astafan8: Addressed in c76baadf7.

src/qcodes/dataset/_raw_data_storage.py
astafan8: Addressed in c76baadf7.

docs/changes/newsfragments/8219.new
astafan8: Addressed in c76baadf7.

docs/changes/newsfragments/8219.new
astafan8: Addressed in c76baadf7.

src/qcodes/dataset/_raw_data_storage.py
astafan8: Note: I moved the query to queries.py as get_datasets_with_raw_data_path() which returns a list of tuples (run_id, guid, raw_data_db_path, sample_name, start_time). The DatasetInfo dataclass construction (with computed 
aw_data_size_bytes from checking file existence/size) is done in _build_dataset_info_list() in the storage module. Using sqlite3 row_factory for a dataclass with a computed field wasn't a clean fit, so I kept the explicit construction but simplified the unpacking.

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -423,40 +379,38 @@ def purge_orphaned_datasets(
...
+        errors: list[tuple[int, Exception]] = []
+
+        if not dry_run and orphaned:
+            for ds_info in orphaned:
+                try:
+                    remove_dataset_from_db(
+                        conn, ds_info.run_id, ds_info.result_table_name
+                    )
+                    removed.append(ds_info)
+                    log.info(
```
astafan8: use log.debug here, and for all lines where there's a report per each processed dataset

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -521,77 +476,79 @@ def cleanup_datasets(
...
+
+        if not dry_run and matching:
+            for ds_info in matching:
+                try:
+                    # Delete the raw data file from disk
+                    if ds_info.raw_data_db_path:
+                        raw_path = Path(ds_info.raw_data_db_path)
+                        if raw_path.is_file():
+                            file_size = raw_path.stat().st_size
+                            os.remove(raw_path)
```
astafan8: isn't shutil.rm/remove whatever more robust?

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -521,77 +476,79 @@ def cleanup_datasets(
...
+                            total_freed += file_size
+                            log.info("Deleted raw data file: %s", raw_path)
+
+                    # Remove dataset records from the main DB
+                    remove_dataset_from_db(
+                        conn, ds_info.run_id, ds_info.result_table_name
+                    )
+                    removed.append(ds_info)
+                    log.info(
```
astafan8: log.debug

src/qcodes/dataset/_raw_data_storage.py
astafan8: Addressed in 84e119d4e.

src/qcodes/dataset/_raw_data_storage.py
astafan8: Addressed in 84e119d4e.

src/qcodes/dataset/_raw_data_storage.py
astafan8: Addressed in 84e119d4e.

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -0,0 +1,571 @@
...
+
+        # Orphaned = raw_data_size_bytes is None (file not found on disk)
+        orphaned = [ds for ds in all_datasets if ds.raw_data_size_bytes is None]
+
+        msg = (
+            f"Found {len(all_datasets)} datasets with raw data references in {db_path}, "
+            f"{len(orphaned)} orphaned (file missing)."
+        )
+        log.info(msg)
+        print(msg)
```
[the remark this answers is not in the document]
astafan8: ok, but then the print of a similar message has to happen in a caller function that calls this library helper.

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -0,0 +1,571 @@
...
+        dry_run=dry_run,
+        errors=errors,
+    )
+
+    if dry_run:
+        msg = f"Dry run: {len(orphaned)} orphaned datasets would be removed from {db_path}."
+    else:
+        msg = f"Removed {len(removed)} orphaned datasets from {db_path}."
+    log.info(msg)
+    print(msg)
```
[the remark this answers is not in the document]
astafan8: ok, but then the print of a similar message has to happen in a caller function that calls this library helper.

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -0,0 +1,571 @@
...
+                    continue
+
+            matching.append(ds)
+
+        msg = (
+            f"Found {len(all_datasets)} datasets with raw data in {db_path}, "
+            f"{len(matching)} match cleanup criteria."
+        )
+        log.info(msg)
+        print(msg)
```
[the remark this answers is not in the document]
astafan8: ok, but then the print of a similar message has to happen in a caller function that calls this library helper.

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -0,0 +1,571 @@
...
+
+    if dry_run:
+        total_size = sum(
+            ds.raw_data_size_bytes for ds in matching if ds.raw_data_size_bytes
+        )
+        msg = f"Dry run: {len(matching)} datasets ({total_size} bytes) would be removed from {db_path}."
+    else:
+        msg = f"Removed {len(removed)} datasets from {db_path}, freed {total_freed} bytes."
+    log.info(msg)
+    print(msg)
```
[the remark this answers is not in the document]
astafan8: ok, but then the print of a similar message has to happen in a caller function that calls this library helper.

docs/dataset/dataset_design.rst
```diff
@@ -75,3 +75,33 @@ We note that the dataset currently exclusively supports storing data in an
...
 measurement layer. It is possible that at a future state support for writing
 to a different backend will be added.
+
+.. _sec:design_split_storage:
+
+Split Raw Data Storage
+======================
+
+As the main SQLite database grows with many datasets, browsing experiments and
+loading metadata can become slower due to the file size. To address this,
```
astafan8:
```suggestion
managing the database file can become inconvenient due to the file size. To address this,
```

docs/dataset/introduction.rst
```diff
@@ -75,3 +75,54 @@ For dataset operations, QCoDeS provides functions for:
 - **Exporting datasets**: :doc:`Exporting data to other file formats <../examples/DataSet/Exporting-data-to-other-file-formats>`
 - **Extracting runs between databases**: :doc:`Extracting runs from one DB file to another <../examples/DataSet/Extracting-runs-from-one-DB-file-to-another>` and :func:`qcodes.dataset.extract_runs_into_db`
 - **Bulk export and metadata-only databases**: :func:`qcodes.dataset.export_datasets_and_create_metadata_db` for creating lightweight metadata-only databases while exporting all data to NetCDF files
+
+.. _sec:intro_split_raw_data:
+
+Split Raw Data Storage
+======================
+
+By default, all measurement data (the results table rows) is stored in the same SQLite database alongside metadata such as experiments, runs, parameter layouts, and dependencies. Over time, the main database file can grow very large, which can slow down operations like browsing experiments and loading metadata.
```
astafan8:
```suggestion
By default, all measurement data (the results table rows) is stored in the same SQLite database alongside metadata such as experiments, runs, parameters and their dependencies. Over time, the main database file can grow very large, which can slow down operations like browsing experiments and make managing a large database file inconvenient.
```

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -0,0 +1,571 @@
+"""
+Module for managing per-dataset raw data SQLite files.
+
+When the ``dataset.raw_data_to_separate_db`` config option is enabled,
+measurement data (results tables) are written to individual SQLite files
+- one per dataset - instead of the main QCoDeS database file.  All metadata
+(runs, experiments, layouts, dependencies) remains in the main database.
```
astafan8:
```suggestion
(runs, experiments, parameters) remains in the main database.
```

src/qcodes/dataset/_raw_data_storage.py
```diff
@@ -0,0 +1,571 @@
...
+def connect_to_raw_data_db(
+    path: str | Path,
+    *,
+    read_only: bool = False,
+) -> AtomicConnection:
+    """Open (or create) a lightweight SQLite connection for raw data.
+
+    Unlike the main QCoDeS :func:`~qcodes.dataset.sqlite.database.connect`,
+    this does **not** create the full metadata schema (experiments, runs, ...).
+    It only registers the numpy/sqlite type adapters that QCoDeS needs to
```
astafan8: ok, so instead of copying this over here from connect funcition, can't you extract that part into a function and re-use it in here and in connect? the same is true for the rest of this function

astafan8: Addressed in ae3c879a6. Extracted the shared logic into `connect_to_sqlite_file()` and `_register_numpy_sqlite_adapters_and_converters()` in `sqlite/database.py`, and now both `connect()` and `connect_to_raw_data_db()` reuse them. I also reused the existing `_create_run_table()` in `create_raw_data_db()` (instead of duplicating the results-table CREATE TABLE logic), and reused `_validate_table_name()` before the DROP TABLE in `remove_dataset_from_db()`.

src/qcodes/dataset/sqlite/queries.py
```diff
@@ -2307,3 +2307,72 @@ def _get_result_table_name_by_guid(conn: AtomicConnection, guid: str) -> str:
...
+                f" OR independent IN ({placeholders})",
+                *layout_ids,
+                *layout_ids,
+            )
+
+        # Delete layouts
+        transaction(aconn, "DELETE FROM layouts WHERE run_id = ?", run_id)
+
+        # Drop the results table in the DB (if it exists)
+        transaction(aconn, f'DROP TABLE IF EXISTS "{result_table_name}"')
```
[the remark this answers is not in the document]
astafan8: Addressed in ae3c879a6 - now validates the table name via `_validate_table_name()` before the DROP TABLE.

docs/dataset/dataset_design.rst
```diff
@@ -75,3 +75,33 @@ We note that the dataset currently exclusively supports storing data in an
...
+- Write paths (``add_results``, ``_BackgroundWriter``) and read paths
+  (``get_parameter_data``, ``DataSetCacheWithDBBackend``, ``number_of_results``,
+  ``__len__``) all go through this single routing point.
+- The per-dataset SQLite file is a lightweight database containing only the
+  results table and numpy type adapters -- no QCoDeS metadata schema.
+- Subscriber triggers (used for real-time data callbacks) are created on the
+  data connection so that they fire regardless of which database holds the
+  results table.
+
+The implementation is contained in ``qcodes.dataset.raw_data_storage`` (helper
```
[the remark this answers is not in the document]
astafan8: Addressed in ae3c879a6 - fixed the module path to `qcodes.dataset._raw_data_storage`.

docs/dataset/introduction.rst
```diff
@@ -75,3 +75,54 @@ For dataset operations, QCoDeS provides functions for:
...
+-----------------
+
+QCoDeS provides helpers for managing per-dataset raw data files over time:
+
+- :func:`~qcodes.dataset.purge_orphaned_datasets` — remove dataset records whose raw data files no longer exist on disk.
+- :func:`~qcodes.dataset.cleanup_datasets` — remove datasets (DB records and raw data files) by age, sample name, or file size.
+
+For usage examples, see the :doc:`Database notebook <../examples/DataSet/Database>`.
+
+For more details on database management, see the :doc:`Database notebook <../examples/DataSet/Database>`.
```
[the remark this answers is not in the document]
astafan8: Addressed in ae3c879a6 - collapsed into a single sentence.

src/qcodes/dataset/sqlite/database.py
```diff
@@ -120,44 +120,92 @@ def _adapt_complex(value: complex | np.complexfloating) -> sqlite3.Binary:
...
-    Connect or create  database. If debug the queries will be echoed back.
-    This function takes care of registering the numpy/sqlite type
-    converters that we need.
+    Open a bare :class:`AtomicConnection` to an sqlite file with the settings
+    that QCoDeS relies on (URI mode, ``PARSE_DECLTYPES``) and the numpy type
+    adapters/converters registered.
+
+    Unlike :func:`connect`, this does **not** create or upgrade any QCoDeS
+    schema; it is the low-level building block shared by :func:`connect` and
+    the per-dataset raw-data connection helper.
```
jenshnielsen, line 160: Then perhaps it should be private?

src/qcodes/dataset/sqlite/queries.py
```diff
@@ -2307,3 +2307,76 @@ def _get_result_table_name_by_guid(conn: AtomicConnection, guid: str) -> str:
     sql = "SELECT result_table_name FROM runs WHERE guid=?"
     formatted_name = one(transaction(conn, sql, guid), "result_table_name")
     return formatted_name
+
+
+def get_datasets_with_raw_data_path(
+    conn: AtomicConnection,
+) -> list[tuple[int, str, str, str, float | None, float | None, str, str]]:
```
jenshnielsen: replace this type with a named tuple or dataclass?

astafan8:
and idea: why don't we leverage the export-info infra, add another export type qcodes-raw-sqlite-db with .db extension isntead of adding a new raw_data_db_path column. what would be the pros/cons of that?

astafan8:
an idea: should this implementation be part of DataSet class or should we add DataSetInSeparateSqliteDbFile class, or a DataSetInSeparateFile class with sqlite-db implementation and a place for further storage implementations? what would be pros-cons of those?

src/qcodes/dataset/sqlite/queries.py
astafan8: Done in 198f91963. Now returns a `RawDataDatasetRecord` dataclass instead of the opaque tuple.

src/qcodes/dataset/sqlite/database.py
astafan8: Agreed - made it private (`_connect_to_sqlite_file`) in 198f91963. It's only used internally by `connect()` and `connect_to_raw_data_db()`.

astafan8:
**Re: reusing the export-info infra (a `qcodes-raw-sqlite-db` export type) instead of a `raw_data_db_path` column**

I researched this and I'd recommend **against** it. Findings:

1. **Semantic mismatch.** `export_info`/`export_paths` is a registry of *post-acquisition exported snapshots*, produced by the one-shot `DataSet.export()` flow (`to_xarray`/`to_pandas` → write a file → record the path). Split raw storage instead writes results **live during acquisition** via `add_results`/the background writer. It is a *storage backend*, not an *export*. Overloading `export_info` would conflate two genuinely different concepts (and e.g. `load_from_exported_file` / `_get_datasetprotocol_from_guid` routing would get tangled between "load an exported netcdf snapshot" and "route live reads to the raw DB").

2. **`DataExportType` is a hard-coded, non-extensible `enum.Enum`** (`export_config.py`, values `nc`/`csv`). Adding `qcodes-raw-sqlite-db` would require touching the enum **and** the hard-coded dispatch in `BaseDataSet._export_data` (`if NETCDF / elif CSV / else`), **and** `load_from_file` (only recognises `.nc`), **and** `ExportInfo.__post_init__` key validation, **and** it interacts with the `export_type` config + `set/get_data_export_type`. That's *more* invasive than one dynamic column, with a larger blast radius on the export/config surface.

3. **`export_info` is user-facing metadata.** It is stored under the `export_info` tag in the `runs` table and appears in `DataSet.metadata`. In the latest review round we deliberately moved `raw_data_db_path` **out** of `metadata` (into a dedicated column read via `get/set_raw_data_db_path_for_run`) precisely to stop it leaking into `the_same_dataset_as` comparisons and into extracted/exported databases. Storing the path inside `export_info` would put it right back into user-facing metadata and reintroduce those exact bugs.

4. **It doesn't even save a schema column.** `raw_data_db_path` is a *dynamic* column, exactly like `export_info` — so there's no schema-cleanliness win to be had.

Net: the dedicated, metadata-excluded column is the cleaner model because raw-sqlite storage is a live backend, not an export. If we ever want a *unified* "where does this dataset's data live" registry, I'd argue for a small purpose-built abstraction rather than overloading `export_info`.

astafan8:
**Re: should this live in `DataSet`, or a `DataSetInSeparateSqliteDbFile` / `DataSetInSeparateFile` class?**

I agree the config-flag-in-`DataSet` approach is the weakest part architecturally — split behaviour is currently ~15 `is_raw_data_storage_enabled()` touch-points sprinkled through `DataSet`. I'm going to implement the **subclass** option in a separate stacked PR so you can review it as an isolated diff.

Design:

- Introduce a small set of **overridable hooks** on `DataSet`, whose base implementations reproduce today's main-DB behaviour:
  - `_creates_main_results_table` (bool) — base `True`.
  - `_data_conn` — base returns `self.conn`.
  - `_setup_results_backend_on_load()` / `_setup_results_backend_on_new_run()` / `_setup_results_backend_on_start(paramspecs)` — base no-ops.
  - `_augment_background_write_item(item)` — base no-op.
  The generic parts already added for the no-empty-table work (`_results_table_exists`, tolerant `number_of_results`/`__len__`, deferred subscribers) stay in the base — they are backend-agnostic.

- Add **`DataSetInSeparateSqliteDbFile(DataSet)`** that overrides only those hooks (create/connect the per-dataset SQLite file, route `_data_conn`, record `raw_data_db_path`). All the `is_raw_data_storage_enabled()` conditionals then leave `DataSet`.

- Move the **class-selection** to the factory boundary:
  - `new_data_set()` and `Measurement` instantiate the subclass when the feature is enabled.
  - `_get_datasetprotocol_from_guid` instantiates the subclass when the run has a `raw_data_db_path` (this is the disambiguation we already added).

This keeps `DataSet` backend-agnostic and gives us the natural extension point for the **future non-SQLite backends** you mentioned: a later `DataSetInSeparateFile` base could sit between `DataSet` and `DataSetInSeparateSqliteDbFile`, with sibling implementations (parquet/zarr/…). I'll start with the concrete SQLite subclass and leave that generalisation as a clean follow-up rather than speculatively abstracting now.

I'll open the stacked PR against this branch shortly.

src/qcodes/dataset/sqlite/queries.py
jenshnielsen: @astafan8 That change never seems to have made it to this branch

PR #8266, opened 2026-07-03
Add get_db_overview for fast, lightweight run listing

astafan8:
## Summary

Adds `get_db_overview`, a fast, lightweight way to list the runs in a QCoDeS
database. It issues a single `JOIN` query against the `runs` and `experiments`
tables to collect run metadata (experiment/sample names, timestamps, record
counts, guids) **without instantiating a `DataSet` object per run**.

This avoids the expensive `experiments()` + `data_sets()` enumeration, so
listing a database with many thousands of runs becomes near-instant instead of
taking minutes. It is intended for tools that display a table of runs (dataset
browsers, e.g. plottr's `inspectr`).

## What's included

- New module `qcodes/dataset/sqlite/db_overview.py` with:
  - `get_db_overview(path_to_db=None, conn=None, start_run_id=0, extra_columns=None)`
  - `RunOverviewDict` — the `TypedDict` describing each run's overview entry
- Both are exported on the public `qcodes.dataset` namespace (added to
  `qcodes/dataset/__init__.py` `__all__`).
- Tests in `tests/dataset/test_db_overview.py`.

## Notes / design choices

- `records` is a best-effort data-point count: for completed runs the run
  description `shapes` are preferred (authoritative final count), falling back
  to the results-table row count, then `result_counter`. For in-progress runs
  the live results-table row count is preferred.
- `start_run_id` supports cheap incremental refresh (only fetch runs newer than
  a known `run_id`).
- `extra_columns` lets callers pull ad-hoc metadata columns added via
  `DataSet.add_metadata` (e.g. plottr's `inspectr_tag`) into the overview.
  Columns that don't exist in the `runs` table are silently skipped.
- Either `path_to_db` (opened read-only, then closed) or an existing `conn`
  (left open) may be supplied.

## Origin

This was prototyped and battle-tested in
[plottr](https://github.com/toolsforexperiments/plottr) and is being upstreamed
so it can be shared. A companion plottr PR switches to this implementation when
available and keeps a local fallback for older QCoDeS versions.

## Status

**Draft** — opening for early feedback on the public API shape (function name,
argument names, and the `RunOverviewDict` key names). Happy to rename
`experiment`/`sample` keys or adjust the `records` heuristic based on review.

---

> [!NOTE]
> This pull request was created by an agent on behalf of @astafan8.

### DIFF
--- a/docs/changes/newsfragments/8266.new
+++ b/docs/changes/newsfragments/8266.new
@@ -4,4 +4,9 @@ record counts and guids) via a single ``JOIN`` query on the ``runs`` and
 ``experiments`` tables, without instantiating a ``DataSet`` object per run,
 making it possible to list databases with many thousands of runs almost
 instantly. The returned :class:`qcodes.dataset.RunOverviewDict` is also exported
-on the public ``qcodes.dataset`` namespace.
+on the public ``qcodes.dataset`` namespace. The function does not return
+snapshots as they can be large and slow down building the database overview.
+The number of results in each dataset is taken from shape information when
+available, and otherwise falls back to a best-effort count of the rows in the
+so-called results table that exists for every dataset, which may NOT be as
+precise as ``DataSet.number_of_results``.
--- a/docs/changes/newsfragments/8266.new
+++ b/docs/changes/newsfragments/8266.new
@@ -1,0 +1,7 @@
+Added :func:`qcodes.dataset.get_db_overview`, a fast way to list the runs in a
+QCoDeS database. It fetches run metadata (experiment/sample names, timestamps,
+record counts and guids) via a single ``JOIN`` query on the ``runs`` and
+``experiments`` tables, without instantiating a ``DataSet`` object per run,
+making it possible to list databases with many thousands of runs almost
+instantly. The returned :class:`qcodes.dataset.RunOverviewDict` is also exported
+on the public ``qcodes.dataset`` namespace.
--- a/src/qcodes/dataset/sqlite/db_overview.py
+++ b/src/qcodes/dataset/sqlite/db_overview.py
@@ -1,0 +1,253 @@
+"""
+This module provides a fast, lightweight overview of the runs stored in a
+QCoDeS database.
+
+:func:`get_db_overview` issues a single ``JOIN`` query against the ``runs`` and
+``experiments`` tables to collect run metadata (experiment/sample names, time
+stamps, record counts, guids, ...) without instantiating a ``DataSet`` object
+per run. This avoids the expensive ``experiments()`` + ``data_sets()``
+enumeration and makes it possible to list the contents of databases with many
+thousands of runs almost instantly. It is primarily intended for tools that
+need to display a table of runs (e.g. dataset browsers).
+
+The queries rely on the stable QCoDeS database schema (the ``runs`` and
+``experiments`` tables) which has been unchanged across many QCoDeS versions.
+"""
+
+from __future__ import annotations
+
+import datetime
+import json
+import logging
+from contextlib import closing, nullcontext
+from typing import TYPE_CHECKING
+
+from typing_extensions import TypedDict
+
+from qcodes.dataset.sqlite.database import conn_from_dbpath_or_conn
+from qcodes.dataset.sqlite.query_helpers import is_column_in_table
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+    from pathlib import Path
+
+    from qcodes.dataset.sqlite.connection import AtomicConnection
+
+log = logging.getLogger(__name__)
+
+
+class RunOverviewDict(TypedDict):
+    """
+    Lightweight overview of a single run.
+
+    Contains only cheap-to-query metadata: no snapshot, no data and no full
+    ``DataSet`` object. Extra ad-hoc metadata columns requested via the
+    ``extra_columns`` argument of :func:`get_db_overview` are added to the
+    dictionary under their column name in addition to the keys documented here.
+    """
+
+    #: ``run_id`` of the run.
+    run_id: int
+    #: Name of the experiment the run belongs to.
+    experiment: str
+    #: Sample name of the experiment the run belongs to.
+    sample: str
+    #: Name of the run.
+    name: str
+    #: Local date the run was started, formatted as ``YYYY-MM-DD`` (empty
+    #: string if unknown).
+    started_date: str
+    #: Local time the run was started, formatted as ``HH:MM:SS`` (empty string
+    #: if unknown).
+    started_time: str
+    #: Local date the run was completed, formatted as ``YYYY-MM-DD`` (empty
+    #: string if the run has not completed).
+    completed_date: str
+    #: Local time the run was completed, formatted as ``HH:MM:SS`` (empty
+    #: string if the run has not completed).
+    completed_time: str
+    #: Best-effort number of data points in the run, see :func:`get_db_overview`.
+    records: int
+    #: guid of the run.
+    guid: str
+
+
+def _format_timestamp(ts: float | None) -> tuple[str, str]:
+    """
+    Convert a unix timestamp into ``(date, time)`` strings in local time.
+
+    Returns a pair of empty strings if the timestamp is missing or invalid.
+    """
+    if ts is None or ts == 0:
+        return "", ""
+    try:
+        dt = datetime.datetime.fromtimestamp(ts)
+    except (OSError, ValueError, OverflowError):
+        return "", ""
+    return dt.strftime("%Y-%m-%d"), dt.strftime("%H:%M:%S")
+
+
+def _records_from_run_description(run_description_json: str | None) -> int:
+    """
+    Extract a data-point count from the ``shapes`` field of a run description.
+
+    A QCoDeS run description may contain a ``shapes`` mapping from dependent
+    parameter names to their shape tuples. The count returned here is the sum
+    over all dependent parameters of the product of their shape dimensions.
+    Returns ``0`` if the run description is missing, cannot be parsed or does
+    not contain shape information.
+    """
+    if not run_description_json:
+        return 0
+    try:
+        desc = json.loads(run_description_json)
+    except (json.JSONDecodeError, TypeError):
+        return 0
+    shapes = desc.get("shapes") if isinstance(desc, dict) else None
+    if not shapes:
+        return 0
+    total = 0
+    for shape in shapes.values():
+        if isinstance(shape, (list, tuple)) and len(shape) > 0:
+            n = 1
+            for dim in shape:
+                n *= dim
+            total += n
+    return total
+
+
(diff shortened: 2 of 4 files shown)


astafan8:
Companion plottr PR that consumes this API (with a local fallback for older QCoDeS): toolsforexperiments/plottr#462

src/qcodes/dataset/sqlite/db_overview.py
```diff
@@ -0,0 +1,253 @@
...
+:func:`get_db_overview` issues a single ``JOIN`` query against the ``runs`` and
+``experiments`` tables to collect run metadata (experiment/sample names, time
+stamps, record counts, guids, ...) without instantiating a ``DataSet`` object
+per run. This avoids the expensive ``experiments()`` + ``data_sets()``
+enumeration and makes it possible to list the contents of databases with many
+thousands of runs almost instantly. It is primarily intended for tools that
+need to display a table of runs (e.g. dataset browsers).
+
+The queries rely on the stable QCoDeS database schema (the ``runs`` and
+``experiments`` tables) which has been unchanged across many QCoDeS versions.
```
jenshnielsen: This comment feels like a left over from plottr and can probably be removed

src/qcodes/dataset/sqlite/db_overview.py
```diff
@@ -0,0 +1,253 @@
...
+                "completed_date": completed_date,
+                "completed_time": completed_time,
+                "records": records,
+                "guid": row[7] or "",
+            }
+            if valid_extra_columns:
+                extra = {
+                    col: row[n_fixed + i] for i, col in enumerate(valid_extra_columns)
+                }
+                entry.update(extra)  # type: ignore[typeddict-item]
```
jenshnielsen, line 256: document why ignore is required

src/qcodes/dataset/sqlite/db_overview.py
```diff
@@ -0,0 +1,253 @@
...
+        if isinstance(shape, (list, tuple)) and len(shape) > 0:
+            n = 1
+            for dim in shape:
+                n *= dim
+            total += n
+    return total
+
+
+def get_db_overview(
+    path_to_db: str | Path | None = None,
```
jenshnielsen, line 118:
```suggestion
    path_to_db: str | Path | None = None,
    *,
```
I think for clarity all of these should be kwargs onyl

src/qcodes/dataset/sqlite/db_overview.py
```diff
@@ -0,0 +1,253 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+    from pathlib import Path
+
+    from qcodes.dataset.sqlite.connection import AtomicConnection
+
+log = logging.getLogger(__name__)
+
+
+class RunOverviewDict(TypedDict):
```
jenshnielsen, line 37: Not a super strong preference but why not a dataclass?

src/qcodes/dataset/sqlite/db_overview.py
```diff
@@ -0,0 +1,253 @@
...
+            if is_completed:
+                records = _records_from_run_description(row[9])
+                if records == 0:
+                    records = row_counts.get(result_table, 0)
+            else:
+                records = row_counts.get(result_table, 0)
+                if records == 0:
+                    records = _records_from_run_description(row[9])
+            if records == 0:
+                records = row[6] or 0
```
jenshnielsen: Looks like tests covering these branches are missing 

Image

src/qcodes/dataset/sqlite/db_overview.py
```diff
@@ -0,0 +1,253 @@
...
+        # ``result_counter`` in the runs table counts INSERT calls, not data
+        # points. For the ``array`` paramtype a single INSERT can contain many
+        # data points, so the real row count of the results table is queried
+        # separately.
+        result_tables = {row[8] for row in rows if row[8]}
+        row_counts: dict[str, int] = {}
+        for table in result_tables:
+            try:
+                count = c.execute(f'SELECT COUNT(*) FROM "{table}"').fetchone()
+            except Exception:
```
jenshnielsen: Should this only capture a more specific exception

src/qcodes/dataset/sqlite/db_overview.py
astafan8: Good catch - removed that paragraph (it was justifying the schema assumption to plottr's external readers, unnecessary inside QCoDeS).

src/qcodes/dataset/sqlite/db_overview.py
astafan8: Done - added a comment explaining the keys come from the runtime `extra_columns` argument, so they can't be part of the closed `RunOverviewDict`.

src/qcodes/dataset/sqlite/db_overview.py
astafan8: Agreed - made `conn`, `start_run_id` and `extra_columns` keyword-only.

src/qcodes/dataset/sqlite/db_overview.py
astafan8: Done - now catches `sqlite3.Error` around both the overview JOIN and the per-table `COUNT(*)`.

src/qcodes/dataset/sqlite/db_overview.py
astafan8: Added tests for the previously-uncovered branches: an in-progress run (live results-table row count), a run whose results table is missing, and a query error (returns an empty overview). Writing these surfaced a bug: the old `result_counter` fallback is actually the run's ordinal within its experiment, not a data-point count (run 2 -> counter 2), so it produced a misleading number. I removed that fallback entirely - the count is now shapes (completed) / live row count (in progress), and 0 when genuinely unknown.

src/qcodes/dataset/sqlite/db_overview.py
astafan8: I kept it a `TypedDict` on purpose: entries are consumed as plain dict rows (e.g. straight into `pandas.DataFrame.from_dict(overview, orient='index')` by dataset browsers), and the `extra_columns` feature adds dynamically-named keys that a fixed dataclass can't represent. A `TypedDict` documents the guaranteed keys while still allowing those runtime extras. Happy to switch if you'd prefer a dataclass + a separate `extra` mapping, though.

docs/changes/newsfragments/8266.new
```diff
@@ -0,0 +1,7 @@
+Added :func:`qcodes.dataset.get_db_overview`, a fast way to list the runs in a
+QCoDeS database. It fetches run metadata (experiment/sample names, timestamps,
+record counts and guids) via a single ``JOIN`` query on the ``runs`` and
+``experiments`` tables, without instantiating a ``DataSet`` object per run,
+making it possible to list databases with many thousands of runs almost
+instantly. The returned :class:`qcodes.dataset.RunOverviewDict` is also exported
+on the public ``qcodes.dataset`` namespace.
```
astafan8: these details are important to mention. also in the documentation, so please add this (and fix any typos that i might have made)
```suggestion
on the public ``qcodes.dataset`` namespace. The function does not return snapshots
as they can be large and slow down making the database overview. Counts of results in
each dataset is taken from shape information when available, and falls back to best-effort
counting of the rows in the so-called results table that exists for every dataset, and
may NOT be as precise as ``DataSet.number_of_results``.
```

tests/dataset/test_db_overview.py
```diff
@@ -0,0 +1,243 @@
...
+def _make_db_with_runs(
+    db_path: str,
+    n_runs: int = 1,
+    n_points: int = 10,
+    experiment_name: str = "test_exp",
+    sample_name: str = "test_sample",
+) -> None:
+    """Create a database at ``db_path`` with ``n_runs`` simple numeric runs."""
+    initialise_or_create_database_at(db_path)
+    load_or_create_experiment(experiment_name, sample_name=sample_name)
```
astafan8: look at other qcodes tests and make sure that here we create a database file in such a way that does not change qcodes config (which initialise_or_create_database_at does). this means that load_or_create_experiment needs to take an explicit connection object to the database.

tests/dataset/test_db_overview.py
```diff
@@ -0,0 +1,243 @@
...
+    interdeps = InterDependencies_(dependencies={p_y: (p_x,)})
+
+    for r in range(n_runs):
+        ds = new_data_set(f"run_{r + 1}")
+        ds.set_interdependencies(interdeps)
+        ds.mark_started()
+        for i in range(n_points):
+            ds.add_results([{p_x.name: float(i), p_y.name: float(i**2)}])
+        ds.mark_completed()
+        ds.conn.close()
```
astafan8: do this with request.addfinalzier (or a simialr mechanism) so that if the test fails, the connections are closed and cleanup correctly.

tests/dataset/test_db_overview.py
```diff
@@ -0,0 +1,243 @@
...
+    conn = sqlite3.connect(db_path)
+    try:
+        for run_id, info in overview.items():
+            (table_name,) = conn.execute(
+                "SELECT result_table_name FROM runs WHERE run_id=?", (run_id,)
+            ).fetchone()
+            (actual,) = conn.execute(f'SELECT COUNT(*) FROM "{table_name}"').fetchone()
+            assert info["records"] == actual == 10
+    finally:
+        conn.close()
```
astafan8: request.addfinalizer is preferred for these cases

docs/changes/newsfragments/8266.new
astafan8: Applied the suggestion in the newsfragment (with a small grammar fix) and also added these caveats to the `get_db_overview` docstring: snapshots are deliberately not read (they can be large and slow), and the record count is best-effort and may be less precise than `DataSet.number_of_results`.

tests/dataset/test_db_overview.py
astafan8: Done. The tests now build the database through an explicit connection object via a `db_conn` fixture and use `load_or_create_experiment(conn=...)` / `new_data_set(exp_id=..., conn=...)`, so the global qcodes config (`db_location`) is no longer mutated by `initialise_or_create_database_at`.

tests/dataset/test_db_overview.py
astafan8: Done - the `db_conn` fixture registers `request.addfinalizer(conn.close)` so the connection is closed even if the test fails, and the datasets reuse that same connection so there are no other connections to manage.

tests/dataset/test_db_overview.py
astafan8: Done - these tests now reuse the `db_conn` fixture (whose connection is closed via `request.addfinalizer`) instead of opening and closing their own connection in a `finally` block.

jenshnielsen (approved):

PR #8355, opened 2026-07-30
fix(timezone): Make datetime usage timezone-aware for DTZ lint rules

jenshnielsen:
- Use datetime.now(timezone.utc).astimezone() for local time with UTC offset
- Update all datetime.fromtimestamp() calls with tz parameter
- Store parameter cache timestamps as timezone-aware local time
- Parameter snapshot 'ts' field now includes UTC offset (e.g. '2024-01-01 13:00:00+01:00')
- Fix buggy manual timezone computation in AWG70000A driver
- Update all test files to use timezone-aware datetimes
- Update snapshot parsing to use fromisoformat() for ISO 8601 format

Fixes DTZ005, DTZ006, DTZ007 ruff lint violations. Existing datasets remain unaffected; only new ones have explicit timezone info.


src/qcodes/instrument_drivers/tektronix/AWG70000A.py
```diff
@@ -1485,17 +1478,10 @@ def _makeSMLFile(
...
-            tz_h *= -1
-        else:
-            signstr = "+"
-        timestr = dt.datetime.now().strftime("%Y-%m-%dT%H:%M:%S.%f")[:-3]
-        timestr += signstr
-        timestr += f"{tz_h:02.0f}:{tz_m:02.0f}"
+        now = dt.datetime.now(dt.UTC).astimezone()
+        tz_str = now.strftime("%z")  # e.g. "+0100"
+        timestr = now.strftime("%Y-%m-%dT%H:%M:%S.%f")[:-3]
+        timestr += tz_str[:3] + ":" + tz_str[3:]  # "+01:00"
```
jenshnielsen, line 1484: This changes the implementation in 2 ways.
time.timezone reports `The offset of the local (non-DST) timezone, in seconds west of UTC (negative in most of Western Europe, positive in the US, zero in the UK). See note below.`

E.g. This means that currently this reports -3600 in Denmark where which is transformed to `-01:00` but normally we would expect the timezone to be positive in Denmark. Furthermore, it does not take daylight saving into account. 

E.g. in Danish summer time this would change the timezone str from `-01:00` to `+02:00`

tests/dataset/test_links.py
```diff
@@ -27,7 +27,9 @@ def _timestamp() -> int:
         Return a random timestamp that is approximately
         one day in the past.
         """
-        timestamp = datetime.now() - timedelta(days=1, seconds=random.randint(1, 1000))
+        timestamp = datetime.now(UTC) - timedelta(
```
jenshnielsen, line 30: Should this not be in the local timezone rather than utc

astafan8 (approved):

PR #8426, opened 2026-08-19
Refresh package-lock

jenshnielsen:


jenshnielsen:
No effect seems to be https://github.com/dependabot/dependabot-core/issues/15721

PR #8440, opened 2026-08-25
Station YAML: set parameters of channels and submodules in nested syntax

astafan8:
This PR adds the ability to configure parameters on channel lists/tuples and submodules directly from the station (YAML) configuration file.

## Motivation

In QCoDeS, setting a parameter on a channel list/tuple applies it to every channel, e.g. `instrument.channels.myparam(27)`. This PR brings the same convenience to the station YAML configuration, and also makes it possible to configure parameters on any submodule of an instrument.

## Usage

In addition to the reserved keys (`type`, `init`, `address`, `port`, `enable_forced_reconnect`, `parameters`, `add_parameters`), an instrument section may now contain keys naming a submodule or channel list/tuple of the instrument. The value is a mapping of parameter names to their **settings**, using the **same options-mapping form as the existing `parameters` section** (e.g. `initial_value`, `label`, `limits`, ...). These mappings can be nested to reach submodules of submodules:

```yaml
instruments:
  instr1:
    type: ...
    mychannels:            # channel list/tuple or submodule
      myparam:
        initial_value: 27
    subbar:                # submodule
      myparam:
        initial_value: 27
```

Configuring a parameter of a channel list/tuple applies it to every channel in the container. The nested form above is equivalent to using a dotted path in the `parameters` section:

```yaml
instruments:
  instr1:
    type: ...
    parameters:
      mychannels.myparam:
        initial_value: 27
```

Note: a bare scalar value (e.g. `myparam: 27`) is **not** accepted — parameter settings must always be an options mapping, keeping the YAML schema consistent across the `parameters` section and the nested blocks.

## Details

- New recursive resolution in `Station.load_instrument` (`set_component_parameters_from_dict`) that reuses the existing `setup_parameter_from_dict`, so parameters configured via the nested form behave exactly like the `parameters` section. Clear `RuntimeError` messages are raised for unresolvable names and for values whose shape does not match the resolved attribute (a mapping is required for both parameters and submodules).
- The station JSON schema now shares a single `parameter_settings` definition between the `parameters` section and the nested component blocks (`component_settings`). Bare scalar values are rejected early at schema-validation time, everywhere.
- Tests covering the channel-tuple, submodule (with parameter attributes), combined-override, and nested cases; an equivalence test (nested form == dotted-path `parameters` form); and schema-rejection tests for scalar values.
- Documentation added to the Station example notebook and the `load_instrument` docstring (both the dotted `parameters` form and the hierarchical nested form).
- Newsfragment included.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


src/qcodes/station.py
```diff
@@ -698,6 +745,64 @@ def add_parameter_from_dict(
...
+                child_path = f"{path}.{name}"
+                try:
+                    attribute = getattr(component, name)
+                except AttributeError:
+                    raise RuntimeError(
+                        f"Cannot resolve `{name}` in `{child_path}`: "
+                        f"{component!r} has no parameter, submodule or "
+                        f"channel list/tuple with that name."
+                    ) from None
+                if isinstance(attribute, (InstrumentBase, ChannelTuple)):
```
astafan8: isn't there "submodule" subclass of InstrumentBase that we can use here? and just to check, ChannelList is a child class of ChannelTuple, right? hence there using tuple is enough?

src/qcodes/station.py
```diff
@@ -718,6 +823,13 @@ def update_monitor() -> None:
             if isinstance(local_instr, ChannelTuple):
                 raise RuntimeError("A parameter cannot be added to an ChannelTuple")
             add_parameter_from_dict(local_instr, parts[-1], options)
+        component_settings = {
+            name: settings
+            for name, settings in instr_cfg.items()
+            if name not in INSTRUMENT_CONFIG_RESERVED_KEYS
+        }
+        if component_settings:
+            set_component_parameters_from_dict(instr, component_settings, identifier)
```
astafan8, line 839: it is importatnt to mention in the docstrings and qcodes documentation page that the order of applying these settings is "parameters:" first and the component/submodule/channel parts after that. and let's also add a test for this.

src/qcodes/station.py
astafan8: Good call - switched to the more precise `InstrumentModule` (the submodule subclass of `InstrumentBase`). And yes: `ChannelList` is a subclass of `ChannelTuple` (verified: `issubclass(ChannelList, ChannelTuple) == True`), so `isinstance(attr, (InstrumentModule, ChannelTuple))` covers channel lists too. The `submodules` dict is even typed `dict[str, InstrumentModule | ChannelTuple]`, so this matches that exactly. Done in 2873b1c.

src/qcodes/station.py
astafan8: Documented in both the `load_instrument` docstring and the Station example notebook: settings are applied `parameters` first, then `add_parameters`, then the nested submodule/channel list/tuple settings, so the nested settings win on conflict (applied last). Added `test_nested_config_applied_after_parameters_section` which sets the same parameter via `parameters` and via the nested block and asserts the nested value wins. Done in 2873b1c.

src/qcodes/station.py
```diff
@@ -74,6 +75,24 @@
...
+INSTRUMENT_CONFIG_RESERVED_KEYS = frozenset(
+    {
+        "type",
+        "driver",
+        "address",
+        "port",
+        "enable_forced_reconnect",
+        "init",
+        "parameters",
+        "add_parameters",
```
jenshnielsen, line 92: What about submodules, functions add_submodule and add_function do they need to be marker reserved?

jenshnielsen (approved):

src/qcodes/station.py
astafan8: they are not supported in the sense that we don't allow adding submodules and functions in the yaml file, they don't needto be reserved (yet)

PR #8445, opened 2026-08-25
Fix xarray inferred-from data ordering

asmull:
Dataset export to xarray currently fails under the condition.

This can occur when array-valued parameters. Users may wish to acquire and save data in a particular order, and no assumptions.

This failure occurs in two cases:
1. With pandas-based exporting to xarray, indexing is not propagated to inferred-from relationships.
2. With direct xarray exporting, the assumption is made that shape metadata can be used to imply that, e.g., the first column of an NxM sweep can be used to infer the axis of one of the sweep values. This contract is not made explicit in the API, and should not be relied upon. If data for an NxM sweep is created in a permuted order, this MR fixes the direct export to directly extract the axis information from the larger multi-dimensional array. Backwards compatibility of ordering is maintained for non-permuted cases.

### DIFF
--- a/src/qcodes/dataset/exporters/export_to_xarray.py
+++ b/src/qcodes/dataset/exporters/export_to_xarray.py
@@ -84,6 +84,17 @@ def _add_inferred_data_vars(
     dep_names = {dep.name for dep in deps}
     dims = tuple(d for d in xr_dataset.dims)
 
+    # ``Series.to_xarray()`` expands the index into one dimension per index
+    # level. That is only compatible with the target dataset if the dataset
+    # dimensions are exactly the levels of the index. It is not the case when
+    # the dataset uses a single ``multi_index`` dimension or the flat index
+    # created by ``DataFrame.reset_index()``. A non unique MultiIndex cannot be
+    # converted at all. In those cases the data is already in index order so it
+    # can be reshaped directly.
+    index_is_expandable = (
+        index is not None and index.is_unique and set(dims) == set(index.names)
+    )
+
     for inf in inferred:
         if inf.name in dep_names:
             continue
@@ -108,7 +119,7 @@ def _add_inferred_data_vars(
         expected_shape = tuple(xr_dataset.sizes[d] for d in dims)
         expected_size = prod(expected_shape)
         if flat.shape[0] == expected_size:
-            if index is not None:
+            if index is not None and index_is_expandable:
                 # If an index is provided, we should align the inferred data with the index.
                 # This is necessary because data may be reordered when transforming from a pandas DataFrame to an xarray Dataset.
                 # Passing an index allows the original data ordering to be preserved on reconstruction.
--- a/src/qcodes/dataset/exporters/export_to_xarray.py
+++ b/src/qcodes/dataset/exporters/export_to_xarray.py
@@ -67,6 +67,7 @@ def _add_inferred_data_vars(
     name: str,
     sub_dict: Mapping[str, npt.NDArray],
     xr_dataset: xr.Dataset,
+    index: pd.Index | pd.MultiIndex | None,
 ) -> xr.Dataset:
     """Add inferred parameters as data variables to an xarray dataset.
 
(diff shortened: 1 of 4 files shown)


asmull:
@microsoft-github-policy-service agree [company="Microsoft"]

asmull:

src/qcodes/dataset/exporters/export_to_xarray.py
```diff
@@ -67,13 +67,15 @@ def _add_inferred_data_vars(
     name: str,
     sub_dict: Mapping[str, npt.NDArray],
     xr_dataset: xr.Dataset,
+    index: pd.Index | pd.MultiIndex | None,
```
jenshnielsen, line 70: Is there a reason to not use a default value of None and avoid modifying the tests that explicitly pass None

src/qcodes/dataset/exporters/export_to_xarray.py
```diff
@@ -234,64 +243,119 @@ def _xarray_data_set_from_pandas_multi_index(
...
     data_vars: dict[str, tuple[tuple[str, ...], npt.NDArray]] = {
-        name: (tuple(dep_axis.keys()), sub_dict[name])
+        name: (tuple(dep_axis), reordered_data[name]),
+        **extra_data_vars,
     }
-    data_vars.update(extra_data_vars)
 
-    ds = xr.Dataset(data_vars, coords=coords)
-    return ds
+    return xr.Dataset(data_vars, coords=coords)
```
jenshnielsen, line 358: I think it would be good to extend the docs in https://microsoft.github.io/Qcodes/examples/DataSet/Working-With-Pandas-and-XArray.html to explain the various ways exports is handled including what is added here

src/qcodes/dataset/exporters/export_to_xarray.py
```diff
@@ -106,7 +108,14 @@ def _add_inferred_data_vars(
...
         if flat.shape[0] == expected_size:
-            xr_dataset[inf.name] = (dims, flat.reshape(expected_shape))
+            if index is not None:
+                # If an index is provided, we should align the inferred data with the index.
+                # This is necessary because data may be reordered when transforming from a pandas DataFrame to an xarray Dataset.
+                # Passing an index allows the original data ordering to be preserved on reconstruction.
+                indexed_data = Series(flat, index=index, name=inf.name).to_xarray()
+                xr_dataset[inf.name] = indexed_data
+            else:
+                xr_dataset[inf.name] = (dims, flat.reshape(expected_shape))
```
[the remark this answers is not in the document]
jenshnielsen: Tying to figure out if this issue can be triggered in real code I ended up having copilot generate https://github.com/microsoft/Qcodes/pull/8454

src/qcodes/dataset/exporters/export_to_xarray.py
```diff
@@ -234,64 +243,119 @@ def _xarray_data_set_from_pandas_multi_index(
...
-    for axis, dep in enumerate(deps):
-        dep_array = sub_dict[dep.name]
-        dep_axis[dep.name] = dep_array[
-            tuple(slice(None) if i == axis else 0 for i in range(dep_array.ndim))
-        ]
+    destination = np.zeros(expected_size, dtype=np.intp)
+
+    for dimension, dep in enumerate(deps):
+        dep_data = sub_dict[dep.name].ravel()
+
```
jenshnielsen, line 270: Do we have any idea of the performance implications for the default case where data is already ordered?

jenshnielsen (approved):

src/qcodes/dataset/exporters/export_to_xarray.py
jenshnielsen: In general my preference would be to extend the metadata such that a dond over lin (or log ) sweeps can be explicitly annotated with information that this a "real" grid but that does not have to happen in this pr

src/qcodes/dataset/exporters/export_to_xarray.py
asmull: Not really, I will fix this one

src/qcodes/dataset/exporters/export_to_xarray.py
asmull: I will follow up on this in another PR

src/qcodes/dataset/exporters/export_to_xarray.py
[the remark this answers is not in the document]
asmull: Good find, merged into this branch

PR #8451, opened 2026-08-26
Improve parameter scaling / offset

### LINKED ISSUE #8450: Scale/offset silently truncate when their length does not match the value
`ParameterBase` converts between a value and its raw value by zipping the value with an iterable `scale`/`offset`:

```python
tuple(val * sub_scale for val, sub_scale in zip(raw_value, scale))
```

in `_scale_raw_value`, `_offset_raw_value`, `_unoffset_value` and `_unscale_value` in `src/qcodes/parameters/parameter_base.py`.

Since `zip` is not strict, a length mismatch is silently truncated. Setting a parameter with `scale = [2, 4]` to `[10, 20, 30]` silently drops the third element rather than telling the user that the scale does not match the value.

We should decide whether to use `zip(..., strict=True)` (or an explicit length check raising a clear `ValueError`) so that a mismatch is reported instead of silently producing a shorter value. This is a behaviour change for anyone relying on the truncation, so it needs a breaking-change newsfragment.

Follow up from #8449.

jenshnielsen:
* Scale Sequence similar to numpy arrays
* Error if scale/offset length is shorter than data length rather than truncating. 

Closes #8450

### DIFF
--- a/src/qcodes/parameters/parameter_base.py
+++ b/src/qcodes/parameters/parameter_base.py
@@ -266,11 +266,18 @@ class ParameterBaseKWArgs(
 # generic ``ParameterDataTypeVar`` out of the arithmetic.
 
 
+_CONVERSION_KIND: dict[Callable[[Any, Any], Any], str] = {
+    operator.mul: "scale",
+    operator.truediv: "scale",
+    operator.add: "offset",
+    operator.sub: "offset",
+}
+
+
 def _apply_elementwise(
     value: Any,
     conversion: Iterable[float],
     operation: Callable[[Any, Any], Any],
-    kind: str,
 ) -> tuple[Any, ...]:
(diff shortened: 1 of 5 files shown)


src/qcodes/parameters/parameter_base.py
```diff
@@ -265,11 +266,55 @@ class ParameterBaseKWArgs(
 # generic ``ParameterDataTypeVar`` out of the arithmetic.
 
 
+def _apply_elementwise(
+    value: Any,
+    conversion: Iterable[float],
+    operation: Callable[[Any, Any], Any],
+    kind: str,
```
jenshnielsen: Rather than passing in this kind just match on the operator and generate the message

esbenh4 (approved):

PR #8498, opened 2026-09-14
style: Enable ruff BLE001 (blind-except)

jenshnielsen:
Rename the private module level loggers `_LOG` to `_LOGGER` so that ruff recognises them as loggers, switch several catch-all handlers to `Logger.exception` so that the traceback is logged, narrow the exceptions caught in the Triton driver, chain the exception raised by AWG70000A and explicitly mark the remaining intentional catch-alls with `noqa: BLE001`. The example notebooks deliberately catch broad exceptions to demonstrate that an operation is rejected, so `docs/*` ignores the rule.


src/qcodes/parameters/multi_channel_instrument_parameter.py
```diff
@@ -61,7 +61,7 @@ def set_raw(self, value: ParamRawDataType | Sequence[ParamRawDataType]) -> None:
         try:
             for chan in self._channels:
                 getattr(chan, self._param_name).set(value)
-        except Exception as err:
+        except Exception as err:  # noqa: BLE001 re-raised below unless the fallback below succeeds
```
CinaArjmand, line 64: Exceptions that are re-raised will not be flagged, this #noqa should not be needed. 
https://docs.astral.sh/ruff/rules/blind-except/

src/qcodes/instrument/instrument.py
```diff
@@ -152,7 +152,7 @@ def __del__(self) -> None:
         """Close the instrument and remove its instance record."""
         try:
             self.close()
-        except BaseException:
+        except BaseException:  # noqa: BLE001 __del__ must never raise
```
CinaArjmand: By default, __del__() will ignore exceptions and print a warning to sys.stderr. Is this try/except here intentionally in order to suppress the Pytest warning which can cause test failure due to the config which treats warnings as errors?

src/qcodes/station.py
```diff
@@ -285,7 +285,7 @@ def add_component(
         try:
             if not (isinstance(component, Parameter) and component.snapshot_exclude):
                 component.snapshot(update="All" if update_snapshot else "Never")
-        except Exception:
+        except Exception:  # noqa: BLE001 a failing snapshot must not prevent adding the component
```
CinaArjmand: Explains why registration should continue, but should the failure be silent? Better to log warning "Could not snapshot component during registration; adding it anyway"?

src/qcodes/station.py
```diff
@@ -948,7 +948,7 @@ def instrument_names_from_module(module: ModuleType) -> tuple[str, ...]:
                 ms = inspect.getmembers(
                     importlib.import_module(s.name), inspect.isclass
                 )
-            except Exception:
+            except Exception:  # noqa: BLE001 an uninspectable submodule should not abort the scan
                 ms = []
```
CinaArjmand, line 960: Do we want unexpected failures to log something here?

tests/parameter/test_parameter_on_set_callback.py
```diff
@@ -70,7 +70,7 @@ def cleanup_db_connections():
     for conn in open_connections:
         try:
             conn.close()
-        except Exception:
+        except Exception:  # noqa: BLE001 best effort cleanup, never fail the test
```
CinaArjmand: Why not narrow down to just SQLite errors (sqlite3.Error): https://docs.python.org/3/library/sqlite3.html#exceptions

src/qcodes/instrument/instrument.py
CinaArjmand: https://docs.python.org/3/reference/datamodel.html#object.__del__

src/qcodes/parameters/multi_channel_instrument_parameter.py
jenshnielsen: In this case the lint is not clever enough to find this out. We already configure unnecessary noqa to be an error so if this improves this we will catch it

src/qcodes/instrument/instrument.py
jenshnielsen: I think that you are right but I really do not want to change this code at the moment. I will update the comment.

tests/parameter/test_parameter_on_set_callback.py
jenshnielsen: Fair, I think the idea was just to not error here but lets give it a try

src/qcodes/parameters/multi_channel_instrument_parameter.py
jenshnielsen: I don't like how this code is implemented but I don't want to change that here

CinaArjmand (approved):

PR #8520, opened 2026-09-21
Fix lakeshore336 snapshot bug

SnX59:
The 336's HTRSET has no output_type field, and its voltage outputs 3 & 4 don't support HTRSET at all. Group._separator_parser then mis-zipped parameter names against the shorter instrument response, producing one Snapshot: Could not update parameter warning per heater-setup parameter.

Model 335 and 372 behavior is unchanged. I scoped the fix to the reported 336 issue but perhaps it might be important to update those drivers too!


src/qcodes/instrument_drivers/Lakeshore/Lakeshore_model_336.py
```diff
@@ -96,7 +105,16 @@ def __init__(
         output_index: int,
         **kwargs: "Unpack[InstrumentBaseKWArgs]",
     ):
-        super().__init__(parent, output_name, output_index, has_pid=False, **kwargs)
+        # Outputs 3 and 4 are unpowered analog voltage outputs that do not
+        # support the heater setup (HTRSET) command, so disable it entirely.
```
jenshnielsen, line 109: Does this mean that we disable it for all 4 outputs (1,2,3,4) is that what we want?

src/qcodes/instrument_drivers/Lakeshore/lakeshore_base.py
```diff
@@ -129,61 +142,82 @@ def __init__(
...
-
-        self.output_heater_resistance: GroupParameter = self.add_parameter(
-            name="output_heater_resistance",
-            docstring="Heater Resistance Setting: 25/50ohm",
-            val_mapping={"25ohm": 1, "50ohm": 2},
-            parameter_class=GroupParameter,
-        )
-        """Heater Resistance Setting: 25/50ohm"""
+        self._has_heater_output_setup = has_heater_output_setup
+        if has_heater_output_setup:
```
jenshnielsen, line 146: I would like to investigate it we can add this in a subclass. As it this would make these parameters not observable statically. I am happy to do this

src/qcodes/instrument_drivers/Lakeshore/Lakeshore_model_336.py
SnX59: The has_heater_output_setup / heater_output_has_output_type flags are per‑output‑instance constructor arguments, not a global/class switch. The Model 336 builds its four outputs from two different subclasses:

output_1, output_2 → LakeshoreModel336CurrentSource -> Keep HTRSET without output_type
output_3, output_4 → LakeshoreModel336VoltageSource -> no HTRSET group at all

SnX59:

SnX59:

